PR #217, opened 2019-02-11
Fix build

krwq:
Fix build and enable building tests by default so that we get the failure in the local build as well.

@joperezr another reason to use P2P instead of PkgRef


krwq:
Merging to unblock people @joperezr please do post review, I'll address feedback later.

joperezr:
Lgtm, thanks for fixing @krwq.

PR #133, opened 2019-01-03
Libgpiod pinvoke C Api Read/Write implementation

buyaa-n:
Please note that following event handling methods implementations not completed yet
`protected internal abstract void AddCallbackForPinValueChangedEvent(int pinNumber, PinEventTypes eventType, PinChangeEventHandler callback);`
`protected internal abstract void RemoveCallbackForPinValueChangedEvent(int pinNumber, PinChangeEventHandler callback);`

-- UPDATE : event handlers above mentioned are ready for review
Issue: https://github.com/dotnet/iot/issues/42

PR #219, opened 2019-02-12
Libgpiod driver without event handling

buyaa-n:
Removed event handling logic by PR comment: https://github.com/dotnet/iot/pull/133#issuecomment-462570635


src/System.Device.Gpio.Tests/UnixDriverTests.Linux.cs
```diff
@@ -8,8 +8,8 @@ namespace System.Device.Gpio.Tests
 {
     public class UnixDriverTests : GpioControllerTestBase
     {
-        protected override GpioDriver GetTestDriver() => new UnixDriver();
+        protected override GpioDriver GetTestDriver() => UnixDriver.CreateUnixDriver();
```
krwq: just `Create`

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                throw new IOException($"No any chip available, error code: {Marshal.GetLastWin32Error()}");
+            }
+
+            _chip = Interop.GetNextChipFromChipIterator(iterator);
+            if (_chip == null)
+            {
+                throw new IOException($"No chip found, error code: {Marshal.GetLastWin32Error()}");
+            }
+            // Freeing other chips opened
```
krwq: enter after `}`

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+                {
+                    success = Interop.RequestLineInput(pinHandle, pinNumber.ToString());
+                }
+                else
+                {
+                    success = Interop.RequestLineOutput(pinHandle, pinNumber.ToString());
+                }
+            }
+
+            if (success == -1) {
```
krwq: brace on new line

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+                if (pinHandle != null)
+                {
+                    Interop.ReleaseGpiodLine(pinHandle);
+                }
+            }
+
+            if (_chip != null)
+            {
+                _chip.Dispose();
+            }
```
krwq: enter after `}`

src/Native/System.Device.Gpio.Native/libgpiod.c
```diff
@@ -0,0 +1,174 @@
...
+	return gpiod_line_request_output(line, consumer, 0);
+}
+
+/**
+ * @brief Request all event type notifications on a single line.
+ * @param line GPIO line object.
+ * @param consumer Name of the consumer.
+ * @return 0 if the operation succeeds, -1 on failure.
+ */
+extern int RequestBothEdgeEventForLine(struct gpiod_line *line, const char *consumer)
```
joperezr: NIT: Should we keep the Eventing related c APIs for the eventing review instead?

src/System.Device.Gpio/Interop/Unix/SafeLineHandle.cs
```diff
@@ -0,0 +1,20 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace System.Device.Gpio
+{
+    internal class SafeLineHandle : SafeHandle
```
joperezr: question: Can we think of a Chip to be equivalent to a GpioController and a Line to be a Pin? Is that more or less what this maps to? It would be great to have that info in a comment above the class just for future reference

src/Native/System.Device.Gpio.Native/libgpiod.c
buyaa-n: we can remove that too, maybe that is better for future review

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -0,0 +1,139 @@
...
+   internal static extern void ReleaseGpiodLine(SafeLineHandle lineHandle);
+
+   /// <summary>
+   /// Request all event type notifications on a single line.
+   /// </summary>
+   /// <param name="line">GPIO line handle</param>
+   /// <param name="consumer">Name of the consumer.</param>
+   /// <returns>0 the operation succeeds, -1 on failure.</returns>
+   [DllImport(LibgpiodLibrary, SetLastError = true)]
+   internal static extern int RequestBothEdgeEventForLine(SafeLineHandle line, string consumer);
```
joperezr: Same here, I would probably keep this for the next PR.

src/Native/System.Device.Gpio.Native/libgpiod.c
krwq: i think this is fine

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
```diff
@@ -0,0 +1,562 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading;
+
+namespace System.Device.Gpio.Drivers
+{
+    /// <summary>
+    /// A GPIO driver for Unix.
+    /// </summary>
+    internal class SysFSDriver : UnixDriver
```
joperezr: Make this public instead, so people can choose to get this behavior instead of libgpiod getting picked

src/System.Device.Gpio/Interop/Unix/SafeLineHandle.cs
buyaa-n: Not sure what level of equivalency you mean, yes SafeChipHandle is maps to gpio controller and SafeLineHandle is map to pin, but at hardware level, just pointers to chip/pin. But cannot be equivalent to our Software level GpioController

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -324,7 +324,7 @@ private void InitializeSysFS()
                 {
                     return;
                 }
-                _sysFSDriver = new UnixDriver();
+                _sysFSDriver = UnixDriver.CreateUnixDriver();
```
joperezr: For now, can you change this to always use the SysFSDriver instead? This is used strictly for eventing, which you don't support for now with libgpiod so let's make sure we force sysfs here for now.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+// See the LICENSE file in the project root for more information
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Runtime.InteropServices;
+using System.IO;
+
+namespace System.Device.Gpio.Drivers
+{
+    public class LibGpiodDriver : UnixDriver
```
joperezr: Given that both this and the SysFSDriver are public, you will need to also have to provide  a Windows implementation for them which just throws PNS Exception

src/System.Device.Gpio.Tests/UnixDriverTests.Linux.cs
joperezr: I think we actually want to change this and have two different drivers, where one is for sysfs driver and one for libgpiod driver. The one for libgpiod can simply skip tests in case the library is not installed.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                if (pinHandle != null)
+                {
+                    Interop.ReleaseGpiodLine(pinHandle);
+                }
+            }
+
+            if (_chip != null)
+            {
+                _chip.Dispose();
```
joperezr: NIT: Set to null after so that if disposed gets called twice, then this won't start having wierd outcomes.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle pinHandle))
+            {
+                Interop.SetGpiodLineValue(pinHandle, (value == PinValue.High) ? 1 : 0);
+            }
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            foreach (SafeLineHandle pinHandle in _pinNumberToSafeLineHandle.Values)
```
joperezr: We should be emptying out the dictionary here. In order to do that, you will probably want to use while loop instead where you pop the next item, and release it and remove it from the collection.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Runtime.InteropServices;
+using System.IO;
+
+namespace System.Device.Gpio.Drivers
+{
+    public class LibGpiodDriver : UnixDriver
+    {
+        private SafeChipHandle _chip;
+
+        private Dictionary<int, SafeLineHandle> _pinNumberToSafeLineHandle;
```
joperezr: readonly?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        protected internal override int PinCount => Interop.GetNumberOfLines(_chip);
+
+        public LibGpiodDriver()
+        {
+            SafeChipIteratorHandle iterator = Interop.GetChipIterator();
+            if (iterator == null)
+            {
+                throw new IOException($"No any chip available, error code: {Marshal.GetLastWin32Error()}");
```
joperezr: NIT: `Unable to find a chip iterator, error code: ...`

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
ahsonkhan: Is the casing of this class name (and the class name itself) appropriate? Should it be `SysfsDriver`?

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
buyaa-n: ok

src/Native/System.Device.Gpio.Native/libgpiod.c
```diff
@@ -0,0 +1,174 @@
...
+extern void FreeChipIterator(struct gpiod_chip_iter *iter) {
+	gpiod_chip_iter_free(iter);
+}
+
+/**
+ * @brief Release all resources allocated for the gpiochip iterator but
+ *        don't close the most recently opened gpiochip (if any).
+ * @param iter The gpiochip iterator object.
+ */
+extern void FreeChipIteratorNoCloseCurrentChip(struct gpiod_chip_iter *iter) {
```
ahsonkhan: nit: inconsistent braces (some at a new line, some on the same line)

src/System.Device.Gpio/Interop/Unix/SafeChipHandle.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public SafeChipHandle() : base(IntPtr.Zero, true) { }
+
+        protected override bool ReleaseHandle()
+        {
+            Interop.CloseChip(handle);
+            return true;
+        }
+
+        public override bool IsInvalid => handle == IntPtr.Zero || handle == new IntPtr(-1);
```
ahsonkhan: Do we always have to new up an IntPtr for every call to IsInvalid or can we cache it in a static?

ahsonkhan: Also, when will handle ever `== IntPtr(-1)`?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
joperezr: great question.. according to kernel.org, `SysfsDriver` might be the right casing for this.

src/System.Device.Gpio/Interop/Unix/SafeLineHandle.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using System.Text;
```
ahsonkhan: Do we need all these using directives (particularly System.Text)?

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -0,0 +1,139 @@
...
+using System.Runtime.InteropServices;
+
+internal partial class Interop
+{
+    private const string LibgpiodLibrary = "System.Device.Gpio.Native";
+
+    /// <summary>
+    /// Create a new gpiochip iterator.
+    /// </summary>
+    /// <returns>Pointer to a new chip iterator object or NULL if an error occurred.</returns>
```
ahsonkhan: nit: `... object or <see langword="null" /> if an error ...`

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            _chip = Interop.GetNextChipFromChipIterator(iterator);
+            if (_chip == null)
+            {
+                throw new IOException($"No chip found, error code: {Marshal.GetLastWin32Error()}");
```
ahsonkhan: Consider using a throw helper pattern to reduce the method size bloat that might come from formatting the error message in line.

You can either call a helper which returns the exception you want to throw (which contains the exception message), or defer throwing all together to the helper.

Example [sharplab.io](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxNMAfAAgBgARYCMA3ALABQWAzAQEx4DCeA3hXuwTcAPbcA2eALKEAFEXwA3CHwCucJHgCWAOxh4+cZQHMYACwCULNhxOKAZnhFTZcPAF47eZTL599xk+1blPvvHoRuAHcnOBCAUQAPAGM4AAcYRW5lEQAiPAA5FwFrOTxU/TIfPwBfD08sAHZ/BDkikzLi9nLqPB5+IVoxQklpOQUVNQ1tPUNvPyULKz7bByds9ybPcYmOAODQiJj4xOSRABJ0rNc8XNtmYZ1dEoL60vKTKpq68saTFq5eAUEqbt6bAaqdSaK5jB4ccyWM72RzOVyLCYrVZ4AAqukCQQAkgB5KKxBJJFKXUbEPDg9hvPxPGC1OB3CnlcpxBCKKTwAiEABsBAALKj0cEcXidoSRINgSMDOUkZ51iFlGE8MKCXtDplsqcZixiddbq8KCUgA=)

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+            // Libgpiod Api do not support pull up or pull down resistors for now.
+            return mode != PinMode.InputPullDown && mode != PinMode.InputPullUp;
+        }
+
+        protected internal override void OpenPin(int pinNumber)
+        {
+            SafeLineHandle pinHandle = Interop.GetChipLineByOffset(_chip, pinNumber);
+            if (pinHandle == null)
+            {
+                throw new IOException($"Pin number {pinNumber} not available for chip: {_chip} error: {Marshal.GetLastWin32Error()}");
```
ahsonkhan: Same here, and elsewhere.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
ahsonkhan: @AndyAyersMS, what are your thoughts on using the format string pattern to throw exception messages within the method which contains the happy path?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        protected internal override void RemoveCallbackForPinValueChangedEvent(int pinNumber, PinChangeEventHandler callback)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected internal override void SetPinMode(int pinNumber, PinMode mode)
+        {
+            int success = -1;
```
ahsonkhan: nit: rename to result?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
```diff
@@ -0,0 +1,562 @@
...
+    {
+        private const string GpioBasePath = "/sys/class/gpio";
+        private int _pollFileDescriptor = -1;
+        private Thread _eventDetectionThread;
+        private int _pinsToDetectEventsCount;
+        private readonly static CancellationTokenSource s_eventThreadCancellationTokenSource = new CancellationTokenSource();
+        private readonly List<int> _exportedPins = new List<int>();
+        private readonly Dictionary<int, UnixDriverDevicePin> _devicePins = new Dictionary<int, UnixDriverDevicePin>();
+        private readonly int _pollingTimeoutInMilliseconds = Convert.ToInt32(TimeSpan.FromMilliseconds(1).TotalMilliseconds);
```
ahsonkhan: Isn't this just 1? Why set it with all this ceremony?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
```diff
@@ -0,0 +1,562 @@
...
+        {
+            if (mode == PinMode.Input)
+            {
+                return "in";
+            }
+            if (mode == PinMode.Output)
+            {
+                return "out";
+            }
+            throw new PlatformNotSupportedException($"{mode} is not supported by this driver.");
```
ahsonkhan: Is this necessary? Just use `Debug.Assert` since it looks like we already validate that mode is either `PinMode.Input` or `PinMode.Output` in the caller.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
```diff
@@ -0,0 +1,562 @@
...
+        }
+
+        private PinMode ConvertSysFsModeToPinMode(string sysfsMode)
+        {
+            sysfsMode = sysfsMode.Trim();
```
ahsonkhan: Consider using `ReadOnlySpan<char>` instead to avoid intermediary allocation.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
```diff
@@ -0,0 +1,562 @@
...
+                }
+            }
+            else
+            {
+                throw new InvalidOperationException("There was an attempt to write to a pin that is not open.");
+            }
+        }
+
+        private string ConvertPinValueToSysFs(PinValue value)
+
```
ahsonkhan: nit: remove extra new line

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
```diff
@@ -0,0 +1,562 @@
...
+        /// <summary>
+        /// Checks if a pin supports a specific mode.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        /// <param name="mode">The mode to check.</param>
+        /// <returns>The status if the pin supports the mode.</returns>
+        protected internal override bool IsPinModeSupported(int pinNumber, PinMode mode)
+        {
+            // Unix driver does not support pull up or pull down resistors.
+            if (mode == PinMode.InputPullDown || mode == PinMode.InputPullUp)
```
ahsonkhan: nit: return `mode != PinMode.InputPullDown && mode != PinMode.InputPullUp`

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
buyaa-n: ok

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
```diff
@@ -0,0 +1,562 @@
...
+            }
+
+            // Ignore first time because it will always return the current state.
+            Interop.epoll_wait(pollFileDescriptor, out _, 1, 0);
+        }
+
+        private unsafe bool WasEventDetected(int pollFileDescriptor, int valueFileDescriptor, out int pinNumber, CancellationToken cancellationToken)
+        {
+            char buf;
+            IntPtr bufPtr = new IntPtr(&buf);
```
ahsonkhan: nit: Move this closer to where it's used.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
```diff
@@ -0,0 +1,562 @@
...
+                valueFileDescriptor = -1;
+            }
+            if (closePollFileDescriptor)
+            {
+                if (cancelEventDetectionThread)
+                {
+                    s_eventThreadCancellationTokenSource.Cancel();
+                    while (_eventDetectionThread != null && _eventDetectionThread.IsAlive)
+                    {
+                        Thread.Sleep(TimeSpan.FromMilliseconds(10)); // Wait until the event detection thread is aborted.
```
ahsonkhan: How do we know that 10 ms is sufficient?

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -0,0 +1,139 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+using System;
+using System.Device.Gpio;
+using System.Runtime.InteropServices;
+
+internal partial class Interop
```
ahsonkhan: Are we missing a namespace?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
ahsonkhan: Do we want to seal these new classes?

buyaa-n: By my understanding the API is not intending to expose them, instead decide which driver to load runtime. By your suggestion I made them public as libgpiod driver event handling is not done yet, but as when event handling is implemented we gonna make them internal back I think we don't need to provide Windows implementation

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: Not sure i got this comment correctly, if you mean if this dictionary should be readonly then i don't thinks so, would like to ask why

src/Native/System.Device.Gpio.Native/libgpiod.c
```diff
@@ -0,0 +1,174 @@
...
+ */
+extern unsigned int GetNumberOfLines(struct gpiod_chip *chip) 
+{
+	return gpiod_chip_num_lines(chip);
+}
+
+/**
+ * @brief Get the handle to the GPIO line at given offset.
+ * @param chip The GPIO chip object.
+ * @param offset The offset of the GPIO line.
```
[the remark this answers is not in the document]
buyaa-n: pin number, expressed like offset address

src/Native/System.Device.Gpio.Native/libgpiod.c
```diff
@@ -0,0 +1,174 @@
...
+extern int GetLineDirection(struct gpiod_line *line) 
+{
+	return gpiod_line_direction(line);
+}
+
+/**
+ * @brief Reserve a single line, set the direction to input.
+ * @param line GPIO line object.
+ * @param consumer Name of the consumer.
+ * @return 0 if the line was properly reserved, -1 on failure.
```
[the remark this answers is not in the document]
buyaa-n: always -1 on error, real error code will be set in the system error code

src/Native/System.Device.Gpio.Native/libgpiod.c
```diff
@@ -0,0 +1,174 @@
...
+ * @param line GPIO line object.
+ */
+extern void ReleaseGpiodLine(struct gpiod_line *line) {
+	gpiod_line_release(line);
+}
+
+/**
+ * @brief Read current value of a single GPIO line.
+ * @param line GPIO line object.
+ * @return 0 or 1 if the operation succeeds. On error this routine returns -1
```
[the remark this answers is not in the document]
buyaa-n: binary values can be read from a pin, i.e. high or low

src/Native/System.Device.Gpio.Native/libgpiod.c
```diff
@@ -0,0 +1,174 @@
...
+extern int GetGpiodLineValue(struct gpiod_line *line) 
+{
+	return gpiod_line_get_value(line);
+}
+
+/**
+ * @brief Set the value of a single GPIO line.
+ * @param line GPIO line object.
+ * @param value New value.
+ * @return 0 is the operation succeeds. In case of an error this routine
```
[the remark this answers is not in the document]
buyaa-n: It is comments copied from gpiod.h header of libgpiod C library, for some methods with some modification. By my understanding it is C style documentation

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
buyaa-n: SysfsDriver code is just copied from UnixDriver, not my work

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
buyaa-n: i don't really know how it calculated, anyways SysFs driver code didn't changed, just move to this new class as Unix driver now has 2 options, libgpiod and sysFs. I am gonna ignore sysFs related comments i dont really know or don't want to update

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
buyaa-n: i don't really know why, but all partial classes for Interop are haven't set namespace. myabe @joperezr can explain

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
AndyAyersMS: It can potentially add overhead to the method prolog and epilog, so if that overhead impacts perf of the happy path then you might consider using a throw helper.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
krwq: I this might be the existing code and the file just got renamed though :)

src/System.Device.Gpio/Interop/Unix/SafeLineHandle.cs
krwq: we havent particualrly paid attention to that in this repo. I'd presume we will eventually clean them all up in one sweep

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
krwq: I think it's fine if we follow existing pattern

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
buyaa-n: yeah, its all just copied from UnixDriver, gonna ignore comments on this file

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
krwq: or make it bool and later `success = Interop....(...) != -1`

src/System.Device.Gpio/Interop/Unix/SafeLineHandle.cs
ahsonkhan: Seems like unnecessary tech debt, but ok :)

src/System.Device.Gpio/Interop/Unix/SafeChipHandle.cs
buyaa-n: > Do we always have to new up an IntPtr for every call to IsInvalid or can we cache it in a static?

Moved to static variable

Not sure what about the 2nd question, if you asking when handle could be equal to IntPtr(-1), then I guess when calling native method for chip pointer it might return -1 in some error nous conditions.

src/System.Device.Gpio/Interop/Unix/SafeLineHandle.cs
buyaa-n: removed

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: @ahsonkhan looked into the sample you referenced, seems good approach when error messages use similar string lateral, but here each errornous case messages are different. 

Also not sure about what you mean by happy path, if you mean about normal program execution path then: no it is not happy path, program execution will stop if any of these exceptions thrown

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: @ahsonkhan this variable is for determining if setPinMode call was successful, i think if it was returning result we better rename it into result

@krwq that can be improvment, as main goal is if it was failed throw exception maybe change it bool failed

src/System.Device.Gpio.Tests/UnixDriverTests.Linux.cs
buyaa-n: Updated

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
krwq: @ahsonkhan we currently don't bother with sealing stuff, there is too high chance we will unseal it in the future (i.e. new version of libgpiod shows up with the new kernel)

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
krwq: readonly will only mean that the reference cannot be changed not the actual content. IMO this keyword is kinda useless since you can workaround it with reflection and only adds unnecessary complexity like const correctness in C++

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -100,20 +100,20 @@ protected internal override void RemoveCallbackForPinValueChangedEvent(int pinNu
 
         protected internal override void SetPinMode(int pinNumber, PinMode mode)
         {
-            int success = -1;
+            bool failed  = false;
```
krwq: should this be true by default?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
ahsonkhan: > this variable is for determining if setPinMode call was successful, i think if it was returning result we better rename it into result

I was thinking more towards what is the result of the request (i.e. Interop.RequestLineInput) rather than it being successful or not. Naming it as success or failure implies one or the other when it could be either.

Changing it to a bool unnecessarily introduces more branches in the code. and increases the method body size. **We should consider reverting it back to int, imo.** The rename suggestion was a nit so either one of these is fine: result/failed/success/requestResult/etc.

Looking at a subset of the method code:
[51 instructions (using integer) vs 57 (now with bool) vs 42 (best - cache local + use integer) - sharplab.io](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxNMAfAAgBgARYCMA3ALABQuBhAdAEoCuAdjAJZhw0DCA9mAA6sANogDKiAG6sAxnChlKAZgIAmPFzwBvCnl0FlWACx5xMAAqsmAWR4ATOAH1+COBIAUlmHkFMAcgzBgRCQ8YB4eITwwOzgASi0dPSTPPCgGaVkoKDwAXjwAWhJEpN1WADM8N2j7WOKS7XISptT0zOy8ujgARwY5GAAZSzg3H39A4O9LMaCEGgAVHlEYBEsAczdY2IVmvABfOqS4ISg4A70GnaS0jLl2vE6evsGmYdGAmZC38dmFpZWmdabbbNfaNZpnUoVNzXNq5PKFWpgpoXS66GAACwQPAA7ngXriAKIAD1k/DYPCYG2BTVBSVpejqWAMxlMFmsMQc+I8LEmfneEzCESiMXiKKagsiZQgwjgtlyeClxzg1JK5Uq1TiEISSMuUplco63V6sGerym/OQvOmiHmi2Waw28Ry8KKOrpWqOJy1Yt10pEBvuRqeQxG5u+nzDM1tfwdmzhBVdO3pTS1arcev9iJ2Pp2GKxuPxeGJpPJlK2WuTezqjOZJjg5ksNnsDiCMHgCG5Xi+H1C4UiGtFqZ5LkesE6aSEMBVSSI+G7YntAPl85+dv+gOnejTA+9WqSI+NMHHDEn8oeh9Nob54atFpjAPLbr0lcOSt3T6aB76x9PhtHAxDFcI2vGZ71WR8ky1VMoS/Mc5BPLxnQTLNmhzZo8xxPE4EJEk4DJVgKSpCs6krOoAG0rHrdE7AASQEIQ3EojFaPogB5fCKSgGhfB4GimCESw1liABdOpnFYCQIHgPAUnPYMXk7W8bzA3kwJQ3Q0N0LAAHYlJmPAAGpVMXVYaH6OAAQxFVQV2IA==)

We should also cache `pinNumber.ToString()` in a local since it's common between the branches.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
ahsonkhan: > but here each errornous case messages are different
In corefx, we have used an exception enum to get the appropriate string message within the throw helper.

ahsonkhan: So something like this:
```C#
private static class ThrowHelper
{
   // Add overloads for other exception types or exceptions with other format string data as parameters
   public static void ThrowIOException(ExceptionResource resource, int errorCode /*, ... */)
   {
      throw GetIOException(GetResourceString(resource, errorCode));
   }
   private static IOException GetIOException(string message)
   {
      return new IOException(message);
   }
   private static string GetResourceString(ExceptionResource resource, int errorCode /*, ... */)
   {
      string message = "";
      switch(resource)
      {
         case ExceptionResource.NoChipFound:
            message = $"No chip found, error code: {errorCode}";
            break;
         // ... More cases
      }
      return message;
   }
}
internal enum ExceptionResource
{
   NoChipFound,
}
```

Usage, in this case, would become something like:
```C#
   _chip = Interop.GetNextChipFromChipIterator(iterator);
   if (_chip == null)
   {
      ThrowHelper.ThrowIOException(ExceptionResource.NoChipFound, Marshal.GetLastWin32Error());
```

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
ahsonkhan: Sounds good to me. However, I am wondering why would a new version of libgpiod require us to unseal the class?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: yes it is thanks, my point is this collection content will be changed even the reference to the collection wouldn't change, doesn't make sense to make readonly

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: ok @ahsonkhan will update, thanks

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: Looks good, will update, thanks

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: well i don't think new version of libgpiod would require to unseal, never heard something like require to be unsealed, instead heard required to be sealed. So i would like to ask why we need to seal the class. I saw classes sealed only if they really need to be sealed or required to be sealed

buyaa-n:
@joperezr @ahsonkhan addressed comments please re review

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -22,13 +22,13 @@ public LibGpiodDriver()
...
             {
-                throw new IOException($"Unable to find a chip iterator, error code: {Marshal.GetLastWin32Error()}");
+                ThrowHelper.ThrowIOException(ThrowHelper.ExceptionResource.NoChipIteratorFound, Marshal.GetLastWin32Error());
             }
 
             _chip = Interop.GetNextChipFromChipIterator(iterator);
             if (_chip == null)
             {
-                throw new IOException($"Unable to find a chip, error code: {Marshal.GetLastWin32Error()}");
+                ThrowHelper.ThrowIOException(ThrowHelper.ExceptionResource.NoChipFound, Marshal.GetLastWin32Error());
```
krwq: You should unify all to either use `throw ThrowHelper.GetIOException` or `ThrowHelper.ThrowIOException)` - I'm seeing both patterns in this PR. Btw, better name for the class could be simply `Error`

src/System.Device.Gpio/System/Device/Gpio/ThrowHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+                    break;
+                default:
+                    Debug.Fail($"The ExceptionResource enum value: {resource} is not part of the switch. Add the appropriate case and exception message.");
+                    break;
+            }
+            
+            return message;
+        }
+
+        internal enum ExceptionResource
```
krwq: might be better to make this a static class with const strings, then you don't need the large switch

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -100,20 +100,21 @@ protected internal override void RemoveCallbackForPinValueChangedEvent(int pinNu
 
         protected internal override void SetPinMode(int pinNumber, PinMode mode)
         {
-            bool failed  = false;
+            int requestResult  = -1;
```
krwq: why reverting this?

ahsonkhan: See https://github.com/dotnet/iot/pull/219#discussion_r256269017

src/System.Device.Gpio/System/Device/Gpio/ThrowHelper.cs
ahsonkhan: How would that work? The code path with the switch only occurs on exceptions and it is somewhat easy to extend to add new exceptions. I don't think const strings would work since a lot of the messages are format strings with custom args.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        protected internal override void RemoveCallbackForPinValueChangedEvent(int pinNumber, PinChangeEventHandler callback)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected internal override void SetPinMode(int pinNumber, PinMode mode)
+        {
+            int requestResult  = -1;
+            string consumer = pinNumber.ToString();
```
ahsonkhan: nit: move this local into the if block (closer to where it's used).

```C#
if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle pinHandle))
{
   string consumer = pinNumber.ToString();
   ...
}
```

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        protected internal override WaitForEventResult WaitForEvent(int pinNumber, PinEventTypes eventType, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected internal override void Write(int pinNumber, PinValue value)
+        {
+            if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle pinHandle))
```
ahsonkhan: nit: invert this (which would avoid chances of bugs with missing return statements)

```C#
            if (!_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle pinHandle))
            {
               throw ThrowHelper.GetInvalidOperationException(ThrowHelper.ExceptionResource.PinNotOpenedError, pin: pinNumber);
            }
            Interop.SetGpiodLineValue(pinHandle, (value == PinValue.High) ? 1 : 0);
           
```

src/System.Device.Gpio/System/Device/Gpio/ThrowHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            NoChipIteratorFound,
+            NoChipFound,
+            PinModeReadError,
+            OpenPinError,
+            ReadPinError,
+            PinNotOpenedError,
+            SetPinModeError,
+            ConvertPinNumberingSchemaError
```
ahsonkhan: nit: It would be cleaner if this enum was outside the ThrowHelper static class. That way you can remove the boilerplate `ThrowHelper.ExceptionResource....` everywhere and use it like a regular enum: `ExceptionResource....`

src/System.Device.Gpio/System/Device/Gpio/ThrowHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+            return GetInvalidOperationException(GetResourceString(resource, -1, pin));
+        }
+
+        private static string GetResourceString(ExceptionResource resource, int errorCode, int pin)
+        {
+            string message = "";
+            switch (resource)
+            {
+                case ExceptionResource.NoChipIteratorFound:
+                    message = $"Unable to find a chip iterator, error code: { errorCode}";
```
ahsonkhan: nit: extra space

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: followed ThrowHelper class in corefx which has only ThrowHelper.ThrowIOException(), but it was not good for using for methods need to return a value, so added ThrowHelper.GetIOException(), but yeah better to use only use ThrowHelper.GetIOException().

For the class name though I don't prefer renaming into Error, it is helper class so I would keep word Helper, but as am gonna use ThrowHelper.GetIOException() pattern instead of ThrowHelper.ThrowIOException() pattern, maybe rename it ExceptionHelper or ErrorHelper

src/System.Device.Gpio/System/Device/Gpio/ThrowHelper.cs
buyaa-n: i think its better to be enum for swith case ...

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: yeah @ahsonkhan showed that this is better

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,159 @@
...
+        protected internal override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber) => throw 
+            ExceptionHelper.GetPlatformNotSupportedException(ExceptionResource.ConvertPinNumberingSchemaError);
+
+        protected internal override PinMode GetPinMode(int pinNumber)
+        {
+            if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle pinHandle))
+            {
+                return (Interop.GetLineDirection(pinHandle) == 1) ? PinMode.Input : PinMode.Output;
+            }
+            throw ExceptionHelper.GetInvalidOperationException(ExceptionResource.PinNotOpenedError, pin:pinNumber);
```
ahsonkhan: nit: spacing `pin: pinNumber`

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,159 @@
...
+        protected internal override void ClosePin(int pinNumber)
+        {
+            if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle pinHandle))
+            {
+                Interop.ReleaseGpiodLine(pinHandle);
+                _pinNumberToSafeLineHandle.Remove(pinNumber);
+            }
+        }
+
+        protected internal override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber) => throw
```
ahsonkhan: nit: move the `throw` keyword to the next line

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -0,0 +1,159 @@
...
+                _pinNumberToSafeLineHandle.Remove(pinNumber);
+            }
+        }
+
+        protected internal override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber) => throw 
+            ExceptionHelper.GetPlatformNotSupportedException(ExceptionResource.ConvertPinNumberingSchemaError);
+
+        protected internal override PinMode GetPinMode(int pinNumber)
+        {
+            if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle pinHandle))
```
ahsonkhan: nit: invert the condition and throw up front for consistency with return after the if block.

krwq:
Thanks @buyaa-n!

buyaa-n:
> Thanks @buyaa-n!

Sure, thank you all for your reviews

PR #222, opened 2019-02-13
Add more Mcp23xxx tests.

JeremyKuhne:
- Align Mcp address validation with datasheets
- Allow passing in IGpioController for master controller
- Add basic tests for reset functionality
- Fix reset caching issues
- Add validation for disabled state and tests
- Remove dead code
- Add deconstructor and bit mask helper for PinValuePair


src/System.Device.Gpio/System/Device/Gpio/PinValuePair.cs
```diff
@@ -17,5 +17,47 @@ public PinValuePair(int pin, PinValue value)
...
+        /// Returns a bit mask of designated pins. Supports pin numbers from 0 to 31.
+        /// </summary>
+        /// <remarks>
+        /// On some devices (such as GPIO extenders) pin numbers are low and often consecutive. This
+        /// is a helper to turn sets of pins into bits/bitmasks for storage and/or sending over the
+        /// wire to devices that want pin numbers in a bit format.
+        /// </remarks>
+        /// <param name="pinValues">A set of pin value pairs.</param>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// Thrown if given a negative pin or pin number of 32 or higher.
```
shaggygi: Could this bite us later?  For example, The [CY8C95xx](https://www.cypress.com/file/37971/download) I plan to work on soon has 20/40/60 I/Os.

src/devices/Mcp23xxx/Mcp23008.cs
```diff
@@ -2,25 +2,42 @@
...
-            : base(new I2cAdapter(i2cDevice), i2cDevice.ConnectionSettings.DeviceAddress, reset, interrupt)
+        /// <param name="reset">
+        /// The output pin number that is connected to the hardware reset, if any. If specified the device
+        /// will start in a disabled state.
+        /// </param>
+        /// <param name="interrupt">The input pin number that is connected to the interrupt, if any.</param>
+        /// <param name="masterController">
+        /// The controller for the reset and interrupt pins. If not specified, the default controller will be used.
+        /// </param>
+        public Mcp23008(I2cDevice i2cDevice, int reset = -1, int interrupt = -1, IGpioController masterController = null)
```
shaggygi: Why change reset/interrupt to use -1 instead of null?  Just curious as I did this in other bindings I might need to change based on your thinking.

src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -205,54 +206,41 @@ public virtual void Dispose()
         {
             _masterGpioController?.Dispose();
             _masterGpioController = null;
-            _device?.Dispose();
+            _bus?.Dispose();
         }
 
         /// <summary>
-        /// Disables the device by setting the reset pin LOW.
+        /// Disables the device by setting the reset pin low.
```
shaggygi: Question: Is the convention better to have uppercase or lowercase for things like OFF/ON, LOW/HIGH, FALSE/TRUE, etc.?  Just curious as I want to make sure I'm following the same in other bindings.

src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -314,20 +304,16 @@ public void SetPinMode(int pinNumber, PinMode mode)
...
         {
             ValidatePin(pinNumber);
-            Span<PinValuePair> values = stackalloc PinValuePair[1];
-            values[0] = new PinValuePair(pinNumber, default);
+            Span<PinValuePair> values = stackalloc PinValuePair[] { new PinValuePair(pinNumber, default) };
             Read(values);
             return values[0].PinValue;
         }
 
         public void Read(Span<PinValuePair> pinValues)
```
shaggygi: Summary since public?

src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -367,25 +353,15 @@ public void Write(int pinNumber, PinValue value)
 
         public void Write(ReadOnlySpan<PinValuePair> pinValues)
```
shaggygi: Summary since public?

src/devices/Mcp23xxx/Mcp23x0x.cs
```diff
@@ -2,15 +2,17 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Device.Gpio;
+
 namespace Iot.Device.Mcp23xxx
 {
     /// <summary>
     /// Wraps 8 bit MCP I/O extenders
```
shaggygi: NIT: Should be "8-bit".  Should extenders be expanders?

src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -4,89 +4,53 @@
...
         /// <param name="interruptA">The input pin number that is connected to the interrupt for Port A (INTA).</param>
         /// <param name="interruptB">The input pin number that is connected to the interrupt for Port B (INTB).</param>
+        /// <param name="masterController">
+        /// The controller for the reset and interrupt pins. If not specified, the default controller will be used.
+        /// </param>
         /// <param name="bankStyle">
         /// The current bank style of the ports. This does not set the bank style- it tells us what the bank style is.
         /// It is *highly* recommended not to change the bank style from the default as there is no direct way to
         /// detect what style the chip is in and most apps will fail if the chip is not set to defaults. This setting
         /// has no impact on 8 bit expanders.
```
shaggygi: NIT: Should be "8-bit".

src/System.Device.Gpio/System/Device/Gpio/PinValuePair.cs
```diff
@@ -17,5 +17,47 @@ public PinValuePair(int pin, PinValue value)
...
+        {
+            pinNumber = PinNumber;
+            pinValue = PinValue;
+        }
+
+        /// <summary>
+        /// Returns a bit mask of designated pins. Supports pin numbers from 0 to 31.
+        /// </summary>
+        /// <remarks>
+        /// On some devices (such as GPIO extenders) pin numbers are low and often consecutive. This
```
shaggygi: Should extenders be expanders?

src/devices/Mcp23xxx/tests/Mcp23xxxTest.cs
```diff
@@ -96,7 +98,10 @@ public I2cDeviceMock(int ports, I2cConnectionSettings settings = null)
             public override byte ReadByte() => throw new NotImplementedException();
         }
 
-        public class Mcp23xxxMock
+        /// <summary>
+        /// Mock the behavior of the chip
+        /// </summary>
+        public class Mcp23xxxChipMock
```
shaggygi: Question: I have not done much mocking (but hopefully plan to as I learn more in this space).  When it comes to bindings, is it preferred to create similar to what you've done here or is it accepted to use other packages like [Moq](https://www.nuget.org/packages/moq/)?

src/devices/Mcp23xxx/tests/Mcp23xxxTest.cs
```diff
@@ -96,7 +98,10 @@ public I2cDeviceMock(int ports, I2cConnectionSettings settings = null)
             public override byte ReadByte() => throw new NotImplementedException();
         }
 
-        public class Mcp23xxxMock
+        /// <summary>
+        /// Mock the behavior of the chip
```
shaggygi: NIT: Period

src/System.Device.Gpio/System/Device/Gpio/PinValuePair.cs
JeremyKuhne: Yeah, I'll change it.

src/System.Device.Gpio/System/Device/Gpio/PinValuePair.cs
JeremyKuhne: It doesn't prevent us from adding another helper. I did find out were not allowed to use tuples in public API so I'll have to change this. `BitVector32` is useful, but we don't have a 64 bit equivalent.

I discussed it a bit and I think the right answer (that will get past the review) is to add a `PinVector32` and `PinVector64` struct and respective helpers here. We can add deconstructors to those which will allow using them tuple style if desired without actually exposing tuples (verboten).

src/devices/Mcp23xxx/Mcp23008.cs
JeremyKuhne: Discussed with API review folks. There is a small perf consideration to using nullables and the general guideline is to not use nullables if you have values that don't have meaning. We don't expect to have valid negative pins, so -1 is the right way to go here.

src/devices/Mcp23xxx/Mcp23x0x.cs
JeremyKuhne: Changing.

src/devices/Mcp23xxx/Mcp23xxx.cs
JeremyKuhne: Fixed

src/devices/Mcp23xxx/Mcp23xxx.cs
JeremyKuhne: I'm just leaning towards how it would be written in the docs. We wouldn't write it in caps there.

src/devices/Mcp23xxx/Mcp23xxx.cs
JeremyKuhne: As a general rule I try not to duplicate comments for overrides or interface implementations. Quite often they end up deviating over time and getting messy. I'm not super strict on that point and I don't know that we have actual guidelines around it.

src/devices/Mcp23xxx/tests/Mcp23xxxTest.cs
JeremyKuhne: I discussed this point briefly with @joperezr. I think we should be using Moq for tests, I just didn't want to consume a specific library without building some consensus first.

We don't do a ton of mocking in CoreFX as we rely on the hardware. In my previous group we did. The recommendation is to use the mocking framework as it helps describe what you're actually testing and allows you to do more extensive validation easily (like only this method was called and it was only called once).

In this particular case I'm not mocking another API I'm simulating some hardware. I probably should have called it `Mcp32xxxChipSimulator` to be clearer? I could have put this functionality in each of the classes I am mocking, and in fact did to begin with.  I found it harder to track what I was doing and understand what was going on in the tests.

shaggygi: We might want to come up with practices around using Moq and bindings to help make consistent.  Maybe some thoughts around the main pin types (Gpio, Pwm, I2c, Spi, etc.) would be good.  Of course, actually trying it on the hardware would tell us if it really worked or not (with timing, connection, etc.), but I could see the "simulators" being helpful... especially getting started on coding and studying the chip datasheet (while waiting on the chips to arrive in the mail :smile:)

JeremyKuhne: > but I could see the "simulators" being helpful.

They do provide quite a bit of value. Actually running on hardware is the proof in the pudding, of course. We can't, however, reasonably run automation on real devices (well, not many of them for sure).  Simulating the hardware can help validate behavior and avoid regressions. While we'll undoubtedly have bugs in our simulations, those will get fixed over time and we'll have much higher confidence in the state of our code base before it gets to the unbounded permutations of the real world.

One way I think of it sometimes is as an emulated or low fidelity record/playback system.

src/System.Device.Gpio/System/Device/Gpio/PinVector32.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// </summary>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// Thrown if <paramref name="pinValues"/> contains negative pin numbers or pin numbers higher than 31.
+        /// </exception>
+        public PinVector32(ReadOnlySpan<PinValuePair> pinValues)
+        {
+            Pins = 0;
+            Values = 0;
+
+            foreach ((int pin, PinValue value) in pinValues)
```
krwq: Does Deconstruct allow using this? Pretty nice if so

src/System.Device.Gpio/System/Device/Gpio/PinVector32.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    /// <summary>
+    /// 32-bit vector of pins and values.
+    /// </summary>
+    public struct PinVector32
+    {
+        /// <summary>
+        /// Bit vector of pin numbers from 0 (bit 0) to 31 (bit 31).
+        /// </summary>
+        public uint Pins { get; set; }
```
krwq: I think this should be called something like PinValueMask (something with Mask in the name I think)

src/System.Device.Gpio/System/Device/Gpio/PinVector32.cs
JeremyKuhne: Yep :)

src/System.Device.Gpio/System/Device/Gpio/PinVector32.cs
```diff
@@ -0,0 +1,69 @@
...
+            Pins = 0;
+            Values = 0;
+
+            foreach ((int pin, PinValue value) in pinValues)
+            {
+                if (pin < 0 || pin >= sizeof(uint) * 8)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(pinValues));
+                }
+                uint bit = (uint)(1 << pin);
```
krwq: nit: enter after brace

src/System.Device.Gpio/System/Device/Gpio/PinVector32.cs
JeremyKuhne: Which? The struct itself or the member? Naming for the struct was meant to align with what we current have for `BitVector32`. I tried more complicated wording for the properties but it got a bit cumbersome, imho.

src/devices/Mcp23xxx/Mcp23008.cs
krwq: I'd rename masterController to parentController :smile:

JeremyKuhne: Yeah, I'm unsure on this one. I'm following existing terminology for this- it is the master node that sends commands down the bus...

src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -205,54 +206,41 @@ public virtual void Dispose()
         {
             _masterGpioController?.Dispose();
             _masterGpioController = null;
-            _device?.Dispose();
+            _bus?.Dispose();
```
krwq: I think this is missing `_bus = null` after

src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -314,20 +304,16 @@ public void SetPinMode(int pinNumber, PinMode mode)
...
         {
-            ushort pins = 0;
-            foreach (PinValuePair pair in pinValues)
-            {
-                ValidatePin(pair.PinNumber);
-                pins |= (ushort)(1 << pair.PinNumber);
-            }
+            (uint pins, _) = new PinVector32(pinValues);
+            if ((pins >> PinCount) > 0)
+                ThrowBadPin(nameof(pinValues));
```
krwq: I'd recommend BadPin return the exception and do `throw BadPin(...)` here - the stack looks better (no strong opinion)

src/devices/Mcp23xxx/samples/Mcp23xxx.Sample.cs
```diff
@@ -81,13 +81,13 @@ private static Mcp23xxx GetMcp23xxxDevice(Mcp23xxxDevice mcp23xxxDevice)
             switch (mcp23xxxDevice)
             {
                 case Mcp23xxxDevice.Mcp23S08:
-                    return new Mcp23S08(s_deviceAddress, spiDevice);
+                    return new Mcp23s08(spiDevice, s_deviceAddress);
```
krwq: My recommendation would be to add SpiMinAddress and SpiMaxAddress consts to all devices, so that users have an easier way to remember what the address should be

src/devices/Mcp23xxx/tests/Mcp23xxxTest.cs
```diff
@@ -157,5 +162,59 @@ public void Write(Span<byte> data)
                 }
             }
         }
+
```
krwq: extra enter

src/devices/Mcp23xxx/Mcp23008.cs
krwq: fine if this is existing terminology

src/devices/Mcp23xxx/Mcp23xxx.cs
JeremyKuhne: I can't as I made it `readonly`. It begs the question of what is the right thing to do with disposable members. `readonly` has value and setting to null here doesn't _guarantee_ that it can't be reused in threaded scenarios.

JeremyKuhne: I'm going to poke around on this subject some more to see if we have / can establish guidelines.

src/devices/Mcp23xxx/Mcp23xxx.cs
JeremyKuhne: While I didn't really need it in this particular case, in some cases it is important to have the throw come from another method. Throws prevent the JIT from inlining methods.

src/System.Device.Gpio/System/Device/Gpio/PinVector32.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Device.Gpio
+{
+    /// <summary>
+    /// 32-bit vector of pins and values.
+    /// </summary>
+    public struct PinVector32
```
joperezr: Let's get an issue filed with this proposal instead of having it on this PR. That way we can review it like we do for other proposals and once we have a design in mind we can submit a PR with the implementation

src/devices/Mcp23xxx/samples/Mcp23xxx.Sample.cs
JeremyKuhne: As it is hardware configured I don't think it is worth it- you should never be programmatically needing it. We call it out in the comments, but you have to physically look at the wiring to know what address to use.

JeremyKuhne:
> Let's remove the main library changes for now form this PR as new classes here have to follow the process of adding APIs by going to API review like in corefx.

I'll wait on resolution of https://github.com/dotnet/iot/issues/230. I don't want to spend time un/reimplementing unless we outright reject the new APIs.

src/System.Device.Gpio/System/Device/Gpio/PinVector32.cs
JeremyKuhne: I've created the issue #230 and moved the structs to \devices\Common pending resolution.

JeremyKuhne:
> I'll wait on resolution of #230. I don't want to spend time un/reimplementing unless we outright reject the new APIs.

After discussing with @joperezr, moving to Common to allow utilizing while we discuss whether or not this is useful public API.

krwq:
@joperezr can we do API review once in a while instead of per API? We are in early stages of development and this feels really cumbersome to go through that every time. I think we should scope API reviews to only GpioController related classes since everything depends on that

JeremyKuhne:
I think I've addressed @joperezr main issue for now. Going to merge so I can update my other pull. As the code is in, just internal, moving it back is easy. If there are any other issues I'm happy to address them.

PR #223, opened 2019-02-13
Fix and improve code

ZhangGaoxing:


src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -40,32 +42,31 @@ public Sonar(int triggerPin, int echoPin)
...
         /// Gets the current distance in cm.
         /// </summary>
-        public double GetDistance()
+        private double GetDistance()
         {
             _timer.Reset();
 
             // Trigger input for 10uS to start ranging
             // ref https://components101.com/sites/default/files/component_datasheet/HCSR04%20Datasheet.pdf
+            _controller.Write(_trigger, PinValue.Low);
```
krwq: why is Low needed here? It should be low already after previous measuring

src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -40,32 +42,31 @@ public Sonar(int triggerPin, int echoPin)
...
             Thread.Sleep(TimeSpan.FromMilliseconds(0.01));
             _controller.Write(_trigger, PinValue.Low);
             
-            _timer.Start();
-            
             while(_controller.Read(_echo) == PinValue.Low)
             {
             }
-
+            _timer.Start();
```
krwq: we should measure time starting from when we end sending the "High" pulse, so previously was correct

src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -40,32 +42,31 @@ public Sonar(int triggerPin, int echoPin)
...
             {
             }
-
             _timer.Stop();
             
             TimeSpan elapsed = _timer.Elapsed;
 
             // distance = (time / 2) × velocity of sound (34300 cm/s)
-            return (double)(elapsed.TotalMilliseconds * 34.3) / 2 - 8;
+            return elapsed.TotalMilliseconds / 2.0 * 34.3;
```
krwq: `- 8` was based on the experiments. Any reasons to remove?

src/devices/Max7219/Register.cs
```diff
@@ -24,4 +24,4 @@ public enum Register : byte
         SHUTDOWN = 0xC,
         DISPLAYTEST = 0xF
     }
-}
```
krwq: since you already are onto it, this class can be private I presume

krwq:
Hey @ZhangGaoxing thanks for doing the fixes, the only comment I have is about Hcsr04 - I think the new logic will not work correctly since you start measuring time too late (the `-8` is spurious though but worked better in practice when I tried it)

src/devices/Hcsr04/Hcsr04.cs
ZhangGaoxing: Start measuring the trigger pin that need to be written LOW -> HIGH -> LOW according to the datasheet. We don't know what the user will do after a measurement, so I think we should follow the datasheet.

src/devices/Hcsr04/Hcsr04.cs
ZhangGaoxing: No, Time should be measured when _echo change to high level. That's why the original distance needs to -8.

src/devices/Hcsr04/Hcsr04.cs
ZhangGaoxing: @krwq When I first saw this formula, I was wondering why this needs to -8, and it's a cm. I think about the thickness of the sensor, but it's not very reliable. Ultimately, it's still a mistake in measuring time. You can try it again with new code.😁

src/devices/Max7219/Register.cs
ZhangGaoxing: This code was not written by me. I changed the type to internal, but many public methods called it, so I gave up changing it. Maybe it was my VS saved the change, and I forgot to discard it.

src/devices/Hcsr04/Hcsr04.cs
krwq: if you start measuring from the _echo being high then you measure the pulse width and not the time between sending and receiving. I'll double read the spec to see if I didn't misunderstand anything.

src/devices/Hcsr04/Hcsr04.cs
krwq: @ZhangGaoxing but we do control pin in this class and assume noone else is writing to it in here and it will be low already - I don't mind setting it explicitly but seems redundant to me since it's already at low

src/devices/Hcsr04/Hcsr04.cs
krwq: @ZhangGaoxing reading: https://cdn.sparkfun.com/datasheets/Sensors/Proximity/HCSR04.pdf

Spec wording is really weird. First they say:

```
The Echo is a distance object that is pulse width and the range in proportion
```

then

```
You can calculate the range through the time interval between sending trigger signal and receiving echo signal
and give formula
range = high level time * velocity (340M/S) / 2
```

I guess both will work then and the 8 I was seeing was some factor multiplied by response time from the sensor and formula still worked since it included actual time. I think this should work as is. Let me check with my sensor to see if I get correct results.

As for the initial low time: We don't need to set it directly since it's already low but we should make sure that at least 60ms elapsed since last measuring (spec says `we suggest to use over 60ms measurement cycle, in order to prevent trigger signal to the echo signal`). So we should check Environment.TickCount after we are finished and another time when we start and if 60ms hasn't elapsed we should wait until it does

krwq: add enter after `}`

ZhangGaoxing: @krwq like this? But I feel that it takes more time than write the _trigger LOW.
```C#
private int lastMeasurment = 0;

private double GetDistance()
{
    _timer.Reset();

    while (Math.Abs(Environment.TickCount - lastMeasurment) < 60)
    {
    }
    _controller.Write(_trigger, PinValue.High);
    Thread.Sleep(TimeSpan.FromMilliseconds(0.01));

    while (_controller.Read(_echo) == PinValue.Low)
    {
    }

    _timer.Start();
    while (_controller.Read(_echo) == PinValue.High)
    {
    }
    _timer.Stop();

    lastMeasurment = Environment.TickCount;

    TimeSpan elapsed = _timer.Elapsed;

    return elapsed.TotalMilliseconds / 2.0 * 34.3;
}
```

krwq: @ZhangGaoxing _trigger LOW does not protect against echo. I will push to your branch with the code.

I've just tested the new measuring code and it works really accurate except for 0-4cm range where spec says it doesn't work well anyway :smile:

krwq: Actually the code above looks correct - also noticed the issue which we can fix since we are on it: when there is timeout (no object) it just hangs

krwq:
@ZhangGaoxing I've pushed to your branch with feedback:
- ive made internal that Max7219 thing I commented on in the beginning
- swapped sides on the elapsed check since TickCount > lastCheck you should be subtracting (also added sleep in the loop)

ZhangGaoxing:
> swapped sides on the elapsed check since TickCount > lastCheck you should be subtracting (also added sleep in the loop)

Thanks. I was careless.🤦‍♂️

krwq:
@ZhangGaoxing no worries, I've done multiple time, I only fixed it since I've started checking out the Max7219 registers and when I finished I noticed you pushed to the branch and then spotted it - I didn't want to cause any merge conflicts so I just changed it

PR #224, opened 2019-02-13
Added Hmc5883 binding

DemoS76:


krwq:
Hey @DemoS76 thank you for contributing! Before I start going deeper into this PR, I'm seeing some overlap with this PR and https://github.com/dotnet/iot/tree/master/src/devices/Hmc5883l I think we could merge these 2 devices together which would be ideal. If you think they are different enough we can look at it later and I'll review as is. (I'm seeing at least some overlap in the registers and names so think this should be possible)

DemoS76:
Hi Krzysztof,

Looks like @ZhangGaoxing outran me a bit:) You are right my PR overlaps the other PR. But my code had some differences. So I've removed my binding and added a couple fixes to the existing implementation.
Could you please review the code?

src/devices/Hmc5883l/Hmc5883l.cs
```diff
@@ -58,7 +76,7 @@ public Hmc5883l(I2cDevice sensor, Gain gain = Gain.Gain1090, MeasuringMode measu
         private void Initialize()
         {
             // Details in Datasheet P12
-            byte configA = (byte)(0x70 + _outputRate << 2);
+            byte configA = (byte)(_samplesAmount | _outputRate << 2 | _measurementConfig);
```
krwq: brackets please :smile:

src/devices/Hmc5883l/MeasurementConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Hmc5883l
+{
+    /// <summary>
+    /// Measurement configuration.
+    /// This enum defines the measurement flow of the device, specifically whether or not to incorporate an applied bias to the sensor into the measurement.
+    /// </summary>
+    public enum  MeasurementConfiguration: byte
```
krwq: nit: space before `:`

src/devices/Hmc5883l/MeasurementConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Positive bias configuration for X and Y axes, negative bias configuration for Z axis.
+        /// In this configuration, a positive current is forced across the resistive load for X and Y axes, a negative current for Z axis.
+        /// </summary>
+        PositiveBiasConfiguration = 0b00000001,
+        /// <summary>
+        /// Negative bias configuration for X and Y axes, positive bias configuration for Z axis.
+        /// In this configuration, a negative current is forced across the resistive load for X and Y axes, a positive current for Z axis.
+        /// </summary>
+        NegativeBias = 0b00000010
```
krwq: What does this mean in terms of output? Will the vector be negated now? (so equivalent to putting it upside down?)

src/devices/Hmc5883l/SamplesAmount.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Number of samples averaged (1 to 8) per measurement output.
+    /// </summary>
+    public enum  SamplesAmount: byte
+    {
+        /// <summary>
+        /// 1 (Default) samples per measurement output.
+        /// </summary>
+        One = 0b00000000,
```
joperezr: NIT: We usually try to use underscore to separate every 4 bits like `0b_0000_0000`. Makes it more readable. Same applies for the rest of the values.

src/devices/Hmc5883l/Status.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// </summary>
+    public enum  Status
+    {
+        /// <summary>
+        /// Regulator Enabled Bit. This bit is set when the internal voltage regulator is enabled.
+        /// This bit is cleared when the internal regulator is disabled.
+        /// </summary>
+        Ren,
+        /// <summary>
+        /// Data output register lock. This bit is set when this some but not all for of the six data output registers have been read.
```
joperezr: NIT: `when this some` -> `when some`

src/devices/Hmc5883l/Status.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// When this bit is set, the six data output registers are locked and any new data will not be placed in these register until
+        /// on of four conditions are met: one, all six have been read or the mode changed, two, a POR is issued, three, the
```
joperezr: NIT: `on` -> `one`

src/devices/Hmc5883l/Status.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// mode is changed, or four, the measurement is changed.
+        /// </summary>
+        LOCK,
+        /// <summary>
+        /// Ready Bit. Set when data is written to all six data registers. Cleared when device initiates a write to the data output
+        /// registers, when in off mode, and after one or more of the data output registers are written to. 
+        /// When RDY bit is clear it shall remain cleared for a minimum of 5 μs. 
+        /// DRDY pin can be used as an alternative to the status register for monitoring the device for measurement data.
```
joperezr: So we don't want to include DRDY as a Status?

src/devices/Hmc5883l/Hmc5883l.cs
```diff
@@ -120,5 +138,33 @@ public void Dispose()
                 _sensor = null;
             }
         }
+
+        /// <summary>
+        /// Reads device statuses .
+        /// </summary>
+        /// <returns>IList<Status></returns>
```
joperezr: the return comment should be more specific mentioning what is it returning and perhaps even possible values and their meaning

src/devices/Hmc5883l/Hmc5883l.cs
```diff
@@ -120,5 +138,33 @@ public void Dispose()
...
+        private IList<Status> GetStatus()
+        {
+            var result = new List<Status>();
+            _sensor.WriteByte((byte)Register.HMC_STATUS_REG_ADDR);
+            byte status = _sensor.ReadByte();
+
+            byte mask = 0b00000001;
```
joperezr: NIT instead of having a variable, can you instead just do the inline check like: `if ((status & 0b_0001) != 0) {` etc.

src/devices/Hmc5883l/Hmc5883l.cs
joperezr: Looks like by setting _samplesAmount to One as default won't match the existing implementation, so presumably we might want to default to a different value? why was the samples amount before `0x70`, and why don't we have that value in your SamplesAmount enum?

src/devices/Hmc5883l/Status.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <summary>
+    /// The status of HMC5883L device
+    /// </summary>
+    public enum  Status
+    {
+        /// <summary>
+        /// Regulator Enabled Bit. This bit is set when the internal voltage regulator is enabled.
+        /// This bit is cleared when the internal regulator is disabled.
+        /// </summary>
+        Ren,
```
krwq: I'd call it full names if possible so just RegulatorEnabled

src/devices/Hmc5883l/Status.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// Data output register lock. This bit is set when this some but not all for of the six data output registers have been read.
+        /// When this bit is set, the six data output registers are locked and any new data will not be placed in these register until
+        /// on of four conditions are met: one, all six have been read or the mode changed, two, a POR is issued, three, the
+        /// mode is changed, or four, the measurement is changed.
+        /// </summary>
+        LOCK,
```
krwq: `LOCK` -> `Lock`

src/devices/Hmc5883l/Hmc5883l.cs
DemoS76: Fixed

src/devices/Hmc5883l/MeasurementConfiguration.cs
DemoS76: Fixed

src/devices/Hmc5883l/MeasurementConfiguration.cs
DemoS76: PositiveBiasConfiguration and NegativeBias values are used for self test operation. You can find more details on page 19 of datasheet.

src/devices/Hmc5883l/SamplesAmount.cs
DemoS76: Fixed

src/devices/Hmc5883l/Status.cs
DemoS76: Fixed

src/devices/Hmc5883l/Status.cs
DemoS76: Fixed

src/devices/Hmc5883l/Status.cs
DemoS76: Sorry for confusion. DRDY is hardware interrupt pin not register value. It was mentioned in the datasheet.

src/devices/Hmc5883l/Hmc5883l.cs
DemoS76: Fixed

src/devices/Hmc5883l/Hmc5883l.cs
DemoS76: Fixed

src/devices/Hmc5883l/Hmc5883l.cs
DemoS76: Could you please find the description for configuring register A on page 12 of datasheet?
I assume that the existing implementation is not working right.
(0x70 + _outputRate << 2) = 11010000 where _outputRate = 0x04 is default value.
According of datasheet the CRA7 bit must be 0. Also CRA6 and CRA5 bits have values 1 and 0 so it is not default value for "number of samples" configuration. So the 0x70 value is not correct for configuration.

In my implementation we have separated parts for each of configuration bits which are summed by "OR" logical operations.
When we set default values for all configs:
_samplesAmount = 0b_0000_0000(One)
_outputRate << 2 = 0b_0001_0000(Rate15)
_measurementConfig = 0b_0000_0000(Normal)

we have got result: 16 (0b_0001_0000) that is correct value for configuration of register A

src/devices/Hmc5883l/Status.cs
DemoS76: Fixed

src/devices/Hmc5883l/Status.cs
DemoS76: Fixed

src/devices/Hmc5883l/Hmc5883l.cs
```diff
@@ -120,5 +138,30 @@ public void Dispose()
...
+        /// Reads device statuses.
+        /// </summary>
+        /// <returns>Device statuses</returns>
+        private IList<Status> GetStatus()
+        {
+            var result = new List<Status>();
+            _sensor.WriteByte((byte)Register.HMC_STATUS_REG_ADDR);
+            byte status = _sensor.ReadByte();
+
+            if ((status & 0b_0000_0001) != 0) {
```
krwq: this might be useful: https://docs.microsoft.com/en-us/dotnet/api/system.enum.hasflag?view=netcore-2.1

krwq: also if status is just `[Flags]` then you don't need the list at all and can just `(Status)status`

src/devices/Hmc5883l/Status.cs
```diff
@@ -0,0 +1,30 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Hmc5883l
+{
+    /// <summary>
+    /// The status of HMC5883L device
+    /// </summary>
+    public enum Status
```
krwq: make this `[Flags]` and match values with what the device returns

src/devices/Hmc5883l/Hmc5883l.cs
DemoS76: Fixed

src/devices/Hmc5883l/Status.cs
DemoS76: Done

krwq:
@DemoS76, thanks for fxing up the PR, LGTM

PR #227, opened 2019-02-13
Added 28BYJ-48 stepper motor binding

johnastaf:
28BYJ-48 Stepper Motor 5V 4-Phase 5-Wire & ULN2003 Driver Board


src/devices/SM28BYJ48/SM28BYJ48.cs
```diff
@@ -0,0 +1,202 @@
...
+        private Action _movementCompleted;
+
+        /// <summary>
+        /// Initialize a StepperMotor class
+        /// </summary>
+        /// <param name="pin1">The GPIO pin number which corresponds pin A on ULN2003 driver board.</param>
+        /// <param name="pin2">The GPIO pin number which corresponds pin B on ULN2003 driver board.</param>
+        /// <param name="pin3">The GPIO pin number which corresponds pin C on ULN2003 driver board.</param>
+        /// <param name="pin4">The GPIO pin number which corresponds pin D on ULN2003 driver board.</param>
+        public SM28BYJ48(int pin1, int pin2, int pin3, int pin4)
```
krwq: do all stepper have 4 windings or can they have i.e. 8 for more precision? (This is fine as is for now regardless of the answer since we can easily change it later)

src/devices/SM28BYJ48/SM28BYJ48.cs
```diff
@@ -0,0 +1,202 @@
...
+using System;
+using System.Device.Gpio;
+using System.Threading;
+
+namespace Iot.Device.SM28BYJ48
+{
+    /// <summary>
+    /// Can move the motor clockwise and counterclockwise.
+    /// </summary>
+    public enum RotationDirection
```
krwq: One file per type (in VS you can use Ctrl-. to extract it automatically)

src/devices/SM28BYJ48/SM28BYJ48.cs
```diff
@@ -0,0 +1,202 @@
...
+            _controller.OpenPin(_pin3, PinMode.Output);
+            _controller.OpenPin(_pin4, PinMode.Output);
+        }
+
+        /// <summary>
+        /// Set the stepper's mode.
+        /// </summary>
+        /// <param name="type">The mode.</param>
+        /// <returns>Current stepper motor.</returns>
+        public SM28BYJ48 SetStepperType(StepperType type)
```
krwq: why return this? Is it convenience to provide multiple settings and chain? We usually provide that kind of stuff with Settings class and then that's usually not needed

src/devices/SM28BYJ48/SM28BYJ48.cs
```diff
@@ -0,0 +1,202 @@
...
+        HalfStep,
+        FullStepSinglePhase,
+        FullStepDualPhase
+    }
+
+    /// <summary>
+    /// This class is for controlling stepper motors that are controlled by a 4 pin controller board.
+    /// </summary>
+    /// <remarks>It is tested and developed using the 28BYJ-48 stepper motor and the ULN2003 driver board.</remarks>
+    public class SM28BYJ48 : IDisposable
```
krwq: Will this work for any stepper motor? if so I'd suspect you can just call it StepperMotor

src/devices/SM28BYJ48/SM28BYJ48.cs
```diff
@@ -0,0 +1,202 @@
...
+namespace Iot.Device.SM28BYJ48
+{
+    /// <summary>
+    /// Can move the motor clockwise and counterclockwise.
+    /// </summary>
+    public enum RotationDirection
+    {
+        Stopped,
+        Clockwise,
+        CounterClockwise
```
krwq: Wondering if just having `Speed` or `StepsPerMinute`/`StepsPerSecond` accepting negative numbers wouldn't be easier to use

src/devices/SM28BYJ48/SM28BYJ48.cs
```diff
@@ -0,0 +1,202 @@
...
+            _type = type;
+            return this;
+        }
+
+        /// <summary>
+        /// Set the stepper's delay. 
+        /// </summary>
+        /// <param name="ms">The milliseconds value.</param>
+        /// <returns>Current stepper motor.</returns>
+        public SM28BYJ48 SetStepperDelay(int ms)
```
krwq: If you use `StepsPer...` then you can get rid of this

src/devices/SM28BYJ48/SM28BYJ48.cs
```diff
@@ -0,0 +1,202 @@
...
+            _delayMs = ms > 1 ? ms : 1;
+            return this;
+        }
+
+        /// <summary>
+        /// This method is being called when the motor has completed the amount of steps.
+        /// </summary>
+        /// <param name="movementCompleted">The callback.</param>
+        /// <returns>Current stepper motor.</returns>
+        public SM28BYJ48 AddMovementCompletionHandler(Action movementCompleted)
```
krwq: `event Action MovementCompleted` (or use Speed/StepsPer...)

src/devices/SM28BYJ48/SM28BYJ48.cs
```diff
@@ -0,0 +1,202 @@
...
+            _controller.Write(_pin2, PinValue.Low);
+            _controller.Write(_pin3, PinValue.Low);
+            _controller.Write(_pin4, PinValue.Low);
+        }
+
+        /// <summary>
+        /// Rotate the motor clockwise.
+        /// </summary>
+        /// <param name="steps">Number of steps.</param>
+        public void RotateClockwise(int steps)
```
krwq: I see, this goes with different design. Wondering if we could make it something like `StepsPerMinute` for continuous movement and also have this API at the same time. I guess having just one for now is fine (if you accept negative numbers here you can get rid of RotationDirection enum)

src/devices/SM28BYJ48/SM28BYJ48.cs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            SetRotationDirection(RotationDirection.Clockwise, steps);
+            Start();
+        }
+
+        /// <summary>
+        /// Rotate the motor counter clockwise.
+        /// </summary>
+        /// <param name="steps">Number of steps.</param>
+        public void RotateCounterClockwise(int steps)
```
krwq: make the RotateSteps take negative numbers and that will reduce it to a single API

src/devices/SM28BYJ48/SM28BYJ48.cs
```diff
@@ -0,0 +1,202 @@
...
+                    _movementCompleted?.Invoke();
+                }
+            }
+        }
+
+        private void ApplyEngineStep()
+        {
+            if (_type == StepperType.HalfStep)
+            {
+                _controller.Write(_pin1, _engineStep == 8 || _engineStep == 1 || _engineStep == 2 ? PinValue.High : PinValue.Low);
```
krwq: you can remove `? PinValue.High : PinValue.Low` now, we accept bool values since few days

krwq:
@johnastaf thank you for contributing! Really nice to see stepper motor :smile: I'll check this code out with my stepper motor whenever I got some time (likely after merging though)

src/devices/SM28BYJ48/SM28BYJ48.cs
johnastaf: Yeah, I wanted to use smth like chain... But no problem, I'll move this settings to new class to be consistent with other devices

src/devices/SM28BYJ48/SM28BYJ48.cs
johnastaf: Good idea :-) Thanks! I'll refactor this

ZhangGaoxing:
I think this should be named Uln2003 for the driver board, not SM28BYJ48 for the stepper motor. Because there are all kinds of motors, and there is only one driver board.

src/devices/SM28BYJ48/SM28BYJ48.cs
johnastaf: Unipolar stepper motors (like 28BYJ-48) have five or six contacts for connection and four electromagnetic coils in the housing (to be more precise, two coils divided into four). I found more info [here](https://learn.adafruit.com/all-about-stepper-motors/types-of-steppers)

src/devices/SM28BYJ48/SM28BYJ48.cs
johnastaf: Moved. Thanks for the useful hotkey :-)

src/devices/SM28BYJ48/SM28BYJ48.cs
johnastaf: Removed

src/devices/SM28BYJ48/SM28BYJ48.cs
johnastaf: Fixed

src/devices/SM28BYJ48/SM28BYJ48.cs
johnastaf: Removed

src/devices/SM28BYJ48/SM28BYJ48.cs
johnastaf: Fixed

johnastaf:
> I think this should be named Uln2003 for the driver board, not SM28BYJ48 for the stepper motor. Because there are all kinds of motors, and there is only one driver board.

For now I renamed as StepperMotor. Guys, I'll rename it as you suggest :-) I've tested motor with ULN2003A and ULN2003APG drivers

src/devices/SM28BYJ48/SM28BYJ48.cs
johnastaf: Renamed

joperezr:
Agreed with @ZhangGaoxing that this should be renamed as per the driver board.

krwq:
@johnastaf if the driver is related to specific board then call it with that, if not (which I suspect) then call it StepperMotor

src/devices/StepperMotor/StepperMotor.cs
```diff
@@ -0,0 +1,130 @@
...
+        private int _pin4;
+        private int _steps = 0;
+        private int _engineStep = 0;
+        private int _currentStep = 0;
+        private StepperMode _mode;
+        private bool _isClockwise = true;
+        private GpioController _controller;
+        private Stopwatch _stopwatch;
+        private uint _rpm;
+        private long stepMicrosecondsDelay;
```
krwq: private variables should start with `_`

src/devices/StepperMotor/StepperMotor.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// <summary>
+        /// Initialize a StepperMotor class.
+        /// </summary>
+        /// <param name="pin1">The GPIO pin number which corresponds pin A on ULN2003 driver board.</param>
+        /// <param name="pin2">The GPIO pin number which corresponds pin B on ULN2003 driver board.</param>
+        /// <param name="pin3">The GPIO pin number which corresponds pin C on ULN2003 driver board.</param>
+        /// <param name="pin4">The GPIO pin number which corresponds pin D on ULN2003 driver board.</param>
+        /// <param name="mode">The stepper's mode.</param>
+        /// <param name="rpm">Revolutions per minute.</param>
+        public StepperMotor(int pin1, int pin2, int pin3, int pin4, StepperMode mode = StepperMode.HalfStep, uint rpm = 15)
```
krwq: I think rpm should be signed number. `rpm = 0` should be no movement, `rpm < 0`, rotation one way, `rpm > 0` rotating the other way

src/devices/StepperMotor/StepperMotor.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            _pin1 = pin1;
+            _pin2 = pin2;
+            _pin3 = pin3;
+            _pin4 = pin4;
+            _mode = mode;
+            _rpm = rpm > 0 ? rpm : 15; // Default revolutions per minute for 28BYJ-48 is approximately 15
```
krwq: I think this is counterintuitive

src/devices/StepperMotor/StepperMotor.cs
```diff
@@ -0,0 +1,130 @@
...
+                    _currentStep++;
+                }
+            }
+        }
+
+        private void ApplyEngineStep()
+        {
+            if (_mode == StepperMode.HalfStep)
+            {
+                _controller.Write(_pin1, _engineStep == 8 || _engineStep == 1 || _engineStep == 2);
```
krwq: you can define a static array for those, i.e.:
`static bool[] s_halfStep = new bool[8] { true, true, true, false, false, false, false, false }`
- change step to be numbered from 0 and then:

```csharp
_controller.Write(_pin1, s_halfStep[_step]);
_controller.Write(_pin2, s_halfStep[(_step + 2) % 8]);
_controller.Write(_pin3, s_halfStep[(_step + 4) % 8]);
_controller.Write(_pin3, s_halfStep[(_step + 6) % 8]);
```

etc.. that should be doable for all modes I believe.

The steps increment can be done like that:

```csharp
// somewhere have
int inc = 1; // or -1 depending on direction
_step = (_step + inc + 8) % 8;
```

src/devices/StepperMotor/StepperMotor.cs
johnastaf: Agree, removed

johnastaf:
> @johnastaf if the driver is related to specific board then call it with that, if not (which I suspect) then call it StepperMotor

Ok, thanks guys! Renamed to Uln2003

src/devices/StepperMotor/StepperMotor.cs
johnastaf: Fixed

src/devices/StepperMotor/StepperMotor.cs
johnastaf: Got it! To be honest the code above is not worked as expected for me :-)

Krzysztof, I use the following sequence for halfstep mode :
Sequence

So for this mode I've added appropriate array:

        static bool[,] _halfStepSequence = new bool[4, 8] {
                 { true, true, false, false, false, false, false, true },
                 { false, true, true, true, false, false, false, false },
                 { false, false, false, true, true, true, false, false },
                 { false, false, false, false, false, true, true, true }
            };
I think this is more obvious. What do you think?

src/devices/StepperMotor/StepperMotor.cs
johnastaf: Hmm I bit doubt about using negative RPM. We already use negative steps for rotate counterclockwise.

                    motor.Mode = StepperMode.HalfStep;
                    motor.RPM = 8;
                    motor.Step(-2048);

Will the following code be misleading If I set RPM as -8?

src/devices/StepperMotor/StepperMotor.cs
krwq: That is fine too but I think the code will work once you change _step to be 0..7 instead of 1..8 (note that swapping step 8 with 0 and putting it as a first column is equivalent will make the table look more obvious)

krwq: I think the formula would only work for the half step and we would need to think how to write it in generic way to work for other types

johnastaf: Thanks! Generic way is good idea. Now I've added 3 matrix for each stepper motor types, not only for half step. Please find these changes in the 3rd commit

johnastaf:
> I think I'm fine with this as is, we can re-iterate and improve later once we get next person playing around with the stepper motor  Thanks for contributing @johnastaf! Please let me know if you want to apply more feedback before I merge this.

@krwq , thanks for review! Please go ahead with merge when you get a chance :-)

src/devices/Uln2003/Uln2003.cs
```diff
@@ -0,0 +1,169 @@
...
+                 { true, false, false, true, true, false, false, true },
+                 { true, true, false, false, true, true, false, false },
+                 { false, true, true, false, false, true, true, false },
+                 { false, false, true, true, false, false, true, true }
+            };
+
+        /// <summary>
+        /// Initialize a Uln2003 class.
+        /// </summary>
+        /// <param name="pin1">The GPIO pin number which corresponds pin A on ULN2003 driver board.</param>
```
joperezr: Can these be any pins? And given in any order? If each pin is controlling something different I would rather have a better name for these parameters than just pin1, pin2, etc.

src/devices/Uln2003/Uln2003.cs
```diff
@@ -0,0 +1,169 @@
...
+        private int _engineStep = 0;
+        private int _currentStep = 0;
+        private StepperMode _mode = StepperMode.HalfStep;
+        private bool[,] _currentSwitchingSequence = _halfStepSequence;
+        private bool _isClockwise = true;
+        private GpioController _controller;
+        private Stopwatch _stopwatch;
+        private long _stepMicrosecondsDelay;
+
+        static bool[,] _halfStepSequence = new bool[4, 8] {
```
joperezr: We do enforce all members in a class to use explicit access modifiers, so all these arrays should call out that they are private.

src/devices/Uln2003/Uln2003.cs
```diff
@@ -0,0 +1,169 @@
...
+            };
+
+        /// <summary>
+        /// Initialize a Uln2003 class.
+        /// </summary>
+        /// <param name="pin1">The GPIO pin number which corresponds pin A on ULN2003 driver board.</param>
+        /// <param name="pin2">The GPIO pin number which corresponds pin B on ULN2003 driver board.</param>
+        /// <param name="pin3">The GPIO pin number which corresponds pin C on ULN2003 driver board.</param>
+        /// <param name="pin4">The GPIO pin number which corresponds pin D on ULN2003 driver board.</param>
+        public Uln2003(int pin1, int pin2, int pin3, int pin4)
```
joperezr: This constructor should take in a IGpioController object, which is defaulted to null. That way, the binding will have support when used with Gpio extenders.

src/devices/Uln2003/Uln2003.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Stop the motor.
+        /// </summary>
+        public void Stop()
+        {
+            _steps = 0;
+            _stopwatch.Stop();
```
joperezr: seems to me you could end up in a case where this is either null or hasn't been started. We should definitely make sure that we check for that here.

src/devices/Uln2003/Uln2003.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        /// <summary>
+        /// Moves the motor. If the number is negative, the motor moves in the reverse direction.
+        /// </summary>
+        /// <param name="steps">Number of steps.</param>
+        public void Step(int steps)
+        {
+            double lastStepTime = 0;
+            _stopwatch = Stopwatch.StartNew();
```
joperezr: I'm not sure if we should care about that scenario, but I don't think that this is thread safe as you could have two threads calling step which would override the stopwatch. Also, it kind of feels that this should be an Async method.

src/devices/Uln2003/samples/Uln2003.Samples.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+
+ <ItemGroup>
+    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
```
joperezr: NIT: This one is not needed as the ProjectReference bellow will import this reference automatically.

joperezr:
@ZhangGaoxing thanks for your contribution! PR was merged before I could take a look but do you mind checking the comments I left to see if you could address those? Also, please remember to add the binding to the binding directory over at src/devices/Readme.md

ZhangGaoxing:
> thanks for your contribution! PR was merged before I could take a look but do you mind checking the comments I left to see if you could address those? Also, please remember to add the binding to the binding directory over at src/devices/Readme.md

@johnastaf

src/devices/Uln2003/Uln2003.cs
krwq: they have to be in specific order, that's what all tables everywhere call them (either pin0..pin3 or pin1..pin4 or something similar - even colors are not consistent) - I was trying to think of better name but there is none...

src/devices/Uln2003/Uln2003.cs
krwq: we do not guarantee thread safety for a single instance unless the doc says otherwise but I agree we should not allocate stopwatch here

johnastaf:
> @ZhangGaoxing thanks for your contribution! PR was merged before I could take a look but do you mind checking the comments I left to see if you could address those? Also, please remember to add the binding to the binding directory over at src/devices/Readme.md

@joperezr , thanks for review! I've sent new PR with fixed your comments and updated readme files (https://github.com/dotnet/iot/pull/247)
Sorry, forgot about src/devices/Readme.md. I'll add the binding after your response.

src/devices/Uln2003/samples/Uln2003.Samples.csproj
johnastaf: Removed

PR #229, opened 2019-02-13
Initial implementation of NXP/TI PCX857X GPIO expanders.

JeremyKuhne:
Tested on PCX8574A with a 1602LCD.

This does not include the readme, sample app, etc. I'll be following up with those. Getting this up pending #222 (which this includes) to illustrate GPIO API consumption.

See the most recent commit https://github.com/dotnet/iot/commit/c1721a47d040e0157b1bf90b0e8cc46932bb6bf1 for this change. I'll rebase once #222 is in.

### DIFF
--- a/src/devices/Pcx857x/Pca8574.cs
+++ b/src/devices/Pcx857x/Pca8574.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Device.Gpio;
+using System.Device.I2c;
+
+namespace Iot.Device.Pcx857x
+{
+    /// <summary>
+    /// Remote 8-bit I/O expander for I2C-bus with interrupt.
+    /// </summary>
+    /// <remarks>
+    /// Fast mode I2C variant of the Pcf8574.
+    /// </remarks>
+    public class Pca8574 : Pcx8574
+    {
+        /// <summary>
+        /// Initializes a new instance of the Pca8574 device.
+        /// </summary>
+        /// <param name="device">The I2C device.</param>
+        /// <param name="interrupt">The interrupt pin number, if present.</param>
+        /// <param name="gpioController">
+        /// The GPIO controller for the <paramref name="interrupt"/>.
+        /// If not specified, the default controller will be used.
+        /// </param>
+        public Pca8574(I2cDevice device, int interrupt = -1, IGpioController gpioController = null)
+            : base(device, interrupt, gpioController)
+        {
+        }
+    }
+}
--- a/src/devices/Pcx857x/Pca8575.cs
+++ b/src/devices/Pcx857x/Pca8575.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Device.Gpio;
+using System.Device.I2c;
+
+namespace Iot.Device.Pcx857x
+{
+    /// <summary>
+    /// Remote 16-bit I/O expander for I2C-bus with interrupt.
+    /// </summary>
+    /// <remarks>
+    /// Fast mode I2C variant of the Pcf8575.
+    /// </remarks>
+    public class Pca8575 : Pcx8575
+    {
+        /// <summary>
+        /// Initializes a new instance of the Pca8575 device.
+        /// </summary>
+        /// <param name="device">The I2C device.</param>
+        /// <param name="interrupt">The interrupt pin number, if present.</param>
+        /// <param name="gpioController">
+        /// The GPIO controller for the <paramref name="interrupt"/>.
+        /// If not specified, the default controller will be used.
+        /// </param>
+        public Pca8575(I2cDevice device, int interrupt = -1, IGpioController gpioController = null)
+            : base(device, interrupt, gpioController)
+        {
+        }
+    }
+}
--- a/src/devices/Pcx857x/Pcx857x.cs
+++ b/src/devices/Pcx857x/Pcx857x.cs
@@ -1,0 +1,204 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.Gpio;
+using System.Device.I2c;
+using System.Runtime.CompilerServices;
+
+namespace Iot.Device.Pcx857x
+{
+    public abstract class Pcx857x : IGpioController
+    {
+        protected I2cDevice Device { get; }
+        private readonly IGpioController _masterGpioController;
+        private readonly int _interrupt;
+
+        // Pin mode bits- 0 for input, 1 for output to match PinMode
+        private ushort _pinModes;
+
+        // Pin value bits, 1 for high
+        private ushort _pinValues;
+
+        /// <summary>
+        /// Remote I/O expander for I2C-bus with interrupt.
+        /// </summary>
+        /// <param name="device">The I2C device.</param>
+        /// <param name="interrupt">The interrupt pin number, if present.</param>
+        /// <param name="gpioController">
+        /// The GPIO controller for the <paramref name="interrupt"/>.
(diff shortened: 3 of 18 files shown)


src/devices/Pcx857x/Pca8574.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Device.Gpio;
+using System.Device.I2c;
+
+namespace Iot.Device.Pcx857x
+{
+    /// <summary>
+    /// Remote 8-bit I/O expander for I2C-bus with interrupt.
+    /// </summary>
+    /// <remarks>
+    /// Fast mode I2c variant of the Pcf8574.
```
shaggygi: NIT: uppercase I2C

src/devices/Pcx857x/Pca8574.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// </remarks>
+    public class Pca8574 : Pcx8574
+    {
+        /// <summary>
+        /// Constructor.
```
shaggygi: Update wording.  Looks like a few others are like this, too.

src/devices/Pcx857x/Pca8575.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Device.Gpio;
+using System.Device.I2c;
+
+namespace Iot.Device.Pcx857x
+{
+    /// <summary>
+    /// Remote 16-bit I/O expander for I2C-bus with interrupt.
+    /// </summary>
+    /// <remarks>
+    /// Fast mode I2c variant of the Pcf8575.
```
shaggygi: NIT: Uppercase I2C

src/devices/Pcx857x/Pcx857x.cs
```diff
@@ -0,0 +1,204 @@
...
+                pinValues[i] = new PinValuePair(pin, (data >> pin) & 1);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void ThrowInvalidPin(string argumentName)
+        {
+            // This is a helper to allow the JIT to inline calling methods.
+            // (Methods with throws cannot be inlined.)
+            throw new ArgumentOutOfRangeException(argumentName, $"Pin numbers must be in the range of 0 to {PinCount - 1}");
```
shaggygi: NIT: Period end of message.

JeremyKuhne:
Merged in master and cleaned up.

src/devices/Pcx857x/Pcx857x.cs
```diff
@@ -0,0 +1,204 @@
...
+            {
+                WriteByte((byte)_pinValues);
+            }
+            else
+            {
+                InternalWriteUInt16(_pinValues);
+            }
+        }
+
+        private ushort SetBit(ushort data, int bitNumber) => data |= (ushort)(1 << bitNumber);
```
krwq: what's the assignment for? Should this be just `data | (ushort)(1 << bitNumber)` or perhaps ref is missing?

krwq: same comment for ClearBit

src/devices/Pcx857x/tests/ConstructionTests.cs
```diff
@@ -0,0 +1,28 @@
...
+        [MemberData(nameof(TestDevices))]
+        public void RegisterInitialization(TestDevice testDevice)
+        {
+            // All pins should be set to output.
+            Pcx857x device = testDevice.Device;
+            if (device.PinCount == 8)
+            {
+                Assert.Equal(0x00, device.ReadByte());
+            }
+            else
```
krwq: I'd add Assert `device.PinCount == 8 || device.PinCount == 16` or do explicit `else if` and `else` which throws

src/devices/Pcx857x/tests/ConstructionTests.cs
```diff
@@ -0,0 +1,28 @@
...
+            {
+                Pcx8575 device16 = (Pcx8575)device;
```
krwq: Is it possible to also do it without casting?

src/devices/Pcx857x/tests/Pcx857xTest.cs
```diff
@@ -0,0 +1,177 @@
...
+            {
+                for (int i = 0; i < pinValues.Length; i++)
+                {
+                    int pin = pinValues[i].PinNumber;
+                    pinValues[i] = new PinValuePair(pin, Read(pin));
+                }
+            }
+
+            public void SetPinMode(int pinNumber, PinMode mode)
+            {
```
krwq: If not expected to be called or expected to be called with specific values might be worth adding some assert here.

src/devices/Pcx857x/Pcx857x.cs
JeremyKuhne: Nice catch. I've updated this and the Mcp23xxx code where this came from. Also moving the helpers inline as they only have one caller.

src/devices/Pcx857x/tests/ConstructionTests.cs
JeremyKuhne: Sure, added the assert.

src/devices/Pcx857x/tests/ConstructionTests.cs
JeremyKuhne: No, as we only expose the 16 bit relevant apis on the 16 bit chips (`ReadUInt16`).

src/devices/Pcx857x/tests/Pcx857xTest.cs
JeremyKuhne: We don't care what it gets called with yet. When we add tests that track that we'll add an implementation.  I'll also be rejiggering this a bit to use MOQ in a later PR.

JeremyKuhne:
Note that the chip tested was on a WaveShare board.  https://www.waveshare.com/wiki/PCF8574_IO_Expansion_Board

src/devices/Pcx857x/Pcx857x.cs
```diff
@@ -0,0 +1,203 @@
...
+        public PinValue Read(int pinNumber)
+        {
+            Span<PinValuePair> values = stackalloc PinValuePair[]{ new PinValuePair(pinNumber, default) };
+            Read(values);
+            return values[0].PinValue;
+        }
+
+        public void Read(Span<PinValuePair> pinValues)
+        {
+            (uint pins, _) = new PinVector32(pinValues);
```
krwq: interesting usage :smile:

src/devices/Pcx857x/tests/Pcx857x.Tests.csproj
```diff
@@ -0,0 +1,28 @@
...
+
+    <IsPackable>false</IsPackable>
+
+    <LangVersion>7.3</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
+    <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
```
krwq: I presume this restores correctly on Linux?

JeremyKuhne: We already have this in other projects.

src/devices/Pcx857x/Pcx857x.cs
JeremyKuhne: Thanks 😝

src/devices/Pcx857x/Pcx857x.csproj
```diff
@@ -0,0 +1,33 @@
...
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems> <!--Disabling default items so samples source won't get build by the main library-->
+    <RootNamespace>Iot.Device.Pcx857x</RootNamespace>
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!--
+      Temporarily making a project reference until the daily has IGpioController
+      <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
```
joperezr: This shouldn't be necessary any longer, do you mind removing this workaround from the bindings that are using it please?

shaggygi:
@joperezr I have a PR for this already with a few other changes.   See #239.

Issue #87, opened 2018-12-08
Write operations should use ReadOnlySpan<T>


arkadiuszwojcik:
Right now write operation use `Span<byte>`. Is there any reason for that? Looks like we should use `ReadOnlySpan<byte>`

joperezr:
This makes sense, there is no particular reason why to use Span, so I agree that we should be using ReadOnlySpan for writing as well. I'm marking this up-for-grabs. @arkadiuszwojcik let me know if this is something that you would be interested in fixing.

JeremyKuhne:
They should also use "buffer" as the term. And single items should be "value". So:

``` C#
    public abstract class I2cDevice : IDisposable
    {
        protected I2cDevice();

        public abstract I2cConnectionSettings ConnectionSettings { get; }

        public void Dispose();
        public virtual void Dispose(bool disposing);
        public abstract void Read(Span<byte> buffer);
        public abstract byte ReadByte();
        // OLD: public abstract void Write(Span<byte> data);
        public abstract void Write(ReadOnlySpan<byte> buffer);
        // OLD: public abstract void WriteByte(byte data);
        public abstract void WriteByte(byte value);
    }
```

See https://apisof.net/ to find existing patterns. https://apisof.net/catalog/System.IO.Stream.WriteByte(Byte) for an example.

shaggygi:
I had a related discussion with @joperezr about data/buffer during an earlier PR.  He might want to chime in on the thinking/naming.

JeremyKuhne:
Note that I am not the API review process myself, but what I'm pointing out is based on my experience with that process. Which is hopefully going to get us closer than we would be otherwise. 😆

krwq:
@JeremyKuhne I've sent PR with the fix but did not include any naming changes

We should re-iterate on all of the APIs at one point separately

joperezr:
I didn't know that we had an existing pattern for Write where we also called it buffer. What @shaggygi and I discussed in the past, was that it made sense for reading for it to be a buffer since that is where the API will be writing into, but thought that for writing it wasn't really a buffer you fill, but instead just data that you write which is why we had it like that. If we have an existing pattern in the framework, then we should use that instead.

PR #232, opened 2019-02-14
Added nRF24L01 binding

ZhangGaoxing:
Single chip 2.4 GHz Transceiver


joshfree:
Thank you for your contribution @ZhangGaoxing 😄 

/cc @krwq @tarekgh @joperezr @buyaa-n

src/devices/Nrf24l01/Command.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Nrf24l01
+{
+    internal enum Command : byte
+    {
+        NRF_R_REGISTER = 0x00,
```
joshfree: could you add comments for these command values?

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,773 @@
...
+    /// </summary>
+    public class Nrf24l01 : IDisposable
+    {
+        private GpioController _gpio = null;
+        private SpiDevice _sensor = null;
+
+        private readonly int _ce;
+        private readonly int _irq;
+
+        private readonly byte[] _empty = new byte[0];
```
joshfree: should this be static

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,773 @@
...
+        }
+
+        /// <summary>
+        /// Send
+        /// </summary>
+        /// <param name="data">Send Data</param>
+        public void Send(byte[] data)
+        {
+            SetWorkingMode(WorkingMode.TX);
+            Thread.Sleep(4);
```
joshfree: how did you pick the values 4, 1, 10, etc?

src/devices/Nrf24l01/Nrf24l01.cs
tarekgh: also can use Array.Empty instead of new byte[0]

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,773 @@
...
+        /// <param name="pipe">Pipe, form 0 to 5</param>
+        /// <param name="address">Address, if (pipe > 1) then (address.Length = 1), else if (pipe = 1 || pipe = 0) then (address.Length ≤ 5)</param>
+        internal void SetRxAddress(byte pipe, Span<byte> address)
+        {
+            if (pipe > 5 || pipe < 0)
+            {
+                throw new ArgumentOutOfRangeException("pipe", "pipe needs to be in the range of 0 to 5!");
+            }
+
+            if (address.Length > 5)
```
joshfree: doesn't address.Length also have to be >= 1?

src/devices/Nrf24l01/WorkingMode.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// nRF24L01 Working Mode
+    /// </summary>
+    public enum WorkingMode
+    {
+        /// <summary>
+        /// Receive Mode
+        /// </summary>
+        RX = 1,
```
joshfree: nit/question: Should these be Receive and Transmit instead of RX/TX?

src/devices/Nrf24l01/Nrf24l01.cs
JeremyKuhne: Note that sleeping for this amount of time isn't really possible, particularly on Windows (12ms+ is likely what you'll get). I'll add some helpers to \Common that will allow you to delay more appropriately as we're hitting this a ton.  I'll try and get these up today.

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,773 @@
...
+        public byte PacketSize
+        {
+            get { return _packetSize; }
+            set { _packetSize = value < 0 || value > 32 ? throw new ArgumentOutOfRangeException("PacketSize needs to be in the range of 0 to 32!") : value; }
+        }
+
+        /// <summary>
+        /// nRF24L01 Send Address
+        /// </summary>
+        public Span<byte> Address { get => ReadTxAddress(); set => SetTxAddress(value); }
```
krwq: I think this should be `void ReadAddress(Span<byte> address)` or `byte[] Address` returning span only makes sense when it is a subspan of the input argument

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,773 @@
...
+            Span<byte> readBuf = new byte[1 + dataLength];
+
+            writeBuf[0] = (byte)((byte)command + (byte)register);
+
+            _sensor.TransferFullDuplex(writeBuf, readBuf);
+
+            return readBuf.Slice(1);
+        }
+
+        internal Span<byte> WriteRead(Command command, Register register, Span<byte> writeData)
```
krwq: I thing this should not return span: Span should be an input to function where user can decide where the data should be written or a subspan of some input span, otherwise it is better to return array (in this case I'd just take this as an input)

src/devices/Nrf24l01/Nrf24l01.cs
JeremyKuhne: In this case `Span<byte>.Empty` at the call site is probably the best choice. I believe this is only trying to pass an empty Span.

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,773 @@
...
+        /// <summary>
+        /// Set nRF24L01 Receive Packet Size
+        /// </summary>
+        /// <param name="pipe">Pipe, form 0 to 5</param>
+        /// <param name="payload">Size, from 0 to 32</param>
+        internal void SetRxPayload(byte pipe, byte payload)
+        {
+            if (payload > 32 || payload < 0)
+            {
+                throw new ArgumentOutOfRangeException("payload", "payload needs to be in the range of 0 to 32!");
```
JeremyKuhne: nit: use `nameof(payload)` when referring to argument names. Also don't use exclamation points. So, in this case:

``` C#
throw new ArgumentOutOfRangeException(nameof(payload), $"{nameof(payload)} needs to be in the range of 0 to 32.");
```

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,773 @@
...
+            return readBuf.Slice(1);
+        }
+
+        internal Span<byte> WriteRead(Command command, Register register, Span<byte> writeData)
+        {
+            Span<byte> writeBuf = stackalloc byte[1 + writeData.Length];
+            Span<byte> readBuf = new byte[1 + writeData.Length];
+
+            writeBuf[0] = (byte)((byte)command + (byte)register);
+            for (int i = 0; i < writeData.Length; i++)
```
JeremyKuhne: This should be `writeData.CopyTo(writeBuf.Slice(1));`  You'll get the best performance that way.

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,773 @@
...
+        #region SpiSettings
+        /// <summary>
+        /// NRF24L01 SPI Clock Frequency
+        /// </summary>
+        public const int SpiClockFrequency = 2000000;
+
+        /// <summary>
+        /// NRF24L01 SPI Mode
+        /// </summary>
+        public const SpiMode SpiMode = System.Device.Spi.SpiMode.Mode0;
```
buyaa-n: don't think you need to specify System.Device.Spi namespace

src/devices/Nrf24l01/Nrf24l01.cs
JeremyKuhne: As @krwq points out the pattern we use is to take all `Span` buffers as input. If you want to trim off the command byte without an extra copy you can also pass back a sliced span of the passed output buffer.

Additionally input buffers should be `ReadOnlySpan<>`. Our `I2cDevice` and `SpiDevice` need to be fixed as well, they should be `Write(ReadOnlySpan<byte> buffer)`. We need to clean up https://github.com/dotnet/iot/issues/87 so we're not setting a bad example here. :)

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,773 @@
...
+            writeBuf[0] = (byte)((byte)command + (byte)register);
+
+            _sensor.TransferFullDuplex(writeBuf, readBuf);
+
+            return readBuf.Slice(1);
+        }
+
+        internal Span<byte> WriteRead(Command command, Register register, Span<byte> writeData)
+        {
+            Span<byte> writeBuf = stackalloc byte[1 + writeData.Length];
```
JeremyKuhne: What sort of buffer sizes can we expect here? If they're large they can overflow the stack. If we get large ones in we can rent from `ArrayPool` to reduce allocations and return to the `ArrayPool` at the end of the method.

Solutions for when the inputs are unbounded:

1. Just create a new array (wasteful, but simplest)
2. Always rent the array from the `ArrayPool` (`ArrayPool<byte>.Shared.Rent()`) and return them at the end of the method.
3. Stackalloc if the buffer is small enough (say, under 32 bytes) and otherwise do (1) or (2).

src/devices/Nrf24l01/Nrf24l01.cs
krwq: the reasoning for that is so that tools can work correctly when refactoring (i.e. VS will rename variable correct) - updating string won't happen automatically while nameof works fine

src/devices/Nrf24l01/Nrf24l01.cs
ZhangGaoxing: To use Span<T>, you need to set it as a property. I will change it to Array.Empty

src/devices/Nrf24l01/Nrf24l01.cs
ZhangGaoxing: got it

src/devices/Nrf24l01/Nrf24l01.cs
ZhangGaoxing: If I don't do this, VS will report an error

src/devices/Nrf24l01/Nrf24l01.cs
ZhangGaoxing: Maximum length is 5 bytes😁

krwq: @ZhangGaoxing in that case please add Debug.Assert directly above :smile:

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -712,11 +712,14 @@ internal byte ReadChannel()
...
 
         #region sensor operation
-        internal void Write(Span<byte> writeData)
+        internal void Write(ReadOnlySpan<byte> writeData)
         {
             Span<byte> readBuf = stackalloc byte[writeData.Length];
+            Span<byte> writeBuf = stackalloc byte[writeData.Length];
 
-            _sensor.TransferFullDuplex(writeData, readBuf);
+            writeData.CopyTo(writeBuf);
```
krwq: can you just directly pas writeData to TransferFullDuplex?

ZhangGaoxing: VS will report wrong

ZhangGaoxing: VS will report wrong

krwq: Interesting, that sounds like a bug in the Spi abstraction: https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Spi/SpiDevice.cs#L51 they should be using ReadOnlySpan

krwq: @ZhangGaoxing I've sent PR with the fix: https://github.com/dotnet/iot/pull/240

krwq: @ZhangGaoxing this is merged now, you can update the PR

ZhangGaoxing:
I think we can continue this PR😄

src/devices/Nrf24l01/Nrf24l01.cs
krwq: @ZhangGaoxing good questions here - delay helpers also merged: https://github.com/dotnet/iot/blob/master/src/devices/Common/System/Device/DelayHelper.cs

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,767 @@
...
+            _gpio.Write(_ce, PinValue.Low);
+
+            Span<byte> writeData = stackalloc byte[length];
+            Span<byte> readData = WriteRead(Command.NRF_R_RX_PAYLOAD, Register.NRF_NOOP, writeData);
+
+            Span<byte> ret = new byte[readData.Length];
+            readData.CopyTo(ret);
+
+            _gpio.Write(_ce, PinValue.Low);
+            Write(Command.NRF_W_REGISTER, Register.NRF_STATUS, 0x4E);
```
krwq: please add comment on what 4E means

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,767 @@
...
+        /// Initialize
+        /// </summary>
+        private void Initialize(PinNumberingScheme pinNumberingScheme, OutputPower outputPower, DataRate dataRate, byte channel)
+        {
+            _gpio = new GpioController(pinNumberingScheme);
+            _gpio.OpenPin(_ce, PinMode.Output);
+            _gpio.OpenPin(_irq, PinMode.Input);
+            _gpio.RegisterCallbackForPinValueChangedEvent(_irq, PinEventTypes.Falling, Irq_ValueChanged);
+
+            Thread.Sleep(50);
```
krwq: some comment why this much

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,767 @@
...
+
+            // Details in the datasheet P53
+            _gpio.Write(_ce, PinValue.Low);
+            Write(Command.NRF_FLUSH_TX, Register.NRF_NOOP, _empty);
+            Write(Command.NRF_FLUSH_RX, Register.NRF_NOOP, _empty);
+            Write(Command.NRF_W_REGISTER, Register.NRF_CONFIG, 0x3B);
```
krwq: comment on 3B

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -0,0 +1,767 @@
...
+        /// Set nRF24L01 Auto Acknowledgment (All Pipe)
+        /// </summary>
+        /// <param name="isAutoAck">Is Enable</param>
+        internal void SetAutoAck(bool isAutoAck)
+        {
+            _gpio.Write(_ce, PinValue.Low);
+
+            if (isAutoAck)
+            {
+                Write(Command.NRF_W_REGISTER, Register.NRF_EN_AA, 0x3F);
```
krwq: please describe bits in this register and if makes sense please use binary literal (`0b0000_0000`)

krwq:
LGTM but needs some comments on magic numbers

src/devices/Nrf24l01/Nrf24l01.cs
ZhangGaoxing: Actually, it doesn't make any special sense. I just don't want the program to go too fast. Now I change these values to 1.  I don't think DelayHelper is needed because these are not exact values.

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -313,15 +318,16 @@ internal byte ReadRxPayload(byte pipe)
         /// <param name="isAutoAck">Is Enable</param>
         internal void SetAutoAck(bool isAutoAck)
         {
+            // Details in the Datasheet P53
             _gpio.Write(_ce, PinValue.Low);
 
             if (isAutoAck)
             {
-                Write(Command.NRF_W_REGISTER, Register.NRF_EN_AA, 0x3F);
+                Write(Command.NRF_W_REGISTER, Register.NRF_EN_AA, 0b_0011_1111);        // all pipe
```
krwq: comments should be before the line on the same indentation level

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -545,11 +562,13 @@ internal WorkingMode ReadWorkwingMode()
...
         internal void SetOutputPower(OutputPower power)
         {
+            // Details in the Datasheet P54
+            // Register: 0x06, bit[2:1]
             _gpio.Write(_ce, PinValue.Low);
 
             Span<byte> readData = WriteRead(Command.NRF_R_REGISTER, Register.NRF_RF_SETUP, 1);
 
-            byte setting = (byte)(readData[0] & (~0x06) | ((byte)power << 1));
+            byte setting = (byte)(readData[0] & (~0b_0000_0110) | ((byte)power << 1));
```
krwq: why not directly write negated bits?

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -200,23 +200,25 @@ private void Irq_ValueChanged(object sender, PinValueChangedEventArgs args)
...
         private void Initialize(PinNumberingScheme pinNumberingScheme, OutputPower outputPower, DataRate dataRate, byte channel)
         {
+            // open pins
             _gpio = new GpioController(pinNumberingScheme);
             _gpio.OpenPin(_ce, PinMode.Output);
             _gpio.OpenPin(_irq, PinMode.Input);
             _gpio.RegisterCallbackForPinValueChangedEvent(_irq, PinEventTypes.Falling, Irq_ValueChanged);
 
-            Thread.Sleep(50);
+            Thread.Sleep(10);       // wait for some time
```
krwq: why do you need to wait? perhaps this is something we can fix on our side?

src/devices/Nrf24l01/Nrf24l01.cs
krwq: actually whichever since this can be thought of as "clear these bits"

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -562,9 +581,10 @@ internal void SetOutputPower(OutputPower power)
         /// <returns>Output Power</returns>
         internal OutputPower ReadOutputPower()
         {
+            // Details in the Datasheet P54
             Span<byte> readData = WriteRead(Command.NRF_R_REGISTER, Register.NRF_RF_SETUP, 1);
 
-            return (OutputPower)(readData[0] & 0x06 >> 1);
+            return (OutputPower)(readData[0] & 0b_0000_0110 >> 1);
```
krwq: brackets please since operator precedence is not immediately obvious

src/devices/Nrf24l01/Nrf24l01.cs
ZhangGaoxing: I just want to make the operation meaningful and readable.

src/devices/Nrf24l01/Nrf24l01.cs
ZhangGaoxing: No, I just think we need to give the hardware a little reaction time to make sure it works correctly.😅

krwq:
Thanks @ZhangGaoxing!

Issue #187, opened 2019-01-31
High precision timer waiting 1 microsecond


Ellerbach:
Hi,
For the implementation of the DHT sensor family, I had to wait for 1 microsecond. The only reliable way I found is a for loop doing nothing.
The values I found to be working on a Raspberry Pi 3 are 99 for release mode and 27 for debug.
So 1 waiting 1 microsecond in release mode would then be:
```csharp
for (byte wt = 0; wt < 99; wt++)
     ;
```

It would be great to have the possibility to find those values at execution time for any platform or have a real way to wait for 1 microsecond with a specific high precision timer. This is needed for a lot of sensors and is quite a challenge in managed code.

The mockup code to find those values is below. The core idea is to use stopwatch. First step is to calibrate how much tricks stopwatch is taking to start and stop.
Second step is to run the for loop with the specific value to check the behavior.
Please note that in all the cases, the first and second time, values are not that accurate, so it’s better to start after a third time. Tests shows that running 100 times is largely enough. Precision is around 5% which is more than acceptable for most sensor.
The next step for this mockup would be to find the values by iteration. But there maybe a much better and much simpler way to do it.
And it would anyway be great to have an inline function or equivalent to do this microsecond, high precision timer. 

```Csharp
using System;
using System.Diagnostics;

namespace TestNetCore
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");
            Stopwatch stopwatch = Stopwatch.StartNew();
            for (int i = 0; i < 30; i++)
                ;
            stopwatch.Stop();
            Console.WriteLine($"First run: {stopwatch.ElapsedTicks}");

            //Run it 10 times and take an average
            double origin = 0;
            for (int steps = 0; steps < 100; steps++)
            {
                stopwatch = new Stopwatch();
                stopwatch.Start();
                stopwatch.Stop();
                // Console.WriteLine($"Start and stop: {stopwatch.ElapsedTicks}");
                origin = (origin * steps + stopwatch.ElapsedTicks)/(steps + 1);
            }
            
            Console.WriteLine($"Origin: {origin}");
            double average = 0;
            for (int steps = 0; steps < 99; steps++)
            {
                stopwatch = new Stopwatch();
                stopwatch.Start();
                for (int i = 0; i < 98; i++)
                    ;
                stopwatch.Stop();
                Console.WriteLine($"Total time: {stopwatch.ElapsedTicks}");
[...]

PR #235, opened 2019-02-15
Add delay helpers.

JeremyKuhne:
`Thread.Sleep()` isn't really what we want when trying to pulse pins, etc. Creating a helper for Common that we can centralize our wait logic and iterate on.

#187 mentions some of the sleep issues. We are not and cannot be a real-time framework, but we can provide tighter waits than `Thread.Sleep` and don't want to have a dozen different delay implementations in the device repo.


src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -525,34 +526,29 @@ private void WriteBits(byte value, int count)
...
+
+            // Leaving the flag here to make sure calling code is describing when it cannot
+            // actually check the busy state should we be able to get performance
             // to the point where checking would be a net benefit.
 
             // Note that on a Raspberry Pi 3B+ we average about 1.5μs when delaying for one μs.
             // 
             // SpinWait currently spins to approximately 1μs before it will yield the thread.
             // Thread.Yield() takes around 1.2μs. This gives us a fidelity of microseconds to
             // microseconds + 1.4.
```
kouvel: Should this part of the comment about `SpinWait`, and `we average about 1.5μs when delaying for one μs`, be removed or modified?

src/devices/Common/System/Device/DelayHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// <summary>
+        /// Delay for at least the specified <paramref name="microseconds"/>.
+        /// </summary>
+        /// <param name="microseconds">The number of microseconds to delay.</param>
+        /// <param name="allowThreadYield">True to allow yielding the thread.</param>
+        public static void DelayMicroseconds(int microseconds, bool allowThreadYield = true)
+        {
+            long start = Stopwatch.GetTimestamp();
+            long end = start + (microseconds * Stopwatch.Frequency / 1_000_000);
```
kouvel: The add may overflow. It is probably safer to compute the duration in ticks and in the loop check the elapsed ticks from the start against the desired duration. Same for `DelayMilliseconds()`.

src/devices/Common/System/Device/DelayHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            if (!allowThreadYield)
+            {
+                while (Stopwatch.GetTimestamp() < end)
```
kouvel: Minor: Maybe use a do/while loop in all of these places, as it's very unlikely that the desired duration would have elapsed between two `GetTimestamp()` calls in quick succession.

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -525,34 +526,29 @@ private void WriteBits(byte value, int count)
             // and 230ns on 5V. (PWeh on page 49/52 and Figure 25 on page 58)
 
             _controller.Write(_enablePin, PinValue.High);
-            DelayMicroseconds(1);
+            DelayMicroseconds(1, allowThreadYield: false);
             _controller.Write(_enablePin, PinValue.Low);
         }
 
-        private void DelayMicroseconds(int microseconds, bool checkBusy = true)
+        private void DelayMicroseconds(int microseconds, bool allowThreadYield = true,  bool checkBusy = true)
```
kouvel: Since there is only one call to this at the moment, is the `allowThreadYield` parameter necessary? If so would a default of `false` or no default (to require explicit specification by caller) be more appropriate?

kouvel:
Also should there be some consideration for single-proc systems? `Thread.SpinWait(1)` on uniproc systems would not allow other work to run during the delay. I don't think there is a good alternative in that case other than `Thread.Sleep(0)`. If a busy-loop is ok then `Thread.SpinWait(1)` would be ok.

src/devices/Common/System/Device/DelayHelper.cs
JeremyKuhne: Ok

JeremyKuhne:
> Also should there be some consideration for single-proc systems?

That's the thing I'm trying for here- is trying to get the timing as close as possible when it really matters. Presumably that means we have to starve the other threads. This area isn't my strongest, so any feedback you have is greatly appreciated. :)

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: I'll make it explicit.

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: Yeah, I'll remove it.

src/devices/Common/System/Device/DelayHelper.cs
JeremyKuhne: I was trying to minimize the amount of work done in the loop. I'll redo it.

krwq: I think if you just keep on subtracting current time and start time (or current and end time) you can safely overflow since it will cancel each other

src/devices/Common/System/Device/DelayHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// <summary>
+        /// Delay for at least the specified <paramref name="milliseconds"/>
+        /// </summary>
+        /// <param name="milliseconds">The number of milliseconds to delay.</param>
+        /// <param name="allowThreadYield">True to allow yielding the thread.</param>
+        public static void DelayMilliseconds(int milliseconds, bool allowThreadYield = true)
+        {
+            long start = Stopwatch.GetTimestamp();
+            long end = start + (milliseconds * Stopwatch.Frequency / 1_000_000);
```
krwq: can the implementation use Thread.Sleep? It doesn't seem to need much resolution and I think this should work for most of the cases

JeremyKuhne: Sleeping the thread takes multiple milliseconds.  I'm not sure if we can count on when the thread comes back either- hence the `allowThreadYield`.

src/devices/Common/System/Device/DelayHelper.cs
JeremyKuhne: > I think if you just keep on subtracting current time and start time

I'm not following

JeremyKuhne: Ok, updated. Please check my brain. XD

JeremyKuhne:
Ok- I've updated again. If I could get another sanity check I'd appreciate it. :)

src/devices/Common/System/Device/DelayHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            long start = Stopwatch.GetTimestamp();
+            ulong elapsed = (ulong)(microseconds * Stopwatch.Frequency / 1_000_000);
+
+            if (!allowThreadYield)
+            {
+                do
+                {
+                    Thread.SpinWait(1);
+                } while (elapsed < (ulong)(Stopwatch.GetTimestamp() - start));
```
kouvel: Should this be `>` instead? The right side is actually the elapsed time since start and the left side is the delay duration. Similarly elsewhere.

kouvel:
> Presumably that means we have to starve the other threads.

I can't think of any other way, you're probably right

src/devices/CharacterLcd/Hd44780.cs
kouvel: Did you intend to make this change?

src/devices/Common/System/Device/DelayHelper.cs
krwq: I think the check is correct but naming is confusing :) as @kouvel points out elapsed is the right side and left side is requested delay

kouvel: I think what we would want is `do ... while(elapsed < duration);`, since the condition is reversed above, the equivalent would then be `do ... while(duration > elapsed);`, isn't it?

JeremyKuhne: Yes, my terms are terrible. Updating to make it correct and hopefully clearer.

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: Yes, there are multiple callers and only one. I made the helper explicit. I should have put this one at the end. To make this clearer I'm changing the name of the helper here to `WaitForNotBusy` and calling the new helpers directly when they aren't looking for not busy (in theory).

JeremyKuhne:
I'm writing some other drivers with this. Once this is approved and I've validated I'll check in.

src/devices/Common/System/Device/DelayHelper.cs
krwq: @kouvel right :smile:

src/devices/Common/System/Device/DelayHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+        // GetTimestamp() currently can take ~300ns. We hope to improve this to get better
+        // fidelity for very tight spins.
+        //
+        // SpinWait currently spins to approximately 1μs before it will yield the thread.
+
+        /// <summary>
+        /// Delay for at least the specified <paramref name="microseconds"/>.
+        /// </summary>
+        /// <param name="microseconds">The number of microseconds to delay.</param>
+        /// <param name="allowThreadYield">True to allow yielding the thread. On single-proc systems this will prevent all other code from running.</param>
```
vancem: I think the comment is was saying the opposite of what it should.   
```suggestion
        /// <param name="allowThreadYield">True to allow yielding the thread. If this is set to false, on a single-proc systems this will prevent all other code from running.</param>
```

src/devices/Common/System/Device/DelayHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+                    spinWait.SpinOnce();
+                } while ((ulong)(Stopwatch.GetTimestamp() - start) < minimumTicks);
+            }
+        }
+
+        /// <summary>
+        /// Delay for at least the specified <paramref name="milliseconds"/>
+        /// </summary>
+        /// <param name="milliseconds">The number of milliseconds to delay.</param>
+        /// <param name="allowThreadYield">True to allow yielding the thread. On single-proc systems this will prevent all other code from running.</param>
```
vancem: Same issue with comment that I mentioned above

vancem:
What do you have DelayMilliseconds?   Seems like DelayMicroseconds is enough and they basically do the same thing.   The only value I could think of is that DelayMilliseconds has a bigger range, but you already get 2000 seconds with DelayMicroseconds, which seems like more than enough.  

Also this is an internal helper at the moment.  Is it intended to be made public eventually?   Is DelayHelper the intended name (seems like adding it to SpinWait  may make more sense).

Also finally when you have a method parameter (allowThreadYield) that completely changes the implementation, it makes you wonder if you should just have two methods (e.g. SpinDelay and YieldingDelay, which makes things more explicit.

JeremyKuhne:
> What do you have DelayMilliseconds?

Clarity/correctness in consuming code is one hope. There are a ton of waits in some of the IOT code. Datasheets use both micro and milliseconds quite often. I've seen a number of transliteration mistakes already.

The other thing is that I expect that we might add additional logic to the millisecond one based on gained experience (like maybe, you're waiting more than 100ms, start here...). When you're waiting 1 microsecond there isn't a lot of time to insert logic.

> Also this is an internal helper at the moment. Is it intended to be made public eventually? Is DelayHelper the intended name (seems like adding it to SpinWait may make more sense).

I presume that we'll want to make it public. DelayHelper isn't a thought-out public name. :)

> it makes you wonder if you should just have two methods

I agree it needs more thought. I'll watch how it gets used and iterate.

vancem:
> What do you have DelayMilliseconds?
>>  Clarity/correctness in consuming code is one hop

I would probably get rid of DelayMilliseconds, but if you do find it valuable I guess it is not a big deal.   Why can't we just declare it to be DelayMicroseconds(time*1000) at least for now (later we may change it but right now they really are the same thing.   Even if DelayMilliseconds ends up having more complex logic, to the degree we can define it in terms of the DelayMicroseconds primitive I think that would ge good.

For what it is worth...

JeremyKuhne:
> Why can't we just declare it to be DelayMicroseconds(time*1000) at least for now

Sure

Issue #157, opened 2019-01-25
Remove DllNotFoundException catching logic after Native Library Loader API deployed 


buyaa-n:
When loading unix driver we try to load libgpiod driver first, if it is not installed DllNotFoundException catched, in that case we load Sys Fs driver: https://github.com/dotnet/iot/pull/133/files#diff-b4f508a799552fbc32cd5d70b229dc86R19. Remove/update this logic with appropriate one after "Native Library Loader API" deployed https://github.com/dotnet/corefx/issues/32015

buyaa-n:
"Native Library Loader API" is shipped with 3.0 preview dotnet/corefx#32015
So i have tried to use 3.0 preview for iot project and fix this issue. But it turns out we cannot use 3.0 preview from iot project. The project TargetFramework is set to netstandard2.0 and Santi suggested to not change it until it is released as stable version. I will lookup if I could do some work around but most likely we will wait for full release of 3.0

joperezr:
I agree with @safern, In fact even when 3.0 ships, I think we will still want to be able to work in netstandard2.0 so we would either multi-target or stay with netstandard 2.0 target only. What I mean is that I don't believe that this is an important enough reason that would cause us want to change the platform target.

safern:
Correct. Actually I meant until .netstandard2.1 is available, rather than .netcoreapp3.0 ships.

krwq:
Specific piece of code this is about: https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs#L81

krwq:
[Triage] Closing because it's been open for 3 years and it doesn't seem we care about it enough

PR #236, opened 2019-02-15
Libgpiod event handling rafactored/comments addressed

buyaa-n:


src/Native/System.Device.Gpio.Native/libgpiod.c
```diff
@@ -138,3 +138,41 @@ extern struct gpiod_chip * GetNextChipFromChipIterator(struct gpiod_chip_iter *i
...
 	return gpiod_chip_iter_next(iter); 
 }
+
+/**
+ * @brief Request all event type notifications on a single line.
+ * @param line GPIO line object.
+ * @param consumer Name of the consumer.
+ * @return 0 if the operation succeeds, -1 on failure.
+ */
+extern int RequestBothEdgeEventForLine(struct gpiod_line *line, const char *consumer)
```
krwq: `Event`->`Events`

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -35,9 +39,79 @@ public LibGpiodDriver()
...
+                else
+                {
+                    throw ExceptionHelper.GetArgumentException(ExceptionResource.InvalidEventType);
+                }
+            }
+
+            if (eventSuccess < 0)
+            {
+                throw ExceptionHelper.GetIOException(ExceptionResource.RequestEventError, pinNumber, Marshal.GetLastWin32Error());
+            }
```
krwq: nit: enter after `}`

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -35,9 +39,79 @@ public LibGpiodDriver()
...
+            return pinHandle;
+        }
+
+        private Task InitializeEventDetectionTask(LibGpiodDriverEventHandler eventHandler, CancellationToken token)
+        {
+            Task pinListeningThread = Task.Run(() =>
```
krwq: you should also pass cancellation token to the Task.Run - this will handle the case where event is already cancelled before task event starts and prevent the task from starting

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -35,9 +39,79 @@ public LibGpiodDriver()
...
+            {       
+                int waitResult = -2;
+                while (!token.IsCancellationRequested)
```
krwq: you should also have a flag cancelling when controller is disposed

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -35,9 +39,79 @@ public LibGpiodDriver()
...
+                throw ExceptionHelper.GetIOException(ExceptionResource.RequestEventError, pinNumber, Marshal.GetLastWin32Error());
+            }
+            return pinHandle;
+        }
+
+        private Task InitializeEventDetectionTask(LibGpiodDriverEventHandler eventHandler, CancellationToken token)
+        {
+            Task pinListeningThread = Task.Run(() =>
+            {       
+                int waitResult = -2;
```
krwq: this value doesn't seem to be used anywhere, any chance it could be local inside of loop?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -35,9 +39,79 @@ public LibGpiodDriver()
...
+                while (!token.IsCancellationRequested)
+                {
+                    waitResult = Interop.WaitForEventOnLine(eventHandler.PinHandle);
```
krwq: could you add a comment what does this return? -1 seem to be the error, 1 that we get an event - what does anything else mean?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -35,9 +39,79 @@ public LibGpiodDriver()
...
+                    {
+                        int readResult = Interop.ReadEventForLine(eventHandler.PinHandle);
+                        if (readResult == -1)
+                        {
+                            throw ExceptionHelper.GetIOException(ExceptionResource.EventReadError, Marshal.GetLastWin32Error());
+                        }
+
+                        PinEventTypes eventType = (readResult == 1) ? PinEventTypes.Rising : PinEventTypes.Falling;
+                        var args = new PinValueChangedEventArgs(eventType, eventHandler.PinNumber);
+                        eventHandler?.OnPinValueChanged(args, eventType);
```
krwq: This should be ThreadPool.QueueUserWorkItem. See this for an example:
https://github.com/dotnet/corefx/blob/master/src/System.IO.Ports/src/System/IO/Ports/SerialStream.Unix.cs#L630

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -35,9 +39,79 @@ public LibGpiodDriver()
...
+                    }
+                }
+            });
+            return pinListeningThread;
+        }
+
         protected internal override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEventTypes eventType, PinChangeEventHandler callback)
         {
-            throw new NotImplementedException();
+            if (!_pinNumberToEventHandler.TryGetValue(pinNumber, out LibGpiodDriverEventHandler eventHandler))
```
krwq: Will this work if user tries to add 2 callbacks for the same event? (i.e. seperate for falling and seperate for raising or i.e. one for logging the other for logic)

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -95,7 +169,25 @@ protected internal override PinValue Read(int pinNumber)
...
-            throw new NotImplementedException();
+            if (_pinNumberToEventHandler.TryGetValue(pinNumber, out LibGpiodDriverEventHandler eventHandler))
+            {
+                eventHandler.ValueFalling -= callback;
+                eventHandler.ValueRising -= callback;
+                if (eventHandler.IsCallbackListEmpty())
+                {
+                    if (_pinNumberToEventHandler.TryRemove(pinNumber, out eventHandler))
+                    {
+                        eventHandler.CancellationTokenSource.Cancel();
```
krwq: this part of code should be locked - if someone adds a new event when we get here they will never get any events

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -142,9 +272,17 @@ protected override void Dispose(bool disposing)
...
                 {
                     Interop.ReleaseGpiodLine(pinHandle);
+                    _pinNumberToSafeLineHandle.Remove(pin);
                 }
+            }
 
-                _pinNumberToSafeLineHandle.Remove(pin);
+            foreach (int pin in _pinNumberToEventHandler.Keys)
+            {
+                if (_pinNumberToEventHandler.TryRemove(pin, out LibGpiodDriverEventHandler pinEventHandler)) {
```
krwq: you should have a function for those few lines of code since you have them in at least 2 places

src/System.Device.Gpio/System/Device/Gpio/LibgpiodDriverEventHandler.Linux.cs
```diff
@@ -0,0 +1,55 @@
...
+            if (PinHandle != null)
+            {
+                Interop.ReleaseGpiodLine(PinHandle);
+                PinHandle.Dispose();
+            }
+            ValueRising = null;
+            ValueFalling = null;
+        }
+    }
+}
```
krwq: I think we should have some tests for at least the events - including couple of stress tests where you try to break your code.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: sure, good catch

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: it is concurrent dictionary, do you think we still need a lock?

krwq: concurrent dictionary only protects against not destroying the internals of the dictionary - not against the state itself. you might want to consider using regular dictionary and lock to have more control on when the locks happen

src/System.Device.Gpio/System/Device/Gpio/LibgpiodDriverEventHandler.Linux.cs
buyaa-n: I have tested some level on my machine, will try to add more test cases to the GpioControllerTestBase

src/System.Device.Gpio/System/Device/Gpio/ExceptionHelper.cs
```diff
@@ -73,6 +73,9 @@ private static string GetResourceString(ExceptionResource resource, int errorCod
                 case ExceptionResource.NotListeningForEventError:
                     message = $"Attempted to remove a callback for a pin that is not listening for events.";
                     break;
+                case ExceptionResource.LibGpiodNotInstalled:
+                    message = $"Libgpiod driver not installed. More information on: https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/about/ ";
```
buyaa-n: @joshfree as you suggested updated to give more meaningful message when libgpiod not installed, but not sure about exact wording of the message...

krwq: Looks fine, There is an extra space in the end though.
Alternative: "Libgpiod driver not present. Kernel version X.Y.Z required."

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -111,5 +111,37 @@ internal partial class Interop
...
+    /// <summary>
+    /// Read the last event from the GPIO line.
+    /// </summary>
+    /// <param name="line">GPIO line handle</param>
+    /// <returns>1 if rising edge event occured, 2 on falling edge, -1 on error.</returns>
+    [DllImport(LibgpiodLibrary, SetLastError = true)]
+    internal static extern int ReadEventForLine(SafeLineHandle line);
+}
+
+internal enum WaitEventResult
```
krwq: move to separate file

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -35,17 +49,89 @@ public LibGpiodDriver()
...
+            }
         }
         
         protected internal override void ClosePin(int pinNumber)
         {
             if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle pinHandle))
             {
-                Interop.ReleaseGpiodLine(pinHandle);
-                _pinNumberToSafeLineHandle.Remove(pinNumber);
+                //Event listeners might still waiting for event, don't release them
```
krwq: nit space after `//`

krwq: - the comment should be inside the inner `if`
- this could be just a single if with `&&`

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -122,7 +220,45 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
...
+                cancelWhenEventOccured.Cancel();
+            };
+            if (eventType.HasFlag(PinEventTypes.Rising))
+            {
+                eventHandler.ValueRising += callback;
+            }
+            else if (eventType.HasFlag(PinEventTypes.Falling))
+            {
+                eventHandler.ValueFalling += callback;
+            }
```
krwq: `else Debug.Assert(false, "eventType has invalid value");`

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -122,7 +220,45 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
...
+                TimedOut = !cancelWhenEventOccured.IsCancellationRequested,
+                EventType = eventType
+            };
+        }
+
+        private void WaitForEventResult(CancellationToken sourceToken, CancellationToken parentToken, CancellationToken eventOccuredToken,  PinEventTypes eventType)
+        {
+            while (!(sourceToken.IsCancellationRequested || parentToken.IsCancellationRequested || eventOccuredToken.IsCancellationRequested))
+            {
+                Thread.Sleep(1_000);
```
krwq: why such a long sleep?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -122,7 +220,45 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
...
+            {
+                eventHandler = new LibGpiodDriverEventHandler(pinNumber, new CancellationTokenSource());
+                if (_pinNumberToEventHandler.TryAdd(pinNumber, eventHandler))
+                {
+                    eventHandler.PinHandle = SubscribeForEvent(pinNumber, eventType);
+                    eventHandler.Task = InitializeEventDetectionTask(eventHandler, eventHandler.CancellationTokenSource.Token);
+                }
+            }
+
+            CancellationTokenSource cancelWhenEventOccured = new CancellationTokenSource();
```
krwq: could this be just a bool?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -137,22 +273,41 @@ protected internal override void Write(int pinNumber, PinValue value)
...
+
+            if (_pinNumberToEventHandler != null)
             {
-                if (_pinNumberToSafeLineHandle.TryGetValue(pin, out SafeLineHandle pinHandle))
+                foreach (var kv in _pinNumberToEventHandler)
                 {
-                    Interop.ReleaseGpiodLine(pinHandle);
+                    int pin = kv.Key;
+                    LibGpiodDriverEventHandler eventHandler = kv.Value;
+                    eventHandler.Dispose();
```
krwq: I'm wondering if event handler should own the pin here - that suggest that if I register for the event and use regular APIs at the same time and then unregister the event then I will find myself with the disposed pin?

src/System.Device.Gpio/System/Device/Gpio/LibgpiodDriverEventHandler.Linux.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace System.Device.Gpio.Drivers
+{
+    internal sealed class LibGpiodDriverEventHandler : IDisposable
+    {
+        public event PinChangeEventHandler ValueRising;
+
+        public event PinChangeEventHandler ValueFalling;
+
+        public int PinNumber;
```
krwq: I don't think this is used anywhere

krwq: except for the events these should be properties with `{ get; private set; }`

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -137,22 +273,41 @@ protected internal override void Write(int pinNumber, PinValue value)
 
         protected override void Dispose(bool disposing)
         {
-            foreach (int pin in _pinNumberToSafeLineHandle.Keys)
+            _disposing = disposing;
+
+            if (_pinNumberToEventHandler != null)
             {
-                if (_pinNumberToSafeLineHandle.TryGetValue(pin, out SafeLineHandle pinHandle))
+                foreach (var kv in _pinNumberToEventHandler)
```
ahsonkhan: nit: don't use var here.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -137,22 +273,41 @@ protected internal override void Write(int pinNumber, PinValue value)
...
 
-            if (_chip != null)
+            
+            if (_pinNumberToSafeLineHandle != null)
             {
-                _chip.Dispose();
-                _chip = null;
+                foreach (int pin in _pinNumberToSafeLineHandle.Keys)
+                {
+                    if (_pinNumberToSafeLineHandle.TryGetValue(pin, out SafeLineHandle pinHandle))
```
ahsonkhan: When would this return false?

buyaa-n: when pin value was not present for that pin

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: For my knowledge it would be great if some information/comparison given about advantage of using real type instead of var

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: I am wondering if that scenario could work for the controller. I have got pin not available error when trying to listening event for a pin set for Input or Output. With that i thinkg one cannot use a pin for regular pin API and event handling simultaniosly. I didn't exactly tested your scenario though ...

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: yes it is, good catch

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: 😛 thought it is 1 millisecond, good catch

src/System.Device.Gpio/System/Device/Gpio/ExceptionHelper.cs
buyaa-n: Thanks, not sure why you added "Kernel version X.Y.Z required", libgpiod still need to be installed manually even kernel version was above 4.8

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: oh I was mistaken, didn't work with Boolean, seems cannot use value type

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
ahsonkhan: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

> 10. We only use `var` when it's obvious what the variable type is (e.g. `var stream = new FileStream(...)` not `var stream = OpenStandardInput()`).

Essentially, use `var` when the code reviewer/reader can easily tell what the type is (statically). This  is useful since it removes redundant type information. Otherwise, explicitly use the type name. The advantage is that it makes reviewing code easier (by making it easy to understand and follow).

krwq: @ahsonkhan the kv is irrelevant since she assigns the Key and Values to locals with explicit type line later. I'm not seeing much value in typing that

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
krwq: that's actually not something I'd expect but if that is the limitation perhaps it might be worth throwing when you try to do both simulatanously

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
krwq: @buyaa-n in the function you will likely need to use ref bool

src/System.Device.Gpio/System/Device/Gpio/ExceptionHelper.cs
krwq: I see, I thought it was included in the kernel by default not a separate kernel module - makes sense in that case

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -20,6 +20,7 @@ public void ControllerCanTurnOnLEDs()
         {
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
+                Console.WriteLine("ControllerCanTurnOnLEDs run");
```
krwq: looks like you got some remainder from debugging?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -35,20 +50,94 @@ public LibGpiodDriver()
...
+            }
+
+            return pinHandle;
+        }
+
+        private Task InitializeEventDetectionTask(LibGpiodDriverEventHandler eventHandler, CancellationToken token)
+        {
+            return Task.Run(() =>
+            {       
+                while (!(token.IsCancellationRequested || _disposing))
```
krwq: alternative: `!token.IsCancellationRequested && !_disposing`

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -35,20 +50,94 @@ public LibGpiodDriver()
...
+            }, token);
+        }
+
         protected internal override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEventTypes eventType, PinChangeEventHandler callback)
         {
-            throw new NotImplementedException();
+            if (!_pinNumberToEventHandler.TryGetValue(pinNumber, out LibGpiodDriverEventHandler eventHandler))
+            {
+                eventHandler = new LibGpiodDriverEventHandler(pinNumber, new CancellationTokenSource());
+                if (_pinNumberToEventHandler.TryAdd(pinNumber, eventHandler))
```
krwq: I think you should put the next two lines in the constructor of the event handler and in here you can use lambda to generate value only when it is not present:
https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=netframework-4.7.2#System_Collections_Concurrent_ConcurrentDictionary_2_GetOrAdd__0_System_Func__0__1__

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: Yeah 😛 I noticed that too, but didn't fixed it because still trying to debug/run testing, will be removed

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: which one is better you think? for me mine better 😛

krwq: I'm fine either way with slight preference toward mine :)

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -31,6 +32,7 @@ public void ControllerCanTurnOnLEDs()
         [Fact]
         public void PinValueReturnsToLowAfterDispose()
         {
+            Console.WriteLine("PinValueReturnsToLowAfterDispose run");
```
joshfree: can we make the tests only output content on failure?  a passing test shouldn't fill the screen with output.

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -130,9 +138,10 @@ public void OpenPinDefaultsModeToInput()
         [Fact]
         public void AddCallbackTest()
         {
-            while (!Debugger.IsAttached)
-                Thread.Sleep(1000);
-            Debugger.Break();
+            Console.WriteLine("AddCallbackTest run");
+            /*          while (!Debugger.IsAttached)
```
joshfree: dead code

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: These Console writing rows committed accidentally, will be removed

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
joshfree: @buyaa-n shouldn't we never get in this state and validate the input parameters as part of accepting pin values prior to gettting all the way to Dispose?

src/System.Device.Gpio/System/Device/Gpio/Drivers/UnixDriver.Linux.cs
```diff
@@ -14,7 +14,7 @@ public static UnixDriver Create()
             {
                 driver = new LibGpiodDriver();
             }
-            catch (DllNotFoundException)
+            catch (PlatformNotSupportedException)
```
joshfree: Is it possible to have a static tryget factory method to return a driver and avoid the exception handling entirely?

src/System.Device.Gpio/System/Device/Gpio/ExceptionHelper.cs
```diff
@@ -53,6 +58,24 @@ private static string GetResourceString(ExceptionResource resource, int errorCod
...
+                    message = $"Error while waiting for event, error code {errorCode}, pin: {pin}";
+                    break;
+                case ExceptionResource.EventReadError:
+                    message = $"Error while reading pin event result, error code {errorCode}";
+                    break;
+                case ExceptionResource.NotListeningForEventError:
+                    message = $"Attempted to remove a callback for a pin that is not listening for events.";
+                    break;
+                case ExceptionResource.LibGpiodNotInstalled:
+                    message = $"Libgpiod driver not installed. More information on: https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/about/ ";
```
joshfree: there's a trailing space in this exception message

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -95,7 +182,19 @@ protected internal override PinValue Read(int pinNumber)
 
         protected internal override void RemoveCallbackForPinValueChangedEvent(int pinNumber, PinChangeEventHandler callback)
         {
-            throw new NotImplementedException();
+            if (_pinNumberToEventHandler.TryGetValue(pinNumber, out LibGpiodDriverEventHandler eventHandler))
+            {
+                eventHandler.ValueFalling -= callback;
+                eventHandler.ValueRising -= callback;
+                if (eventHandler.IsCallbackListEmpty())
```
krwq: should you also remove the handler from the dictionary?

buyaa-n: Definitely, good catch, thanks

src/System.Device.Gpio/System/Device/Gpio/Drivers/UnixDriver.Linux.cs
buyaa-n: Seems we don't need to catch DllNotFoundException anymore as https://github.com/dotnet/corefx/issues/32015 is implemented and we already have issue to fix this https://github.com/dotnet/iot/issues/157. So I am gonna address your comment with the #157

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -153,7 +150,75 @@ void callback(object sender, PinValueChangedEventArgs pinValueChangedEventArgs)
...
         }
 
+        [Fact]
+        public void AddCallbackRemoveCallbackTest()
+        {
+            int risingEventOccuredCount = 0, fallingEventOccuredCount = 0;
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+            {
+                controller.OpenPin(InputPin, PinMode.Input);
+                controller.OpenPin(OutputPin, PinMode.Output);
```
krwq: set the initial value for the output pin so it is more obvious what it is.

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -153,7 +150,75 @@ void callback(object sender, PinValueChangedEventArgs pinValueChangedEventArgs)
...
+                    }
+                });
+
+                controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Falling, (o, e) => {
+                    fallingEventOccuredCount++;
+                });
+                for (int i = 0; i < 10; i++)
+                {
+                    controller.Write(OutputPin, PinValue.High);
+                    Thread.Sleep(100);
```
krwq: why is sleep needed?

buyaa-n: Sleep needed to make sure hardware able to detect the event, about the large number -> was not really sure how much sleep is enough, so just put enough duration 😛

buyaa-n: set to sleep 10 milliseconds

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -153,7 +150,81 @@ void callback(object sender, PinValueChangedEventArgs pinValueChangedEventArgs)
...
+                controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Rising, (o, e) =>
+                {
+                        tokenSource.Cancel();
+                        cancelled = true;
+                });
+
+                Task.Run(() =>
+                {
+                    controller.Write(OutputPin, PinValue.High);
+                    Thread.Sleep(10);
```
krwq: what's the setting to low for? You seem to be waiting only for raising

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -153,7 +150,81 @@ void callback(object sender, PinValueChangedEventArgs pinValueChangedEventArgs)
...
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+            {
+                bool cancelled = false;
+                CancellationTokenSource tokenSource = new CancellationTokenSource();
+                controller.OpenPin(InputPin, PinMode.Input);
+                controller.OpenPin(LedPin, PinMode.Input);
+                controller.OpenPin(OutputPin, PinMode.Output);
+                controller.Write(OutputPin, PinValue.Low);
+
+                controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Rising, (o, e) =>
```
krwq: I'm not sure if you are testing RegisterCallbackForPinValueChangedEvent or WaitForEvent here - you should probably make 2 separate tests out of that

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -153,7 +150,81 @@ void callback(object sender, PinValueChangedEventArgs pinValueChangedEventArgs)
...
+                    controller.Write(OutputPin, PinValue.High);
+                    Thread.Sleep(10);
+                    controller.Write(OutputPin, PinValue.Low);
+                });
+
+                WaitForEventResult result = controller.WaitForEvent(LedPin, PinEventTypes.Rising, tokenSource.Token);
+                Assert.True(cancelled);
+                Assert.True(result.TimedOut);
+            }
+        }
```
krwq: other scenarios which might be worth testing:
- register event falling/rising and trigger opposite event and expect the callback not to be called
- register few events, unregister them all and expect not to be called
- register, unregister, register again - expect only 1 event
- WaitForEvent falling
- cancellation token scenarios

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: Well this is testing cancellation of WaitForEvent, so that it cancelled when cancellation token cancelled. Also as WaitForEvent is blocking event we still need to have some thread which could give the signal for the pin it is listening, but yeah we need separate test for that

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: just for saving energy 😛

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: Yeah, there could be many scenarios, tried to test at least these to work, will add more

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: removed

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: added more scenarios

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -150,6 +150,29 @@ void callback(object sender, PinValueChangedEventArgs pinValueChangedEventArgs)
...
+            ManualResetEvent mre = new ManualResetEvent(false);
+            bool wasCalled = false;
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+            {
+                controller.OpenPin(InputPin, PinMode.Input);
+                controller.OpenPin(OutputPin, PinMode.Output);
+                controller.Write(OutputPin, PinValue.Low);
+                controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Falling, callback);
+                controller.Write(OutputPin, PinValue.High);
+                mre.WaitOne(TimeSpan.FromSeconds(5));
```
krwq: 2 things:
- WaitOne returns bool so you don't need additional bool flag
- Test seem to always last 5 seconds - this seems quite too long IMO - 100ms seems reasonably enough

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -195,35 +218,121 @@ void callback(object sender, PinValueChangedEventArgs e)
...
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+            {
+                CancellationTokenSource tokenSource = new CancellationTokenSource();
+                controller.OpenPin(InputPin, PinMode.Input);
+                controller.OpenPin(OutputPin, PinMode.Output);
+                controller.Write(OutputPin, PinValue.Low);
+
                 Task.Run(() =>
-                {
+                {
```
krwq: nit: random extra space

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -130,30 +131,203 @@ public void OpenPinDefaultsModeToInput()
...
-            ManualResetEvent mre = new ManualResetEvent(false);
             bool wasCalled = false;
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 controller.OpenPin(InputPin, PinMode.Input);
                 controller.OpenPin(OutputPin, PinMode.Output);
                 controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Rising, callback);
                 controller.Write(OutputPin, PinValue.High);
-                mre.WaitOne(TimeSpan.FromSeconds(5));
+                Thread.Sleep(10);
```
joshfree: how did we pick 10?  I'm worried about having a flaky/unreliable test here if the value is too small...

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -130,30 +131,203 @@ public void OpenPinDefaultsModeToInput()
...
+        {
+            bool wasCalled = false;
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+            {
+                controller.OpenPin(InputPin, PinMode.Input);
+                controller.OpenPin(OutputPin, PinMode.Output);
+                controller.Write(OutputPin, PinValue.Low);
+                controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Falling, callback);
+                controller.Write(OutputPin, PinValue.High);
+                Thread.Sleep(10);
```
joshfree: same here.  Should these values be hoisted into a static test variable that can be changed in one spot?

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -130,30 +131,203 @@ public void OpenPinDefaultsModeToInput()
...
+                controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Rising, (o, e) => {
+                    risingEventOccuredCount++;
+                });                        
+                controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Rising, callback);
+                controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Rising, (o, e) => {
+                    risingEventOccuredCount++;
+                    if (fallingEventOccuredCount == 4)
+                    {
+                        controller.UnregisterCallbackForPinValueChangedEvent(InputPin, callback);
+                    }
```
joshfree: should there be an else-case here that causes the test to fail?

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -130,30 +131,203 @@ public void OpenPinDefaultsModeToInput()
...
+                });
+                controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Falling, (o, e) => {
+                    fallingEventOccuredCount++;
+                });
+
+                for (int i = 0; i < 10; i++)
+                {
+                    controller.Write(OutputPin, PinValue.High);
+                    Thread.Sleep(10);
```
joshfree: same here with these sleep values.

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -67,7 +67,7 @@ internal partial class Interop
     /// Read the GPIO line direction setting.
     /// </summary>
     /// <param name="line">GPIO line handle</param>
-    /// <returns>GPIOD_DIRECTION_INPUT or GPIOD_DIRECTION_OUTPUT.</returns>
+    /// <returns>1 if INPUT otherwise 2 (OUTPUT).</returns>
```
joshfree: This feels like a wrong change; why get rid of the enum/consts?

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -106,10 +106,43 @@ internal partial class Interop
     [DllImport(LibgpiodLibrary, SetLastError = true)]
     internal static extern int GetGpiodLineValue(SafeLineHandle line);
 
+    /// <summary>
+    /// Check if line is no used (not set as Input or Output, not listening events).
+    /// </summary>
+    /// <param name="line">GPIO line handle</param>
+    /// <returns>0 or 1 if the operation succeeds. On error this routine returns -1 and sets the last error number.</returns>
```
joshfree: this returns comment is either wrong or the return type of the function is wrong (bool)

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -122,7 +181,46 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
...
+            {
+                throw ExceptionHelper.GetArgumentException(ExceptionResource.InvalidEventType);
+            }
+        }
+
+        private void WaitForEventResult(CancellationToken sourceToken, CancellationToken parentToken, ref bool eventOccured)
+        {
+            while (!(sourceToken.IsCancellationRequested || parentToken.IsCancellationRequested || eventOccured))
+            {
+                Thread.Sleep(1);
```
joshfree: this feels like it's going to burn the CPU with such a low value.

/cc @stephentoub

stephentoub: Do we need to spin?  There's nothing we can block waiting for, a callback we can receive, etc.?

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: Not really sure how to choose this value, by my understanding it is really hardware dependent.
So I first used 100 millisec, just thought it will be enough, Krzysztof suggested to lower it, so changed to 10, worked for the Pi am using ...

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: sure that is better, will use a static variable

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: No, that is not test failing scenario, its just random value for removing one of the event listeners in the middle of execution

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
buyaa-n: Will use same statice variable, suggestion for appropriate value is appreciated

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
buyaa-n: That enum is define and used in C library, doc copied from C library, forgot to update these values

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
buyaa-n: Good catch, copy paste issue 😛

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: > this feels like it's going to burn the CPU with such a low value.

Put a value used in lbgpiod samples (same as we use for hardware level timeout https://github.com/dotnet/iot/pull/236/files#diff-dc75afeeda145b47dfc2bd34d7f04d21R171)

> Do we need to spin? There's nothing we can block waiting for, a callback we can receive, etc.?

Reused logic used for AddCallbackForPinValueChangedEvent(...). It can be implemented separately and listen for event directly from hardare. But in that case the pin could only be used for that purpose. In other words if the pin have been used for listening other events this will just add another callback.

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
krwq: I think for direct GPIO writes 1us is what we theoretically should aim for but since we are not running on real-time OS 10ms sounds reasonable. For events I don't have strong opinion but I think at the current state I wouldn't expect anything less than 12ms (approximate Thread.Sleep resolution) to work reliably

buyaa-n: Now using one variable for all sleep duration, will see how it works on repo test server and update accordingly

PR #240, opened 2019-02-16
Convert spi/i2c to use ReadOnlySpan when Writing

krwq:
Fixes: https://github.com/dotnet/iot/issues/87

cc: @joperezr @JeremyKuhne 

<s>I could swear I've seen the issue about it at one point but could not find it :smile: </s>


joperezr:
I'm not super familiar with Span, is this a breaking change? or can bindings that are already sending Spans stay the way they are?

krwq:
@joperezr only binary breaking breaking because we change public surface. If someone overrides SpiDevice/I2cDevice their code will break since the signature doesn't match (I have only seen that in mocks in our tests). Otherwise source compatible.

(`Span<byte>` implicitly converts to `ReadOnlySpan<byte>`)

Issue #170, opened 2019-01-30
Add bindings for E466616 (NeoPixel Stick - 8x 5050 RGB LED with integrated driver)


krwq:

maloo:
Hi  @krwq, just found this repo. I was planning to try this out since I'm just using the Neo 8x panel for an embedded system interface. I have to do this within the next week anyway, so if you have any guidelines, APIs or general ideas it would be helpful. I'm planning to use the SPI way of driving them, because I found that to be the only stable (timing) way in linux without writing a kernel driver.
Been using .Net since day one and have a solid background in embedded SW and drivers working for a mobile phone platform company. Let me know if there is anything in the area of drivers/graphics/displays you need help with. An ex coworker is the maintainer of linux gpio so can probably help with that too.
Please let me know how to proceed, first time grabbing ...

joperezr:
Hey @maloo, so glad you are joining the community. You can find some docs on how to add a new binding here: https://github.com/dotnet/iot/tree/master/src/devices#contributing-a-binding

I do recommend using [our template](https://github.com/dotnet/iot/blob/master/tools/templates/DeviceBindingTemplate/README.md) when creating the device binding so that you can start directly into coding. I would also recommend to check out [Scott Hanselman's blog post](https://www.hanselman.com/blog/RemoteDebuggingWithVSCodeOnWindowsToARaspberryPiUsingNETCoreOnARM.aspx) for your innerloop testing so that you can test things faster while developing. Other  than that we have a few examples for device bindings in the repo already so feel free to take a look in order to give you a sense of how yours should look like 😄 

Let us know if you have any issues/questions.

krwq:
@maloo awesome, thanks for interest in this project :-) other option thank what @joperezr mention is to simply copy the existing project and modify appropriately - please let us know if you find instructions for setup not clear or any other issues you might hit - we are trying to incrementally improve it (so PRs clarifying instructions or issues with a first time experience problems are also welcome).

maloo:
First thing I'm wondering about is the classic Copyright/License. Do you have any guidelines for attribution, copyrights etc when porting (as in changing from C++ to C#). I'm used to do stuff from scratch. Couldn't find any existing library ports in the repo at first glance.
Second question is if maybe a library like ADA Gfx should have its own dll/proj.
But first step I probably will just add some code to existing bindings proj.
Also I couldn't find any sln for bindings proj. Are you all using VS Code? I'm used to the VS sln way of working. Like the Gpio proj.

[...]

Issue #189, opened 2019-02-01
Need OLED Graphics API


shaggygi:
We recently had a new binding PR (#188) for the SSD1306 OLED display controller.  However, it is somewhat limited as it doesn't offer a good solution for text/images.  We need a graphics library that can produce the respective bytes and possibly work with various other displays.

Also referencing #178 as it is somewhat related.

The [Adafruit GFX graphics library](https://learn.adafruit.com/adafruit-gfx-graphics-library/overview?gclid=EAIaIQobChMI4uOo4vmV4AIVGYzICh00OwkhEAAYASAAEgJVY_D_BwE) might be a good place to start to understand what is needed.

maloo:
Hi,
I can port the ADA Gfx library if you want. Or you want a .Net Core designed one? Do you want just bindings or a pure C# library?

joperezr:
What we are looking for here would be trying to port the [GFX library](https://github.com/adafruit/Adafruit-GFX-Library) so that we can use it with other bindings like the one that @shaggygi pointed out.

maloo:
I could take a look at it. Maybe just use a Span<byte> as target for the GFX lib, then it could work with fbdev et al too. Since the color space on these displays is quite limited it might be enough to start with RGB565 support which most I2C/SPI displays support.
Do you have any guidelines for contributions/PRs? Coding style, unit test requirements, process etc

joperezr:
> Do you have any guidelines for contributions/PRs? Coding style, unit test requirements, process etc

I replied to this question [on this other issue](https://github.com/dotnet/iot/issues/170#issuecomment-459840491). Regarding Coding style, we have an .editorConfig that will cause VS to have you follow some guidelines and we follow the same guidelines that [corefx enforces](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md)

BrianA-DK:
@maloo 
You might take a peek at https://github.com/Bodmer/TFT_eSPI and https://github.com/Bodmer/TFT_eFEX
It has some extra TFT functions that is quite handy. Like support for sprites, antialiased fonts, BMP & JPEG load etc.

/Brian

maloo:
Current plan is to write a bitmap/image class like RenderTargetBitmap with some basic draw ops like Adafruit Gfx lib. Then maybe add some bitblt ops.
As for images I plan a command line tool to convert bmp/png/jpg etc to a .cs file in native format.
[...]

PR #241, opened 2019-02-17
Add WS2812B binding

maloo:
fixes: #170 
cc: @krwq @joperezr


joperezr:
> using System;

Please move these two files under src/devices/Common/BitmapImage.cs and also change the namespace so that it is like: `Iot.Device.Common`

---
Refers to: src/Iot.Device.Bindings/Utils/BitmapImage.cs:1 in bbf7c15. [](commit_id = bbf7c1594a49a88c003bed385df4c60cd7ad99b4, deletion_comment = False)

joperezr:
> using System;

Also, can you please add the liscensing headers into these two files as well?

---
Refers to: src/Iot.Device.Bindings/Utils/BitmapImage.cs:1 in bbf7c15. [](commit_id = bbf7c1594a49a88c003bed385df4c60cd7ad99b4, deletion_comment = False)

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -18,6 +18,11 @@
     <Compile Include="$(DeviceRoot)**/*.cs" Exclude="@(_ExcludeFromCompile)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="Utils\BitmapImage.cs" />
```
joperezr: As per my other comment, these two files shouldn't leave here so please remove these edits. Once you move the files to the place I suggested, then they will be automatically compiled by this project.

src/devices/Ws2812B/BitmapImageNeo3.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Iot.Device.Bindings.Utils;
+
+namespace Iot.Device.Bindings.WS2812B
```
joperezr: Namespace should be just `Iot.Device.Ws2812b`

joperezr:
> <Project Sdk="Microsoft.NET.Sdk">

These samples project should instead live under the binding you added in a samples folder. You can check all of our existing bindings for examples, but basically what we require for a binding is: a project under `src/devices/<yourBinding>/` folder, and a Readme.md with information on the binding(again, you can check all of our existing bindings' Readme files for the content that should go in it). Then we also ask for a samples folder containing a sample project with the usage of your binding, along with another readme containing both a fritzing diagram and an explanation of what the samples would do and how to wire it. Seems like the samples contents you have here is right, so you just need to move it under the binding.

---
Refers to: samples/neo-pixels-x8/neo-pixels-x8.csproj:1 in bbf7c15. [](commit_id = bbf7c1594a49a88c003bed385df4c60cd7ad99b4, deletion_comment = False)

samples/neo-pixels-x8/neo-pixels-x8.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <RootNamespace>neo_pixels_x8</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Iot.Device.Bindings\Iot.Device.Bindings.csproj" />
```
joperezr: Don't reference the full bindings project directly, instead reference your bidning project only.

joperezr:
Hey @maloo, thanks for your contribution! I left some initial comments on the PR, in order to produce the right layout/structure we are looking for in a binding, would you mind either copying the structure of an existing binding or using the steps outlined [here](https://github.com/dotnet/iot/tree/master/tools/templates/DeviceBindingTemplate) in order to create your binding project using our CLI template?

samples/neo-pixels-x8/Program.cs
```diff
@@ -0,0 +1,41 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Iot.Device.Bindings.Utils;
+using Iot.Device.Bindings.WS2812B;
+
+namespace led_blink
```
krwq: namespace should match the device

samples/neo-pixels-x8/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    class Program
+    {
+        static void Main()
+        {
+            var spi = new System.Device.Spi.Drivers.UnixSpiDevice(new System.Device.Spi.SpiConnectionSettings(0, 0));
```
krwq: `using System.Device.Spi`

samples/neo-pixels-x8/Program.cs
```diff
@@ -0,0 +1,41 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
krwq: we usually put device samples under `src/devices/<yourdevice>/samples` - see other devices for reference

samples/neo-pixels-x8/README.md
```diff
@@ -0,0 +1,31 @@
+# Drive Neo pixel strip (x8) from Raspberry Pi
+
+This [sample](Program.cs) demonstrates how to use the [Bitmap library](../../src/Iot.Device.Bindings/Utils/BitmapImage.cs) and the [Neo pixel binding](../../src/devices/Ws2812B/Ws2812B.cs) to drive an 8 Neo Pixel stick from a Raspberry Pi.
+
+## Run the sample
+
+This sample can be built and run with .NET Core 3.0. Use the following commands from the root of the repo:
```
krwq: This should be elsewhere in the docs since it is not related to this device (fine to add a link or contribute to the generic docs: https://github.com/dotnet/iot/tree/master/Documentation)

src/Iot.Device.Bindings/Utils/BitmapImage.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
```
krwq: 1. copyright header is missing
2. this should be internal in your device, for generic purpose APIs please create an issue and after we get an agreement from other people on APIs then we can move it to shared directory or common folder in the devices

src/Iot.Device.Bindings/Utils/Color.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System.Runtime.InteropServices;
```
krwq: System.Drawing already defines Color class

src/devices/Ws2812B/BitmapImageNeo3.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public BitmapImageNeo3(int width, int height)
+            : base(new byte[width * height * BytesPerPixel + ResetDelayInBytes], width, height, width * BytesPerPixel)
+        {
+        }
+
+        public void SetPixel(int x, int y, uint color) => SetPixel(x, y, new Color(color));
+        public void SetPixel(int x, int y, byte level) => SetPixel(x, y, new Color(level));
+
+        public override void SetPixel(int x, int y, Color c)
```
krwq: this coincidentally exactly matches the signature I'm using in the LED matrix for SenseHat (except the Color where I use the System.Drawing.Color) - we should create an issue to figure out the common abstraction

src/devices/Ws2812B/BitmapImageNeo3.cs
```diff
@@ -0,0 +1,57 @@
...
+
+using Iot.Device.Bindings.Utils;
+
+namespace Iot.Device.Bindings.WS2812B
+{
+    /// <summary>
+    /// Special 24bit RGB format for Neo pixel LEDs where each bit is converted to 3 bits.
+    /// A one is converted to 110, a zero is converted to 100.
+    /// </summary>
+    public class BitmapImageNeo3 : BitmapImage
```
krwq: Name should match the device name

src/devices/Ws2812B/BitmapImageNeo3.cs
```diff
@@ -0,0 +1,57 @@
...
+            Data[offset++] = _lookup[c.R * BytesPerComponent + 0];
+            Data[offset++] = _lookup[c.R * BytesPerComponent + 1];
+            Data[offset++] = _lookup[c.R * BytesPerComponent + 2];
+            Data[offset++] = _lookup[c.B * BytesPerComponent + 0];
+            Data[offset++] = _lookup[c.B * BytesPerComponent + 1];
+            Data[offset++] = _lookup[c.B * BytesPerComponent + 2];
+        }
+
+        private static readonly byte[] _lookup = new byte[256 * BytesPerComponent];
+        static BitmapImageNeo3()
```
krwq: nit: our coding guidelines require explicit `private`

src/devices/Ws2812B/BitmapImageNeo3.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    /// <summary>
+    /// Special 24bit RGB format for Neo pixel LEDs where each bit is converted to 3 bits.
+    /// A one is converted to 110, a zero is converted to 100.
+    /// </summary>
+    public class BitmapImageNeo3 : BitmapImage
+    {
+        const int BytesPerComponent = 3;
+        const int BytesPerPixel = BytesPerComponent * 3;
+        const int ResetDelayInBytes = 30; // 100us @ 2.4Mbps
```
krwq: please use time units instead (TimeSpan or miliseconds) unless datasheet defines it like that then please use actual SPI speed.

krwq: btw this feels like should be in the device itself and not in the bitmap

krwq:
Hey @maloo thanks for contributing, couple of comments (mostly naming and layout) other than that looks good!

src/Iot.Device.Bindings/Utils/BitmapImage.cs
maloo: I moved it to Common as requested by @joperezr. You can create an issue. But this is actually part of my other work item, porting Adafruit GFX lib. So maybe we could use that issue instead. For now I drop it in Common, once GFX lib is merged it will have found its final resting place.

src/Iot.Device.Bindings/Utils/Color.cs
maloo: I know :$, I just don't like it ;). Anyway, I ported lib and sample to use System.Drawing. I loose a bunch of helpers. But maybe we can get those in System.Drawing later.

src/devices/Ws2812B/BitmapImageNeo3.cs
maloo: See above, add your requirements to issue #189.

src/devices/Ws2812B/BitmapImageNeo3.cs
maloo: I don't think that is allowed in C#

src/devices/Ws2812B/BitmapImageNeo3.cs
maloo: The bitmap format is local to the binding, so I consider it already part of the device. Would it be better to make it a private class? I feel it is a bit messy to add this constant to the constructor of the bitmap.
As for the speed, I'm not sure what you mean. Freq should always be 2.4Mbps for these LEDs. Do you mean I should calculate this constant from say `internal const int Ws2812b.SpiSpeed = 2_400_000`? What if someone add a BitmapImageNeo1 (1Bpp, requires extra HW)? Then I think it is better to move the Spi freq to BitmapImageNeo3, since that is what is dictating the freq.

src/devices/Ws2812B/BitmapImageNeo3.cs
krwq: @maloo I meant this: `private static BitmapImageNeo3()` - that is allowed

src/Iot.Device.Bindings/Utils/Color.cs
krwq: @maloo not a fan of Color and and Vector/Vector2/Vector3 as well - see my sample in SenseHat PR - the code could be at least 3 times shorter if we could do something similar to GLSL. You should create an issue on https://github.com/dotnet/corefx

src/devices/Ws2812B/BitmapImageNeo3.cs
krwq: Let's do it private for now and leave the delay as is. Once we figure out the common interface I suspect we will need to refactor a lot of similar classes all over the place

src/devices/Ws2812B/BitmapImageNeo3.cs
```diff
@@ -2,27 +2,31 @@
...
+        private const int ResetDelayInBytes = 30;
 
         public BitmapImageNeo3(int width, int height)
             : base(new byte[width * height * BytesPerPixel + ResetDelayInBytes], width, height, width * BytesPerPixel)
         {
         }
 
-        public void SetPixel(int x, int y, uint color) => SetPixel(x, y, new Color(color));
-        public void SetPixel(int x, int y, byte level) => SetPixel(x, y, new Color(level));
+        public void SetPixel(int x, int y, uint color) => SetPixel(x, y, Color.FromArgb((int)color));
```
krwq: I'd stick with just Color overload for now, those 2 feel a bit confusing

src/devices/Ws2812B/Ws2812b.csproj
```diff
@@ -0,0 +1,17 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <!--Disabling default items so samples source won't get build by the main library-->
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <Compile Include="../Common/**/*.cs" />
```
krwq: pulling from Common should be more explicit, so please remove globs unless there is directory related directly with this device

src/devices/Ws2812B/BitmapImageNeo3.cs
maloo: Tell that to the compiler ;)
[The almighty](https://stackoverflow.com/questions/4119365/static-constructor-in-c-sharp)

Issue #196, opened 2019-02-01
Sense HAT


RPagels:
Is there support in Dot Net Core C# for the Raspberry Pi Sense HAT?

There is a Python implementation of it here: https://github.com/RPi-Distro/python-sense-hat.  It would be awesome for it to be ported to C#.

Thank you,

Randy

richlander:
Not yet. We'd love one. Anyone else want support for this device? Please vote on the issue.

emmellsoft:
I have a working C# implementation (currently UWP) that I am porting to .NET Core.
https://github.com/emmellsoft/RPi.SenseHat
Almost completed.

joshfree:
Thanks @emmellsoft for looking into porting the SenseHat binding!

/cc @krwq

krwq:
@emmellsoft really cool! I've also added some basic implementation in the PR over the weeken: https://github.com/dotnet/iot/pull/242 - I've seen you have a lot of stuff we currently don't have - any chance we could merge this together? (calibration, drawing utilities, more control over the sensors etc)

emmellsoft:
Yes of course! Let me just wrap things up a bit!

krwq:
Thank you! @emmellsoft let's do it bit by bit - so let's start with my merging my basic stuff, then perhaps calibration and more config options from your library and perhaps changing some conventions

For the drawing utilities we will have a separate conversation on some common interface. @gloveboxes and couple of people here have couple ideas about it and there were couple of proposition for libraries - I don't believe there is a reason to limit to just one library - we can have one for text related stuff, one for lines and simple geometric stuff and others and figure out similarities and right abstraction along the way.

I'd start with the calibration and more configuration because this is lacking completely and also will give you a hang of conventions in our repo - we will need to figure out if we want to put it under devices or tools - depending how you have created your APIs. Let's make sure we provide as easy experience as possible but also have some more powerful slightly more complex options.

emmellsoft:
I just pushed my ported code. Although I had totally missed the `System.Devices.Gpio` project, so I based the I2C communication on the `Unosquare` implementation. However, I have limited the usage of I2C, so it should be easy to replace.

I will look into the iot project here -- I would love to contribute! :-)

krwq:
Hey @emmellsoft my stuff is merged - feel free to add PRs/issues with anything you find useful and we are missing. I'm really curious to see how far we can stretch sense hat 😄

PR #242, opened 2019-02-17
SenseHAT [LSM9DS1, HTS221, LPS25H included] - LED matrix, joystick, accelerometer, gyroscope, magnetometer, thermometer, humidity, pressure

krwq:
Fixes: https://github.com/dotnet/iot/issues/196
Fixes: https://github.com/dotnet/iot/issues/253
Fixes: https://github.com/dotnet/iot/issues/245
Fixes: https://github.com/dotnet/iot/issues/252

cc: @joperezr @JeremyKuhne


joperezr:
Lol you beat me to it!! I was working on this today as well :smile: my implementation though was using I2c to talk to the device, but looks like instead you used the same implementation that sense hat python library uses. Do you plan to add the rest of the sensors to this same binding? Or will that come in a future PR instead?

krwq:
@joperezr not in this PR :smile: We can split remaining sensors if you want :smile: You can also add the I2c implementation - I didn't know it was possible (do you connect just ground and I2C pins for that?)

I couldn't find much documentation except for python source code so went with slightly simpler version of python code - it should be easy to build on top of that with current APIs

joperezr:
All of the sensors in the Hat are I2c, so when you connect it they will all show up in the I2c bus. If you run i2cdetect 1 on your pi you’ll see how all of these are connected in there, the LED matrix in particular is 0x46, and is very easy to control using I2c, literally you just have to send it the bitmap of the leds and it will light them up. I believe that internally what your implementation will do is the same. The nice thing of using I2c directly, is that you don’t require any library installed on the pi, the way your implementation is, might require sense-hat library installed in order to find the device

krwq:
@joperezr isn't it faster to use the sysfs in this case since it has designated driver? We can have both and compare perf later and then if sysfs is slower then we can remove it. I didn't know about i2cdetect. Do you have some datasheet explaining how talk to devices or are they all some standard devices connected to i2c bus on a single board?

joperezr:
Here is the info on the different sensors on the hat https://pinout.xyz/pinout/sense_hat

As you see it has the address of each sensor and the chip name, which should help finding each datasheet.

krwq:
ok, added I2C implementation as well. Initially I was planning to remove one of them after testing but there are some differences in how colors look like:
- I2C - colors are more pale but does not require driver
- driver/sysfs - colors look more vivid but require driver

sysfs implementation seems slightly faster but not 100% sure if this isn't because of FS as internally I suspect it uses I2C either way and since the matrix is just 8x8 it probably doesn't matter either way

krwq:
Added accelerometer + gyroscope from LSM9DS1, magnetometer TODO - this is generally 2 devices in 1 since they work on 2 different I2C buses

krwq:
@ZhangGaoxing please take a look at least on LSM9DS1 - this is very similar to stuff from the different device you were adding

joperezr:
Awesome, I’ll take a look at this tomorrow. FYI @richlander

krwq:
To summarize this is still missing:
- LPS25H - Pressure/Temperature
- HTS221 - Humidity/Temperature
- LSM9DS1 - Magnetometer is missing (gyro and accelerometer in PR)
- some simple SenseHat abstraction which contains all of them
- some better samples (gyro/joystick+LCD should make a fun sample)

src/devices/Lsm9Ds1/Lsm9Ds1.AccelerometerAndGyroscope.cs
```diff
@@ -0,0 +1,138 @@
...
+
+namespace Iot.Device.Lsm9Ds1
+{
+    public class Lsm9Ds1AccelerometerAndGyroscope : IDisposable
+    {
+        private const byte ReadMask = 0x80;
+        private I2cDevice _i2c;
+        private AccelerationScale _accelerometerScale;
+        private AngularRateScale _angularRateScale;
+
```
ZhangGaoxing: I think devices like SPI and I2C should provide information such as I2cAddress, SpiClockFrequency in class. If I am a user, I would probably not look at examples or datasheet. For example, if the I2C device has multiple addresses, it can be provided by enum type.😁

ZhangGaoxing: code like this:
```C#
// SPI
public const int SpiClockFrequency = 5000000;
public const SpiMode SpiMode = System.Device.Spi.SpiMode.Mode3;

// I2C
public const byte I2cAddress = 0x68;
public enum I2cAddress
{

}
```

krwq: I couldn't find any I2C address in the datasheet except for the SenseHat where this sensor is - for sense HAT I've added that in the SenseHat version of the device

src/devices/Lsm9Ds1/AccelerationScale.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Lsm9Ds1
+{
+    public enum AccelerationScale : byte
```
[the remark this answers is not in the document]
krwq: isn't range related to distance? scale seems like generic term and also that's what datasheet calls it.

krwq:
@joperezr ready for review, I currently don't have any ideas (or energy) for a cooler sample - I was thinking snake perhaps but we could add that in a separate PR since this one grew quite a bit

src/devices/Hts221/Hts221.cs
```diff
@@ -0,0 +1,148 @@
...
+        private (short t0, short t1) GetTemperatureCalibrationPointsRaw()
+        {
+            Span<byte> t0t1 = stackalloc byte[4];
+            Read(Register.Temperature0Raw, t0t1);
+            short t0 = BinaryPrimitives.ReadInt16LittleEndian(t0t1.Slice(0, 2));
+            short t1 = BinaryPrimitives.ReadInt16LittleEndian(t0t1.Slice(2, 2));
+            return (t0, t1);
+        }
+
+        (byte h0, byte h1) GetHumidityCalibrationPointsRH()
```
buyaa-n: NIT: don't need to specify access modifier?

src/devices/Hts221/README.md
```diff
@@ -0,0 +1,13 @@
...
+- Air conditioning, heating and ventilation
+- Air humidifiers
+- Refrigerators
+- Wearable devices
+- Smart home automation
+- Industrial automation
+- Respiratory equipment
+- Asset and goods tracking
+
+See [samples](samples/README.md) for information about usage.
```
buyaa-n: Can we have link to data sheets?

src/devices/SenseHat/SenseHatJoystick.cs
```diff
@@ -0,0 +1,72 @@
...
+            _i2c.WriteByte(StateRegister);
+            return (JoystickState)_i2c.ReadByte();
+        }
+
+        [Flags]
+        private enum JoystickState : byte
+        {
+            Down = 1,
+            Right = 1 << 1,
+            Up = 1 << 2,
```
buyaa-n: Isn't it better to set the numbers directly (2, 4 ...) instead of shifting 1  inti left each time?

src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
```diff
@@ -0,0 +1,121 @@
...
+using System;
+using System.IO;
+using System.Device.Gpio;
+using System.Diagnostics;
+using System.Drawing;
+using System.Runtime.CompilerServices; 
+
+namespace Iot.Device.SenseHat
+{
+    public class SenseHatLedMatrixSysFs : SenseHatLedMatrix
```
buyaa-n: Is this specific to SysFs driver only or generic UnixDriver

src/devices/SenseHat/SenseHatJoystick.cs
krwq: this will be compile constant anyway and this way I can tell which bit it is quickly

src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
krwq: not sure I understand - SenseHat driver exposes its kernel objects using sysfs. This implementation requires the SenseHat driver - I2C does not. This implementation will produce better colors than the I2C

src/devices/Hts221/Hts221.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// <summary>
+        /// Hts221 - Temperature and humidity sensor
+        /// </summary>
+        public Hts221(I2cDevice i2cDevice)
+        {
+            _i2c = i2cDevice;
+
+            // Highest resolution for both temperature and humidity sensor:
+            // 0.007 Celsius and 0.03 percentage of relative humidity respectively
+            byte resolution = 0b00111111;
```
ahsonkhan: nit: use digit separators

Here and elsewhere

src/devices/Hts221/Hts221.cs
```diff
@@ -0,0 +1,148 @@
...
+    {
+        private const byte ReadMask = 0x80;
+        private I2cDevice _i2c;
+
+        /// <summary>
+        /// Hts221 - Temperature and humidity sensor
+        /// </summary>
+        public Hts221(I2cDevice i2cDevice)
+        {
+            _i2c = i2cDevice;
```
ahsonkhan: What if `i2cDevice` is `null`?

src/devices/Lps25h/Lps25h.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        private static int ReadInt24LittleEndian(ReadOnlySpan<byte> buff)
+        {
+            Debug.Assert(buff.Length == 3);
+
+            Span<byte> b = stackalloc byte[4]
+            {
+                buff[0],
+                buff[1],
+                buff[2],
```
ahsonkhan: Cache `buff[2]` in a local, up front?
It will probably remove the bounds checks on `buff[0]`, `buff[1]` later on as well.

src/devices/Lsm9Ds1/Lsm9Ds1.AccelerometerAndGyroscope.cs
```diff
@@ -0,0 +1,138 @@
...
+        public Lsm9Ds1AccelerometerAndGyroscope(
+            I2cDevice i2cDevice,
+            AccelerationScale accelerationScale = AccelerationScale.Scale02G,
+            AngularRateScale angularRateScale = AngularRateScale.Scale0245Dps)
+        {
+            _i2c = i2cDevice;
+            _accelerometerScale = accelerationScale;
+            _angularRateScale = angularRateScale;
+
+            Span<byte> buff = stackalloc byte[2];
```
ahsonkhan: This isn't used anywhere.

src/devices/Lsm9Ds1/Lsm9Ds1.Magnetometer.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            byte disableI2c = 0; // enable I2C
+            byte lowPowerMode = 0; // higher power mode
+            byte spiAllowReads = 0; // only writes allowed
+            byte operatingMode = 0b00; // continous mode
+
+            byte operativeModeForZAxis = 0b11; // ultra-high performance mode
+            byte bigEndianEnabled = 0; // little endian
+
+            Span<byte> buff = stackalloc byte[]
```
ahsonkhan: specify the size?

src/devices/Lsm9Ds1/Lsm9Ds1.Magnetometer.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            short x = BinaryPrimitives.ReadInt16LittleEndian(vec.Slice(0, 2));
+            short y = BinaryPrimitives.ReadInt16LittleEndian(vec.Slice(2, 2));
+            short z = BinaryPrimitives.ReadInt16LittleEndian(vec.Slice(4, 2));
+            return new Vector3(x, y, z);
+        }
+
+        private void Write(RegisterM register, ReadOnlySpan<byte> data)
+        {
+            Debug.Assert(data.Length <= 10);
```
ahsonkhan: Where does this 10 come from?

src/devices/Lsm9Ds1/samples/Lsm9Ds1.Magnetometer.Sample.cs
```diff
@@ -0,0 +1,91 @@
...
+                    min = Vector3.Min(min, sample);
+                    max = Vector3.Max(max, sample);
+                    Thread.Sleep(50);
+                }
+
+                Console.WriteLine("Stop moving for some time...");
+                Thread.Sleep(3000);
+
+                const int intervals = 32;
+                bool[,] data = new bool[32,32];
```
ahsonkhan: nit: `new bool[intervals, intervals]`

src/devices/Lsm9Ds1/samples/Lsm9Ds1.Sample.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Device.I2c;
+using System.Device.I2c.Drivers;
+
+namespace Iot.Device.Lsm9Ds1.Samples
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            // AccelerometerAndGyroscope.Run();
```
ahsonkhan: Remove commented out code or add a comment explaining why one is commented out.

src/devices/SenseHat/SenseHatLedMatrix.cs
```diff
@@ -0,0 +1,48 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static (int x, int y) IndexToPosition(int index)
+        {
+            return (index % NumberOfPixelsPerRow, index / NumberOfPixelsPerRow);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int PositionToIndex(int x, int y)
+        {
+            Debug.Assert(x >= 0 && x < NumberOfPixelsPerRow);
```
ahsonkhan: What's guaranteeing this on a public API?

src/devices/SenseHat/SenseHatLedMatrix.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        public abstract void Write(ReadOnlySpan<Color> colors);
+
+        public abstract void Clear(Color color = default(Color));
+
+        public abstract void SetPixel(int x, int y, Color color);
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static (int x, int y) IndexToPosition(int index)
```
ahsonkhan: Do these statics have to be public?

src/devices/SenseHat/SenseHatLedMatrixI2c.cs
```diff
@@ -0,0 +1,145 @@
...
+            Span<byte> buffer = stackalloc byte[FrameBufferLength + 1];
+
+            // Register address of first pixel
+            buffer[0] = 0;
+
+            Span<byte> frameBuffer = buffer.Slice(1);
+
+            for (int i = 0; i < NumberOfPixels; i++)
+            {
+                (int x, int y) = IndexToPosition(i);
```
ahsonkhan: There are some redundant operations going on between `IndexToPosition` and `PositionToAddress` for each of these iterations.

src/devices/SenseHat/SenseHatLedMatrixI2c.cs
```diff
@@ -0,0 +1,145 @@
...
+                int address = PositionToAddress(x, y);
+                (byte r, byte g, byte b) = DestructColor(colors[i]);
```
ahsonkhan: `Debug.Assert` that `colors.Length <= NumberOfPixels`

src/devices/SenseHat/SenseHatLedMatrixI2c.cs
```diff
@@ -0,0 +1,145 @@
...
+            encoded[1] = r;
+            _i2c.Write(encoded);
+
+            encoded[0] = (byte)(address + GOffset);
+            encoded[1] = g;
+            _i2c.Write(encoded);
+
+            encoded[0] = (byte)(address + BOffset);
+            encoded[1] = b;
+            _i2c.Write(encoded);
```
ahsonkhan: Is it necessary to send 3 separate writes? Would sending a single 6-byte write work?

src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            for (int i = 0; i < NumberOfPixels; i++)
+            {
+                WriteColor(colors[i]);
+            }
+
+            EndWriting();
+        }
+
+        public override void Clear(Color color = default(Color))
```
ahsonkhan: Why does Clear take a Color? Should there be two separate APIs? `Clear()` and `FilI(Color)`?

src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            for (int i = 0; i < NumberOfPixels; i++)
+            {
+                WriteColor(color);
+            }
+
+            EndWriting();
+        }
+
+        public override void SetPixel(int x, int y, Color color)
```
ahsonkhan: Validate that the input is `>= 0`?

src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void WriteColor(Color color)
+        {
+            // Writes color in RGB565 format
+            byte r = (byte)(color.R >> 3);
+            byte g = (byte)(color.G >> 2);
+            byte b = (byte)(color.B >> 3);
+            ushort col = (ushort)((r << 11) | (g << 5) | b);
+            Span<byte> encoded = stackalloc byte[2] { (byte)(col & 0xff), (byte)(col >> 8) };
```
ahsonkhan: nit: inconsistent formatting of stackalloc bytes

src/devices/SenseHat/samples/LedMatrix.Sample.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        public static void Run()
+        {
+            // another implementation which can be used is: SenseHatLedMatrixSysFs
+            // I2C implementation does not require installing anything
+            // SysFs implementation is faster and has arguably better colors
+            using (var m = new SenseHatLedMatrixI2c())
+            {
+                WriteDemo(m);
+                // SetPixelDemo(m);
```
ahsonkhan: commented out code

src/devices/SenseHat/samples/LedMatrix.Sample.cs
```diff
@@ -0,0 +1,91 @@
...
+        static Color Pixel(Vector2 uv, float time)
+        {
+            return Color.FromArgb(
+                Col(0.5 + 0.5 * Math.Cos(time + uv.X)),
+                Col(0.5 + 0.5 * Math.Cos(time + uv.Y + 2.0)),
+                Col(0.5 + 0.5 * Math.Cos(time + uv.X + 4.0))
+            );
+        }
+
+        static byte Col(double x)
```
ahsonkhan: add missing access modifiers (here/elsewhere)

src/devices/SenseHat/samples/Magnetometer.Sample.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Iot.Device.SenseHat.Samples
+{
+    class Magnetometer
+    {
+        public const int I2cAddress = 0x1C;
+
+        public static void Run()
+        {
+            using (var m = new SenseHatMagnetometer())
```
ahsonkhan: nit: avoid single letter variable names

src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
buyaa-n: Ok, seems sysfs specific

src/devices/SenseHat/SenseHatJoystick.cs
joperezr: In different bindings we are using a different style for these, we do: `0b_0001`, `0b_0010`, etc. I think that is better and addresses @buyaa-n concern.

src/devices/SenseHat/SenseHatLedMatrix.cs
krwq: no, but user of this class will have to do this calculation themselves

src/devices/SenseHat/SenseHatLedMatrixI2c.cs
krwq: there is a `(colors.Length != NumberOfPixels)` in the very begining of the function

ahsonkhan: Sorry, missed that part.

src/devices/SenseHat/SenseHatLedMatrixI2c.cs
krwq: could you clarify? note `address != index` - index points to the buffer, address is in the device memory. There are only 64 pixels here so this doesn't really matter that much anyway

src/devices/SenseHat/SenseHatLedMatrixI2c.cs
krwq: yes, those are 3 separate transactions (I2C numbers the data it sends) - I actually spent quite a bit of time debugging why writing 6 bytes at once doesn't do what I want :smile:

ahsonkhan: > I actually spent quite a bit of time debugging why writing 6 bytes at once doesn't do what I want

Care to share your findings? What does it end up doing instead if you write at once? Do you know why it doesn't work?

src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
krwq: I changed to Fill but with default arg as is - Clear feels redundant

krwq:
Ok, this has been open long enough - merging - I'll address any feedback in separate PR

PR #244, opened 2019-02-18
Dispose() method implemented

garciaolais:
Hi

I noticed that the method is empty, so I add this little fix

### DIFF
--- a/src/devices/Si7021/Si7021.cs
+++ b/src/devices/Si7021/Si7021.cs
@@ -18,10 +18,6 @@ namespace Iot.Device.Si7021
             _i2cDevice = i2cDevice;
         }
 
-        public void Dispose()
-        {
-        }
-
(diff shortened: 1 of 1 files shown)


src/devices/Si7021/Si7021.cs
```diff
@@ -59,5 +55,15 @@ public double ReadHumidity()
             
             return humidity;
         }
+
+        public void Dispose()
+        {
+            if (_i2cDevice != null)
```
johnastaf: _i2cDevice?.Dispose() ? :-)

garciaolais: I like your propose

garciaolais: > _i2cDevice?.Dispose() ? :-)

@johnastaf  done 👍

joperezr:
Thanks @garciaolais !

PR #246, opened 2019-02-18
Adding GrovePi support

Ellerbach:
- Adding supportfor GrovePi from Dexter Industries
- Adding support for all core GrovePi functions
- Adding support for native GrovePi sensors like Ultrasound, Led Bar, DHT
- Adding support for generic Analogic Input, Digital Input and Output, PWM Output
- Adding support for high level sensors like buzzer, led, relay
- Adding documentation
- Adding sample with usage of multiple sensors, all present in the Grove starter kit


src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+    public class GrovePi : IDisposable
+    {        
+        private I2cDevice _i2cDevice;
+        private readonly bool _autoDispose;
+
+        /// <summary>
+        /// The default GrovePi I2C address is 0x04
+        /// Other addresses can be use, see GrovePi documentation
+        /// </summary>
+        public static byte GrovePiSefaultI2cAddress => 0x04;
```
krwq: `I2cAddress` should be good enough

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        /// <summary>
+        /// The maximum ADC, 12 bit so 2013 on GrovePi
```
krwq: confused by the comment, since value is 1023

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// </summary>
+        public int MaxAdc => 1023;
+        /// <summary>
+        /// Contains the GrovePi key information
+        /// </summary>
+        public GrovePiInfo GrovePiInfo { get; internal set; }
```
krwq: `Info` is sufficient since this is on GrovePi class

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+        public GrovePi(I2cDevice i2cDevice, bool autoDispose = true)
+        {
+            _i2cDevice = i2cDevice ?? throw new ArgumentException("I2C device can't be null");
+            _autoDispose = autoDispose;
+            GrovePiInfo = new GrovePiInfo() { SoftwareVersion = GetFirmwareVerion() };
+        }
+
+        public void Dispose()
+        {
+            if(_autoDispose)
```
krwq: space before `(`

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+            {
+                _i2cDevice.Dispose();
```
krwq:
```
_i2cDevice?.Dispose();
_i2cDevice = null;
```

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        /// <summary>
+        /// Get the firmware version
+        /// </summary>
+        /// <returns>GroovePi firmware version</returns>
+        public Version GetFirmwareVerion()
+        {
+            WriteCommand(GrovePiCommands.Version, 0, 0, 0);
+            Thread.Sleep(10);
```
krwq: Why is sleep and 3 zeros needed needed?

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+            var inArray = ReadCommand(GrovePiCommands.Version, 0);
+            return new Version(inArray[1], inArray[2], inArray[2]);
```
krwq: `inArray[1], inArray[2], inArray[3]`?

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// <summary>
+        /// Write a GrovePi command
+        /// </summary>
+        /// <param name="commands">The GrovePi command</param>
+        /// <param name="pin">The pin to write the command</param>
+        /// <param name="param1">First parameter</param>
+        /// <param name="param2">Second parameter</param>
+        public void WriteCommand(GrovePiCommands commands, GrovePort pin, byte param1, byte param2)
+        {
+            byte[] outArray = new byte[4] { (byte)commands, (byte)(pin), param1, param2 };
```
krwq: `Span<byte> outArray = stackalloc byte[4]`

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+            try
+            {
+                _i2cDevice.Write(outArray.AsSpan());
```
krwq: AsSpan is not needed since `byte[]` is also a Span

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+            }
+            catch (Exception ex) when (ex is IOException)
```
krwq: `catch (IOException ex)`

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+                    numberBytesToRead = 4;
+                    break;
+                case GrovePiCommands.DhtTemp:
+                    numberBytesToRead = 9;
+                    break;
+                // No other commands are for read
+                default:
+                    return null;
+            }
+            Span<byte> inArray = new Span<byte>(new byte[numberBytesToRead]);
```
krwq: This should be:

```
byte[] outArray = new byte[numberBytesToRead];
_i2cDevice.Read(outArray);
return outArray;
```

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// <returns>Returns the level either High or Low</returns>
+        public PinLevel DigitalRead(GrovePort pin)
+        {
+            WriteCommand(GrovePiCommands.DigitalRead, pin, 0, 0);
+            Thread.Sleep(10);
+            try
+            {
+                return (PinLevel)_i2cDevice.ReadByte();
+            }
+            catch (Exception ex) when (ex is IOException)
```
krwq: exception directly in catch

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        /// <summary>
+        /// Read a digital pin, equivalent of digitalRead on Arduino
+        /// </summary>
+        /// <param name="pin">The GroovePi pin to read</param>
+        /// <returns>Returns the level either High or Low</returns>
+        public PinLevel DigitalRead(GrovePort pin)
+        {
+            WriteCommand(GrovePiCommands.DigitalRead, pin, 0, 0);
+            Thread.Sleep(10);
```
krwq: why is the sleep needed? if datasheet requires it - should it be part of WriteCommand instead?

src/devices/GrovePi/Models/GrovePiCommands.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Iot.Device.GrovePiDevice.Models
+{
+    /// <summary>
+    /// GroovePi commands to read, write, setup pins and access special sensors
+    /// Note that those commands are supports in most of the recent firmware with version higher than 1.2.1
+    /// </summary>
+    public enum GrovePiCommands
```
krwq: should this be private?

src/devices/GrovePi/Models/GrovePiCommands.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
```
krwq: copyright header

krwq: same comment for other files where it is missing

src/devices/GrovePi/Models/PinMode.cs
```diff
@@ -0,0 +1,20 @@
...
+    public enum PinMode
+    {
+        Input = 0,
+        Output = 1
+    }
+
+    /// <summary>
+    /// Defines the pin level, either high, either low
+    /// </summary>
+    public enum PinLevel
```
krwq: why not use PinValue?

src/devices/GrovePi/Models/PinMode.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Iot.Device.GrovePiDevice.Models
+{
+    /// <summary>
+    /// Defines the pin mode, either input, either output
+    /// </summary>
+    public enum PinMode
```
krwq: why not use `PinMode` from Gpio?

src/devices/GrovePi/Sensors/AnalogSensor.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        /// <param name="grovePi">The GrovePi class</param>
+        /// <param name="port">The grove Port, need to be in the list of SupportedPorts</param>
+        public AnalogSensor(GrovePi grovePi, GrovePort port)
+        {
+            if (!SupportedPorts.Contains(port))
+                throw new ArgumentException($"Error: grove Port not supported");
+            _grovePi = grovePi;
+            Port = port;
+            _grovePi.PinMode(Port, PinMode.Input);
```
krwq: I feel like GrovePi could probably implement IGpioController - might be worth taking a look

src/devices/GrovePi/Sensors/DhtSensor.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                ReadSensor();
+                return _lastTemHum;
+            }
+        }
+
+        /// <summary>
+        /// You need to read the sensorbefore getting the value
+        /// </summary>
+        public void ReadSensor()
```
krwq: I'd go with just `Read`

src/devices/GrovePi/Sensors/DhtSensor.cs
```diff
@@ -0,0 +1,125 @@
...
+        Dht11 = 0,
+        Dht22 = 1,
+        Dht21 = 2,
+        Am2301 = 3
+    }
+
+    /// <summary>
+    /// DhtSensor supports DHT familly sensors
+    /// </summary>
+    public class DhtSensor : ISensor<double[]>
```
krwq: could this perhaps share some code with https://github.com/dotnet/iot/tree/master/src/devices/Dhtxx ?

src/devices/GrovePi/Sensors/GrooveTemperatureSensor.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Returns the temperature formated in Celsius
+        /// </summary>
+        /// <returns>Returns the temperature formated in Celsius</returns>
+        public override string ToString() => $"{Temperature} °C";
+
+        /// <summary>
+        /// Get the name grove Temperature Sensor
+        /// </summary>
+        public new string SensorName => "grove Temperature Sensor";
```
krwq: grove -> Grove

src/devices/GrovePi/Sensors/ISensor.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        /// <summary>
+        /// Returns the raw value of the sensort as a formated string
+        /// </summary>
+        string ToString();
+
+        /// <summary>
+        /// Gets the name of the sensor.
+        /// </summary>
+		string SensorName { get; }
```
krwq: nic: spacing

src/devices/GrovePi/Sensors/ISensor.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Iot.Device.GrovePiDevice.Models;
+using System.Collections.Generic;
+
+namespace Iot.Device.GrovePiDevice.Sensors
+{
+    /// <summary>
+	/// Interface for a sensor 
+	/// </summary>
+    public interface ISensor<T>
```
krwq: if this can be internal please change (can't remember if this was in any public APIs or just base classes)

src/devices/GrovePi/Sensors/LedBar.cs
```diff
@@ -0,0 +1,156 @@
...
+        private GrovePi _grovePi;
+        private byte _level;
+        private LedBarOrientation _orientation;
+
+        /// <summary>
+        /// LedBar constructor
+        /// </summary>
+        /// <param name="grovePi">The GrovePi class</param>
+        /// <param name="port">The grove Port, need to be in the list of SupportedPorts</param>
+        public LedBar(GrovePi grovePi, GrovePort port) : this(grovePi, port, LedBarOrientation.GreenToRed)
```
krwq: I feel like most of the sensors do not need to depend on the GrovePi and could just take I2c as an input - this way you could use them separately and they could live outside of GrovePi directory

src/devices/GrovePi/Sensors/LedBar.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        /// <summary>
+        /// Set one led
+        /// </summary>
+        /// <param name="led">The led number from 0 to 10</param>
+        /// <param name="status">true for on, false for off</param>
+        public void SetOneLed(byte led, bool status)
+        {
+            led = Math.Clamp(led, (byte)0, (byte)10);
```
krwq: I'd rather throw in here since it likely would be unintentional to light non-existing led

src/devices/GrovePi/Sensors/SoundSensor.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Iot.Device.GrovePiDevice.Models;
+
+namespace Iot.Device.GrovePiDevice.Sensors
+{
+    /// <summary>
+    /// SoundSensor class to support basic Grove microphones
+    /// </summary>
+    class SoundSensor : AnalogSensor
```
krwq: should this be public?

src/devices/GrovePi/samples/GrovePiDevice.sample.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease.19078.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MSBuildThisFileDirectory)\..\GrovePiDevice.csproj" />
```
krwq: `$(MSBuildThisFileDirectory)\` is not needed

src/devices/GrovePi/GrovePi.cs
Ellerbach: typo, it's 1023

src/devices/GrovePi/GrovePi.cs
Ellerbach: The write command is taking 4 bytes. The version does not require the last 3 bytes. So they have to be set to 0 and sent. 
For sleep, I can try to remove it and run more tests as I I implemented a second retry. In tests versions, asking too many I2C commands would raise exceptions.

src/devices/GrovePi/GrovePi.cs
Ellerbach: the first inArray is always 0. And it is not part of the version. Only the last 3 bytes are

src/devices/GrovePi/GrovePi.cs
Ellerbach: Yes, it's needed to have the operation done correctly on the GrovePi side. Now, As  implemented a retry method in the write command, I can try without and run tests like this.

src/devices/GrovePi/Models/GrovePiCommands.cs
Ellerbach: No as if you want to implement your own sensors, you may need it

src/devices/GrovePi/Models/GrovePiCommands.cs
Ellerbach: pfff, totally forgot them… Will add

src/devices/GrovePi/Models/PinMode.cs
Ellerbach: Yes, same for PinLevel (PinValue). They are actually the same values in the enum. So all will work the same way

src/devices/GrovePi/Sensors/AnalogSensor.cs
Ellerbach: It actually can but the IGpioController is only for Digital Pins and not for analogic pins. And most people who are using Grove sensors which are not implemented will look at the Arduino code. And in the Arduino code, they'll find analogRead, digitalWrite, etc. And not the ones from the IGpioController. So in this case, I think it will bring more confusion than Clarity. If I would implement the IGpioController, I would anyway keep the DigitalWrite and DigitalRead functions.
GrovePi is here to bring analogic sensors to the RaspberyPi world and makes things easy. So I'll keep the philosophie of GrovePi.

src/devices/GrovePi/Sensors/DhtSensor.cs
Ellerbach: At best, it can share the interface. The rest is not working the same way.
I think I'll open an issue to start a discussion on standard interface for various sensors. They will appear with different interface and can create confusion. We will have the same for few others.

src/devices/GrovePi/Sensors/ISensor.cs
Ellerbach: If people want to build their own class with this interface, I would leave it public.

src/devices/GrovePi/Sensors/LedBar.cs
Ellerbach: Unfortunately, none of those sensors are I2C. they all need GrovePi. Those sensors are purely Analogic or Digital sensors. For the Digital ones, the way GrovePi works won't make it easy to have it in a generic way. But for most of the digital sensors, there maybe a way to do it.
and in this case, based on your feedback on IGpioController, it would then be possible. 
I'll update this PR first like it is and will dig into implementing the IGpioController and see how I can make all the digital sensors (including the PWM ones) generic

src/devices/GrovePi/GrovePi.cs
krwq: @Ellerbach What I meant is that you got `[2]` twice

src/devices/GrovePi/GrovePi.cs
krwq: perhaps add this delay before doing the retry

src/devices/GrovePi/Models/PinMode.cs
krwq: I meant that we already have it in Gpio namespace

src/devices/GrovePi/Sensors/AnalogSensor.cs
krwq: @Ellerbach I'd suggest to create an issue that we should have support for analog pins in Gpio

src/devices/GrovePi/Sensors/DhtSensor.cs
krwq: that's fine. what about moving this sensor outside of GrovePi directory so i.e. making it i.e.:
`src/devices/GroveDhtXXXX/`?

src/devices/GrovePi/Models/PinMode.cs
Ellerbach: yes, I got it. I used the Gpio ones.

src/devices/GrovePi/Sensors/AnalogSensor.cs
Ellerbach: I will do. also the IGpioInterface is not present in the nuget 0.1.0-prerelease.19078.2. Would be great to have a new nuget release that would include it.

src/devices/GrovePi/GrovePi.cs
Ellerbach: Yes, exactly what I was trying to test. I did a stress test wihout any wait. And after 15 minutes it failed. Will run a new one with the wait before the retry

src/devices/GrovePi/GrovePi.cs
Ellerbach: hooooooooooooooo

src/devices/GrovePi/Sensors/DhtSensor.cs
Ellerbach: I would leave it with GrovePi as it's really GrovePi specific. That said, if we fully reorganize the sensors, then why not. Still, this one is linked to GrovePi. The reading are done in GrovePi and there is a special function for that.

src/devices/GrovePi/Sensors/AnalogSensor.cs
Ellerbach: Issue for analogic support: https://github.com/dotnet/iot/issues/255

src/devices/GrovePi/Sensors/ISensor.cs
krwq: This is interesting idea to represent all sensors. I think we should also have specific types which use that as well. So like in my PR: https://github.com/dotnet/iot/pull/256 I presume ITemperatureSensor could also imply `ISensor<float>`? But currently that would also imply that we always return "Temperature sensor" in the SensorName field or should this be specific device name? If we go with more generic then interface can't do that in C# (AFAIK)

src/devices/GrovePi/GrovePi.cs
Ellerbach: I ran more tests and still failing after 30 minutes with older firmware (1.2.2). I'll implement multiple retries with the wait pattern. Let's see.

src/devices/GrovePi/Sensors/ISensor.cs
Ellerbach: Absolutely. The temperature sensor should implement ISensor<float>. I like to have at least a standard way to get access to the "raw" data of the sensor like Value.  And for the sensor name. Makes it easy to display. 
Now, yes, it's a bit broader discussion with the types as well. And yes, some sensors can have multiple interfaces like a temperature one, humidity one and we can think of few others. And I'll catch up in the other PR.

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,226 @@
+﻿using Iot.Device.GrovePiDevice.Models;
```
krwq: copyright header

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,226 @@
...
+            byte tries = 0;
+            IOException innerEx = new IOException();
+            while (tries < MaxRetries)
+            {
+                try
+                {
+                    _i2cDevice.Write(outArray);
+                    return;
+                }
+                catch (IOException ex)
```
krwq: you can do `catch (IOException ex) when (tries < MaxRetries)`
and then `while (true)` - it will just throw on the last try - fine to leave as is too

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,226 @@
...
+        /// <summary>
+        /// Read a digital pin, equivalent of digitalRead on Arduino
+        /// </summary>
+        /// <param name="pin">The GroovePi pin to read</param>
+        /// <returns>Returns the level either High or Low</returns>
+        public PinValue DigitalRead(GrovePort pin)
+        {
+            WriteCommand(GrovePiCommands.DigitalRead, pin, 0, 0);
+            byte tries = 0;
+            IOException innerEx = new IOException();
```
krwq: Why not IOException `innerEx = null` or if compiler allows just declaration? I don't think this value is ever used

Ellerbach: yes, it does accept null. The simple declaration was not enough. So I jsut created a new exception which is not needed as null is working.

src/devices/GrovePi/GrovePi.cs
Ellerbach: I4ll leave it like this so far. But I keep the pattern in mind for other cases where it may Apply better.

src/devices/GrovePi/GrovePi.cs
krwq: isn't that 10bit though?

Ellerbach: yes, it is. copy/paste from GoPiGo with mistakes. correcting.

JeremyKuhne:
FYI: I just got the GrovePi+ starter kit. I'm going through the RGB LCD and revamping it a bit. I've found a datasheet for what I believe the IC is copying.  It _should_ be a clone of the NXP PCF2119x, which was introduced back in 1997.  (AIP31068L is the clone part number, but I cannot find the datasheet for it.)

src/devices/GrovePi/GrovePi.cs
JeremyKuhne: Can we write something here describing why we're retrying? Ideally we don't swallow `IOException`, of course, but if there is a valid reason we should clearly state why.

src/devices/GrovePi/Sensors/DigitalOutput.cs
```diff
@@ -0,0 +1,82 @@
...
+        internal PinValue _value;
+
+        /// <summary>
+        /// DigitalOutput constructor
+        /// </summary>
+        /// <param name="grovePi">The GrovePi class</param>
+        /// <param name="port">The grove Port, need to be in the list of SupportedPorts</param>
+        public DigitalOutput(GrovePi grovePi, GrovePort port)
+        {
+            if (!SupportedPorts.Contains(port))
```
JeremyKuhne: The name of the argument should be included in the exception and you're not actually using the interpolated string.

``` C#
throw new ArgumentException($"Error: grove port {port} not supported.", nameof(port));
```

krwq: if we are changing it then I'd get rid of the `Error: ` prefix since that is implied by throwing an exception

krwq:
@JeremyKuhne - any chance you could take another look?

JeremyKuhne:
I'll review shortly

src/devices/GrovePi/Sensors/DigitalInput.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <summary>
+        /// Returns the value as a string
+        /// </summary>
+        /// <returns>Returns the value as a string</returns>
+        public override string ToString() => Value.ToString();
+
+        /// <summary>
+        /// Get the name Digital Input
+        /// </summary>
+        public string SensorName => "Digital Input";
```
JeremyKuhne: The concept of `SensorName` isn't one I think would actually provide value to solutions built on this. The type itself tells you what it is and I can't see how someone would use this in production code.

src/devices/GrovePi/Sensors/DigitalInput.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        /// <summary>
+        /// grove sensor port
+        /// </summary>
+        public GrovePort Port { get; internal set; }
```
JeremyKuhne: In general we haven't been exposing the pins, controllers, etc. passed to drivers. This should probably be an internal field.

src/devices/GrovePi/Sensors/LightSensor.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Iot.Device.GrovePiDevice.Sensors
+{
+    /// <summary>
+    /// LightSensor class for analogic luminosity sensors
+    /// </summary>
+    public class LightSensor : AnalogSensor
+    {
+        /// <summary>
+        ///
```
JeremyKuhne: Empty

src/devices/GrovePi/Sensors/PotentiometerSensor.cs
```diff
@@ -0,0 +1,30 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Iot.Device.GrovePiDevice.Models;
+
+namespace Iot.Device.GrovePiDevice.Sensors
+{
+    public class PotentiometerSensor : AnalogSensor
+    {
+        /// <summary>
+        ///
```
JeremyKuhne: Empty

src/devices/GrovePi/Sensors/Relay.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="grovePi">The GoPiGo3 class</param>
+        /// <param name="port">The grove Port, need to be in the list of SupportedPorts</param>
+        public Relay(GrovePi grovePi, GrovePort port) : this(grovePi, port, false)
+        { }
+
+        /// <summary>
+        ///
```
JeremyKuhne: Empty

src/devices/GrovePi/Sensors/Relay.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// </summary>
+        /// <param name="grovePi">The GoPiGo3 class</param>
+        /// <param name="port">The grove Port, need to be in the list of SupportedPorts</param>
+        public Relay(GrovePi grovePi, GrovePort port) : this(grovePi, port, false)
+        { }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="goPiGo">The GoPiGo3 class</param>
```
JeremyKuhne: Doesn't match

src/devices/GrovePi/Sensors/Relay.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <param name="inverted">If inverted, the relay is on when output is low and off when output is high</param>
+        public Relay(GrovePi grovePi, GrovePort port, bool inverted) : base(grovePi, port)
+        {
+            IsInverted = inverted;
+        }
+
+        /// <summary>
+        /// Switch on the relay
+        /// </summary>
+        public void On()
```
JeremyKuhne: This should probably be a property `public void IsOn { get... set... }`. Or `IsEnabled`. Right now there is no easy way to read if the state is on/off.

src/devices/GrovePi/Sensors/Relay.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        /// <summary>
+        /// Switch off the relay
+        /// </summary>
+        public void Off()
+        {
+            Value = PinValue.High;
+        }
+
+        public bool IsInverted { get; set; }
```
JeremyKuhne: Comment

src/devices/GrovePi/Sensors/Relay.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <summary>
+        /// Get/set the state of the relay, 0 for off, 1 for on. Wehn set, anything not 0 will be considered as on
+        /// </summary>
+        public new PinValue Value
+        {
+            get { return _value; }
+            set
+            {
+                _value = value;
+                _grovePi.DigitalWrite(Port, ((_value == PinValue.High) && !IsInverted) ? PinValue.High : PinValue.Low);
```
JeremyKuhne: If we're exposing the raw value we should also set the raw value. This probably isn't even needed- you can infer the `PinValue` if you need to from `IsOn/IsEnabled` and `IsInverted`. I presume needing to know this would be very rare and having just one get/set would make it easier to use this type correctly.

src/devices/GrovePi/Sensors/Relay.cs
JeremyKuhne: Also, should set be public? Presumably flipping this is something you shouldn't need to do.

src/devices/GrovePi/Sensors/Relay.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Get "On" when relay if on and "Off" when relay is off
+        /// </summary>
+        public override string ToString() => (_value == PinValue.High) ? "On" : "Off";
```
JeremyKuhne: This isn't right as it doesn't take inverted into account.

src/devices/GrovePi/Sensors/DhtSensor.cs
```diff
@@ -0,0 +1,128 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Iot.Device.GrovePiDevice.Sensors
+{
+    /// <summary>
+    /// Type of DHT sensors
+    /// </summary>
+    public enum DhtType
```
JeremyKuhne: Nit: This should be in a separate file.

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        /// <summary>
+        /// Get the firmware version
+        /// </summary>
+        /// <returns>GroovePi firmware version</returns>
+        public Version GetFirmwareVerion()
+        {
+            WriteCommand(GrovePiCommands.Version, 0, 0, 0);
+            var inArray = ReadCommand(GrovePiCommands.Version, 0);
```
JeremyKuhne: Per our [coding styles](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md), don't use `var` here.

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,239 @@
...
+        public Info GrovePiInfo { get; internal set; }
+
+        /// <summary>
+        /// GrovePi constructor
+        /// </summary>
+        /// <param name="i2cDevice">The I2C device. Device address is 0x04</param>
+        /// <param name="autoDispose">True to dispose the I2C device when disposing GrovePi</param>
+        public GrovePi(I2cDevice i2cDevice, bool autoDispose = true)
+        {
+            _i2cDevice = i2cDevice ?? throw new ArgumentException("I2C device can't be null", nameof(i2cDevice));
```
JeremyKuhne: Nit- the description isn't needed. The argument name and the exception tell you everything you need to know.

qwertoyo:
🐱‍🐉 I'd like to see this merged, I'm using python with a weird dockerfile to use these sensors at the moment, but my ideal solution would be to use dotnet.
Is the PR opener working on the comments? If not, may I help?

Ellerbach:
> 🐱‍🐉 I'd like to see this merged, I'm using python with a weird dockerfile to use these sensors at the moment, but my ideal solution would be to use dotnet.
> Is the PR opener (kudos) working on the comments? If not, may I help?

I need to find the time to work on this PR. In the mean time, you can use the PR by checking out the branch of this PR and work on it. Looking at the feedbacks, not everything will change. You may have later to adjust some of the elements but not everything. And your feedback will be very valuable.

Ellerbach:
> 🐱‍🐉 I'd like to see this merged, I'm using python with a weird dockerfile to use these sensors at the moment, but my ideal solution would be to use dotnet.
> Is the PR opener (kudos) working on the comments? If not, may I help?

@qwertoyo, other option is that you can take this PR, look at feedbacks from Jeremy and work on the PR ,as well. As I wrote in the previous comment, you can take the PR as a branch and work on it.

src/devices/GrovePi/Models/GrovePiCommands.cs
```diff
@@ -0,0 +1,64 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.GrovePiDevice.Models
+{
+    /// <summary>
+    /// GroovePi commands to read, write, setup pins and access special sensors
+    /// Note that those commands are supports in most of the recent firmware with version higher than 1.2.1
+    /// </summary>
+    public enum GrovePiCommands
```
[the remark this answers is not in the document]
Ellerbach: For simplicity reasons I would just keep all the commands in the same enum. Splitting is possible of course but can create a bit of confusion. And the commands are present mainly for advanced users. Most will just use PinMode, AnalogRead, AnalogWrite, DigitalWrite, DigitalRead.

src/devices/GrovePi/Models/GrovePiInfo.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.GrovePiDevice.Models
+{
+    public class Info
```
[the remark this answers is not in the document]
Ellerbach: I write BrickPi and GoPiGo3 as well, for those 2 others, there is a string that can be read from the hardware. And the classes are quite similar. I've decided to return those 2 strings for consistency.

src/devices/GrovePi/Models/GrovePort.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// The supported Grove Ports
+    /// See README.md files for the exact location of each Port
+    /// </summary>
+    public enum GrovePort
+    {
+        AnalogPin0 = 0,
+        AnalogPin1 = 1,
+        AnalogPin2 = 2,
+        DigitalPin2 = 2,
```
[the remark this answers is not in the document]
Ellerbach: Yes, I can split the enum into 2 different enums to make it easier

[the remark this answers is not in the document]
Ellerbach: Well, actually, it is creating more complication than not doing it. The WriteCommand for example will need to be double. So quite some additional code for very little gain and creating quite a lot of confusion. So I'll keep it with only 1 enum.

src/devices/GrovePi/Sensors/ISensor.cs
```diff
@@ -0,0 +1,34 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Iot.Device.GrovePiDevice.Models;
+
+namespace Iot.Device.GrovePiDevice.Sensors
+{
+    /// <summary>
+	/// Interface for a sensor 
+	/// </summary>
+    public interface ISensor<T>
```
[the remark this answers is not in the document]
Ellerbach: I can get rid of the interface, that's no issue. I will anyway keep the functions as they bring consistency between the sensors.

src/devices/GrovePi/Sensors/Relay.cs
Ellerbach: I've added an IsOn property. Will keep this Value for consistency with the rest of the sensors.

src/devices/GrovePi/Sensors/LedBar.cs
krwq: @Ellerbach I think this comment got missed

src/devices/GrovePi/samples/Program.cs
```diff
@@ -10,6 +10,7 @@
 using System.Device.I2c;
 using System.Device.I2c.Drivers;
 using System.Device.Gpio;
+using Iot.Device.GroovePiDevice.Models;
```
krwq: Grove

krwq:
@Ellerbach except for one missed comment from previous and renaming `Groove` -> `Grove` looks like this should be ready. @JeremyKuhne, do you want to take another look?

src/devices/GrovePi/samples/Program.cs
Ellerbach: arg, will fix this one.

src/devices/GrovePi/Sensors/LedBar.cs
Ellerbach: yes, missed that one, good catch! Fixed

src/devices/GrovePi/GrovePiDevice.csproj
```diff
@@ -0,0 +1,22 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <RootNamespace>Iot.Device.GroovePiDevice</RootNamespace>
+    <LangVersion>7.3</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease.19078.2" />
```
joperezr: This is not right. Can you please instead use version $(SystemDeviceGpioPackageVersion) ? That is what we are doing in the rest of the repo

src/devices/GrovePi/Sensors/DhtSensor.cs
```diff
@@ -0,0 +1,118 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Iot.Device.GrovePiDevice.Models;
+using Iot.Device.GrovePiDevice.Models;
```
joperezr: Build is failing because you have this using namespace twice.

Ellerbach: yes, correcting it

src/devices/GrovePi/GrovePiDevice.csproj
Ellerbach: will adjust, the PR is so old that it was not the case originally. Thanks for checking! Will adjust as well

krwq:
@joperezr want to review? I think this has been open long enough and is in reasonably good state - we can fix any remaining feedback separately

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,240 @@
...
+    /// </summary>
+    public class GrovePi : IDisposable
+    {
+        private I2cDevice _i2cDevice;
+        private readonly bool _autoDispose;
+        private const byte MaxRetries = 4;
+
+        /// <summary>
+        /// The default GrovePi I2C address is 0x04
+        /// Other addresses can be use, see GrovePi documentation
```
joperezr: NIT: I wonder if it would be worth it to add here the alternative I2c addresses as well. It might be nice so that users don't have to look into datasheet in case they want to use an alternate address. Anyways, you don't have to fix it, just thinking out loud.

src/devices/GrovePi/GrovePi.cs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        /// <summary>
+        /// Write a GrovePi command
+        /// </summary>
+        /// <param name="command">The GrovePi command</param>
+        /// <param name="pin">The pin to write the command</param>
+        /// <param name="param1">First parameter</param>
+        /// <param name="param2">Second parameter</param>
+        public void WriteCommand(GrovePiCommand command, GrovePort pin, byte param1, byte param2)
```
joperezr: Should we consider making these lower-level APIs internal instead? seems like you have a wrapper API for all commands, so perhaps we should hide Write and Read from the public surface area just to make API easier to use.

src/devices/GrovePi/GrovePi.cs
krwq: the way this device is built it's rather unlikely anyone will want to change this (basically there is only one reasonable way to connect it)

joperezr:
Left small comments but this looks good otherwise. Main thing for me is that I would try to simplify the surface area with GroovePi, maybe the best way would be to simply have a WriteCommand and ReadCommand, and don't have those DigitalReads and instead simply force people to manually call the WriteCommand and specify what is it that they want to read.

src/devices/GrovePi/GrovePi.cs
Ellerbach: There is a way to change the address programmatically thru some scripts. To implement that, it would need to add SPI support and basically do what the script is doing. I Don't see much interest. Usually what people do is changing it before they start using it. If they want to have multiple GrovePi, they need anyway to change one by one with the script.

src/devices/GrovePi/GrovePi.cs
Ellerbach: As I have implemented some sensors only, I would live this part public so anyone who want to implement their own sensor can really use all low level as well. Now Read and Write will make it in 90% of the cases. Just living the door open for the 10%

joperezr:
I think this is good to go now. Thanks for all the great work here @Ellerbach!

PR #247, opened 2019-02-18
Update readme for stepper motor

johnastaf:
The changes related to https://github.com/dotnet/iot/pull/227 pull request.

There are updated readme files and small fixes.

### DIFF
--- a/src/devices/Uln2003/Uln2003.cs
+++ b/src/devices/Uln2003/Uln2003.cs
@@ -127,7 +127,7 @@ namespace Iot.Device.Uln2003
         public void Step(int steps)
         {
             double lastStepTime = 0;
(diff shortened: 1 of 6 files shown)


src/devices/Uln2003/Uln2003.cs
```diff
@@ -126,7 +127,7 @@ public void Stop()
         public void Step(int steps)
         {
             double lastStepTime = 0;
-            _stopwatch = Stopwatch.StartNew();
+            _stopwatch.Start();
```
krwq: `Restart()` will be equivalent to previous logic

johnastaf: Fixed

PR #248, opened 2019-02-18
Added third party notices, fixed pin numbers.

SergeyRazmyslov:
Added third party notices for Grove LCD RGB Backlight. Also fixed pin numbers for LCD sample.


krwq:
@joperezr I think we should add it for everything even if it's not required. Merging this - we can fix up anything later

PR #250, opened 2019-02-18
PCA9685: 16-channel, 12-bit PWM Fm+ I²C-bus LED controller

HumJ0218:


src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// </summary>
+        public void SetPwmFrequency(double freq_hz)
+        {
+            SetPwmFrequency(GetPrescale(freq_hz));
+        }
+
+        /// <summary>
+        /// Get prescale of specified PWM frequency
+        /// </summary>
+        public byte GetPrescale(double freq_hz)
```
ZhangGaoxing: ditto

src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            device.Write(new byte[] { (byte)ALL_LED_ON_L, (byte)(on & 0xFF) });
+            device.Write(new byte[] { (byte)ALL_LED_ON_H, (byte)(on >> 8) });
+            device.Write(new byte[] { (byte)ALL_LED_OFF_L, (byte)(off & 0xFF) });
+            device.Write(new byte[] { (byte)ALL_LED_OFF_H, (byte)(off >> 8) });
+        }
+
+        public void Dispose()
+        {
+            device.Dispose();
```
ZhangGaoxing: device?.Dispose();
device=null;

src/devices/Pca9685/samples/Pca9685.Sample.cs
```diff
@@ -0,0 +1,102 @@
...
+using Iot.Device.Pca9685;
+using System;
+
+namespace Iot.Device.Pca9685.Samples
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            var pca9685 = new Pca9685();
```
ZhangGaoxing: using(Pca9685 pca9685 = new Pca9685())
{

}

ZhangGaoxing:
Left some feedback notes😁

src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,187 @@
...
+            device = i2cDevice;
+
+            Init();
+        }
+
+        private void Init()
+        {
+
+            SetPwm(0, 0);
+            device.Write(new byte[] { MODE2, OUTDRV });
```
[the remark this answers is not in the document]
krwq: To expand this:
```csharp
Span<byte> buffer = stackalloc byte[] { ... }; // or some better name than buffer
device.Write(buffer);
```

src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,187 @@
...
+            return ClockRate / 4096 / (prescale + 1);
+        }
+
+        /// <summary>
+        /// Set PWM frequency by using prescale
+        /// </summary>
+        public void SetPwmFrequency(byte prescale)
+        {
+            var oldmode = device.ReadByte();
+            var newmode = (oldmode & 0x7F) | 0x10; // sleep
```
krwq: please at least add the comment what both flags mean

src/devices/Pca9685/samples/Pca9685.Sample.cs
```diff
@@ -0,0 +1,102 @@
...
+                                Console.WriteLine($"PWM Frequency has been set to about {freq}Hz with prescale is {prescale}");
+                                break;
+                            }
+                        case 's':
+                            {
+                                switch (command.Length)
+                                {
+                                    case 2:
+                                        {
+                                            var off = int.Parse(command[1]);
```
krwq: please add functions with descriptive names for each case

src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,206 @@
...
+    /// <summary>
+    /// PCA9685 PWM LED/servo controller
+    /// </summary>
+    public class Pca9685 : IDisposable
+    {
+
+        /// <summary>
+        /// I2C Device
+        /// </summary>
+        public I2cDevice Device { get; private set; }
```
ZhangGaoxing: Private field
```c#
private I2cDevice _device;
```

src/devices/Pca9685/Bit.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Pca9685
+{
+    internal enum Bit : byte
```
ZhangGaoxing: please add some comments😀

src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,206 @@
...
+        public Pca9685(I2cDevice i2cDevice)
+        {
+            // Setup I2C interface for the device.
+            Device = i2cDevice;
+
+            SetPwm(0, 0);
+
+            Span<byte> buffer = stackalloc byte[2];
+
+            buffer[0] = (byte)MODE2;
```
ZhangGaoxing: try this 😀
```c#
Span<byte> buffer = stackalloc byte[2] { (byte)MODE2, (byte)OUTDRV };

File: src/devices/Pca9685/samples/Pca9685.Sample.cs [hunk 10]
```diff
@@ -0,0 +1,122 @@
...
+    /// <summary>
+    /// This sample is for Raspberry Pi Model 3B+
+    /// </summary>
+    class Program
+    {
+        static void Main(string[] args)
+        {
+
+            var busId = 1;  // /dev/i2c-1
+            var deviceAddress = 0x40;
```
ZhangGaoxing: As far as I know, this board I2C address is customizable. Can we add a class to help calculate the address? Of course, it's not necessary. I can do it after the code merge.

src/devices/Pca9685/Pca9685.cs
HumJ0218: I have a question - enum Register extends byte. Why I need explicitly convert MODE2 to byte when putting it into a byte array?

ZhangGaoxing: if you don't extend byte, the code will look like this
```C#
public enum Register : int
{

}
```

HumJ0218: but why enum:byte to byte needs **explicit** conversion?

src/devices/Pca9685/Bit.cs
krwq: this needs different name describing what this means - `Bit` is a bit generic :smile:

src/devices/Pca9685/Pca9685.cs
krwq: @HumJ0218 this is a good question and I find this really annoying as well, might be worth asking this question here: https://github.com/dotnet/roslyn/issues

src/devices/Pca9685/samples/Pca9685.Sample.cs
krwq: @ZhangGaoxing is it customizable with external pins? If so I'd recommend adding const for whatever the default value is and put that info in the README instead (not necessary to go into details just mention that and point to the datasheet).

ZhangGaoxing: The universal address of the board is 0x70. Change the address by changing the hardware address pin on the board. Custom address calculate method: **address = 0x40 | 0b_A5A4_A3A2A1A0**
@HumJ0218 @krwq 
42022263-6e072926-7aef-11e8-8193-c27a6e64700b

src/devices/Pca9685/Pca9685.cs
HumJ0218: @krwq I found it in .NET document that I really need an explicit conversion.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/enum#example-1

    enum Range : long { Max = 2147483648L, Min = 255L };
    static void Main()
    {
        long x = (long)Range.Max;
        ...
    }

src/devices/Pca9685/samples/Pca9685.Sample.cs
krwq: I'd with just a const `I2cAddress`/`DefaultI2cAddress` and you feel like it perhaps also add an enum `CustomAddressSetting` with `[Flags]` with `A0`-`A5` values and a static function `GetCustomAddress(CustomAddressSetting addressSetting)`

src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,192 @@
...
+using static Iot.Device.Pca9685.Mode2;
+
+namespace Iot.Device.Pca9685
+{
+    /// <summary>
+    /// PCA9685 PWM LED/servo controller
+    /// </summary>
+    public class Pca9685 : IDisposable
+    {
+
```
krwq: extra enter

src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,192 @@
...
+            set
+            {
+                Prescale = GetPrescale(value);
+            }
+        }
+
+        /// <summary>
+        /// Set PWM frequency using prescale value or get the value.
+        /// </summary>
+        public byte Prescale
```
krwq: I think this could be internal since you can already do it with the other property - personally find it a bit confusing but fine to leave if you think it adds enough value

src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,192 @@
...
+            Thread.Sleep(5); // wait for oscillator
+
+            int mode1 = _device.ReadByte();
+            mode1 &= ~(byte)SLEEP; // wake up (reset sleep)
+
+            buffer = stackalloc byte[2] { (byte)MODE1, (byte)mode1 };
+            _device.Write(buffer);
+
+            Thread.Sleep(5); // wait for oscillator
+
```
krwq: extra enter

src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            buffer = stackalloc byte[2] { (byte)MODE1, oldmode };
+            _device.Write(buffer);
+
+            Thread.Sleep(5);
+
+            buffer = stackalloc byte[2] { (byte)MODE1, (byte)(oldmode | (byte)RESTART) };
+            _device.Write(buffer);
+
```
krwq: extra enter

src/devices/Pca9685/Pca9685.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
```
krwq: extra enter

src/devices/Pca9685/Register.cs
```diff
@@ -0,0 +1,86 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Pca9685
+{
+    internal enum Register : byte
+    {
+
```
krwq: extra enter

src/devices/Pca9685/Register.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <summary>
+        /// load all the LEDn_OFF registers, byte 1
+        /// </summary>
+        ALL_LED_OFF_H = 0xFD,
+
+        /// <summary>
+        /// prescaler for PWM output frequency
+        /// </summary>
+        PRESCALE = 0xFE,
+
```
krwq: extra enter

src/devices/Pca9685/samples/Pca9685.Sample.cs
```diff
@@ -0,0 +1,125 @@
...
+namespace Iot.Device.Pca9685.Samples
+{
+    /// <summary>
+    /// This sample is for Raspberry Pi Model 3B+
+    /// </summary>
+    class Program
+    {
+        static void Main(string[] args)
+        {
+
```
krwq: extra enter

src/devices/Pca9685/samples/Pca9685.Sample.cs
```diff
@@ -0,0 +1,125 @@
...
+            var deviceAddress_fixed = 0x40;
+            var deviceAddress_selectable = 0b000000;    // A5 A4 A3 A2 A1 A0
+            var deviceAddress = deviceAddress_fixed | deviceAddress_selectable;
+
+            var settings = new I2cConnectionSettings(busId, deviceAddress);
+            var device = new UnixI2cDevice(settings);
+
+            using (var pca9685 = new Pca9685(device))
+            {
+
```
krwq: ditto

src/devices/Pca9685/samples/Pca9685.Sample.cs
```diff
@@ -0,0 +1,125 @@
...
+                Console.WriteLine("Command:  F {freq_hz}             set PWM frequency");
+                Console.WriteLine("          P {prescale}            set PRE_SCALE register");
+                Console.WriteLine("          S {off}                 set off step with on step is 0 to all channels");
+                Console.WriteLine("          S {on} {off}            set on step and off step to all channels");
+                Console.WriteLine("          S {on} {off} {channel}  set on step and off step to specified channel");
+
+                Console.WriteLine();
+                while (true)
+                {
+
```
krwq: ditto

src/devices/Pca9685/samples/Pca9685.Sample.cs
```diff
@@ -0,0 +1,125 @@
...
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        Console.ForegroundColor = ConsoleColor.Red;
+                        Console.WriteLine(ex.GetBaseException().Message);
+                        Console.ResetColor();
+                    }
+                }
+
```
krwq: ditto

src/devices/Pca9685/samples/Pca9685.Sample.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+
```
krwq: ditto

src/devices/Pca9685/Pca9685.cs
HumJ0218: PCA9685's PWM frequency is directly linked to `Prescale`. Users can change it or change `Frequency`(indirectly, but calculating creates deviation). If actual frequency is different from users want, they can know why from `Prescale`.

HumJ0218:
`System.Device.Gpio.Tests` Build FAILED. ?????

HumJ0218:
`System.Device.Gpio.Tests` Build FAILED. ???
@zhanggaoxing @krwq

HumJ0218:
@ZhangGaoxing @krwq
System.Device.Gpio.Tests Build FAILED.

because of

@shaggygi Updated parameter naming https://github.com/dotnet/iot/commit/31cf85c9a98f26d07c01d85330ec75b3f084d3cb

shaggygi:
@HumJ0218 just created a PR to fix the DeviceApiTester code. #290

krwq:
Thanks @HumJ0218 - I have restarted all builds after taking @shaggygi's fix and it is green now

shaggygi:
I've heard @jaredpar talking about this some.  does this mean i get the squirrel for breaking?  🤕

jaredpar:
Generally the break has to be merged before you qualify for the squirrel ;)

krwq:
It was merged and fixed :-) I blame using PackageReference vs ProjectReference rather than @shaggygi though

jaredpar:
Definitely squirrel worthy then.

PR #254, opened 2019-02-19
Improve template

krwq:
Improve template

### DIFF
--- a/tools/templates/DeviceBindingTemplate/dotnet_new_device-binding_csharp/_DeviceBinding.csproj
+++ b/tools/templates/DeviceBindingTemplate/dotnet_new_device-binding_csharp/_DeviceBinding.csproj
@@ -4,10 +4,11 @@
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <!--Disabling default items so samples source won't get build by the main library-->
     <EnableDefaultItems>false</EnableDefaultItems>
+    <LangVersion>Latest</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


tools/templates/DeviceBindingTemplate/dotnet_new_device-binding_csharp/_DeviceBinding.csproj
```diff
@@ -4,10 +4,11 @@
     <TargetFramework>netcoreapp2.1</TargetFramework>
```
joperezr: I know you didn't change this in your PR, but we should probably think if it's worth it to change the default target framework to be netstandard2.0 instead. That way, we will have bindings use API that will be available for boards that don't run .NET Core. We can address this in a different PR if we think it's valuable.

krwq: @joperezr we remove csprojes and create an uber project for the package later anyway so does it matter? We would have to go all or nothing, right? I think this might impede the development at this point - perhaps we should bulk address it later?

joperezr: that is true today, but we might want to not have all the bindings in one package in the future, or we might have customers that would just want to copy paste the code from the binding into their project, in which case it might just work if they targeted netstandard2.0 because the API used in the binding would have that already. In any case, I do agree that this is something we can worry about in the future.

PR #256, opened 2019-02-19
T - things

krwq:
I've created a stub for repo with abstraction for sensors. I wanted to start to standardize naming and conventions through this interfaces

Right now only added temperature - please let me know if I should go further with this or if this is a bad idea.

Feels like it is better to discuss on the interface rather than in the issue.

@joperezr @JeremyKuhne @shaggygi @ZhangGaoxing @Ellerbach

### DIFF
--- a/src/devices/Dhtxx/DhtSensor.cs
+++ b/src/devices/Dhtxx/DhtSensor.cs
@@ -3,12 +3,13 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Device;
 using System.Device.Gpio;
 using System.Diagnostics;
 
 namespace Iot.Device.DHTxx
 {
-    public class DHTSensor : IDisposable
+    public class DHTSensor : ITemperatureSensor, IDisposable
     {
         private const int MAX_TIME = 85;
         private const uint MAX_WAIT = 255;
@@ -22,7 +23,7 @@ namespace Iot.Device.DHTxx
 
         /// <summary>
         /// Wait for a specific number of milliseconds
-        /// 
+        ///
         /// </summary>
         /// <param name="milliseconds">Number of milliseconds to wait</param>
         /// <remarks>
@@ -51,7 +52,7 @@ namespace Iot.Device.DHTxx
         /// <remarks>
         /// If last read was not successfull, it returns double.MaxValue
         /// </remarks>
-        public double Temperature => (_dhtType == DhtType.Dht11) ? GetTempDht11() : GetTempDht22();
+        public float Temperature => (float)((_dhtType == DhtType.Dht11) ? GetTempDht11() : GetTempDht22());
 
         /// <summary>
         /// Get the last read temperature in Farenheit
@@ -210,7 +211,7 @@ namespace Iot.Device.DHTxx
                 if (counter == MAX_WAIT)
                     break;
 
-                // top 3 transistions are ignored   
+                // top 3 transistions are ignored
                 if ((i >= 4) && (i % 2 == 0))
                 {
                     _dht11Val[j / 8] <<= 1;
--- a/src/devices/Dhtxx/DhtSensor.cs
+++ b/src/devices/Dhtxx/DhtSensor.cs
@@ -52,7 +52,7 @@ namespace Iot.Device.DHTxx
         /// <remarks>
         /// If last read was not successfull, it returns double.MaxValue
         /// </remarks>
-        public float Temperature => (float)((_dhtType == DhtType.Dht11) ? GetTempDht11() : GetTempDht22());
+        public double Temperature => (_dhtType == DhtType.Dht11) ? GetTempDht11() : GetTempDht22();
 
         /// <summary>
         /// Get the last read temperature in Farenheit
--- a/src/devices/things/ITemperatureSensor.cs
+++ b/src/devices/things/ITemperatureSensor.cs
@@ -1,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device
+{
+    /// <summary>
+    /// Abstraction for temperature
+    /// </summary>
+    public interface ITemperatureSensor
+    {
+        float Temperature { get; }
+    }
+}
--- a/src/devices/things/ITemperatureSensor.cs
+++ b/src/devices/things/ITemperatureSensor.cs
@@ -5,10 +5,10 @@
 namespace Iot.Device
 {
     /// <summary>
-    /// Abstraction for temperature
+    /// Abstraction for temperature sensor
     /// </summary>
     public interface ITemperatureSensor
     {
-        float Temperature { get; }
+        double Temperature { get; }
     }
 }
--- a/src/devices/things/things.csproj
+++ b/src/devices/things/things.csproj
@@ -1,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <!--Disabling default items so samples source won't get build by the main library-->
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <None Include="README.md" />
+    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
+  </ItemGroup>
+
+</Project>
--- a/src/devices/Bmp280/Bmp280.cs
+++ b/src/devices/Bmp280/Bmp280.cs
@@ -12,7 +12,7 @@ using System.Buffers.Binary;
 
 namespace Iot.Device.Bmp280
 {
-    public class Bmp280 : IDisposable
+    public class Bmp280 : ITemperatureSensor, IDisposable
     {
         private const byte Signature = 0x58;
 
@@ -37,6 +37,8 @@ namespace Iot.Device.Bmp280
             _calibrationData = new CalibrationData();
             _communicationProtocol = CommunicationProtocol.I2c;
         }
+
+        public double Temperature => ReadTemperatureAsync().Result;
 
         private void Begin()
         {
--- a/src/devices/Bmp280/Bmp280.csproj
+++ b/src/devices/Bmp280/Bmp280.csproj
@@ -13,6 +13,7 @@
     <Compile Include="Register.cs" />
     <Compile Include="Sampling.cs" />
     <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
+    <ProjectReference Include="..\things\things.csproj" />
   </ItemGroup>
   
   <ItemGroup> 
--- a/src/devices/Dhtxx/Dhtxx.csproj
+++ b/src/devices/Dhtxx/Dhtxx.csproj
@@ -13,7 +13,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
-    <ProjectReference Include="../things/things.csproj" />
+    <ProjectReference Include="..\things\things.csproj" />
     <None Include="README.md" />
   </ItemGroup>
 
--- a/src/devices/Dhtxx/Dhtxx.csproj
+++ b/src/devices/Dhtxx/Dhtxx.csproj
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
+    <ProjectReference Include="../things/things.csproj" />
     <None Include="README.md" />
   </ItemGroup>
 
--- a/src/devices/Ds3231/Ds3231.cs
+++ b/src/devices/Ds3231/Ds3231.cs
@@ -10,7 +10,7 @@ namespace Iot.Device.Ds3231
     /// <summary>
     /// Realtime Clock DS3231
     /// </summary>
-    public class Ds3231 : IDisposable
+    public class Ds3231 : ITemperatureSensor, IDisposable
     {
         /// <summary>
         /// DS3231 I2C Address
--- a/src/devices/Ds3231/Ds3231.csproj
+++ b/src/devices/Ds3231/Ds3231.csproj
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
+    <ProjectReference Include="..\things\things.csproj" />
     <Compile Include="Ds3231.cs" />
     <Compile Include="Ds3231Data.cs" />
     <Compile Include="Register.cs" />
--- a/src/devices/Si7021/Si7021.cs
+++ b/src/devices/Si7021/Si7021.cs
@@ -7,7 +7,7 @@ using System.Device.I2c;
 
 namespace Iot.Device.Si7021
 {
-    public class Si7021 : IDisposable
+    public class Si7021 : ITemperatureSensor, IDisposable
     {
         private const byte ReadTemperatureCommand = 0xE3;
         private const byte ReadHumidityCommand = 0xE5;
@@ -22,15 +22,17 @@ namespace Iot.Device.Si7021
         {
         }
 
+        public double Temperature => ReadTemperatureInCelsius();
+
         public double ReadTemperatureInFahrenheit()
         {
-            double tempCelcius = ReadTemperatureInCelcius();
+            double tempCelcius = ReadTemperatureInCelsius();
             double tempFahrenheit = (tempCelcius * (9 / 5)) + 32;
             
             return tempFahrenheit;
         }
         
-        public double ReadTemperatureInCelcius()
+        public double ReadTemperatureInCelsius()
         {
             byte[] buffer = new byte[2];
 
--- a/src/devices/Si7021/Si7021.csproj
+++ b/src/devices/Si7021/Si7021.csproj
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
+    <ProjectReference Include="..\things\things.csproj" />
     <Compile Include="Si7021.cs" />
     <None Include="README.md" />
   </ItemGroup>
--- a/src/devices/Si7021/samples/Si7021.sample.cs
+++ b/src/devices/Si7021/samples/Si7021.sample.cs
@@ -24,7 +24,7 @@ namespace Iot.Device.Samples
             using (i2CSi7021)
             {
                 Console.WriteLine($"Temperature in fahrenheit: {i2CSi7021.ReadTemperatureInFahrenheit()}");
-                Console.WriteLine($"Temperature in celcius: {i2CSi7021.ReadTemperatureInCelcius()}");
+                Console.WriteLine($"Temperature in celsius: {i2CSi7021.ReadTemperatureInCelsius()}");
                 Console.WriteLine($"Relative humidity is: {i2CSi7021.ReadHumidity()}%");
             }
         }
--- a/src/devices/things/README.md
+++ b/src/devices/things/README.md
@@ -1,0 +1,5 @@
+﻿# things
+
+T - things
+
+Set of interfaces to standardize naming and types across this repo.


src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -51,7 +52,7 @@ private void Wait(double milliseconds)
         /// <remarks>
         /// If last read was not successfull, it returns double.MaxValue
         /// </remarks>
-        public double Temperature => (_dhtType == DhtType.Dht11) ? GetTempDht11() : GetTempDht22();
+        public float Temperature => (float)((_dhtType == DhtType.Dht11) ? GetTempDht11() : GetTempDht22());
```
krwq: didn't want to make further refactoring before we decide on returned type. I think that float is enough since abstraction should be a common denominator and float's precision reflects that well IMO (we can always have IHighPrecisionTemperatureSensor there in the future which could take BigInteger decimal)

src/devices/things/ITemperatureSensor.cs
```diff
@@ -0,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device
```
krwq: Perhaps this should be Iot.Thing since that's what IOT stands for

src/devices/things/ITemperatureSensor.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    /// <summary>
+    /// Abstraction for temperature
+    /// </summary>
+    public interface ITemperatureSensor
```
krwq: or maybe just ITemperature

Ellerbach:
Based on the discussion we had here https://github.com/dotnet/iot/pull/246#pullrequestreview-205087060, I would go fo a generic interface for sensors like the one I proposed ISensor<T>. So we will have at least the Value, the SensorName. Then add few more specific interfacves like Temperature, Humidity, etc.

src/devices/things/ITemperatureSensor.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        float Temperature { get; }
```
ZhangGaoxing: Doesn't it need to specify method name as well? like
```C#
float GetTemperature();
```

Ellerbach: If there is a property, I don't think we need a function especially if the function does not bring anything different.

src/devices/things/ITemperatureSensor.cs
shaggygi: This could have static helper methods to convert between Fahrenheit, Celsius, Kelvin, etc.

shaggygi:
This could possible be located in the Common directory and not separate.

Ellerbach:
I've been looking at all the sensors posted so far everywhere on the src/device. And I came with the Following interfaces that will cover most cases and can be reusable for similar families of sensors:

```csharp
/// <summary>
	/// Interface for a sensor 
	/// </summary>
    public interface ISensor<T>
    {
        /// <summary>
        /// Property to get the raw value of the sensor as an int
        /// </summary>
        T Value { get; }

        /// <summary>
        /// Returns the raw value of the sensort as a formated string
        /// </summary>
        string ToString();

        /// <summary>
        /// Gets the name of the sensor.
        /// </summary>
        string SensorName { get; }

        /// <summary>
        /// Just read the sensor
        /// </summary>
        T ReadRaw();

        /// <summary>
        /// Some sensors may not raise an exception
        /// So that can help to check is measures are correct
        /// </summary>
        bool IsValidData { get; }
    }

    public interface ITemperature : ISensor<float>
    {
        float Temperature { get; }
        float TemperatureInFarenheit { get; }
    }

    public interface IHumidity : ISensor<float>
    {
        float Humidity { get; }
    }

    // Reason is if you mix the interface with the sensors, for Value, you'll get only 1
    // not the 2 values. Maybe there is a better wau to do it
    public interface ITemperatureHymidity : ISensor<float[]>
    {
        float Temperature { get; }
        float TemperatureInFarenheit { get; }
        float Humidity { get; }
    }

    public interface IAcceleration : ISensor<Vector3>
    {
        Vector3 Acceleration { get; }
    }

    public interface ISonar: ISensor<float>
    {        
        float Distance { get; }
        // All distance are in cm ? or mm?
        // So this should give .01 is in cm
        float UnitToMeter { get; }
    }

    public interface ICompas:ISensor<Vector3>
    {
        Vector3 DirectionVector { get;  }
    }

    public interface IScreen
    {
        // Something for rotation
        TypeForRotation Rotation { get; set; }
        void Write(string value);
        void Write(byte value);
        void Write(Span<byte> data);
        void SetCursor(Point point);
        void Clear();
        void Home();
        void ShitScreenLeft();
        void ShitScreenRight();
        void ShitScreenUp();
        void ShitScreenDown();
        void ShiftCursorRight();
        void ShiftCursorLeft();
        void ShiftCursorUp();
        void ShiftCursorDown();
        bool LeftToRight { get; set; }
    }
```

Ellerbach:
I've been looking at all the sensors posted so far everywhere on the src/device. And I came with the Following interfaces that will cover most cases and can be reusable for similar families of sensors:

```csharp
/// <summary>
	/// Interface for a sensor 
	/// </summary>
    public interface ISensor<T>
    {
        /// <summary>
        /// Property to get the raw value of the sensor as an int
        /// </summary>
        T Value { get; }

        /// <summary>
        /// Returns the raw value of the sensort as a formated string
        /// </summary>
        string ToString();

        /// <summary>
        /// Gets the name of the sensor.
        /// </summary>
        string SensorName { get; }

        /// <summary>
        /// Just read the sensor
        /// </summary>
        T ReadRaw();

        /// <summary>
        /// Some sensors may not raise an exception
        /// So that can help to check is measures are correct
        /// </summary>
        bool IsValidData { get; }
    }

    public interface ITemperature : ISensor<float>
    {
        float Temperature { get; }
        float TemperatureInFarenheit { get; }
    }

    public interface IHumidity : ISensor<float>
    {
        float Humidity { get; }
    }

    // Reason is if you mix the interface with the sensors, for Value, you'll get only 1
    // not the 2 values. Maybe there is a better wau to do it
    public interface ITemperatureHymidity : ISensor<float[]>
    {
        float Temperature { get; }
        float TemperatureInFarenheit { get; }
        float Humidity { get; }
    }

    public interface IAcceleration : ISensor<Vector3>
    {
        Vector3 Acceleration { get; }
    }

    public interface ISonar: ISensor<float>
    {        
        float Distance { get; }
        // All distance are in cm ? or mm?
        // So this should give .01 is in cm
        float UnitToMeter { get; }
    }

    public interface ICompas:ISensor<Vector3>
    {
        Vector3 DirectionVector { get;  }
    }

    public interface IScreen
    {
        // Something for rotation
        TypeForRotation Rotation { get; set; }
        void Write(string value);
        void Write(byte value);
        void Write(Span<byte> data);
        void SetCursor(Point point);
        void Clear();
        void Home();
        void ShitScreenLeft();
        void ShitScreenRight();
        void ShitScreenUp();
        void ShitScreenDown();
        void ShiftCursorRight();
        void ShiftCursorLeft();
        void ShiftCursorUp();
        void ShiftCursorDown();
        bool LeftToRight { get; set; }
    }
```

Ellerbach:
Similar work will need to be done for Motors, Vehicle. And those are just interfaces for couple of very common sensors. There are tons of them. So idea is not to cover all cases but most common ones. Barometer missing for example but there are very easy to do if we agree on some interfaces like that.

src/devices/things/ITemperatureSensor.cs
Ellerbach: Same for meter to imperial system.

src/devices/things/ITemperatureSensor.cs
krwq: @ZhangGaoxing property getter is technically a method (at least on IL level)

krwq:
> This could possible be located in the Common directory and not separate.

Unfortunately if you define it in Common without a project then every library using the source reference will have a separate copy of the interface and even though they are called the same they technically will be different types

@Ellerbach - for `ISensor<T>` - I do believe we need abstraction which provides similar capabilities as your proposal since it allows multiple things like enumerating, auto-discovering devices etc but I'm currently unsure of how it should look like and not convinced proposed is the right design - the issues with `ISensor<T>` I'm seeing:
- If we go with stuff like ITemperature to implement ISensor it will now have to have 2 properties meaning the same thing
- Some devices include more than one sensor, i.e.: pressure and temperature - which one does Value refer to?

Perhaps this is fine but I think we should have some more targeted conversation related to this - can we keep this out of this PR for now?

For now I'd focus on a more stable abstraction and stuff we know need like the fact that thermometer needs `Temperature` property.

Open questions I have:
- ITemperature vs IThermometer vs ISomethingWithVerb - IMO ITemperature makes sense the most sincle IThermometer would suggest that device is a thermometer where the device may be pressure sensor which happens to also have temperature measuring capabilities - I'm open for conversations here
- `ITemperature` vs `ITemperature<T>` - instead of tying to specific output type make it generic? IMO: I like keeping stuff simple so ITemperature does not exclude possibility of adding `ITemperature<T>` later and float gives good enough resolution for all cases I've seen so far
  - if we go with non-generic version do we prefer float or double?

src/devices/things/ITemperatureSensor.cs
krwq: @shaggygi I agree we should have them - let's not do it in this PR though - once we get the design for the low level sensors we can propose those and figure out which library should they be at and what should be the namespace

src/devices/things/ITemperatureSensor.cs
krwq: Let's keep this out of this PR regardless of opinion

src/devices/Dhtxx/DhtSensor.cs
joshfree: Let's keep this as Double for now; as Tanner said offline, all Single values are cleanly representable by Double.

krwq: fair enough - most temperature sensors I've seen were 16bit int scaled and shifted by some constant factor so was basing it on that. I don't feel strongly either way

krwq: also the other reason for float is that Vector3 uses float internally and and Vector3 is what we use for Acceleration and other 3D vectors in this repo

krwq: another alternative is to make ITemperature/IThermometer generic and not depend on any specific type

src/devices/things/ITemperatureSensor.cs
krwq: https://github.com/dotnet/iot/issues/258

src/devices/Dhtxx/DhtSensor.cs
joperezr: I think we should stick with double for now as well and then if we ever see a scenario where float would work better then we can change this.

src/devices/things/things.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <!--Disabling default items so samples source won't get build by the main library-->
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
```
joperezr: Instead of having a separate project for this, I believe this should go under Common instead. So have a src/devices/Common/SensorInterfaces folder where all of these guys live, and sensors that implement them would just add the Compile item to it.

src/devices/Dhtxx/DhtSensor.cs
krwq: fair enough, I honestly wish we'd have Vector3 with double internals 😄

src/devices/things/things.csproj
krwq: when 2 projects use the same interface they cannot exchange it this way since type is in both assemblies and would not unify correctly - do you mean to make it internal for now?

joperezr: That is true as long as we ship different assemblies for different sensors. At least the plan for now, is to only have one IoT.Device.Bindings.dll that has everything includding the interfaces, so they would be able to exchange it.

krwq:
@Ellerbach I think this sounds fine but couple of more questions/thoughts:
- let's say we do ISensor<T>, now we still did not fully enforce conventions since anyone can pick different T and now one thermometer can return double, float or some other decimal and we can't also reason about units the value represent - with ITemperature/IThermometer we can document that Temperature is double/float and document it that unit is Celsius - also allows us to add extension methods to convert to other units - if someone picks different unit this suggest a bug since interface is documented where with ISensor<T> this is arguable what the unit should be and the extension method doesn't work really well
- ITemperatureHumidity vs `ITemperature, IHumidity` - if we pick ISensor for some stuff and some other custom interface for other things we don't have much benefit of adding abstraction. ISensor on its own makes sense to me but that assumes that each device has single concern - perhaps there is a way to solve these both issues by i.e. creating:

```csharp
interface ISensorCollection
{
  IReadOnlyList<ISensor> Sensors;
}

internal struct TemperatureSensorABS : ITemperature, ISensor<double> // similar for humidity
{
  private TemperatureAndHumiditySensorABC _parent;
  public TemperatureSensorABS(TemperatureAndHumiditySensorABC parent)
  {
    _parent = parent;
  }

  public double Temperature => _parent.Temperature;
  public double Value => Temperature;
}

public class TemperatureAndHumiditySensorABC : ITemperature, IHumidity, ISensorCollection
{
  public double Temperature => 123; // replace with real logic
  public double Humidity => 333;
  public IReadOnlyList<ISensor> => new ISensor[] { new  TemperatureSensorABS(this), ... };
}
```

I think we will eventually need abstraction such as ISensor/ISensor<T> because it allows a lot of cool things which simple stuff does not but I think we currently don't have simple stuff so more advanced/powerful stuff should IMO come later - we should have conversation about ISensor separately since we should also have to cover subjects like unit representation, naming conventions (or perhaps name should be represented by some kind of class which contains vendor name etc.), ids, think of some more higher level scenarios and how this can be used and how it combines with other stuff - perhaps we can also figure out how to make it so that you don't have to implement ISensor to use it (i.e. you can have extension method on the ITemperature interface which creates adapter automatically - this way you can hide details and have goodies of both worlds).

I'm not opposing to that idea - I simply think it was not what I'm trying to solve here and that needs a bit more design and thought

Ellerbach:
> I'm not opposing to that idea - I simply think it was not what I'm trying to solve here and that needs a bit more design and thought

Got it. So for the units, I guess it will be hard to fully standardize. That said, most sensors have some précisions and float should be all the time sufficient. 
And then it's about recommendations like temperature should be float, the main values have always to be in international system (metric, °C).

src/devices/things/ITemperatureSensor.cs
arkadiuszwojcik: Indeed property getter is a method but properties syntax suggests operations that are very fast, basically returning already stored variable. If such call is intended to call sensor it should be explicit method: GetTemperature suggesting a bit more complex operation.

joperezr: mmm I'm not sure that is necessarily a true every time. While most of the time this is true because properties would only provide the value of a private member, I do believe we have some examples in the framework where properties have a bit more logic behind them. 

That said, I do have as feedback that we should be explicit with the units that these properties return, so this should be something like: `TemperatureInCelcius`. That way, bindings that consume types that implement this interface would know exactly what to expect out of the values provided by these properties.

Ellerbach: It should be explicit for what is not international standard. Acutally Temperature is in Celcius (in theory, it should even been in Kelving as it's the real international one :-)) 
Where it's more problematic, it's distance sensors like ultrasound. In the international metric system, the unit is the meter. So if there is a default value, it should be in meter. Now, most of those sensors in Arduino and Python are usually returning centimeters. So in this case, yes, it would make sense to have Something like DistanceInCentimeter because it's not meter. And it's usually centimeter because it's the default unit and does not require a float.
but if the "Distance" returns meter, then no need to add anything.
Other option I proposed up is to have Something called "UnitToMeter" which is then a float/double to give the conversion rate to the main international unit. So in case of "Distance" returning centimeter, the "UnitToMeter" would then be 0.01.

src/devices/things/things.csproj
krwq: @joperezr that is true for the package but not true for development where there are still multiple assemblies

krwq:
@Ellerbach I think we will go with double: as @tannergooding mentioned every float is directly representable by double where oppose is not true - the only issue is the Vector3 type which has floats and is kinda a pain to have double everywhere and float in the Vector3

src/devices/Dhtxx/DhtSensor.cs
tannergooding: > fair enough, I honestly wish we'd have Vector3 with double internals

That would be https://github.com/dotnet/corefx/issues/25334, which is ready for review but will likely be post 3.0

krwq:
Is the proposal reasonable here, what are the conclusions?
- double vs float: double seems to be preferred option? (IMO float should be enough, also @Ellerbach who added the sensor I changed above to return float also mentioned above that float should be fine) - @tannergooding what is the difference in perf with using float vs double on arm32? If it is not much then more precision sounds better to me
- ITemperature vs IThermometer - @joperezr any thoughts? IMO ITemperature as it is easier to find what you're looking for

Let's vote
- 👍 for float
- 👎 for double
- ❤️ for ITemperature
- 🚀 for IThermometer

tannergooding:
> * what is the difference in perf with using float vs double on arm32? If it is not much then more precision sounds better to me

It likely varies from device to device. `double` takes twice the space and so on modern computers, when vectorizing, you can generally do twice as many `float` operations per cycle as you can `double`. The throughput for scalar operations is generally limited by memory throughput and so `float` is some measure faster there as well.

However, this is for IOT devices and communicating with the device, over the pins, is likely going to be significantly more expensive than converting the device output to double and converting a double back to the device input (whatever that is).

If the user, after getting the input, wants to work with it as `float` (for whatever reason), the conversion of `float<->double` is also non-expensive.

tannergooding:
> * ❤️ for ITemperature
* 🚀 for IThermometer

I'm not sure either of these are "accurate" names. I would expect maybe `ITemperatureSensor`, which is what I have typically seen these branded as.

krwq:
@tannergooding added the option for that, will vote myself since ITemperature and ITemperatureSensor both seem fine to me.

src/devices/things/ITemperatureSensor.cs
krwq: @arkadiuszwojcik this may or may not be fast - depends on the implementation - how fast is fast? In many of the sensors you can do hundreds of reads per second. Just `Temperature` is much shorter and more intuitive

krwq:
Seems like it concludes to use ITemperatureSensor and double as a backing type.
I've updated all temperature sensors to the new API. Please let me know if project name `things` is fine.
Note: the name won't be used anywhere - our package currently does not use any projects and produces just one dll for all devices

src/devices/things/ITemperatureSensor.cs
tarekgh: I am wondering if we should make this more Generic interface for the sensors report data (e.g. humidity, odometers...etc.) something like 


public interface ISensorData<T>
{
    T Data { get; }
}

krwq: This is what @Ellerbach proposed as well. I generally like this idea but I think there are some issues with that.

See my response here:
- https://github.com/dotnet/iot/pull/256#issuecomment-465308671
- https://github.com/dotnet/iot/pull/256#issuecomment-465462304

the main thing being that some sensors have multiple embedded sensors (i.e. humidity & temeperature)

JeremyKuhne: > the main thing being that some sensors have multiple embedded sensors (i.e. humidity & temeperature)

Yeah, that would be a problem. :)

src/devices/things/ITemperatureSensor.cs
```diff
@@ -0,0 +1,14 @@
...
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device
+{
+    /// <summary>
+    /// Abstraction for temperature sensor
+    /// </summary>
+    public interface ITemperatureSensor
+    {
+        double Temperature { get; }
```
JeremyKuhne: So what is `double`? This needs to be explained thoroughly. Is there an error state for a return value?

krwq: We're thinking Celsius since this will be the most intuitive for most of the people in the world perhaps except US, other alternative is Kelvin as International System of Units suggest - but that seems that it will be inconvenient for everyone except for physicist.
I'll add summary saying Celsius. (Fahrenheit/Kelvin can be an extension method)

For error - couple of options:
- NaN
- double.Max/double.Min
- InvalidOperationException

JeremyKuhne:
Can you please describe where this is useful? Where do we need to pass around an abstraction of the temperature sensor?

src/devices/things/ITemperatureSensor.cs
tannergooding: Error should be `NaN` or an exception.

Ellerbach: I would go for Celsius as it's really what is used everywhere (except very few countries). 
For error, double.Max is a good way to check esilly in the code. Having try/catch is not the very best way to manage thos kind of readings.
Other option is to have a "ValidData" property. So you can do
```csharp
if (tempClass.IsValid)
{
    Console.WriteLine(tempClass.Temerature)
}
```

krwq:
@JeremyKuhne this is for enforcing naming conventions and return types across the repo, IAdc abstraction will probably be the most useful of the all since that once will be able to simplify couple of things which take ADC output as an input

src/devices/things/ITemperatureSensor.cs
tannergooding: > For error, double.Max is a good way to check esilly in the code.

Using any finite value as the error case is bound to lead to confusion or errors, especially if there exists a sensor that legitimately returns that value.

Using any infinite value as the error case is slightly better, but can also lead to failures in the case of overflow or underflow.

`NaN` is an explicit value that, outside of very specific domains (such as complex numbers), generally represents an error value and should be the value used in the case of error (if we are not throwing an exception). As from the IEEE 754 specification:
> Quiet NaNs should, by means left to the implementer’s discretion, afford retrospective diagnostic information inherited from invalid or unavailable data and results.
> ...
> Under default exception handling, any operation signaling an invalid operation exception and for which a floating-point result is to be delivered shall deliver a quiet NaN.

The runtime does not support the IEEE exception handling behavior and explicitly disables it, so the effective behavior is that the floating-point result deliver should return `double.NaN`

JeremyKuhne:
> this is for enforcing naming conventions and return types across the repo

We might get some push back from design review if there isn't a composition or testability case for this. I don't know of any other case where we use interfaces to enforce naming consistency. We have to have a good argument as to why documentation doesn't suffice. Could wiki pages or markdown documents describing conventions be sufficient? If not, why? Interfaces aren't free- it is a linear lookup when dispatching if I recall correctly. We need to be careful to not put ourselves on a track where we start implementing several (or more) interfaces on everything.

krwq:
After some internal conversations replacing this PR with stub for the document with conventions https://github.com/dotnet/iot/pull/284 

Some reasoning:
- interfaces must be useful - currently we think they will be useful - this is not enough
- too many interfaces are confusing for users - how many interfaces will we have in the future? We don't want to end up with 200 interfaces

PR #259, opened 2019-02-20
Adding System.Device.Gpio.Native to the package

joperezr:
cc: @safern @ericstj @krwq @buyaa-n

Adding the native shim into the System.Device.Gpio package. I had to separate the build of the libraries from the build of the packages, so that I could aggregate all of the output (managed and native) and then gather everything in one single package.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -49,8 +49,8 @@ phases:
     - task: PublishBuildArtifacts@1
       displayName: Publish Windows managed assets
       inputs:
-        pathToPublish: $(Build.SourcesDirectory)/artifacts
-        artifactName: artifacts
+        pathToPublish: $(Build.SourcesDirectory)/artifacts/bin
+        artifactName: bin
         artifactType: container
       condition: eq(variables['_BuildConfig'], 'Release')
     variables:
@@ -170,8 +170,8 @@ phases:
     - task: DownloadBuildArtifacts@0
       displayName: Download Built Managed Assets
       inputs:
-        artifactName: artifacts
-        downloadPath: $(Build.SourcesDirectory)
+        artifactName: bin
+        downloadPath: $(Build.SourcesDirectory)/artifacts
     - task: DownloadBuildArtifacts@0
       displayName: Download Built Native Assets
       inputs:
@@ -188,13 +188,14 @@ phases:
       name: Build
       displayName: Build
       condition: succeeded()
-    - task: PublishBuildArtifacts@1
-      displayName: Publish Windows package assets
-      inputs:
-        pathToPublish: $(Build.SourcesDirectory)/artifacts/packages/$(_BuildConfig)/Shipping
-        artifactName: BuiltPackageOutputs
-        artifactType: container
-      condition: eq(variables['_BuildConfig'], 'Release')
+    - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+      - task: PublishBuildArtifacts@1
+        displayName: Publish Windows package assets
+        inputs:
+          pathToPublish: $(Build.SourcesDirectory)/artifacts/packages/$(_BuildConfig)/Shipping
+          artifactName: BuiltPackageOutputs
+          artifactType: container
+        condition: eq(variables['_BuildConfig'], 'Release')
     variables:
       _HelixBuildConfig: $(_BuildConfig)
       ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -40,7 +40,7 @@ phases:
     ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
       enableMicrobuild: true
     steps:
-    - script: build.cmd -ci -sign
+    - script: build.cmd -ci
         -configuration $(_BuildConfig)
         -prepareMachine
       name: Build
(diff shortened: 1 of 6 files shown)


joperezr:
As FYI, this is how the contents of the package look like after my changes:

image

Directory.Build.targets
```diff
@@ -46,6 +46,14 @@
     </ItemGroup>
   </Target>
 
+  <Target Name="_GetNativeAssetsForPackage">
+    <ItemGroup>
+      <TfmSpecificPackageFile Include="$(BaseOutputPath)../Native/Release/*.so">
```
krwq: double checking - if you build this on 64 bit linux, this will work correctly with arm32?

krwq:
assuming this works correctly on ARM32 (Raspberry Pi specifically) LGTM

Directory.Build.targets
joperezr: Yes, we technically build it using a crossrootfs that is targeting arm, which is what makes it work. This is the same way we do it for corefx and coreclr 😄

ericstj: I'm a bit surprised that nothing in the source path is unique for architecture, yet you're putting everything under "linux-arm" RID in the package.

ericstj:
What happens when using this package on OSX?  On desktop? On Xamarin? On linux-x64?  It's not ok to just not provide an assembly / let a PInvoke fail to load a library.  You need to fix these things or have issues tracking the fix.

joperezr:
You are right, I just logged #263 to track the work of generating a turd netstandard2.0 implementation using GenAPI. I flagged that with the 3.0 milestone as it will need to happen before release.

ericstj:
How about the architectures where you aren't providing a native lib?

Directory.Build.targets
joperezr: Right now, we only have one native build which is linux arm. We only plan to have one more specific RID native binary (arm64) so I had planned that this would have to get fixed once that is added.

joperezr:
The native lib only works on arm32 and arm64. The way our internal logic works is that we attempt to PInvoke, and if we succeed we will use those shims, if we can't then we will fallback to use a different implementation:

https://github.com/dotnet/iot/blob/58be7cfeddff7c06db58a8ae11191f141800a6c5/src/System.Device.Gpio/System/Device/Gpio/Drivers/UnixDriver.Linux.cs#L10-L21

Directory.Build.targets
ericstj: I see.  Consider a comment here that explains what's going on and how you intend for it to change.  Both @krwq and myself were confused by it.

.vsts-ci.yml
```diff
@@ -40,20 +40,19 @@ phases:
...
-        inputs:
-          pathToPublish: $(Build.SourcesDirectory)/artifacts/packages/$(_BuildConfig)/Shipping
-          artifactName: BuiltPackageOutputs
-          artifactType: container
-        condition: eq(variables['_BuildConfig'], 'Release')
+    - task: PublishBuildArtifacts@1
+      displayName: Publish Windows managed assets
+      inputs:
+        pathToPublish: $(Build.SourcesDirectory)/artifacts
+        artifactName: artifacts
```
ericstj: should you give this a more specific name?

.vsts-ci.yml
```diff
@@ -130,23 +129,83 @@ phases:
...
+        release_configuration:
+          _BuildConfig: Release
+    ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+      enableMicrobuild: true
+    steps:
+    - task: DownloadBuildArtifacts@0
+      displayName: Download Built Managed Assets
+      inputs:
+        artifactName: artifacts
+        downloadPath: $(Build.SourcesDirectory)
```
ericstj: should this have a artifacts in the path, or is the task appending the artifactName?  (same below)

Directory.Build.targets
joperezr: Will do.

.vsts-ci.yml
joperezr: I wanted it to be the same as the directory, since then I would be able to just unpack it in the same location for my packages build and it would have the same folder structure as the original build. Basically by doing this and then unpacking it on $(Build.SourcesDirectory), I will automatically get a folder called artifacts under my source root so I won't have to rename for my package build to find the right assets.

.vsts-ci.yml
joperezr: yup it is, this will create a folder called artifacts under the sources directory, which is what I want so that it respects the output as it was where the assets where generated.

.vsts-ci.yml
safern: Actually, this should only publish anything under artifacts/bin because you don't need obj, tmp and other directories that get created under artifacts. That will make the artifacts container be fatter and also slow down the build. Then, when you download the assets you can tell the download task to download it into `$(Build.SourcesDirectory)/artifacts/`

ericstj: It seems OK to preserve / restore to the same structure, but I would hope that you could do so without naming the containers the same as the folder, since the folder name alone isn't very descriptive.

safern: The only way to preserve directory structure and renaming the container to something else, is to download the artifacts to another intermediate location like Build.ArtifactsStagingDirectory and then adding an extra step to copy the files over to /bin or whatever destination directory.

joperezr: yeah that's what I thought. @ericstj do you still think that its worth adding the extra step for this? or given that reason is it ok to keep the artifacts name as the destination folder?

Directory.Build.targets
safern: Should you add validation for official builds and CI builds thjat this itemgroup is not empty and error if it is? This to guard your builds and make sure your package always contains this native assets.

PR #264, opened 2019-02-21
Set copyright and package license

tmat:
Prepares repo for change https://github.com/dotnet/arcade/pull/2003 by setting `Copyright` and `PackageLicenseExpression` properties. These values will be required to be set by each repository once https://github.com/dotnet/arcade/pull/2003 is merged.

In order to not break the current builds this change sets the properties conditionally. This condition can be removed once all repos switch to Arcade that has https://github.com/dotnet/arcade/pull/2003.

@markwilkie

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,14 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
+  <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
+
+  <PropertyGroup Condition="'$(CopyrightNetFoundation)' != ''">
+    <Copyright>$(CopyrightNetFoundation)</Copyright>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+  </PropertyGroup>
+  
   <PropertyGroup>
-  	<UsingToolMicrosoftNetCompilers>false</UsingToolMicrosoftNetCompilers>
-  	<PublishWindowsPdb>false</PublishWindowsPdb>
-    <ImportNetSdkFromRepoToolset>false</ImportNetSdkFromRepoToolset>
+    <PublishWindowsPdb>false</PublishWindowsPdb>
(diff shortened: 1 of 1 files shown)


tmat:
`UsingToolMicrosoftNetCompilers` is false by [default](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/tools/DefaultVersions.props#L47). It used to be different, so now it's not necessary to override it anymore.

The [recommendation](https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md#directorybuildprops) is to import Sdk.props first:

> The Sdk.props file sets various properties and item groups to default values. It is recommended to perform any customizations after importing the SDK.

BTW, setting `<Language>C#</Language>` seems unnecessary. The language is determined from the project extension (e.g. `.csproj`).

tmat:
OT: All the logic in `/build.proj` seems unnecessary as well. This is replicating/circumventing similar Arcade build logic. Why not have a solution that contains all the projects and build that solution?

safern:
> BTW, setting `<Language>C#</Language>` seems unnecessary. The language is determined from the project extension (e.g. `.csproj`).

Agreed.

> OT: All the logic in `/build.proj` seems unnecessary as well. This is replicating/circumventing similar Arcade build logic. Why not have a solution that contains all the projects and build that solution?

I think @joperezr wanted control over what targets are run on which projects and he's concerned that this repo is potentially going to grow to a handful amount of projects as they add bindings and drivers, potentially something like corefx, over 500 projects, from last time I spoke to him, that where the insights he told me.

tmat:
@safern @joperezr I'd rather avoid replicating CoreFX build structure unless absolutely necessary and instead I'd strongly suggest to follow what all other repos are doing. We have 180 projects in Roslyn repo and there haven't been any issues. If any issues arise we need to address them in the product and/or in Arcade SDK, not in each repo separately.

PR #268, opened 2019-02-21
Added SHT3x binding

ZhangGaoxing:
Humidity and Temperature Sensor SHT30/SHT31/SHT35


src/devices/Sht3x/I2cAddress.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// SHT3x I2C Address
+    /// </summary>
+    public enum I2cAddress : byte
+    {
+        /// <summary>
+        /// ADDR (pin2) connected to logic low (Default)
+        /// </summary>
+        AddrLow =0x44,
```
krwq: nit: space after `=`

src/devices/Sht3x/Sht3x.cs
```diff
@@ -0,0 +1,169 @@
...
+            if (isOn)
+                Write(Register.SHT_HEATER_ENABLE);
+            else
+                Write(Register.SHT_HEATER_DISABLE);
+        }
+
+        /// <summary>
+        /// Read Temperature and Humidity
+        /// </summary>
+        private void ReadTempAndHumi()
```
krwq: does it make sense to split this into two functions (for reading temperature and humidity separately)? you call for both of them either way so feels like there is no benefit of having them together

src/devices/Sht3x/Sht3x.cs
```diff
@@ -0,0 +1,169 @@
...
+                for (int j = 8; j > 0; j--)
+                {
+                    if ((crc & 0x80) != 0)
+                        crc = (byte)((crc << 1) ^ CRC_POLYNOMIAL);
+                    else
+                        crc = (byte)(crc << 1);
+                }
+            }
+
+            if (crc == crc8)
```
krwq: this can be just `return crc == crc8;`

src/devices/Sht3x/Sht3x.cs
```diff
@@ -0,0 +1,169 @@
...
+            _humidity = Math.Round(srh * 100 / 65535.0, 1);
+        }
+
+        /// <summary>
+        /// 8-bit CRC Checksum Calculation
+        /// </summary>
+        /// <param name="data">Raw Data</param>
+        /// <param name="crc8">Raw CRC8</param>
+        /// <returns>Checksum is true or false</returns>
+        private bool CRC8(ReadOnlySpan<byte> data, byte crc8)
```
krwq: since you take 2 parameters you should call it: `CheckCrc8` or `ValidateCrc8`

src/devices/Sht3x/Sht3x.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            // Details in the Datasheet P13
+            byte crc = CRC_INIT;
+            for (int i = 0; i < 2; i++)
+            {
+                crc ^= data[i];
+
+                for (int j = 8; j > 0; j--)
```
krwq: why do you go backwards instead of forward? I'd let JIT/compiler take care of any optimizations like that if that was the motivation

src/devices/Sht3x/Register.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Sht3x
+{
+    /// <summary>
+    /// SHT3x Register
+    /// </summary>
+    internal enum Register : short
```
krwq: go with the ushort - shifting short right for negative numbers does something else that what might be expected (in your case doesn't make much difference but someone might copy the project and do it by accident)

src/devices/Sht3x/Sht3x.cs
ZhangGaoxing: If I could use two methods, I would use them. This sensor is strange. To read the data, I need to send an address containing four hexadecimal bits, and then the sensor will directly return six bytes of data. In other words, temperature and humidity return together.

src/devices/Sht3x/Register.cs
ZhangGaoxing: 👍Awesome! Actually, I didn't think about that in detail.

src/devices/Sht3x/Sht3x.cs
krwq: haha 😄 won't argue - this is hardware - there are weird things here everywhere!

PR #269, opened 2019-02-22
Adding suppport for TCS3472x

Ellerbach:
Adding support for TCS3472x familly:
* Implemented reading color
* Implemented functions to access all registers
* Documentation


ZhangGaoxing:
I've written some code for TCS34725, and I don't think I need it now.😄

ZhangGaoxing:
left some comments😄

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,215 @@
...
+
+using System;
+using System.Buffers.Binary;
+using System.Device.I2c;
+using System.Drawing;
+using System.Threading;
+
+namespace Iot.Device.Tcs3472x
+{
+    public class Tcs3472xSensor : IDisposable
```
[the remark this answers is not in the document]
Ellerbach: Well, I like to add Sensor when it's a sensor. I guess we should align on naming convention.

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,215 @@
...
+    {
+        public const byte DefaultAddress = 0x29;
```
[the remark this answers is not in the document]
Ellerbach: DefaultI2cAddress will work. At the same time it's the only address and the sensor is I2C. So maybe just Address? 
I got feedback on another PR to just call it Address but I think it's more clear with DefaultAddress and would be even better with DefaultI2cAddress. Let's see if anyone else will comment.

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,215 @@
...
+            enable = state ? enable |= (byte)Registers.ENABLE_AIEN : enable = (byte)(enable & ~(byte)Registers.ENABLE_AIEN);
+            _i2cDevice.Write(new byte[] { (byte)(Registers.COMMAND_BIT | Registers.ENABLE), enable });
+        }
+
+        /// <summary>
+        /// Get the color
+        /// </summary>
+        /// <param name="delay">Wait to read the data that the integration time is passed</param>
+        /// <returns></returns>
+        public Color GetColor(bool delay = true)
```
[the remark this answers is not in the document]
Ellerbach: I can add a property. The Reason it's a function is because you can force the reading withtout waiting (if you waited before for example for sopme reasons) or you just measure the data time to time.

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+
+            if (delay)
+                Thread.Sleep((int)(IntegrationTime * 1000));
```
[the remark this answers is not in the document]
Ellerbach: it doesn't need to be very precise. So Thread.Sleep will work great. Other operations can be done in parallel. Or did I miss Something?

src/devices/Tcs3472x/Registers.cs
```diff
@@ -0,0 +1,64 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.Tcs3472x
+{
+    [Flags]
```
krwq: do you combine those that `[Flags]` is useful? feels kinda unusual for registers :smile:

src/devices/Tcs3472x/TCS3472Type.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Tcs3472x
+{
+    public enum TCS3472Type
+    {
+        TCS34721 = 0x44,
+        TCS34725 = 0x44,
```
krwq: values are doubled (here and below) - is this intentional?

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,227 @@
...
+        /// <summary>
+        /// Set/Get the gain
+        /// </summary>
+        public Gain Gain
+        {
+            get { return _gain; }
+            set
+            {
+                _gain = value;
+                _i2cDevice.Write(new byte[] { (byte)(Registers.COMMAND_BIT | Registers.CONTROL), (byte)_gain });
```
krwq: 1. feels like the command bit should not be in the register and rather some const in this class (the `[Flags]` on the Register is tiny bit confusing)
2. you can use non-allocating pattern here

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,227 @@
...
+            _i2cDevice.WriteByte((byte)(Registers.COMMAND_BIT | Registers.STATUS));
+            var stat = _i2cDevice.ReadByte();
+            return ((Registers)(stat & (byte)Registers.STATUS_AINT) == Registers.STATUS_AINT);
+        }
+
+        /// <summary>
+        /// Set the integration (sampling) time for the sensor
+        /// </summary>
+        /// <param name="timeSeconds">Time in seconds for each sample. 0.0024 second(2.4ms) increments.Clipped to the range of 0.0024 to 0.6144 seconds.</param>
+        public void SetIntegrationTime(double timeSeconds)
```
krwq: since you already got property this should be internal/private

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,227 @@
...
+            SetIntegrationTime(integrationTime);
+            Gain = gain;
+            PowerOn();
+        }
+
+        /// <summary>
+        /// Returns true is there are valid data
+        /// </summary>
+        /// <returns>Returns true is there are valid data</returns>
+        public bool IsValidData()
```
krwq: I'd just make this property since we use that in most of the places for similar things

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,227 @@
...
+            _i2cDevice.WriteByte((byte)(Registers.COMMAND_BIT | Registers.STATUS));
+            var stat = _i2cDevice.ReadByte();
+            return ((Registers)(stat & (byte)Registers.STATUS_AVALID) == Registers.STATUS_AVALID);
+        }
+
+        /// <summary>
+        /// Returns true if RGBC is clear channel interrupt
+        /// </summary>
+        /// <returns>Returns true if RGBC is clear channel interrupt</returns>
+        public bool IsClearInterrupt()
```
krwq: property?

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,227 @@
...
+            if (timeSeconds <= 700)
+            {
+                if (_isLongTime)
+                {
+                    SetConfigLongTime(false);
+                }
+                _isLongTime = false;
+                var timeByte = (int)(0x100 - (timeSeconds / 0.0024));
+                timeByte = Math.Clamp(timeByte, 0, 255);
+                _i2cDevice.Write(new byte[] { (byte)(Registers.COMMAND_BIT | Registers.ATIME), (byte)timeByte });
```
krwq: you can use non-allocating pattern

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            SetInterrupt(InterruptState.All, state);
+        }
+
+        /// <summary>
+        /// Set/clear a specific interrupt persistence
+        /// </summary>
+        /// <param name="interupt">The percistence cycles</param>
+        /// <param name="state">True to set the interrupt, false to clear</param>
+        public void SetInterrupt(InterruptState interupt, bool state)
```
krwq: how should the user use this? do you set interrup flag first and then check if interrup is clear with the other property? perhaps might be worth to add some comment about it - also might be worth renaming to something like:
StartWaitingFor...
Is...Ready

src/devices/Tcs3472x/Tcs3472x.cs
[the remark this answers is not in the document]
krwq: I find the delay a bit confusing from usability point of view - is it better to wait or not? if yes/no what's the benefit disadvantage (please add a comment)

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,227 @@
...
+            a = Math.Clamp(a, 0, 255);
+            return Color.FromArgb(a, r, g, b);
+        }
+
+        /// <summary>
+        /// Get the color
+        /// </summary>
+        public Color Color => GetColor();
+
+        private UInt16 I2cRead16(Registers reg)
```
krwq: UInt16 -> ushort

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,227 @@
...
+            _i2cDevice.WriteByte((byte)(Registers.COMMAND_BIT | reg));
+            return _i2cDevice.ReadByte();
+        }
+
+        public void Dispose()
+        {
+            PowerOff();
+            if (_autoDisposable)
+            {
+                _i2cDevice?.Dispose();
```
krwq: `_i2cDevice = null` missing

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,227 @@
...
+            b = Math.Clamp(b, 0, 255);
+            int a = (int)(I2cRead16(Registers.CDATAL) * 255 / divide);
+            a = Math.Clamp(a, 0, 255);
+            return Color.FromArgb(a, r, g, b);
+        }
+
+        /// <summary>
+        /// Get the color
+        /// </summary>
+        public Color Color => GetColor();
```
krwq: GetColor should be internal/private

src/devices/Tcs3472x/Registers.cs
Ellerbach: Correct, I used to but changed the architecture. Forgot to remove the [Flags] :-)

src/devices/Tcs3472x/TCS3472Type.cs
Ellerbach: Yes, it is intentional. Some sensors have the same signature

src/devices/Tcs3472x/Tcs3472x.cs
Ellerbach: Yes, I wrote a function for it.

src/devices/Tcs3472x/Gain.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Tcs3472x
+{
+    /// <summary>
+    /// The gain used to integrate the colors
+    /// </summary>
+    public enum Gain
```
krwq: since this is public please rename values to i.e. Gain01X (make sure it sorts from lowest to highest and apply casing rules)

src/devices/Tcs3472x/InterruptState.cs
```diff
@@ -0,0 +1,42 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Tcs3472x
+{
+    public enum InterruptState
+    {
+        // Every RGBC cycle generates an interrupt
+        All = 0b0000,
+        // 1 clear channel value outside of threshold range
```
krwq: I think one larger comment explaining what this means would be useful (no need for similar comment per value)

src/devices/Tcs3472x/Registers.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.Tcs3472x
+{
+    public enum Registers
```
krwq: internal

src/devices/Tcs3472x/Tcs3472x.cs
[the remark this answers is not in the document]
krwq: Currently we don't add suffixes anywhere - almost everything under Iot.Device is assumed to be a Sensor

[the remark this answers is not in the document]
Ellerbach: I'll fix this.

src/devices/Tcs3472x/Tcs3472x.cs
```diff
@@ -0,0 +1,236 @@
...
+        /// <summary>
+        /// Create a TCS4272x sensor
+        /// </summary>
+        /// <param name="i2cDevice">The I2C Device class</param>
+        /// <param name="integrationTime">The time to wait for sensor to read the data, minimum is 0.024 seconds, maximum in the constructor is 0.7 seconds</param>
+        /// <param name="gain">The gain when integrating the color measurement</param>
+        /// <param name="autoDisposable">true to dispose the I2C Device class at dispose</param>
+        public Tcs3472x(I2cDevice i2cDevice, double integrationTime = 0.0024, Gain gain = Gain.Gain16X, bool autoDisposable = true)
+        {
+            _i2cDevice = i2cDevice ?? throw new ArgumentException($"{nameof(i2cDevice)} can't be null");
```
krwq: `new ArgumentException(...)` => `new ArgumentNullException(nameof(i2cDevice))`

src/devices/Tcs3472x/Tcs3472x.cs
Ellerbach: It's about the number of cycle you wait to generate an interruption. So it's more to use instead of the previous one

src/devices/Tcs3472x/Tcs3472x.cs
Ellerbach: I leave it public if you want to read the colors even if integration cycle is not over. So read the previous value for example

src/devices/Tcs3472x/samples/Tcs3472x.sample..cs
```diff
@@ -24,7 +24,7 @@ static void Main(string[] args)
                     Console.WriteLine($"ID: {tcs3472X.ChipId} Gain: {tcs3472X.Gain} Time to wait: {tcs3472X.IsClearInterrupt}");
                     var col = tcs3472X.GetColor();
                     Console.WriteLine($"R: {col.R} G: {col.G} B: {col.B} A: {col.A} Color: {col.Name}");
-                    Console.WriteLine($"Valid data: {tcs3472X.IsValidData} Clear Interrupt: {tcs3472X.IsValidData}");
+                    Console.WriteLine($"Valid data: {tcs3472X.IsValidData} Clear Interrupt: {tcs3472X.IsClearInterrupt}");
```
krwq: there is double dot in the file name (`..cs`)

src/devices/Tcs3472x/Tcs3472x.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <LangVersion>latest</LangVersion>
```
krwq: please remove `latest` that should be the default - I have removed that recently in all projects

src/devices/Tcs3472x/samples/Tcs3472x.sample.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease.19078.2" />
```
krwq: please use `0.1.0-prerelease*`

src/devices/Tcs3472x/Tcs3472x.csproj
```diff
@@ -0,0 +1,15 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease.19078.2" />
```
krwq: `0.1.0-prerelease*`

krwq:
LGTM except for the few comments I've just put (only blocking thing is the `..` in the file name)

krwq:
Thanks @Ellerbach!!!

PR #270, opened 2019-02-22
Adding support for BME280

Ellerbach:
Adding support for BME280 sensor:
* Refactored the existing BMP280 sensor
* Removed the Bmp280 directory, created Bmx280 directory
* Added sample for BME280 sensor
* Added default I2C Address part of the classes
* Updated the documentation


ZhangGaoxing:
Question: Should we increase our support for BMP180 and BMP085? How to extend it if it is designed like this?

src/devices/Bmx280/Bme280.cs
```diff
@@ -0,0 +1,22 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Device.I2c;
+
+namespace Iot.Device.Bmx280
+{
+    public class Bme280 : BmxBase
+    {
+        private const byte Signature = 0x60;
+        public const byte Address = 0x76;
```
ZhangGaoxing: I think it should be called **I2cAddress**

src/devices/Bmx280/BmxBase.cs
```diff
@@ -53,6 +46,7 @@ private void Begin()
             _calibrationData.ReadFromDevice(this);
         }
 
+
         /// <summary>
         /// Sets the power mode to the given mode
         /// </summary>
```
ZhangGaoxing: I think we should use property to read data and set settings, and change Task.Delay() to DelayHelper.DelayMilliseconds().
https://github.com/dotnet/iot/blob/3c55f83a2fd60557d1373ca126abcacfaaeea9de/src/devices/Common/System/Device/DelayHelper.cs#L13

Ellerbach:
> Question: Should we increase our support for BMP180 and BMP085? How to extend it if it is designed like this?

Do you have pointers on the documentation? I can check but I won't be able to test the sensors

src/devices/Bmx280/Bme280.cs
Ellerbach: we should agree on naming. What about DefaultI2cAddress for all those cases? At least it's very explicit.

src/devices/Bmx280/BmxBase.cs
```diff
@@ -6,44 +6,37 @@
 //Formulas and code examples can also be found in the datasheet http://www.adafruit.com/datasheets/BST-BMP280-DS001-11.pdf
 
 using System;
+using System.Buffers.Binary;
 using System.Device.I2c;
 using System.Threading.Tasks;
-using System.Buffers.Binary;
 
-namespace Iot.Device.Bmp280
+namespace Iot.Device.Bmx280
```
krwq: I'm wondering if X should be capital - otherwise Bmx sounds like it is part of the name 😄 (no strong opinion either way though)

Ellerbach: Everywhere else we are using lower case x to replace characters. So I went for the lower case one

krwq: fair enough

joperezr:
@garciaolais was working on the binding for BMP180, so one thing we could do is to get this in as is and then @garciaolais can adapt his PR to use this abstraction to add support for that as well.

src/devices/Bmx280/Bme280.cs
joperezr: I like @Ellerbach proposal, `DefaultI2cAddress` makes sense to me

src/devices/Bmx280/Bme280.cs
```diff
@@ -0,0 +1,22 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Device.I2c;
+
+namespace Iot.Device.Bmx280
+{
+    public class Bme280 : BmxBase
+    {
+        private const byte Signature = 0x60;
```
joperezr: Similar to the Address comment, should this be something more explicit like `DeviceID` or somehting like that?

src/devices/Bmx280/Bme280.cs
```diff
@@ -0,0 +1,22 @@
...
+        public const byte Address = 0x76;
+
+        public Bme280(I2cDevice i2cDevice)
```
joperezr: Do we plan to add SPI support for these devices in this PR? Or do we want to do that in a future PR instead?

src/devices/Bmx280/BmxBase.cs
joperezr: Interesting point though, We can log an issue and change this to capital X in the places where we do this as I agree with @krwq that it would be more clear that we are talking about a base class instead of a individual binding.

src/devices/Bmx280/BmxBase.cs
```diff
@@ -6,44 +6,37 @@
...
 {
-    public class Bmp280 : IDisposable
+    public class BmxBase : IDisposable
     {
-        private const byte Signature = 0x58;
+        internal I2cDevice _i2cDevice;
+        internal byte _signature;
```
joperezr: readonly?

src/devices/Bmx280/BmxBase.cs
```diff
@@ -6,44 +6,37 @@
...
+        internal bool _initialized = false;
+        internal CommunicationProtocol _communicationProtocol;
```
joperezr: readonly as well?

src/devices/Bmx280/BmxBase.cs
```diff
@@ -6,44 +6,37 @@
...
-        private I2cDevice _i2cDevice;
-        private readonly CommunicationProtocol _communicationProtocol;
-        private bool _initialized = false;
-        private readonly CalibrationData _calibrationData;
         /// <summary>
         /// The variable _temperatureFine carries a fine resolution temperature value over to the
         /// pressure compensation formula and could be implemented as a global variable.
         /// </summary>
-        private int _temperatureFine;
+        internal int _temperatureFine;
```
joperezr: Why internal? I would have them be protected if needed by the subclasses, but if not I would keep as private.

src/devices/Bmx280/BmxBase.cs
```diff
@@ -6,44 +6,37 @@
...
-        private void Begin()
+        internal void Begin()
         {
             _i2cDevice.WriteByte((byte)Register.CHIPID);
             byte readSignature = _i2cDevice.ReadByte();
 
-            if (readSignature != Signature)
+            if (readSignature != _signature)
             {
                 return;
```
joperezr: Should this throw an exception? I know you didn't write this part, but I guess that it makes a bit more sense to let consumer know that the connection did not work as expected instead of failing silently.

src/devices/Bmx280/BmxBase.cs
```diff
@@ -88,7 +81,6 @@ public PowerMode ReadPowerMode()
                 return PowerMode.Forced;
             }
         }
-
```
joperezr: NIT: We shouldn't remove this line, as we do want to have one empty line separating each method.

src/devices/Bmx280/BmxBase.cs
```diff
@@ -332,7 +324,6 @@ private long CompensatePressure(int adcPressure)
             p = ((p + var1 + var2) >> 8) + ((long)_calibrationData.DigP7 << 4);
             return p;
         }
-
```
joperezr: Same comment about empty line here.

src/devices/Bmx280/BmxBase.cs
```diff
@@ -342,7 +333,7 @@ private long CompensatePressure(int adcPressure)
         /// <returns>
         ///  Value from register
         /// </returns>
-        private byte Read8BitsFromRegister(byte register)
+        internal byte Read8BitsFromRegister(byte register)
```
joperezr: Same comment about making internal. I don't see why you would change these so they should all remain private.

src/devices/Bmx280/README.md
```diff
@@ -0,0 +1,33 @@
+# BMx280 Sensors
+
+## Summary
+
+BMx280 is a device family that read barometric pressure, altitude and temperature. SPI and I2C can be used to communicate with the device (only I2C implemented so far).
```
joperezr: Can we add to this readme a checkbox of supported protocols that show that the device supports both I2c and SPI where we only check the i2c box as that is the only one we have for now? Here is an example of where we already do that: [https://github.com/dotnet/iot/blob/e9529392e86b8e0e31e96ba1a5f83110ecf55639/src/devices/Ssd1306/README.md#connection-type](https://github.com/dotnet/iot/blob/e9529392e86b8e0e31e96ba1a5f83110ecf55639/src/devices/Ssd1306/README.md#connection-type)

src/devices/Bmx280/samples/Bme280.sample.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <EnableDefaultItems>false</EnableDefaultItems>
```
joperezr: NIT: we don't need this property on sample projects as we don't mind in this case if bindings include *.cs to the Compile item.

joperezr:
Left a few comments, but this looks good otherwise.

garciaolais:
> 
> 
> @garciaolais was working on the binding for BMP180, so one thing we could do is to get this in as is and then @garciaolais can adapt his PR to use this abstraction to add support for that as well.

I plan to send the PR this weekend.

src/devices/Bmx280/Bme280.cs
krwq: i think we should add it once someone requests it - also I expect we will eventually have some common interface between those 2

src/devices/Bmx280/BmxBase.cs
krwq: yes, the previous case happened to end with number just before the `x` so it didn't matter. fine to keep as is for this PR - let's observe this closer once we get more devices

src/devices/Bmx280/Bme280.cs
garciaolais: What about `DefaultI2cAddressPrim`  and `DefaultI2cAddressSec`?

The Bme280 and Bmp280 have two address, dependes of SDO pin conection
[BMP280](https://cdn-shop.adafruit.com/datasheets/BST-BMP280-DS001-11.pdf
) Page 28
[BME280](https://cdn-shop.adafruit.com/datasheets/BST-BME280_DS001-10.pdf) Page 31

SDO to GND    -> 0x76
SDO to VDDIO -> 0x77

ZhangGaoxing: use enum type
```C#
public enum I2cAddress : byte
{

}
```

garciaolais:
Question ❔  : Do we plan to add 💦 `GetHumidity()`  for bme device?

https://github.com/Ellerbach/iot/blob/9c685715e6bd624f4a5dd2112ccd0965703c80f7/src/devices/Bmx280/Bme280.cs#L9

src/devices/Bmx280/Bme280.cs
krwq: I'd go with the just Default - if user already looks in the datasheet to set physical pins they can as well check the address

src/devices/Bmx280/BmxBase.cs
krwq: if going with protected then use naming convention as with public (`TemperatureFine`)

src/devices/Bmx280/BmxBase.cs
```diff
@@ -391,7 +382,7 @@ internal ushort Read16BitsFromRegister(byte register)
         /// <returns>
         ///  Value from register
         /// </returns>
-        private uint Read24BitsFromRegister(byte register)
+        internal uint Read24BitsFromRegister(byte register)
```
krwq: we should add some helpers for 24-bit numbers - I've seen this like 3 times already at least and every one does it differently

src/devices/Bmx280/Bme280.cs
Ellerbach: > I'd go with the just Default - if user already looks in the datasheet to set physical pins they can as well check the address

Fair. I'll remove the second one. It's anyway a specific setup

src/devices/Bmx280/Bme280.cs
Ellerbach: Yes, I can rename this to ```DeviceId```

src/devices/Bmx280/Bme280.cs
Ellerbach: We Don't have this sensor to test in SPI mode. So yes, I'll leave it up to the point we either find one to test, either someone demands it. Looks like most popular implémentations as using I2C.

src/devices/Bmx280/BmxBase.cs
Ellerbach: if I do understand correctly, readonly won't work as it won't be assigned in the base controller but in the child.

src/devices/Bmx280/BmxBase.cs
Ellerbach: same as above.

src/devices/Bmx280/BmxBase.cs
Ellerbach: I'll go for protected then. Can be useful for other sensors from the same familly. Won't change much for this PR.

src/devices/Bmx280/BmxBase.cs
Ellerbach: I was hesitating to change it. Now I Don't anymore :-)

src/devices/Bmx280/BmxBase.cs
Ellerbach: yes, will switch back to private.

src/devices/Bmx280/samples/Bme280.sample.csproj
Ellerbach: I have 2 projects in the same directory, so I guess it's needed. And ```async main``` is used so need one of the latest version of language.

krwq: for latest compiler  you can go ahead and use that everywhere

PR #271, opened 2019-02-22
Added LM75 binding

ZhangGaoxing:
Digital Temperature Sensor LM75


src/devices/Lm75/Lm75.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Device.I2c;
+
+namespace Iot.Device.Lm75
+{
+    /// <summary>
+    /// Digital Temperature Sensor LM75
+    /// </summary>
+    public class Lm75 : IDisposable
+    {
+        I2cDevice _sensor;
```
krwq: explicit private

src/devices/Lm75/samples/Program.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Iot.Device.Lm75.Samples
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            I2cConnectionSettings settings = new I2cConnectionSettings(1, Lm75.I2cAddress);
+            UnixI2cDevice device = new UnixI2cDevice(settings);
+
+            using(Lm75 sensor=new Lm75(device))
```
krwq: spaces around `=`

src/devices/Lm75/samples/Program.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
```
krwq: license header

src/devices/Lm75/Lm75.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// <summary>
+        /// LM75 Temperature (℃)
+        /// </summary>
+        public double Temperature { get => GetTemperature(); }
+
+        private bool _shutdown;
+        /// <summary>
+        /// LM75 Shutdown
+        /// </summary>
+        public bool Shutdown { get { return _shutdown; } set { SetShutdown(value); _shutdown = value; } }
```
krwq: I would go with some word with past tense i.e. Disabled. Shutdown sounds like a function name

ZhangGaoxing: Aha, the datasheet says shutdown. I will change the name.😄

src/devices/Lm75/Lm75.cs
```diff
@@ -0,0 +1,103 @@
...
+            double temp = 0;
+            ushort raw = (ushort)((readBuff[0] << 3) | (readBuff[1] >> 5));
+            if ((readBuff[0] & 0x80) == 0)
+            {
+                temp = raw * 0.125;
+            }
+            else
+            {
+                // two's complement
+                raw |= 0xF800;
```
krwq: isn't two's complement be equivalent to using BinaryPrimitives.ReadLittle/BigEndian?

src/devices/Lm75/Register.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Lm75
+{
+    public enum Register : byte
```
krwq: this should be private

src/devices/Lm75/Lm75.cs
ZhangGaoxing: Aha, I don't quite understand what you mean. `using BinaryPrimitives.ReadInt16Little/BigEndian` need to pass in a Span<T>. This is a binary complement for negative numbers. Temperature register data contains undefined bits, and BinaryPrimitives.ReadInt16Little/BigEndian cannot be used directly😄. @krwq

krwq: @ZhangGaoxing I'm fine with this as is for now but I suspect if you just cast raw to short you might get same result - not sure what the `raw |= 0xF800` does though - I'd have to dig more to check it so fine to leave as is if you get correct results

src/devices/Lm75/Lm75.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                // temperature < 0
+                // two's complement
+                raw |= 0xF800;
+                raw = (ushort)(~raw + 1);
+
+                temp = raw * (-1) * 0.125;
+            }
+
+            return Math.Round(temp, 1);
```
krwq: why are you rounding?

src/devices/Lm75/Lm75.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <param name="isShutdown">Shutdown when value is true.</param>
+        private void SetShutdown(bool isShutdown)
+        {
+            _sensor.WriteByte((byte)Register.LM_CONFIG);
+            byte config = _sensor.ReadByte();
+
+            config &= 0xFE;
+            if (isShutdown)
+                config |= 0x01;
+            else
```
krwq: else can be removed completely `config |= 0x00` is a nop, isn't it?

src/devices/Lm75/Lm75.cs
ZhangGaoxing: The temperature of this sensor is accurate to 0.1

src/devices/Lm75/Lm75.cs
ZhangGaoxing: Yes, I just want to keep the code readable. I'll keep this operation as a comment.

src/devices/Lm75/Lm75.cs
ZhangGaoxing: `0xF800` just sets the high five unused bits

src/devices/Lm75/Lm75.cs
```diff
@@ -89,8 +89,11 @@ private void SetShutdown(bool isShutdown)
             config &= 0xFE;
             if (isShutdown)
                 config |= 0x01;
-            else
-                config |= 0x00;
+            //else
```
krwq: remove the commented out code 😄

PR #277, opened 2019-02-24
Added MAX44009 binding

ZhangGaoxing:
Ambient Light Sensor MAX44009


src/devices/Max44009/IntegrationTime.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    /// <summary>
+    /// Measurement Cycle
+    /// </summary>
+    public enum IntegrationTime
+    {
+        /// <summary>
+        /// 800ms
+        /// </summary>
+        TIM_800 = 0b000,
```
krwq: could we perhaps use property which just translates this or interpolates this into a float? I think it would be so much easier to use

krwq: cc: @tannergooding any ideas for this?

ZhangGaoxing: The different working modes of sensors will lead to the validity of IntegrationTime setting. The IntegrationTime setting is invalid in default mode, so I used two constructors instead of attributes to set IntegrationTime.😄

tannergooding: I think a "getter" property that returns this as a `TimeSpan` might be useful. But, the datasheet shows that the IntegrationTime can only be one of these 8 explicit values, so I think an enum probably makes sense (not a fan of the names, however).

krwq: Thanks @tannergooding, I agree the name should be at least something like Time800... The underscore for `_25` and `_5` sounds fine

Issue #111, opened 2018-12-21
WriteRead using I2C and MAX7311


bigdnf:
I’m trying to communicate with MAX7311 using UnixI2cDevice on Raspians. Generally I2C with other devices works but in this case I need to write some data and then immediately read response from MAX7311 device. 
When I try to do something like this:

```
var buffer = new Span<byte>();
var device = new System.Device.I2c.Drivers.UnixI2cDevice(new I2cConnectionSettings(BusId, address));
device.Write(new byte[] { 0x00 });
device.Read(buffer);

```
I get exception “Error when attempting to perform the I2c data transfer”
When I use some custom implementation of I2C from this repo

https://github.com/chkr1011/Wirehome.Core/blob/master/Wirehome.Core/Hardware/I2C/I2CBusService.cs

and use WriteRead method I can communicate with success. 

 ```
var device = new LinuxI2CBusAdapter(Logger);
 device.Enable();
 var readBuffer = new byte[2];
 device.WriteRead(address, new byte[] { 0x00 }, readBuffer);
```

It would be nice if this WriteRead could be supported – or I’m doing something wrong?

**Comment by shaggygi** [comment]
I didn't look into this much, but you might need to set the buffer size to 1.  Not completely sure.  Also, make sure the BusId is correct.  I believe it should be 1.

```csharp
var device = new System.Device.I2c.Drivers.UnixI2cDevice(new I2cConnectionSettings(BusId, address));
device.Write(new byte[] { 0x00 });
byte[] readBuffer = new byte[1];
device.Read(buffer);
```

joperezr:
Because I2C protocol only uses 1 data pin, there is not a method like the TransferFullDuplex that we have in SPI. Internally what we are doing is to just do ioctl on the I2c device and send the information, so I would expect that as long as you set the right size of the read and write buffers that it should work correctly. Did you try what @shaggygi suggested?

shaggygi:
I tested this (only on Unix) and can repro this when sending Write or Read commands with an empty array or Span as shown in @bigdnf example.

@joperezr I don't understand why the ioctl wouldn't take 0 bytes, but I guess it is expecting something when doing a write/read function.  To fix this, we could put a check in `ReadWriteInterfaceTransfer` and `FileInterfaceTransfer` to just return if `writeBufferLength` or `readBufferLength` are both 0.

[...]

PR #278, opened 2019-02-24
I2C/SPI Read should not allow empty reads

krwq:
Fixes https://github.com/dotnet/iot/issues/111

Making experience better when empty buffer is accidentally passed to Read

### DIFF
--- a/src/System.Device.Gpio/System/Device/I2c/Drivers/UnixI2cDevice.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/I2c/Drivers/UnixI2cDevice.Linux.cs
@@ -174,6 +174,9 @@ namespace System.Device.I2c.Drivers
         /// </param>
         public override unsafe void Read(Span<byte> buffer)
         {
+            if (buffer.Length == 0)
+                throw new ArgumentException($"{nameof(buffer)} cannot be empty.");
+
             Initialize();
 
             fixed (byte* bufferPointer = buffer)
--- a/src/System.Device.Gpio/System/Device/I2c/Drivers/Windows10I2cDevice.Windows.cs
+++ b/src/System.Device.Gpio/System/Device/I2c/Drivers/Windows10I2cDevice.Windows.cs
@@ -65,6 +65,9 @@ namespace System.Device.I2c.Drivers
         /// </param>
         public override void Read(Span<byte> buffer)
         {
+            if (buffer.Length == 0)
+                throw new ArgumentException($"{nameof(buffer)} cannot be empty.");
+
             byte[] byteArray = new byte[buffer.Length];
             _winI2cDevice.Read(byteArray);
             new Span<byte>(byteArray).CopyTo(buffer);
--- a/src/System.Device.Gpio/System/Device/Spi/Drivers/UnixSpiDevice.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Spi/Drivers/UnixSpiDevice.Linux.cs
@@ -129,6 +129,9 @@ namespace System.Device.Spi.Drivers
         /// </param>
         public override unsafe void Read(Span<byte> buffer)
         {
+            if (buffer.Length == 0)
+                throw new ArgumentException($"{nameof(buffer)} cannot be empty.");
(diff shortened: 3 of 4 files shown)


src/System.Device.Gpio/System/Device/I2c/Drivers/UnixI2cDevice.Linux.cs
```diff
@@ -174,6 +174,9 @@ public override unsafe byte ReadByte()
         /// </param>
         public override unsafe void Read(Span<byte> buffer)
         {
+            if (buffer.Length == 0)
```
tarekgh: >Length [](start = 23, length = 6)

why we throw? should we just return?

tarekgh: this comment apply in all changed places.

---
In reply to: [259633550](https://github.com/dotnet/iot/pull/278#discussion_r259633550) [](ancestors = 259633550)

krwq: zero length is pretty much always a bug on the caller side - why read nothing?

tarekgh: I would suggest to have this method return the read count instead of void and don't throw here. return 0 for 0 length buffer is good indication what could be the issue. 

My point is, this is a public API and you never know the caller scenario which may end up passing zero length buffer. throwing exception is always a headache in such scenarios. we have similar notion like in Encoding and passing empty string, just a valid input and we don't throw. 


---
In reply to: [259636817](https://github.com/dotnet/iot/pull/278#discussion_r259636817) [](ancestors = 259636817)

krwq: @tarekgh this is slightly different to regular Stream Reads/Writes - you have to know up front how many bytes you want to read and scenarios where you don't read the whole buffer are unexpected cc: @JeremyKuhne to double check I'm not saying anything incorrectly

krwq: if reading less is possible then I agree we should return int - so far I haven't seen such case

tarekgh: This API takes Span which I expect the callers will just do Slicing, ending up with empty span is possible. passing empty span and returning 0 is reasonable to me :-) 

> if reading less is possible then I agree we should return int - so far I haven't seen such case

What happen if I allocate 1 GB and send span with that length? does the read block till reading 1 GB?

krwq: @tarekgh I do not believe there currently any devices which support reading like that - currently pretty much every device I worked with required writing an address just before doing any read and number of bytes is known in the specification - I don't think there are any scenarios for this kind of streaming here

tarekgh: Are you saying, the span length always has to match the number of bytes we read? this doesn't sound right to me.

krwq: @tarekgh I'm almost sure that is this case. I think considering the current shape of the API the empty read does not make sense - we can reiterate on the return value of this API later - does that sound reasonable? So far I haven't seen a case where you don't know much about what the other device will send you

tarekgh: go ahead and don't block on my comment. Just make sure you track discussing this more and get some feedback if we need to change the behavior. thanks!

krwq: created https://github.com/dotnet/iot/issues/288 to track this

PR #281, opened 2019-02-24
Dispose CancellationTokenSource

CosminLazar:


src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -231,8 +233,10 @@ public WaitForEventResult WaitForEvent(int pinNumber, PinEventTypes eventTypes,
         /// <returns>A task representing the operation of getting the structure that contains the result of the waiting operation.</returns>
         public ValueTask<WaitForEventResult> WaitForEventAsync(int pinNumber, PinEventTypes eventTypes, TimeSpan timeout)
         {
-            CancellationTokenSource tokenSource = new CancellationTokenSource(timeout);
-            return WaitForEventAsync(pinNumber, eventTypes, tokenSource.Token);
+            using (CancellationTokenSource tokenSource = new CancellationTokenSource(timeout))
+            {
+                return WaitForEventAsync(pinNumber, eventTypes, tokenSource.Token);
```
sharwell: This one seems problematic without `await`.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -201,8 +201,10 @@ public void Write(int pinNumber, PinValue value)
         /// <returns>A structure that contains the result of the waiting operation.</returns>
         public WaitForEventResult WaitForEvent(int pinNumber, PinEventTypes eventTypes, TimeSpan timeout)
         {
-            CancellationTokenSource tokenSource = new CancellationTokenSource(timeout);
-            return WaitForEvent(pinNumber, eventTypes, tokenSource.Token);
+            using (CancellationTokenSource tokenSource = new CancellationTokenSource(timeout))
+            {
+                return WaitForEvent(pinNumber, eventTypes, tokenSource.Token);
```
sharwell: If `WaitForEventResult` is awaitable, this could be disposing too early.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
CosminLazar: Sorry for the sloppy work. Will await for WaitForEventAsync. What about ConfigureAwait(false)?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
CosminLazar: True story, that is the case for the `WaitForEventAsync`, this one is fine because the `WaitForEventResult` is not awaitable

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
sharwell: I'm not sure what the best practice is for this repository. It seems like the ideal behavior would match `ConfigureAwaitRunInline`, but I don't know how to achieve that in the most efficient manner.

CosminLazar: Implemented the changes, can you please take a look?

sharwell:
@richlander I don't see any problems but this is outside my working area so please treat it only as the secondary approval 😄

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFSDriver.Linux.cs
```diff
@@ -432,6 +432,7 @@ protected override void Dispose(bool disposing)
             if (_eventDetectionThread != null && _eventDetectionThread.IsAlive)
             {
                 s_eventThreadCancellationTokenSource.Cancel();
+                s_eventThreadCancellationTokenSource.Dispose();
```
joperezr: Should this wait till the cancellation token takes effect? @stephentoub do you mind taking a quick look at this?

stephentoub: `Cancel` invokes all of the registered cancellation handlers synchronously, so "wait till cancellation token takes effect" is immediate.  From that perspective, calling Dispose here is fine.  The main thing you need to look out for is making sure that after the CTS is disposed, no one else tries to use the CTS (using the CT is fine), e.g. I see on L417 another call to Cancel on this source... you need to make sure that can't happen after this Dispose call.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
krwq: I'm curious - will the timer still cancel even though the CTS is disposed? seems a bit confusing but I'm not an expert on this

stephentoub: Disposing a CTS disposes of its Timer.

Issue #282, opened 2019-02-24
Update compiler


krwq:
I've encountered this bug yesterday: https://github.com/dotnet/corefx/issues/35550
when trying to remove some allocations from one of the projects - this is really confusing and much more likely to happen in our repo than in any other (due to writing over SPI/I2C).

We should update our dependencies to make sure that changeset fixing https://github.com/dotnet/roslyn/issues/29092 is included.

cc: @safern

joshfree:
Is there a newer drop of buildtools with the updated compiler we can just pick up?

safern:
> Is there a newer drop of buildtools with the updated compiler we can just pick up?

I don't think iot uses buildtools at all. 

I guess we will have to do something like we did in corefx to pick up newer compilers:
https://github.com/dotnet/corefx/pull/35134

@krwq please let me know if you need me to send a PR to pick up a new compiler.

krwq:
Thanks again @safern!

PR #283, opened 2019-02-26
Update compilers to use the latest 3.0.0-beta3-final

safern:
Fixes: #282 

cc: @joshfree


krwq:
Thanks @safern - LGTM - I'll test at home if it fixes the original issue

Directory.Build.props
```diff
@@ -13,5 +13,15 @@
     <TargetsWindows Condition="$(RuntimeIdentifier.StartsWith('win'))">true</TargetsWindows>
     <TargetsLinux Condition="$(RuntimeIdentifier.StartsWith('linux'))">true</TargetsLinux>
     <LangVersion>Latest</LangVersion>
+    <!-- We use the compiler toolset that comes from NuGet Packages rather than the SDK built-in.
+    This one sets UseSharedCompilation to false by default. -->
```
joperezr: The comment doesn’t seem to match whith the value you set, should this be false instead?

Directory.Build.props
```diff
@@ -13,5 +13,15 @@
...
+    <UseSharedCompilation>true</UseSharedCompilation>
+
+    <CompilerVersion>3.0.0-beta3-final</CompilerVersion>
   </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Condition="'$(MSBuildRuntimeType)' != 'Core'" Include="Microsoft.Net.Compilers" Version="$(CompilerVersion)" IsImplicitlyDefined="true" />
```
joperezr: What does the IsImplicitlyDefined do?

safern:
Hmm, build is failing with:

```
/Users/vsts/agent/2.147.1/work/1/s/.packages/microsoft.netcore.compilers/3.0.0-beta3-final/tools/Microsoft.Managed.Core.targets(90,5): error MSB4062: The "Microsoft.CodeAnalysis.BuildTasks.MapSourceRoots" task could not be loaded from the assembly /Users/vsts/agent/2.147.1/work/1/s/.packages/microsoft.netcore.compilers/3.0.0-beta3-final/tools/Microsoft.Build.Tasks.CodeAnalysis.dll. Assembly with same name is already loaded Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. [/Users/vsts/agent/2.147.1/work/1/s/src/System.Device.Gpio/System.Device.Gpio.csproj]
```

I thought nodeReuse was set to false on CI. I'll see if it is really false, if not I'll set it to false.

Directory.Build.props
safern: No, actually that is the point, the NuGet package compilers set UseSharedCompilation to false by default, the sdk one sets it to true.

Directory.Build.props
safern: It is for the IDE so that people can't remove the package reference from VisualStudio UI.

Directory.Build.props
joperezr: Oh I see. I guess the term “this one” is a bit confusing. If you could be a bit more specific there saying we require it to be true but the package one sets the default to false it would be better.

safern: Ok, I'll add it.

joperezr:
As an aside question, is there a reason why we need compilers feom a package? Why not just update the CLI we use to the very latest RTM? That way the compilers will be crossgened

safern:
Just FYI per the compilers team, state that there isn’t a big perf change on using the crossgenned compiler vs non cross gen. https://github.com/dotnet/corefx/pull/35184#issuecomment-461963586

Are we OK with using a pre release version of the SDK? If that is the case then I can just update global.json file, but I followed the packages approach since I didn’t know if we wanted to use a pre release version of the SDK

krwq:
@joperezr does that version include the fix for the issue this is fixing? (broken stackalloc initializer)

@safern I'd go with the fix similar as is for now and we can update it later. Updating SDK is fine too but we need to make sure the issue is actully fixed there too.

I think using pre-release SDK is fine if that is the only way to fix the issue.

safern:
@agocke I'm trying to use the new compilers through packages, but that is causing the build to fail when -maxcpucount is enabled. It seems like the multiple nodes are trying to load a Task from the packages that is already loaded. cc: @rainersigwald as this could be msbuild's issue rather than roslyn.

agocke:
There should be no problem trying to load the task multiple times *as long as you're trying to load the same task*. It's likely the problem is that you're not using the exact same compiler package in all the projects in your solution.

safern:
> It's likely the problem is that you're not using the exact same compiler package in all the projects in your solution.

Hmm, I just doubled checked and there is no other project referencing another version of the compiler package. Actually I added the package references to Directory.Build.props to be shared across all projects.

```
F:\vsagent\16\s\.packages\microsoft.netcore.compilers\3.0.0-beta3-final\tools\Microsoft.Managed.Core.targets(90,5): error MSB4062: The "Microsoft.CodeAnalysis.BuildTasks.MapSourceRoots" task could not be loaded from the assembly F:\vsagent\16\s\.packages\microsoft.netcore.compilers\3.0.0-beta3-final\tools\Microsoft.Build.Tasks.CodeAnalysis.dll. Assembly with same name is already loaded Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. [F:\vsagent\16\s\src\System.Device.Gpio\System.Device.Gpio.csproj]
```

Maybe related to this is that this project is built twice at the same time in two separate MSBuild nodes?  Because it has multiple RIDs? 

https://github.com/dotnet/iot/blob/de5821242385a500207d8766d1d1de7dfb2b97db/Directory.Build.targets#L29

rainersigwald:
Arcade does scary things with compiler package versions; I think @agocke is correct that the problem is that the build is trying to use more than one at once. It shouldn't be a problem (in this way) to build the same project multiple times concurrently.

safern:
It seems like indeed, there where different versions of the compiler being used. The samples projects have a Directory.Build.props in their root that doesn't import the root compiler which was causing them to use the compiler shipped within the SDK. However that adds duplication on the repo if we want to add the packages. So I went with updating the SDK and CLI to latest preview3. @krwq could you please confirm that has the fix you need?

Directory.Build.props
```diff
@@ -13,5 +13,17 @@
     <TargetsWindows Condition="$(RuntimeIdentifier.StartsWith('win'))">true</TargetsWindows>
     <TargetsLinux Condition="$(RuntimeIdentifier.StartsWith('linux'))">true</TargetsLinux>
     <LangVersion>Latest</LangVersion>
+
+    <!-- Packaging props -->
+    <PackageLicenseFile>LICENSE</PackageLicenseFile>
+    <!-- We need to clear PackageLicenseUrl since arcade sets it, and since we're using the new compilers
```
ericstj: nit: Compilers should have nothing to do with NuGet so this comment may be a little unusual to future folks outside the context of this change.  I'd merely state that the property is deprecated so we ensure it isn't set.

global.json
```diff
@@ -1,9 +1,9 @@
 {
   "tools": {
-    "dotnet": "2.1.403"
+    "dotnet": "3.0.100-preview3-010395"
```
ericstj: Is this a publicly released preview?  I believe whatever you list here will be the minimum required by contributors when working in VS as well.

Directory.Build.props
safern: Ok, sounds good.

krwq:
I run local build `./build.sh` with this changeset and it failed. I did `git clean -fdx` and removed `~/.nuget/`, repeated and failed again.

Here are the errors I'm getting (very similar for almost every project) - is this expected that it asks me to install 2.1.0 globally?:

```
/home/krwq/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error : It was not possible to find any compatible framework version [/home/krwq/work/iot/src/devices/Mcp3008/Mcp3008.csproj]
/home/krwq/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error : The specified framework 'Microsoft.NETCore.App', version '2.1.0' was not found. [/home/krwq/work/iot/src/devices/Mcp3008/Mcp3008.csproj]
/home/krwq/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error :   - Check application dependencies and target a framework version installed at: [/home/krwq/work/iot/src/devices/Mcp3008/Mcp3008.csproj]
/home/krwq/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error :       /home/krwq/work/iot/.dotnet/ [/home/krwq/work/iot/src/devices/Mcp3008/Mcp3008.csproj]
/home/krwq/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error :   - The .NET Core Runtime and SDK can be installed from: [/home/krwq/work/iot/src/devices/Mcp3008/Mcp3008.csproj]
/home/krwq/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error :       https://aka.ms/dotnet-download [/home/krwq/work/iot/src/devices/Mcp3008/Mcp3008.csproj]
/home/krwq/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error :   - The following versions are installed: [/home/krwq/work/iot/src/devices/Mcp3008/Mcp3008.csproj]
/home/krwq/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error :       3.0.0-preview3-27427-8 at [/home/krwq/work/iot/.dotnet/shared/Microsoft.NETCore.App] [/home/krwq/work/iot/src/devices/Mcp3008/Mcp3008.csproj]
/home/krwq/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error :   - Installing .NET Core prerequisites might help resolve this problem: [/home/krwq/work/iot/src/devices/Mcp3008/Mcp3008.csproj]
/home/krwq/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error :       https://go.microsoft.com/fwlink/?linkid=2063370 [/home/krwq/work/iot/src/devices/Mcp3008/Mcp3008.csproj]
```

safern:
It is possible that you have multilevel lookup enabled and it is trying to use the SDK/CLI from your path rather than the one bootstrapped for the repo.

But yes I would expect the new compilers to not support 2.0 since it is EOL already. Try disabling multilevel lookup with the environment variable.

krwq:
@safern I do have muti level lookup enabled but it is enforcing `<repo-root>/.dotnet/` and not my local dotnet version

krwq:
I just tried 3 combinations:
- `DOTNET_MULTILEVEL_LOOKUP=1`
- `DOTNET_MULTILEVEL_LOOKUP=0`
- `DOTNET_MULTILEVEL_LOOKUP=`

none of them builds with this changeset

krwq:
perhaps this might be useful

```
> dotnet --info
A compatible SDK version for global.json version: [3.0.100-preview3-010395] from [/home/krwq/work/iot/global.json] was not found

Host (useful for support):
  Version: 2.2.2
  Commit:  a4fd7b2c84

.NET Core SDKs installed:
  2.1.403 [/usr/share/dotnet/sdk]
  2.1.504 [/usr/share/dotnet/sdk]
  2.2.104 [/usr/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.8 [/usr/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.2 [/usr/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.8 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.2.2 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.5 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.8 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.2.2 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
```

Directory.Build.props
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
+  <Import Project="$(RepoRoot)eng/Compilers.props" />
```
ericstj: nit: I think there may be a prop defined by arcade for the eng directory.

ericstj:
@krwq it looks to me like your .dotnet is different than specified by the repo.  https://github.com/dotnet/iot/blob/750f2c149b34d516c43cab87a58ff9d48e4ae01d/global.json#L3 calls for 2.1.0, however 
https://github.com/dotnet/iot/pull/283#issuecomment-470420455 indicates that the .dotnet in your repo folder is a 3.0 version.  Make sure you delete your .dotnet folder, double check your global.json, then retry build.

krwq:
@ericstj I did `git clean -fdx` and have purely this changeset so not sure what to check...

krwq:
The first version which I signed off was green on my box and red on CI - now it is opposite

krwq:
Is the recommended fix to install 2.1.0 globally?

safern:
> A compatible SDK version for global.json version: [3.0.100-preview3-010395] from [/home/krwq/work/iot/global.json] was not found

Did you double check the global.json in your local repo that it actually points to 2.1.403?

It seems like you got on a weird state where your global.json is specifying another version of the SDK which can't be found.

Directory.Build.props
safern: Oh right: RepositoryEngineeringDir

```suggestion
  <Import Project="$(RepositoryEngineeringDir)Compilers.props" />
```

PR #284, opened 2019-02-26
Initial stub for conventions for devices

krwq:
Replaces: https://github.com/dotnet/iot/pull/256

This is an attempt to formalize some conventions we should follow for the devices.

This document should be as short possible so that everyone can read it quickly and later can point to the larger documents with more details.

Please let me know if there is something I should add there.

cc: @tarekgh @JeremyKuhne @joperezr

### DIFF
--- a/Documentation/Devices-conventions.md
+++ b/Documentation/Devices-conventions.md
@@ -1,0 +1,20 @@
+# Conventions for devices APIs
+
+## Value returned by sensor
+
+- prefer property (i.e. Temperature) rather than method (i.e. GetTemperature)
+- use `double` when you need to return any floating point value
+- use `Vector2/3/4` for returning vectors`*`
+- value should conform to units conventions (see below)
+
+`*` -  Vector3 is currently backed by float, this may be changed in the future: https://github.com/dotnet/corefx/issues/25334
+
+## Units
+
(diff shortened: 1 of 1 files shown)


Documentation/Devices-conventions.md
```diff
@@ -0,0 +1,25 @@
+# Conventions for devices APIs
+
+## Value returned by sensor
+
+- prefer property (i.e. Temperature) rather than method (i.e. GetTemperature)
```
tarekgh: nit: start with upper case letter in every paragraph/line

Documentation/Devices-conventions.md
```diff
@@ -0,0 +1,25 @@
...
+- use `double` when you need to return any floating point value
+- use `Vector2/3/4` for returning vectors`*`
+- value should conform to units conventions (see below)
+- only the most useful APIs should be public, anything else which may be useful but will unlikely get used by most of the people should be protected (inheriting the class allows you to use it but it is not visible by default) or internal/private
```
tarekgh: I think we should tell more when protected should be used. like API is believed not going to be used much but there is some specific non main stream scenario can be used so making it protected can allow such scenarios to be achieved.

krwq: I'll let someone else fix the wording and merge as is for now (with casing fix)

Documentation/Devices-conventions.md
krwq: 👍

PR #285, opened 2019-02-26
Support for reading CPU temperature on Raspberry Pi.

gloveboxes:
This reads the CPU temperature on the Raspberry Pi. The telemetry can be used by itself but also for calibrating the temperature of the Raspberry Pi Sense HAT.


src/devices/RpiCpuTemperature/README.md
```diff
@@ -0,0 +1,11 @@
+# RpiCpuTemperature
```
ZhangGaoxing: I don't think it's supposed to be a binding, put in samples? Waiting for comments from other members😄

src/devices/RpiCpuTemperature/RpiCpuTemperature.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Iot.Device.Rpi
+{
+    public class CpuTemperature
+    {
+        public double? ReadTemperature()
```
ZhangGaoxing: please use property and change to private
```C#
public double? Temperature => ReadTemperature();
```

ZhangGaoxing: Or use static methods?

src/devices/RpiCpuTemperature/README.md
krwq: I think this is fine but it should not be specific to Raspberry Pi - I suspect we will be able to make it work for other boards as well at one point

src/devices/RpiCpuTemperature/RpiCpuTemperature.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Iot.Device.Rpi
```
krwq: Iot.Device.CpuTemperature/CpuThermometer

src/devices/RpiCpuTemperature/RpiCpuTemperature.cs
krwq: I'd just go with double and throw elsewhere and perhaps have another static property `IsAvailable` or something along these lines

src/devices/RpiCpuTemperature/RpiCpuTemperature.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class CpuTemperature
+    {
+        public double? ReadTemperature()
+        {
+            double? _temperature = null;
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && File.Exists("/sys/class/thermal/thermal_zone0/temp"))
+            {
+                using (FileStream _fileStream = new FileStream("/sys/class/thermal/thermal_zone0/temp", FileMode.Open, FileAccess.Read))
+                using (StreamReader _reader = new StreamReader(_fileStream))
```
krwq: nit: StreamReader already disposes stream (there is no harm in double disposing) but you can just construct file stream directly in this call

src/devices/RpiCpuTemperature/RpiCpuTemperature.cs
```diff
@@ -0,0 +1,32 @@
...
+                {
+                    string _data = _reader.ReadLine();
```
krwq: local variables should be camelCase without prefixes. (only private fields should be _fooBar)

src/devices/RpiCpuTemperature/RpiCpuTemperature.cs
```diff
@@ -0,0 +1,32 @@
...
+                    if (!string.IsNullOrEmpty(_data))
+                    {
+                        int _temp;
+                        if (int.TryParse(_data, out _temp))
+                        {
+                            _temperature = Math.Round(_temp / 1000F, 1);
```
krwq: why rounding?

src/devices/RpiCpuTemperature/RpiCpuTemperature.cs
```diff
@@ -0,0 +1,32 @@
...
+                        }
+                    }
+                }
+            }
```
krwq: enter after brace

src/devices/RpiCpuTemperature/RpiCpuTemperature.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
```
krwq: license header missing (copy from other cs file)

src/devices/RpiCpuTemperature/samples/RpiCpuTemperature.Sample.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
```
krwq: license header

krwq:
Thanks for putting this up @gloveboxes!

src/devices/RpiCpuTemperature/RpiCpuTemperature.cs
gloveboxes: hey I round as the data below one decimal point is so variable as to not mean anything.

src/devices/CpuTemperature/CpuTemperature.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Runtime.InteropServices;
+
+namespace Iot.Device.CpuTemperature
+{
+    public class CpuTemperature
+    {
+        public double Temperature => ReadTemperature();
+        private double ReadTemperature()
+        {
+            double temperature = 0.0;
```
krwq: I'd go with max/min/NaN when thermometer cannot be found - 0 seems like a valid temperature and might be misinterpreted

src/devices/CpuTemperature/CpuTemperature.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Iot.Device.CpuTemperature
+{
+    public class CpuTemperature
+    {
+        public double Temperature => ReadTemperature();
```
krwq: nit: enter before function

src/devices/CpuTemperature/CpuTemperature.cs
```diff
@@ -0,0 +1,37 @@
...
+                    if (!string.IsNullOrEmpty(data))
+                    {
+                        int temp;
+                        if (int.TryParse(data, out temp))
+                        {
+                            temperature = temp / 1000F;
+                        }
+                    }
+                }
+            }
```
krwq: nit: enter after the brace

gloveboxes:
Added IsAvailable property to confirm the validity of the CPU temp

src/devices/CpuTemperature/CpuTemperature.cs
krwq: I think:
- disregardless of having or not having IsAvailable property this should be NaN or double.Max when not available
- IsAvailable can be cached and calculated in the constructor and then just opened - assuming the file once existed it will not just disappear

krwq:
So to clarify my comment since it might have got lost:

I think:
    * disregardless of having or not having IsAvailable property Temperature should be NaN or double.Max when not available
    * IsAvailable can be cached and calculated in the constructor and then file just opened without any checks - assuming the file once existed it will not just disappear

gloveboxes:
added cached implementation to IsAvailable

gloveboxes:
added cached IsAvailable

krwq:
@gloveboxes looks good - I've fixed up braces in your PR to avoid another cycle :smile:

PR #286, opened 2019-02-26
Add Mpr121 binding.

SergeyRazmyslov:
The 12-channels I2C proximity capacitive touch sensor controller.


src/devices/Mpr121/Mpr121.cs
```diff
@@ -0,0 +1,252 @@
...
+
+        private Dictionary<Channels, bool> _statuses;
+
+        private int _periodRefresh;
+
+        /// <summary>
+        /// Notifies about a property has been changed.
+        /// Refresh period can be changed by setting PeriodRefresh property.
+        /// </summary>
+        public event PropertyChangedEventHandler PropertyChanged;
```
ZhangGaoxing: Change to a more meaningful name? I think you can implement it with a custom event instead of using INotifyPropertyChanged. In my impression, INotifyPropertyChanged is commonly used in programs with interfaces. 😄

src/devices/Mpr121/README.md
```diff
@@ -0,0 +1,67 @@
...
+#### Channel statuses auto refresh
+
+```csharp
+var i2cDevice = new UnixI2cDevice(new I2cConnectionSettings(busId: 1, deviceAddress: 0x5A));
+
+// Initialize controller with default configuration and auto-refresh the channel statuses every 100 ms.
+var mpr121 = new Mpr121(i2cDevice, 100);
+
+// Subscribe to channel statuses updates.
+mpr121.PropertyChanged += (object sender, PropertyChangedEventArgs e) => {
```
ZhangGaoxing:
```C#
mpr121.PropertyChanged += (object sender, PropertyChangedEventArgs e) => 
{

}
```

ZhangGaoxing:
left some comments😄

src/devices/Mpr121/Channels.cs
```diff
@@ -0,0 +1,70 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Mpr121
+{
+    // Mpr121 channels.
```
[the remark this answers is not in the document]
krwq: wondering if we should just go with `int` instead and explain valid ranges with static property and exception

src/devices/Mpr121/Mpr121.cs
```diff
@@ -0,0 +1,252 @@
...
+        {
+            get { return _periodRefresh; }
+
+            set
+            {
+                _periodRefresh = value;
+
+                if (_periodRefresh > 0 && _timer == null)
+                {
+                    _timer = new Timer(RefreshChannelStatuses, this, TimeSpan.FromMilliseconds(_periodRefresh), TimeSpan.FromMilliseconds(_periodRefresh));
```
krwq: does it make sense to instead read status only when user tries to access it?

src/devices/Mpr121/Mpr121.cs
```diff
@@ -0,0 +1,252 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Initialize a MPR121 controller.
+        /// </summary>
+        /// <param name="device">The i2c device.</param>
+        /// <param name="periodRefresh">The period in milliseconds of refresing the channel statuses.</param>
+        /// <param name="configuration">The controller configuration.</param>
+        public Mpr121(I2cDevice device, int periodRefresh, Mpr121Configuration configuration)
```
krwq: I'd leave just this constructor with default values. I.e.:

```csharp
public Mpr121(I2cDevice device, int periodRefresh = 0, Mpr121Configuration configuration = null)
{
    configuration = configuration ?? GetDefaultConfiguration();
...
```

src/devices/Mpr121/Channels.cs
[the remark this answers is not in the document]
SergeyRazmyslov: Having this enum, users don't need to check values in the channels status dictionary. Please let me know if it's better to replace this by ```int```.

src/devices/Mpr121/Mpr121.cs
```diff
@@ -0,0 +1,252 @@
...
+    {
+        private static readonly int CHANNELS_NUMBER = Enum.GetValues(typeof(Channels)).Length;
+
+        private I2cDevice _device;
+        private Timer _timer = null;
+
+        private Dictionary<Channels, bool> _statuses;
+
+        private int _periodRefresh;
+
```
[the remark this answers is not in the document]
SergeyRazmyslov: Fixed. I think it's better to use ```static readonly``` instead of ```const```.

src/devices/Mpr121/Mpr121.cs
SergeyRazmyslov: I'm afraid we can get performance issues if we will refresh the status implicitly. Let's imagine situation when user's code looks like:
```
var channel1 = mpr121.ChannelStatuses[Channels.CH_1];
var channel2 = mpr121.ChannelStatuses[Channels.CH_2];
```

In that case the channel statuses will be read twice.

src/devices/Mpr121/Channels.cs
```diff
@@ -0,0 +1,70 @@
...
+    public enum Channels
+    {
+        /// <summary>
+        /// Channel 0.
+        /// </summary>
+        CH_0 = 0,
```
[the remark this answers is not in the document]
SergeyRazmyslov: Fixed.

src/devices/Mpr121/Mpr121.cs
SergeyRazmyslov: Fixed.

src/devices/Mpr121/Mpr121.cs
SergeyRazmyslov: I made some refactoring and now we don't need the ChannelStatuses property anymore. We can get the channel statuses from event args in the auto-refresh mode. Otherwise, we can call ReadChannelStatuses() method which read the status and return the value.

src/devices/Mpr121/Mpr121.cs
SergeyRazmyslov: Fixed.

src/devices/Mpr121/README.md
SergeyRazmyslov: Fixed.

src/devices/Mpr121/Channels.cs
```diff
@@ -0,0 +1,70 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Mpr121
+{
+    // MPR121 channels.
```
krwq: I think the comments here and above the values can be removed - they are not super useful anyway 😄

src/devices/Mpr121/Mpr121.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// <summary>
+        /// Notifies about a the channel statuses have been changed.
+        /// Refresh period can be changed by setting PeriodRefresh property.
+        /// </summary>
+        public event EventHandler<ChannelStatusesChangedEventArgs> ChannelStatusesChanged;
+
+        /// <summary>
+        /// Gets ot sets the period in milliseconds to refresh the channels statuses.
```
krwq: `Gets ot sets` -> `Gets or sets`

src/devices/Mpr121/Mpr121.cs
```diff
@@ -0,0 +1,220 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Initialize a MPR121 controller.
+        /// </summary>
+        /// <param name="device">The i2c device.</param>
+        /// <param name="periodRefresh">The period in milliseconds of refresing the channel statuses.</param>
+        /// <param name="configuration">The controller configuration.</param>
+        public Mpr121(I2cDevice device, int? periodRefresh = null, Mpr121Configuration configuration = null)
```
krwq: This is fairly recent agreement on the conventions and likely most of the places in the code still use nullable: if negative value for integer doesn't make sense use `-1`  instead of null

src/devices/Mpr121/Mpr121.cs
```diff
@@ -0,0 +1,220 @@
...
+            {
+                _timer.Dispose();
+                _timer = null;
+            }
+        }
+
+        /// <summary>
+        /// Reads the channel statuses of MPR121 controller.
+        /// </summary>
+        public IReadOnlyDictionary<Channels, bool> ReadChannelStatuses()
```
krwq: I don't know much about this device but does it make sense to just create:
`bool ReadChannelStatus(Channel channel)` or `GetChannelStatus`?

src/devices/Mpr121/Channels.cs
SergeyRazmyslov: Fixed.

src/devices/Mpr121/Mpr121.cs
SergeyRazmyslov: Fixed.

src/devices/Mpr121/Mpr121.cs
SergeyRazmyslov: Fixed.

src/devices/Mpr121/Mpr121.cs
SergeyRazmyslov: Yes, this method could be useful. I added it with some remarks in xmldoc.

SergeyRazmyslov:
Thank you Guys for review! I resolved comments and also I refactored interaction with Timer. Could you please take a look this?

krwq:
@SergeyRazmyslov thanks for fixing comments - I think after you have removed:
```csharp
configuration = configuration ?? GetDefaultConfiguration();
```

it will now throw NullReferenceException when configuration is not set (you might want to not call InitializeController or add it back)

krwq:
Looks good, thanks @SergeyRazmyslov!

Issue #197, opened 2019-02-01
Support one-wire protocol


ericstj:
During the IOT hackathon I was playing with some [DS18B20](https://cdn-shop.adafruit.com/datasheets/DS18B20.pdf) temperature sensors.  These use the one-wire protocol which is supported on the RaspberyPi by remapping a GPIO pin to act as a one-wire bus.  For more information see
https://www.raspberrypi.org/documentation/configuration/device-tree.md
https://www.kernel.org/doc/Documentation/gpio/drivers-on-gpio.txt
https://www.kernel.org/doc/html/latest/driver-api/w1.html
https://www.kernel.org/doc/Documentation/w1/

The DS18B20 is a pretty popular and inexpensive digital thermometer so it seems worthwhile supporting this.

I haven't yet found official Windows support for one wire, there are a few hacks that use the serial port or run a software timer in  user mode but I can't imagine they are reliable.  I think Windows would need a kernel driver for this like Linux has (though I could be missing something).  

For now I'm going to proceed trying to implement the controller for OneWire on linux along with device binding for the DS18B20.

maloo:
Hi,
please keep me up to date. I'm just about to use a MAX31820 which is very similar. I use this as an ambient temp meter on my latest PCB design.

joperezr:
note that #172 is already tracking the work to add the binding for DS18B20 in case you end up working in both we should close both issues

ericstj:
It would seem that linux automatically installs kernel device drivers for the DS18B20 and similar devices: https://www.kernel.org/doc/Documentation/w1/slaves/w1_therm

From the docs https://www.kernel.org/doc/Documentation/w1/w1.generic
```
When a device is found on the bus, w1 core tries to load the driver for its family
and check if it is loaded. If so, the family driver is attached to the slave.
If there is no driver for the family, default one is assigned, which allows to perform
almost any kind of operations. Each logical operation is a transaction
in nature, which can contain several (two or one) low-level operations.
```

Still researching to understand what's the right way to surface all this in the GPIO library and device bindings.

maloo:
Looks like the kernel is handling the known bridges already.

ericstj:
I believe the current design of the GPIO library is to be a lower-level component which doesn't "know" about specific device types. 

I can imagine an API for OneWire controller might look like enumerating the bus'es, enumerating the devices on the bus in a generic way, and exposing any generic commands that are available for devices without a family driver.

[...]

Issue #172, opened 2019-01-30
Add bindings for DS18B20 - Digital temperature sensor


krwq:

maloo:
Hi, I can help,
I need to write a driver for a MAX31820, I guess these are very similar. Do you have any API. Do you want a 1-wire device and a device driver or just a device driver?

joperezr:
Hello @maloo, I know that @ericstj was planning on working on this one as well. He is planning to add support for 1-Wire into our main library. I'm not sure if he is planning to also write a device binding for the actual sensor, so just an FYI so that you don't both end up doing the same work :smile:

krwq:
let me update this and other issues so that we don't end up investigating

joshfree:
removing the up-for-grabs label as @ericstj grabbed this along with general 1-WIRE support.

@ericstj if your plans change please flip this back to up-for-grabs.  Thanks

MaklaCof:
Is there something new. I would like to replace really my old device for measuring hot water in boiler with a new one, based on RaspberryPI. The best suited sensor for this is DS18B20.

maloo:
Waiting for @krwq to hit the merge button of #793

krwq:
This is merged already, thanks @maloo!

PR #289, opened 2019-02-26
1-wire controller and thermometer binding

maloo:
Proof of concept implementation of proposed API in #197.

Basic code sample to scan for devices, enumerate buses and devices and print temperature of all thermometer family devices found.

Limitations:
- Only Linux support
- Only Thermometer (family) binding
- Not much error handling
- No docs

### DIFF
--- a/src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.Linux.cs
+++ b/src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.Linux.cs
@@ -1,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Iot.Device.OneWire
+{
+    public partial class OneWireThermometerDevice : OneWireDevice
+    {
+        private async Task<float> ReadTemperatureInternalAsync()
+        {
+            // Expected data format:
+            // 7f 01 4b 46 7f ff 01 10 33 : crc=33 YES
+            // 7f 01 4b 46 7f ff 01 10 33 t=23937
+            var data = await File.ReadAllTextAsync(Path.Combine(OneWireController.SysfsDevicesPath, Bus.DeviceId, DeviceId, "w1_slave"));
+            if (!data.Contains("YES"))
+                throw new IOException("Unable to read temperature from device.");
+            var tempIdx = data.LastIndexOf("t=");
+            if (tempIdx == -1 || tempIdx + 2 >= data.Length || !int.TryParse(data.AsSpan(tempIdx + 2), out var temp))
+                throw new FormatException("Invalid sensor data format.");
+            return temp * 0.001f;
+        }
+    }
+}
--- a/src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
+++ b/src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
@@ -1,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading.Tasks;
+
+namespace Iot.Device.OneWire
+{
+    public partial class OneWireThermometerDevice : OneWireDevice
+    {
+        protected internal OneWireThermometerDevice(OneWireBus bus, string deviceId, int family)
+            : base(bus, deviceId, family)
+        {
+        }
+
+        public Task<float> ReadTemperatureAsync()
+        {
+            return ReadTemperatureInternalAsync();
+        }
+    }
+}
--- a/src/devices/OneWire/OneWireBus.cs
+++ b/src/devices/OneWire/OneWireBus.cs
@@ -1,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Iot.Device.OneWire
+{
+    public class OneWireBus
+    {
+        internal OneWireBus(string deviceId)
+        {
+            DeviceId = deviceId;
+        }
+
+        public string DeviceId { get; }
+
+        public IEnumerable<OneWireDevice> EnumerateDevices(int family = -1)
+        {
+            return OneWireController.EnumerateDevices(this, family);
+        }
+
+        public Task ScanForDevicesAsync(int numDevices = 5, int numScans = -1)
+        {
+            return OneWireController.ScanForDevicesInternal(this, numDevices, numScans);
+        }
+    }
+}
--- a/src/devices/OneWire/OneWireController.cs
+++ b/src/devices/OneWire/OneWireController.cs
@@ -1,0 +1,44 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace Iot.Device.OneWire
+{
+    public static partial class OneWireController
+    {
+        public static IEnumerable<OneWireBus> EnumerateBuses()
+        {
+            return EnumerateBusesInternal();
+        }
+
+        public static IEnumerable<OneWireDevice> EnumerateDevices(int family = -1)
+        {
+            foreach (var bus in EnumerateBuses())
+            {
+                foreach (var dev in bus.EnumerateDevices(family))
+                {
+                    yield return dev;
+                }
+            }
+        }
+
+        internal static OneWireDevice CreateDeviceByFamily(OneWireBus bus, string deviceId, int family)
+        {
+            switch (family)
+            {
+                case 0x10:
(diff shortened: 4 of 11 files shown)


krwq:
cc: @ericstj 
ref: https://github.com/dotnet/iot/issues/172

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.Linux.cs
```diff
@@ -0,0 +1,27 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Iot.Device.OneWire
+{
+    public partial class OneWireThermometerDevice : OneWireDevice
```
krwq: I'lll take a deeper look later but for a quick feedback, I think this device should go by name (i.e. DS18B20) and be under Iot.Device.Ds18B20. The 1-wire should be a separate directory which this project would reference.

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Iot.Device.OneWire
+{
+    public partial class OneWireThermometerDevice : OneWireDevice
+    {
+        protected internal OneWireThermometerDevice(OneWireBus bus, string deviceId, int family)
+            : base(bus, deviceId, family)
+        {
+        }
+
+        public Task<float> ReadTemperatureAsync()
```
krwq: Please make sure to read our conventions https://github.com/dotnet/iot/pull/284 (still WIP but should give rough idea of the expectations). TL;DR is that we currently prefer property `Temperature` which returns `double`. We do not have yet opinion on having or not having async APIs so fine to keep it along the property (the return type should be double though)

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.Linux.cs
krwq: Actually second look - not seeing much device specific. Please change only if this is DS18B20 specific

krwq: (please at least mention DS18B20 somewhere in the readme since it will likely cause that someone will duplicate the effort)

src/devices/OneWire/README.md
```diff
@@ -0,0 +1,18 @@
+﻿# OneWire
+
+## Summary
+Provide a brief description on what the component is and its functonality.
```
krwq: please add some comment on how to connect it or set it up on linux.

Description can be really brief and other bullets can be removed, i.e.:

```
# 1-wire

Allows using 1-wire devices such as digital thermometers (i.e. DS18B20).

## How to setup 1-wire

(put description here)
```

if there is nothing needed for setup at least mention if it can be connected to any pin and if not then which one can be used

src/devices/OneWire/OneWireBus.cs
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        public string DeviceId { get; }
+
+        public IEnumerable<OneWireDevice> EnumerateDevices(int family = -1)
+        {
+            return OneWireController.EnumerateDevices(this, family);
+        }
+
+        public Task ScanForDevicesAsync(int numDevices = 5, int numScans = -1)
```
krwq: personally I'm not fan of async APIs for stuff which is fairly fast - would be nice to have a sync and async versions if you strongly feel about having async.

src/devices/OneWire/OneWireController.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Iot.Device.OneWire
+{
+    public static partial class OneWireController
+    {
+        public static IEnumerable<OneWireBus> EnumerateBuses()
+        {
+            return EnumerateBusesInternal();
+        }
+
+        public static IEnumerable<OneWireDevice> EnumerateDevices(int family = -1)
```
krwq: Should family be enum instead? I think there is just few 1-wire devices total. How does user know what value to use for say Thermometer?

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.Linux.cs
maloo: Yes, I actually wrote this for the MAX31820. The binding supports 5 1-wire device families (temp sensors). MAX31820 and DS18B20 are both of the same family. So I can't use the device name in the binding. But can definitely add some device names in the readme.
My intent was not to hijack ericstj's issue #197. I had to implement this anyways and just uploaded it as a WIP PR in case eric could make use of it. I can continue the work, but then I want eric's blessing ;). Now it has all I need for my project.

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
maloo: Will fix double, but async is needed. It can take up to 1 sec to read one sensor. I usually go by MS WinRT standards of anything over 50ms is async.

src/devices/OneWire/README.md
maloo: If you (and eric) decide I should push forward on this PR, then of course I will fix comments, readme and sample.

src/devices/OneWire/OneWireBus.cs
maloo: This is WIP. I didn't find a specific notification of when the Linux driver is done searching for devices. I will have to read the kernel driver in more detail to find a possible trigger of the search ending. Worst case is that I have to read the driver timeout and poll a bit after that to see if it is done.
Also, there is a continuous search/scan mode in Linux. Not sure if that should have its' own API. Now I just used the driver one using -1 attempts for continuous.
And if using scan with multiple searches, should the operation finish after the first one when the devices are most likely found or the last one?
As for async, I'm a fan ;) And in this case the single search timeout is 10 seconds (read only) in Linux. Which is enough for motivating async. But as I said, I don't wait that long yet.

src/devices/OneWire/OneWireController.cs
maloo: I started with that, but there is no good name for the families. And many families have overlapping feature set, like temp sensors are in 5 families.
[Devices](https://github.com/owfs/owfs-doc/wiki/1Wire-Device-List)
[Families](http://owfs.sourceforge.net/simple_family.html)

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.Linux.cs
krwq: I've just talked with @ericstj and he is fine with you picking this up 😄

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
krwq: fair enough - usually it is much less than 1ms - btw. any chance we could reduce that timing? If you don't have ideas I can tale a look later but I have some flashbacks from years ago that 1-wire requires some large delays so you might be right. Let's leave it as is and I'll bring this up next time I'll be talking about conventions and update the doc accordingly

src/devices/OneWire/README.md
krwq: yes you got Eric's blessing 😄

src/devices/OneWire/OneWireBus.cs
krwq: np :) Hardware is weird and some devices are more `special` than others - 1-wire seems to be like that 😄

src/devices/OneWire/OneWireController.cs
krwq: you can put 2 names for the same value in the enum - i.e.:
```csharp
enum DeviceFamily : byte
{
    Ds18B20 = 5,
    Max31820 = 5,
    DigitalThermometer = 5,
}

```

this way you can have both names but in the implementation use whichever one you have tested. if we ever find any difference we can branch on that

joperezr:
@maloo do you think you'll have some time to incorporate @krwq's feedback into your PR? Just want to make sure that this won't go stale.

maloo:
> @maloo do you think you'll have some time to incorporate @krwq's feedback into your PR? Just want to make sure that this won't go stale.

Yes, sorry, have been "offline" for a while. I'm working with this right now in a product, so I definitely want this merged. I will try to finish this during my vacation July/Aug.

joperezr:
Sounds good, thanks a lot! I will go ahead and close the PR for now just so that we keep up only active PRs open, but please feel free to reopen this once you are ready to jump back to it and we will gladly take a look and get the changes in.

lloydjatkinson:
When this gets picked up again I'd definetly like to contribute to this - I've used these devices before and I identified a series of edge cases in the underlying `w1-gpio`/`w1-therm` kernel modules relating to the electrical nature of the 1 wire bus that results in incorrect/invalid readings. I compensated for this by handling it in the C# I wrote for it.

maloo:
I hope to pick it up during the next couple of weeks. For now you are welcome to add a pull request to https://github.com/maloo/iot/tree/w1ctrl where I will continue from.

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
maloo: Changed to Units.Temperature

src/devices/OneWire/OneWireController.cs
maloo: Added enum

maloo:
Moved to #793

krwq:
@maloo why not just force push instead of opening new PR?

PR #290, opened 2019-02-27
Update GpioButtonWait.cs

shaggygi:
Fixed the change to EventTypes.  DeviceApiTester build broke after the change with #279.

PR #291, opened 2019-02-27
update Si7021 binding

ZhangGaoxing:
1. use property
2. add Heater, Revision, Resolution
3. modify Temp and Humi get method
4. add and update README


ZhangGaoxing:
In this PR, I removed the method of converting Celsius to Fahrenheit. Normally, temperature sensors read data in Celsius. I wonder if we can provide a tool class for temperature conversion instead of the same method in every binding.

shaggygi:
@ZhangGaoxing just be aware builds will fail until #290 is updated.

src/devices/Si7021/README.md
```diff
@@ -1,21 +1,29 @@
-﻿# Using Si7021 (temperature and humidity sensor)
-
-## Summary
-
+﻿# Si7021
```
krwq: Perhaps `SI7021 - temperature and humidity` good call on removing stuff like `using` etc here

src/devices/Si7021/Resolution.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Si7021 Measurement Resolution
+    /// </summary>
+    public enum Resolution : byte
+    {
+        /// <summary>
+        /// Humidity 12-bit, Temperature 14-bit
+        /// </summary>
+        Resolution1 = 0b00,
```
krwq: I wish the only useful name I can think of wasn't super long 😄

src/devices/Si7021/Si7021.cs
```diff
@@ -3,59 +3,192 @@
...
         {
-            byte[] buffer = new byte[2];
+            Span<byte> readbuff = stackalloc byte[2];
 
             // Send temperature command, read back two bytes
-            _i2cDevice.WriteByte(ReadTemperatureCommand);
-            _i2cDevice.Read(buffer.AsSpan());
-            
+            _i2cDevice.WriteByte((byte)Register.SI_TEMP);
+            // wait SCL free
```
krwq: SCL?

src/devices/Si7021/samples/README.md
```diff
@@ -0,0 +1,36 @@
+# Example of Si7021
```
krwq: Example for

krwq: I'd change to `SI7021 - samples` though

src/devices/Si7021/Resolution.cs
ZhangGaoxing: Actually, I haven't come up with any suitable name. This Resolution can not be simply described by high and low.😂

src/devices/Si7021/samples/README.md
ZhangGaoxing: Aha, I found out that you didn't take a close look at README in my previous PRs. 😄

krwq: I usually just skim through READMEs if they have the right structure - they are easier to change after they are added vs the code which someone can start to depend on 😄 I'll try to look more but I'm currently the only one checking PRs and it takes quite a bit of my time

src/devices/Si7021/Si7021.cs
```diff
@@ -3,59 +3,192 @@
...
+            return Math.Round(humidity);
+        }
+
+        /// <summary>
+        /// Get Si7021 Firmware Revision
+        /// </summary>
+        /// <returns>Firmware Revision</returns>
+        private byte GetRevision()
+        {
+            Span<byte> writeBuff = stackalloc byte[2] { (byte)Register.SI_REVISION_MSB, (byte)Register.SI_REVISION_LSB };
```
krwq: slightly confused - you are asking for 2 bytes but reading 1?

ZhangGaoxing: Yes, you can see the `5.4. Firmware Revision` in the datasheet. Actually, this should be called command.😄

Issue #218, opened 2019-02-11
We should use project references instead of package references


krwq:
Recently we have broken the build but the check in which broke it was green on the CI. This could have been avoided if we have uses project-to-project (P2P) reference instead of package reference. Right now build will be broken only after the package with breaking changes is produced.

Technically we could use the the local nuget package source to be able to use package reference but arcade doesn't support it...

Previous reasoning to use package ref is to allow people to copy paste our projects as an example but I think developer workflow is more important and we can add comments to improve that.

joperezr:
Using the local NuGet package would be hard to accomplish. We could instead try using project references instead but it won't be as simple as just changing that, since the main library currently cross-compiles for different RIDs so you would have to specify a RuntimeIdentifier every time you have a P2P. We could solve this by adding some more complex infrastructure like the one corefx has in order to deal with P2Ps, but for now I would keep this issue open and in the backlog, and if we get enough number of hits or it really becomes annoying then we can go ahead and do all the infrastructure work in order to fix this.

krwq:
another build break because of this

JeremyKuhne:
This is proving to be a time sync, I think we should change it. We may end up being forced to build from source in the future anyway.

joperezr:
@krwq pointed out yesterday that now that we've moved to use Arcade for updating our dependencies on the Gpio library, our original desire of having sample projects be copy-pastable is broken so having P2Ps might be considered now. That said, the one good advantage that we still have by using packages is that we still use this as testing and validation that our package is well constructed.

Ellerbach:
This has been implemented everywhere in the repo. I'll close this one!

PR #292, opened 2019-02-27
Make it less likely to break build next time

krwq:
Fix build break and try to make it happen less frequently.

Current issue is that we use PackageReference in many projects. This is causing that the build is green locally and on the CI but when the official build produces new package it may delay break the build

Ref: https://github.com/dotnet/iot/issues/218

### DIFF
--- a/tools/DevicesApiTester/DeviceApiTester.csproj
+++ b/tools/DevicesApiTester/DeviceApiTester.csproj
@@ -12,7 +12,13 @@
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.3.0" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Scripting" Version="2.10.0" />
-    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
+    <!-- Normally package reference should be used: -->
+    <!-- <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" /> -->
+    <!-- We use project reference in our repository -->
+    <!-- so that our build breaks immediately and not on the next day when the official build happens -->
+    <ProjectReference Include="..\..\src\System.Device.Gpio\System.Device.Gpio.csproj">
+      <AdditionalProperties>RuntimeIdentifier=linux</AdditionalProperties>
+    </ProjectReference>
   </ItemGroup>
 
 </Project>


tools/DevicesApiTester/DeviceApiTester.csproj
```diff
@@ -12,7 +12,13 @@
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.3.0" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Scripting" Version="2.10.0" />
-    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
+    <!-- Normally package reference should be used: -->
+    <!-- <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" /> -->
+    <!-- We use project reference in our repository -->
+    <!-- so that our build breaks immediately and not on the next day when the official build happens -->
+    <ProjectReference Include="..\..\src\System.Device.Gpio\System.Device.Gpio.csproj">
+      <AdditionalProperties>Configuration=Linux-Debug;RuntimeIdentifier=linux</AdditionalProperties>
```
joperezr: No need to set Configuration, RuntimeIdentifier should be good enough.

krwq: I followed the pattern for existing projects - not sure why they did that but assuming it fixes some issue 😄 will do it though

JeremyKuhne: Because I wasn't getting it to work right in VS without it. YMMV.

joperezr: The Configuration part is only for VS if you had both projects loaded in the solution which you don’t in this case so that should not be needed

krwq: actually seems like a build breaks when I remove this - it cannot find many files - I will take a look at this

krwq: Actually - misread your comment and removed RuntimeIdentifier as well... why is this actually needed? can we do something that it is not?

krwq: is it because you reference rid specific project from rid=any? Seems kinda weird that I have to pass that by hand

Issue #175, opened 2019-01-30
Add bindings for RFID-RC522 (MF-RC522)


krwq:
This is currently WIP https://github.com/dotnet/iot/pull/293 - please pick up from here when implementing (PR is closed)

robertsundstrom:
@krwq I just started to port a module that I had previously contributed to another project, then I saw that you already had started making one. What is the status?

Anyway, here is my port: https://github.com/robertsundstrom/iot/commit/7cfba1f35dd5f5d5eb0b16347c4453eb2fab229d

It needs some polishing on the interface, as it is based on older code written in Java and Python.

krwq:
@robertsundstrom I have only barebones (the super low level stuff), reading UIDs from MIFARE tags and detecting presence of card (no authentication) but was mostly porting https://github.com/miguelbalboa/rfid with some improvement and the code is currently very far from pretty (although I have made it mostly non-allocating) - I still do not fully understand some of the pieces from that code (I only started working on this again today after not touching the code for 2 weeks). You can find the code in here (I usually push to the private fork first though but can push directly to public) - I didn't even call my project correctly at first (RfidRc522 instead of MFRC522): https://github.com/krwq/iot/tree/rfid_rc522/src/devices/RfidRc522

Your code seems much simpler than mine so what we can do is perhaps:
- figure out what APIs do we want
- combine both pieces of code together (assuming there is anything useful in mine which you don't have)

what I don't like in my version of code is that there is a lot of magic numbers all around and basically impossible to read without the spec (although yours is simpler and much easier to read still has many "magic" pieces).

I currently lack of higher level context on how exactly RFID tags are used. I could only think of reading tag IDs and I know that at least some tags support some form of authentication but have very little knowledge about how this exactly works... MFRC522 has fairly hard to read datasheet (at least for me) although have found some more interesting collection of datasheets in here: https://github.com/dalmirdasilva/ArduinoRadioFrequencyIdentification/tree/master/MifareClassic/datasheet but haven't had time to skim through that yet

krwq:
Seems https://github.com/dalmirdasilva/ArduinoRadioFrequencyIdentification/blob/master/ReaderMFRC522/datasheet/TagTypes.pdf has some more high level description on different types of tags. All tags have some kind of UID and some devices allow to change the data and then permanently lock the data. Depending on the tag type they have different protocols on what can be done on them with some common stuff.

[...]

PR #293, opened 2019-02-27
MFRC522

robertsundstrom:
Adding support for MFRC522 (RFID modules) modules.

- [X] Basic functionality for reading card UID
- [ ] Functionality for writing and reading card data
- [ ] FInalizing the interface

Based on code that I ported from Python and contributed to unosquare/raspberryio. I got inspiration from their component.

Data Sheet: https://www.nxp.com/docs/en/data-sheet/MFRC522.pdf

Related issue: #175


src/devices/Mfrc522/Command.cs
```diff
@@ -0,0 +1,13 @@
+namespace Iot.Device.Mfrc522
```
krwq: copyright header missing (add to all cs files)

src/devices/Mfrc522/Mfrc522Controller.cs
```diff
@@ -0,0 +1,447 @@
...
+using static Iot.Device.Mfrc522.Status;
+using static Iot.Device.Mfrc522.Register;
+using static Iot.Device.Mfrc522.RequestMode;
+using static Iot.Device.Mfrc522.Command;
+
+namespace Iot.Device.Mfrc522
+{
+    public class Mfrc522Controller : IDisposable
+    {
+        const byte NRSTPD = 22;
```
krwq: this needs rename - also any reason this is not configurable? I presume this is pin for hard reset?

src/devices/Mfrc522/Mfrc522Controller.cs
```diff
@@ -0,0 +1,447 @@
...
+        private GpioController _controller;
+
+        public Mfrc522Controller(SpiDevice spiDevice)
+        {
+            _spiDevice = spiDevice;
+
+            _controller = new GpioController();
+
+            _controller.OpenPin(NRSTPD, PinMode.Output);
+            _controller.Write(NRSTPD, PinValue.High);
```
krwq: this is set to high twice in a row (here and Init)

src/devices/Mfrc522/Mfrc522Controller.cs
```diff
@@ -0,0 +1,447 @@
...
+        }
+
+        private void WriteSpi(byte address, byte value)
+        {
+            Span<byte> buffer = stackalloc byte[2] {
+                (byte) ((address << 1) & 0x7E),
+                value
+            };
+            _spiDevice.Write(buffer);
+        }
```
krwq: nit: enter after the brace

src/devices/Mfrc522/Mfrc522Controller.cs
```diff
@@ -0,0 +1,447 @@
...
+
+        private void WriteSpi(Register register, Command command)
+        {
+            WriteSpi((byte)register, (byte)command);
+        }
+
+        private byte ReadSpi(byte address)
+        {
+            Span<byte> buffer = stackalloc byte[2] {
+                (byte) (((address << 1) & 0x7E) | 0x80),
```
krwq: create a constant for 0x80 (call it i.e. ReadMask)

src/devices/Mfrc522/Mfrc522Controller.cs
robertsundstrom: Yes. I’m aware of that. Planned to be fixed.

src/devices/Mfrc522/Mfrc522Controller.cs
```diff
@@ -0,0 +1,447 @@
...
+            }
+
+            WriteSpi(CommandReg, (byte)command);
+
+            if (command == Transceive)
+            {
+                SetBitMask(BitFramingReg, 0x80);
+            }
+
+            i = 2000;
```
krwq: check my changeset I shared with you - you can do this with Stopwatch instead of loop with magic number of iterations :smile:

src/devices/Mfrc522/Mfrc522Controller.cs
robertsundstrom: Not got around doing that yet.

src/devices/Mfrc522/Mfrc522Controller.cs
```diff
@@ -0,0 +1,447 @@
...
+
+            var i = 0;
+            if (status == OK)
+            {
+                i = 0;
+            }
+
+            if (backData.Length == 5)
+            {
+                while (i < 4)
```
krwq: I presume this is some kind of CRC calculation? if so please create a function from that

src/devices/Mfrc522/Mfrc522Controller.cs
```diff
@@ -0,0 +1,447 @@
...
+            byte i = 0;
+            while (i < pIndata.Length)
+            {
+                WriteSpi(FIFODataReg, pIndata[i]);
+                i++;
+            }
+
+            WriteSpi(CommandReg, (byte)CalculateCRC);
+            i = 0xFF;
+            while (true)
```
krwq: `for (int i = 0; i < 255; i++)`? Also any reason why 255 times?

src/devices/Mfrc522/Mfrc522Controller.cs
```diff
@@ -0,0 +1,447 @@
...
+
+            (status, backData, backLen) = SendCommand(Transceive, buf.ToArray());
+
+            if (status != OK || backLen != 4 || (backData[0] & 0x0F) != 0x0A)
+            {
+                throw new Mfrc522Exception("Failed to write to");
+            }
+        }
+
+        public void DumpClassic1K(byte[] key, byte[] uid)
```
krwq: Please use Span/ReadOnlySpan as arguments, also might be worth expanding the name a bit - I presume this refers to Mifare Classic?

src/devices/Mfrc522/Register.cs
```diff
@@ -0,0 +1,73 @@
+namespace Iot.Device.Mfrc522
+{
+    enum Register : byte
```
krwq: nit: explicit `internal`

src/devices/Mfrc522/Register.cs
```diff
@@ -0,0 +1,73 @@
...
+        Status2Reg = 0x08,
+        FIFODataReg = 0x09,
+        FIFOLevelReg = 0x0A,
+        WaterLevelReg = 0x0B,
+        ControlReg = 0x0C,
+        BitFramingReg = 0x0D,
+        CollReg = 0x0E,
+        Reserved01 = 0x0F,
+
+        Reserved10 = 0x10,
```
krwq: no need to add reserved registers - there is not much use from them

src/devices/Mfrc522/samples/Mfrc522.Sample.csproj
```diff
@@ -0,0 +1,20 @@
...
+  <PropertyGroup>
+    <LangVersion>7.1</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MSBuildThisFileDirectory)\..\Mfrc522.csproj" />
```
krwq: you can remove `$(MSBuildThisFileDirectory)\` from the path

src/devices/Mfrc522/samples/Mfrc522.sample.cs
```diff
@@ -0,0 +1,107 @@
...
+      Console.WriteLine($"Card UID: {string.Join(", ", cardUid)}");
+
+      mfrc522Controller.SelectTag(cardUid);
+
+      if (mfrc522Controller.Authenticate(RequestMode.Authenticate1A, 7, Mfrc522Controller.DefaultAuthKey, cardUid) == Status.OK)
+      {
+        var data = new byte[16 * 3];
+        for (var x = 0; x < data.Length; x++)
+        {
+          data[x] = (byte)(x + 65);
```
krwq: `x + 'A'` (+ cast if needed) so it is clearer what you're doing

src/devices/Mfrc522/samples/Mfrc522.sample.cs
```diff
@@ -0,0 +1,107 @@
...
+        return;
+
+      Console.WriteLine("Detected card");
+
+      var (status2, cardUid) = mfrc522Controller.AntiCollision();
+      Console.WriteLine($"Card UID: {string.Join(", ", cardUid)}");
+
+      mfrc522Controller.SelectTag(cardUid);
+
+      if (mfrc522Controller.Authenticate(RequestMode.Authenticate1A, 7, Mfrc522Controller.DefaultAuthKey, cardUid) == Status.OK)
```
krwq: 7?

joperezr:
@robertsundstrom do you still plan on addressing @krwq's feedback? Just want to make sure this PR won't go stale.

robertsundstrom:
> @robertsundstrom do you still plan on addressing @krwq's feedback? Just want to make sure this PR won't go stale.

robertsundstrom:
@joperezr I have not had the time. I would happily let someone else help me finish this.

Sorry, for accidentally closing and then reopening the PR.

joperezr:
@robertsundstrom no worries at all! We can leave this open for a bit more in case somebody wants to pick it up. If not, what we can do is to close this PR but to link it from the issue we have for adding this binding, and mention in there that a big part of the work was done by this PR so that any new work should go on top of the great work you provided here 😄

krwq:
@joperezr other option is we can merge as is since some of the functionality works but add an explicit comment in the README that this is WIP and put a list of checkboxes as in the PR description

joperezr:
I'm fine with us merging partial implementation of the device, but we should at least make sure that the partial implementation has passed the review. From what I can see, it looks like a lot of your comments on the PR haven't yet been addressed. Once those are addressed, I'm fine with merging and having the list of things yet to be supported on the README

krwq:
sounds good, @robertsundstrom do you think you will have time to at least address the feedback in this PR and update README? (fine if not)

robertsundstrom:
Yes. I'll have a look at it ASAP.

robertsundstrom:
@krwq I have had a look at the code again. There is a lot that has to be fixed. I would not accept it myself because of the design and incomplete implementation. I don't have my test setup running either.

How should we proceed? 

I have added the headers. Should I commit and push?

joperezr:
I see, thanks a lot for going and taking a look again. In that case, I would take the proposed idea of closing this PR and linking it as a WIP on the issue for anybody that wants to take on the task.

krwq:
Sounds good to me as well, thanks for investigating @robertsundstrom!

PR #299, opened 2019-03-01
Delete one of the copies of Mcp3008.sample.csproj

krwq:
This fixes the issue with the Arcade auto-update PR: https://github.com/dotnet/iot/pull/280

Currently we have two copies of Mcp3008.sample.csproj differing by casing:
- Mcp3008.sample.csproj
- Mcp3008.Sample.csproj

this has probably happened during some rename.


krwq:
Merging since it feels harmless.

PR #304, opened 2019-03-03
Added AGS01DB Binding

ZhangGaoxing:
MEMS VOC Gas Sensor ASG01DB

(BTW: This sensor was launched in the second half of last year, and I think adafruit will sell it soon. The price I bought was about two dollars. This is a very very cheap VOC sensor. 😄But I am sorry that I have not found the English version of the datasheet, because the manufacturer has not provided it for the time being.🙁)


src/devices/Ags01db/README.md
```diff
@@ -0,0 +1,31 @@
...
+    // read VOC concentration
+    double concentration = sensor.Concentration;
+}
+
+```
+
+## References
+Products Page in English: http://www.aosong.com/en/products-33.html
+
+Datasheet in English: [To be supplemented]()
```
joshfree: nit: dead link; I'm assuming you'll fix this in a supplement PR?

PR #309, opened 2019-03-04
Factor LCD interaction

JeremyKuhne:
Add a layer to allow abstracting the difference between using direct pins or built-in I2C support for lcd driver chips that support it.

Factoring set up to facilitate adding read functionality.

Clarify I2c behavior/flags and add multi-byte send overloads. Improves sample perf test from 2048ms to 660ms.

Factor setup code to include only appropriate steps for each interface.

cc: @SergeyRazmyslov


src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -16,46 +17,54 @@ namespace Iot.Device.CharacterLcd
...
+        /// headroom over the by-the-book waits.
+        /// 
+        /// This is more useful if you have a slow GpioAdapter as time spent in the
+        /// adapter may eat into the need to wait as long for a command to complete.
+        /// 
+        /// There is a busy signal that can be checked that could make this moot, but
+        /// currently we are unable to check the signal fast enough to make gains (or
+        /// even equal) going off hard timings. The busy signal also requires having a
+        /// r/w pin attached. And lastly, as already stated, it takes time to set up
+        /// for another byte of data to be pulsed in.
```
krwq: Is this expected to be set to value between 0..1 where 1.0 is the slowest possible display (the most compat) and some much lower value for faster displays or reverse (1+ and current value optimizes for most of the cases)? I think the summary should clarify that since this seems a bit confusing property.

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -68,181 +77,312 @@ public class Hd44780 : Hd44780Base, IDisposable
         /// <param name="backlightBrightness">The brightness of the backlight. 0.0 for off, 1.0 for on.</param>
         /// <param name="readWrite">The optional pin that controls the read and write switch.</param>
         /// <param name="controller">The controller to use with the LCD. If not specified, uses the platform default.</param>
```
krwq: the params descriptions are outdated

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -68,181 +77,312 @@ public class Hd44780 : Hd44780Base, IDisposable
...
+        /// Clears the LCD, returning the cursor to home and unshifting if shifted.
+        /// Will also set to Increment.
         /// </summary>
-        /// <param name="data">True to send data, otherwise sends a command.</param>
-        protected override void Send(byte value, bool data = false)
+        public void Clear()
         {
-            _controller.Write(_rsPin, data ? PinValue.High : PinValue.Low);
+            SendCommand(ClearDisplayCommand);
+            WaitForNotBusy(3000);
```
krwq: why 3000? Also 3000 seems to be used in more than one place, consider using a constant for it

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -68,181 +77,312 @@ public class Hd44780 : Hd44780Base, IDisposable
...
+        /// <param name="left">The column position from left to right starting with 0.</param>
+        /// <param name="top">The row position from the top starting with 0.</param>
+        public void SetCursorPosition(int left, int top)
+        {
+            int rows = _rowOffsets.Length;
+            if (top < 0 || top >= rows)
+                throw new ArgumentOutOfRangeException(nameof(top));
+
+            int newAddress = left + _rowOffsets[top];
+            if (left < 0 || (rows == 1 && newAddress >= 80) || (rows > 1 && newAddress >= 104))
```
krwq: I think the numbers should have some comment or reference to datasheet

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -68,181 +77,312 @@ public class Hd44780 : Hd44780Base, IDisposable
...
+            SendCommand((byte)(SetDDRamAddressCommand | newAddress));
+        }
+
+        /// <summary>
+        /// Enable/disable the display.
+        /// </summary>
+        public bool DisplayOn
+        {
+            get => (_displayControl & DisplayControl.DisplayOn) > 0;
+            set => SendCommand((byte)(value ? _displayControl |= DisplayControl.DisplayOn
```
krwq: consider using `[Flags]` enum and HasFlag and similar utilities for these

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -68,181 +77,312 @@ public class Hd44780 : Hd44780Base, IDisposable
...
+        /// book 5x7 and 5x10 is used (Advance Copy #AP4 from July 1985). Perhaps it was a
+        /// marketing change?
+        /// 
+        /// As only 11 bytes are used in 5x10 mode, but 16 bytes are reserved, the last 5 bytes
+        /// are useless. The datasheet helpfully suggests that you can store your own data there.
+        /// The same would be true for bits 5-7 of lines that matter for both 5x8 and 5x10.
+        /// </remarks>
+        /// <param name="location">Should be between 0 and 7</param>
+        /// <param name="characterMap">Provide an array of 8 bytes containing the pattern</param>
+        public void CreateCustomCharacter(byte location, params byte[] characterMap)
```
krwq: I see the `params` here but this feels like should be `ReadOnlySpan` since I think this in many cases will be static array with possibly all characters which you can slice on

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -68,181 +77,312 @@ public class Hd44780 : Hd44780Base, IDisposable
...
+        {
+            if (location > 7)
+                throw new ArgumentOutOfRangeException(nameof(location));
+
+            if (characterMap.Length != 8)
+                throw new ArgumentException(nameof(characterMap));
+
+            SendCommand((byte)(SetCGRamAddressCommand | (location << 3)));
```
krwq: some comment on 3

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -68,181 +77,312 @@ public class Hd44780 : Hd44780Base, IDisposable
...
+            SendData(characterMap);
         }
 
-        private void WriteBits(byte value, int count)
+        /// <summary>
+        /// Write text to display.
+        /// </summary>
+        /// <remarks>
+        /// There are only 256 characters available. There are chip variants
```
krwq: This is actually interesting. One one hand you want  simple API which takes string on the other hand this is actually is a byte array which partially maps to ASCII... I think this is fine as is but wondering if just for clarity we should not also have `ReadOnlySpan<byte>` overload which describes that this is actually custom character set 8-bit encoding which mostly overlaps with ASCII

src/devices/CharacterLcd/LcdInterface.I2c.cs
```diff
@@ -0,0 +1,105 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Buffers;
+using System.Device.I2c;
+using System.Diagnostics;
+
+namespace Iot.Device.CharacterLcd
+{
+    public abstract partial class LcdInterface : IDisposable
```
krwq: I think something like this (minus LCD specific stuff) would be useful in many other devices - we should look closely for similar patterns

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -16,46 +17,54 @@ namespace Iot.Device.CharacterLcd
...
+        /// This is more useful if you have a slow GpioAdapter as time spent in the
+        /// adapter may eat into the need to wait as long for a command to complete.
+        /// 
+        /// There is a busy signal that can be checked that could make this moot, but
+        /// currently we are unable to check the signal fast enough to make gains (or
+        /// even equal) going off hard timings. The busy signal also requires having a
+        /// r/w pin attached. And lastly, as already stated, it takes time to set up
+        /// for another byte of data to be pulsed in.
+        /// </remarks>
+        public double TimingMultiplier { get; set; } = 1.0;
```
krwq: I didn't see this being used in some of the initialization code later - is this intentional?

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: I changed the name to `WaitMultiplier` and moved to LcdInterface (it got disconnected).

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: Whoops- fixed.

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: Nope, was an oversight. Fixed.

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: It is a bit conservative. I've added more detailed comments on where this number comes from. (As well as all other timing numbers)

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: The row setup code explains the addressing in detail. I'll add a comment here.

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: DisplayControl is a `[Flags]` enum. `.HasFlag` had terrible [perf](https://blogs.msdn.microsoft.com/dotnet/2018/04/18/performance-improvements-in-net-core-2-1/) before 2.1 so I generally avoid it. I also find it weird as we don't have `SetFlag` or `ClearFlag`, so I lean towards doing everything with operators for consistency.

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: With `params byte[]` you can do this (and I do):

``` C#
            lcd.CreateCustomCharacter(0,
                0b_00110,
                0b_00110,
                0b_01100,
                0b_10111,
                0b_00100,
                0b_01110,
                0b_01010,
                0b_10001);
```

While it isn't _terrible_ to have a `new [] {}` wrapped around it, it is less clean. I don't see the harm in letting people write this way.

I'll add a `ReadOnlySpan<>` overload to allow slicing from a larger data source.

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: Sure- that is where the address goes in the command byte.

src/devices/CharacterLcd/Hd44780.cs
JeremyKuhne: There are a ton of character sets out there, including custom chips. I give quite a bit of detail in the remarks.  Taking a `byte` would require contortion more often than not- .NET uses `string` and `char`.  If you want to pass "raw" data you can do so, I just don't expect that to be common. When we get UTF-8 core types we should review all of our APIs that take strings.
 
Note that I'm not concerned about perf here as the character display is really quite small (80 characters at max).

src/devices/CharacterLcd/LcdInterface.I2c.cs
JeremyKuhne: I have been looking for patterns. So far I'm not finding enough commonality to warrant pulling anything up.

PR #312, opened 2019-03-05
Adding VL53L0X distance sensor

Ellerbach:
Adding support for VL53L0X
* Implementation of all main functions to initialize, calibrate, change the settings of the sensor
* Documentation and example
Please note that registers are almost not documented in the ST Microelectronics documentation. So there are quite a few magic numbers which can't be easily converted to registers as it's even unknow what they are doing and what is the size of the data sent.


src/devices/Vl53L0X/Info.cs
```diff
@@ -0,0 +1,20 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.Vl53L0X
+{
+    /// <summary>
+    /// Store the infromation regarding the sensor
+    /// </summary>
+    public class Info
```
ZhangGaoxing: Use struct instead of class. (and change the `Info` to `Information`?)

ZhangGaoxing: also add some comments to explain the properties

src/devices/Vl53L0X/Info.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.Vl53L0X
```
ZhangGaoxing: `Iot.Device.Vl53l0x` 😄

src/devices/Vl53L0X/PeriodPulse.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Iot.Device.Vl53L0X
+{
+    /// <summary>
+    /// The period used for the pulse when setting the 
+    /// VCSEL (vertical cavity surface emitting laser) pulse period
+    /// </summary>
+    public enum PeriodPulse
+    {
+        Period8 = 8,
```
ZhangGaoxing: please change `Period8` to `Period08` so that VS sorts the values better

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1114 @@
...
+using System.Buffers.Binary;
+using System.Device.I2c;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Threading;
+
+namespace Iot.Device.Vl53L0X
+{
+    public class Vl53L0X : IDisposable
```
ZhangGaoxing: Vl53l0x 😄

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1114 @@
...
+    {
+        /// <summary>
+        /// The default I2C Address, page 18 of the documentation 0x52 >> 1 = 0x29
```
ZhangGaoxing: >  page 18 of the documentation 0x52 >> 1 = 0x29

I don't think this sentence should be put in `<summary>`.

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1114 @@
...
+            var range = ReadUInt16((byte)Registers.RESULT_RANGE_STATUS + 10);
+            WriteRegister((byte)Registers.SYSTEM_INTERRUPT_CLEAR, 0x01);
+            return range;
+        }
+
+        /// <summary>
+        /// Get a distance in millimeters from the continous measurement feature.
+        /// It is recommended to used this method to gethigher quality measurements
+        /// </summary>
+        public UInt16 DistanceContinousMillimeters
```
ZhangGaoxing: I think you can add an enum type `MeasuringMode`, change the measurement by setting properity, rename `DistanceContinousMillimeters` to `Distance`.

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1114 @@
...
+            PerformSingleRefCalibration((byte)Registers.SYSRANGE_START);
+            WriteRegister((byte)Registers.SYSTEM_SEQUENCE_CONFIG, sequence_config);
+            return true;
+        }
+
+        /// <summary>
+        /// Set the type of precision needed for measurement
+        /// </summary>
+        /// <param name="precision">The type of precision needed</param>
+        public void SetPrecision(Precision precision)
```
ZhangGaoxing: please use properity😄

ZhangGaoxing:
> Please note that registers are almost not documented in the ST Microelectronics documentation. So there are quite a few magic numbers which can't be easily converted to registers as it's even unknow what they are doing and what is the size of the data sent.

Haha, the sensor datasheet does not provide any useful information. It's very strange. But ST provides 
software guide and libraries written in C. I wonder if you found it. This library has so much content that I only implemented continuous measurement in Win10 IoT last year. I've considered migrate it to. net core, but there are many registry addresses I don't understand, and the functions I implement is too simple. Now I'm glad to see someone implement it. I'll try it when the code merges.😄

ZhangGaoxing:
Also I see `Int16` and `Int32` in the code. I suggest `short`, `int` is better.😄

src/devices/Vl53L0X/Info.cs
Ellerbach: why not capital L and capital X? Asking for coherence with the rest

src/devices/Vl53L0X/Vl53L0X.cs
Ellerbach: ok, I'll move it to comments. Dépends of the PR and people commenting, sometimes, it is asked to be there, sometimes not. So we should aligne :-)

src/devices/Vl53L0X/Vl53L0X.cs
Ellerbach: Yes, good idea, will do it like that. I was hesitating but it will be much simpler.

Ellerbach:
> ide any useful information. It's very strange. But ST provides
> software guide and libraries written in C. I wonder if you found it. This library has so much content that I only implemented continuous measurement in Win10 IoT last year.

Yes I found it. Look at the initialization :-) it's even less readable than what I did! I named more than there is in the init file.

Ellerbach:
> I'll try it when the code merges.😄

Great! You can even try it before by pooling out the PR branch. Feedback more than welcome!

Ellerbach:
> Also I see Int16 and Int32 in the code. I suggest short, int is better.😄

I really found Int16, Int32 much more readable for specific usage than the short and int. But if more comments to have it renamed, I'll do it :-)

src/devices/Vl53L0X/Info.cs
krwq: @ZhangGaoxing there is very high likelyhood we will rename all projects with `x` to `X` so I'd leave as is (reason is because in case of `Bmx280` it looks like part of the device name and not like a placeholder)

src/devices/Vl53L0X/Information.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        public Version Revision { get; set; }
+        /// <summary>
+        /// The product ID
+        /// </summary>
+        public string ProductId { get; set; }
+        /// <summary>
+        /// Raw measurement of the signal rate fixed point 400 micrometers
+        /// </summary>
+        public UInt32 SignalRateMeasFixed1104_400_Micrometers { get; set; }
```
krwq: I'd recommend to make the raw readings internal unless there is a reason to make them public

src/devices/Vl53L0X/Information.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <summary>
+        /// Raw measurement of the distance measurement fixed point 400 micrometers
+        /// </summary>
+        public UInt32 DistMeasFixed1104_400_Micrometers { get; set; }
```
krwq: `UInt32` -> uint

src/devices/Vl53L0X/Information.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <summary>
+        /// Get the offset in micrometers
+        /// </summary>
+        public int OffsetMicrometers => SignalRateMeasFixed1104_400_Micrometers != 0 ? (((int)SignalRateMeasFixed1104_400_Micrometers - (400 << 4) * 1000) >> 4) * -1 : 0;
```
krwq: some comments explaining the formula would be useful

src/devices/Vl53L0X/README.md
```diff
@@ -0,0 +1,49 @@
+# VL53L0X Sensor
+
+## Summary
+
+The VL53L0X sensor is a Time-to-Flight sensor measuring precisely distances. The sensor allows you to get precise short distance measurement (from 5 millimeters to 2 meters) as well as long distance measurement (up to 8 meters but with a decreased precision). This sensor is a laser ranging sensor. So it is using laser pulses to measure the distances.
```
krwq: `So` feel out of place here - I'd just remove it

src/devices/Vl53L0X/README.md
```diff
@@ -0,0 +1,49 @@
...
+You can adjust the type of measurement you want. By default, the sensor is put into long range. This does allow to have a quite accurate (+-5%) precision within 2 meters but still can measure distances up to 8 meters (accuracy much less correct). You can switch to short range using the ```SetPrecision``` function. If you want to setup your own mode, you should use in conjunction the ```SetSignalRateLimit``` and the ```SetVcselPulsePeriod``` functions plus the ```HighResolution``` property like in the below example: 
+
+```csharp
+HighResolution = false;
+SetSignalRateLimit(0.25);
+SetVcselPulsePeriod(VcselType.VcselPeriodPreRange, PeriodPulse.Period14);
+SetVcselPulsePeriod(VcselType.VcselPeriodFinalRange, PeriodPulse.Period10);
+```
+
+Please refer to the documentation to understand the impact of changing the various pluses as well as using the high resolution precision measurement. The sensor can't be precise in long range and in general, the longer it can see, the less precise are the data. High resolution will return a more precise measurement but mainly in short distance.
```
krwq: `pluses` -> `pulses`

src/devices/Vl53L0X/SquadInfo.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Vl53L0X
+{
+    internal class SpadInfo
```
krwq: What's spad? https://en.wikipedia.org/wiki/Single-photon_avalanche_diode ?

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+        public int MaxTryReadSingle { get; set; }
+
+        /// <summary>
+        /// Set/Get high resolution measurement
+        /// </summary>
+        public bool HighResolution
+        {
+            get
+            {
+                return _highResolution;
```
krwq: does it make sense to not have any cache and read this value from register?

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+            // Set longer range
+            Precision = Precision.LongRange;
+        }
+
+        /// <summary>
+        /// The sensor can be changed for other I2C Address, this function allows to change it
+        /// </summary>
+        /// <param name="i2cDevice">The current I2C Device</param>
+        /// <param name="newAddress">The new I2C Address from 0x00 to 0x7F</param>
+        static public void ChangeI2cAddress(I2cDevice i2cDevice, byte newAddress)
```
krwq: `static public` -> `public static`

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+        {
+            if (newAddress > 0x7F)
+                throw new ArgumentException($"{nameof(newAddress)} can't exceed 0x7F");
+            try
```
krwq: enter above `try`

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+            {
+                i2cDevice.Write(new byte[] { (byte)Registers.I2C_SLAVE_DEVICE_ADDRESS, newAddress });
+            }
+            catch (IOException ex)
+            {
+
```
krwq: extra enter

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+                return ReadContinuousMeasrurementMillimeters();
+            }
+        }
+
+        /// <summary>
+        /// Get a distance in millimeters
+        /// </summary>
+        /// <param name="multipleReads">True if you want multiple try to get a clean value</param>
+        /// <returns>Returns the maximum range so 8190</returns>
+        public UInt16 GetDistanceSingleMillimeters(bool multipleReads = false)
```
krwq: The name feels a bit weird - perhaps something along the lines of `GetDistanceOnce` and document the unit (we will very likely we will add a Distance struct in the future so this will avoid more changes)

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+                }
+                else
+                {
+                    return value;
+                }
+                return goodCount != 0 ? (UInt16)(average / goodCount) : (UInt16)OperationRange.OutOfRange;
+            }
+            catch (IOException)
+            {
+                return (UInt16)OperationRange.OutOfRange;
```
krwq: mixed feelings about this but fine for now until we got units more stabilized

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+        /// Set the VCSEL (vertical cavity surface emitting laser) pulse period for the
+        /// given period type (pre-range or final range) to the given value in PCLKs.
+        /// Longer periods seem to increase the potential range of the sensor.
+        /// Valid values are (even numbers only):
+        /// pre:  12 to 18 (initialized default: 14)
+        /// final: 8 to 14 (initialized default: 10)
+        /// based on official API
+        /// </summary>
+        /// <param name="type">The type of VCSEL</param>
+        /// <param name="periodPclks">The period paert of the supported periods. Be aware periods are a bit different depending on the VCSEL you are targetting.</param>
```
krwq: `paert` -> `part1

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+                }
+            }
+            // Write back the SPAD configuration
+            ReferenceSpadMap[0] = (byte)Registers.GLOBAL_CONFIG_SPAD_ENABLES_REF_0;
+            _i2cDevice.Write(ReferenceSpadMap);
+
+            // Those commands come from the offical API C code from STM
+            // Documentation does not provide register names
+            // In the main API, this block of command are hard coded
+            // The names of Registers provided are based on know ones.
```
krwq: bummer but doesn't seem like we can do much about it without spending a lot of time trying to to dig where this info comes from

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+            {
+                case VcselType.VcselPeriodPreRange:
+                    return DecodeVcselPeriod(ReadByte((byte)Registers.PRE_RANGE_CONFIG_VCSEL_PERIOD));
+                case VcselType.VcselPeriodFinalRange:
+                    return DecodeVcselPeriod(ReadByte((byte)Registers.FINAL_RANGE_CONFIG_VCSEL_PERIOD));
+                default:
+                    // Should not arrive
+                    return byte.MaxValue;
+            }
+
```
krwq: nit: extra enter

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+        /// <returns>the period encoded</returns>
+        private byte EncoreVcselPeriod(byte periodPclks) => (byte)((periodPclks >> 1) - 1);
+
+        /// <summary>
+        /// Decode sequence step timeout in MCLKs from register value 
+        /// format: (LSByte * 2^MSByte) + 1
+        /// </summary>
+        /// <param name="valueToDecode"></param>
+        /// <returns>The decoded value</returns>
+        private UInt32 DecodeTimeout(int valueToDecode) => (UInt32)(((valueToDecode & 0x00FF) << ((valueToDecode & 0xFF00) >> 8)) + 1);
```
krwq: should this return TimeSpan instead?

src/devices/Vl53L0X/Vl53L0X.cs
```diff
@@ -0,0 +1,1163 @@
...
+                if (stopWatch.ElapsedMilliseconds > expirationMilliseconds)
+                    return false;
+            }
+
+            // Set interrupt and get ready for measurement
+            WriteRegister((byte)Registers.SYSTEM_INTERRUPT_CLEAR, 0x01);
+            WriteRegister((byte)Registers.SYSRANGE_START, 0x00);
+            return true;
+        }
+        public void Dispose()
```
krwq: enter above

src/devices/Vl53L0X/Vl53L0X.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease.19078.2" />
```
krwq: `0.1.0-prerelease*`

src/devices/Vl53L0X/Information.cs
Ellerbach: will do

src/devices/Vl53L0X/Information.cs
Ellerbach: I really prefer UInt32 as it's always super clear. But I'll replace everywhere.

src/devices/Vl53L0X/Information.cs
Ellerbach: It's coming from the official API and they' Don't explain the formula. It's just the formula, Nothing else. So I'll just add a reference that it's coming from the official API

src/devices/Vl53L0X/SquadInfo.cs
Ellerbach: Yes as defined line 749 in the comments for the ```GetSpadInfo()``` function

src/devices/Vl53L0X/Vl53L0X.cs
Ellerbach: Yes, in this situation. At initialization the register is changed and the internal value is set. Later, when you change again the resolution, same thing, the internal value is updated. Looks like there is no other way to change the resolution than writting them. 
And as the documentation is not perfect, the reference I found is writting it and not reading the register.
In all the cases, the main code will be the master on this sensor and it won't be changed by anything else. 
With all that, I think it's the best way to handle this.

src/devices/Vl53L0X/Vl53L0X.cs
Ellerbach: Yes, I can change the name. There are 2 modes, continous and single reading. So I kept close to them. But I can clearly rename it like you propose. 
OK to remove the unit and document it then.

src/devices/Vl53L0X/Vl53L0X.cs
Ellerbach: The sensor has a maximum range and returns it when object is too far or when there is an issue

src/devices/Vl53L0X/Vl53L0X.cs
Ellerbach: I've already renamed some and for the rest, the original API is just a large array of bytes. So unfortunately, we'll have to keep it like this. If at some point there is a new documentation, a new API version, with a better documentation, we'll be able to adjust.

src/devices/Vl53L0X/Vl53L0X.cs
Ellerbach: No, what is used is the raw value after. So better keep it raw. TimeSpan will add complexity and it's not necessary in this case.

PR #314, opened 2019-03-05
Buzzer

Veujin:
Adding piezo buzzer binding


src/devices/Buzzer/Buzzer.cs
```diff
@@ -0,0 +1,82 @@
...
+        private readonly int _buzzerPin;
+        private readonly int _pwmChannel;
+        private readonly PwmController _pwmController;
+
+        /// <summary>
+        /// Create Buzzer class instance with output on specified pin with specified channel.
+        /// </summary>
+        /// <param name="pinNumber">The GPIO pin number in case of a software PWM. The chip in case of a hardware PWM.</param>
+        /// <param name="pwmChannel">The channel to use in case of a hardware PWM.</param>
+        public Buzzer(int pinNumber, int pwmChannel)
```
krwq: I'd recommend instead of catching exceptions etc just take PwmController and channel as an input.

src/devices/Buzzer/Buzzer.cs
```diff
@@ -0,0 +1,82 @@
...
+                _pwmController = new PwmController(new SoftPwm(true));
+                _pwmController.OpenChannel(_buzzerPin, _pwmChannel);
+            }
+        }
+
+        /// <summary>
+        /// Start playing tone of specific frequency.
+        /// </summary>
+        /// <param name="frequency">Tone frequency in Hertz.</param>
+        public void PlayTone(double frequency)
```
krwq: I'm wondering if this should be renamed to SetFrequency - reason is that PlayTone suggest you might be able to play 2 frequencies at the same time while this will override previously played frequency

krwq: (no strong opinion either way - just something to consider)

src/devices/Buzzer/Buzzer.cs
joshfree: having both ctors would be nice

src/devices/Buzzer/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Device Family
+
+This binding was tested on two types of piezo buzzers. First type of buzzer has two pins *vcc* and *gnd*. Second type of buzzers has addition *signal* pin.
+  
+
+
+## Binding Notes
+
+The  `Buzzer`  class can use either software either hardware PWM. This is done fully transparently by the initialization.
```
joshfree: "can use either software **either** hardware PWM" -> "can use either software **or** hardware PWM"

src/devices/Buzzer/README.md
```diff
@@ -0,0 +1,58 @@
+﻿
+# Buzzer
+
+  
+
+## Summary
+
+This device binding allow to play certain tone using piezo buzzer. It uses PWM with 50% duty cycle and various frequencies.
```
joshfree: "allow to play certain tone using piezo buzzer" -> "allow**s playing** certain tone**s** using piezo buzzer"

src/devices/Buzzer/Buzzer.cs
Veujin: Added new constructor with three parameters next to old one.

src/devices/Buzzer/Buzzer.cs
Veujin: Thanks. Good suggestion. Renamed to `SetFrequency`. Also renamed `StopTone` to `StopPlaying`.

src/devices/Buzzer/README.md
Veujin: Fixed.

src/devices/Buzzer/README.md
Veujin: Also fixed.

PR #318, opened 2019-03-08
Added DHT12 Binding

ZhangGaoxing:
Temperature and Humidity Sensor DHT12

DHT12, unlike DHT11/22, is an I2C protocol sensor. I think this PR and the existing `Dhtxx` must have one that needs to be modified. Let us discuss.😄


ZhangGaoxing:
DHT12 reads and processes data differently from existing Dhtxx. I recommend renaming Dhtxx.

ZhangGaoxing:
I thought about it carefully and decided to merge the code into Dhtxx.

krwq:
@ZhangGaoxing I was just gonna ask since I felt I've similar code before when I just started reviewing :smile: We need to start thinking about some listing of the devices...

ZhangGaoxing:
@krwq @shaggygi 
I tested the existing DHT code, and my sensors didn't work. So I refer to the [code](https://github.com/ZhangGaoxing/windows-iot-demo/blob/master/src/DHT11/03_Source/DHT11Demo/DHT11.cs) I wrote before, which works well on Win10 IoT. But embarrassingly, these codes don't work on Linux right now. I don't know why. I tested DHT11, DHT12(1-wire) on Arduino to make sure my sensors are not broken. If you have these sensors, could you help me test it or find out some mistakes?

Now, DHT12 I2C protocol ensures work.

shaggygi:
@ZhangGaoxing are you saying you can communicate with the device via I2C on Win IoT, but not Linux?  Or is it some conversion logic that is blowing up?  I would try doing some basic tests to see if you can at least read a register/address.  Then we could try to narrow down the issue.

ZhangGaoxing:
I think there may be some problems with my words...😅

I mean that I2C works well, 1-wire reads abnormal (including existing DHTxx code).

krwq:
Hey @ZhangGaoxing is this ready to take `WIP` label off? :smile:

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -3,43 +3,35 @@
...
     {
-        private const int MAX_TIME = 85;
-        private const uint MAX_WAIT = 255;
-        private byte[] _dht11Val = new byte[5];
-        private int _pin;
-        private DhtType _dhtType;
+        /// <summary>
+        /// DHT12 Default I2C Address
+        /// </summary>
+        public const byte Dht12DefaultI2cAddress = 0x5C;
```
krwq: should this be just `DefaultI2cAddress` or are other Dhts different? If different I'd still rename to `DefaultI2cAddressDh12` so that they are sorted in such a way that it is easier to find

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -3,43 +3,35 @@
...
-            long initialTick = _stopwatch.ElapsedTicks;
-            long initialElapsed = _stopwatch.ElapsedMilliseconds;
-            double desiredTicks = milliseconds / 1000.0 * Stopwatch.Frequency;
-            double finalTick = initialTick + desiredTicks;
-            while (_stopwatch.ElapsedTicks < finalTick)
-            {
-                //nothing than waiting
-            }
-        }
+        private int _lastMeasurment = 0;
```
krwq: nit: missing `e` in variable name

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -52,21 +44,27 @@ private void Wait(double milliseconds)
...
-            return ret;
+            get
+            {
+                ReadData();
+
+                Temperature temperature = Temperature.FromCelsius(double.MaxValue);
+                switch (_dhtType)
+                {
+                    case DhtType.DHT11:
+                        temperature = Temperature.FromCelsius(GetTempDht11());
```
krwq: you'll save couple of lines of code if you do

```csharp
               switch (_dhtType)
                {
                    case DhtType.DHT11:
                        return Temperature.FromCelsius(GetTempDht11());
                    // ...
                   default:
                        return Temperature.FromCelsius(double.MaxValue);
```

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -75,116 +73,174 @@ public bool TryGetTemperature(out Temperature temperature)
         /// <remarks>
         /// If last read was not successfull, it returns double.MaxValue
         /// </remarks>
-        public double Humidity => (_dhtType == DhtType.Dht11) ? GetHumidityDht11() : GetHumidityDht22();
+        public double Humidity
```
krwq: starts to feel like we should have a base class (no strong opinion yet but still looking at the PR)

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -75,116 +73,174 @@ public bool TryGetTemperature(out Temperature temperature)
...
 
         /// <summary>
-        /// Start a reading
+        /// Read through One-Wire
         /// </summary>
         /// <returns>
         /// <c>true</c> if read is successfull, otherwise <c>false</c>.
         /// </returns>
-        public bool ReadData()
+        private bool ReadThroughOneWire()
```
krwq: would be nice to merge this with: https://github.com/dotnet/iot/pull/289 eventually (not in this PR though)

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -195,28 +251,32 @@ public bool ReadData()
...
+        private double GetHumidityDht22() => IsLastReadSuccessful ? _readBuff[0] + _readBuff[1] * 0.1 : double.MaxValue;
 
+        /// <summary>
+        /// Cleanup
+        /// </summary>
         public void Dispose()
         {
-            _controller.ClosePin(_pin);
+            _controller?.Dispose();
+            _sensor?.Dispose();
```
krwq: they should not be null unless you nullify them :smile:

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -195,28 +251,32 @@ public bool ReadData()
...
                 // if MSB = 1 we have negative temperature
-                return ((_dht11Val[2] & 0x80) == 0 ? temp : -temp);
+                return ((_readBuff[3] & 0x80) == 0 ? temp : -temp);
             }
             else
                 return (double.MaxValue);
         }
 
-        private double GetHumidityDht22() => IsLastReadSuccessful ? (double)((_dht11Val[0] << 8) | _dht11Val[1]) * 0.1F : double.MaxValue;
+        private double GetHumidityDht22() => IsLastReadSuccessful ? _readBuff[0] + _readBuff[1] * 0.1 : double.MaxValue;
```
krwq: please add some comment on 0.1 (or make it a const with descriptive name)

src/devices/Dhtxx/DhtType.cs
```diff
@@ -9,7 +9,15 @@ namespace Iot.Device.DHTxx
     /// </summary>
     public enum DhtType
     {
-        Dht11,
-        Dht22
+        DHT11,
```
krwq: `Dht11`, `Dht12`...

src/devices/Dhtxx/README.md
```diff
@@ -1,66 +1,37 @@
...
-#endif
-```
-
-This value is used to wait 1 microsecond in a for simple loop. This value is platform dependent.
-
-```csharp
-for (byte wt = 0; wt < waitMS; wt++)
-    ;
-```
+The DHT temperature and humidity sensors are very popular. This project support DHT11, DHT12, DHT21(AM2301), DHT22(AM2302).
```
krwq: nit: supports

src/devices/Dhtxx/samples/DhtSensor.sample.cs
```diff
@@ -16,27 +13,19 @@ static void Main(string[] args)
     {
         Console.WriteLine("Hello DHT!");
 
-        DHTSensor dht = new DHTSensor(26, DhtType.Dht22);
+        // Init DHT12 through I2C
```
krwq: would be nice to put this in a separate function and just comment out the call so that we at least know this builds

src/devices/Dhtxx/samples/README.md
```diff
@@ -1,91 +1,40 @@
...
-
-Note: Humidity is in relative percentage air humidity. Humidity is a value between 0.0 and 100.0. 100.0 represents 100% humidity in the air.
-
-Here is a full example:
+* SCL - SCL
+* SDA - SDA
+* VCC - 5V
+* GND - GND
 
+## Code
```
krwq: I'd recommend always putting extra enter before and after triple ticks (otherwise it sometimes breaks down markdown)

src/devices/Dhtxx/DhtSensor.cs
krwq: ok after finishing I think it's fine as is - I was expecting more switches later

src/devices/Dhtxx/DhtSensor.cs
ZhangGaoxing: What if Dispose() is called many times?😄

ZhangGaoxing: If initialized with i2c, gpio is null, and on the contrary....

src/devices/Dhtxx/DhtType.cs
ZhangGaoxing: The name of this kind of sensor is DHTxx. I don't think it's necessary to use Dhtxx.😄

ZhangGaoxing:
@krwq  
Please do not merge this code first, because there is still a problem with 1-wire reading. I have some questions. I initialized the sensor using `DelayHelper.DelayMicroseconds()`. Because initialization needs to be delayed by 30 microseconds, can our runtime be accurate to 30 or smaller microseconds? `DelayHelper.DelayMicroseconds()` is effective in how many microseconds?

I'm going to do some more testing until 1-wire works.

krwq:
@ZhangGaoxing I think this depends on the board - for Raspberry Pi I think this is currently around 200-500ns but occasionally you may get a hickup (i.e. GC or threading context switch) since this is not real time OS.

For 1-wire I recommend to create seperate device binding using SysFs but if this is not currently supported and you cannot get that to work I'm fine with removing this part of the code

src/devices/Dhtxx/DhtType.cs
krwq: I personally prefer to use exact casing as you did above but our [coding guidelines](https://github.com/dotnet/iot/blob/master/Documentation/Coding-guidelines.md) is fairly clear about using `Dht` :( Once we have enough evidence this doesn't work for this repo we can reiterate :smile:

ZhangGaoxing:
@krwq I think we can continue this PR. The 1-wire protocol is working, and I tested it with DHT11 and DHT12.
I think some of the code still needs to improve. The code now returns `double.MaxValue` if it reads data incorrectly, which is not friendly on the interface. Do you have any good suggestions?😄
360截图20190318220019159

krwq:
@ZhangGaoxing change it to return NaN instead. initially it was MaxValue so it sorts better but both are easy to filter out and NaN is clearer that it is invalid value

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -154,41 +156,82 @@ private void ReadData()
...
             // wait 20 - 40 microseconds
-            DelayHelper.DelayMicroseconds(30, true);
+            DelayHelper.DelayMicroseconds(40, true);
 
-            _controller.SetPinMode(_pin, PinMode.Input);
+            _controller.SetPinMode(_pin, PinMode.InputPullUp);
 
-            // DHT corresponding signal - LOW - 80 microseconds
+            // DHT corresponding signal - LOW - about 80 microseconds
+            count = _loopCount;
```
krwq: any chance you could use similar logic DelayHelper is doing here? I'd prefer to avoid adding magic  loops since they will likely not work correctly if we add new board. If it doesn't work please at least add a comment what board this was measured on and if you use Debug or Release build

krwq: also if you're leaving as is also add a comment in the README file that I2C was tested only on `<board name here>` and might not work elsewhere and perhaps open an issue tracking that

ZhangGaoxing: This loop is considered from the timing of DHT, just to prevent deadly loops caused by time chaos. `_loopCount` does not need to be precise. And it can't wait directly with DelayHelper, because this 80 microseconds is also about, not exactly accurate.

ZhangGaoxing: Sensor timing can be affected by many factors, such as wire, voltage, temperature, humidity and so on. So actually I don't like using 1-wire protocol very much. Now most of the new sensors produced by the company are I2C protocols.😄

JeremyKuhne:
> can our runtime be accurate to 30 or smaller microseconds

I believe +/- 1.5 microseconds is the best you can hope for. The resolution of the clock is 1us on the Pi I believe. We're also limited by the overhead to actually get the ticks.

krwq:
@ZhangGaoxing I think given above it makes sense to try to convert magic loops even if they don't need to be accurate

Issue #183, opened 2019-01-31
Cannot read from output pin?


JinShil:
This exception doesn't seem right to me:

https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs#L171-L174

Is this a mistake, or is it my understanding that needs correction?

shaggygi:
I’m not sure the exact issue you’ve encountered.  Could you provide an example?  Basically, this code checks and if the pin is an output, it throws to state you’re reading a output pin.  You can only read input pins.

JinShil:
> You can only read input pins.

Yes, that's the issue.  Why can you only read input pins?  I've never used a microcontroller or any other embedded device where reading an output pin was not permitted.  What's the justification for it?

JinShil:
To illustrate a use case, one may want to read the output pin when the program first starts in order to update their UI with the output's last state.  Initializing the pin to a default value may violate the software's requirements.

joperezr:
I'm not sure that all drivers would be able to support this. For example, the RaspberryPi driver operates with registers in the memory directly, and there are different bits in memory that are used for the value depending if you are in Input vs Output mode, and I would really doubt that the Input bit will be updated when the pin is set to output and the value is changed.

JinShil:
> there are different bits in memory that are used for the value depending if you are in Input vs Output mode, and I would really doubt that the Input bit will be updated when the pin is set to output and the value is changed.

I would expect the driver to return the value of the Output bit if it is in Output mode.

joperezr:
That is not how the Output bits work on the RPi driver unfortunately. There are two registers that are used to control an output pin's value in the RPi: 

https://github.com/dotnet/iot/blob/c1a3493abc77e6988e8625e2c46b3ea35eac3a91/src/System.Device.Gpio/Interop/Unix/Libc/Interop.mmap.cs#L44-L50

The way this works internally, is that if you want to set the pin value to low then you flip a bit of the GPCLR register, and when you want to set it high then you flip a bit on the GPSET register, and by modifying one register, the other one won't be modified. What this means, is that only internally does the raspberry pi knows which was the register that was last modified in that bit, which would tell it whether the value is high or low. There is now way that by reading those registers one would be able to tell whether the value is high or low.

JinShil:
Doesn't GPLEV give you the state of the output pin?

[...]

PR #323, opened 2019-03-10
Expose Raspberry register addresses

tarekgh:
This is need for the perf critical scenarios

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -127,7 +127,7 @@ namespace System.Device.Gpio.Drivers
             ValidatePinNumber(pinNumber);
 
             /*
-             * There are two registers that contain the value of a pin. Each hold the value of 32 
+             * There are two registers that contain the value of a pin. Each hold the value of 32
              * different pins. 1 bit represents the value of a pin, 0 is PinValue.Low and 1 is PinValue.High
              */
 
@@ -220,14 +220,14 @@ namespace System.Device.Gpio.Drivers
              * This is the process outlined by the BCM2835 datasheet on how to set the pull mode.
              * The GPIO Pull - up/down Clock Registers control the actuation of internal pull-downs on the respective GPIO pins.
              * These registers must be used in conjunction with the GPPUD register to effect GPIO Pull-up/down changes.
-             * The following sequence of events is required: 
-             * 
+             * The following sequence of events is required:
+             *
              * 1. Write to GPPUD to set the required control signal (i.e.Pull-up or Pull-Down or neither to remove the current Pull-up/down)
-             * 2. Wait 150 cycles – this provides the required set-up time for the control signal 
+             * 2. Wait 150 cycles – this provides the required set-up time for the control signal
              * 3. Write to GPPUDCLK0/1 to clock the control signal into the GPIO pads you wish to modify
              *    – NOTE only the pads which receive a clock will be modified, all others will retain their previous state.
-             * 4. Wait 150 cycles – this provides the required hold time for the control signal 
-             * 5. Write to GPPUD to remove the control signal 
+             * 4. Wait 150 cycles – this provides the required hold time for the control signal
+             * 5. Write to GPPUD to remove the control signal
              * 6. Write to GPPUDCLK0/1 to remove the clock
              */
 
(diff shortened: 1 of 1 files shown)


tarekgh:
CC @krwq @JeremyKuhne @joperezr @joshfree

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -313,6 +313,14 @@ protected internal override void Write(int pinNumber, PinValue value)
             *registerPointer = register;
         }
 
+        public void GetRegisters(out ulong* setRegister, out ulong* clearRegister)
```
krwq: hmm, wondering if there is perhaps way to not expose pointers i.e.:

```csharp
public ulong SetRegister
{
  get { *_registerViewPointer->GPSET; }
  set {  *_registerViewPointer->GPSET = value; }
}

public ulong ClearRegister
{
  get { *_registerViewPointer->GPCLR; }
  set {  *_registerViewPointer->GPCLR = value; }
}
```

+ perhaps call to `Initialize()` (or just move it to constructor)

shaggygi: > I think it was from @shaggygi
I don't recall anything related, but probably slept since then :smile: 
The only thing I can remember when it comes to registers is the reading output pin issue (#183)

tarekgh: @krwq the only concern is we used to call Initialize() before when opening the Pin. I am not sure if we call Initialize() in the constructors will have any side effect. There should be a good reason we delay the initialization but I don't know what is it. if you are sure there is no any side effect, I can go with your idea.

tarekgh: The other idea is to not do the initialization at all and anyone calling these registers properties has to open the pins (and that will do the initialization)

krwq: @joperezr if you happen to read this - got any clues about the `Initialize` being delayed? cc: @buyaa-n

JeremyKuhne:
I would want to see actual perf numbers before we open up "unsafe" APIs. We should also be comparing against piping the multi-pin apis from the controllers to the drivers:

https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs#L309

My _assumption_ is that most of the overhead is in multiple calls to setting/getting the registers, not validating/prepping inputs.

tarekgh:
@JeremyKuhne what @krwq proposed is no unsafe any more and I am inclining to his proposal. and as I mentioned to you before, in my case it made a 200x time faster than using the current APIs. Piping multi-pin apis are good but it can be address in some other PR and I think it shouldn't be expose on the level of the driver. in other word, we'll need to discuss more about such API, you may start having a proposal and we go from there.

JeremyKuhne:
@tarekgh, @krwq We already have an API for passing multiple values. It simply isn't piped all the way down, and should be. We shouldn't be adding yet another way to do things that doesn't apply to all drivers until we've finished that and have data that compares against our existing solution.

I spoke with @joperezr at length about using bitmasks instead of the API that we did introduce. He was against using them and for good reasons:

1. Not all devices will fit into 32 bits, let alone 64 bits (or even 128)
2. Pin numbers aren't sequential and have multiple address schemes
3. Using such APIs are difficult and error prone, even without 1 & 2

Let's get our existing API fully implemented and then look very carefully at performance data to understand precisely what the remaining overhead is and where it is. I'm happy to dig into other APIs once we have that data.

tarekgh:
@krwq and @JeremyKuhne I have updated the PR according to our conversation. Please let me know if you have any more feedback. Thanks

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -313,12 +314,20 @@ protected internal override void Write(int pinNumber, PinValue value)
             *registerPointer = register;
         }
 
-        public void GetRegisters(out ulong* setRegister, out ulong* clearRegister)
+        protected virtual ulong SetRegister
         {
-            Initialize();
+            get { return *(ulong*)(_registerViewPointer->GPSET); }
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
krwq: only setter needs it and not getter? 😄

tarekgh: I added to the setter and also I removed the virtual specifier to ensure the in-lining will occur.

PR #327, opened 2019-03-12
Make package reference to compiler toolset opt-in feature

safern:
@tarekgh said he was having issues when using a test project outside the repo that is referencing the Gpio project as a ProjectReference. So the Gpio project will use the package compilers but his project a different version of the compiler causing problems when loading Compilers tasks. With this property people can opt-in to use the Sdk built-in compilers, restore and build their project that references Gpio and not hit any issues.

cc: @krwq @joperezr

### DIFF
--- a/eng/Compilers.props
+++ b/eng/Compilers.props
@@ -7,6 +7,6 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


tarekgh:
Thanks @safern

PR #328, opened 2019-03-12
Batch builds on commit triggers

safern:
cc: @tarekgh @krwq @joperezr


safern:
> I suppose this doesn’t change official builds since they have the trigger overwritten right?

Actually it does, because as per @tarekgh and @krwq request we disabled the scheduled triggers and wanted to try the per commit triggers. Which is what every repo is doing now, even corefx and we haven’t hit any problems with having a per commit official build it actually has helped to have a faster package flow to the world

joperezr:
I see, I don’t see why we would need to produce more than 2 official builds daily, but I guess it makes sense if all the other repos are doing it as well

PR #329, opened 2019-03-13
Added support for apds9300 device

garciaolais:
add  apds9300 ambient light sensor (ALS) and proximity sensor


src/devices/Apds9930/Apds9930.cs
```diff
@@ -0,0 +1,337 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+//Ported from https://github.com/Depau/python-apds9930/blob/master/apds9930/__init__.py
```
krwq: this is a bit awkward but this seems to be GPL v2 (https://github.com/Depau/python-apds9930/blob/master/LICENSE)

which I believe allows porting but only if it is also GPL v2 which I do not believe we can take. cc: @richlander

garciaolais: 😢 I'm really sorry to hear that

krwq: let's see with @richlander perhaps there is solution (but we were usually advised not to even look at GPL to avoid any legal issues :cry:) - perhaps would be fine to put it as a sample on GPL but doubt we can put one class in the same package on different license than the rest of the code so it will be sample or we will unfortunately have to close :-(

richlander: We cannot take GPL-related samples. While they are not a big problem for hobbyists, they are for companies who use the code in this repo. We don't want them to have to get their legal staff involved to use the code we produce for them.

Is there a non-GPL implementation for you to port? If so, you can port that instead, but you need to delete the code you've written and then wait at least 30 days to start that. This is a "cooling off period". These are the rules for managing between difference license types.

This [page](https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/contributing.md#copying-files-from-other-projects) talks about our rules for contributions. I'm sorry we didn't make that more obvious.

I'm sorry that you used up a bunch of time on this.

We can continue to discuss this, but we need to close this PR for the reasons given.

garciaolais: Thank you for your feedback I closed this PR  and I will delete the ported code,  😓  Software licenses are very complex
> 
> Is there a non-GPL implementation for you to port? 
There is a few ones ... 
MIT - [https://github.com/makecode-extensions/APDS9930](https://github.com/makecode-extensions/APDS9930)
Apache 2.0 - [https://github.com/N-Storm/autolight](https://github.com/N-Storm/autolight)
>

krwq: @garciaolais MIT is 100% fine, Apache 2.0 I think is also fine but might need some extra steps in 3rd party notices - (I'd go with MIT)

garciaolais: Great! , 30 Days it is enough time for working in this

krwq: Thanks @garciaolais and sorry for the inconvenience!

joperezr: @garciaolais just poking the PR since it looks like 30 days have passed, so it should be fine now to port the MIT library 😄

garciaolais: @joperezr I hope send the PR very soon 😄

Issue #202, opened 2019-02-02
SpiConnectionSettings should have a property for data flow


shaggygi:
This proposal is to add a data flow property to the SpiConnectionSettings class.

## Rationale and Usage
While most SPI devices us Most Significant Bit (MSB), there are devices that can use Least Significant Bit (LSB). To overcome this today (to my current knowledge) is adding additional software logic to reverse the bits during reads/writes.

## Proposed Changes
The proposed change is to add a new property for data flow.

This could be named `DataLsbFirst` or similar. It could have a default value of false that represents LSB since this is the minority use.  It also doesn't have to be included in the `SpiConnectionSettings` constructor as is similar to other properties like `ClockFrequency`, `Mode`, etc..

```csharp
var spiConnectionSettings = new SpiConnectionSettings(0, 0)
{
    ClockFrequency = 1000000,
    DataLsbFirst= true,
    Mode = SpiMode.Mode0
};
```

## References
- I commented on similar topic in another proposal [Provide a new GpioSpiDevice](https://github.com/dotnet/iot/issues/122).

## Open Questions
1. Windows 10 IoT appears to only allow for MSB.  Is there a way to change this in a system setting or within Windows.Devices.Spi API?
2. It appears you can change this in Linux, but I have not tested.  See https://01.org/linuxgraphics/gfx-docs/drm/driver-api/spi.html.  Need to review if this can be performed programmatically or must be hardcoded in SPI settings.

krwq:
@shaggygi just for reference, do you perhaps know of any specific devices needing this?

shaggygi:
@krwq Reviewing this appears most cases are MSB first, but similar to #201... many processors, APIs and tools appear to be configurable.  So this is one of those areas we might want to support sooner than later.

I came across the following that mentioned it supports LSB first, but didn't dig too much of the details:

- [PS2 Controller](http://roboticsforevery1.blogspot.com/2014/08/interfacing-playstation-controller-with.html)
- [WTV020](https://os.mbed.com/forum/mbed/topic/2170/?page=1#comment-10992)
- [Arduino](https://arduino.stackexchange.com/questions/16348/how-do-you-use-spi-on-an-arduino) This is more of an example for the board and API that allows you to specify the flow.

krwq:
I've opened a PR for both please take a look - naming to be discussed in the PR - I'm not particular fan of name DataFlow but on the other hand I could not come up with anything better. Python calls it `FirstBit` which does not seem good either.

[...]

Issue #201, opened 2019-02-02
SpiConnectionSettings should have a property for chip select active state


shaggygi:
This proposal is to add a chip select active state property to the SpiConnectionSettings class.

## Rationale and Usage
While most SPI devices are selected with an active-low chip select pin, there are devices that are active-high.  To overcome this today (to my current knowledge) is adding additional hardware logic to inverse the select line between master/slave devices.

## Proposed Changes
The proposed change is to add a new property for a chip select active state.

- This could be named `ChipSelectActiveHigh` or similar.  It could have a default value of `false` that represents active-high since this is the minority use.
- It also doesn't have to be included in the `SpiConnectionSettings` constructor as is similar to other properties like `ClockFrequency`, `Mode`, etc..  

```csharp
var spiConnectionSettings = new SpiConnectionSettings(0, 0)
{
    ChipSelectActiveHigh = true,
    ClockFrequency = 1000000,
    Mode = SpiMode.Mode0
};
```

## References
- I commented on similar topic in another proposal [Provide a new GpioSpiDevice](https://github.com/dotnet/iot/issues/122).

## Open Questions
1. Windows 10 IoT appears to only allow for active-low state.  Is there a way to change this in a system setting or within Windows.Devices.Spi API?
2. It appears you can change this in Linux, but I have not tested.  See https://01.org/linuxgraphics/gfx-docs/drm/driver-api/spi.html.  Need to review if this can be performed programmatically or must be hardcoded in SPI settings.

joshfree:
@krwq could you please investigate for 3.0?

krwq:
@shaggygi got any specific devices in mind? (just curious)

shaggygi:
@krwq I came across this when trying to use with bit-banging GPIO pins.  You can use SPI with a 74595 logic chip.  While I have a partial binding started for this chip, it could be used with something like [I2C/SPI Character LCD Backpack](https://www.adafruit.com/product/292) using SPI (with active-high CS).

While reviewing this topic, I found that many processors allow CS enable to be configured to Low or High.  I also came across various SPI tools where they can be configured and blogs/forums where users state they have a device that needs to be active-high enabled.  However, there was never a specific device name mentioned where I could verify the CS interface.

I agree most cases will be active-low CS as that appears to be the preferred method.  I'm just thinking this option might need to be available sooner than later.  Hope this helps.

joperezr:
[...]

PR #331, opened 2019-03-14
Add support for SPI data flow and chip select active state on linux

krwq:
Fixes: https://github.com/dotnet/iot/issues/201
Fixes: https://github.com/dotnet/iot/issues/202

cc: @tarekgh @JeremyKuhne @shaggygi @ZhangGaoxing

This enables using devices which define CS active as high and makes it much easier to use devices which use LSB data flow (no need to invert bits)

### DIFF
--- a/src/System.Device.Gpio/System/Device/Spi/DataFlow.cs
+++ b/src/System.Device.Gpio/System/Device/Spi/DataFlow.cs
@@ -5,7 +5,7 @@
 namespace System.Device.Spi
 {
     /// <summary>
-    /// Defines order in which data is send over the wire.
+    /// Specifies order in which bits are transferred first on the SPI bus.
     /// </summary>
     public enum DataFlow
     {
--- a/src/System.Device.Gpio/System/Device/Spi/DataFlow.cs
+++ b/src/System.Device.Gpio/System/Device/Spi/DataFlow.cs
@@ -1,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Device.Spi
+{
+    /// <summary>
+    /// Defines order in which data is send over the wire.
+    /// </summary>
+    public enum DataFlow
+    {
+        MsbFirst,
+        LsbFirst,
+    }
+}
--- a/src/System.Device.Gpio/System/Device/Spi/DataFlow.cs
+++ b/src/System.Device.Gpio/System/Device/Spi/DataFlow.cs
@@ -9,7 +9,13 @@ namespace System.Device.Spi
     /// </summary>
     public enum DataFlow
     {
+        /// <summary>
+        /// Most significant bit will be sent first (most of the devices use this value).
+        /// </summary>
         MsbFirst,
+        /// <summary>
+        /// Least significant bit will be sent first.
+        /// </summary>
         LsbFirst,
     }
 }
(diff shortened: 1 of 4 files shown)


src/System.Device.Gpio/System/Device/Spi/DataFlow.cs
```diff
@@ -0,0 +1,15 @@
...
+// See the LICENSE file in the project root for more information.
+
+namespace System.Device.Spi
+{
+    /// <summary>
+    /// Defines order in which data is send over the wire.
+    /// </summary>
+    public enum DataFlow
+    {
+        MsbFirst,
```
tarekgh: could you please add Xml comments for these here? I know Most/Least significant bits but this may be not obvious for some people

krwq: @tarekgh most of the people should not change this - when they need to they will likely know what that means :-) Will add though

src/System.Device.Gpio/System/Device/Spi/DataFlow.cs
```diff
@@ -0,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Device.Spi
+{
+    /// <summary>
+    /// Defines order in which data is send over the wire.
+    /// </summary>
+    public enum DataFlow
```
shaggygi: I left a comment on the issue, but I'll leave here just in case.

Found a [reference ](https://python-periphery.readthedocs.io/en/latest/spi.html)where a Python implementation names it BitOrder.  Not sure I'm set on any of these so far.  Just throwing it out for discussion.

src/System.Device.Gpio/System/Device/Spi/Drivers/Windows10SpiDevice.Windows.cs
```diff
@@ -23,6 +24,11 @@ public class Windows10SpiDevice : SpiDevice
         /// </param>
         public Windows10SpiDevice(SpiConnectionSettings settings)
         {
+            if (settings.DataFlow != DataFlow.MsbFirst || settings.ChipSelectLineActiveState != PinValue.Low)
+            {
+                throw new PlatformNotSupportedException($"Changing {nameof(settings.DataFlow)} or {nameof(settings.ChipSelectLineActiveState)} options is not supported on the current platform.");
```
shaggygi: I had a question related to this and when a platform doesn't support something another does.  Is this a feature the team might ping the Windows team to possibly add in a future release (and WinRT APIs)?  Just curious.

src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
```diff
@@ -49,5 +51,13 @@ public SpiConnectionSettings(int busId, int chipSelectLine)
         /// The frequency in which the data will be transfered.
         /// </summary>
         public int ClockFrequency { get; set; }
+        /// <summary>
+        /// Specifies which bits will be transfered first over the wire.
+        /// </summary>
+        public DataFlow DataFlow { get; set; } = DataFlow.MsbFirst;
```
shaggygi: NIT: I prefer more like "... transferred first on the SPI bus.".  We use similar wording in other SPI/I2C files.

src/System.Device.Gpio/System/Device/Spi/Drivers/Windows10SpiDevice.Windows.cs
krwq: We can send request to Windows team but we cannot force or promise they will do it. This seems like a simple API addition though and useful so I suspect we might be able to get it. I think in order to avoid spamming them with various requests we should create one issue which summarizes them all and I can send that forward to them.

shaggygi: Should we create an issue to log some of these for future review?  A couple of others that come to mind are possible CAN and I2S.

PR #332, opened 2019-03-14
Easy CLI workflow on linux and Raspberry PI running Raspbian

krwq:
Sharing some simple script for interacting with Raspberry PIs from CLI.

See README.md file for usage.

Side effect that on your Raspbian you will now have published binaries under:

```
/home/pi/src/iot-partial/
```

(I usually don't change user name on PI and just allow asymmetric key access so never bothered to change that)

### DIFF
--- a/tools/scripts/helpers-rpi-linux.sh
+++ b/tools/scripts/helpers-rpi-linux.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+
+# This file is supposed to be sourced when in iot repo i.e.:
+# source helpers-rpi-linux.sh
+# or just
+# . helpers-rpi-linux.sh
+# while you are in the root folder of your repo
+
+# is-git-repo <dir>
+is-git-repo() {
+    test -d "$1/.git"
+}
+
+# git-find-repo-root <dir>
+git-find-repo-root() {
+    dir=$1
+    if [ "$dir" != "" ]; then
+        if is-git-repo $dir; then
+            echo $dir
+        else
+            parent_dir=`dirname $dir`
+            git-find-repo-root $parent_dir
+        fi
+    fi
+}
+
+repo_root="$(git-find-repo-root "$(pwd)")"
+export PATH=$repo_root/.dotnet:$PATH
+export DOTNET_MULTILEVEL_LOOKUP=0
+
+find-csproj() {
+    basename "$(find . -maxdepth 1 -type f -name "*.csproj")"
+}
+
+find-project-name() {
+    find-csproj | sed 's/\.[^\.]*$//'
+}
+
+get-pi() {
+    if [ $# -eq 0 ]; then
+        if [ -z "${rpi}" ]; then
+            echo 'publish <name>' >&2
+            echo 'or export rpi=<name>' >&2
+            return 1
+        else
+            echo "$rpi"
+        fi
+    else
+        echo "$1"
+    fi
+}
+
+publish() {
+    name="$(get-pi "$@")"
+    if [ $? -ne 0 ]; then
+        return 1
+    fi
+
+    projname=`find-project-name`
+    if [ -z "${projname}" ]; then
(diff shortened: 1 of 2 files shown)


tools/scripts/helpers-rpi-linux.sh
```diff
@@ -0,0 +1,112 @@
...
+        return 1
+    fi
+
+    projname=`find-project-name`
+    if [ -z "${projname}" ]; then
+        echo 'Project file not found.' >&2
+        return 1
+    fi
+
+    dir=/home/pi/src/iot-partial/$projname
```
wfurt: Should this be part of the rpi variable? This makes fixed assuptions about project layout.

tools/scripts/helpers-rpi-linux.sh
```diff
@@ -0,0 +1,112 @@
...
+    exe=$dir/$projname
+
+    if ssh $name test -e $exe; then
+        # -t - pseudo-terminal allocation
```
wfurt: I don't understand relation between -t and zombies? AFAIK, the pseudo terminal allows you to use some more of System.Console functions but it should not really have impact on process management.

tools/scripts/helpers-rpi-linux.sh
```diff
@@ -0,0 +1,112 @@
...
+    projname=`find-project-name`
+    if [ -z "${projname}" ]; then
+        echo 'Project file not found.' >&2
+        return 1
+    fi
+
+    dir=/home/pi/src/iot-partial/$projname
+    exe=$dir/$projname
+
+    if ssh $name test -e $exe; then
```
wfurt: Do we really need this as separate execution? You may simply try to execute and report back return code. 
Also the -e tests presence but that still may not be executable.

tools/scripts/helpers-rpi-linux.sh
```diff
@@ -0,0 +1,112 @@
...
+    dir=/home/pi/src/iot-partial/$projname
+    exe=$dir/$projname
+
+    echo 'Publishing...'
+    dotnet publish -v q -r linux-arm /nologo || return 1
+
+    ssh $name mkdir -p $dir
+    rsync -r ./bin/Debug/netcoreapp2.1/linux-arm/publish/ $name:$dir || return 1
+
+    ssh $name chmod +x $exe &>/dev/null
```
wfurt: Do we really need this? If so I would suggest to change flags before rsync to avoid extra ssh session. 
Each ssh command can possibly ask you for typing password.

tools/scripts/helpers-rpi-linux.sh
```diff
@@ -0,0 +1,112 @@
...
+        return 1
+    fi
+
+    dir=/home/pi/src/iot-partial/$projname
+    exe=$dir/$projname
+
+    echo 'Publishing...'
+    dotnet publish -v q -r linux-arm /nologo || return 1
+
+    ssh $name mkdir -p $dir
```
wfurt: I think rsync should be able to create directories as needed.

tools/scripts/helpers-rpi-linux.sh
krwq: @wfurt - without `-t` my process was not always killed when I `Ctrl-C` over ssh. With `-t` I never got any issues

tools/scripts/helpers-rpi-linux.sh
krwq: never thought about people using password but that's a good point

tools/scripts/helpers-rpi-linux.sh
wfurt: ^C really does not do much on Unix. It is terminal control (-t) what converts that sequence to SIGINT. Without it, it is job for sshd or init to kill and cleanup process group.

krwq:
closing this as I haven't had time to finish it. Will re-open whenever I got time

PR #333, opened 2019-03-16
Moved FSR binding PR into sample

buyaa-n:


samples/force-sensitive-resistor/FsrWithAdcSample.cs
```diff
@@ -0,0 +1,77 @@
...
+
+using System;
+using System.Threading;
+using Iot.Device.Mcp3008;
+
+namespace force_sensitive_resistor
+{
+    class FsrWithAdcSample
+    {
+        private int _pinNumber = 0; // pin number to which FSR output connected
```
krwq: I'd remove the word `output` here since it depends on which side are you on (from perspective of the driver this will be input)

samples/force-sensitive-resistor/FsrWithAdcSample.cs
```diff
@@ -0,0 +1,77 @@
...
+        private int _resistance { get; set; } = 10_000; // kOhm
+        private int _voltageSupplied { get; set; } = 3_300;  // 3300mV = 3.3V
+
+
```
krwq: extra enter

samples/force-sensitive-resistor/FsrWithAdcSample.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            // This sample used Mcp3008 ADC which analog voltage read output ranges from 0 to 1023 (10 bit) 
+            // mapping it to corresponding milli voltage, update output range if you use different ADC
+            return _voltageSupplied * readValue / 1023;
+        }
+
+        private int CalculateFsrResistance(int fsrVoltage)
+        {
+            // Formula: FSR = ((Vcc - V) * R) / V
+            if (fsrVoltage > 0)
```
krwq: why do you need to check if > 0? negative voltage is fine too

samples/force-sensitive-resistor/FsrWithAdcSample.cs
```diff
@@ -0,0 +1,77 @@
...
+                {
+                    force = fsrConductance - 1000;
+                    force /= 30;
+                }
+                return force;
+            }
+            return 0;
+        }
+
+        public void StartReading() {
```
krwq: brace on new line

samples/force-sensitive-resistor/FsrWithAdcSample.cs
```diff
@@ -0,0 +1,77 @@
...
+            return 0;
+        }
+
+
+        private int CalculateForce(int resistance)
+        {
+            if (resistance > 0)
+            {
+                int force;
+                int fsrConductance = 1_000_000 / resistance; // in micro ohms
```
krwq: the units don't match - if you divide by resistance which is already in ohms then 1M would have to be in ohms squared which doesn't make much sense

samples/force-sensitive-resistor/FsrWithCapacitorSample.cs
```diff
@@ -0,0 +1,61 @@
...
+                count++;
+                if (count == 30000)
+                {   // if count goes too high it means FSR in highest resistance which means no pressure, do not need to count more
+                    break;
+                }
+            }
+            return count;
+        }
+
+        public void StartReading()
```
krwq: this would be more useful if it didn't print and instead returned bool (i.e. IsPressed)

samples/force-sensitive-resistor/FsrWithCapacitorSample.cs
```diff
@@ -0,0 +1,61 @@
...
+        private long ReadCapacitorChargingDuration()
+        {
+            int count = 0;
+            _controller.SetPinMode(_pinNumber, PinMode.Output);
+            // set pin to low
+            _controller.Write(_pinNumber, PinValue.Low);
+            // Prepare pin for input and wait until high read
+            _controller.SetPinMode(_pinNumber, PinMode.Input);
+
+            while (_controller.Read(_pinNumber) == PinValue.Low)
```
krwq: can you connect it in such a way that it will be low when it is pressed and high when it is not pressed and ignore the delays completely?

samples/force-sensitive-resistor/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine("Hello Fsr408 capacitor Sample!");
+            // Use this sample when using ADC for reading
+/*            FsrWithAdcSample fsrWithAdc = new FsrWithAdcSample();
+            fsrWithAdc.StartReading();
+*/
+            // Use this sample if using capacitor for reading
+            FsrWithCapacitorSample fsrWithCapacitor = new FsrWithCapacitorSample();
```
krwq: I think all printing should happen here

samples/force-sensitive-resistor/FsrWithAdcSample.cs
buyaa-n: read value would never be negative => voltage would never be negative => if it is negative that means something wrong

samples/force-sensitive-resistor/FsrWithAdcSample.cs
buyaa-n: Yeah it is confusing, conductance = 1/R i.e. inverse of resistance, here 2 calculations made, getting inverse of resistance and converting into micro ohms (1/μΩ), updated comment

samples/force-sensitive-resistor/Program.cs
buyaa-n: ok

samples/force-sensitive-resistor/FsrWithCapacitorSample.cs
buyaa-n: That actually could work just for determining on/off, but still people could use the count number for determining how much it is pressed ...

krwq: what count? it should be 0 or 1 since you read digital pin?

buyaa-n: The count measuring how long it took to get 1. Yes it is digital input and can be used directly without counting or delay to get 0/1, but that count could be useful for measuring pressure intensity, really depend on usage

samples/force-sensitive-resistor/FsrWithAdcSample.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using System;
+using Iot.Device.Mcp3008;
+
+namespace force_sensitive_resistor
+{
+    class FsrWithAdcSample
+    {
+        private int _resistance = 10_000; // kOhm
+        private int _voltageSupplied = 3_300;  // 3300mV = 3.3V\
```
krwq: extra `\` (last character on the line)

samples/force-sensitive-resistor/FsrWithAdcSample.cs
```diff
@@ -0,0 +1,68 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Iot.Device.Mcp3008;
+
+namespace force_sensitive_resistor
```
krwq: ForceSensitiveResistor

samples/force-sensitive-resistor/FsrWithAdcSample.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public FsrWithAdcSample()
+        {
+            // Create a ADC convertor instance you are using depending how you wired ADC pins to controller
+            // in this example used ADC Mcp3008 with "bit-banging" wiring method.
+            // please refer https://github.com/dotnet/iot/tree/master/src/devices/Mcp3008/samples for more information
+            _adcConvertor = new Mcp3008(18, 23, 24, 25);
+        }
+
+        public int CalculateVoltage(int readValue)
```
krwq: I think this should return double - especially because you divide by 1023

samples/force-sensitive-resistor/FsrWithAdcSample.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+            return 0;
+        }
+
+        public int CalculateForce(int resistance)
+        {
+            if (resistance > 0)
+            {
+                int force;
+                int fsrConductance = 1_000_000 / resistance; // conductance is inverse of resistance which converted to micro ohms (1/??)
```
krwq: remove the comment since it's confusing

samples/force-sensitive-resistor/FsrWithAdcSample.cs
```diff
@@ -0,0 +1,68 @@
...
+
+                // Use the two FSR guide graphs to approximate the force
```
krwq: where do you get the guide graphs from?

samples/force-sensitive-resistor/FsrWithCapacitorSample.cs
krwq: why not use stopwatch for this? The `count` term is confusing since it actually usually is referred as ticks. This function should return double and the unit should be clear (second? millisecond? microsecond?)

samples/force-sensitive-resistor/Program.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine("Hello Fsr408 capacitor Sample!");
+            // Use this sample when using ADC for reading          
+            StartReadingWithADC();
+
+            // Use this sample if using capacitor for reading
+            // StartReadingWithCapacitor();
+
```
krwq: nit: extra enter

samples/force-sensitive-resistor/FsrWithAdcSample.cs
buyaa-n: I am actually followed naming structure of existing samples https://github.com/dotnet/iot/blob/master/samples/led-blink/Program.cs#L9

samples/force-sensitive-resistor/FsrWithAdcSample.cs
buyaa-n: ok

samples/force-sensitive-resistor/FsrWithAdcSample.cs
buyaa-n: ok

samples/force-sensitive-resistor/FsrWithAdcSample.cs
buyaa-n: Graph is in the datasheet page 6 https://cdn-learn.adafruit.com/assets/assets/000/010/126/original/fsrguide.pdf.
But I didn't really invented this formula, adafruit sample for same FSR https://learn.adafruit.com/force-sensitive-resistor-fsr/using-an-fsr used below calculation devised from the graph

samples/force-sensitive-resistor/FsrWithCapacitorSample.cs
buyaa-n: stopwatch is not appropriate because the thread sleeps in the calling app https://github.com/dotnet/iot/pull/333/files/33f81b8a125095dab07cf203d9d139ed2ad72d29#diff-145d59a2af55d0554553d3f750daf878R55 which causing interruption between stopwatch count and capasitor charging count which leads to invalid result.

PR #337, opened 2019-03-17
Auto-generate devices listing

krwq:
Improve list of devices and make it so that it is auto-regenerated when `dotnet run` is run on the `tools/device-listing`.

Currently it is based on the first line of the README.md file in a device description.

Note: markdown does not display custom tags which I use in here and therefore they can be used as a placeholder for where generated content can be.

<custom>I.e. this text is in the custom tag which you cannot see (owners of this repo can edit this post to see) </custom>


tools/device-listing/device-listing.cs
```diff
@@ -0,0 +1,127 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.Text;
+using System.Threading;
+
+namespace Iot.Device.device_listing.Samples
```
tarekgh: >device_listing [](start = 21, length = 14)

could you use something like Listing or DeviceListing? 
Also, the word Device became redundant here.

tools/device-listing/device-listing.cs
```diff
@@ -0,0 +1,127 @@
...
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            string repoRoot = FindRepoRoot(Environment.CurrentDirectory);
```
tarekgh: >repoRoot [](start = 19, length = 8)

what happen if FindRepoRoot returns null?

tools/device-listing/device-listing.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            if (!string.IsNullOrEmpty(dir))
+            {
+                if (Directory.Exists(Path.Combine(dir, ".git")))
+                {
+                    return dir;
+                }
+                else
+                {
+                    return FindRepoRoot(new DirectoryInfo(dir).Parent.FullName);
```
tarekgh: >FindRepoRoot [](start = 27, length = 12)

should we ensure we are not going outside the repo root?

krwq: I'm not sure what do you mean.. If we don't find ".git" folder we will return null

tools/device-listing/device-listing.cs
krwq: forgot to change it

tools/device-listing/device-listing.cs
krwq: unless you meant that Parent is null when in root dir - this is now fixed

tarekgh: you have fixed this in https://github.com/dotnet/iot/pull/337/commits/4c506bb76782bad6db6ceeddb79789bacd19920b :-)

joperezr:
Thanks for adding this! @krwq should we have this automatically run as part of a build so that the list gets updated every PR that is adding a new binding?

krwq:
@joperezr we can run this separately and auto create PRs whenever there is any diff if that is easy to add. I was also initially planning to add tiny bit more there but haven't had time (categorization by type of sensor and stuff)

PR #346, opened 2019-03-25
Add a Method to select default GPIO bank

microhobby:
Some SoCs have more than one GPIO bank. For these we gave more than one entry in /dev/: /dev/gpiochip0, /dev/gpiochip1 ... /dev/gpiochipX

This PR add and implement a method to select the default GPIO bank to be used on the GpioController.
If the hardware has only one GPIO bank (/dev/gpiochip0 like the rpi) it will be automatically set in
the GpioController constructor. But if the hardware has more than one GPIO bank the bank to be used
for GpioController operations should be defined.

I made a test on a Toradex Colibri iMX7D, i.MX 7 has 7 GPIO banks, accessing two lines from different GPIO banks:

```c#
using System;
using System.Threading;
using System.Device.Gpio.Drivers;
using System.Device.Gpio;

namespace core_test
{
    class Program
    {
        {
            Console.WriteLine(".net iot");

            var pin1 = 1;
            var pin17 = 17;
            var lightTimeInMilliseconds = 200;
            var dimTimeInMilliseconds = 200;

            using (GpioController gpiochip0 = new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver()),
                        gpiochip3 = new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver(3)))
            {
                // open /dev/gpiochip0 offset 1
                gpiochip0.OpenPin(pin1, PinMode.Output);
                Console.WriteLine($"GPIO pin enabled for use: {pin1}");

                // open /dev/gpiochip3 offset 17
                gpiochip3.OpenPin(pin17, PinMode.Output);
                Console.WriteLine($"GPIO pin enabled for use: {pin17}");

                Console.CancelKeyPress += (object sender, ConsoleCancelEventArgs eventArgs) =>
                {
                    gpiochip0.Write(pin1, PinValue.Low);
                    gpiochip0.ClosePin(pin1);

                    gpiochip3.Write(pin17, PinValue.Low);
                    gpiochip3.ClosePin(pin17);

                    Console.WriteLine("Bye...");

                    gpiochip0.Dispose();
                    gpiochip3.Dispose();
                };

                while (true)
                {
                    if (gpiochip0.IsPinOpen(pin1) && gpiochip3.IsPinOpen(pin17)) {
                        Console.WriteLine($"Light for {lightTimeInMilliseconds}ms");
                        gpiochip0.Write(pin1, PinValue.High);
                        gpiochip3.Write(pin17, PinValue.Low);
                        Thread.Sleep(lightTimeInMilliseconds);
                        Console.WriteLine($"Dim for {dimTimeInMilliseconds}ms");
                        gpiochip3.Write(pin17, PinValue.High);
                        gpiochip0.Write(pin1, PinValue.Low);
                        Thread.Sleep(dimTimeInMilliseconds);
                    }
                }
            }
        }
    }
}
```

### DIFF
--- a/src/Native/System.Device.Gpio.Native/libgpiod.c
+++ b/src/Native/System.Device.Gpio.Native/libgpiod.c
@@ -189,30 +189,6 @@ extern int ReadEventForLine(struct gpiod_line *line)
 }
 
 /**
- * @brief List all GPIO chips.
- * @return The number of GPIO chips, -1 if an error occurred.
- */
-extern int GetNumberOfChips()
-{
-    struct gpiod_chip_iter *iter;
-    struct gpiod_chip *chip;
-    int sum = 0;
-
-    iter = gpiod_chip_iter_new();
-    if (!iter)
-        return -1;
-
-    gpiod_foreach_chip(iter, chip) 
-    {
-        sum++;
-    }
-
-    gpiod_chip_iter_free(iter);
-
-    return sum;
-}
-
-/**
  * @brief Open a gpiochip by number.
  * @return GPIO chip pointer handle or NULL if an error occurred.
  */
--- a/src/Native/System.Device.Gpio.Native/libgpiod.c
+++ b/src/Native/System.Device.Gpio.Native/libgpiod.c
@@ -194,19 +194,20 @@ extern int ReadEventForLine(struct gpiod_line *line)
  */
 extern int GetNumberOfChips()
 {
-	struct gpiod_chip_iter *iter;
-	struct gpiod_chip *chip;
-	int sum = 0;
+    struct gpiod_chip_iter *iter;
+    struct gpiod_chip *chip;
+    int sum = 0;
 
-	iter = gpiod_chip_iter_new();
-	if (!iter)
-		return -1;
-	
-	gpiod_foreach_chip(iter, chip) {
-		sum++;
-	}
+    iter = gpiod_chip_iter_new();
+    if (!iter)
+        return -1;
 
-	gpiod_chip_iter_free(iter);
+    gpiod_foreach_chip(iter, chip) 
+    {
+        sum++;
+    }
 
-	return sum;
+    gpiod_chip_iter_free(iter);
+
+    return sum;
 }
--- a/src/Native/System.Device.Gpio.Native/libgpiod.c
+++ b/src/Native/System.Device.Gpio.Native/libgpiod.c
@@ -187,3 +187,26 @@ extern int ReadEventForLine(struct gpiod_line *line)
 	}
 	return response;
 }
+
+/**
+ * @brief List all GPIO chips.
+ * @return The number of GPIO chips, -1 if an error occurred.
+ */
+extern int GetNumberOfChips()
+{
+	struct gpiod_chip_iter *iter;
+	struct gpiod_chip *chip;
+	int sum = 0;
+
+	iter = gpiod_chip_iter_new();
+	if (!iter)
+		return -1;
+	
+	gpiod_foreach_chip(iter, chip) {
+		sum++;
+	}
+
+	gpiod_chip_iter_free(iter);
+
+	return sum;
+}
--- a/src/Native/System.Device.Gpio.Native/libgpiod.c
+++ b/src/Native/System.Device.Gpio.Native/libgpiod.c
@@ -211,3 +211,12 @@ extern int GetNumberOfChips()
 
     return sum;
 }
+
+/**
+ * @brief Open a gpiochip by number.
+ * @return GPIO chip pointer handle or NULL if an error occurred.
+ */
+extern struct gpiod_chip * OpenChipByNumber(int number)
+{
+    return gpiod_chip_open_by_number(number);
+}
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -145,4 +145,11 @@ internal partial class Interop
     /// <returns>1 if rising edge event occured, 2 on falling edge, -1 on error.</returns>
     [DllImport(LibgpiodLibrary, SetLastError = true)]
     internal static extern int ReadEventForLine(SafeLineHandle line);
+
+    /// <summary>
+    /// List all GPIO chips.
+    /// </summary>
+    /// <returns>The number of GPIO chips, -1 if an error occurred.</returns>
+    [DllImport(LibgpiodLibrary, SetLastError = true)]
+    internal static extern int GetNumberOfChips();
 }
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -152,4 +152,11 @@ internal partial class Interop
     /// <returns>The number of GPIO chips, -1 if an error occurred.</returns>
     [DllImport(LibgpiodLibrary, SetLastError = true)]
     internal static extern int GetNumberOfChips();
+
+    /// <summary>
+    /// Open a gpiochip by number.
+    /// </summary>
+    /// <returns>GPIO chip pointer handle or NULL if an error occurred.</returns>
+    [DllImport(LibgpiodLibrary, SetLastError = true)]
+    internal static extern SafeChipHandle OpenChipByNumber(int number);
 }
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -147,13 +147,6 @@ internal partial class Interop
     internal static extern int ReadEventForLine(SafeLineHandle line);
 
     /// <summary>
-    /// List all GPIO chips.
-    /// </summary>
-    /// <returns>The number of GPIO chips, -1 if an error occurred.</returns>
-    [DllImport(LibgpiodLibrary, SetLastError = true)]
-    internal static extern int GetNumberOfChips();
-
-    /// <summary>
     /// Open a gpiochip by number.
     /// </summary>
     /// <returns>GPIO chip pointer handle or NULL if an error occurred.</returns>
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
@@ -20,29 +20,20 @@ namespace System.Device.Gpio.Drivers
 
         protected internal override int PinCount => Interop.GetNumberOfLines(_chip);
 
-        public LibGpiodDriver()
+        public LibGpiodDriver(int gpioChip = 0)
         {
-            // set /dev/gpiochip0 as default if hardware has only one GPIO bank
-            if (Interop.GetNumberOfChips() == 1)
-                SetDefaultGpioChip(0);
-        }
-
-        protected internal override void SetDefaultGpioChip(int gpioChip)
-        {
-            if (!_gpioChipIsSet)
+            try
             {
                 _chip = Interop.OpenChipByNumber(gpioChip);
                 if (_chip == null)
                 {
                     throw ExceptionHelper.GetIOException(ExceptionResource.NoChipFound, Marshal.GetLastWin32Error());
                 }
-
-                _gpioChipIsSet = true;
                 _pinNumberToSafeLineHandle = new Dictionary<int, SafeLineHandle>(PinCount);
             }
-            else
+            catch (DllNotFoundException)
             {
-                throw ExceptionHelper.GetInvalidOperationException(ExceptionResource.GpioChipControllerAlreadySet);
+                throw ExceptionHelper.GetPlatformNotSupportedException(ExceptionResource.LibGpiodNotInstalled);
             }
         }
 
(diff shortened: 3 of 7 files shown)


src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -12,6 +12,7 @@ namespace System.Device.Gpio.Drivers
     public class LibGpiodDriver : UnixDriver
     {
         private SafeChipHandle _chip;
+		private bool _gpiochip_seted = false;
```
shaggygi: Change to _gpioChipIsSet throughout update.

src/System.Device.Gpio/System/Device/Gpio/ExceptionHelper.cs
```diff
@@ -100,6 +103,7 @@ internal enum ExceptionResource
         EventWaitError,
         EventReadError,
         NotListeningForEventError,
-        LibGpiodNotInstalled
+        LibGpiodNotInstalled,
+		GpioChipControllerAlreadySeted,
```
shaggygi: Change to GpioChipControllerAlreadySet.

src/System.Device.Gpio/System/Device/Gpio/ExceptionHelper.cs
```diff
@@ -76,6 +76,9 @@ private static string GetResourceString(ExceptionResource resource, int errorCod
                 case ExceptionResource.LibGpiodNotInstalled:
                     message = $"Libgpiod driver not installed. More information on: https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/about/";
                     break;
+				case ExceptionResource.GpioChipControllerAlreadySeted:
+					message = $"GpioController Default gpiochip already seted.";
```
shaggygi: Maybe change to "Default GPIO controller chip already set."?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -56,6 +56,11 @@ private int GetLogicalPinNumber(int pinNumber)
             return (NumberingScheme == PinNumberingScheme.Logical) ? pinNumber : _driver.ConvertPinNumberToLogicalNumberingScheme(pinNumber);
         }
 
+		public void SetDefaultGpioChip(int devGpioChip)
```
shaggygi: Change parameter to gpioChip.

src/Native/System.Device.Gpio.Native/libgpiod.c
```diff
@@ -186,4 +186,27 @@ extern int ReadEventForLine(struct gpiod_line *line)
...
+{
+	struct gpiod_chip_iter *iter;
+	struct gpiod_chip *chip;
+	int sum = 0;
+
+	iter = gpiod_chip_iter_new();
+	if (!iter)
+		return -1;
+	
+	gpiod_foreach_chip(iter, chip) {
```
shaggygi: NIT: "{" on new line.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
krwq: also nit: indentation off by 1

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
```diff
@@ -14,6 +14,14 @@ public abstract class GpioDriver : IDisposable
         /// </summary>
         protected internal abstract int PinCount { get; }
 
+		/// <summary>
+        /// Set the GPIO bank to be used /dev/gpiochip0 ... /dev/gpiochipX
+        /// </summary>
+		protected internal virtual void SetDefaultGpioChip(int devGpioChip)
```
krwq: is it virtual for back compat? I'd go with abstract since we haven't shipped yet.

src/System.Device.Gpio/System/Device/Gpio/IGpioController.cs
```diff
@@ -6,6 +6,7 @@ namespace System.Device.Gpio
 {
     public interface IGpioController : IDisposable
     {
+		void SetDefaultGpioChip(int devGpioChip);
```
krwq: I'm wondering if instead we should perhaps have i.e. (or just a constructor argument):
`LibGpiodController.Create(chip = 1)` - this way we don't need to change the interface and you can still work correctly in your scenario

buyaa-n: Agree better to have constructor argument

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -21,29 +22,48 @@ public class LibGpiodDriver : UnixDriver
...
+				}
+				catch (DllNotFoundException)
+				{
+					throw ExceptionHelper.GetPlatformNotSupportedException(ExceptionResource.LibGpiodNotInstalled);
+				}
+
+				// here we are getting the default controller /dev/gpiochipX
+				for (int i = 0; i <= devGpioChip; i++) 
+				{
+					_chip = Interop.GetNextChipFromChipIterator(iterator);
```
buyaa-n: you can load gpio chips directly by name or chim number https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/tree/include/gpiod.h#n416.
So for me we don't need to iterate by each chip, further seems don't see use of GetNumberOfChips method and related logic, can be just load chip 0 by default if chip number not provided

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
tarekgh: >SetDefaultGpioChip [](start = 14, length = 18)

two things here: 

1- should we allow passing the default gpio chip to the controller constructor? even as a default parameter?

2- should we convert this method to a property with setter and getter? something like 


public int DefaultGpioChip { set; get; }

src/Native/System.Device.Gpio.Native/libgpiod.c
microhobby: Thanks for the review @shaggygi

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Thanks for the review @buyaa-n 
In b6cb398 I changed the iterator way to gpiod_chip_open_by_number

microhobby:
> Some notes:
> 
> * I think we should instead have a way of passing that in the constructor of LibGpiodController without modifying interface (IMO the controller should represent a specific chip and not have more subchips)
> * indentation is slightly off all around the code
> 
> cc: @buyaa-n @tarekgh @JeremyKuhne
> 
> The usage with that API would be i.e.:
> 
> ```cs
> using (GpioController gpiochip4 = new LibGpiodController(chipId = 4))
> {
> //...
> }
> ```
> 
> perhaps we could also add that somewhere in the GpioController constructor - but IMO interface doesn't have to have it.


Thanks for the review @krwq 
I hope the code style are correct now 2a40a89

src/System.Device.Gpio/System/Device/Gpio/IGpioController.cs
microhobby: This approach appears good to me

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
microhobby: Thanks for the review @tarekgh 
I liked the approach of passing that in the constructor of LibGpiodController

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
buyaa-n: i think we should set the chip number on driver i guess we don't want to update all drivers with this method, preferably set it with driver constructor argument.

src/System.Device.Gpio/System/Device/Gpio/IGpioController.cs
buyaa-n: oh i misread @krwq's comment, I actually meant pass to the LibGpiodDriver constructor not to the controller. @krwq might meant the same too.  Our Drivers are public, so you can create an instance and set to the controller directly 
`GpioController controller = new GpioController(PinNumberingScheme.Logical , new LibGpiodDriver(chip=1))`
Don't need to set chip number controller level https://github.com/dotnet/iot/pull/346/files#diff-7fe07ed0a542d5e8abc343a8cd469bc1R9

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -21,27 +22,28 @@ public class LibGpiodDriver : UnixDriver
 
         public LibGpiodDriver()
         {
-            SafeChipIteratorHandle iterator = null;
-            try
+            // set /dev/gpiochip0 as default if hardware has only one GPIO bank
+            if (Interop.GetNumberOfChips() == 1)
```
buyaa-n: Some Linux version expose more than one chips even there were only one gpio chip. In that case by this logic any chip not loaded until below SetDefaultGpioChip(..) method called from controller.
What i would prefer is load the chip on driver constructor only (i don't think we would need to change a driver or chip for a controller, if we want to access another chip create another controller with a driver pointing to that chip). Actually all other driver methods written for only one same chip usage, chip shouldn't be changed after driver creation. So chip number should be set with driver constructor once, shouldn't be updatable further

buyaa-n: Basically something like this
```
public LibGpiodDriver(int chipId = 0) {
    try
    {
        _chip = Interop.OpenChipByNumber(gpioChip);
        if (_chip == null)
        {
            throw ExceptionHelper.GetIOException(ExceptionResource.NoChipFound, Marshal.GetLastWin32Error());
        }
        _pinNumberToSafeLineHandle = new Dictionary<int, SafeLineHandle>(PinCount);
    } 
    catch (DllNotFoundException)
    {
        throw ExceptionHelper.GetPlatformNotSupportedException(ExceptionResource.LibGpiodNotInstalled);
     }
 }
}
```

src/System.Device.Gpio/System/Device/Gpio/ExceptionHelper.cs
```diff
@@ -76,6 +76,9 @@ private static string GetResourceString(ExceptionResource resource, int errorCod
                 case ExceptionResource.LibGpiodNotInstalled:
                     message = $"Libgpiod driver not installed. More information on: https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/about/";
                     break;
+                case ExceptionResource.GpioChipControllerAlreadySet:
+                    message = $"GpioController Default gpiochip already set.";
```
buyaa-n: Seems you had idea that first loaded chip shouldn't be changed, by setting and loading the chip on the constructor we wouldn't need to worry about further chip changes and wouldn't need this error

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -145,4 +145,18 @@ internal partial class Interop
     /// <returns>1 if rising edge event occured, 2 on falling edge, -1 on error.</returns>
     [DllImport(LibgpiodLibrary, SetLastError = true)]
     internal static extern int ReadEventForLine(SafeLineHandle line);
+
+    /// <summary>
+    /// List all GPIO chips.
+    /// </summary>
+    /// <returns>The number of GPIO chips, -1 if an error occurred.</returns>
+    [DllImport(LibgpiodLibrary, SetLastError = true)]
+    internal static extern int GetNumberOfChips();
```
krwq: I think this isn't used anymore so you can perhaps remove it

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Thanks @buyaa-n I put it in 940bc17

src/System.Device.Gpio/System/Device/Gpio/IGpioController.cs
microhobby: I edited the description of PR with the new example of use

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
microhobby: 1 was implemented on 940bc17

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
microhobby: Thanks for the review @krwq 
I remove it in d37106a

krwq:
Thanks for contributing @microhobby! It's really cool you had a chance to test with multiple GPIO banks! (I don't have any boards like that)

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: Thank you @microhobby

Issue #350, opened 2019-03-31
Device submission guidance 


PRIMETSS:
After starting work on a possible improvement to a device, I thought to check that what I was doing followed the Template for submission on Device. 
It deviated a bit, and had me go spot check more devices for an understanding of how things like setting Registers are being handled.

Typically Devices (Chips) are commonly a bunch of control registers, which are bit toggled through serial reads and writes through some interface  (e.g. I2C).
Commonly this is done through bit bashing of Registers & Register Parameter enums through bit masking.
e.g. [BNO0555.h](https://github.com/adafruit/Adafruit_BNO055/blob/master/Adafruit_BNO055.h)

These Registers are generally set 'on mass' by writing to the register after OR'ing up all the parameters you wish to set

` // Initalise ADS1115 default hardware config as per datasheet, Default Config Register = 0x8583
            _config = (ConfigParams.ADS1015_REG_CONFIG_OS_NOTBUSY | 
                ConfigParams.ADS1015_REG_CONFIG_MUX_DIFF_0_1 | ConfigParams.ADS1015_REG_CONFIG_PGA_2_048V |
                ConfigParams.ADS1015_REG_CONFIG_MODE_SINGLE | ConfigParams.ADS1015_REG_CONFIG_DR_8SPS |
                ConfigParams.ADS1015_REG_CONFIG_CMODE_TRAD | ConfigParams.ADS1015_REG_CONFIG_CPOL_ACTVLOW |
                ConfigParams.ADS1015_REG_CONFIG_CLAT_NONLAT | ConfigParams.ADS1015_REG_CONFIG_CQUE_NONE);`

Setting or Checking a single Register parameter could be done using a Bit Mask
`if ((this._config & ConfigParams.ADS1015_REG_CONFIG_MODE_SINGLE) == ConfigParams.ADS1015_REG_CONFIG_MODE_SINGLE)`

My question is that, looking at [Ads1115.cs](https://github.com/dotnet/iot/blob/master/src/devices/Ads1115/Ads1115.cs) for example, it seems that each 'group' of register parameters are abstracted out to their own separate classes,  and any parameter that the 'driver' wanted to change is brought in as a local private field or public Property.

My concern is, how does the approach scale to very complicated devices like the uBlox GPS, which can have many registers and hundreds of Register Properties.
[...]

PR #352, opened 2019-04-02
ADS1115 Further Improvements

PRIMETSS:
Adding to [ZhangGaoxing](https://github.com/ZhangGaoxing/iot/tree/ads1115-fix) good work

- Converted limited chip config public properties/fields to ENUMS & Masks to allow for better flexibility for configuring chip using bit masking.
- Added ability to configure ADS1115 for successive reading of multiple channels. (typically this is done quite often as the device has 4SE/2Diff channels and need to be able read each channel sequentially if required (after a conversion has finished).
- Reading the conversion register is now aware if the device is in SingleShot mode, and first triggers a SingleShot conversion to happen, and then waits for conversion to complete. If in continuous mode the last conversion is returned, which seems fair for continuous non blocking conversions.
- Added Config Register support methods for better control and direct reading of chips Config Register (not just Ads1115's object, as may be different for real time monitoring purposes)
- Removed unnecessary sleep on Config write (possibly fixed by below)
- Added blocking wait for chip to perform a conversion (as per device spec)
- Fixed error in performing Val to Volts conversion in RawToVoltage that was using MUX config to set resolution


krwq:
cc: @ZhangGaoxing

ZhangGaoxing:
@PRIMETSS What do you think of the readability of the code? If you are a user, what do you think of the usability and comprehensibility of the code?

Let me start with my point of view.
1. You delete most of the enum classes and merge them into a `Config` class. I think this violates the Single Responsibility Principle of object-oriented design. `Config` class becomes very bulky and unrelated. Leave this aside, setting registers like this is not user friendly too.
2. Naming is not standardized and poor readability. You should use Pascal-Case.
3. As for bugs, I'll read the datasheet carefully. You did a good job.

Left some comments, and we can discuss this further😄. I'm not a professional hardware programmer. My job requires me to consider the problem from the user's point of view, which makes my design different from the usual C programmer. Code design should be as user friendly as possible.

PRIMETSS:
Fair points!

The reason for the consolidation spawns a guidance question I had [#350](https://github.com/dotnet/iot/issues/350) re relying more on 'bit flipping' & masks for chip registers more, as opposed to separating out each Registers Parameter individually and having a class field/Property for each individual one.

The single Config enum I merge into one, as they all relate to one single hardware register, the Config register. Its unlikely they would be used outside of this context.

The ADS1115 is a relatively 'simple' chip with only a small number of register & parameters.
My concern was, for example a chip like uBlox Differential GPS literally has many registers & 100's of parameters.
Separating all those out into individual enum files would probably go into the 100's of files.
Also, using either constructors or class properties to set register fields would be messy? Using bit masks for hardware registers seems to make sense to me.

Agree, the naming of the Register parameters could be made more succinct. I followed the BNO005 as an example. [BNO005] (https://github.com/dotnet/iot/blob/master/src/devices/Bno055/Models/Registers.cs)

To just highlight something, as semantics is difficult. I’m not suggesting the use of bit masks should be used prolifically. Just on Hardware Register or where it makes sense. 
Properties or Methods should be used to provide general functionality.
eg, a chip may have certain parameters in registers that could be damaging under some conditions, so this sort of case is where setting of hardware register might be best to pass through a method for validation.

I guess Im approaching this from the chip level up, and It might not be the right approach.
Some guidance around how to best handle setting hardware register and that separation from chip functions might be helpful. But it might be my bias point of view.

Happy to have RP scrapped,  and maybe just work on the few items I think are in error, the 

- Waiting for conversion to complete when in SingleShot mode
- 10ms wating for config to 'settle': as Ive never seen this with this chip (suspect this is seen if not waiting for conversion to complete)
- Setting of the Volts conversion shouldn't be based on MUX config but PGA gain.

krwq:
@PRIMETSS @ZhangGaoxing what's the status on this? Should this PR be closed/cleaned up? What are the next steps?

PRIMETSS:
Thanks @krwq PR can be closed.
Bit masking probably wont gain traction, and might not be good mix when only implementing chosen part functionality of chips features.

joperezr:
Closing giving the above discussion. @PRIMETSS feel free to reopen if you think otherwise.

PR #356, opened 2019-04-10
Update License

krwq:
Update License to match https://github.com/dotnet/arcade/commit/1e49672184ca450fbf3ea73e1122924ed4bb2ccc
and fix issue with https://github.com/dotnet/iot/pull/339

@chcosta I'm not 100% sure if this is fixing the issue with that PR but seems that's what the error is about, please let me know if that's not it (that's the only difference I could think of)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,25 +3,20 @@
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
   <Import Project="$(RepositoryEngineeringDir)Compilers.props" />
 
-  <PropertyGroup Condition="'$(CopyrightNetFoundation)' != ''">
-    <Copyright>$(CopyrightNetFoundation)</Copyright>
+  <PropertyGroup>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
   </PropertyGroup>
-  
+
   <PropertyGroup>
     <PublishWindowsPdb>false</PublishWindowsPdb>
     <Language>C#</Language>
     <TargetsWindows Condition="$(RuntimeIdentifier.StartsWith('win'))">true</TargetsWindows>
     <TargetsLinux Condition="$(RuntimeIdentifier.StartsWith('linux'))">true</TargetsLinux>
     <LangVersion>Latest</LangVersion>
-
-    <!-- Packaging props -->
-    <PackageLicenseFile>LICENSE</PackageLicenseFile>
   </PropertyGroup>
 
   <ItemGroup>
     <!-- Include license and third party files to packages -->
     <Content Include="$(RepoRoot)THIRD-PARTY-NOTICES.TXT" Pack="true" PackagePath="\" />
-    <Content Include="$(RepoRoot)LICENSE" Pack="true" PackagePath="\" />
   </ItemGroup>
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,6 +4,7 @@
   <Import Project="$(RepositoryEngineeringDir)Compilers.props" />
 
   <PropertyGroup>
+    <Copyright>$(CopyrightNetFoundation)</Copyright>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
   </PropertyGroup>
 
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,8 @@
-MIT License
+The MIT License (MIT)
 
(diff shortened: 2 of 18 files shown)


chcosta:
That looks like a partial fix.  When you  update the sdk in your repo's global.json (to 2.1.503 or newer), you'll need to apply [this](https://github.com/dotnet/arcade/pull/2003) change as well.  I'd suggest you just do that now unless there's some strong reason you can't upgrade your repo's SDK version.

FYI @tmat

tmat:
Remove 
https://github.com/dotnet/iot/blob/master/Directory.Build.props#L25
https://github.com/dotnet/iot/blob/master/Directory.Build.props#L19
since we are using license expression rather than embedding license file.

You can also remove the condition on line https://github.com/dotnet/iot/blob/master/Directory.Build.props#L6.

joperezr:
Once you make the fixes that @chcosta and @tmat suggested this LGTM. Thanks for fixing! it will be great to start getting arcade updates again!

krwq:
not sure if I understood correctly what needs to be changed - I'm not too familiar with Arcade (presumably since we have default copyright now I don't need to define it anymore)?

Directory.Build.props
```diff
@@ -3,25 +3,20 @@
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
   <Import Project="$(RepositoryEngineeringDir)Compilers.props" />
 
-  <PropertyGroup Condition="'$(CopyrightNetFoundation)' != ''">
-    <Copyright>$(CopyrightNetFoundation)</Copyright>
```
joperezr: Do we also want to remove the Copyright? I only ask because that wasn't part of what @tmat suggested.

krwq: per this: https://github.com/dotnet/arcade/pull/2003/files#diff-cdcd303385132649c3cafdaa56c23e01R11 it should be default now, right?

tmat: No, keep the copyright.

tmat: I only suggested removing the condition `Condition="'$(CopyrightNetFoundation)' != ''"`

krwq:
I'm getting errors about:

```
C:\Program Files\dotnet\sdk\2.1.503\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(202,5): error NU5035: The PackageLicenseUrl is being deprecated and cannot be used in conjunction with the PackageLicenseFile or PackageLicenseExpression. [F:\vsagent\134\s\src\Iot.Device.Bindings\Iot.Device.Bindings.csproj]
C:\Program Files\dotnet\sdk\2.1.503\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(202,5): error NU5035: The PackageLicenseUrl is being deprecated and cannot be used in conjunction with the PackageLicenseFile or PackageLicenseExpression. [F:\vsagent\134\s\src\System.Device.Gpio\System.Device.Gpio.csproj]
```

I suspect I just need to update tools to higher version, do you perhaps have the right number?

tmat:
2.1.503 should be good enough.

Your build doesn't collect log artifacts, so we can't look at a build log :(
image

BTW, delete https://github.com/dotnet/iot/blob/master/src/Iot.Device.Bindings/Directory.Build.props#L7

joperezr:
I've merged my changes to publish the build logs, so I'm restarting the builds so that we can better analyze why is the build failing

joperezr:
restarting didn't take my changes, I'll try closing and reopening the PR. If that doesn't work, We may need that @krwq rebase to get my yml changes.

joperezr:
Ok, so according to the logs, the thing you need to remove is line

https://github.com/dotnet/iot/blob/c6a421bd862f99e5e151f628f1c54743b94ea006/Directory.Build.props#L8

Simply remove the line that sets the expression property, and build should pass now.

tmat:
That is not the right fix. Please keep the license expression.

joperezr:
@tmat I just reverted the change. Do you know what is the right fix then? You should be able to take a peek at the logs now, the one you care about is `Windows_NT_release_configuration/Build.binlog`

joperezr:
Looks like all we needed after your suggestions was actually to perform an update of arcade. @tmat is this good to go now?

tmat:
> actually to perform an update of arcade.

Ah, that explains. I assumed the update already happened.

Issue #353, opened 2019-04-02
Add Mcp3428 ADC bindings


kmate95:
4 channel ADC chip with I2C interface. Datasheet here: http://ww1.microchip.com/downloads/en/DeviceDoc/22226a.pdf
I'm currently working on it. Gonna make PR when I have time to finish it.

joshfree:
Thank you for the (future) contribution @kmate95 😄

krwq:
Closing this since PR is merged. Please let me know if there is anything left.

PR #357, opened 2019-04-11
Mcp3428

kmate95:
Implements #353. 

I've mostly detached it from my project. I left in the async API as read times can go up to 60ms on 16 bit resolution. The return value type can be useful for push based continuous background reading.

Still needs a bit of cleanup.


src/devices/Mcp3428/ConversionResult.cs
```diff
@@ -0,0 +1,58 @@
+﻿// Mcp3428::Mcp3428
```
krwq: We require copyright header on all .cs files. I.e. https://github.com/dotnet/iot/blob/master/src/devices/Bno055/Bno055Sensor.cs#L1-L3

src/devices/Mcp3428/Helpers.cs
```diff
@@ -0,0 +1,149 @@
...
+        public static byte AddressFromPins(PinState Adr0, PinState Adr1)
+        {
+            byte addr = 0b1101000; // Base value from doc
+
+            var idx = (byte)Adr0 << 4 + (byte)Adr1;
+
+            // switch (new ValueTuple<PinState, PinState>(Adr0, Adr1))
+            switch (idx)
+            {
+                case (0):
```
krwq: brackets around the numbers in `case` are not needed

src/devices/Mcp3428/Mcp3427.cs
```diff
@@ -0,0 +1,24 @@
...
+        public Mcp3427(I2cDevice i2CDevice) : base(i2CDevice, 2)
+        {
+        }
+
+        /// <inheritdoc />
+        public Mcp3427(I2cDevice i2CDevice, ModeEnum mode = ModeEnum.Continuous, ResolutionEnum resolution = ResolutionEnum.Bit12, GainEnum pgaGain = GainEnum.X1) : this(i2CDevice)
+        {
+            SetConfig(0, resolution: resolution, mode: mode, pgaGain: pgaGain);
+        }
+
```
krwq: nit: extra enters

src/devices/Mcp3428/Mcp3428.cs
```diff
@@ -0,0 +1,301 @@
...
+            {
+                conf = SetChannelBits(conf, channel);
+            }
+
+            WriteConfig(conf);
+            using (var source = new CancellationTokenSource(TimeSpan.FromMilliseconds(WaitTime*5)))
+            {
+                WaitForConversion(TimeSpan.FromMilliseconds(WaitTime), cancellationToken:source.Token);
+
+            if (!_isReadyBit)
```
krwq: nit: indent

src/devices/Mcp3428/Mcp3428.cs
```diff
@@ -0,0 +1,301 @@
...
+
+            _i2cDevice.WriteByte(conf);
+
+            _i2cDevice.Read(_readBuffer);
+            ReadConfigByte(LastConfigByte);
+
+            
+            if (_lastChannel != channel)
+            {
+                Console.WriteLine($"Channel update failed from {_lastChannel} to {channel}");
```
krwq: I'd recommend throwing or returning an exception instead. Writing to output is almost always not expected by user of the library

src/devices/Mcp3428/ConversionResult.cs
```diff
@@ -0,0 +1,58 @@
...
+// Filename: ConversionResult.cs
+// Created: __
+// Edited: 20190405
+// Creator: Máté Kullai
+
+using System;
+
+namespace Iot.Device.Mcp3428
+{
+    public interface IAdcResult
```
krwq: this and ConversionResultHandler don't seem to be used anywhere here. I'd recommend removing the interface unless there is a good reason to have it. For ADC `Func<double>` should be enough to abstract it

src/devices/Mcp3428/Helpers.cs
```diff
@@ -0,0 +1,149 @@
...
+        // From datasheet 5.2
+        /// <summary>
+        /// Address from pin configuration. Based on documentation TABLE 5-3-
+        /// </summary>
+        /// <param name="Adr1">The adr1.</param>
+        /// <param name="Adr0">The adr0.</param>
+        /// <returns>System.Byte.</returns>
+        /// <exception cref="ArgumentException">Invalid combination</exception>
+        /// <autogeneratedoc />
+        public static byte AddressFromPins(PinState Adr0, PinState Adr1)
```
krwq: I think the name should be clearer what this address refers to

src/devices/Mcp3428/Mcp3428Async.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Collections.Generic;
+using System.Device.I2c;
+using System.IO;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Iot.Device.Mcp3428
+{
+    using static Helpers;
```
joperezr: We usually try to avoid the using static directive so that it is more obvious when reading the code where do static methods come from, so please remove this.

src/devices/Mcp3428/Mcp3428Async.cs
```diff
@@ -0,0 +1,127 @@
...
+    /// <summary>
+    /// Async API is mostly useful for greater resolutions and one-shot mode, when conversion time can be significant.
+    /// Implements the <see cref="Iot.Device.Mcp3428.Mcp3428" />
+    /// </summary>
+    /// <seealso cref="Iot.Device.Mcp3428.Mcp3428" />
+    /// <autogeneratedoc />
+    public class Mcp3428Async : Mcp3428
```
joperezr: Any particular reason why we chose to separate Async code into a different class? I would rather follow the .NET patterns and have them all in the same class, and simply naming the methods that are Async with the `Async` suffix.

kmate95: I wasn't sure in the async interface for a long time. Decided to wait for review and fix later.   
The other reason was that I have some caching and background update ideas for my use case, where I have to poll multiple devices. It seamed to be reasonable to cut the API to a "sync and just works" and a more feature-rich implementation. This never made it into the PR...  
Fixing it right now.

src/devices/Mcp3428/Mcp3428.cs
kmate95: I forgot this one too.
Dropped in dummy error list as input. Was useful to see when more than one errors were made on the conversion.

src/devices/Mcp3428/Helpers.cs
kmate95: Renamed to I2CAddressFromPins + better doc

src/devices/Mcp3428/ConversionResult.cs
kmate95: Would be nice to have some common ADC interfaces for the library.

src/devices/Mcp3428/ConversionResult.cs
kmate95: Only missed one :)

kmate95:
I hope I fixed all your remarks. Thanks for the help.  
If you see anything else, please tell.

src/devices/Mcp3428/ConversionResult.cs
krwq: @kmate95 - I agree and I've already tried to add one in the past but got outvoted that we should just use `Func<double>` instead of the interface since that's what ADC produces and value doesn't have any specific unit so it's not suitable for the interface

src/devices/Mcp3428/ConversionResult.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <value>The channel.</value>
+        /// <autogeneratedoc />
+        public byte Channel { get; }
+
+        /// <summary>
+        /// Gets the raw data.
+        /// </summary>
+        /// <value>The raw data.</value>
+        /// <autogeneratedoc />
+        public Int16 RawData { get; }
```
krwq: nit: `Int16` => `short`

src/devices/Mcp3428/ConversionResult.cs
```diff
@@ -0,0 +1,48 @@
...
+        private ResolutionEnum Resolution { get; }
+
+        public double LSBValue => Helpers.LSBValue(Resolution);
+
+        /// <summary>
+        /// Gets the voltage.
+        /// </summary>
+        /// <value>The voltage.</value>
+        /// <autogeneratedoc />
+        public double Voltage => Helpers.LSBValue(Resolution) * RawData;
```
krwq: `Voltage => LSBValue * RawValue`

src/devices/Mcp3428/ConversionResult.cs
```diff
@@ -0,0 +1,48 @@
...
+        public Int16 RawData { get; }
+
+        /// <summary>
+        /// Gets the resolution.
+        /// </summary>
+        /// <value>The resolution.</value>
+        /// <autogeneratedoc />
+        private ResolutionEnum Resolution { get; }
+
+        public double LSBValue => Helpers.LSBValue(Resolution);
```
krwq: this should have some comment or use full name since it's not clear what that is. Also probably the proper casing is `LsbValue` assuming `LSB` is what the industry uses as a name

src/devices/Mcp3428/Helpers.cs
```diff
@@ -0,0 +1,146 @@
...
+                case ResolutionEnum.Bit16:
+                    return 15;
+
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(res), res, null);
+            }
+        }
+
+        // From datasheet 5.2
+        public static class Masks
```
krwq: should this be internal? is this useful outside of this library?

krwq: nvm, this is nested under internal class

src/devices/Mcp3428/Helpers.cs
```diff
@@ -0,0 +1,146 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.Mcp3428
+{
+    public enum GainEnum : byte { X1 = 0, X2 = 1, X4 = 2, X8 = 3 }
```
krwq: since those enums are in Helpers.cs should they be internal? If not they should probably be in a separate file(s)

src/devices/Mcp3428/Mcp3426.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Device.I2c;
+
+namespace Iot.Device.Mcp3428
+{
+    public class Mcp3426 : Mcp3428
+    {
+        public const int Address = 0x68;
+
+        /// <inheritdoc />
+        public Mcp3426(I2cDevice i2CDevice) : base(i2CDevice, 2)
```
krwq: please create private constant for 2 since it's not immediately obvious (without looking at the base)  what that is

src/devices/Mcp3428/Mcp3426.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+        }
+
+        /// <inheritdoc />
+        public Mcp3426(I2cDevice i2CDevice, ModeEnum mode = ModeEnum.Continuous, ResolutionEnum resolution = ResolutionEnum.Bit12, GainEnum pgaGain = GainEnum.X1) : this(i2CDevice)
+        {
+            SetConfig(0, mode: mode, resolution: resolution, pgaGain: pgaGain);
+        }
+
+#pragma warning disable RCS1163 // Unused parameter.
```
krwq: I think you should either create a base class for these two and put static method in only in the derived classes or remove this method - in the current state this is really confusing

src/devices/Mcp3428/Mcp3428.cs
```diff
@@ -0,0 +1,409 @@
...
+        /// Protected constructor to easily generate MCP3426/7 devices whose only difference is channel count and I2C address
+        /// </summary>
+        /// <param name="i2CDevice">The i2 c device.</param>
+        /// <param name="channels">The channels.</param>
+        /// <autogeneratedoc />
+        protected Mcp3428(I2cDevice i2CDevice, int channels)
+        {
+            _i2cDevice = i2CDevice;
+            ChannelCount = channels;
+            ReadValue(); // Don't like this in constructor, makes sure props are valid
```
krwq: not sure I understand the comment, what happens if you remove this?

src/devices/Mcp3428/Mcp3428.cs
```diff
@@ -0,0 +1,409 @@
...
+        /// <value>The resolution.</value>
+        /// <autogeneratedoc />
+        public ResolutionEnum Resolution
+        {
+            get => _resolution;
+            set
+            {
+                WriteConfig(Helpers.SetResolutionBits(LastConfigByte, value));
+                _resolution = value;
+                //WaitTime = (int)(1000.0 / UpdateFrequency(_resolution));
```
krwq: remove dead code

src/devices/Mcp3428/Mcp3428.cs
```diff
@@ -0,0 +1,409 @@
...
+        /// <returns>System.Double.</returns>
+        /// <autogeneratedoc />
+        public double ReadChannel(int channel)
+        {
+            return ReadValue(channel);
+        }
+
+        protected const int MaxRetries = 50;
+        protected readonly I2cDevice _i2cDevice;
+        protected readonly byte[] _readBuffer = new byte[3];
```
krwq: 1. why not private?
2. Protected members should start with capital letters (i.e. ReadBuffer)

src/devices/Mcp3428/Mcp3428.csproj
```diff
@@ -0,0 +1,15 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <!--Disabling default items so samples source won't get build by the main library-->
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <None Include="README.md" />
+    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
```
krwq: recently we changed all Version to `$(SystemDeviceGpioPackageVersion)`, see i.e.: https://github.com/dotnet/iot/blob/master/src/devices/Bmp180/Bmp180.csproj#L13

krwq:
@kmate95 sorry for the late re-review - lost the notification somewhere (please ping us if we don't review)

src/devices/Mcp3428/Mcp3428.cs
kmate95: I don't like initiating communication in the constructor.
This method ensures that if we are not after reset and previous changes were made to the device configuration then these changes are reflected in the created object.
The class is designed to always mirror hardware configuration state so this read call is required to initialize an instance.

src/devices/Mcp3428/ConversionResult.cs
kmate95: Fixed with comment. LSB (= least significant bit) is indeed the name used in the industry.

src/devices/Mcp3428/ConversionResult.cs
kmate95: Tried to save a function call. But it's the work of the compiler...
Fixed.

src/devices/Mcp3428/ConversionResult.cs
kmate95: This one had more meaning for me. Will read more about why is this recommended.

src/devices/Mcp3428/ConversionResult.cs
kmate95: I'm not sure I agree with this reasoning. Every ADC unit on every uC **always** produces a voltage value. That voltage may have other meaning like current measurement, weight, pressure, anything.   
That is the concern of the consuming software. Yet the ADC will have an intermediate voltage value to convert. Like this 4 channel converter, the first channel can be temperature, the second humidity, the third pot-meter and the forth a rocket launch countdown. But they always have to go through a (channel number)=>(voltage)=>(anything) interface. And if the first hop is abstracted away, only the business relevant part is left to the consumer.

src/devices/Mcp3428/Mcp3428.cs
kmate95: Leftover form previous derived async implementation. Fixed.

src/devices/Mcp3428/Mcp3428.csproj
kmate95: So is it no longer possible to build the project standalone?  
Makes it inconvenient to develop devices, especially when you have the project included in some of your own projects.

src/devices/Mcp3428/Mcp3426.cs
kmate95: From the above selection I'm not sure what you refer to. Most likely it's about the I2CAddressFromPins in Mcp3426, this was always a weak point of the design. I didn't want to implement Mcp3426/7 until I realized that they only differ in channel count and address configuration. May introduce base class for all three.

src/devices/Mcp3428/Helpers.cs
kmate95: They shall not be internal, they are used by the consuming software to set device configuration. 
I came up with a separate file, but I think this is just one more file that has no real purpose.

kmate95:
Thank you @krwq for the review.  
I think I addressed all of your concerns, hopefully not introducing new problems. I kind of blindly implemented most of your suggestions, yet I'm left with some questions. I commented them above and I'd be thankful if you had time to answer some of them.

I'm kind of inexperienced with Github PRs so I hope it's not too big of a mess.

src/devices/Mcp3428/ConversionResult.cs
krwq: @kmate95 - didn't relate because the type is double :-) perhaps should be called just "Value"

src/devices/Mcp3428/ConversionResult.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <value>The resolution.</value>
+        /// <autogeneratedoc />
+        private ResolutionEnum Resolution { get; }
+
+        /// <summary>
+        /// Voltage value of the least significant bit of the data
+        /// </summary>
+        /// <value>The LSB value.</value>
+        /// <autogeneratedoc />
+        public double LsbValue => Helpers.LSBValue(Resolution);
```
krwq: this is super confusing.

IMO this should be renamed to i.e. Scale or store reverse unit and call it i.e. Divisor/VoltageDivisor as long (instead of small double store the factor you divide by). Voltage could still be double.

src/devices/Mcp3428/ConversionResult.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <value>The channel.</value>
+        /// <autogeneratedoc />
+        public byte Channel { get; }
+
+        /// <summary>
+        /// Gets the raw data.
+        /// </summary>
+        /// <value>The raw data.</value>
+        /// <autogeneratedoc />
+        public short RawData { get; }
```
krwq: I'd recommend to change name to RawValue

src/devices/Mcp3428/Config.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    /// Possible connection states for the address pins
+    /// </summary>
+    public enum PinState : byte { Low = 0, High = 1, Floating = 2 }
+
+    /// <summary>
+    /// Possible resolution values of the ADC
+    /// </summary>
+    public enum ResolutionEnum : byte { Bit12 = 0, Bit14 = 4, Bit16 = 8 }
+
```
krwq: extra enter

src/devices/Mcp3428/Config.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// Possible gain values of the ADC
+    /// </summary>
+    public enum GainEnum : byte { X1 = 0, X2 = 1, X4 = 2, X8 = 3 }
+
+    /// <summary>
+    /// Possible operation modes of the ADC
+    /// </summary>
+    public enum ModeEnum : byte { OneShot = 0, Continuous = 16 }
```
krwq: Can we remove `Enum` suffix? we usually don't use such prefixes. Perhaps just `Gain` and `Mode` or `AdcGain`/`AdcMode`

src/devices/Mcp3428/Mcp3428.cs
```diff
@@ -0,0 +1,36 @@
...
+            SetConfig(0, mode: mode, resolution: resolution, pgaGain: pgaGain);
+        }
+
+        /// <summary>
+        /// Determine device I2C address based on the configuration pin states.
+        /// </summary>
+        /// <param name="Adr0">The adr0 pin state</param>
+        /// <param name="Adr1">The adr1 pin state</param>
+        /// <returns>System.Int32.</returns>
+        public static int I2CAddressFromPins(PinState Adr0, PinState Adr1) { return Helpers.I2CAddressFromPins(Adr0, Adr1); }
```
krwq: argument names should start with small letter (`adr0`, `adr1`)

src/devices/Mcp3428/Mcp342x.cs
```diff
@@ -0,0 +1,387 @@
...
+                _mode = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the input gain.
+        /// </summary>
+        /// <value>The pga gain.</value>
+        /// <autogeneratedoc />
+        public GainEnum PGAGain
```
krwq: if I read correctly PGA stands for `Programmable-gain amplifier` which means this name reads weird: `Programmable gain amplifier gain`

src/devices/Mcp3428/Mcp342x.cs
```diff
@@ -0,0 +1,387 @@
...
+        {
+            get => _resolution;
+            set
+            {
+                WriteConfig(Helpers.SetResolutionBits(LastConfigByte, value));
+                _resolution = value;
+            }
+        }
+
+        public void Dispose()
```
krwq: we usually implement this as:

```
_i2cDevice?.Dispose();
_i2cDevice = null;
```

src/devices/Mcp3428/Mcp342x.cs
```diff
@@ -0,0 +1,387 @@
...
+            {
+                if (channel > 0 && channel != LastChannel)
+                {
+                    var conf = Helpers.SetChannelBits(LastConfigByte, channel);
+                    WriteConfig(conf);
+                }
+
+                using (var source = new CancellationTokenSource(TimeSpan.FromMilliseconds(WaitTime * 5)))
+                {
+                    WaitForConversion(TimeSpan.FromMilliseconds(WaitTime / 5), cancellationToken: source.Token);
```
krwq: why is WaitTime divided by 5? this means the token is valid 25x longer, is this intentional?

src/devices/Mcp3428/Mcp342x.cs
```diff
@@ -0,0 +1,387 @@
...
+                }
+            }
+
+            //var value = BinaryPrimitives.ReadInt16LittleEndian(_readBuffer.AsSpan().Slice(1, 2));
```
krwq: remove commented out code

src/devices/Mcp3428/Mcp342x.cs
```diff
@@ -0,0 +1,387 @@
...
+        }
+
+        public event EventHandler<ConversionResult> OnConversion;
+
+        protected void WriteConfig(byte configByte)
+        {
+            _i2cDevice.WriteByte(configByte);
+        }
+
+        private static int _asyncThreshold = 20;
```
krwq: should this be const?

src/devices/Mcp3428/Mcp342x.cs
```diff
@@ -0,0 +1,387 @@
...
+        /// Sets the lower time limit in ms. If the current configuration reads data faster the synchronous API is used.
+        /// This can save some overhead.
+        /// </summary>
+        /// <remarks>
+        /// The default configuration is 20ms. This means that only in 16 bit resolution read waits asynchronously.
+        /// Setting it to 0 or lower disables the function.
+        /// </remarks>
+        /// <param name="thresh">Time limit in ms. Default: 20ms</param>
+        /// <autogeneratedoc />
+        public static void SetAsyncThreshold(int thresh) { _asyncThreshold = thresh; }
```
krwq: why is this static? what if you have two devices connected but want different values for them?

src/devices/Mcp3428/Mcp342x.cs
```diff
@@ -0,0 +1,387 @@
...
+        /// <param name="channel">The channel.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>ValueTask.</returns>
+        /// <exception cref="System.IO.IOException">
+        /// Device is not in One-Shot mode
+        /// or
+        /// ADC Conversion was not ready after {tries}
+        /// </exception>
+        /// <autogeneratedoc />
+        protected ValueTask OneShotReadAsync(int channel = -1, CancellationToken cancellationToken = default)
```
krwq: should methods taking `channel` throw when invalid value is used?

kmate95: I think it does in the bit setter function.

src/devices/Mcp3428/Mcp342x.cs
kmate95: It's called this in the device documentation.

src/devices/Mcp3428/Mcp342x.cs
kmate95: Was thinking if I got it from outside then I shouldn't dispose it.

src/devices/Mcp3428/Mcp342x.cs
kmate95: The `WaitTime / 5` is for when the start of the reading is not in sync with the conversion.  So I poll it multiple times in a period because it may have completed. `WaitForConversion` Actually waits forever if not cancelled, but after every `waitSpan` (input) time it polls the device.

The code actually means it tries to read five times in every nominal conversion period. And if it does not complete after five nominal periods the read gets cancelled.

src/devices/Mcp3428/Mcp342x.cs
krwq: we can fix it here but treat as optional if that's what the doc calls it :)

src/devices/Mcp3428/Mcp342x.cs
krwq: let's do it for consistency - we will likely re-iterate on the dispose pattern in the future and fix it everywhere if needed

src/devices/Mcp3428/Mcp342x.cs
kmate95: Then I will name it InputGain.

src/devices/Mcp3428/Mcp3428.csproj
shaggygi: I also came across this issue and didn't realize.  I usually create my binding projects outside of the iot structure at first.  This new property makes us have to remember to update before creating a PR.

joperezr: You can always have your own project outside of the repo to define this property as well. There are two main reasons why we want to keep the property instead of the wildcard version:
 - You make sure that you reference the last produced build, as opposed to maybe a feed being down or a version to be broken and we silently get what ever version NuGet resolves without us noticing it.
 - Most importantly, it enables us to be able to have deterministic builds to any point in time. If today you do a git checkout and go back in time to when we used the wildcard notation, you will find that you won't be able to build. This is because when ingesting new versions, it may require changes in the repo side in order to get those versions working, and when using wildcards, you will always get the latest even though the repo might not have the required changes yet in order to consume this version.

krwq:
@kmate95 do you think you'll have time to address feedback?

src/devices/Mcp3428/Mcp342x.cs
kmate95: Because I imagined `_asyncThreshold ` as an application wide design decision where the developer can decide if their use case values non-blocking read but nondeterministic timing or blocking call with more strict timing.

I can't see real benefits of having different values for separate instances. Read speed itself is settable per instance by setting the resolution.

I can change it if you want, but I would find the above function confusing on an instance.

kmate95:
@krwq Sorry for the delay. I got caught up in other projects and exams. 

Since I have to keep separate branches for build and PR and my own project it became a tedious task to maintain it. It's a problem I want to solve because there will be other drivers like this and I plan to contribute those too.

Plus you made some design decisions I'm not really comfortable with. Like the dispose pattern (I hope it gets cleared up consistently on the whole project) And I'm not decided yet if I want to do these for the PR and keep a separate version for my project or modify my architecture to build around the _official_ API.

I will certainly do it as the requested changes are not big. I'll try to patch it up in the coming days.

src/devices/Mcp3428/ConversionResult.cs
kmate95: I renamed it to Accuracy it's the same value but I hope it's less confusing.
I introduced a VoltageDivisor property as it fits into ushort and removed the Resolution as it's redundant this way.

src/devices/Mcp3428/Mcp342x.cs
kmate95: I still don't like it but it's there now. 
Should I remove readonly modifier from the field to be able to set it to null?

kmate95:
@krwq In the end I did it today. Please check if you have time. I hope I addressed your concerns.

src/devices/Mcp3428/Mcp342x.cs
krwq: let's leave it as is

PR #358, opened 2019-04-11
Migrating iot yml from phases to jobs

joperezr:
cc: @safern @MattGal @chcosta @krwq 

Migrating builds to the BYOC and Hosted model.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -94,14 +94,9 @@ jobs:
     agentOs: Linux
     name: Linux_libgpiod
     enableTelemetry: $(_enableTelemetry)
-    enablePublishBuildArtifacts: true
     helixRepo: dotnet/iot
     pool:
       name: Hosted Ubuntu 1604
-    strategy:
-      matrix:
-        release_configuration:
-          _BuildConfig: Release
     steps:
     - script: src/Native/copyLibgpiod.sh
       name: CopyLibgpiod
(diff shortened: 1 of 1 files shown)


.vsts-ci.yml
```diff
@@ -1,12 +1,15 @@
...
-  _HelixType: build/product
-  _HelixSource: pr/dotnet/arcade-minimalci-sample/$(Build.SourceBranch)
-  _enableTelemetry: true
-  ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
-    OfficialBuildId: $(Build.BuildNumber)
-    DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
-    DOTNET_MULTILEVEL_LOOKUP: 0
+- name: Build.Repository.Clean
+  value: true
+- name: _enableTelemetry
```
safern: Is this variable still needed? I think this is now a parameter on each Job but don't remember.

joperezr: I am sending it to each job and setting it to true since the job defaults to false.

safern: Ah I missed that, this is just a shared variable for your multiple jobs, this is one of the downsides of using the `Job` template, that you duplicate a lot of the parameters, when using the `Jobs` template you can pass multiple jobs down with the same set of parameters.

PR #362, opened 2019-04-15
Onboarding on auto-updates of System.Device.Gpio with darc instead of wildcards

joperezr:
Currently we were getting updates for the System.Device.Gpio package by using wildcards, which is not a great model since you can't ensure where your dependencies are comming from or which version are they. This PR will move to instead use arcade auto-update model so that our samples will now depend on the daily build produced by our official pipelines.

cc: @safern @krwq

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -3,6 +3,7 @@
   <packageSources>
     <clear />
     <add key="arcade" value="https://dotnetfeed.blob.core.windows.net/dotnet-tools-internal/index.json" />
+    <add key="dotnet-iot" value="https://dotnetfeed.blob.core.windows.net/dotnet-iot/index.json" />
     <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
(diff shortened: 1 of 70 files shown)


NuGet.config
```diff
@@ -3,6 +3,7 @@
   <packageSources>
     <clear />
     <add key="arcade" value="https://dotnetfeed.blob.core.windows.net/dotnet-tools-internal/index.json" />
+    <add key="dotnet-iot" value="https://dotnetfeed.blob.core.windows.net/dotnet-iot/index.json" />
     <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
     <add key="myget.org" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
```
safern: Do you still need the myget dependency?

joperezr: I don't think so, but forgot to remove it. I'll take care of that in a subsequent PR. Thanks for the review!

PR #367, opened 2019-04-16
Changes in order to make windows implementation Api compatible with Linux

joperezr:
cc: @buyaa-n @tarekgh @krwq 

We are not currently running ApiCompat as part of the build. Because of this, we did check in some changes in the Linux side that exposed API not present in the Windows side of the implementation. This would lead to runtime errors since we build the reference assembly out of the Linux implementation. These changes will make the Windows and Linux side ApiCompatible. After this is in, I'll make sure to add an ApiCompat check on the WIndows implementation build so that we don't go out of sync any longer.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Windows.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Windows.cs
@@ -1,0 +1,78 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+
+namespace System.Device.Gpio.Drivers
+{
+    public class LibGpiodDriver : UnixDriver
+    {
+        public LibGpiodDriver()
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        public LibGpiodDriver(int gpioChip = 0)
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        protected internal override int PinCount => throw new PlatformNotSupportedException();
+
+        protected internal override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEventTypes eventTypes, PinChangeEventHandler callback)
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        protected internal override void ClosePin(int pinNumber)
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        protected internal override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber)
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        protected internal override PinMode GetPinMode(int pinNumber)
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        protected internal override bool IsPinModeSupported(int pinNumber, PinMode mode)
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        protected internal override void OpenPin(int pinNumber)
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        protected internal override PinValue Read(int pinNumber)
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        protected internal override void RemoveCallbackForPinValueChangedEvent(int pinNumber, PinChangeEventHandler callback)
+        {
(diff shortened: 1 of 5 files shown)


src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Windows.cs
```diff
@@ -6,5 +6,16 @@ namespace System.Device.Gpio.Drivers
 {
     public partial class RaspberryPi3Driver : Windows10Driver
     {
+        protected ulong ClearRegister
+        {
+            get { throw new PlatformNotSupportedException(); }
```
joperezr: @tarekgh PTAL. Windows raspberry pi driver won't expose the device registers, so I thought that throwing would be the best option when this method is called on Windows. Do you agree?

src/System.Device.Gpio/System/Device/Gpio/Drivers/UnixDriver.Windows.cs
```diff
@@ -14,6 +14,11 @@ public UnixDriver()
             throw new PlatformNotSupportedException();
         }
 
+        public static UnixDriver Create()
+        {
+            return new UnixDriver();
```
tarekgh: >UnixDriver [](start = 23, length = 10)

is this intended to throw?

joperezr: yes, it will eventually since it will call the constructor which will throw. Are you saying I should simply throw directly instead?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Windows.cs
tarekgh: do you have any more information about the design of the pi support on Windows? maybe they can have something else.

---
In reply to: [276028609](https://github.com/dotnet/iot/pull/367#discussion_r276028609) [](ancestors = 276028609)

tarekgh: Also, RGBLedMatrix project is marked as linux only but it got compiled with the rest of devices which is wrong to do 

---
In reply to: [276032568](https://github.com/dotnet/iot/pull/367#discussion_r276032568) [](ancestors = 276032568,276028609)

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Windows.cs
```diff
@@ -0,0 +1,78 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+
+namespace System.Device.Gpio.Drivers
+{
+    public class LibGpiodDriver : UnixDriver
```
krwq: is UnixDriver much different from IGpioDriver? I'm wondering if we should instead have something like NotImplementedDriver and reuse it in places like this to reduce lines of code

krwq: Seems we only have IGpioController but no interface for driver. Same question but GpioDriver

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Windows.cs
joperezr: > RGBLedMatrix project is marked as linux only

That is ok, that library will only get compiled once for linux and windows. The main library is the one that has 2 different implementations.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Windows.cs
joperezr: > I'm wondering if we should instead have something like NotImplementedDriver and reuse it in places like this to reduce lines of code

That could work, we can address that in a different PR.
About UnixDriver vs GpioDriver, they are pretty much the same, except that in Linux UnixDriver class has some implementation and the reference assembly has to be produced out of that, which means that in here we can't derive from GpioDriver, and have to derive from UnixDriver instead in order to be compatible.

Issue #343, opened 2019-03-22
Sysfs driver pin numbering offset is different for arm64


buyaa-n:
Sysfs driver was not working in 64 bit Debian for raspberry pi. Recently find out offset numbering of gpio pins start from 458, so for example if you want to access pin 18 you need to use 458 + 18 = 476. Need to find a way to determine and set offset pin numbering withing driver so that users could use consistent numbering.  https://wiki.debian.org/RaspberryPi3

joperezr:
This needs to be fixed for 3.0. Should be easy enough to fix. Flagging as up-for-grabs since @buyaa-n won't have time to look at this any time soon. If no one else picks this up, either her or me can pick this one.

PR #386, opened 2019-04-24
Updates for supporting 64 bit linux pin nimbering in Sysfs driver

buyaa-n:
Creating PR as a draft as want to get feedback with this solution.

Fixes https://github.com/dotnet/iot/issues/343

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
@@ -15,6 +15,10 @@ namespace System.Device.Gpio.Drivers
     public class SysFsDriver : UnixDriver
     {
         private const string GpioBasePath = "/sys/class/gpio";
+        private const string GpioChip = "gpiochip";
+        private const string GpioLabel = "/label";
+        private const string GpioContoller = "pinctrl";
+        private const string GpioOffsetBase = "/base";
         private int _pollFileDescriptor = -1;
         private Thread _eventDetectionThread;
         private int _pinsToDetectEventsCount;
@@ -22,6 +26,22 @@ namespace System.Device.Gpio.Drivers
         private readonly List<int> _exportedPins = new List<int>();
         private readonly Dictionary<int, UnixDriverDevicePin> _devicePins = new Dictionary<int, UnixDriverDevicePin>();
         private readonly int _pollingTimeoutInMilliseconds = Convert.ToInt32(TimeSpan.FromMilliseconds(1).TotalMilliseconds);
+        private int _pinOffset = 0;
+
+        public SysFsDriver() {
+            IEnumerable<string> fileNames = Directory.EnumerateFileSystemEntries(GpioBasePath);
+            foreach (string name in fileNames)
+            {
+                if (name.Contains(GpioChip))
+                {
+                    if (File.ReadAllText($"{name}{GpioLabel}").StartsWith(GpioContoller))
+                    {
+                        _pinOffset = int.Parse(File.ReadAllText($"{name}{GpioOffsetBase}"));
+                        break;
+                    }
+                }
+            }
+        }
 
         /// <summary>
         /// The number of pins provided by the driver.
@@ -41,13 +61,14 @@ namespace System.Device.Gpio.Drivers
         /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
         protected internal override void OpenPin(int pinNumber)
         {
-            string pinPath = $"{GpioBasePath}/gpio{pinNumber}";
+            int pinOffset = pinNumber + _pinOffset;
+            string pinPath = $"{GpioBasePath}/gpio{pinOffset}";
             // If the directory exists, this becomes a no-op since the pin might have been opened already by the some controller or somebody else.
             if (!Directory.Exists(pinPath))
             {
                 try
                 {
-                    File.WriteAllText(Path.Combine(GpioBasePath, "export"), pinNumber.ToString());
+                    File.WriteAllText(Path.Combine(GpioBasePath, "export"), pinOffset.ToString());
                     _exportedPins.Add(pinNumber);
                 }
                 catch (UnauthorizedAccessException e)
@@ -64,14 +85,15 @@ namespace System.Device.Gpio.Drivers
         /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
         protected internal override void ClosePin(int pinNumber)
         {
-            string pinPath = $"{GpioBasePath}/gpio{pinNumber}";
+            int pinOffset = pinNumber + _pinOffset;
+            string pinPath = $"{GpioBasePath}/gpio{pinOffset}";
             // If the directory doesn't exist, this becomes a no-op since the pin was closed already.
             if (Directory.Exists(pinPath))
             {
                 try
                 {
                     SetPinEventsToDetect(pinNumber, PinEventTypes.None);
-                    File.WriteAllText(Path.Combine(GpioBasePath, "unexport"), pinNumber.ToString());
+                    File.WriteAllText(Path.Combine(GpioBasePath, "unexport"), pinOffset.ToString());
                     _exportedPins.Remove(pinNumber);
                 }
                 catch (UnauthorizedAccessException e)
@@ -92,7 +114,7 @@ namespace System.Device.Gpio.Drivers
             {
                 throw new PlatformNotSupportedException("This driver is generic so it does not support Input Pull Down or Input Pull Up modes.");
             }
-            string directionPath = $"{GpioBasePath}/gpio{pinNumber}/direction";
+            string directionPath = $"{GpioBasePath}/gpio{pinNumber + _pinOffset}/direction";
             string sysFsMode = ConvertPinModeToSysFsMode(mode);
             if (File.Exists(directionPath))
             {
@@ -146,7 +168,7 @@ namespace System.Device.Gpio.Drivers
         protected internal override PinValue Read(int pinNumber)
         {
             PinValue result = default;
-            string valuePath = $"{GpioBasePath}/gpio{pinNumber}/value";
+            string valuePath = $"{GpioBasePath}/gpio{pinNumber + _pinOffset}/value";
             if (File.Exists(valuePath))
             {
                 try
@@ -193,7 +215,7 @@ namespace System.Device.Gpio.Drivers
         /// <param name="value">The value to be written to the pin.</param>
         protected internal override void Write(int pinNumber, PinValue value)
         {
-            string valuePath = $"{GpioBasePath}/gpio{pinNumber}/value";
+            string valuePath = $"{GpioBasePath}/gpio{pinNumber + _pinOffset}/value";
             if (File.Exists(valuePath))
             {
                 try
@@ -257,14 +279,14 @@ namespace System.Device.Gpio.Drivers
 
         private void SetPinEventsToDetect(int pinNumber, PinEventTypes eventTypes)
         {
-            string edgePath = Path.Combine(GpioBasePath, $"gpio{pinNumber}", "edge");
+            string edgePath = Path.Combine(GpioBasePath, $"gpio{pinNumber + _pinOffset}", "edge");
             string stringValue = PinEventTypeToStringValue(eventTypes);
             File.WriteAllText(edgePath, stringValue);
         }
 
         private PinEventTypes GetPinEventsToDetect(int pinNumber)
         {
-            string edgePath = Path.Combine(GpioBasePath, $"gpio{pinNumber}", "edge");
+            string edgePath = Path.Combine(GpioBasePath, $"gpio{pinNumber + _pinOffset}", "edge");
             string stringValue = File.ReadAllText(edgePath);
             return StringValueToPinEventType(stringValue);
         }
@@ -323,7 +345,7 @@ namespace System.Device.Gpio.Drivers
 
             if (valueFileDescriptor == -1)
             {
-                string valuePath = Path.Combine(GpioBasePath, $"gpio{pinNumber}", "value");
+                string valuePath = Path.Combine(GpioBasePath, $"gpio{pinNumber + _pinOffset}", "value");
                 valueFileDescriptor = Interop.open(valuePath, FileOpenFlags.O_RDONLY | FileOpenFlags.O_NONBLOCK);
                 if (valueFileDescriptor < 0)
                 {
@@ -367,7 +389,7 @@ namespace System.Device.Gpio.Drivers
                 }
                 if (waitResult > 0)
                 {
-                    pinNumber = events.data.pinNumber;
+                    pinNumber = events.data.pinNumber - _pinOffset;
 
                     // valueFileDescriptor will be -1 when using the callback eventing. For WaitForEvent, the value will be set.
                     if (valueFileDescriptor == -1)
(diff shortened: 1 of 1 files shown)


src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -15,13 +15,33 @@ namespace System.Device.Gpio.Drivers
...
+        private const string GpioContoller = "pinctrl";
+        private const string GpioOffsetBase = "/base";
         private int _pollFileDescriptor = -1;
         private Thread _eventDetectionThread;
         private int _pinsToDetectEventsCount;
         private readonly static CancellationTokenSource s_eventThreadCancellationTokenSource = new CancellationTokenSource();
         private readonly List<int> _exportedPins = new List<int>();
         private readonly Dictionary<int, UnixDriverDevicePin> _devicePins = new Dictionary<int, UnixDriverDevicePin>();
         private readonly int _pollingTimeoutInMilliseconds = Convert.ToInt32(TimeSpan.FromMilliseconds(1).TotalMilliseconds);
+        private int _pinOffset = 0;
```
tarekgh: >= 0 [](start = 31, length = 3)

I don't think we need to have this explicit initialization to 0.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -15,13 +15,33 @@ namespace System.Device.Gpio.Drivers
...
+
+        public SysFsDriver() {
```
tarekgh: >{ [](start = 29, length = 1)

please fix the formatting.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -15,13 +15,33 @@ namespace System.Device.Gpio.Drivers
...
+            IEnumerable<string> fileNames = Directory.EnumerateFileSystemEntries(GpioBasePath);
+            foreach (string name in fileNames)
+            {
+                if (name.Contains(GpioChip))
+                {
+                    if (File.ReadAllText($"{name}{GpioLabel}").StartsWith(GpioContoller))
```
tarekgh: >StartsWith [](start = 63, length = 10)

should you add StringComparison.Ordinal? This is more performant.

tarekgh: >File [](start = 24, length = 4)

should we probe for the files existence before reading them?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
tarekgh: >SysFsDriver [](start = 15, length = 11)

what happen if we enumerated ass items under GpioBasePath and didn't find any? should we throw?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -64,14 +85,15 @@ protected internal override void OpenPin(int pinNumber)
...
+            int pinOffset = pinNumber + _pinOffset;
+            string pinPath = $"{GpioBasePath}/gpio{pinOffset}";
             // If the directory doesn't exist, this becomes a no-op since the pin was closed already.
             if (Directory.Exists(pinPath))
             {
                 try
                 {
                     SetPinEventsToDetect(pinNumber, PinEventTypes.None);
-                    File.WriteAllText(Path.Combine(GpioBasePath, "unexport"), pinNumber.ToString());
+                    File.WriteAllText(Path.Combine(GpioBasePath, "unexport"), pinOffset.ToString());
```
tarekgh: >ToString() [](start = 88, length = 10)

nit: maybe better to pass CultureInfo.InvariantCulture to ToString?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -367,7 +389,7 @@ private unsafe bool WasEventDetected(int pollFileDescriptor, int valueFileDescri
                 }
                 if (waitResult > 0)
                 {
-                    pinNumber = events.data.pinNumber;
+                    pinNumber = events.data.pinNumber - _pinOffset;
```
tarekgh: >pinNumber [](start = 20, length = 9)

could you please add Assert for not getting negative values?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -15,13 +15,33 @@ namespace System.Device.Gpio.Drivers
...
+
+        public SysFsDriver() {
+            IEnumerable<string> fileNames = Directory.EnumerateFileSystemEntries(GpioBasePath);
+            foreach (string name in fileNames)
+            {
+                if (name.Contains(GpioChip))
+                {
+                    if (File.ReadAllText($"{name}{GpioLabel}").StartsWith(GpioContoller))
+                    {
+                        _pinOffset = int.Parse(File.ReadAllText($"{name}{GpioOffsetBase}"));
```
joperezr: Should we have a tryParse here and handle the case where it fails? Should we trim the string we get back? I also wonder if we should optimize this file and stop using File.ReadAllText and WriteAllText into something more performant.

joperezr:
Is this only applicable on x64? And is the offset always the same? If so, I would rather use RuntimeInformation to see if we are running on x64 and if so just set the offset, that way the code is cleaner and we don't have to enumerate and parse files.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
krwq: I think it's unexpected for the file to not exist so I think error would be fine if that happened

tarekgh: Yes but I believe we should throw describable exception and not just having the caller see IO Exception for a file not exist while not having any idea why getting that.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
buyaa-n: Could you share the reason? 
For me if no any offset number found it should be 0 as this number is added to pin number for any further functionality.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
buyaa-n: not sure about that, i checked these path only on Raspberry pi with 32, 64bit installed, guess we could still try with the default pin numbering

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
buyaa-n: With several Linux versions (Rasbian, Debian) on Raspberry Pi i have tested, these path existed, but cannot guarantee the same for all linux versions. So agree with checking file existance, but don't agree with throwing, I think we should continue with default `0` offset in that case

tarekgh: ok, just write a comment to clarify that. Thanks.

---
In reply to: [278330737](https://github.com/dotnet/iot/pull/386#discussion_r278330737) [](ancestors = 278330737)

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
tarekgh: >_pinOffset [](start = 24, length = 10)

we are reading the same value all the time. I think it would be better to convert this to a static value that you read it once.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
joperezr: ints are value types, so they will always get initialized with 0.

buyaa-n: yeah I know that, just didn't think from that side 😛

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
buyaa-n: sure

buyaa-n:
> Is this only applicable on x64? And is the offset always the same? If so, I would rather use RuntimeInformation to see if we are running on x64 and if so just set the offset, that way the code is cleaner and we don't have to enumerate and parse files.

on version of 64 bit Debian i had installed on Rasbian had normal pin numbering (i mean no offset, same as 32 bit), couldn't guarantee for any 64 bit linux

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
buyaa-n: it is reading once in constuctor, will make it static.

buyaa-n: > Should we have a tryParse here and handle the case where it fails? Should we trim the string we get back? 

Yes, thanks

>  I also wonder if we should optimize this file and stop using File.ReadAllText and WriteAllText into something more performant.

Not sure about that ...

tarekgh: I would say, just wrap the whole code inside the constructor inside try-catch and then don't worry about checking the files or parsing the number.

krwq:
@buyaa-n should this fix be scoped to ARM64/RaspberryPi then?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
buyaa-n: turns out we don't need to subtract this value, this pin number is not set from device it is set here: https://github.com/dotnet/iot/pull/386/files/d8d60bade8e87c0d82510f04525e23656949f798#diff-f9b9a68ada4ceabc2a9b052f1c52a2cbR362 where offset address not added

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
buyaa-n: pin numbers are integer and would not be scaled so high, do you think its better to have CultureInfo.InvariantCulture or it parses faster with that culture?

buyaa-n:
@krwq the goal is to cover any unix driver numbering variance, just havent had tried with any other

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
tarekgh: Please use InvariantCulture. it is faster.

buyaa-n: sure, thank you

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -15,13 +16,40 @@ namespace System.Device.Gpio.Drivers
...
+        private const string GpioContoller = "pinctrl";
+        private const string GpioOffsetBase = "/base";
         private int _pollFileDescriptor = -1;
         private Thread _eventDetectionThread;
         private int _pinsToDetectEventsCount;
         private readonly static CancellationTokenSource s_eventThreadCancellationTokenSource = new CancellationTokenSource();
         private readonly List<int> _exportedPins = new List<int>();
         private readonly Dictionary<int, UnixDriverDevicePin> _devicePins = new Dictionary<int, UnixDriverDevicePin>();
         private readonly int _pollingTimeoutInMilliseconds = Convert.ToInt32(TimeSpan.FromMilliseconds(1).TotalMilliseconds);
+        private static int _pinOffset;
```
krwq: this is a static field but you set it in the constructor - this doesn't sound right even if the read value will be the same. You should probably use static constructor or even better - field initializer

tarekgh: just make it delay initialized.

krwq: do you mean `int?` or `Lazy<int>`? I think field initializer will also lazy initialize on first usage

tarekgh: both can work. 

> I think field initializer will also lazy initialize on first usage

That is true too. 

```C#
private static int _pinOffset = ReadOffset();

private static int ReadOffset()
{
    // write the reading code
}
```

buyaa-n:
@tarekgh @krwq @joperezr comments addressed, please re review

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -15,13 +16,41 @@ namespace System.Device.Gpio.Drivers
...
+            IEnumerable<string> fileNames = Directory.EnumerateFileSystemEntries(GpioBasePath);
+            foreach (string name in fileNames)
+            {
+                if (name.Contains(GpioChip))
+                {
+                    try
+                    {
+                        if (File.ReadAllText($"{name}{GpioLabel}").StartsWith(GpioContoller, StringComparison.Ordinal))
+                        {
+                            if (int.TryParse(File.ReadAllText($"{name}{GpioOffsetBase}"), out _pinOffset))
```
tarekgh: >_pinOffset [](start = 94, length = 10)

please use a local variable here instead of _pinOffset 

                            if (int.TryParse(File.ReadAllText($"{name}{GpioOffsetBase}"), out int pinOffset))
                                return pinOffset;

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -15,13 +16,41 @@ namespace System.Device.Gpio.Drivers
...
+        private const string GpioContoller = "pinctrl";
+        private const string GpioOffsetBase = "/base";
         private int _pollFileDescriptor = -1;
         private Thread _eventDetectionThread;
         private int _pinsToDetectEventsCount;
         private readonly static CancellationTokenSource s_eventThreadCancellationTokenSource = new CancellationTokenSource();
         private readonly List<int> _exportedPins = new List<int>();
         private readonly Dictionary<int, UnixDriverDevicePin> _devicePins = new Dictionary<int, UnixDriverDevicePin>();
         private readonly int _pollingTimeoutInMilliseconds = Convert.ToInt32(TimeSpan.FromMilliseconds(1).TotalMilliseconds);
+        private static int _pinOffset = ReadOffset();
```
tarekgh: >_pinOffset [](start = 27, length = 10)

please name this as s_pinOffset

PR #407, opened 2019-05-05
Update to accept content from CLI

richlander:


joperezr:
Lgtm. Can we update the sample’s readme file to say that you can now pass in the message you want to print?

PR #421, opened 2019-05-15
CharacterLCD PCF8574T I2C Sample

monty124:
Hi All

finally got around to adding the sample to the ReadMe and working sample code

@krwq  I think what i've done here is right! if not please do let me know where I've gone wrong!

There is another LCD i will be testing but I'll need a bit of time to do that.


src/devices/CharacterLcd/samples/PCF8574T Sample/PCF8574T I2C LCD Sample.sln
```diff
@@ -0,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
krwq: I'd prefer if you've replaced all spaces in file and directory names with dashes

src/devices/CharacterLcd/samples/PCF8574T Sample/PCF8574T Test/Program.cs
```diff
@@ -0,0 +1,357 @@
...
+using System.Device.I2c;
+using System.Device.I2c.Drivers;
+using System.Diagnostics;
+using System.Drawing;
+using System.Text;
+using System.Timers;
+using Iot.Device.Mcp23xxx;
+using Iot.Device.CharacterLcd;
+
+
```
krwq: nit: extra enters

src/devices/CharacterLcd/samples/PCF8574T Sample/PCF8574T Test/Program.cs
```diff
@@ -0,0 +1,357 @@
...
+
+namespace test
+{
+    class Program
+    {
+        public class ExtendedSample
```
krwq: is this sample mostly the same as original? (except for using extender) If so, it might make sense to `#ifdef` those few lines of code and only create a separate csproj in the same folder which defines the constant or just extract one method which differs to separate file and include file which differs in respective project

src/devices/CharacterLcd/samples/PCF8574T Sample/PCF8574T I2C LCD Sample.sln
joperezr: Apart from removing spaces on path names, can we also just remove the solution?

src/devices/CharacterLcd/samples/PCF8574T Sample/PCF8574T Test/Program.cs
joperezr: Instead of creating new project, why not just add a new file to the samples project that already exists, named Pcf8574tSample.cs and have this code there? To me there is very low value on having two different sample projects.

monty124: hi guys

I created a new solution as it was the only efficient way to test the numerous itinerations of changes i had to do to get this working. but esentially yes this is the extended sample with correct pin assignments for the PCF IC.

sure i can put just a sample file there, do i just make the changes and re push to my fork? or does a new PR need to be made too? 

@krwq I'm by no means a developer this is a little above my head i'd probably need help here if you wanted to go this route.

the only other suggestion i can make is omit the sample and just have the correct definition in the readme? essentially the code is all the same, you just need to change the setup for the lcd really.

re: spaces in file names and extra crlf thats easy enough fixed

joperezr: > sure i can put just a sample file there, do i just make the changes and re push to my fork? or does a new PR need to be made too?

Correct, all you have to o is to make the changes locally, commit them to your local branch, and then push to your fork. This PR will be auto updated with those changes so that we can take another look 😄. Feel free to ask any questions about the process, we are happy to help.

monty124: Hi @joperezr 

OK, I've changed the sample and it should be part of the current sample project; its named Pcf8574tSample.cs.

The extra crlf's should be all fixed too.

let me know if you think it needs anything else or any suggestions on improvements.

src/devices/CharacterLcd/samples/Pcf8574tSample.cs
```diff
@@ -0,0 +1,353 @@
...
+using System.Diagnostics;
+using System.Drawing;
+using System.Text;
+using System.Timers;
+using Iot.Device.Mcp23xxx;
+using Iot.Device.CharacterLcd;
+
+namespace Iot.Device.CharacterLcd.Samples
+{
+    class Pcf8574tSample
```
joperezr: Can we remove the nesting of classes? I would only keep one class called Pcf8574tSample with the main method inside.

src/devices/CharacterLcd/samples/Pcf8574tSample.cs
```diff
@@ -0,0 +1,353 @@
...
+            const string Twenty = "123456789\u0008123456789\u0009";
+            const string Thirty = Twenty + "123456789\u000a";
+            const string Fourty = Thirty + "123456789\u000b";
+            const string Eighty = Fourty + "123456789\u000c123456789\u000d123456789\u000e123456789\u000f";
+
+            static void Main()
+            {
+                Console.WriteLine("Starting...");
+
+            var i2cDevice = new UnixI2cDevice(new I2cConnectionSettings(busId: 1, deviceAddress: 0x27));
```
joperezr: Indenting, seems to be incorrect here.

src/devices/CharacterLcd/samples/Pcf8574tSample.cs
```diff
@@ -0,0 +1,353 @@
...
+using Iot.Device.Mcp23xxx;
+using Iot.Device.CharacterLcd;
+
+namespace Iot.Device.CharacterLcd.Samples
+{
+    class Pcf8574tSample
+    {
+        public class ExtendedSample
+        {
+            const string Twenty = "123456789\u0008123456789\u0009";
```
joperezr: Do you mind adding a comment here explaining what are these supposed to be?

joperezr:
Left a few comments but this is looking great, thanks for adding it!

src/devices/CharacterLcd/samples/Pcf8574tSample.cs
monty124: this is not my code; i copied it from Hd44780.ExtendedSample.cs into a new project to debug; the original author would be best placed to comment here

monty124:
suggestions fixed/implemented 
re: nested classes this was a new-> project using VS2017; fixed in any case

monty124:
if the changes are all ok, whats the next step?

joperezr:
This LGTM. THanks a lot for following up @monty124 and for the contribution.

PR #430, opened 2019-05-17
Try to fix DHT 1-wire read error

### LINKED ISSUE #416: Dht22 example doesn't work
I have DHT22 sensors which works normally. I checked that by the default python script from Adafriut. I made a decision to give .net core on rpi platform a try with a simple example from the examples folder. I used Dhtxx example but it doesn't work.

This is my code:
```
using System;
using System.Threading;
using Iot.Device.DHTxx;

class Program
{
    {

        using (DhtSensor dht = new DhtSensor(26, DhtType.Dht22))
        {
            while (true)
            {
                Console.WriteLine($"Temperature: {dht.Temperature.Celsius} °C, Humidity: {dht.Humidity} %");

                Thread.Sleep(2000);
            }
        }
    }
}
```

Project file:
```

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.2</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="IoT.Device.Bindings" Version="0.1.0-prerelease.19251.2" />
    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease.19251.2" />
  </ItemGroup>

</Project>
```
I wasn't able to install System.Device.Gpio and IoT.Device.Bindings by using commands from the main readme.md so I just used the nearest proposed version by dotnet add.

I installed .net core on the rpi by using commands:
```
sudo apt-get install curl libunwind8 gettext
wget https://download.visualstudio.microsoft.com/download/pr/9650e3a6-0399-4330-a363-1add761127f9/14d80726c16d0e3d36db2ee5c11928e4/dotnet-sdk-2.2.102-linux-arm.tar.gz
wget https://download.visualstudio.microsoft.com/download/pr/9d049226-1f28-4d3d-a4ff-314e56b223c5/f67ab05a3d70b2bff46ff25e2b3acd2a/aspnetcore-runtime-2.2.1-linux-arm.tar.gz
mkdir -p $HOME/dotnet && tar zxf dotnet-sdk-2.2.102-linux-arm.tar.gz -C $HOME/dotnet
export DOTNET_ROOT=$HOME/dotnet 
export PATH=$PATH:$HOME/dotnet
tar zxf aspnetcore-runtime-2.2.1-linux-arm.tar.gz -C $HOME/dotnet
```
On top of that I wasn't able to run it on rpi without `dotnet publish -r linux-arm` since I get FileNotFoundExceptions and runtime isn't able to find System.Device.Gpio.

The result of the program:
```
pi@raspberrypi:~/publish $ dotnet test_dht22.dll
Hello DHT!
Temperature: NaN °C, Humidity: NaN %
Temperature: NaN °C, Humidity: NaN %
Temperature: NaN °C, Humidity: NaN %

```

I tried other library versions and sometimes got values like 123e-308.

Do I do something wrong or the project isn't ready yet for real usage (only people with deep .net knowledge can use it)?

ZhangGaoxing:
try to fix #416 

try this @katzz0 @MarkCiliaVincenti

It's OK for me.
360截图20190517201547169


src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -173,15 +173,31 @@ private bool ReadThroughOneWire()
             // pull up data line
             _controller.Write(_pin, PinValue.High);
             // wait 20 - 40 microseconds
-            DelayHelper.DelayMicroseconds(40, true);
+            DelayHelper.DelayMicroseconds(30, true);
```
joperezr: Is this wait or timing offset specified in the datasheet? If so it would be great to use the value they provide.

ZhangGaoxing: Waiting time is between 20 - 40 microseconds. If I write 40 microseconds in the code, the actual waiting time will be greater than 40 microseconds. The extra microseconds will wait in the next step loop.

joperezr: I see that makes sense. I don't have the hardware here to test it, but given that seems like the current implementation wasn't really working and this seems to be fixed for you, I'll go ahead and merge this.

joperezr:
@katzz0 @MarkCiliaVincenti Once we have a new version of the bindings package, do you guys mind checking if this fixes your issues?

MarkCiliaVincenti:
Sure @joperezr, just let me know when.

joperezr:
@MarkCiliaVincenti Iot.Device.Bindings package version 0.1.0-prerelease.19271.3 should now be in our blob feed, and that one should have the fix in it. In order to add our blob feed to your restore sources, simply drop a NuGet.config file on the same directory as your .csproj that looks something like:

```xml
<configuration>
  <packageSources>
    <add key="dotnet-iot" value="https://dotnetfeed.blob.core.windows.net/dotnet-iot/index.json" />
  </packageSources>
</configuration>
```

MarkCiliaVincenti:
Sorry @joperezr but I can't really test this. For some reason the code reverts to the previous NuGet package whenever I rebuild. Possibly has to do with this using a preview version of .NET Core 3, which I'm not keen on using.

joperezr:
> For some reason the code reverts to the previous NuGet package whenever I rebuild

Sounds like that would be happening if you weren't using the source I suggested above. Can you try removing your bin and obj folders, and then calling `dotnet restore yourProject.csproj -s https://dotnetfeed.blob.core.windows.net/dotnet-iot/index.json`? That should in theory force the use of the blob feed which will have the version I suggested. Also, you shouldn't need to use .NET Core 3 preview, by restoring the project as I suggest, you will only be using a new version of our package, but the rest of the framework will still remain whatever version of the sdk you have installed.

MarkCiliaVincenti:
I'm doing it through the package manager GUI in VS2019 and if I try to upgrade to the latest version it tells me it also needs to upgrade to .NET Core 3 preview.

RobinTTY:
Just got around to testing this (on linux-arm) and here are my results:

SensorTest

There is a pause of 1 second between each measurement. The device used was a dht11. It might be interesting to get some other results since it might just be my sensor getting a lot of invalid readings. System.Device.Gpio version used is 19274.1.

YarekTyshchenko:
This is not fixed. The delays are obviously different on different hardware. They should be exposed via configuration

PR #434, opened 2019-05-17
use grovepi+ terminology

CleoQc:
As Seeed came up with the Grove Base Hat that also works with the Pi, we've seen user confusion. I changed the text a bit to make it clearer.

Also I'm not too sure what is meant on line 26. *multiple versions of GrovePi*  or *multiple versions of GrovePi firmware * ?

### DIFF
--- a/src/devices/GrovePi/README.md
+++ b/src/devices/GrovePi/README.md
@@ -2,12 +2,12 @@
 
 ## Summary
 
-You will need a GrovePi from [Dexter Industries](https://www.dexterindustries.com/grovepi/) and a Raspberry Pi. That said, any board will work with the GrovePi as it is using I2C. You’ll just need to link the pins from your board to the GrovePi ones the correct way.
+You will need a GrovePi+ from [Dexter Industries](https://www.dexterindustries.com/grovepi/) and a Raspberry Pi. That said, any board will work with the GrovePi+ as it is using I2C. You’ll just need to link the pins from your board to the GrovePi+ ones the correct way.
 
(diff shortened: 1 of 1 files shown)


src/devices/GrovePi/README.md
```diff
@@ -2,12 +2,12 @@
...
+You will need a GrovePi+ from [Dexter Industries](https://www.dexterindustries.com/grovepi/) and a Raspberry Pi. That said, any board will work with the GrovePi+ as it is using I2C. You’ll just need to link the pins from your board to the GrovePi+ ones the correct way.
 
-GrovePi have the ability to use Grove sensors, analogic, digital. GrovePi provide as well extensions for I2C sensors. In order to take advantage of GrovePi, you'll need Grove compatible sensors. There are lots existing like on [Seeed Studio](http://wiki.seeedstudio.com/Grove/).
+GrovePi+ have the ability to use Grove sensors, analogic, digital. GrovePi+ provide as well extensions for I2C sensors. In order to take advantage of GrovePi, you'll need Grove compatible sensors. There are lots existing like on [Seeed Studio](http://wiki.seeedstudio.com/Grove/).
 
 - [Device family](./README.md#device-family)
-- [GrovePi requirements](./README.md#make-sure-you-have-a-GrovePi)
+- [GrovePi requirements](./README.md#make-sure-you-have-a-GrovePi+)
```
krwq: Should this also be `GrovePi+ requirements` and similarly below (`Accessing GrovePi+ information`)?

CleoQc: Only if you want to be picky about it. 
I flagged the places that were specific about identifying the platform.  Once it's identified, it's fine to just call it GrovePi. We do. :)

PR #435, opened 2019-05-18
BME280 - Humidity Support

### LINKED ISSUE #399: Add Hummidity and Compensate Hummidity to the BME280 device binding.
For BME280, you can read the pressure and the temperature (including compensate values) but so far there's nothing for humidity. Any chance for this to be implemented please?

Silvenga:
This pr adds support for humidity for BME280 devices.

- Moved the responsibility of reading calibration data to the implemented devices, away from `CalibrationData`.
- Added additional registers.
- Added method to read humidity data and to apply calibrations.

This resolves #399.

### DIFF
--- a/src/devices/Bmx280/Bmx280.csproj
+++ b/src/devices/Bmx280/Bmx280.csproj
@@ -11,6 +11,7 @@
     <Compile Include="Bmp280.cs" />
     <Compile Include="BmxBase.cs" />
     <Compile Include="CalibrationData.cs" />
+    <Compile Include="FilteringMode.cs" />
     <Compile Include="PowerMode.cs" />
     <Compile Include="Register.cs" />
     <Compile Include="Sampling.cs" />
--- a/src/devices/Bmx280/Bmx280.csproj
+++ b/src/devices/Bmx280/Bmx280.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <EnableDefaultItems>false</EnableDefaultItems>
+    <RootNamespace>Iot.Device.Bmx280</RootNamespace>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/devices/Bmx280/Bmx280.csproj
+++ b/src/devices/Bmx280/Bmx280.csproj
@@ -16,6 +16,7 @@
     <Compile Include="Register.cs" />
     <Compile Include="Sampling.cs" />
     <Compile Include="DeviceStatus.cs" />
+    <Compile Include="StandbyTime.cs" />
     <PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
     <ProjectReference Include="..\Units\Units.csproj" />
   </ItemGroup>
--- a/src/devices/Bmx280/Bmx280.csproj
+++ b/src/devices/Bmx280/Bmx280.csproj
@@ -14,6 +14,7 @@
     <Compile Include="PowerMode.cs" />
     <Compile Include="Register.cs" />
     <Compile Include="Sampling.cs" />
+    <Compile Include="DeviceStatus.cs" />
     <PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
     <ProjectReference Include="..\Units\Units.csproj" />
   </ItemGroup>
--- a/src/devices/Bmx280/BmxBase.cs
+++ b/src/devices/Bmx280/BmxBase.cs
@@ -61,7 +61,6 @@ namespace Iot.Device.Bmx280
             status = (byte)(status & 0b1111_1100);
             status = (byte)(status | (byte)powerMode);
             _i2cDevice.Write(new[] { (byte)Register.CTRL_MEAS, status });
-
         }
 
         /// <summary>
@@ -325,6 +324,15 @@ namespace Iot.Device.Bmx280
             var2 = ((long)CalibrationData.DigP8 * p) >> 19;
             p = ((p + var1 + var2) >> 8) + ((long)CalibrationData.DigP7 << 4);
             return p;
+        }
+
+        /// <summary>
+        /// When called, the device is reset using the complete power-on-reset procedure.
+        /// </summary>
+        public void Reset()
+        {
+            const byte resetCommand = 0xb6;
+            _i2cDevice.Write(new[] { (byte)Register.RESET, resetCommand });
         }
 
         /// <summary>
--- a/src/devices/Bmx280/BmxBase.cs
+++ b/src/devices/Bmx280/BmxBase.cs
@@ -356,6 +356,40 @@ namespace Iot.Device.Bmx280
         }
 
         /// <summary>
+        /// Sets the IIR filter mode.
+        /// </summary>
+        /// <param name="filteringMode"></param>
+        public void SetFilterMode(FilteringMode filteringMode)
+        {
+            byte current = Read8BitsFromRegister((byte)Register.CONFIG);
+            current = (byte)((current & 0b1110_0011) | (byte)filteringMode << 2);
+            _i2cDevice.Write(new[] { (byte)Register.CONFIG, current });
+        }
+
+        /// <summary>
+        /// Reads the current IIR filter mode the device is running in.
+        /// </summary>
+        /// <returns></returns>
+        public FilteringMode ReadFilterMode()
+        {
+            byte current = Read8BitsFromRegister((byte)Register.CONFIG);
+            var mode = (byte)((current & 0b0001_1100) >> 2);
+            switch (mode)
+            {
+                case 0b000:
+                    return FilteringMode.Off;
+                case 0b001:
+                    return FilteringMode.X2;
+                case 0b010:
+                    return FilteringMode.X4;
+                case 0b011:
+                    return FilteringMode.X8;
+                default:
+                    return FilteringMode.X16;
+            }
+        }
+
+        /// <summary>
         ///  Reads an 8 bit value from a register
         /// </summary>
         /// <param name="register">
--- a/src/devices/Bmx280/BmxBase.cs
+++ b/src/devices/Bmx280/BmxBase.cs
@@ -390,6 +390,48 @@ namespace Iot.Device.Bmx280
         }
 
         /// <summary>
+        /// Sets the standby time mode the device will used when operating in normal mode.
+        /// </summary>
+        /// <param name="standbyTime"></param>
+        public void SetStandbyTime(StandbyTime standbyTime)
+        {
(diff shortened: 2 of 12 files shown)


Silvenga:
Hopefully did everything correctly. First time reading data sheets and messing with iot sensors. Plus it turns out to be almost midnight here. 😸

Silvenga:
And whoa, that code was really bad. Re-read the data sheets and fixed the bugs. Compared against the python version for correctness.

src/devices/Bmx280/samples/Bme280.sample.cs
```diff
@@ -23,53 +23,70 @@ static async Task Main(string[] args)
             //set this to the current sea level pressure in the area for correct altitude readings
             const double defaultSeaLevelPressure = 1033.00;
 
-            var i2cSettings = new I2cConnectionSettings(busId, Bme280.DefaultI2cAddress);
+            var i2cSettings = new I2cConnectionSettings(busId, Bme280.DefaultI2cAddress + 1);
```
joperezr: I did also find that the default I2C Address is incorrect. Do you mind changing the constant on the same PR to use 0x77 instead?

src/devices/Bmx280/samples/Bme280.sample.cs
```diff
@@ -23,53 +23,70 @@ static async Task Main(string[] args)
...
             var i2cDevice = new UnixI2cDevice(i2cSettings);
             var i2CBmpe80 = new Bme280(i2cDevice);
 
             using (i2CBmpe80)
             {
                 while (true)
                 {
-                    ////set mode forced so device sleeps after read
+                    Console.WriteLine("Let's get this party started!");
```
joperezr: NIT: Nice! but we should probably remove it :)

src/devices/Bmx280/Bmx280.csproj
```diff
@@ -3,16 +3,19 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <EnableDefaultItems>false</EnableDefaultItems>
+    <RootNamespace>Iot.Device.Bmx280</RootNamespace>
```
joperezr: Why did you need to add this?

src/devices/Bmx280/BmxBase.cs
```diff
@@ -13,13 +13,14 @@
...
-    public class BmxBase : IDisposable
+    public abstract class BmxBase : IDisposable
     {
         internal I2cDevice _i2cDevice;
         internal byte _deviceId;
         internal bool _initialized = false;
         internal CommunicationProtocol _communicationProtocol;
-        internal CalibrationData _calibrationData;
+
+        internal CalibrationData CalibrationData { get; private set; }
```
joperezr: Why do we want to change it to a property instead of keeping it as a member of the class?

src/devices/Bmx280/BmxBase.cs
```diff
@@ -305,24 +307,87 @@ private long CompensatePressure(int adcPressure)
...
+        /// <summary>
+        /// Get the current status of the device.
+        /// </summary>
+        /// <returns></returns>
+        public DeviceStatus ReadStatus()
+        {
+            var status = Read8BitsFromRegister((byte)Register.STATUS);
+
+            // Bit 3
+            var measuring = (status & 0x0000_1000) >> 3 == 1;
```
joperezr: NIT: You could do the shift of status first and then just make & 1 check.

src/devices/Bmx280/BmxBase.cs
```diff
@@ -305,24 +307,87 @@ private long CompensatePressure(int adcPressure)
...
+
+            return new DeviceStatus
+            {
+                ImageUpdating = imUpdate,
+                Measuring = measuring
+            };
+        }
+
+        /// <summary>
+        /// Sets the IIR filter mode.
```
joperezr: Do you mind explaining what IIR means?

src/devices/Bmx280/Register.cs
```diff
@@ -23,14 +23,25 @@ internal enum Register : byte
...
+        RESET = 0xE0,
 
         CAL26 = 0xE1,  // R calibration stored in 0xE1-0xF0
 
+        // BME280 Only
+        CTRL_HUM = 0xF2,
+
         STATUS = 0xF3,
-        CONTROL = 0xF4,
+        CTRL_MEAS = 0xF4,
```
joperezr: I suppose this is the name in the DataSheet?

src/devices/Bmx280/DeviceStatus.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Iot.Device.Bmx280
+{
+    public class DeviceStatus
```
joperezr: Do you mind adding doc comments to this class and to the FilteringMode as well?

joperezr:
Left a few comments but overall this looks great! Thanks so much for adding this @Silvenga

src/devices/Bmx280/samples/Bme280.sample.cs
Silvenga: Oops, didn't mean to commit that. I'll update the constant.

src/devices/Bmx280/samples/Bme280.sample.cs
Silvenga: I did that mainly so I could see what loop we were on (I was using the sample app to test with). I'll pull it out.

src/devices/Bmx280/Bmx280.csproj
Silvenga: That was to ensure that newly created/refactored files used the correct namespace. Otherwise the default, defaults to just `Bmx280` IIRC. This also makes static analysis happy (Resharper really wanted to fix that namespace).

Silvenga: Should I pull it out?

src/devices/Bmx280/BmxBase.cs
Silvenga: I wanted `set` restrictions on this field (hence the property refactor). 

The base class owns `_calibrationData`, but implementations need read access too. Since the `Init` operation that creates  `_calibrationData` is both owned by `BmxBase` and is lazily initialized by it, it made sense to me to refactor `_calibrationData` to be a property on `BmxBase`.

src/devices/Bmx280/Register.cs
Silvenga: Yep.

src/devices/Bmx280/BmxBase.cs
Silvenga: It's what the datasheet uses to refer to the internal logic that reduces noise at the cost of accuracy - Infinite impulse response (IIR). The datasheet talks about different recommended filter settings based on usage. The IIR circuits are only used for temperature and and pressure.

src/devices/Bmx280/DeviceStatus.cs
Silvenga: Will do.

Silvenga:
Thanks @joperezr - also had a question about the timing intervals. I wanted to add support, but I have no idea how to name the enum members. This config flag controls the detection/sleep interval when in `normal` mode. The values are kind of odd to build into an enum. I wanted to use an abstraction like the power saving enum for over sampling, but can't find a compatible name.
```
0.5 (ReallyFast)
62.5ms (KindOfFast)
125ms (LessFastThenKindOfFastButSlowerThanFast)
250ms (Fast)
...
10ms
20ms
```

I was think we could alternatively support a dictionary of possible values (with the mapping to their byte version), and have the user pass in a ms that they want to use, but that's not exactly intuitive.

src/devices/Bmx280/Bmx280.csproj
joperezr: Oh ok, no it's fine to leave it. I was just wondering so that if it was useful we would add it to the binding template. Sounds like it may be good to add it there at some point.

src/devices/Bmx280/BmxBase.cs
joperezr: oh ok, that makes sense.

joperezr:
Regarding your timing intervals question, does the DataSheet provide a naming for the different intervals? If not, what about having the enum literally tell what the delay is? Another option would be to have this be a TimeSpan property and when set, we would simply kind of round to the closest one that the hardware allows, although I'm not a fan of that approach.

Silvenga:
@joperezr I think I got everything. I added support for standby times - the enum is a little odd. I also added  the licensed header that was missing from a couple of files.

src/devices/Bmx280/samples/Bme280.sample.cs
```diff
@@ -23,53 +23,68 @@ static async Task Main(string[] args)
             //set this to the current sea level pressure in the area for correct altitude readings
             const double defaultSeaLevelPressure = 1033.00;
 
-            var i2cSettings = new I2cConnectionSettings(busId, Bme280.DefaultI2cAddress);
+            var i2cSettings = new I2cConnectionSettings(busId, Bme280.DefaultI2cAddress + 1);
```
joperezr: I suppose you don't need this +1 any longer now that you've fixed the address?

joperezr:
Thanks for adding this @Silvenga ! I will go ahead and merge as soon as you remove that +1 from the sample as I have no further comments.

src/devices/Bmx280/samples/Bme280.sample.cs
Silvenga: Oops, fixed!

Silvenga:
@joperezr I rebased onto master. Anything preventing a merge at this point?

joperezr:
I'm so sorry about not merging this earlier, I lost track of the PR since I had already signed off on it. We are actually in the middle of a somewhat big refactoring for this binding here: #477. Is there any chance that we can incorporate your changes into that PR since most of the files that you change here are getting moved? cc: @georgemathieson It would be great if we could add this code to your PR so that Bme280 would also have Humidity Support.

Silvenga:
Not a problem, I can rebase onto georgemathieson:bme680 if that's where we are headed. Just let me know what to do.

joperezr:
That would be awesome! I'm very sorry again for not merging this earlier so that it got picked up by his refactoring 😨. @georgemathieson I suppose you are ok with @Silvenga pushing into your PR branch? Or do you prefer to merge yours first and have him create a new one on top of that?

georgemathieson:
Hey @joperezr I think if this branch went into master then I would be happy to bring master into my branch, and bring in the functionality that way. What are your thoughts? I'm just conscious that #482 and @RobinTTY are also blocked by the refactor.

joperezr:
@georgemathieson That sounds good to me. In that case, I'll go ahead and merge this PR. Thanks a lot for understanding and for your refactoring work!

PR #437, opened 2019-05-18
Added DS1307 Binding

ZhangGaoxing:
Realtime Clock DS1307


src/devices/Ds1307/Ds1307.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// DS1307 Default I2C Address
+        /// </summary>
+        public const byte DefaultI2cAddress = 0x68;
+
+        private I2cDevice _sensor = null;
+
+        /// <summary>
+        /// DS1307 DateTime
+        /// </summary>
+        public DateTime DateTime { get => ReadTime(); set => SetTime(value); }
```
joperezr: Any particular reason why you went with a property instead of just making your Read and Set time methods public? I don't have a strong preference on any, I just think that given you are actually executing some code that might throw when reading or writing bytes, it might be more intuitive to have this be a method.

src/devices/Ds1307/Ds1307.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// Cleanup
+        /// </summary>
+        public void Dispose()
+        {
+            _sensor?.Dispose();
+            _sensor = null;
+        }
+
+        /// <summary>
+        /// BCD To Int
```
joperezr: What does BCD stand for? Is that how the datasheet refers to the representation of the numbers in the register?

src/devices/Ds1307/Ds1307.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        /// <summary>
+        /// Read Time from DS1307
+        /// </summary>
+        /// <returns>DS1307 Time</returns>
+        private DateTime ReadTime()
+        {
+            Span<byte> readBuffer = stackalloc byte[7];
+
+            _sensor.WriteByte((byte)Register.RTC_SEC_REG_ADDR);
```
joperezr: NIT: should we add a small comment here saying that you are reading all registers at the time instead of one by one?

joperezr:
Left a few comments but this looks great otherwise, thanks for the contribution @ZhangGaoxing !

src/devices/Ds1307/Ds1307.cs
ZhangGaoxing: > I don't have a strong preference on any

Me too. My first `iot` submission was using the public method, but some members prevented me from doing so, haha. Now, using properties is a feature of `iot` 😄

src/devices/Ds1307/Ds1307.cs
ZhangGaoxing: BCD is Binary-coded decimal. Time is stored in BCD code in the register.

src/devices/Ds1307/Ds1307.cs
ZhangGaoxing: ok

src/devices/Ds1307/Ds1307.cs
joperezr: I see, that's fine then, let's keep it as a property for now.

PR #438, opened 2019-05-19
Basic SocketCan implementation

krwq:
Note: This is implemented in Iot.Device.Bindings since this requires much more testing on different kinds of devices which I currently don't have.

Multiple scenarios of CanBus require using CanRaw protocol which covers:

- listening for multiple devices
- writing to a single id
- hardware/driver filtering

cc: @richlander @shaggygi @joperezr

There exist higher level protocols based on this which are not covered in this PR (some of them might be already implemented directly in the kernel for free but that is also not covered here - I haven't got any devices I could test it on).

### DIFF
--- a/src/devices/SocketCan/CanFlags.cs
+++ b/src/devices/SocketCan/CanFlags.cs
@@ -7,7 +7,7 @@ using System;
 namespace Iot.Device.SocketCan
 {
     [Flags]
-    public enum CanFlags : uint
+    internal enum CanFlags : uint
     {
         ExtendedFrameFormat = 0x80000000,
         RemoteTransmissionRequest = 0x40000000,
--- a/src/devices/SocketCan/CanFlags.cs
+++ b/src/devices/SocketCan/CanFlags.cs
@@ -1,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.SocketCan
+{
+    [Flags]
+    public enum CanFlags : uint
+    {
+        ExtendedFrameFormat = 0x80000000,
+        RemoteTransmissionRequest = 0x40000000,
+        Error = 0x20000000,
+    }
+}
--- a/src/devices/SocketCan/CanFrame.cs
+++ b/src/devices/SocketCan/CanFrame.cs
@@ -1,0 +1,145 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Net.Sockets;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace Iot.Device.SocketCan
+{
+    [StructLayout(LayoutKind.Sequential)]
+    public unsafe struct CanFrame
+    {
+        private const int CAN_MAX_DLEN = 8;
+
+        // RawId (can_id) includes EFF, RTR and ERR flags
+        public uint RawId { get; set; }
+
+        // data length code (can_dlc)
+        // see: ISO 11898-1 Chapter 8.4.2.4
+        private byte _length;
+        private byte _pad;
+        private byte _res0;
+        private byte _res1;
+        private fixed byte _data[CAN_MAX_DLEN];
+
+        public uint Id => ExtendedFrameFormat ? ExtendedId : StandardId;
+
+        public uint StandardId
+        {
+            get
+            {
+                if (ExtendedFrameFormat)
+                    throw new InvalidOperationException($"{nameof(StandardId)} can be obtained only when {nameof(ExtendedFrameFormat)} is not set.");
+                
+                return RawId & Interop.CAN_SFF_MASK;
+            }
+            set
+            {
+                if ((value & ~Interop.CAN_SFF_MASK) != 0)
+                    throw new InvalidOperationException($"{nameof(value)} must be 11 bit identifier.");
+
+                ExtendedFrameFormat = false;
+                RawId &= ~Interop.CAN_SFF_MASK;
+                RawId |= value;
+            }
+        }
+
+        public uint ExtendedId
+        {
+            get
+            {
+                if (!ExtendedFrameFormat)
+                    throw new InvalidOperationException($"{nameof(ExtendedId)} can be obtained only when {nameof(ExtendedFrameFormat)} is set.");
+                
+                return RawId & Interop.CAN_EFF_MASK;
+            }
+            set
+            {
+                if ((value & ~Interop.CAN_EFF_MASK) != 0)
+                    throw new InvalidOperationException($"{nameof(value)} must be 29 bit identifier.");
+
+                ExtendedFrameFormat = true;
+                RawId &= ~Interop.CAN_EFF_MASK;
+                RawId |= value;
+            }
+        }
+
+        public bool Error
+        {
+            get => ((CanFlags)RawId).HasFlag(CanFlags.Error);
+            set => SetCanFlag(CanFlags.Error, value);
+        }
+
+        public bool ExtendedFrameFormat
+        {
+            get => ((CanFlags)RawId).HasFlag(CanFlags.ExtendedFrameFormat);
+            set => SetCanFlag(CanFlags.ExtendedFrameFormat, value);
+        }
+
+        public bool RemoteTransmissionRequest
+        {
+            get => ((CanFlags)RawId).HasFlag(CanFlags.RemoteTransmissionRequest);
+            set => SetCanFlag(CanFlags.RemoteTransmissionRequest, value);
+        }
+
+        public bool IsValid
+        {
+            get
+            {
+                uint idMask = ExtendedFrameFormat ? Interop.CAN_EFF_MASK : Interop.CAN_SFF_MASK;
+                return !Error && _length <= CAN_MAX_DLEN && (RawId & ~idMask) == 0;
+            }
+        }
+
+        public ReadOnlySpan<byte> Data
+        {
+            get
+            {
+                if (_length > CAN_MAX_DLEN)
+                {
+                    throw new InvalidOperationException($"Length {_length} exceed maximum allowed: {CAN_MAX_DLEN}");
+                }
+
+                fixed (byte* d = _data)
+                {
+                    return new ReadOnlySpan<byte>(d, _length);
+                }
+            }
+            set
+            {
+                if (value.Length > CAN_MAX_DLEN)
+                    throw new InvalidOperationException($"Data length cannot exceed {CAN_MAX_DLEN} bytes");
+                
+                _length = (byte)value.Length;
+
+                fixed (byte* d = _data)
+                {
+                    var buff = new Span<byte>(d, value.Length);
+                    value.CopyTo(buff);
+
+                    // fill remainder of the buffer with zeros for sanity
+                    var remainder = new Span<byte>(d + value.Length, CAN_MAX_DLEN - value.Length);
+                    remainder.Fill(0);
+                }
+            }
+        }
+
+        private void SetCanFlag(CanFlags flag, bool value)
+        {
+            if (value)
+            {
+                RawId |= (uint)flag;
+            }
+            else
+            {
+                RawId &= ~(uint)flag;
+            }
+        }
+    }
+}
--- a/src/devices/SocketCan/CanFrame.cs
+++ b/src/devices/SocketCan/CanFrame.cs
@@ -13,132 +13,26 @@ using System.Threading;
 namespace Iot.Device.SocketCan
 {
     [StructLayout(LayoutKind.Sequential)]
-    public unsafe struct CanFrame
+    internal unsafe struct CanFrame
     {
-        private const int CAN_MAX_DLEN = 8;
+        public const int MaxLength = 8;
 
         // RawId (can_id) includes EFF, RTR and ERR flags
-        public uint RawId { get; set; }
+        public CanId Id;
 
         // data length code (can_dlc)
         // see: ISO 11898-1 Chapter 8.4.2.4
-        private byte _length;
+        public byte Length;
         private byte _pad;
         private byte _res0;
         private byte _res1;
-        private fixed byte _data[CAN_MAX_DLEN];
-
-        public uint Id => ExtendedFrameFormat ? ExtendedId : StandardId;
-
-        public uint StandardId
-        {
-            get
-            {
-                if (ExtendedFrameFormat)
-                    throw new InvalidOperationException($"{nameof(StandardId)} can be obtained only when {nameof(ExtendedFrameFormat)} is not set.");
-                
-                return RawId & Interop.CAN_SFF_MASK;
-            }
-            set
-            {
-                if ((value & ~Interop.CAN_SFF_MASK) != 0)
-                    throw new InvalidOperationException($"{nameof(value)} must be 11 bit identifier.");
-
-                ExtendedFrameFormat = false;
-                RawId &= ~Interop.CAN_SFF_MASK;
-                RawId |= value;
-            }
-        }
-
-        public uint ExtendedId
-        {
-            get
-            {
-                if (!ExtendedFrameFormat)
-                    throw new InvalidOperationException($"{nameof(ExtendedId)} can be obtained only when {nameof(ExtendedFrameFormat)} is set.");
-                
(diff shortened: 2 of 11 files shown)


src/devices/SocketCan/samples/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Writing a frame
+
+
+```csharp
+using (CanRaw can = new CanRaw())
+{
+    byte[] buffer = new byte[8] { 1, 2, 3, 40, 50, 60, 70, 80 };
+    CanFrame frame = new CanFrame();
+    frame.StandardId = Id;
```
krwq: `0x1A` instead of Id (copy paste fail)

src/devices/SocketCan/CanFrame.cs
```diff
@@ -0,0 +1,145 @@
...
+
+namespace Iot.Device.SocketCan
+{
+    [StructLayout(LayoutKind.Sequential)]
+    public unsafe struct CanFrame
+    {
+        private const int CAN_MAX_DLEN = 8;
+
+        // RawId (can_id) includes EFF, RTR and ERR flags
+        public uint RawId { get; set; }
```
tarekgh: >set [](start = 33, length = 3)

Can't you do more validation on the RawId here? or any passed value will be valid?

src/devices/SocketCan/CanFrame.cs
```diff
@@ -0,0 +1,145 @@
...
+using System.Linq;
+using System.Net.Sockets;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace Iot.Device.SocketCan
+{
+    [StructLayout(LayoutKind.Sequential)]
+    public unsafe struct CanFrame
```
tarekgh: >CanFrame [](start = 25, length = 8)

would make sense to have a constructor for this struct which take the RawId? I think this struct is useless without RawId. And may be the RawId property be getter only?

src/devices/SocketCan/CanFlags.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.SocketCan
+{
+    [Flags]
+    public enum CanFlags : uint
```
joperezr: Can we add xmldoc comments to both the enum and the actual values?

src/devices/SocketCan/CanFrame.cs
```diff
@@ -0,0 +1,145 @@
...
+        public bool IsValid
+        {
+            get
+            {
+                uint idMask = ExtendedFrameFormat ? Interop.CAN_EFF_MASK : Interop.CAN_SFF_MASK;
+                return !Error && _length <= CAN_MAX_DLEN && (RawId & ~idMask) == 0;
+            }
+        }
+
+        public ReadOnlySpan<byte> Data
```
krwq: As we spoke with @tarekgh offline I'll change this struct to ref struct to emphasize that struct is stack only

src/devices/SocketCan/SocketCan.csproj
```diff
@@ -0,0 +1,17 @@
...
+    <!--Disabling default items so samples source won't get build by the main library-->
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <LangVersion>Latest</LangVersion>
+    <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <None Include="README.md" />
+    <PackageReference Include="System.Device.Gpio" Version="0.1.0-prerelease*" />
```
joperezr: Please reference the property version instead here, so that you don't have a wildcard reference.

src/devices/SocketCan/SafeCanRawSocketHandle.cs
```diff
@@ -0,0 +1,28 @@
...
+using System;
+using System.IO;
+using System.Net.Sockets;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Runtime.Serialization;
+
+namespace Iot.Device.SocketCan
+{
+    internal class SafeCanRawSocketHandle : SafeHandle
```
joperezr: I don't have much experience with SafeHandles, but shouldn't this have a finalizer that calls ReleaseHandle? Also, can we add xml docs for the class?

src/devices/SocketCan/Interop.cs
```diff
@@ -0,0 +1,235 @@
...
+using System.IO;
+using System.Net.Sockets;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Runtime.Serialization;
+using System.Text;
+
+namespace Iot.Device.SocketCan
+{
+    internal class Interop
```
joperezr: static?

src/devices/SocketCan/Interop.cs
```diff
@@ -0,0 +1,235 @@
...
+        internal unsafe struct ifreq
+        {
+            internal const int IFNAMSIZ = 16;
+            public fixed byte ifr_name[IFNAMSIZ];
+            public int ifr_ifindex;
+            private fixed byte _padding[IFNAMSIZ - sizeof(int)];
+
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
```
joperezr: For these structs with sequential layout, should we be concerned for running on x85 vs x64 processes?

src/devices/SocketCan/SocketCan.csproj
shaggygi: @joperezr side topic, but related... I'm assuming you're talking about the following?

```xml
<PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
```

I get build errors when using this local on other projects.  What should be updated to get the projects to build correctly?

src/devices/SocketCan/SafeCanRawSocketHandle.cs
tarekgh: He is subclassing Safehande which has the finalizer which will call ReleaseHandle

---
In reply to: [288320949](https://github.com/dotnet/iot/pull/438#discussion_r288320949) [](ancestors = 288320949)

src/devices/SocketCan/SafeCanRawSocketHandle.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public SafeCanRawSocketHandle(string networkInterface) : base(Interop.CreateCanRawSocket(networkInterface), true)
+        {
+        }
+
+        public override bool IsInvalid => (int)handle == -1;
+
+        protected override bool ReleaseHandle()
+        {
+            Interop.CloseSocket(handle);
```
tarekgh: >Interop.CloseSocket(handle) [](start = 12, length = 27)

do you want to check IsValid before closing the socket?

src/devices/SocketCan/SocketCan.csproj
joperezr: > I'm assuming you're talking about the following?

Yup, that's what I'm talking about.

> I get build errors when using this local on other projects. What should be updated to get the projects to build correctly?

When you say other projects do you mean projects outside of the repo? If so, that is expected, since this is a variable that is defined in our repo which is imported by all projects in the repo, so it won't work for projects that are outside of it. Here is where we define it:

https://github.com/dotnet/iot/blob/26e1b07557ce650d3110769262a27b50f38650be/eng/Versions.props#L5

If you see, we update this guy daily with an auto-generated PR everytime we have a new build of our packages.

src/devices/SocketCan/Interop.cs
tarekgh: can't be x85 at all :-)

but it doesn't matter the architecture as the size of the fields inside the struct not changing between x86 and x64

src/devices/SocketCan/Interop.cs
```diff
@@ -0,0 +1,235 @@
...
+                throw new IOException($"Cannot bind to socket to `{interfaceName}`");
+            }
+        }
+
+        private static unsafe int GetInterfaceIndex(int fd, string name)
+        {
+            const uint SIOCGIFINDEX = 0x8933;
+            const int MaxLen = ifreq.IFNAMSIZ - 1;
+
+            if (Encoding.ASCII.GetByteCount(name) >= MaxLen)
```
tarekgh: >Encoding.ASCII.GetByteCount(name) [](start = 16, length = 33)

I believe you need to use the ASCII encoding which will throw on the invalid characters. right?

tarekgh: >Encoding.ASCII.GetByteCount(name) [](start = 16, length = 33)

Also, you don't want to callGetByteCount. can't you just check name.Length?

src/devices/SocketCan/SocketCan.csproj
shaggygi: > When you say other projects do you mean projects outside of the repo?

That's what it was.  Thanks as usual!

src/devices/SocketCan/CanFrame.cs
krwq: will add some validation (although i.e. setting ERR flag makes frame invalid I will not do validation there as it may be expected - I will only validate that extended frame has ID no larger than 29 bit and standard frame has up to 11 bit id)

src/devices/SocketCan/CanFrame.cs
krwq: one of the expected scenarios is that ID changes frequently and the same message might be sent to multiple people - I think it makes sense to leave this mutable

src/devices/SocketCan/Interop.cs
krwq: This struct is being sent directly over the wire so 32/64 bit doesn't matter

src/devices/SocketCan/SafeCanRawSocketHandle.cs
krwq: SafeHandle already checks that:
https://source.dot.net/#System.Private.CoreLib/shared/System/Runtime/InteropServices/SafeHandle.cs,220

src/devices/SocketCan/Interop.cs
krwq: actually I think this should be UTF8 (and yes it should throw) - will double check and fix

src/devices/SocketCan/CanFrame.cs
```diff
@@ -0,0 +1,145 @@
...
+using System.IO;
+using System.Linq;
+using System.Net.Sockets;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace Iot.Device.SocketCan
+{
+    [StructLayout(LayoutKind.Sequential)]
```
JeremyKuhne: Structs are always `LayoutKind.Sequential`.

src/devices/SocketCan/CanFrame.cs
JeremyKuhne: You can't do this safely. You have to create an array here to be safe. I've been discussing a way to handle this safely in the future with @jaredpar.

If you've got a processing loop you can deal with this limitation the same way I did `FileSystemEntry` for file enumeration. It is a ref struct that you never directly create- it gets passed to your handlers in `FileSystemEnumerator`/`Enumerable` which makes it difficult to elevate the native view out of scope.

src/devices/SocketCan/CanFrame.cs
krwq: talked offline - I will create CanId struct and make CanFrame internal and have WriteFrame and TryReadFrame methods instead

src/devices/SocketCan/CanFrame.cs
krwq: on the second thought, making this internal

src/devices/SocketCan/CanFrame.cs
krwq: struct's still mutable but I've made RawId internal now. Standard and Extended address can still be changed

src/devices/SocketCan/CanFlags.cs
krwq: making this internal

src/devices/SocketCan/Interop.cs
krwq: leaving this ASCII until since we only expect name in format `can[0-9]`

src/devices/SocketCan/CanFrame.cs
krwq: actually seems there is a tiny difference with and without attribute: no attribute generates warning for unused field such as `_padding`. I'll only use it when needed

src/devices/SocketCan/CanRaw.cs
```diff
@@ -0,0 +1,116 @@
...
+            ReadOnlySpan<CanFrame> frameSpan = MemoryMarshal.CreateReadOnlySpan(ref frame, 1);
+            ReadOnlySpan<byte> buff = MemoryMarshal.AsBytes(frameSpan);
+            Interop.Write(_handle, buff);
+        }
+
+        public bool TryReadFrame(Span<byte> buffer, out int frameLength, out CanId id)
+        {
+            CanFrame frame = new CanFrame();
+            
+            {
```
tarekgh: >  { [](start = 10, length = 3)

nit: why you have this?

src/devices/SocketCan/CanRaw.cs
```diff
@@ -0,0 +1,116 @@
...
+                Span<CanFrame> frameSpan = MemoryMarshal.CreateSpan(ref frame, 1);
+                Span<byte> buff = MemoryMarshal.AsBytes(frameSpan);
+                while (buff.Length > 0)
+                {
+                    int read = Interop.Read(_handle, buff);
```
tarekgh: >Interop.Read [](start = 31, length = 12)

does Read can ever return negative numbers?

src/devices/SocketCan/CanRaw.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                // insufficient buffer
+                // bytesWritten will tell how much is needed
+                return false;
+            }
+
+            unsafe
+            {
+                Span<byte> frameData = new Span<byte>(frame.Data, frame.Length);
+                frameData.CopyTo(buffer);
```
tarekgh: >frameData.CopyTo( [](start = 16, length = 17)

would be nice if you can add a detailed comment why we cannot just use the input buffer directly and doing the copy. I know the reason but the code readers wouldn't figure it out easily :-)

src/devices/SocketCan/CanRaw.cs
```diff
@@ -0,0 +1,116 @@
...
+                // we will leave id filled in case it is useful for anyone
+                frameLength = 0;
+                return false;
+            }
+            
+            if (frame.Length > buffer.Length)
+            {
+                // insufficient buffer
+                // bytesWritten will tell how much is needed
+                return false;
```
tarekgh: > return false; [](start = 15, length = 14)

what is expected from the caller when getting this case? should pass bigger span? and if so, we'll use what we have read before or we'll go and re-read again?

src/devices/SocketCan/Interop.cs
```diff
@@ -0,0 +1,232 @@
...
+        [DllImport("libc", EntryPoint = "ioctl", CallingConvention = CallingConvention.Cdecl)]
+        private static extern int Ioctl3(int fd, uint request, ref ifreq ifr);
+
+        [DllImport("libc", EntryPoint = "bind", CallingConvention = CallingConvention.Cdecl)]
+        private static extern int BindSocket(int fd, ref CanSocketAddress addr, uint addrlen);
+
+        [DllImport("libc", EntryPoint = "close", CallingConvention = CallingConvention.Cdecl)]
+        private static extern int CloseSocket(int fd);
+
+        [DllImport("libc", EntryPoint = "write", CallingConvention = CallingConvention.Cdecl, SetLastError = true)]
```
tarekgh: >SetLastError = true [](start = 94, length = 19)

why we set last error while I am not seeing you are using it?

src/devices/SocketCan/CanRaw.cs
krwq: This is a tiny wrapper over the actual Read which throws when the returned value is negative

src/devices/SocketCan/CanRaw.cs
krwq: good point, I'll throw if user does not pass at minimum 8 byte long buffer, otherwise everything gets more complicated

src/devices/SocketCan/Interop.cs
krwq: removing, error messaging is something we might want to improve in the future but I'd prefer to wait until this code can use Socket underneath instead of writing that logic right now (this will be possible once iot re-targets to 3.0)

PR #439, opened 2019-05-20
Update DS3231

ZhangGaoxing:


src/devices/Ds3231/Ds3231.cs
```diff
@@ -45,26 +45,18 @@ public Ds3231(I2cDevice sensor)
...
-            data.Date = Bcd2Int(rawData[4]);
-            data.Month = Bcd2Int((byte)(rawData[5] & 0x1F));
-            data.Century = rawData[5] >> 7;
-            if (data.Century == 1)
-                data.Year = 2000 + Bcd2Int(rawData[6]);
-            else
-                data.Year = 1900 + Bcd2Int(rawData[6]);
-
-            return new DateTime(data.Year, data.Month, data.Date, data.Hour, data.Min, data.Sec);
+            return new DateTime(rawData[5] >> 7 == 1 ? 2000 + Bcd2Int(rawData[6]) : 1900 + Bcd2Int(rawData[6]),
```
joperezr: > `rawData[5] >> 7 == 1`

Aren't you missing a `& 1` just to check if the bit is turned on? Is there ever a case where rawData >> 7 can be more than 0 or 1 ?

joperezr: Looks like before you were not passing in Century to DateTime constructor but you are now, and also you don't seem to be passing data.Date any longer.

joperezr:
Left a couple of small comments but looks good otherwise

src/devices/Ds3231/Ds3231.cs
ZhangGaoxing: The right shift of byte type by 7 bits should not be greater than 1. Do you think this step should be added? If necessary, I'll add it.😄

ZhangGaoxing: `Bcd2Int((byte)(rawData[5] & 0b_0001_1111))` in this step, I removed Century. 

> you don't seem to be passing data.Date any longer

My mistake

joperezr:
Changes look good, that said, why is it that we are changing the DateTime logic on the PR, was there an actual bug with the previous code? It would be great if these PRs had some description that explained better why were the changes being made on the binding, so that it makes more sense when checking history of changes.

joperezr:
CI wasn't running on PRs so I'm closing and reopening to make sure CI runs again.

src/devices/Ds3231/Ds3231.cs
krwq: I think it's fine as is

src/devices/Ds3231/Ds3231.cs
```diff
@@ -45,26 +45,18 @@ public Ds3231(I2cDevice sensor)
...
+                                Bcd2Int((byte)(rawData[5] & 0b_0001_1111)),
+                                Bcd2Int(rawData[4]),
+                                Bcd2Int(rawData[2]),
```
krwq: what's in `rawData[3]`? it seems to be missing here

ZhangGaoxing:
> Changes look good, that said, why is it that we are changing the DateTime logic on the PR, was there an actual bug with the previous code? It would be great if these PRs had some description that explained better why were the changes being made on the binding, so that it makes more sense when checking history of changes.

I will submit some new RTC device bindings, so I checked the previous code to match the new code. I removed the `Ds3231Data` class because it was logically redundant. Assigning raw data to Ds3231Data and then instantiating DateTime with Ds3231Data is not as good as instantiating DateTime directly with raw data.

src/devices/Ds3231/Ds3231.cs
ZhangGaoxing: `rawData[3]` is weekday, and I don't think DateTime needs this data.😄

PR #442, opened 2019-05-21
Add Ssd1351 Binding

Frankenslag:
Here is the PR for the proposed Ssd1351 binding.

I look forward to hearing your comments.

Note that it shows a change to .gitignore. This is there as files in the native folder were showing up visual studio as changes to be pushed.

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -29,7 +29,7 @@ You need to use at least [.NET Core 2.1](https://www.microsoft.com/net/download/
 
 Many of these samples use the [Raspberry Pi](https://www.raspberrypi.org/), however, .NET Core can be used with other devices. A [Raspberry Pi Starter Pack](https://www.adafruit.com/product/3058) contains enough electronics to get started on many projects.
 
-.NET Core is supported on Raspberry Pi 2B and 3A+/B/B+ models (ARMv7/v8). .NET Core is not supported on Raspberry Pi Zero or Arduino.
+.NET Core is supported on Raspberry Pi 2B as well as 3 and 4 models (ARMv7/v8). .NET Core is not supported on Raspberry Pi Zero or Arduino.
 
 ## Resources
 


.gitignore
```diff
@@ -336,3 +336,4 @@ ASALocalRun/
 
 # Ignore Mac DS_Store files
 .DS_Store
+/src/Native/out/build/x64-Debug
```
joperezr: mm not sure why you had these files here, but I don't think we want to change the .gitignore as nothing should get dropped in this specific folder. Try removing this, and running a `git clean -fdx` from the root of the repo in order to delete all the files that are not checked in. Then run a `build.cmd` and you should see that nothing gets dropped there.

src/devices/Ssd1351/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Binding Notes
+
+This binding currently only supports commands and raw data.  Eventually, the plan is to create a graphics library that can send text and images to the device.
+
+### Connection Type
+
+The following connection types are supported by this binding.
+
+- [ ] I2C
```
joperezr: Does this mean that I2C is supported but not includded in the binding yet, or that I2C is not supported by the binding?

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,329 @@
...
+                // send the slice of data off set by the index and of length len.
+                _spiDevice.Write(data.Slice(index, len));
+                // add the length just sent to the index
+                index += len;
+            } while (index < data.Length); // repeat until all data sent.
+        }
+
+        public void Dispose()
+        {
+            if(_gpioDevice != null)
```
joperezr: Instead of doing this, it would be better to just dispose the controller like:
```cs
_gpioDevice?.Dispose();
_gpioDevice = null;
```

That will close open pins and it would also make sure that all that needs to be cleaned up is done correctly.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,329 @@
...
+        /// <summary>
+        /// Initializes new instance of Ssd1351 device that will communicate using SPI bus.
+        /// A single-chip CMOS OLED/PLED driver with controller for organic/polymer
+        /// light emitting diode dot-matrix graphic display system. 
+        /// </summary>
+        /// <param name="spiDevice">>The SPI device used for communication.</param>
+        /// <param name="gpioDevice">>The GPIO controller device used for communication.</param>
+        /// <param name="dcPinId">>The id of the GPIO pin used to control the DC line (data/command).</param>
+        /// <param name="resetPinId">>The id of the GPIO pin used to control the /RESET line (data/command).</param>
+        public Ssd1351(SpiDevice spiDevice, GpioController gpioDevice, int dcPinId, int resetPinId)
```
joperezr: >gpioDevice [](start = 59, length = 10)

Why do you need a gpio controller to get passed in? usually what we do in other bindings, is that we initialize this gpio controller within the class and then dispose it once we dispose the object.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,329 @@
...
+            SPIBufferSize = DefaultSPIBufferSize;
+
+            SendCommand(new SetSegmentReMapColorDepth());
+
+        }
+
+        /// <summary>
+        /// The size of the SPI buffer. If data larger than the buffer is sent then it is split up into multiple transmissions
+        /// </summary>
+        public int SPIBufferSize
```
joperezr: Do we intentionally want to allow modifications on the buffer size? or is the most common case to use the same buffer size throught the lifespan of the object. I ask because if that's the case, I would just pass this in to the constructor instead and don't allow changing it.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,329 @@
...
+
+                _spiBufferSize = value;
+            }
+        }
+
+        /// <summary>
+        /// Convert a color structure to a byte tuple represening the colour in 565 format.
+        /// </summary>
+        /// <param name="color">The color to be converted.</param>
+        private (byte, byte) Color565(Color color)
```
joperezr: @krwq didn't we have this already somewhere in the repo? I kind of feel that we already did this and had a common library for this but I can't seem to find it. I think we did it for sense hat, but I might be wrong.

src/devices/Ssd1351/README.md
Frankenslag: I2C is not supported. I will remover the tickbox

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: O.K. I will do that.... I don't see any reason for it being a property either and so I will remove the SPIBufferSize property.

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: I think there was a discussion regarding this that I saw. I'll have a look around. It is a bit more complex than first it seems though as with the Ssd1351 you can change the byte order of the colour representation and this is also handled within the method.

.gitignore
Frankenslag: It seems that CMake is being run when visual studio opens the folder. This is then creating the files. in /src/Native/out/build-X64-Debug. I'll investigate further.

Frankenslag: I have a bit more clarity now. This problem appears to have been introduced with Visual Studio 16.1 (It seems ok with the initial version of VS2019 and the latest VS2017). The CMake command line is as below.. the  -DCMAKE_INSTALL_PREFIX:PATH seems to be pointing into the Native\out directory whereas in earlier versions of VS then it is a temporary directory. I will try and see if this is intentional or otherwise....

 "C:\PROGRAM FILES (X86)\MICROSOFT VISUAL STUDIO\2019\ENTERPRISE\COMMON7\IDE\COMMONEXTENSIONS\MICROSOFT\CMAKE\CMake\bin\cmake.exe"  -G "Ninja" -DCMAKE_INSTALL_PREFIX:PATH="d:\source\dotnet\iot\src\native\out\install\x64-Debug" -DCMAKE_CXX_COMPILER:FILEPATH="C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.21.27702/bin/HostX64/x64/cl.exe" -DCMAKE_C_COMPILER:FILEPATH="C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.21.27702/bin/HostX64/x64/cl.exe"  -DCMAKE_BUILD_TYPE="Debug" -DCMAKE_MAKE_PROGRAM="C:\PROGRAM FILES (X86)\MICROSOFT VISUAL STUDIO\2019\ENTERPRISE\COMMON7\IDE\COMMONEXTENSIONS\MICROSOFT\CMAKE\Ninja\ninja.exe" "D:\source\dotnet\iot\src\Native"

src/devices/Ssd1351/tests/Ssd1351Tests.cs
```diff
@@ -0,0 +1,17 @@
...
+using Xunit;
+
+namespace Iot.Device.Ssd1351.Tests
+{
+    public class Ssd1351Tests
+    {
+        [Fact]
+        public void Test1()
+        {
+            // NOTE: THIS TEST PROJECT IS NOT REQUIRED AND CAN BE DELETED IF NOT USED.
```
krwq: please delete the test project if not used

src/devices/Ssd1351/ssd1351Imaging.cs
```diff
@@ -0,0 +1,148 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+
```
krwq: nit: extra enter

src/devices/Ssd1351/ssd1351Imaging.cs
```diff
@@ -0,0 +1,148 @@
...
+namespace Iot.Device.Ssd1351
+{
+    /// <summary>
+    /// A single-chip CMOS OLED/PLED driver with controller for organic/polymer
+    /// light emitting diode dot-matrix graphic display system. 
+    /// </summary>
+    public partial class Ssd1351 : IDisposable
+    {
+
```
krwq: extra enter

src/devices/Ssd1351/ssd1351Imaging.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// <summary>
+        /// Send a bitmap to the ssd1351 display specifying the starting position and destination clipping rectangle.
+        /// </summary>
+        /// <param name="bm">The bitmap to be sent to the display controller note that only Pixel Format Format32bppArgb is supported.</param>
+        /// <param name="sourcePoint">A coordinate point in the source bitmap where copying starts from.</param>
+        /// <param name="destinationRect">A rectangle that defines where in the display the bitmap is written. Note that no scaling is done.</param>
+        public void SendBitmap(Bitmap bm, Point sourcePoint, Rectangle destinationRect)
+        {
+
```
krwq: nit: extra enter (not commenting anymore)

src/devices/Ssd1351/samples/Ssd1351.Sample.cs
```diff
@@ -0,0 +1,105 @@
...
+                        {
+                            ssd1351.ResetDisplayAsync().Wait();
+                            InitDisplayForAdafruit(ssd1351);
+
+                            using (System.Drawing.Graphics g = System.Drawing.Graphics.FromImage(dotnetBM))
+                            {
+
+                                while(true)
+                                {
+                                    foreach (string filepath in Directory.GetFiles(@"images", "*.png").OrderBy(f => f))
```
krwq: that's a lot of indentation in here. I'd recommend simplifying this a bit:
- all usings one after another i.e.:

```csharp
using (...)
using (...)
{
   ...
}
```

- `(displayGPIO = new GpioController()) != null` will never be false so if statement is not necessary
- `(displaySPI = new UnixSpiDevice(new SpiConnectionSettings(0, 0) ....) != null` - similarly will never be null

that should reduce most of the nesting

.gitignore
```diff
@@ -336,3 +336,4 @@ ASALocalRun/
 
 # Ignore Mac DS_Store files
```
krwq: is this code much different from https://github.com/dotnet/iot/tree/master/src/devices/Ssd1306 ? if they are similar might be worth to unify them into Ssd13XX or something similar

src/devices/Ssd1351/README.md
Frankenslag: I have removed the I2C in the lastest PR

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: I have made this change in the latest PR

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: I have made this modification. I initially felt that as I was passing in the SPI device then passing in the GPIO controller would be correct but I take your lead.

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: I have added a parameter to the constructor to cover the buffer size.

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: I had a look and couldn't find reference to this right now.

src/devices/Ssd1351/tests/Ssd1351Tests.cs
Frankenslag: Done in the latest PR

src/devices/Ssd1351/ssd1351Imaging.cs
Frankenslag: Done in the latest PR

src/devices/Ssd1351/ssd1351Imaging.cs
Frankenslag: O.K. I wasn't aware that the line after the opening brace shouldn't be blank ?

src/devices/Ssd1351/ssd1351Imaging.cs
Frankenslag: Not sure what rule is being broken here but I have removed the line after the brace.

src/devices/Ssd1351/samples/Ssd1351.Sample.cs
Frankenslag: Done in the latest PR

.gitignore
Frankenslag: I looked at this but felt that the differences between are fairly major i.e.

- Ssd1306 is currently I2C only and Ssd1351 is SPI only with no option for I2C.
- The way that commands are sent to the device is quite different. The Ssd1351 distinguishes between the command and following parameter bytes with the used of the DC pin whereas the Ssd1306 uses the DC ping to distinguish between commands and data.
- The commands are quite different and the way that they address data.

We could of course use flow control statements to cover both devices but I think that the differences are such that most of the complexity would be in the flow control statements and not in the code that actually controls the binding.

Personally I think that what needs to be done in the future is to come to a definitive decision about what interfaces etc. we expect display devices to support but I think other contributors will have a better insight into that then myself.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,312 @@
...
+                // send the slice of data off set by the index and of length len.
+                _spiDevice.Write(data.Slice(index, len));
+                // add the length just sent to the index
+                index += len;
+            } while (index < data.Length); // repeat until all data sent.
+        }
+
+        public void Dispose()
+        {
+            if(_gpioDevice != null)
```
joperezr: You don't need to check for null, that is what the question mark is doing before calling dispose. Also, don't you also need to dispose the spidevice? I know that one gets passed in, but the pattern we have been following in the bindings is that we dispose the Spi or I2c device when passed in. So basically the code of the dispose method should look like:

```cs
        public void Dispose()
        {
            _gpioDevice?.Dispose();
            _gpioDevice = null;
            _spiDevice?.Dispose();
            _spiDevice = null;
        }
```

Frankenslag: I thought the last change to the PR had implemented this for the GPIO device. I left out the SPI device from this pattern as I read the opinion that the caller should dispose as the caller created the device. ... Anyway happy to follow your lead. Will change this at the next opportunity.

joperezr: yeah I would prefer to have all of the bindings following the same pattern for now (disposing the devices) and if we then decide to change this, we'll go through all the bindings and change them in one PR, but ideally we want to aim for consistency 😉

Frankenslag: The latest PR should now include the changes to the dispose pattern. Hope it looks good. Should I be pressing the [Resolve conversation] button or is that for you to do ?

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: I have found the method doing an RGB565 conversion. It was in the WriteColor method implemented in SenseHatLedMatrixSysFs.cs It doesn't look like this has been broken out itno a common library and is not really a generic function but one that is tied to writing to the "RPi-Sense FB" device and so is not very generic in its current form.

src/devices/Ssd1351/Ssd1351.cs
krwq: @Frankenslag thanks! Personal preference - when I have few comments I usually let the comment author resolve them (and resolve them after a day if no one does it) - for larger number of comments I resolve them (there are usually a few comments which need more discussion - without resolving the noise it's hard to see them)

src/devices/Ssd1351/Command/Commands.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
```
krwq: copyright header missing (same comment for all files where it applies - see other .cs files for example)

src/devices/Ssd1351/Command/Commands.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Iot.Device.Ssd1351.Command
+{
+    public enum SSD1351Command : byte
+    {
+        SSD1351_CMD_SETCOLUMN = 0x15, // used
```
krwq: unsure what `//used` refers to but presumably you might want to remove the comments

src/devices/Ssd1351/Command/Locking.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Text;
+
+namespace Iot.Device.Ssd1351.Command
+{
+    public class Lock : ICommand
+    {
+        /// <summary>
+        /// This command is used to lock the OLED driver IC from accepting any command except itself.
+        /// </summary>
+        public Lock()
```
krwq: Seems to me that lock and unlock could be one command with a bool flag (similarly to how the hardware works)

src/devices/Ssd1351/Command/Commands.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Iot.Device.Ssd1351.Command
+{
+    public enum SSD1351Command : byte
```
krwq: casing: `Ssd1351Command`

src/devices/Ssd1351/Command/Commands.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        SSD1351_CMD_SETCOLUMN = 0x15, // used
+        SSD1351_CMD_SETROW = 0x75, // used
+        SSD1351_CMD_WRITERAM = 0x5C, //used
+        SSD1351_CMD_READRAM = 0x5D,
```
krwq: I'd recommend renaming the values or alternatively making this enum private. If you choose renaming:
- drop prefix (`SSD1351_CMD_` seems redundand since it's already in the name of the enum)
- expand names where possible and makes sense. I.e.: `SetRow`, `DisplayOn`, `HorizontalScroll` etc

src/devices/Ssd1351/Command/Commands.cs
```diff
@@ -0,0 +1,56 @@
...
+        SSD1351_CMD_CONTRASTABC = 0xC1, //used
+        SSD1351_CMD_CONTRASTMASTER = 0xC7, //used
+        SSD1351_CMD_MUXRATIO = 0xCA, //used
+        SSD1351_CMD_COMMANDLOCK = 0xFD, // used
+        SSD1351_CMD_HORIZSCROLL = 0x96,
+        SSD1351_CMD_STOPSCROLL = 0x9E,
+        SSD1351_CMD_STARTSCROLL = 0x9F,
+    }
+
+    public interface ICommand
```
krwq: Does having this interface buys anything more rather than having methods directly on Ssd1351. I.e. something along the lines of

```csharp
private void SendCommand(Command cmd, ReadOnlySpan<byte> additionalData)
{
  if (additionalData.Length > MaxSpanLengthInExample64Bytes)
    throw new Exception("TODO: regular allocation or change to Debug.Assert if not needed");
  Span<byte> buff = stackalloc byte[additionalData.Length + 1];
  buff[0] = (byte)cmd;
  additionalData.CopyTo(buff.Slice(1));
  _spiOrI2c.Write(buff);
}

private void SendCommand(Command cmd, byte additionalData)
{
  Span<byte> data = stackalloc byte[1];
  data[0] = additionalData;
  SendCommand(cmd, data);
}
public void Lock(bool lockState)
{
  SendCommand(Command.Lock, lockState ? 0x12 : 0x16); // or whatever the values were
}
```

src/devices/Ssd1351/Ssd1351.cs
krwq: Yes, we might want to move this to common folder:
https://github.com/dotnet/iot/blob/master/src/devices/SenseHat/SenseHatLedMatrixSysFs.cs#L95

you can leave as is for now though - it might be worth creating an issue to unify this and have some conversation how to make it generic purpose (i.e. create Color565 struct or just represent the value as i.e. ushort or tuple as in here).

src/devices/Ssd1351/Command/Commands.cs
Frankenslag: Well I sort of agree with you regarding this but I was trying to follow the way that the existing Ssd1306 was engineered which uses an ICommand interface. I would be happy to refactor if that is what is needed. Please confirm which way you would like me to go,

krwq: In general it's rather matter of preference, internally I've rather seen people pushing back on the interfaces. We have allowed some of that in the past reviews since it's much better to have a support for a new device than not having it and it seemed like quite a bit of work to refactor that.

Before doing any work I'd recommend creating a separate issue to decide which direction people think we should go (please cc at minimum everyone which added/modified previous bindings, me and @joperezr).

For this PR you can leave as is or if you think the non-interface version feels better then go ahead and refactor only this code so that we can compare usage before & after and decide which one is better (note that implementation detail is less important than user experience).

Note that we are still trying to establish generic patterns in this repo and therefore we don't have specific guidelines for this case (so far most of the devices do not use ICommand interface or any similar concept and rather directly expose properties or methos but Ssd1306 do).

Most common and generally (IMO) easier to use pattern would be to have properties (i.e. instead of method Lock I shared above)

```csharp
bool Locked
{
  get
  {
    // TODO: either read cached value or if possible ask hardware what is the true state
  }
  set
  {
    SendCommand(Command.Lock, value ? 0x12 : 0x16); // or whatever the values were
  }
}
```

this way you can write code like following:

```csharp
ssd1351.Locked = true;
ssd1351.Locked = false;
```

vs

```
ssd1351.SendCommand(new LockCommand());
ssd1351.SendCommand(new UnlockCommand());
```

except for syntax being easier and more discoverable (since intellisense will help you with list of available commands when you type `ssd1351.`) you also don't need to allocate the command (although this is arguable since you can cache all popular commands)

shaggygi: > ssd1351.Locked = true;
> ssd1351.Locked = false;

While this does make it cleaner sometimes, I think this type should stay in method format... especially since it is related to processing data via protocol for the binding device to update hardware.  It will confuse binding creators on when/not to use. For example, one might code up...

```csharp
// used as property
ssd1351.Lock = true;
// and for another command as method
ssd1351.SetColumnAddress(0x00, 0x7F);
```

Below shows in method format...
```csharp
ssd1351.Lock();
// or could use a flag to perform Lock and Unlock commands.
ssd1351.Lock(true);
ssd1351.Lock(false);
```

I like how we used a property to control a pin where it was a master GPIO controller that interacted with the binding pins directly and not through the device.  Use [Mcp25xxx pin](https://github.com/dotnet/iot/blob/26e1b07557ce650d3110769262a27b50f38650be/src/devices/Mcp25xxx/Mcp25xxx.cs#L124) as example.

So if the Ssd1351 device had control pins, like Output Enable or Lock, I could see being used as a property.

```csharp
ssd1351.Lock = true;  // or ssd1351.Lock = PinValue.High;
ssd1351.OutputEnable = false;
```

This might be another detail we can add in guidance related to #463.

PR #443, opened 2019-05-21
Fix Build break

joperezr:
For some reason CI didn't catch the build break caused by a recent change that went in that was adding an example for the CharacterLcd display. This change will fix that.


src/devices/CharacterLcd/samples/Pcf8574tSample.cs
```diff
@@ -1,7 +1,11 @@
 ﻿using System;
+using System.Device.I2c;
```
krwq: add a license header since you're touching it (and presumably anywhere else in this dir which it might be missing)

joperezr: good point, will fix.

PR #452, opened 2019-05-24
Added WriteRead() method in I2cDevice

### LINKED ISSUE #129: I2C API should support Restart/Repeat condition
It appears you currently can only read and write using I2C API where it provides a Start/Stop at beginning and end of command.  Many I2C components will work by writing the address with a Write command and then follow with a ReadByte() or Read() for multiples.  However, there are components that require a Reset condition between writing and reading.

@bigdnf mentioned this in https://github.com/dotnet/iot/issues/111

Windows.Devices.I2c appears to have related support with a WriteRead():
https://docs.microsoft.com/en-us/uwp/api/windows.devices.i2c.i2cdevice.writeread#Windows_Devices_I2c_I2cDevice_WriteRead_System_Byte___System_Byte___

It appears the `UnixI2cDevice` has some started work [here](https://github.com/dotnet/iot/blob/0bb80590d409b0c974b4012c484828d7593268bc/src/System.Device.Gpio/System/Device/I2c/Drivers/UnixI2cDevice.Linux.cs#L59).  Have not found anything public to use similar to SPI [TransferFullDuplex](https://github.com/dotnet/iot/blob/0bb80590d409b0c974b4012c484828d7593268bc/src/System.Device.Gpio/System/Device/Spi/SpiDevice.cs#L14).

**Couple Linux references**
* https://stackoverflow.com/questions/35078131/why-repeated-start-based-i2c-operation-are-not-supported-in-linux
* https://stackoverflow.com/questions/505023/reading-writing-from-using-i2c-on-linux

ZhangGaoxing:
fix #129

I tested the code on Raspberry Pi 3B+ running `Raspbian`, and it worked well. The sensor that requires I2C Repeated Start is working properly, such as `MLX90614` and `KT0803L`  I mentioned in #129. **BCM2835** can use `WriteRead()` directly, but **BCM2708** needs to refer to this [post](https://www.raspberrypi.org/forums/viewtopic.php?f=44&t=15840&sid=8c2fff5ce4395676800b4587f2a71b4e&start=25).

If the code is merged, I can add some new bindings.


src/System.Device.Gpio/System/Device/I2c/Drivers/UnixI2cDevice.Linux.cs
```diff
@@ -214,6 +214,33 @@ public override unsafe void Write(ReadOnlySpan<byte> data)
...
+        /// <param name="writeBuffer">
+        /// The buffer that contains the data to be written to the I2C device.
+        /// The data should not include the I2C device address.</param>
+        /// <param name="readBuffer">
+        /// The buffer to read the data from the I2C device.
+        /// The length of the buffer determines how much data to read from the I2C device.
+        /// </param>
+        public override unsafe void WriteRead(ReadOnlySpan<byte> writeBuffer, Span<byte> readBuffer)
+        {
+            if (readBuffer.Length == 0)
```
joperezr: Please also check if readBuffer is not null. I aslo think that we should do both checks for the write buffer.

src/System.Device.Gpio/System/Device/I2c/Drivers/UnixI2cDevice.Linux.cs
```diff
@@ -214,6 +214,33 @@ public override unsafe void Write(ReadOnlySpan<byte> data)
...
+                throw new ArgumentException($"{nameof(readBuffer)} cannot be empty.");
+
+            Initialize();
+
+            fixed (byte* writeBufferPointer = writeBuffer)
+            {
+                fixed (byte* readBufferPointer = readBuffer)
+                {
+                    Transfer(writeBufferPointer, readBufferPointer, writeBuffer.Length, readBuffer.Length);
```
joperezr: In order for this to work, should the Write and read buffer be of the same size?

joperezr:
Left some small comments but this looks great otherwise. Thanks a lot for adding this @ZhangGaoxing!

src/System.Device.Gpio/System/Device/I2c/Drivers/UnixI2cDevice.Linux.cs
ZhangGaoxing: Existing code does not do null checks on `readbuffer` and `writeBuffer`, need I fix it together?

src/System.Device.Gpio/System/Device/I2c/Drivers/UnixI2cDevice.Linux.cs
ZhangGaoxing: Existing code does not do null checks on `readbuffer` and `writeBuffer`, need I fix it together?

src/System.Device.Gpio/System/Device/I2c/Drivers/UnixI2cDevice.Linux.cs
ZhangGaoxing: No, it works well.

src/System.Device.Gpio/System/Device/I2c/Drivers/UnixI2cDevice.Linux.cs
joperezr: since you are already touching this code, do you mind adding a null check as well to both this and the Read method?

joperezr: Actually nvm, just realized that we don't need to do this because Spans are ref structs so they can't be null.

PR #459, opened 2019-05-29
GPIO-access without root-permissions

mickew:
Gets this when running DeviceApiTester.
pi@raspberrypi:~/tests $ ./DeviceApiTester gpio-button-event --button-pin 12 --led-pin 17
Driver=Default, Scheme=Logical, ButtonPin=12, LedPin=17, PressedValue=Rising, OnValue=1
Listening for button presses on GPIO Logical pin 12 . . .
Error: Setting a mode to a pin requires root permissions.

Added delay between _sysFSDriver.OpenPin and_sysFSDriver.SetPinMode so udev have time to set the permissions on all the virtual files.
According to this article [https://forum.up-community.org/discussion/2141/solved-tutorial-gpio-i2c-spi-access-without-root-permissions](url)

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -67,6 +67,7 @@ namespace System.Device.Gpio.Drivers
             InitializeSysFS();
 
             _sysFSDriver.OpenPin(pinNumber);
+            Thread.Sleep(100);
             _sysFSDriver.SetPinMode(pinNumber, GetModeForUnixDriver(_sysFSModes[pinNumber]));
 
             _sysFSDriver.AddCallbackForPinValueChangedEvent(pinNumber, eventTypes, callback);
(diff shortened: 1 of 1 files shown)


krwq:
@mickew how do you know udev hasn't had enough time to set the permissions yet? Assuming that is the culprit is there more reliable way of testing that rather than timing? I'd prefer if we removed any timing based code (unless spec calls out to do so and mentions the specific minimum time). Perhaps query it somehow? Or check the permissions? Presumably if you don't have permissions the openning pin won't work at all and if user had rights then the file should have permissions so that user can also open it? (otherwise seems like a linux driver bug and I don't believe we should fix it here)

@joperezr does libgpiod also require root permissions? perhaps that could be used as a workaround

joperezr:
RaspberryPi driver doesn't require root permissions. I'm not 100% sure about libgpiod.

buyaa-n:
Libgpiod driver doesn't require root permission too

krwq:
@mickew so the workaround for you would be to use:
https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs#L12
with
https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs#L32
instead of GpioController directly.

Please let me know if that helps

PR #462, opened 2019-05-29
Updated to switch expressions

shaggygi:
Updated to use switch expressions based on new C# 8.0 feature.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -201,21 +201,13 @@ namespace System.Device.Gpio.Drivers
         /// <param name="mode">The mode of a pin to set the resistor pull up/down mode.</param>
         private void SetInputPullMode(int pinNumber, PinMode mode)
         {
-            byte modeToPullMode;
-            switch (mode)
+            byte modeToPullMode = mode switch
             {
-                case PinMode.Input:
-                    modeToPullMode = 0;
-                    break;
-                case PinMode.InputPullDown:
-                    modeToPullMode = 1;
-                    break;
-                case PinMode.InputPullUp:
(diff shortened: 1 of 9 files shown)


src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -201,21 +201,13 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
...
-                case PinMode.InputPullDown:
-                    modeToPullMode = 1;
-                    break;
-                case PinMode.InputPullUp:
-                    modeToPullMode = 2;
-                    break;
-                default:
-                    throw new ArgumentException($"{mode} is not supported as a pull up/down mode.");
-            }
+                PinMode.Input => (byte)0,
```
joperezr: NIT: Do you need the explicit cast here? I wonder why that wasn't required before.

joperezr:
Left a small question but looks good to me otherwise. Thanks a lot for doing this @shaggygi

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
shaggygi: Yeah, I was scratching my head on this one.  I would get the following error if I didn't cast.

"Error	CS0266	Cannot implicitly convert type 'int' to 'byte'. An explicit conversion exists (are you missing a cast?)"

I tried a few ways, but this approach was the only thing I found to satisfy.  Is there someone on the Roslyn team to ping that could take a quick look?

joperezr: I see, I guess that makes sense as we would be loosing precision I just don't understand why that wasn't the case before, in any case I don't think this should block your PR from merging.

PR #469, opened 2019-06-02
Resolved issue #416 - translating the dht22 bytes to temperature

### LINKED ISSUE #416: Dht22 example doesn't work

Rickvanderwaal:
Fixes #416

### DIFF
--- a/src/devices/Dhtxx/Devices/Dht11Device.cs
+++ b/src/devices/Dhtxx/Devices/Dht11Device.cs
@@ -1,8 +1,12 @@
-﻿using Iot.Units;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Iot.Units;
 
 namespace Dhtxx.Devices
 {
-    public class Dht11Device : IDhtDevice
+    internal class Dht11Device : IDhtDevice
     {
         public double GetHumidity(byte[] readBuff)
         {
--- a/src/devices/Dhtxx/Devices/Dht11Device.cs
+++ b/src/devices/Dhtxx/Devices/Dht11Device.cs
@@ -1,0 +1,18 @@
+﻿using Iot.Units;
+
+namespace Dhtxx.Devices
+{
+    public class Dht11Device : IDhtDevice
+    {
+        public double GetHumidity(byte[] readBuff)
+        {
+            return readBuff[0] + readBuff[1] * 0.1;
+        }
+
+        public Temperature GetTemperature(byte[] readBuff)
+        {
+            var temp = readBuff[2] + readBuff[3] * 0.1;
+            return Temperature.FromCelsius(temp);
+        }
+    }
+}
--- a/src/devices/Dhtxx/Devices/Dht22Device.cs
+++ b/src/devices/Dhtxx/Devices/Dht22Device.cs
@@ -1,8 +1,12 @@
-﻿using Iot.Units;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Iot.Units;
 
 namespace Dhtxx.Devices
 {
-    public class Dht22Device : IDhtDevice
+    internal class Dht22Device : IDhtDevice
     {
         public double GetHumidity(byte[] readBuff)
         {
--- a/src/devices/Dhtxx/Devices/Dht22Device.cs
+++ b/src/devices/Dhtxx/Devices/Dht22Device.cs
@@ -1,0 +1,21 @@
+﻿using Iot.Units;
+
+namespace Dhtxx.Devices
+{
+    public class Dht22Device : IDhtDevice
+    {
+        public double GetHumidity(byte[] readBuff)
+        {
+            return (readBuff[0] * 256 + readBuff[1]) * 0.1;
+        }
+
+        public Temperature GetTemperature(byte[] readBuff)
+        {
+            var temp = (readBuff[2] & 0x7F) + readBuff[3] * 0.1;
+            // if MSB = 1 we have negative temperature
+            temp = ((readBuff[2] & 0x80) == 0 ? temp : -temp);
+
+            return Temperature.FromCelsius(temp);
+        }
+    }
+}
--- a/src/devices/Dhtxx/Devices/Dht22Device.cs
+++ b/src/devices/Dhtxx/Devices/Dht22Device.cs
@@ -10,7 +10,7 @@ namespace Dhtxx.Devices
     {
         public double GetHumidity(byte[] readBuff)
         {
-            return (readBuff[0] * 256 + readBuff[1]) * 0.1;
+            return (readBuff[0] << 8 | readBuff[1]) * 0.1;
         }
 
         public Temperature GetTemperature(byte[] readBuff)
--- a/src/devices/Dhtxx/Devices/IDhtDevice.cs
+++ b/src/devices/Dhtxx/Devices/IDhtDevice.cs
@@ -1,0 +1,11 @@
+﻿using Iot.Units;
+
+namespace Dhtxx.Devices
+{
+    public interface IDhtDevice
+    {
+        double GetHumidity(byte[] readBuff);
+
+        Temperature GetTemperature(byte[] readBuff);
+    }
+}
--- a/src/devices/Dhtxx/Devices/IDhtDevice.cs
+++ b/src/devices/Dhtxx/Devices/IDhtDevice.cs
@@ -1,4 +1,8 @@
-﻿using Iot.Units;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Iot.Units;
 
 namespace Dhtxx.Devices
 {
--- a/src/devices/Dhtxx/DhtSensor.cs
+++ b/src/devices/Dhtxx/DhtSensor.cs
@@ -294,15 +294,15 @@ namespace Iot.Device.DHTxx
         {
             if (IsLastReadSuccessful)
             {
-                var temp = _readBuff[2] + (_readBuff[3] & 0x7F) * 0.1;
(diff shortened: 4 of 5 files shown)


src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -294,15 +294,15 @@ private double GetTempDht22()
         {
```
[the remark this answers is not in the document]
shaggygi: > This projects support DHT11, DHT12, DHT21(AM2301), DHT22(AM2302).
> Please rename the existing method GetTempDht22 to GetTempDht12.

I haven't looked at these devices or the current code, but would it be better to create an abstract and just call `GetTemp`?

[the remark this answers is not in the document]
ZhangGaoxing: The current model is not complicated, because there are only four devices and three methods. But I don't have a strong preference on any. 😄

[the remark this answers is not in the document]
joperezr: I agree with @shaggygi, having one abstract method implemented on two different devices would be preferable.

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -294,15 +294,15 @@ private double GetTempDht22()
...
             if (IsLastReadSuccessful)
             {
-                var temp = _readBuff[2] + (_readBuff[3] & 0x7F) * 0.1;
+                var temp = ((_readBuff[2] & 0x7F) * 256 + _readBuff[3]) * 0.1;
```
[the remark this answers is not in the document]
Rickvanderwaal: you are right. The * 256 part is not necessary here. I just tested it. I will remove that extra step.

src/devices/Dhtxx/DhtSensor.cs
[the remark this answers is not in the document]
Rickvanderwaal: I agree that would be better. I will make a proposal in this PR based on the Strategy Pattern.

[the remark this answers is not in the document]
Rickvanderwaal: I have updated the pull request. Please let me know what you think. Feel free to give feedback.
I tested this version on my Dht22 device on raspberry pi model 3 b+ with raspbian. 
image

@ZhangGaoxing can you do some tests with your devices too?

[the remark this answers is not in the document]
ZhangGaoxing: Sorry, I only have DHT11 and DHT12. But now the results seem to be correct.👍

src/devices/Dhtxx/Devices/Dht11Device.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Iot.Units;
```
ZhangGaoxing: missing dotnet license

src/devices/Dhtxx/Devices/Dht11Device.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Dhtxx.Devices
+{
+    public class Dht11Device : IDhtDevice
```
ZhangGaoxing: I think the class should be set to `internal`.😄

src/devices/Dhtxx/Devices/Dht22Device.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Iot.Units;
+
+namespace Dhtxx.Devices
+{
+    public class Dht22Device : IDhtDevice
+    {
+        public double GetHumidity(byte[] readBuff)
+        {
+            return (readBuff[0] * 256 + readBuff[1]) * 0.1;
```
ZhangGaoxing: readBuff[0] + readBuff[1] * 0.1;

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -25,24 +25,24 @@ public class DhtSensor : IDisposable
...
 
         /// <summary>
         /// How last read went, <c>true</c> for success, <c>false</c> for failure
         /// </summary>
         public bool IsLastReadSuccessful { get; private set; }
 
+        /// <summary>
+        /// Get the device for which this sensor instance is configured
+        /// </summary>
+        public IDhtDevice Device { get; set; }
```
ZhangGaoxing: private field

src/devices/Dhtxx/Devices/Dht22Device.cs
Rickvanderwaal: @ZhangGaoxing That formula is not working for me with DHT22.
The first 2 bytes (16 bits) of humidity is a 16-bit representation of 10 times the actual humidity value. So when `readBuff[0] = 00000010 (2)` and `readBuff[1] = 10011011 (155)` your formula gives the following output `2 + 155 * 0.1 = 17.5 %`. 

The right humidity should be `0000001010011010 (667) * 0.1 = 66.7 %`.

This can be achieved with the following formulas:
* `(readBuff[0] << 8 | readBuff[1]) * 0.1;`
* `(readBuff[0] * 256 + readBuff[1]) * 0.1;`

I don't know why it is working on the DHT11, do you have any suggestion?

ZhangGaoxing: You are right. I just read the DHT22 datasheet and found that the formula for calculating the results is different from DHT11.

Rickvanderwaal: I do have both devices now. Tested with DHT11 and DHT22 and both working fine.

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -106,7 +80,7 @@ public DhtSensor(int pin, DhtType dhtType, PinNumberingScheme pinNumberingScheme
             _protocol = CommunicationProtocol.OneWire;
             _controller = new GpioController(pinNumberingScheme);
             _pin = pin;
-            _dhtType = dhtType;
+            SetDevice(dhtType);
```
joperezr: Why do we even have a DhtType still, can't we just instead change the constructor to take a IDhtDevice instead? Seems to me that DhtType is not really useful anymore but I may be mistaken.

joperezr: Oh nvm, I just saw that dhtType is simply an enum. I like this new refactoring 👍

src/devices/Dhtxx/Devices/IDhtDevice.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Iot.Units;
+
+namespace Dhtxx.Devices
+{
+    public interface IDhtDevice
```
joperezr: I suppose that this is the only two things that all dht devices can do? if there are any Dhtxx that have more functionality than these two things, I would suggest using an abstract base class instead of interface, since abstract base class can add members in the future, but interface can't.

ZhangGaoxing: At present, the DHT series sensors only have the function of temperature and humidity, and the DHT series has not been updated in these years. Sensors with other functions are not named DHT by ASAIR. So I think all of them can.

Rickvanderwaal: Well i can imagine that we have to add some generic methods for all devices in the future. So an abstract class wouldn’t be an bad idea.

joperezr: I'm fine with this as is today, so I'll go ahead and merge in order for people currently hitting an issue with this binding to have a new version that works. We can decide later if we want to have an abstract base class instead of an interface.

PR #477, opened 2019-06-06
Add Bme680 binding

georgemathieson:
Initial attempt at adding a binding for the Bme680 sensor. This includes core temperature, pressure and humidity functionality, unit testing, and simple sample.

This only includes the I2C interface for now. I would like to do the gas sensor functionality too but I think it would be best to get some initial feedback first on what I've done so far 😀


shaggygi:
@georgemathieson thanks for the interest and working on this binding 👍 

If I could provide some quick feedback... I have not looked at this device, but it appears to be a family of [BME-type sensors](https://www.bosch-sensortec.com/bst/products/all_products/.  In addition, there is already a [Bmx280 binding](https://github.com/dotnet/iot/tree/master/src/devices/Bmx280).  

Would it be possible to review both and see if it is worth creating a binding that works with all under that family?  Thx

georgemathieson:
Thanks @shaggygi, looking at [BmxBase.cs](https://github.com/dotnet/iot/blob/master/src/devices/Bmx280/BmxBase.cs) it seems like there are some fundamental differences, for instance, the registers for accessing temperature and pressure data are different.

On the other hand, the code used to calculate the pressure and humidity is the same and it seems sensible to only have once copy that's shared between the BmXX80 family (I guess that's a suitable name? 😄).

Would it still be a good idea to try and refactor in this branch? The [Bmp180](https://github.com/dotnet/iot/tree/master/src/devices/Bmp180) seems to fall under the same family but is currently separate of the Bmx280 binding and I'm unsure if there's a reason for that.

shaggygi:
@georgemathieson yup, it looks like the register mapping is different between them.  It's a shame they didn't at least keep the related sensors (temp, etc.)  the same.  I guess it might be worth separating the bindings then, but I'll let @joperezr and @krwq chime in.

Regarding the Bmp180, it appears the Bme280 is compatible and next generation.  I'm not sure why there was a separate binding between them.

Just a quick note of what looks like each one supports (but should be double-checked).

BMP280 
	I2C, SPI
	Temperature, Barometric pressure
BME280 
	I2C, SPI
	Temperature, Barometric pressure, Humidity
BME680
	I2C, SPI
	Temperature, Barometric pressure, Humidity, VOC gas

joperezr:
Hey @georgemathieson thanks a lot for you contribution! I think I agree with @shaggygi here and it would be great if we could avoid code duplication as much as possible, as long as it still makes sense. If the differences are the representation of registers and the new sensor data (VOC gas plus its calibration data) then I would suggest to try to refactor the existing code in order to add this binding. From what I can see at least it doesn't look like it would be too bad to just refactor how we handle those registers, but I may be wrong. If we end up having to refactor much more than that in order to add this support, then I would probably prefer to keep this binding by itself, but again, I believe that might not be the case. @georgemathieson what do you think?

georgemathieson:
@joperezr yep, that sounds sensible.

georgemathieson:
Hey @joperezr I've had an initial go at refactoring the BME680 binding into the existing Bmx280 project. For it to make sense, I've renamed the Bmx280 project to Bmxx80.

I've tried to keep the names of registers the same as they were in the Bmx280 project. The `CompensatePressure` method was _so close_ to being the same, but it seems the BME680 has an extra coefficient (DigP10). The `CompensateTemperature` formula is the same for all devices though.

Anything that is common among all devices can be found in [Bmxx80Base.cs](https://github.com/dotnet/iot/pull/477/files#diff-7f1da8c456dd0b49d433e8f718fdb671) and things that are common in the BME280 and BMP280 can be found in [Bmx280Base.cs](https://github.com/dotnet/iot/pull/477/files#diff-12c6dd318d70152daa1d4fde59a140d6).

Hopefully this will shed some light on what the commonalities are between these sensors. The BME280 and BMP280 devices still need testing (I don't own them so can't test them myself), but I welcome any feedback on the refactor 😄

joperezr:
Thanks a lot for the refactoring @georgemathieson! I will take a look at this as soon as I finish one thing I have in progress.

georgemathieson:
Thanks @joperezr. I have brought in the fix for compensate temperature #484 by @RobinTTY too.

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -0,0 +1,207 @@
...
+        /// </summary>
+        protected int TemperatureFine;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Bmxx80Base"/> class.
+        /// </summary>
+        /// <param name="deviceId">The ID of the device.</param>
+        /// <param name="i2cDevice">The <see cref="I2cDevice"/> to create with.</param>
+        /// <exception cref="ArgumentNullException"></exception>
+        /// <exception cref="IOException"></exception>
```
joperezr: NIT: While we are adding it, do you mind also adding a brief text here explaining when are the exceptions thrown?

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        /// <summary>
+        /// Sets the pressure sampling.
+        /// </summary>
+        /// <param name="sampling">The <see cref="Sampling"/> to set.</param>
+        public void SetPressureSampling(Sampling sampling)
+        {
+            byte status = Read8BitsFromRegister(_controlRegister);
+            status = (byte)(status & 0b1110_0011);
```
joperezr: while I'm fine with this, ususally we do something like:
```c#
status &= (byte)~(0b111 << 2);
```

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -0,0 +1,207 @@
...
+using Iot.Units;
+
+namespace Iot.Device.Bmxx80
+{
+    /// <summary>
+    /// Represents the core functionality of the Bmxx80 family.
+    /// </summary>
+    public abstract class Bmxx80Base : IDisposable
+    {
+        internal Bmxx80CalibrationData _calibrationData;
```
joperezr: Should all of these internal fields be protected instead? That way, we can have people define their own BME980 (for example) binding in their own project in the future and still use the logic in this base class.

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -0,0 +1,207 @@
...
+
+        /// <summary>
+        /// Compensates the temperature.
+        /// </summary>
+        /// <param name="adcTemperature">The temperature value read from the device.</param>
+        /// <returns>The <see cref="Temperature"/>.</returns>
+        internal Temperature CompensateTemperature(int adcTemperature)
+        {
+            //Formula from the datasheet
+            //The temperature is calculated using the compensation formula in the BMP280 datasheet
```
joperezr: NIT: can you point to the page where you got this from in the datasheet? that can save some time in the future.

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        /// <summary>
+        /// Recommended wait timings from the datasheet.
+        /// </summary>
+        /// <param name="sampleMode">The <see cref="Sampling"/> to get for.</param>
+        /// <returns>The time it takes for the chip to read data in milliseconds rounded up.</returns>
+        private int GetMeasurementTimeForForcedMode(Sampling sampleMode)
+        {
+            if (sampleMode == Sampling.UltraLowPower)
```
joperezr: NIT: probably we can use the new switch c# 8 feature. Like:
```c#
return sampleMode switch
{
  Sampling.UltraLowPower => 7,
  Sampling.LowPower => 9,
  Sampling.Standard => 14
  //..
  _ => 0
};
```

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// </summary>
+        /// <returns>The current <see cref="Bmx280PowerMode"/>.</returns>
+        public Bmx280PowerMode ReadPowerMode()
+        {
+            byte read = Read8BitsFromRegister(_controlRegister);
+
+            // Get only the power mode bits.
+            var powerMode = (byte)(read & 0b_0000_0011);
+
+            return (Bmx280PowerMode)powerMode;
```
joperezr: NIT: Looks like PowerMode only supports 0b00 and 0b01 so we should validate that what we got is not 0b10 or 0b11.

src/devices/Bmxx80/Bme680.cs
```diff
@@ -0,0 +1,235 @@
...
+            _i2cDevice.Write(new[] { _controlRegister, (byte)(cleared | (byte)powerMode) });
+        }
+
+        /// <summary>
+        /// Compensates the humidity.
+        /// </summary>
+        /// <param name="temperature">The temperature to use.</param>
+        /// <param name="adcHumidity">The humidity value read from the device.</param>
+        /// <returns>The percentage relative humidity.</returns>
+        private double CompensateHumidity(double temperature, int adcHumidity)
```
joperezr: I suppose these compensation methods are very different to the ones in the 280Base?

src/devices/Bmxx80/Bme680.cs
```diff
@@ -0,0 +1,235 @@
...
+{
+    /// <summary>
+    /// Represents a BME680 temperature, pressure, relative humidity and VOC gas sensor.
+    /// </summary>
+    public class Bme680 : Bmxx80Base
+    {
+        /// <summary>
+        /// Gets a value indicating whether new data is available.
+        /// </summary>
+        public bool HasNewData => ReadHasNewData();
```
joperezr: In order to be a bit more similar compared to the 280 devices, can we instead of having a property accesor to instead have ReadHumidity, ReadPreassure and all the rest public and remove these properties? It would actually be great if we could just have an internal interface that all devices implemented so that way they have a shared API design. What do you think?

joperezr:
Left a few comments, most of them are very little NITs and very small changes. The only design-like question is the last one where I propose we have a similar shape of accessing sensor data across all three bindings as opposed to one using properties and the other two read methods. Other than that, this looks awesome @georgemathieson thanks so much for putting the effort here 😄

src/devices/Bmxx80/Bme680.cs
georgemathieson: Something like `ITemperature`, `IPressure` `IHumidity`?

krwq: @georgemathieson I've already tried: #256 - I still think we will eventually end up doing it but it is better to spend a bit more time trying to figure out if we need some more data structures (i.e. we have alredy added Temperature struct since then) - once people start using interfaces it will be hard to make any changes to them

georgemathieson:
Thanks for the review @joperezr I shall begin making those changes 😀 🚧

src/devices/Bmxx80/Bmxx80Base.cs
georgemathieson: This looks like it would need to be `status &= unchecked((byte)~(0b111 << 2));`, or am I missing something obvious? 😅

RobinTTY: I wrestled with the same issue and that's down to how C# handles byte values, so your ```0b111``` gets treated as an integer and the unchecked bitwise complement will therefore also return an integer __which is negative__ so this will give you a wrong result.

The way you can remedy this is to define the ```0b111``` explicitly as a byte or what I did was to define Masks in a enum as byte values which also removes a bit of redundancy when you are using those values for both getters and setters.

This helped me wrap my head around this: https://www.oreilly.com/library/view/c-cookbook/0596003390/ch01s04.html

RobinTTY: To be a little more precise it's not exactly down to how C# handles byte values but that 0b111 is just another way of expressing an int value.

krwq: @georgemathieson: `status &= unchecked((byte)~(0b111 << 2));` looks correct, although I think the original version is more readable in this case so I'd leave as is

RobinTTY: oh @krwq is right, ignore my comment I had a wrong understanding of the unchecked syntax 😨. Just tested it and your code would give the right result, my bad!

src/devices/Bmxx80/Bmx280Base.cs
georgemathieson: Would the expected behavior be at that point to throw an exception? e.g.
```csharp
if (Enum.IsDefined(typeof(Bmx280PowerMode), powerMode) == false)
{
    throw new IOException("Read unexpected power mode");
}
```

src/devices/Bmxx80/Bmx280Base.cs
georgemathieson: Oh that's cool!

src/devices/Bmxx80/Bme680.cs
georgemathieson: I believe they are. The only common one is temperature.

src/devices/Bmxx80/Bmxx80Base.cs
georgemathieson: Shall leave as is for now as I agree with the original being more readable.

georgemathieson:
@joperezr @krwq Thanks guys, I have made those changes.

@Silvenga I've merged in your #435 for humidity support. It would be cool if you could double check I've not just ruined all your good work 😅. I noticed you fixed a few things like the default i2c address and added some extra support like chip reset/IIR filter which is common among the Bmxx80 family of sensors. So where it makes sense, I've been able to merge that functionality so the other sensors can also use it.

Silvenga:
Looks good @georgemathieson, tested with my BME280. Just three things I noticed:

- It might be a good idea to bring the sample project changes for `Bme280.sample` over.
- I fixed a rounding error in Bmx280Base.ReadPressureAsync, I don't think that fix made it over.
- I updated the PowerMode comments to match the docs. It might be good to bring those over too.

joperezr:
@georgemathieson changes look good to me now 😄 Thanks so much for your hard work. I agree on all three comments from @Silvenga though, so as soon as you can add tackle those three, I think this will be good to go.

georgemathieson:
Thank you @Silvenga and @joperezr for your comments, I have made those changes.

georgemathieson:
Hi @joperezr need your help as I'm not sure why the tests are failing 🤔

krwq:
@georgemathieson we will remove `System.Device.I2c.Devices` very soon (I2c will be moved to just `System.Device.I2c`). While we transition you might see some weird failures. I'd refrain from changing those right now.

joperezr:
@krwq it's fine to change them in case his PR goes in first. If not, then I will push the fix directly into this PR.

joperezr:
Hey @georgemathieson sorry that this PR went a bit stale in terms of reviews, I'm taking a look at this now. One initial comment is that looks like we added the category.txt file in a different PR into the old Bmx280 directory and that is not being moved to the new folder with your PR. Do you mind moving that?

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -0,0 +1,215 @@
...
+                return Sampling.UltraHighResolution;
+            }
+
+            return (Sampling)value;
+        }
+
+        /// <summary>
+        /// Cleanup.
+        /// </summary>
+        public void Dispose()
```
joperezr: Can we instead implement the Dispose pattern here?

```c#
public void Dispose()
{
  Dispose(true);
  GC.SuppressFinalize(this);
}

{
  _i2cDevice?.Dispose();
  _i2cDevice = null;
}
```

That way if nested classes want to do their own dispose, all of the parent's ones will get called as well.

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -0,0 +1,291 @@
...
+        /// <returns>Atmospheric pressure in Pa.</returns>
+        public async Task<double> ReadPressureAsync()
+        {
+            if (ReadPowerMode() == Bmx280PowerMode.Forced)
+            {
+                await Task.Delay(GetMeasurementTimeForForcedMode(ReadPressureSampling()));
+            }
+
+            //Read the temperature first to load the t_fine value for compensation
+            if (TemperatureFine == int.MinValue)
```
joperezr: I might be missing it but where is TemperatureFine being initialized to int.MinValue? I don't see that happening in the constructor so technically I believe this guy will be defaulted to 0 right? Should we initialize it to int.MinValue in the constructor of this base class?

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -0,0 +1,291 @@
...
+        /// <summary>
+        /// Sets the power mode to the given mode
+        /// </summary>
+        /// <param name="powerMode">The <see cref="Bmx280PowerMode"/> to set.</param>
+        public void SetPowerMode(Bmx280PowerMode powerMode)
+        {
+            byte status = Read8BitsFromRegister(_controlRegister);
+
+            // Clear last two bits.
+            status = (byte)(status & 0b_1111_1100);
```
joperezr: NIT: This could technically also be converted into a one-line as the other sets where you combine both bitwise operations.

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -0,0 +1,291 @@
...
+        /// Reads the currently configured standby time mode the device will used when operating in normal mode.
+        /// </summary>
+        /// <returns>The current <see cref="StandbyTime"/>.</returns>
+        public StandbyTime ReadStandbyTime()
+        {
+            byte current = Read8BitsFromRegister((byte)Bmx280Register.CONFIG);
+            var time = (byte)((current & 0b_1110_0000) >> 5);
+
+            return time switch
+            {
```
joperezr: Can't you just cast the byte into a StandbyTime here?

src/devices/Bmxx80/Bme280.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <returns>Returns a percentage from 0 to 100.</returns>
+        public async Task<double> ReadHumidityAsync()
+        {
+            if (ReadPowerMode() == Bmx280PowerMode.Forced)
+            {
+                await Task.Delay(GetMeasurementTimeForForcedMode(ReadHumiditySampling()));
+            }
+
+            // Read the temperature first to load the t_fine value for compensation.
+            if (TemperatureFine == int.MinValue)
```
joperezr: Same comment about TemperatureFine

src/devices/Bmxx80/Bme680.cs
```diff
@@ -0,0 +1,230 @@
...
+            var hasNewData = (byte)(read & 0b_1000_0000);
+
+            return (hasNewData >> 7) == 1;
+        }
+
+        /// <summary>
+        /// Read the humidity.
+        /// </summary>
+        /// <returns>Calculated humidity.</returns>
+        public double ReadHumidity()
```
joperezr: For consistency sake and given that we also support PowerMode with the 680, should we make these methods Async and return a Task<double> which does something like:

```c#
            if (ReadPowerMode() == Bmx280PowerMode.Forced)
            {
                await Task.Delay(GetMeasurementTimeForForcedMode(ReadHumiditySampling()));
            }
```

src/devices/Bmxx80/Bme680.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// <returns>The current <see cref="Bme680PowerMode"/>.</returns>
+        /// <exception cref="NotImplementedException">Thrown when the power mode does not match a defined mode in <see cref="Bme680PowerMode"/>.</exception>
+        public Bme680PowerMode ReadPowerMode()
+        {
+            byte read = Read8BitsFromRegister(_controlRegister);
+
+            // Get only the power mode bits.
+            var powerMode = (byte)(read & 0b_0000_0011);
+
+            return powerMode switch
```
joperezr: Cast instead of a switch?

src/devices/Bmxx80/Bmx280Base.cs
georgemathieson: Yup, good spot! Shall initialise in the constructor of this base class.

src/devices/Bmxx80/Bme280.cs
georgemathieson: Resolved by Bmx280base's constructor which this inherits from

src/devices/Bmxx80/Bme680.cs
georgemathieson: Agree with making it consistent but the BME680 doesn't have any specified wait times in the datasheet for reading data like the BMx280 sensors do. Instead, there's a flag that indicates when new data is available to be read. Thoughts?

src/devices/Bmxx80/Bmxx80Base.cs
georgemathieson: Done! 😃

georgemathieson:
@joperezr thanks for the review 😊. I have left one comment though on the power mode timings. Otherwise, I think this is good.

src/devices/Bmxx80/Bme680.cs
joperezr: If we really wanted to follow the framework design guidelines and designed this like a corefx class in here, we would have something like this:

```c#
public abstract class SensorBaseClass
{
    public double ReadHumidity()
    {
        // Perform Setup, validation and other things that all devices have in common.
        return ReadHumidityCore();
    }

    public Task<double> ReadHumidityAsync()
    {
        //Perform Setup, validation and other things that all devices have in common.
        return ReadHumidityAsyncCore();
    }

    protected abstract double ReadHumidityCore();
    protected virtual Task<double> ReadHumidityAsyncCore()
    {
        return Task.FromResult(ReadHumidityCore());
    }
}

public class MySensorWhichDoesntWorkAsync : SensorBaseClass
{
    protected override double ReadHumidityCore()
    {
        // perform sync read here
    }
}

public class MySensorWhichWorksAsync : SensorBaseClass
{
    protected override Task<double> ReadHumidityAsyncCore()
    {
        // perofrm async read.
    }
    protected override double ReadHumidityCore()
    {
        return ReadHumidityAsyncCore().GetAwaiter().GetResult();
    }
}
```

All of that said, you don't have to do this now in this PR, for now, what you can do is to simply also offer a `ReadHumidityAsync` method (and all other values) which simply return a `Task.FromResult(ReadHumidity())`. I guess you get to choose if you want to do the full design patter from the framework guidelines, or if you just want to add consistency for now.

georgemathieson: I think ultimately the corefx class pattern is the way to go... However, I've gone with consistency _for now_ as there's quite a lot in this PR and it would be great to get this in 😄

georgemathieson:
Thanks @joperezr, I've made those changes 😀

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -0,0 +1,279 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Bmx280Base"/> class.
+        /// </summary>
+        /// <param name="deviceId">The ID of the device.</param>
+        /// <param name="i2cDevice">The <see cref="I2cDevice"/> to create with.</param>
+        protected Bmx280Base(byte deviceId, I2cDevice i2cDevice)
+            : base(deviceId, i2cDevice)
+        {
+            _controlRegister = (byte)Bmx280Register.CTRL_MEAS;
+            TemperatureFine = int.MinValue;
```
joperezr: Won't this also need to be assigned in Bme680 case? that one doesn't inherit from this base class, and it does use it on Bme680.CompensatePressure. Perhaps you want to initialize this on Bmxx80Base?

src/devices/Bmxx80/Bme680.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        /// <summary>
+        /// Compensates the pressure.
+        /// </summary>
+        /// <param name="adcPressure">The pressure value read from the device.</param>
+        /// <returns>The pressure in Pa.</returns>
+        private double CompensatePressure(int adcPressure)
+        {
+            // Calculate the pressure.
+            var var1 = (TemperatureFine / 2.0) - 64000.0;
```
joperezr: Will TemperatureFine have the right value here? Shouldn't CompensatePressure call CompensateTemperature first in order to ensure that TemperatureFine has been correctly initialized?

src/devices/Bmxx80/Bmx280Base.cs
georgemathieson: Thinking about this, do we really want this check for the TemperatureFine value? Ideally, we always want to have a fresh reading of temperature for pressure.
```
            //Read the temperature first to load the t_fine value for compensation
            if (TemperatureFine == int.MinValue)
            {
                await ReadTemperatureAsync();
            }
```
I think the following code is trying to save a call to `ReadTemperatureAsync()` but once we've made that call, this is essentially never true again. Potentially, pressure will use an old temperature reading here which is going to effect the accuracy. I guess it depends on how the sensor class is used in an application (i.e. is it a new instance every time or is it injected in as a singleton) but I think in principle, it would be better to get an up-to-date reading of temperature.

joperezr: I think I agree, and I don't believe it will be a very expensive call, it is just performing another read and that way you will get a more accurate reading of pressure. I believe calling ReadTemperatureAsync always might be better.

src/devices/Bmxx80/Bme680.cs
georgemathieson: Ah, yes it should!

PR #479, opened 2019-06-06
Adding netstandard ridless configuration

### LINKED ISSUE #263: Generate notimpl netstandard 2.0 assembly for System.Device.Gpio
Today we have a reference assembly for netstandard2.0, but we don't have a runtime assembly for each specific platform/RID where netstandard2.0 is supported (currently, we only have a windows and a linux implementation). We should add a not implemented netstandard 2.0 assembly to the package so that if it is restored when targeting a different platform (like OSX or Xamarin) we can provide a runtime asset.

For this, we could leverage GenApi tool, which can help with generating the notimpl assembly for us.

joperezr:
@ericstj @safern 

fixes #263 

This is adding the RID-less configuration to the package that just throws PNSE.


eng/Tools.props
```diff
@@ -0,0 +1,16 @@
...
+    <!--When we are doing SDK validation we place the locally built packages in a custom path. This has to be propagated so the next
+        build can restore the local packages.-->
+    <RestoreSources Condition="'$(AdditionalRestoreSources)' != ''">
+        $(RestoreSources);
+        $(AdditionalRestoreSources)
+    </RestoreSources>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.DotNet.GenAPI" Version="$(MicrosoftDotNetGenApiPackageVersion)" />
```
ericstj: Can you instead reference this in the project itself?  The Tools.proj.nuget.g.targets import hack isn't something we should do in other places...

src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -42,4 +43,54 @@
       <IsWinMDFile>true</IsWinMDFile>
     </Reference>
   </ItemGroup>
+
+  <PropertyGroup>
+    <NotSupportedSourceFile>$(IntermediateOutputPath)$(MSBuildProjectName).notsupported.cs</NotSupportedSourceFile>
+  </PropertyGroup>
+  <Target Name="GenerateNotSupportedAssemblyForNetStandardSource"
```
ericstj: Rather than running this in the runtime-specific project build, I think you can run https://github.com/dotnet/arcade/blob/a1805103791e43031355e11c0d037bca803a9593/src/Microsoft.DotNet.GenAPI/build/Microsoft.DotNet.GenAPI.targets#L73 in the outer build and just get the reference assembly built in the rid-specific build.  That way the only thing you need to couple to is the ref assembly path, which I believe you already have.

src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -42,4 +43,54 @@
...
+
+  <Target Name="SetupEnvironmentForNotSupportedAssembly">
+    <PropertyGroup>
+      <OutDir>$(OutDir)notsupported\</OutDir>
+      <ProjectDepsFilePath>$(OutDir)$(ProjectDepsFileName)</ProjectDepsFilePath>
+    </PropertyGroup>
+    <ItemGroup>
+      <Compile Remove="@(Compile)" />
+      <Compile Include="$(IntermediateOutputPath)..\$(MSBuildProjectName).notsupported.cs" />
+      <CreateDirectory Include="$(OutputPath)notsupported\" />
```
ericstj: Here you use OutputPath and above you use OutDir.  I wonder if you even need to change this since nothing else will be building to a path without a RID?

joperezr: I don't need to change it, but I just like the output better when it is clear to me which assembly is which, today in the artifacts/bin/Debug/System.Device.Gpio/netstandard2.0 folder I have four sub folders which make their content obvious: ref, notsupported, linux, win. If I remove this, then it would look like: ref, linux, win and then have a System.Device.Gpio.dll in that directory which wouldn't be to obvious that is the turd assembly.

That said, I can still change it if you think it's better.

joperezr:
Merging now in order to avoid merge conflicts when arcade update comes in. I will address any additional feedback on a subsequent PR. Thanks for the review @ericstj

PR #482, opened 2019-06-07
Add new functionality to Bme680 binding

RobinTTY:
This pull request adds a driver for the Bosch Bme680 temperature, humidity, pressure and air quality sensor. I ported it from the offical [Bosch C driver](https://github.com/BoschSensortec/BME680_driver).

I ported the full API but not the full SPI binding since I don't have a device to test it on, the template is there but some of the address translation that is necessary is missing.

When working on this I saw that the binding is very similar to the Bme280/Bmp280/Bmp180. Are there any thoughts on combining these because as of right now this is a lot of redundant functionality.

I tested the binding on Windows Iot Core and Raspbian and got matching readings to other libraries for this driver so I think everything should work fine.


RobinTTY:
Oh just saw there is another recent pull request on this exact binding #477 😆...

joperezr:
Hey @RobinTTY thaks a lot for your contribution and the interest in this binding! 😄 

As you noticed, we do already have an open PR for this same binding, so is there any chance that you can join efforts with @georgemathieson in order to add support for this binding?

RobinTTY:
Hey @joperezr,

my binding differs a bit from @georgemathieson but generally he could incorporate my implementation for reading the gas resistance and the additional functionality or I could do that for him, whatever is better for him.

joperezr:
I believe @georgemathieson will attempt to do some refactoring on the existing Bmx280 binding so that we could leverage the code already there for this new binding, which if possible, is ideally what we want. Once he does that, then we could leverage your implementation for the new functionality into it.

RobinTTY:
@joperezr sounds good to me!

georgemathieson:
Awesome!

joperezr:
@RobinTTY hey just as FYI, we finally merged the Bme680 PR, so we should be good to go if you now want to rebase this and add the extra functionality that the sensor provides.

RobinTTY:
@joperezr Alright, I'll do that in the next few days.

RobinTTY:
@joperezr One question regarding the design of the API, if there is a setting that can be either true or false like turning on a heater module, would you prefer a property or a function? I saw different implementations in the existing bindings.

So the options I think about would look like this:

```C#
public void SetHeaterEnabled(bool status){ ... }     // true enables, false disables
```

or

```C#
public bool HeaterIsEnabled{ get { ... } set{ ... } }
```

joperezr:
I think for that particular case having a property should be fine. We recently had a chat about properties vs. methods but it was more around when trying to get data out of the sensor, like temperature for example. The guidelines say that for readings like that, we should use methods as properties in c# are expected to always return you the same value for consecutive reads (assuming obviously that the consumer didn't call the setter in between both calls) so we decided that for Temperature and other data that the sensors return, methods would be the best choice. For things like switches to turn on or off something like the one you propose, properties should be fine.

joperezr:
@RobinTTY Please ping me once this is ready for review 😄

krwq:
@richlander this was ported from repo with this license https://github.com/BoschSensortec/BME680_driver/blob/master/LICENSE - do we need to do anything special for using this?

RobinTTY:
@joperezr This is ready for review now. Doing some testing currently and measurements are looking good to me.

One improvement I haven't implemented yet is caching of values like the current sampling modes. Right now they are always read from the device when needed but they don't change until they are written trough the API (or on reset). So theoretically we wouldn't need to read them each time from the device which would improve performance. Should I add this?

joperezr:
> Should I add this?

That is up to you if you want to add it to this PR or if you want to add that as a separate change.

RobinTTY:
Alright I'll add it in this PR, since it's not a huge change anyways.

RobinTTY:
Now this is really ready for review 😄

src/devices/Bmxx80/Bme680.cs
```diff
@@ -51,49 +58,252 @@ public Bme680(I2cDevice i2cDevice)
...
-        public void SetHumiditySampling(Sampling sampling)
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <see cref="Bme680HeaterProfile"/> is set to an undefined profile.</exception>
+        public Bme680HeaterProfile CurrentHeaterProfile
         {
-            var register = (byte)Bme680Register.CTRL_HUM;
-            byte read = Read8BitsFromRegister(register);
+            get => _currentHeaterProfile;
+            set
+            {
+                if (_heaterConfigs.Exists(config => config.HeaterProfile == value))
```
RobinTTY: One note on this: The Bme680 can save up to 10 heater profiles in its registers. One of these 10 profiles can be set as current profile which will influence the gas resistance measurement. A heater profile needs to be configured by the user before being used to produce meaningful measurements (the registers are just initialized to 0x00).

If an assignment to CurrentHeaterProfile is done with an unconfigured profile, the assignment is ignored and remains set to the old profile. Theoretically we could also switch to an unconfigured profile or configure it automatically. Just wasn't sure what would be expected here.

src/devices/Bmxx80/Bme680.cs
```diff
@@ -51,49 +58,248 @@ public Bme680(I2cDevice i2cDevice)
...
+        /// <summary>
+        /// Gets or sets the humidity sampling.
+        /// </summary>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <see cref="Sampling"/> is set to an undefined mode.</exception>
+        public Sampling HumiditySampling
+        {
+            get => _humiditySampling;
+            set
+            {
+                if (!Enum.IsDefined(typeof(Sampling), value))
```
RobinTTY: Also wasn't sure if these setter should throw exceptions or just retain their value if an exception would be thrown.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property states:

> DO preserve the previous value if a property setter throws an exception. 

Just wasn't sure if these rules are used for this repo.

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -53,61 +53,54 @@ protected Bmxx80Base(byte deviceId, I2cDevice i2cDevice)
...
-        /// </summary>
-        /// <param name="sampling">The <see cref="Sampling"/> to set.</param>
-        public void SetPressureSampling(Sampling sampling)
-        {
-            byte status = Read8BitsFromRegister(_controlRegister);
-            status = (byte)(status & 0b1110_0011);
-            status = (byte)(status | (byte)sampling << 2);
-            _i2cDevice.Write(new[] { _controlRegister, status });
-        }
+        private Sampling _pressureSampling;
```
joperezr: NIT: very dumb thing, but would you mind moving these backup fields to the top of the class? That is usually how we have most of the bindings so just in order to be consistent.

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -66,9 +84,12 @@ public FilteringMode ReadFilterMode()
         /// <returns>Calculated temperature.</returns>
         public async Task<Temperature> ReadTemperatureAsync()
         {
+            if (TemperatureSampling == Sampling.Skipped)
```
joperezr: I guess I've never thought of it through with the sampling, but is there any scenario where somebody that have Sampling.Skipped set, and would call this method? I guess I don't understand why would somebody call this method when sampling is skipped, but it might be that they actually care about reading the temperature even when sampling was disabled.

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -200,34 +213,12 @@ public void SetPowerMode(Bmx280PowerMode powerMode)
...
-            return (StandbyTime)((current & 0b_1110_0000) >> 5);
-        }
-
         /// <summary>
         /// Recommended wait timings from the datasheet.
         /// </summary>
         /// <param name="sampleMode">The <see cref="Sampling"/> to get for.</param>
         /// <returns>The time it takes for the chip to read data in milliseconds rounded up.</returns>
-        internal int GetMeasurementTimeForForcedMode(Sampling sampleMode)
+        internal static int GetMeasurementTimeForForcedMode(Sampling sampleMode)
```
joperezr: why did we have to change this to static? all places where we call it we are already in an instance method.

src/devices/Bmxx80/Bmp280.cs
```diff
@@ -27,6 +27,8 @@ public Bmp280(I2cDevice i2cDevice)
             _communicationProtocol = CommunicationProtocol.I2c;
             _calibrationData = new Bmp280CalibrationData();
             _calibrationData.ReadFromDevice(this);
+
+            SetDefaultConfiguration();
```
joperezr: NIT: can we just add this call to the Bmxx80base constructor instead since presumably we want to set this for every device?

src/devices/Bmxx80/Bme680HeaterProfile.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Iot.Device.Bmxx80
+{
+    /// <summary>
+    /// 10 addressable heater profiles stored on the Bme680.
+    /// </summary>
+    public enum Bme680HeaterProfile : byte
+    {
+        /// <summary>
+        /// Heater Profile 1.
```
joperezr: Is there any other info on what does each profile mean in the datasheet? Given that this enum is public, I think it might be confusing to know which one to pick. Unless consumers would likely understand what each profile mean.

src/devices/Bmxx80/Bme680.cs
```diff
@@ -51,49 +58,248 @@ public Bme680(I2cDevice i2cDevice)
...
+        }
+
+        private Bme680HeaterProfile _currentHeaterProfile;
+
+        /// <summary>
+        /// Gets or sets the heater profile to be used for measurements.
+        /// Current heater profile is only set if the chosen profile is configured.
+        /// </summary>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <see cref="Bme680HeaterProfile"/> is set to an undefined profile.</exception>
+        public Bme680HeaterProfile CurrentHeaterProfile
```
joperezr: NIT: `CurrentHeaterProfile` -> `HeaterProfile`

src/devices/Bmxx80/Bmx280Base.cs
RobinTTY: My thought with this was that it's just the most unambiguous way to indicate that no measurement was taken. There is probably less room for error if the method returns NaN instead of the last value measured so an out of date measurement is not mistakenly used.

The scenario where someone wants to read the last measured value even if its from many cycles ago seems unlikely to me since you would probably just log them when they are created.

src/devices/Bmxx80/Bmx280Base.cs
RobinTTY: You're right this is actually not necessary.

src/devices/Bmxx80/Bmp280.cs
RobinTTY: Yes and no, as it currently stands calling it from the base would result in an Exception because the calibrationData is not yet read (which happens in the derived constructor) and `SetDefaultConfiguration` in the Bme680 needs some of those values. Plus I thought it would be bad practice in general to call a virtual method from the base constructor but I'm still fairly new to C# so I could be wrong.

Reading the calibration data in the base constructor would theoretically be possible and would allow us to call SetDefaultConfiguration after but would require some changes as well.

src/devices/Bmxx80/Bme680HeaterProfile.cs
RobinTTY: These heater profiles are perfectly equal, they have no inherent settings associated with them. All profiles must be configured by the user through `ConfigureHeatingProfile`. It doesn't matter which profile is used.

RobinTTY: To make this a little clearer, there are in total 10x2 registers for these settings.
So for each profile a target temperature and duration for the heater. An additional register decides which of these profiles is used (by writing one of the values of the enum to it).

src/devices/Bmxx80/Bme280.cs
```diff
@@ -36,35 +39,33 @@ public Bme280(I2cDevice i2cDevice)
...
-            _i2cDevice.Write(new[] { (byte)Bmx280Register.CTRL_MEAS, measureState });
+            get => _humiditySampling;
+            set
+            {
+                if (!Enum.IsDefined(typeof(Sampling), value))
+                    throw new ArgumentOutOfRangeException();
+
+                byte status = Read8BitsFromRegister((byte)Bme280Register.CTRL_HUM);
+                status = (byte)(status & 0b_1111_1000);
+                status = (byte)(status | (byte)value);
```
krwq: nit: `value & 0b111`

krwq: actually nvm, you got `Enum.IsDefined(typeof(Sampling), value)`

src/devices/Bmxx80/Bme280.cs
```diff
@@ -73,12 +74,15 @@ public void SetHumiditySampling(Sampling sampling)
         /// <returns>Returns a percentage from 0 to 100.</returns>
         public async Task<double> ReadHumidityAsync()
         {
+            if (HumiditySampling == Sampling.Skipped)
```
krwq: When you provide async method you should as well provide synchronous version of the same function

src/devices/Bmxx80/Bme680.cs
```diff
@@ -51,49 +64,238 @@ public Bme680(I2cDevice i2cDevice)
...
+                throw new ArgumentOutOfRangeException();
 
-            // Get only the power mode bit.
-            var hasNewData = (byte)(read & 0b_1000_0000);
+            // read ambient temperature for resistance calculation
+            var heaterResistance = CalculateHeaterResistance(targetTemperature, (short)ambientTemperature);
+            var heaterDuration = CalculateHeaterDuration(duration);
 
-            return (hasNewData >> 7) == 1;
+            _i2cDevice.Write(new[] { (byte)((byte)Bme680Register.RES_HEAT_0 + profile), heaterResistance });
```
krwq: nit: stackalloc

src/devices/Bmxx80/Bme680.cs
```diff
@@ -51,49 +64,238 @@ public Bme680(I2cDevice i2cDevice)
...
+        /// Gets the required time in ms to perform a measurement. The duration of the gas
+        /// measurement is not considered if <see cref="GasConversionIsEnabled"/> is set to false
+        /// or the chosen <see cref="Bme680HeaterProfile"/> is not configured.
+        /// The precision of this duration is within 1ms of the actual measurement time.
+        /// </summary>
+        /// <param name="profile">The used <see cref="Bme680HeaterProfile"/>. </param>
+        /// <returns></returns>
+        public int GetMeasurementDuration(Bme680HeaterProfile profile)
+        {
+            var osToMeasCycles = new byte[] { 0, 1, 2, 4, 8, 16 };
```
krwq: if this is not being modified you might want to make this static

src/devices/Bmxx80/Bme680.cs
```diff
@@ -102,38 +304,32 @@ public bool ReadHasNewData()
         /// <returns>Calculated humidity.</returns>
         public async Task<double> ReadHumidityAsync()
         {
+            if (HumiditySampling == Sampling.Skipped)
+                return double.NaN;
+
             // Read humidity data.
-            byte msb = Read8BitsFromRegister((byte)Bme680Register.HUMIDITYDATA_MSB);
-            byte lsb = Read8BitsFromRegister((byte)Bme680Register.HUMIDITYDATA_LSB);
+            var msb = Read8BitsFromRegister((byte)Bme680Register.HUMIDITYDATA_MSB);
```
krwq: nit: many devices allow reading more bytes on the register address (so you pass address of whichever address is lower (MSB/LSB) and read 16bits and then you can use https://docs.microsoft.com/en-us/dotnet/api/system.buffers.binary.binaryprimitives?view=netstandard-2.1 if needed.

I'm not sure if this applies here but might be worth checking

src/devices/Bmxx80/Bme680.cs
```diff
@@ -150,17 +346,54 @@ public async Task<double> ReadPressureAsync()
         /// <returns>Calculated temperature.</returns>
         public Task<Temperature> ReadTemperatureAsync()
```
krwq: I have mixed feelings if this should be returning NaN equivalen or perhaps it would be better to do `bool TryReadTemperature(out Temperature temperature)`

src/devices/Bmxx80/Bme680.cs
```diff
@@ -150,17 +346,54 @@ public async Task<double> ReadPressureAsync()
...
+        /// Reads the gas resistance.
+        /// </summary>
+        /// <returns>Gas resistance in Ohm. NaN if last gas measurement is invalid.</returns>
+        public Task<double> ReadGasResistanceAsync()
+        {
+            if (!ReadGasMeasurementIsValid() || !ReadHeaterIsStable())
+                return Task.FromResult(double.NaN);
+
+            // Read 10 bit gas resistance value from registers
+            var g1 = Read8BitsFromRegister((byte)Bme680Register.GAS_RES);
```
krwq: this will very likely allow you to read 16bits directly

src/devices/Bmxx80/Bme680.cs
```diff
@@ -223,6 +456,67 @@ private double CompensatePressure(int adcPressure)
...
+        {
+            var gasMeasValid = Read8BitsFromRegister((byte)Bme680Register.GAS_RANGE);
+            gasMeasValid = (byte)((gasMeasValid & (byte)Bme680Mask.GAS_VALID) >> 5);
+
+            return Convert.ToBoolean(gasMeasValid);
+        }
+
+        private double CalculateGasResistance(ushort adcGasRes, byte gasRange)
+        {
+            var k1Lookup = new[] { 0.0, 0.0, 0.0, 0.0, 0.0, -1.0, 0.0, -0.8, 0.0, 0.0, -0.2, -0.5, 0.0, -1.0, 0.0, 0.0 };
```
krwq: if doesn't change make this static

src/devices/Bmxx80/Bme280.cs
RobinTTY: I would actually change these to synchronous methods and not provide asynchronous ones thinking about it. These methods just have an async signature and no real async behavior. The signature was chosen to keep in line with the Bmx280 implementation.

But the Bme680 actually works a little different compared to the Bmx280 right now:

```C#
...
var measurementDuration = bme680.GetMeasurementDuration(bme680.HeaterProfile);
bme680.SetPowerMode(PowerMode.Forced);          // kick off measurement
Thread.Sleep(measurementDuration);              // wait for measurement
var temp = await bme680.ReadTemperatureAsync(); // this just reads from the register
...
```

compared to:

```C#
...
bmx280.SetPowerMode(Bmx280PowerMode.Forced);          // kick off measurement
var tempValue = await bmx280.ReadTemperatureAsync();  // wait and read after
...
```
The behavior of the bmx280 is actually a little bit suboptimal since the ReadHumidityAsnyc (same for Pressure) works like this:

```C#
if (ReadPowerMode() == Bmx280PowerMode.Forced)
{
    await Task.Delay(GetMeasurementTimeForForcedMode(HumiditySampling));
}

await ReadTemperatureAsync();

byte msb = Read8BitsFromRegister((byte)Bme280Register.HUMIDDATA_MSB);
byte lsb = Read8BitsFromRegister((byte)Bme280Register.HUMIDDATA_LSB);
```

So it always queries the power mode first to check if a measurement is still being taken and then waits for the time the current measurement will take. This breaks down for instance if:

TemperatureSampling = X1, PressureSampling = X1, HumiditySampling = X16

If you call

```C#
...
bmx280.SetPowerMode(Bmx280PowerMode.Forced);          // this will take 7 + 7 + 44ms
var tempValue = await bmx280.ReadTemperatureAsync();  // this will only wait for 7ms
...
```

So this actually doesn't make sense, I just did not think about this until further looking into why this method signature is actually async.
I would change the behavior of the Bmx280 to the one of the Bme680 if you agree and drop the async signatures.

krwq: Yes, I agree async is not particularly useful here

src/devices/Bmxx80/Bmx280Base.cs
RobinTTY: Capture

Actually reading the registers after the measurements are skipped just returns wrong data so in case we want to return the last measurement we would have to save that through a field too.

src/devices/Bmxx80/Bme680.cs
RobinTTY: @joperezr had a similar thought on this, would you prefer returning the last measured result or the TryReadTemperature method?

krwq: TryRead or NaN both sound good. I'll need to have some conversations internally what is preferred (I suspect TryRead will be the guidelines)

RobinTTY:
Addressed the remaining feedback. I changed the Read methods to TryRead with out parameter.

src/devices/Bmxx80/Bme280.cs
```diff
@@ -71,20 +71,28 @@ public Sampling HumiditySampling
...
 
         /// <summary>
-        /// Reads the Humidity from the sensor as %rH.
+        /// Reads the humidity. A return value indicates whether the reading succeeded.
         /// </summary>
-        /// <returns>Returns a percentage from 0 to 100.</returns>
-        public double ReadHumidity()
+        /// <param name="humidity">
+        /// Contains the measured humidity as %rH if the <see cref="HumiditySampling"/> was not set to <see cref="Sampling.Skipped"/>.
+        /// Contains <see cref="double.NaN"/> otherwise.
```
krwq: doc needs update

src/devices/Bmxx80/Bme680.cs
```diff
@@ -307,70 +307,106 @@ public int GetMeasurementDuration(Bme680HeaterProfile profile)
...
+        // TODO: what happens if temp measurement is skipped and humidity is read
         /// <summary>
-        /// Read the humidity.
+        /// Reads the humidity. A return value indicates whether the reading succeeded.
         /// </summary>
-        /// <returns>Calculated humidity.</returns>
-        public double ReadHumidity()
+        /// <param name="humidity">
+        /// Contains the measured humidity as %rH if the <see cref="HumiditySampling"/> was not set to <see cref="Sampling.Skipped"/>.
+        /// Contains <see cref="double.NaN"/> otherwise.
```
krwq: ditto

RobinTTY: Other than the forgotten Todo am I missing something? The out parameter will still contain NaN if the measurement was skipped.

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -48,67 +52,71 @@ protected Bmxx80Base(byte deviceId, I2cDevice i2cDevice)
...
         /// <summary>
-        /// When called, the device is reset using the complete power-on-reset procedure.
+        /// Read the pressure.
         /// </summary>
-        public void Reset()
-        {
-            const byte resetCommand = 0xB6;
-            _i2cDevice.Write(new[] { (byte)Bmxx80Register.RESET, resetCommand });
-        }
+        /// <returns>Calculated pressure in Pa.</returns>
```
joperezr: NIT: doc is not correct. This actually returns a bool which I suppose means success, and has an out parameter which is the one that has the calculated pressure.

RobinTTY: Ah yes good catch

src/devices/Bmxx80/Bme680.cs
krwq: `Contains <see cref="double.NaN"/> otherwise.`

krwq: I see, I'd expect the call to return false when you're skipping but I'm fine either way

RobinTTY: Yeah the call will return false and the out parameter contains NaN if the measurement was skipped

krwq:
@RobinTTY [my PR](https://github.com/dotnet/iot/pull/685) could have caused conflicts here. Feel free to revert any of my docs changes but please rebase your PR and make sure to fix all of the docs issues (previously they might not have been reported)

RobinTTY:
@krwq Added the missing docs.

src/devices/Bmxx80/Bmp280.cs
joperezr: Calling a virtual method from the base constructor should be fine in this case, and if you know that you need to perform some calibration data, you could also just have a ReadCalibrationData abstract method which you call from the base constructor, and define that on the subclasses with what calibration needs o be loaded. In general, the idea would be to bake in the most logic we can on the base class, and only have device specific code in the concrete devices/constructors, and seems like setting the default configuration is something that all devices would need to do.

RobinTTY: Alright changed it so that calibration data and default configuration gets loaded in the base.

RobinTTY:
> This LGTM. @RobinTTY I'm assumming you tested this with the Bmx280 devices to ensure that scenarios still worked there?

Yeah I tested on a Bme680 and Bme280, so it should work on all devices since Bmp280 has no additional logic compared to the Bme280.

joperezr:
One last thing before we merge this @RobinTTY, do we need to update any of the readme.md files for either the samples or the main binding family? I believe that some of it might be out of date with your changes.

RobinTTY:
@joperezr The readme is up to date.

joperezr:
Thanks for all your hard work here @RobinTTY! 🎆

PR #483, opened 2019-06-07
Remove ref assembly from package

joperezr:
cc: @ericstj 

The reference assembly is not required any longer since we now have the not supported assembly in the package.


ericstj:
Do you want to remove `ProduceReferenceAssembly` as well?

joperezr:
I was planning on moving into a project-reference ref-only way of producing the ref assembly in a subsequent PR where I will also incorporate APICompat into the build so I was planning on doing it there. If you rather have me do it here, I can remove that code as well.

PR #486, opened 2019-06-08
Fix DS3231 set RTC date/time bug

maloo:
Device binding is not using user supplied DateTime, instead it reads the RTC time and sets that again.


joperezr:
Great catch, thanks @maloo !

PR #489, opened 2019-06-09
Refactoring RTC Devices & Adding PCF8563 Binding

ZhangGaoxing:
Realtime Clock PCF8563


src/devices/Pcf8563/Pcf8563.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// <summary>
+        /// DS3231 DateTime
+        /// </summary>
+        public DateTime DateTime { get => ReadTime(); set => SetTime(value); }
+
+        /// <summary>
+        /// Creates a new instance of the PCF8563
+        /// </summary>
+        /// <param name="sensor">I2C Device, like UnixI2cDevice or Windows10I2cDevice</param>
+        public Pcf8563(I2cDevice sensor)
```
krwq: for consistency please rename to i2cDevice

src/devices/Pcf8563/Pcf8563.cs
```diff
@@ -0,0 +1,118 @@
...
+            _sensor?.Dispose();
+            _sensor = null;
+        }
+
+        /// <summary>
+        /// BCD To Int
+        /// </summary>
+        /// <param name="bcd">BCD Code</param>
+        /// <returns>int</returns>
+        private int Bcd2Int(byte bcd)
```
krwq: do we have this logic elsewhere? might be worth moving this to `common` directory

src/devices/Pcf8563/README.md
```diff
@@ -0,0 +1,25 @@
...
+The PCF8563 is a CMOS1 Real-Time Clock (RTC) and calendar optimized for low power consumption. A  programmable clock output, interrupt output, and voltage-low detector are also provided. All addresses and data are transferred serially via a two-line bidirectional I2C-bus.
+
+## Sensor Image
+![](sensor.jpg)
+
+## Usage
+```C#
+I2cConnectionSettings settings = new I2cConnectionSettings(1, Pcf8563.DefaultI2cAddress);
+// get I2cDevice (in Linux)
+UnixI2cDevice device = new UnixI2cDevice(settings);
```
krwq: We have `I2cDevice.Create` since today :smile:

src/devices/Pcf8563/samples/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Iot.Device.Pcf8563.Samples
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            I2cConnectionSettings settings = new I2cConnectionSettings(1, Pcf8563.DefaultI2cAddress);
+            // get I2cDevice (in Linux)
+            UnixI2cDevice device = new UnixI2cDevice(settings);
```
krwq: ditto

src/devices/Pcf8563/samples/README.md
```diff
@@ -0,0 +1,37 @@
+# PCF8563 - Samples
+
+## Hardware Required
+* PCF8563
+* Male/Female Jumper Wires
+
+## Circuit
+![](pcf8563_circuit.jpg)
+
+* SCL - SCL
```
krwq: this doesn't seem to be particularly useful except for maybe VCC=5V

src/devices/Pcf8563/Pcf8563.cs
shaggygi: I was thinking the same as it seems common in RTCs.

krwq: optionally join all RTC into one if they are similar enough

krwq:
@ZhangGaoxing you need to remove/replace references to I2c/Spi.Drivers namespace, the subnamespace does not exist anymore, that should fix your build issue. If it won't build locally I recommend to squash your commits and then rebase against latest master

ZhangGaoxing:
@krwq @joperezr

src/devices/Pcf8563/Pcf8563.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Device.I2c;
+
+namespace Iot.Device.Pcf8563
+{
+    /// <summary>
+    /// Realtime Clock PCF8563
+    /// </summary>
+    public class Pcf8563 : IDisposable
+    {
+        private I2cDevice _i2cDevice = null;
```
krwq: initialization is not needed since you overwrite this in the constructor

ZhangGaoxing: @krwq Is that critical? I'm used to giving variables a null value. It's the same code style I've submitted. Should I make a unified revision to this?

joperezr: Not critical, but it is one extra instruction, so I would also remove it 😄

src/devices/Pcf8563/Pcf8563.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        /// <summary>
+        /// DS3231 DateTime
+        /// </summary>
+        public DateTime DateTime { get => ReadTime(); set => SetTime(value); }
+
+        /// <summary>
+        /// Creates a new instance of the PCF8563
+        /// </summary>
+        /// <param name="i2cDevice">The <c>System.Device.I2c.I2cDevice</c></param>
```
joperezr: `<c>System.Device.I2c.I2cDevice</c>` -> `<see cref="System.Device.I2c.I2cDevice"/>`

src/devices/Pcf8563/Pcf8563.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        private I2cDevice _i2cDevice = null;
+
+        /// <summary>
+        /// PCF8563 Default I2C Address
+        /// </summary>
+        public const byte DefaultI2cAddress = 0x51;
+
+        /// <summary>
+        /// DS3231 DateTime
```
joperezr: Looks like copy paste from other binding. Also, could we just rename it to somehting like `The device's <see cref="DateTime"/>`

src/devices/Pcf8563/Pcf8563.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <summary>
+        /// Creates a new instance of the PCF8563
+        /// </summary>
+        /// <param name="i2cDevice">The <c>System.Device.I2c.I2cDevice</c></param>
+        public Pcf8563(I2cDevice i2cDevice)
+        {
+            _i2cDevice = i2cDevice;
+
+            // Set "Normal Mode"
+            Span<byte> ctrl1Config = stackalloc byte[] { (byte)Register.PCF_CTRL1_ADDR, 0x00 };
```
joperezr: Is there a way for this binding to check if the deviceID of the device matches what we expect? Other I2c devices have a device id that we can check in initialization just to make sure that the device we are talking to is indeed a PCF8563

src/devices/Pcf8563/Pcf8563.cs
```diff
@@ -0,0 +1,98 @@
...
+                                NumberTool.Bcd2Dec((byte)(readBuffer[3] & 0b_0011_1111)),
+                                NumberTool.Bcd2Dec((byte)(readBuffer[2] & 0b_0011_1111)),
+                                NumberTool.Bcd2Dec((byte)(readBuffer[1] & 0b_0111_1111)),
+                                NumberTool.Bcd2Dec((byte)(readBuffer[0] & 0b_0111_1111)));
+        }
+
+        /// <summary>
+        /// Cleanup
+        /// </summary>
+        public void Dispose()
```
joperezr: NIT: Do you expect this class to ever be a base class of something else? If so, we would probably want to implement the dispose pattern. If not, then we should just mark this class as sealed.

joperezr:
Left a few comments but this PR looks great overall. Thanks for putting this together @ZhangGaoxing !

src/devices/Pcf8563/Pcf8563.cs
ZhangGaoxing: My fault😅. Over time, I really can't remember how the code was named, such as `ReadTime` or `GetTime`...

src/devices/Pcf8563/Pcf8563.cs
ZhangGaoxing: I looked through the datasheet carefully and found no registers like DeviceID

ZhangGaoxing:
I want to make a `RtcBase` class like this, let RTC devices inherit this class, and move all RTC devices to `RealtimeClock` folder. What do you think? @krwq @joperezr 
```C#
namespace Iot.Device.Rtc
{
    public abstract class RtcBase : IDisposable
    {
        public virtual DateTime DateTime { get => ReadTime(); set => SetTime(value); }
        internal abstract void SetTime(DateTime time);
        internal abstract DateTime ReadTime();
        public abstract void Dispose();
    }
}
```

src/devices/Pcf8563/Pcf8563.cs
joperezr: ok, I guess that should be fine then. I wonder if cases like this we should add a try catch on the first read/write and send a helpful error message in case of having the wrong device, or perhaps connecting it to the board incorrectly. You don't have to add this in your PR, this is just food for thought :)

joperezr:
> I want to make a RtcBase class like this, let RTC devices inherit this class, and move all RTC devices to RealtimeClock folder. 

Sounds like a great plan to me. the only things I would change is one to make the Set and ReadTime protected, instead of internal, so that people can add a subclass in their own application or in a different assembly, and I would also properly implement the dispose pattern like:
```c#
namespace Iot.Device.Rtc
{
    public abstract class RtcBase : IDisposable
    {
        public virtual DateTime DateTime { get => ReadTime(); set => SetTime(value); }
        protected abstract void SetTime(DateTime time);
        protected abstract DateTime ReadTime();
        public void Dispose()
        {
          Dispose(true);
          GC.SupressFinalize(this);
        }
        {
           //..
        }
    }
}
```

Do you plan to do this on this PR?

ZhangGaoxing:
@joperezr @krwq

src/devices/Common/Iot/Device/NumberTool.cs
```diff
@@ -0,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device
+{
+    /// <summary>
+    /// Helpers for number.
+    /// </summary>
+    internal static class NumberTool
```
krwq: optional: I'd prefer if we called it something like NumberHelpers or ConversionHelpers or something along these lines but since it's internal I'm fine with this as is. I guess the "Tool" is the most confusing part of the name for me since tool sounds like a standalone application to me

src/devices/RealtimeClock/Devices/Ds1307/Ds1307.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            if (!_disposedValue)
+            {
+                if (disposing)
+                {
+                    _i2cDevice?.Dispose();  
+                }
+
+                _i2cDevice = null;
+                _disposedValue = true;
```
krwq: I don't think `_disposedValue` is needed since it's value is equal exactly `_i2cDevice == null`

src/devices/RealtimeClock/Devices/Ds1307/Ds1307Register.cs
```diff
@@ -2,12 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-namespace Iot.Device.Ds1307
+namespace Iot.Device.Rtc
```
krwq: I like this. We should do similar stuff for Mcp3008 where Mcp3008 is namespace and class name which causes it to be really annoying to use as you always have to type both (not in this PR)

src/devices/RealtimeClock/Devices/Ds3231/Ds3231.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// </summary>
+    public class Ds3231 : RtcBase
+    {
+        /// <summary>
+        /// DS3231 Default I2C Address
+        /// </summary>
+        public const byte DefaultI2cAddress = 0x68;
+
+        private I2cDevice _i2cDevice;
+        private bool _disposedValue = false;
```
krwq: ditto

src/devices/RealtimeClock/Devices/Pcf8563/Pcf8563.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <returns>PCF8563 Time</returns>
+        protected override DateTime ReadTime()
+        {
+            // Sec, Min, Hour, Date, Day, Month & Century, Year
+            Span<byte> readBuffer = stackalloc byte[7];
+
+            _i2cDevice.WriteByte((byte)Pcf8563Register.PCF_SEC_ADDR);
+            _i2cDevice.Read(readBuffer);
+
+            return new DateTime(readBuffer[5] >> 7 == 1 ? 2000 + NumberTool.Bcd2Dec(readBuffer[6]) : 1900 + NumberTool.Bcd2Dec(readBuffer[6]),
```
krwq: nit: this seems equivalent to `1900 + (readBuffer[5] >> 7) * 100 + NumberTool.Bcd2Dec(readBuffer[6])`

src/devices/Common/Iot/Device/NumberTool.cs
```diff
@@ -0,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device
```
joperezr: Can we move this into an additional namespace? Perhaps Iot.Device.Helpers or Iot.Device.Common? Currently Iot.Device doesn't contain any classes, only other namespaces, and I would like to keep it like that if possible.

src/devices/RealtimeClock/RealtimeClock.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
```
joperezr: We recently changed this so that new .csprojs use:

```xml
    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
      <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
    </ProjectReference>
```

Can you please update this project, and your sample projects as well?

src/devices/RealtimeClock/RtcBase.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.Rtc
```
joperezr: Can we rename this to RealtimeClock? It will make it easier for people to find where the namespace is defined on the repo if the namespace matches the folder and project it is located in.

src/devices/RealtimeClock/Devices/Ds1307/Ds1307.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            _i2cDevice.Write(writeBuffer);
+        }
+
+        /// <summary>
+        /// Cleanup
+        /// </summary>
+        protected override void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
```
joperezr: This is not needed. given that you have `_i2cDevice?.Dispose()` it will only call that in case the device hasn't been disposed as it is assigned to null in the next line.

src/devices/RealtimeClock/Devices/Ds1307/Ds1307.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                if (disposing)
+                {
+                    _i2cDevice?.Dispose();  
+                }
+
+                _i2cDevice = null;
+                _disposedValue = true;
+            }
```
joperezr: We are also missing here the call to the base like `base.Dispose(true);`

src/devices/RealtimeClock/Devices/Ds3231/Ds3231.cs
```diff
@@ -0,0 +1,119 @@
...
+            // datasheet Temperature part
+            return data[0] + (data[1] >> 6) * 0.25;
+        }
+
+        /// <summary>
+        /// Cleanup
+        /// </summary>
+        protected override void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
```
joperezr: Same disposing comment

src/devices/RealtimeClock/RtcBase.cs
krwq: I actually think the current namespace makes sense, I'd rather rename the folder to RTC - RTC is really well known abbreviation in hardware so doesn't make sense to expand it

ZhangGaoxing: rename to Rtc

src/devices/Common/Iot/Device/NumberTool.cs
ZhangGaoxing: change it to `Iot.Device.Common`, because the class in the `Common` folder and I don't know if `Iot.Device.Helpers` should be `Helper` or `Helpers`😂

src/devices/RealtimeClock/RealtimeClock.csproj
ZhangGaoxing: Got it, but I hate this decision. Every time I need to put all `.csproj` in one `.sln`. Maybe there is something wrong with my VS settings.😅

joperezr: Oh that is because you have package restore enabled in your VS. Go to `Tools -> Options`  and then to `NuGet Package Manager -> General` and make sure that `Allow NuGet to download missing packages` and `Automatically check for missing packages during build in Visual Studio` are both **unchecked**. After doing this, all you will need for setup is to have at least run `dotnet restore` on the project once before opening VS (which we do when you call build.cmd) and then everything should just work fine without the need of pulling stuff into your solution.

PR #497, opened 2019-06-12
Add Windows CI legs

joperezr:
cc: @safern 

This will add Windows runs into our CI and official builds

FYI: @ulisesh @joshfree

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -54,6 +54,18 @@ jobs:
       name: Build
       displayName: Build
       condition: succeeded()
+    - script: powershell -ExecutionPolicy ByPass -NoProfile eng\common\msbuild.ps1 -warnaserror:0 -ci
+        eng/sendToHelix.proj
+        /t:Test
+        /p:TestOS=Windows_NT
+        /p:Configuration=$(_BuildConfig)
+        /p:HelixBuild=$(Build.BuildNumber)
+        /bl:$(Build.SourcesDirectory)/artifacts/log/$(_BuildConfig)/SendToHelix.binlog
+        /p:HelixAccessToken=$(HelixApiAccessToken)
+      displayName: Send to Helix
+      condition: succeeded()
+      env:
+        SYSTEM_ACCESSTOKEN: $(System.AccessToken)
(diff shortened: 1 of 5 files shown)


.vsts-ci.yml
```diff
@@ -54,6 +54,18 @@ jobs:
...
       displayName: Build
       condition: succeeded()
+    - script: powershell -ExecutionPolicy ByPass -NoProfile eng\common\msbuild.ps1 -warnaserror:0 -ci
+        eng/sendToHelix.proj
+        /t:Test
+        /p:TestOS=Windows_NT
+        /p:Configuration=$(_BuildConfig)
+        /p:HelixBuild=$(Build.BuildNumber)
+        /bl:$(Build.SourcesDirectory)/artifacts/log/$(_BuildConfig)/SendToHelix.binlog
+        /p:HelixAccessToken=$(HelixApiAccessToken)
```
safern: Will `HelixApiAccessToken` always be defined? What happens in PR builds were the variable group is not available?

eng/sendToHelix.proj
```diff
@@ -16,36 +21,54 @@
     <FailOnMissionControlTestFailure>true</FailOnMissionControlTestFailure>
   </PropertyGroup>
 
-  <ItemGroup>
-    <!-- For now we only have a queue for Linux. Once we have a queue for Windows we will run 
-    those tests too. -->
+  <PropertyGroup Condition=" '$(HelixAccessToken)' == '' ">
+    <IsExternal>true</IsExternal>
```
safern: I think `IsExternal` is no longer needed. I think the Helix SDK will set it to true if `HelixAccessToken` is empty

.vsts-ci.yml
joperezr: It is evaluated as empty. Which basically means that it won't be set.

safern: Ok, cause I thought when a variable wasn't defined, azure devops would replace it with the literal `$(...)` value.

eng/sendToHelix.proj
joperezr: Ok, I'll check and if so I'll remove. Thanks

Issue #347, opened 2019-03-27
Disposing pins is not always expected


krwq:
I think we should have a way of telling controller that we don't want any cleanup to happen for the specific pin (in my case `reset` pin)

Here is the scenario I've encountered recently:
I've worked on oLED bindings - one of the oLED pins is `reset`.
Scenario is following:
- start .NET app to draw something and close it

Here is what happens instead:
- .NET app starts, draws whatever user requested
- Dispose is called
- Reset PIN is set to Low
- oLED is reset
- .NET app finishes
- oLED is black

current solution for this problem was not to call Dispose at all (uses the fact that .NET Core doesn't call dispose on cleanup).

joperezr:
Interesting scenario. I kind of feel that if we wanted to do this we might leak some files for specific set of drivers. For example, the SysFS driver would need to leave some files behind in order to persist the value. I'm not 100% sure  on this, but I believe that is the case.

shaggygi:
While I'm just speculating with the work on `I2cController`, we could eventual evolve the IGpioController to manage GpioPins (or IGpioPins) instead.  This way, you could possibly have a `ShouldDispose` property like we might be adding to `I2cDevice`.

joperezr:
I don't think we want to optimize for the scenario you specified above with the OLED displays. As for the disposing pattern, we have that covered in another issue so I will close this as a dupe as the fix will be tracked by that other one

joperezr:
duplicate of #419

Issue #204, opened 2019-02-03
PwmController/PwmDevice should be bound to specific pin


krwq:
Currently the  PwmController class requires passing channel and chip every time.

There are couple of disadvantages of such approach:
- APIs are inconvenient and require user to store pin value and pass it on every call
- PWM settings along with pin have to be dragged along to all depending classes which also need to access PWM settings
- it is not known if channel is open or not
- other PWM implementation (i.e. SoftwarePwm does not use channel and chip and uses pin instead)

my proposition is to have simpler design which looks more or less like this:

```csharp
public abstract PwmDevice : IDisposable
{
  protected PwmDevice() {}
  public abstract double FrequencyHz { get; } // note: can't change after setting up
  public abstract double DutyCyclePercentage { get; set; } // 0.0 - 1.0
  public static PwmDevice Create(int chip, int channel, double frequencyHz = 50.0, double dutyCyclePercentage = 0.0) { /* create default pwm device depending on the OS, default values are arbitrary and TBD */ }
}
```

Notes:
- change range of DutyCycle from 0 - 100 to 0-100% because % in math equals to 0.01 and 100%=1.0. Also it combines with any other math done in the app much better. I.e.: `dutyCycle = a/b` and not `dutyCycle = a/b*100`
- channel will be open when constructed by default, when dutyCycle is 0 it technically does nothing anyway

shaggygi:
If this is implemented, my vote would be to create IPwmPin/PwmPin similar to what is [suggested for IGpioPin/GpioPin](https://github.com/dotnet/iot/issues/215).

shaggygi:
I was thinking about this a little while we were working on I2C/SPI and wanted to see your thoughts before attempting to prototype.

Ultimately the plan would be to have a final structure below.  I understand this includes big breaking changes... but better now than never.

```console
Pwm
  Channels                        
    PwmChannel.cs                      // All classes have namespace of 'namespace System.Device.Pwm'
    UnixPwmChannel.Linux.cs            // Internal
    Windows10PwmChannel.Windows.cs     // Internal
    Windows10PwmChip.Windows.cs        // Internal
  IPwmController.cs
  PwmController.cs
```

### Toughts

- You'd have the option of using a `PwmChannel` or `PwmController`.

#### PwmChannel

Similar to what you mentioned above, but a `PwmChannel`.

```csharp
public abstract PwmChannel : IDisposable
{
  protected PwmChannel() {}
  public abstract double FrequencyHz { get; }
[...]

PR #505, opened 2019-06-16
Added new I2cController and related support

### LINKED ISSUE #118: [Proposal] I2cController and II2cController
Similar to the ```GpioController```, there needs to be a mechanism to determine the best ```I2cDevice``` to use.

#### Updated 6/18/2019

`I2cDevice` allows for an easy way to create an I2C connection between a master controller and one device binding.  However, there are scenarios where a new type of **controller** can manage related devices more efficiently.

# Rationale and Usage

**System.Device.Gpio Consistency** - Other APIs provide controllers that work with low-level drivers to perform the functional operations.  For example, `GpioController` informs related driver to write logic-levels to pins and `PwmController` informs related driver to write variable frequency to pins.  The `I2cController` will inform which I2C device to read/write data.

**Multiple Device Management** - Other controllers also can open and close multiple pins that will be controlled.  Sometimes hardware configurations require a master to communicate with multiple devices on the same I2C bus.  This is a current limitation with `I2cDevice` as it requires the bus ID & device address when instantiating and the read/write methods communicate with only the one device.

**Default OS Implementation** - Most current device binding samples specifically define the `I2cDevice` related to executing OS (e.g. `UnixI2cDevice` or `Windows10I2cDevice`).  This forces users to update sample code when running on different hardware and OS.  This proposal will offer a method to generate the default `I2cDevice` for OS similar to how a `GpioController` provides a default `GpioDriver`.

**Controller Interface** - In addition to providing a default `I2cController`, this proposal also includes an `II2cController` as there are use-cases where bindings could be a master I2C controller that can control other I2C devices.

# Proposed Changes

## 1. Add New II2cController and I2cController

Add a new `II2cController` interface that includes the same read/write methods as an `I2cDevice` with the addition of specified bus ID and address parameters.  This allows the controller to know which device to use for communications.  There are also methods to open/close devices.

```csharp
public interface II2cController : IDisposable
{
    void OpenDevice(I2cConnectionSettings settings, bool shouldDispose = true);
    void OpenDevice(I2cDevice device, bool shouldDispose = true);
    void CloseDevice(I2cConnectionSettings settings);
    void CloseDevice(I2cDevice device);
    void ClearDevices();
    bool TryGetDevice(int busId, int address, out I2cDevice device);
    void Read(int busId, int address, Span<byte> buffer);
    byte ReadByte(int busId, int address);
    void Write(int busId, int address, ReadOnlySpan<byte> buffer);
    void WriteByte(int busId, int address, byte value);
    void WriteRead(int busId, int address, ReadOnlySpan<byte> writeBuffer, Span<byte> readBuffer);
}
```

The `I2cController` class will implement the new `II2cController` interface and be the default for most apps and device bindings.  The controller is instantiated with zero I2C devices to control.  More devices can be opened/closed similar to pins on other controllers.

```csharp
// Class
public class I2cController : II2cController

// Constructor
public I2cController()
```

The creation of devices with a new static helper method that attempts to pick the best I2C device for OS app is executing on.  This static method will be added to a new `sealed partial class` respective to OS.

```csharp
// I2cController.Linux.cs
public static I2cDevice Create(I2cConnectionSettings settings) => new UnixI2cDevice(settings);
// I2cController.Windows.cs
public static I2cDevice Create(I2cConnectionSettings settings) => new Windows10I2cDevice(settings);
```
    
### Example 1.1
Instantiate a controller and open a device.

```csharp
int busId = 1;
int address = 0x21;
var settings = new I2cConnectionSettings(busId, address);
var i2cController = new I2cController();
i2cController.OpenDevice(settings);
```

### Example 1.2
Open a new device to controller that already contains device with specified address.

```csharp
int busId = 1;
int address = 0x21;
var settings = new I2cConnectionSettings(busId, address);
var i2cController = new I2cController();
i2cController.OpenDevice(settings);
i2cController.OpenDevice(settings); // Throws exception because device with address already exists.
```

### Example 1.3
Open two devices with same address but on different buses.

```csharp
int busId0 = 0;
int busId1 = 1;
int address = 0x21;
var settings0 = new I2cConnectionSettings(busId0, address);
var settings1 = new I2cConnectionSettings(busId1, address);
var i2cController = new I2cController();
i2cController.OpenDevice(settings0);
i2cController.OpenDevice(settings1);
```

### Example 1.4
Open a new device that implements `I2cDevice` to controller.

**NOTE**: This I2C software implementation doesn't currently exist but is the same concept as the SPI software implementation.  See [[Proposal] Provide a Software implementation of the SPI protocol](https://github.com/dotnet/iot/issues/122) for related details.

```csharp
int busId = 1;
int address = 0x21;
var settings = new I2cConnectionSettings(busId, address);
int sdaPin = 17;
int sclPin = 18;
var gpioController = new GpioController();

var i2cDevice = new GpioI2cDevice(
    settings,
    sdaPin,
    sclPin,
    gpioController);

var i2cController = new I2cController();
i2cController.OpenDevice(i2cDevice);
```

### Example 1.5
Close a device from controller that doesn't contain device with specified address.

```csharp
int busId = 1;
int address = 0x21;
var settings = new I2cConnectionSettings(busId, address);
var i2cController = new I2cController();

i2cController.OpenDevice(settings);
i2cController.CloseDevice(new I2cConnectionSettings(busId, 0x45));  // Throws exception because no device with specified address exists.
```

### Example 1.6
Reading from a specific device in controller.

```csharp
int busId = 1;
int address = 0x21;
var settings = new I2cConnectionSettings(busId, address);
var i2cController = new I2cController();
i2cController.OpenDevice(settings);

byte value = i2cController.ReadByte(busId, address);

// Or...

Span<byte> buffer = stackalloc byte[3];
i2cController.Read(busId, address, buffer);
```

### Example 1.7
Write to a specific device in controller.

```csharp
int busId = 1;
int address = 0x21;
var settings = new I2cConnectionSettings(busId, address);
var i2cController = new I2cController();
i2cController.OpenDevice(settings);

byte value = 0x23;
i2cController.WriteByte(busId, address, value);

// Or...

ReadOnlySpan<byte> buffer = new byte[] { 0x12, 0x23, 0x45 };
i2cController.Write(busId, address, buffer);
```

## 2. Disposing of I2cControllers and I2cDevices

- [ ] WIP: Need to review topic and create examples here.  There are some initial dispose notes below (delete after adding to this section).

**NOTE**: This dispose pattern needs to fit other controllers to not confuse others, as well.

A new `public bool ShouldDispose` property is added to `I2cDevice` class for different disposing scenarios.  This property will allow controllers/bindings to know if it should dispose device(s) if/when the controller/binding is being disposed.  By default, this property should be set to `true` as this is the intention in most cases.

```csharp
public abstract class I2cDevice : IDisposable
{
    public bool ShouldDispose { get; set; } => true;

    // other current code...
}
```

### Example 2.1
Use controller with default device (ShouldDispose = true).

### Example 2.2
Use controller with provided device (ShouldDispose = false).

## 3. Device Binding Guidance
This section points out a few areas bindings should use that use `I2cDevice`.

### Binding Constructors

Bindings that use I2C should be the owners of opening the `I2cDevice` on the controller.  Therefore, the bindings should provide at least 2 constructor overloads that offer the option of using `I2cConnectionSettings` or `I2cDevice`.  This will make it easier when instantiating the bindings with other apps and APIs.  It is assumed there will be no other I2C device with the same bus ID and address as it will throw an exception when the binding adds the device to the controller.

### Example 3.1
Instantiate a device binding with specified settings without a defined controller.

**NOTE**: The default controller would be instantiated similar to a master `IGpioController` when not defined.  Since the controller is not provided and only settings, there is no way to add a specified `I2cDevice`.  Therefore, the intent is to also create a default `I2cDevice` when it is added to controller.

```csharp
int busId = 1;
int address = 0x21;
var settings = new I2cConnectionSettings(busId, address);
var mcp23018 = new Mcp23018(settings);  // Default controller and default device is opened within.
```

### Example 3.2
Instantiate a device binding with specified device without a defined controller.

**NOTE**: The default controller would be instantiated similar to a master `IGpioController` when not defined.  The specified device will then be opened.

```csharp
int busId = 1;
int address = 0x21;
var settings = new I2cConnectionSettings(busId, address);
MyI2cDevice device = new MyI2cDevice(settings);  // New device that implements I2cDevice.
var mcp23018 = new Mcp23018(myI2cDevice);  // Default controller and specified device is opened within.
```

### Example 3.3
Instantiate a device binding with specified settings (or device) and specified controller.

```csharp
int busId = 1;
int address = 0x20;
var i2cController = new WhizBangI2cController();

// By settings.
var settings = new I2cConnectionSettings(busId, address);
var mcp23018 = new Mcp23018(settings, i2cController);

// Or...

// By device.
var device = I2cController.Create(new I2cConnectionSettings(busId, address));
i2cController.OpenDevice(device, i2cController);
```

### Example 3.4
Instantiate a controller for a binding that has multiple devices with fixed addresses.

See [Adding GHI Fez Cream driver](https://github.com/dotnet/iot/issues/379) for related details.

**NOTE**: This hardware is a RPi HAT with hard-wired pins on I2C bus 1.

```csharp
var ghiFezCream = new GhiFezCream(); // Adds all 4 on-board devices to binding.

// Or...

// Adds all 4 on-board devices to binding using default devices on specified controller.
var i2cController = new I2cController();
var ghiFezCream = new GhiFezCream(i2cController);

// Or...

// Adds all 4 on-board devices to binding using specified devices and specified controller.
int busId = 1;

var ghiFezCream = new GhiFezCream(
    new GpioI2cDevice(new I2cConnectionSettings(busId, GhiFezCream.Pca9535AddressIc1)),
    new GpioI2cDevice(new I2cConnectionSettings(busId, GhiFezCream.Pca9535AddressIc2)),
    new GpioI2cDevice(new I2cConnectionSettings(busId, GhiFezCream.Ads7830IpwtAddress)),
    new GpioI2cDevice(new I2cConnectionSettings(busId, GhiFezCream.Pca9685Address)),
    i2cController);

public class GhiFezCream
{
    // Constructor shows it adding fixed devices.
    public GhiFezCream(II2cController? i2cController = null)
    {
        int busId = 1;
        i2cController = i2cController ?? new I2cController();

        i2cController.OpenDevice(new I2cConnectionSettings(busId, GhiFezCream.Pca9685Address)),
        i2cController.OpenDevice(new I2cConnectionSettings(busId, GhiFezCream.Pca9535AddressIc2)),
        i2cController.OpenDevice(new I2cConnectionSettings(busId, GhiFezCream.Ads7830IpwtAddress)),
        i2cController.OpenDevice(new I2cConnectionSettings(busId, GhiFezCream.Pca9685Address)),
    }
    
    // Or you could overload with specified devices.
    public GhiFezCream(
        I2cDevice pca9535Ic1,
        I2cDevice pca9535Ic2 ,
        I2cDevice ads7830Ipwt ,
        I2cDevice pca9685 ,
        II2cController? i2cController = null)
    {
        i2cController = i2cController ?? new I2cController();
        i2cController.OpenDevice(pca9535Ic1);
        i2cController.OpenDevice(pca9535Ic2);
        i2cController.OpenDevice(ads7830Ipwt);
        i2cController.OpenDevice(pca9685);
    }

    public static int Pca9535AddressIc1 => 0x23;
    public static int Pca9535AddressIc2 => 0x25;
    public static int Ads7830IpwtAddress => 0x48;
    public static int Pca9685Address => 0x70;
}
```

### Example 3.5
Instantiate a controller, opens another I2C device on bus, and then opens a binding.

```csharp
int busId = 1;
int address = 0x21;
var settings = new I2cConnectionSettings(busId, address);
var i2cController = new I2cController();
i2cController.OpenDevice(settings);  // This opens a default device using specified settings.

var bindingSettings = new I2cConnectionSettings(busId, 0x20);
var mcp23018 = new Mcp23018(bindingSettings, i2cController);  // Specified controller opens default device within on same bus as other device.
```

### Disposing of Bindings

Bindings should implement `IDisposable` where it disposes the related devices.

```csharp
public class MyBinding : IDisposable
{
    public void Dispose()
    {
        // This will remove the device from controller and only dispose if ShouldDispose is true. 
        _i2cController.CloseDevice(_i2cSettings);
    }
}
```

# Update I2C Device Namespace

:warning: This topic is related to a future PR that will be a breaking change.

There have been discussions how we should structure/name some of the folders and classes throughout API.  I originally thought we should rename I2C devices to use 'Driver' instead to be consistent with other types.  For example, `I2cDevice` would be renamed `I2cDriver`.

My opinions have changed as the new I2C controller fits quite nicely with the current device naming.  Although, there is a breaking change needed with the folder structure and related namespace.  This could be implemented before the release of v1 in a future PR.   So this proposal is in-line to the current device class naming.  It only states we need to change the namespace later (hopefully soon after).

```csharp
// From...
namespace System.Device.I2c.Drivers
// To...
namespace System.Device.I2c.Devices
```

See [[Proposal] Move classes and rename to be consistent throughout API](https://github.com/dotnet/iot/issues/126) for related details.

# Related Issues

- [I2C APIs should take address on Read/Write](https://github.com/dotnet/iot/issues/214)
- [Adding GHI Fez Cream driver](https://github.com/dotnet/iot/issues/379)
- [[Proposal] Provide controller provider interfaces](https://github.com/dotnet/iot/issues/125)
- [[Proposal] Move classes and rename to be consistent throughout API](https://github.com/dotnet/iot/issues/126)
- [SenseHat does not create appropriate I2C devices](https://github.com/dotnet/iot/issues/498)
- [Reconsider Dispose pattern across devices](https://github.com/dotnet/iot/issues/419)

shaggygi:
Resolves #118 

This adds new Create method to return `I2cDevice` based on OS app is executing.


joperezr:
Thanks for putting this up @shaggygi. I left some comments on the proposal so lets talk about those first and once we are in agreement I'll go ahead and review this. From just a very quick look, some feedback would be that you don't really need the Interop file you added to check the current OS, since we currently build two System.Device.Gpio.dll implementations, one for Linux and one for Windows. When we are building for Linux, we will compile all files that don't end like `*.Windows.cs` and similar for windows, we will build everything but the `*.Linux.cs`. This means that you can add all the shared logic for the controller in a file called `I2cController.cs` and then simply add the Open methods that will create the actual I2cDevices on separate files like `I2cController.Linux.cs`. You can see an example of this in our GpioController : https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs

shaggygi:
@joperezr as you probably noticed, this is breaking now as I'm using some of the new C# 8 nullable stuff.  Would it be possible to add the <Nullable>enable</Nullable> to the main project?

Not sure what that would cause, but we might need to begin preparing the APIs based on that to get ahead of the game?

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        /// <summary>
+        /// Opens an I2C device in order for it to be ready to use with the controller.
+        /// </summary>
+        /// <param name="device">The I2C device to open.</param>
+        /// <param name="shouldDispose">Determines if the device should be disposed by a controller or binding.</param>
+        public void OpenDevice(I2cDevice device, bool shouldDispose = true)
+        {
+            // TODO: Should this compare type with other devices added to makes sure they are all the same (at least on same bus).
```
shaggygi: Need to complete this.

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
```diff
@@ -0,0 +1,206 @@
...
+            if (device.ShouldDispose)
+            {
+                device.Dispose();
+            }
+        }
+
+        /// <summary>
+        /// Removes all I2C devices from controller.
+        /// </summary>
+        public void ClearDevices() => _openDevices.Clear();
```
shaggygi: Need to also dispose device(s) that have `ShouldDispose`.

src/System.Device.Gpio/System/Device/I2c/II2cController.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// Determines if this should be disposed.
+        /// </summary>
+        bool ShouldDispose { get; set; }
+
+        /// <summary>
+        /// Opens an I2C device with specified settings in order for it to be ready to use with the controller.
+        /// </summary>
+        /// <param name="settings">The connection settings of a device on an I2C bus.</param>
+        /// <param name="shouldDispose">Determines if this should be disposed.</param>
+        void OpenDevice(I2cConnectionSettings settings, bool shouldDispose = true);
```
shaggygi: Is this optional parameter supposed to be set in the interface?  Was not sure if that is the norm.

joperezr:
> as you probably noticed, this is breaking now as I'm using some of the new C# 8 nullable stuff. Would it be possible to add the Nullable = enable to the main project?

Right, enabling this project wide will generate some warnings as today we are setting some values to null without annotating, so I wouldn't set this yet project-wide. For the mean time, what you can do is to add `#nullable enable` on the top of the .cs files where you are adding annotations so that the compiler would not complain and annotations will still be taken into account. Usually the pattern followed for annotating nulls, one would go file-by-file enabling nullable with `#nullable enable` and once all files are anotated, then you can go and remove all of those and enable this project wide by setting the following property on the csproj: `<NullableContextOptions>enable</NullableContextOptions>`

src/System.Device.Gpio/System/Device/I2c/II2cController.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// Gets the I2C device associated with the specified bus ID and address.
+        /// </summary>
+        /// <param name="busId">The bus ID the I2C device is connected to.</param>
+        /// <param name="address">The bus address of the I2C device to write.</param>
+        /// <param name="device">
+        /// When this method returns, contains the I2C device associated with the specified bus ID and address, if they are found;
+        /// otherwise, the default value for the type of the value parameter. This parameter is passed uninitialized.
+        /// </param>
+        bool TryGetDevice(int busId, int address, out I2cDevice? device);
+
```
shaggygi: Should we also have a GetDevices() method that returns all opened devices?

joperezr: we could do that, and if so we should probably implement IEnumerable as well so that it may be used in a foreach. I'm fine with either that, or the GetDevices() method returning an IEnumerable<I2cDevice>

shaggygi: Let's go with GetDevices().

shaggygi:
@joperezr @krwq my brain hurts 🤕 😄 

I made a few more changes and removed WIP from PR title.  Please give it a look.  Be gentle, but I know there are things missed and improvements to be made.

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            return device;
+        }
+
+        /// <summary>
+        /// Opens an I2C device with specified settings in order for it to be ready to use with the controller.
+        /// </summary>
+        /// <param name="settings">The connection settings of a device on an I2C bus.</param>
+        /// <param name="shouldDispose">Determines if the device should be disposed by a controller or binding.</param>
+        public void OpenDevice(I2cConnectionSettings settings, bool shouldDispose = true)
```
krwq: there is something confusing about this name/API. Usually `Open` methods return whatever they try to open and never dispose it.

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
```diff
@@ -0,0 +1,218 @@
...
+            if (device.ShouldDispose)
+            {
+                device.Dispose();
+            }
+        }
+
+        /// <summary>
+        /// Removes all I2C devices.
+        /// </summary>
+        public void ClearDevices() => Dispose(true);
```
krwq: This seems redundant considering it does the same what Dispose

krwq:
@shaggygi I like the fact that you had a time to tackle on this issue but it feels to me like this API could be collapsed to just:

```csharp
namespace System.Device.I2c
{
    public static class I2cController
    {
        public static I2cDevice Create(I2cConnectionSettings settings);
    }
}
```

Or just directly:

```csharp
public class I2cDevice
{
  public static I2cDevice Create(I2cConnectionSettings settings);
}
```

I don't find much value in having the controller representation. Do you have any use cases where this would be useful?

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
```diff
@@ -0,0 +1,218 @@
...
+// See the LICENSE file in the project root for more information.
+
+#nullable enable
+using System.Collections.Generic;
+
+namespace System.Device.I2c
+{
+    public sealed partial class I2cController : II2cController
+    {
+        private readonly Dictionary<(int, int), I2cDevice> _openDevices;
```
krwq: this probably should be ConcurrentDictionary or locked everywhere

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
shaggygi: I was trying to use similar naming between controllers.  See [OpenPin for GpioController](https://github.com/dotnet/iot/blob/33432162e8b7dbc64f457571a9be43767033df51/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs#L63) where it doesn't return anything.

The ShouldDispose is to allow controllers or bindings to know if they should dispose of device when they are being removed.

Something for future thought (and understanding it might not be feasible)...  We could eventually modify the GpioController to manage GpioPin (or GpioDevice) and update  to something like...

```csharp
public void OpenPin(int pin, bool shouldDispose = true)
// Or...
public void OpenPin(GpioConnectionSettings settings, bool shouldDispose = true)
```

Where this could create default GpioPin and offer a flag to dispose of pin when done with controller or binding.  This might address your issue #347.  But I'm not completely sure.

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
shaggygi: Added change for this.

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
krwq: the main difference is that pin currently does not have a representing object vs I2cDevice does.

I'm personally not in favor of keeping track of all open I2c/Spi devices just to add few helper functions for edge case scenarios

shaggygi: Agreed, but would be an object if GPIO was expanded upon.  Same thing for possibly changing PwmController.  Currently, we manage chips/channels.  We could expand and manage PwmDevices as you mentioned in #204.

```csharp
public void OpenChannel(int chip, int channel, bool shouldDispose = true)
// Or...
public void OpenChannel(PwmConnectionSettings settings, bool shouldDispose = true)
```

krwq: I think I'm leaning toward changing every `Open*` method to return something which can be disposed and remove all `Close*` functions. I think with the current path we will get ourselves into really annoying state. I'd leave all helper functions as they are (WritePin etc).

In general I'd prefer if we did not track lifetime of objects in their parents at all but shouldDispose is acceptable to me. cc: @joperezr 

Initially I was really pushing back on having any objects for pins since it seemed heavyweight to me but I think I don't feel that way anymore consdering all special cases we've hit (i.e. #347 and weird abstraction for PWM).

If we're gonna go this path we should change it before we ship 3.0. I think we should make the changes mostly source compatible (so we can consider leaving the Close methods although I'd prefer if we've removed them)

shaggygi:
@krwq 

> I don't find much value in having the controller representation. Do you have any use cases where this would be useful?

Maybe you're right and it is being over-engineered.  I thought about it some more today while reviewing the [TCA9548 binding](https://github.com/dotnet/iot/issues/508):

The TCA9548 doesn't work how first thought and would be a little more complex to code or would require unnecessary writes for reading/writing selected I2C devices.  So let's step back and try a scenerio.

```csharp
// Create an I2C device used with Tca9548.
I2cDevice tca9548I2cDevice = I2cDevice.Create(new I2cConnectionSettings(1, 0x40));   // Will dispose when disposing Tca9548.
Tca9548 tca9548 = new Tca9548(tca9548I2cDevice, ...);

// Create another I2C device used with Mcp23008.
I2cDevice mcp23008I2cDevice = I2cDevice.Create(new I2cConnectionSettings(1, 0x20));  // Will dispose when disposing Mcp23008.
Mcp23008 mcp23008 = new Mcp23008(mcp23008I2cDevice, ...);

// Select the Mcp23008 that is hard-wired on channel 3 of Tca9548.
tca9548.Select(Channel.Three);

// At this point, the Mcp23008 should look like it is connected on the same bus as the Tca9548.
// So doing the GPIO commands are still using the default I2C work without having to interact with an I2C controller.
mcp23008.WritePin(1, PinValue.High);

```

I like the `ShouldDispose` property so you have option to use the same I2C device for multiple devices on the buses connected to the Tca9548.  For example, you could have a Mcp23008 connected to channels 1-3 with the same settings.  This would allow one of the Mcp23008 to be disposed, but allow the shared I2C device to continue working on the other 2 Mcp23008 devices.

```csharp
I2cDevice mcp23008I2cDevice = I2cDevice.Create(new I2cConnectionSettings(1, 0x20) { ShouldDispose = false } );  // Don't dispose when disposing of any Mcp23008.
Mcp23008 mcp23008_1 = new Mcp23008(mcp23008I2cDevice, ...);
Mcp23008 mcp23008_2 = new Mcp23008(mcp23008I2cDevice, ...);
Mcp23008 mcp23008_3 = new Mcp23008(mcp23008I2cDevice, ...);

tca9548.Select(Channel.One);
// Do work with mcp23008_1.
mcp23008_1.Dispose();  // Clean up other resources like GPIO pins if used.

tca9548.Select(Channel.Two);
// Do work with mcp23008_2 and mcp23008_3.

mcp23008_2.Dispose();  // Clean up other resources like GPIO pins if used.
mcp23008_3.Dispose();  // Clean up other resources like GPIO pins if used.

mcp23008I2cDevice.Dispose();  // Finially releases I2C device.
```

If we end up not adding an `I2cController`, we should at least consider the following.

1. Add `ShouldDipose` property to `I2cConnectionSettings` with default value of true.  Seems like a better fit than in I2cDevice.
2. Bindings should continue to implement IDisposable.
3. The binding guidance should be to provide two constructors where one allows for `I2cConnectionSettings` (for default) and another allows for `I2cDevice` (for non-default types).

```csharp
// For default.  Creates based on OS app executing.
Tca9548 tca9548 = new Tca9548(new I2cConnectionSettings(1, 0x40), ...);

// For non-default.
GpioI2cDevice gpioI2cDevice = new GpioI2cDevice(new I2cConnectionSettings(1, 0x40), ...);
Tca9548 tca9548 = new Tca9548(gpioI2cDevice, ...);

```

krwq:
@shaggygi, IMO we should remove the shouldDispose overall and that would remove weird abstraction (and also it is still possible to add it in the future without breaking anything).

I feel like I2cSettings should not have any board specific things so specifically it should only have address and perhaps some other I2c generic stuff. I feel like everything else should have either specific implementation which takes settings and extra info: for SoftI2c that would be SDA/SCL pin numbers and GpioController, for hardware implementation it would be busId and settings

shaggygi:
@krwq 

> public class I2cDevice
> {
>   public static I2cDevice Create(I2cConnectionSettings settings);
> }

I couldn't figure a way to use since I2cDevice is abstract class (without doing the which OS logic.  So in local work I have it in I2cController.[OS].cs files.

And should `Create` be `Open` to follow your previous comment?

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
shaggygi: Not included any longer.

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
shaggygi: Not included any longer.

src/System.Device.Gpio/System/Device/I2c/I2cController.cs
shaggygi: Not included any longer.

src/System.Device.Gpio/System/Device/I2c/II2cController.cs
shaggygi: Not included any longer.

krwq:
@joperezr I think the current version looks reasonable, we only got APIs we are sure we want to have which should be sufficient for 3.0. Do we need any process to push this through somehow?

joperezr:
This looks good. It is true that we were overengineering it a bit, since while it makes a lot of sense to have more complex controllers for gpio and PWM, it probably doesn't make that much sense to have it for I2c and Spi. My only comment here is: should we have the Create static method in the I2cController? or simply just add it to I2cDevice itself?

shaggygi:
> or simply just add it to I2cDevice itself?

While I can't explain/justify some of the work we didn't add, I think having in controller gives us room to grow.  I still believe there will be a need for II2cController/ISpiController one day and the default controllers could add future support.

Other than that, we can put the `Create` wherever.  I guess we just need to make sure we want `Create` or  `Open` as @krwq mentioned.

joperezr:
> I still believe there will be a need for II2cController/ISpiController one day and the default controllers could add future support.

I agree that we need an interface for I2c and SPI functionalities, but we could also just have II2cDevice and ISpidevice right?

shaggygi:
> I agree that we need an interface for I2c and SPI functionalities, but we could also just have II2cDevice and ISpidevice right?

I suppose it wouldn't be a problem.

krwq:
My vote is to put Create under SpiDevice. Anything else we can add later

shaggygi:
Referencing: https://github.com/dotnet/iot/pull/505#issuecomment-504683779

I know I'm overlooking something, but not sure how to proceed while moving under `I2cDevice` class.

krwq:
@shaggygi does it work if you make it so that you have 3 copies of I2cDevice similarly as we have for other classes (make sure to mark class as partial):
- I2cDevice.cs - non-os specific version
- I2cDevice.Windows.cs - windows specific implementation
- I2cDevice.Unix.cs - Unix specific implementation

you will need to modify csproj and move the `<Compile` under correct `<ItemGroup` with appropriate conditionals (follow whatever we do for other os specific files)

shaggygi:
@krwq comparing to other files we use *.[OS].cs convention on, they are `sealed partial` where I2cDevice is `abstract`.

krwq:
just do `partial` without sealed (basically repeat all decorators just add partial everywhere)

shaggygi:
Gotcha.  It's the way VS and red squiggles trip me up and how this project is structure.

krwq:
Thanks @shaggygi!

krwq:
@shaggygi planning to also do matching PR on SPI?

shaggygi:
@krwq @joperezr i’ll be honest i didn’t necessarily like the outcome of this, but understand it was probably best to make it simpler than originally thought.  Working with a team, especially remotely, is sometimes difficult to express and discuss goals effectively.  But hey, that’s why we need gate keepers to keep us all inline.  Thank you for this.

One thing I’ve definitely learned about OSS projects is you don’t always get what you want.  But that is not what it’s supposed to be about.  Even though sometimes frustrating, for me it’s about learning from others and trying to understand their point of view to improve my coding skills. I usually learn something new each PR and this one was no different.  Again, thanks for taking the time while we worked on this.

Until next time... i already have the next PR setup for SPI Create like we did for I2C.  I’ll push that soon.  Once those are checked in, i can scan/update the bindings to make them check device type in samples.

Thx again 😄 👍

shaggygi:
@krwq 

> planning to also do matching PR on SPI?

Added #527 just now.

joperezr:
> i’ll be honest i didn’t necessarily like the outcome of this, but understand it was probably best to make it simpler than originally thought.

Hey @shaggygi, just a thought, if you ever feel like some path we took is wrong you should feel free and empowered to say it and push for a fix. This project is as much yours as it is ours, hence we decided to make it open source 😄. I think I understand your sentiment, where  you wanted all 4 protocols to feel like part of the same ecosystem, so having controllers for all 4 was a great way to do that. I think that the main question that @krwq and I raised here is that having a controller for Spi and I2c might be a bit of an overkill as those controllers wouldn't have much functionallity around them, and would really only serve as "Collections" of devices. There were also not many scenarios where having a collection of devices would be useful, as mostly all consumers will be working with device bindings and those would only take one device. That is kind of why we thought that perhaps it wasn't worth to create a Controller just for one method (Create), so putting it as a static method in the device made more sense. That said, I'm just explaining our reasoning, but it is totally fair for you to not agree with this, and if so, let's start a discussion to find the best design :)

PR #511, opened 2019-06-19
Add Bh1745 device binding

RobinTTY:
This adds a device binding for the Bh1745 digital color sensor.

Some of the things I wasn't sure about and that might need to be changed:

1. I used the Color struct from System.Drawing to return a color, I don't know if you would want to implement something separate from that in Iot.Units
2. Currently there is only a method to get the compensated color, to get uncompensated values you would need to get the different color register values from their properties. I didn't see a big need to implement an extra method to get all 3 values since it's probably rare that you would want uncompensated values
<s>3. I used properties instead of methods pretty heavily, this is personal preference and I'm open to change them to methods if that's preferential</s>
<s>4. The sensor offers the ability to read a register used for internal calculation (dint register), I exposed it trough the API but don't really see much use for this value</s>


joperezr:
> Also not really sure why that one build task fails since the same issue was fixed for the other tasks with my last commit.

@RobinTTY the problem is that we have two projects that build the same files, one is the binding project which you have fixed, but the other one is the IoT.Device.Bindings.csproj project, so you would have to add your fix/property into that project as well.

RobinTTY:
@joperezr thanks for the help! Should be good now.

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -8,6 +8,7 @@
     <DeviceRoot>$(MSBuildThisFileDirectory)../devices/</DeviceRoot>
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
     <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);PackageRefAssembly</TargetsForTfmSpecificContentInPackage>
+	  <LangVersion>8</LangVersion>
```
joperezr: Seems like you have a tab here. Also, we already did this in a different PR so you don't need to do it any longer.

src/devices/Bh1745/samples/Bh1745.Sample.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Iot.Device.Bh1745.Samples
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            // bus id on the raspberry pi 3
+            const int busId = 1;
+
+			// create device
```
joperezr: looks like you have some tabs in your .cs and .csproj files, do you mind changing all of them to spaces instead?

RobinTTY: yeah seems like I didn't check the files before comitting, shame on me 😄 Should be fixed

joperezr:
@RobinTTY Is this binding similar to our existing Bh1750 binding?

RobinTTY:
> @RobinTTY Is this binding similar to our existing Bh1750 binding?

Not really, this sensor can measure 3 channels of visible light and illuminance. The Bh1750 only measures illuminance from looking at the code and outputs the value in the SI unit lux while this sensor does not convert it too lux.

RobinTTY:
Looking at the Bh1750 binding a little more I think these two have very limited similarities if any. This device does not have the same measuring modes or commands.

src/devices/Bh1745/Bh1745.cs
```diff
@@ -0,0 +1,430 @@
...
+        /// <summary>
+        /// Gets or sets the state of the software reset.
+        /// True indicates that the initial reset is started, false that it is not started.
+        /// On reset all registers are reset and the sensor goes to power down mode.
+        /// </summary>
+        public bool SoftwareReset
+        {
+            get
+            {
+                var status = Read8BitsFromRegister((byte)Register.SYSTEM_CONTROL);
```
joperezr: would it make sense to cache this value once it's been read once? Presumably the only way to change it is by calling set, so after you have read once, then you will be able to know when changes to the cached value happen right?

src/devices/Bh1745/Bh1745.cs
```diff
@@ -0,0 +1,430 @@
...
+
+        /// <summary>
+        /// Gets or sets the currently set measurement time.
+        /// </summary>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown if invalid MeasurementTime is set.</exception>
+        public MeasurementTime MeasurementTime
+        {
+            get
+            {
+                var time = Read8BitsFromRegister((byte)Register.MODE_CONTROL1);
```
joperezr: Same question for all properties, can these guys be cached? In case not, we should probably change these methods to be a GetMeasurementTime and SetMeasurementTime because of the discussion we are having here: #626

src/devices/Bh1745/Bh1745.cs
```diff
@@ -0,0 +1,430 @@
...
+                var valid = Read8BitsFromRegister((byte)Register.MODE_CONTROL2);
+                valid = (byte)(valid & (byte)Mask.VALID);
+                return Convert.ToBoolean(valid);
+            }
+        }
+
+        /// <summary>
+        /// Gets the data in the red data register.
+        /// </summary>
+        public ushort RedDataRegister => Read16BitsFromRegister((byte)Register.RED_DATA);
```
joperezr: I think all of these should instead be methods instead of readonly properties. See #626 for more info

RobinTTY:
Oh just as a note, I will get to this after #482.

RobinTTY:
This is now finally ready for review too!

krwq:
@RobinTTY [my PR](https://github.com/dotnet/iot/pull/685) could have caused conflicts here. Please rebase your PR and make sure to fix all of the docs issues (previously they might not have been reported and PR might have stayed green)

RobinTTY:
@krwq done

src/devices/Bh1745/Bh1745.cs
```diff
@@ -0,0 +1,394 @@
...
+        /// <summary>
+        /// Gets or sets the channel compensation multipliers which are used to compensate the measurements.
+        /// </summary>
+        public ChannelCompensationMultipliers ChannelCompensationMultipliers { get; set; }
+
+        /// <summary>
+        /// Resets the device to the default configuration.
+        /// On reset the sensor goes to power down mode.
+        /// </summary>
+        public void Reset()
```
krwq: Should you also update any of the private fields too to reflect the state?

src/devices/Bh1745/Bh1745.cs
```diff
@@ -0,0 +1,394 @@
...
+        /// </summary>
+        /// <returns></returns>
+        public Color GetCompensatedColor()
+        {
+            var clearDataRaw = ReadClearDataRegister();
+            if (clearDataRaw == 0)
+                return Color.FromArgb(0, 0, 0);
+
+            // apply channel multipliers and normalize
+            var compensatedRed = ReadRedDataRegister() * ChannelCompensationMultipliers.Red / MeasurementTime.ToMilliseconds() * 360;
```
krwq: please use explicit types here (especially it seems it's same number of characters)

src/devices/Bh1745/Bh1745.cs
RobinTTY: Oh yeah good catch forgot about that when I changed from actual device polling to caching through fields.

RobinTTY:
@krwq addressed the feedback

PR #518, opened 2019-06-22
Refactoring DHT & Re-adding DHT12 & Adding DHT10 Binding

ZhangGaoxing:
DHT10 is a new product released by ASAIR in May this year. It is used to replace DHT11 with the same precision as DHT22. In README, I compared the technical details of DHT sensors. I bought one for about 5 RMB. When I added the binding, I found that the read logic of DHT10 was different from that of previous DHT sensors, so I refactored the code. At present, only the Chinese version of the datasheet is available. Once I find that the English version is updated, I will update README in time.

I hope you can get it as soon as possible, haha😄

image


src/devices/Dhtxx/Devices/Dht10.cs
```diff
@@ -0,0 +1,98 @@
+﻿using System;
```
krwq: copyright header

src/devices/Dhtxx/Devices/Dht12.cs
```diff
@@ -0,0 +1,52 @@
...
+        internal override double GetHumidity(byte[] readBuff)
+        {
+            return readBuff[0] + readBuff[1] * 0.1;
+        }
+
+        internal override Temperature GetTemperature(byte[] readBuff)
+        {
+            var temp = readBuff[2] + (readBuff[3] & 0x7F) * 0.1;
+            // if MSB = 1 we have negative temperature
+            temp = (readBuff[3] & 0x80) == 0 ? temp : -temp;
```
krwq: double checking this is the right number format and not [2's complement](https://en.wikipedia.org/wiki/Two%27s_complement)?

src/devices/Dhtxx/Devices/Dht21.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System.Device.Gpio;
+using Iot.Units;
+
+namespace Iot.Device.DHTxx
+{
+    /// <summary>
+    /// Temperature and Humidity Sensor DHT21
+    /// </summary>
+    public class Dht21 : DhtBase
```
krwq: how's this code different from Dht12

src/devices/Dhtxx/Devices/Dht22.cs
```diff
@@ -0,0 +1,38 @@
...
+        public Dht22(int pin, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical)
+            : base(pin, pinNumberingScheme)
+        { }
+
+        internal override double GetHumidity(byte[] readBuff)
+        {
+            return (readBuff[0] << 8 | readBuff[1]) * 0.1;
+        }
+
+        internal override Temperature GetTemperature(byte[] readBuff)
```
krwq: are those all the same?

src/devices/Dhtxx/DhtBase.cs
```diff
@@ -128,7 +115,7 @@ private void ReadData()
         /// <returns>
         /// <c>true</c> if read is successfull, otherwise <c>false</c>.
         /// </returns>
-        private bool ReadThroughOneWire()
+        internal virtual void ReadThroughOneWire()
```
joperezr: Wasn't the whole point of IsLastReadSuccessful to make sure that we got a valid accurate reading? if you are removing it, how will we be certain of that now? Also, the xml docs above the method will have to be reflected with this change.

src/devices/Dhtxx/Devices/Dht21.cs
ZhangGaoxing: They are different sensors, and I listed the parameters in readme.

src/devices/Dhtxx/Devices/Dht22.cs
ZhangGaoxing: No, because I don't have all the sensors, I overlooked this issue in previous PRs. So `GetTemperature` needs to be override.

src/devices/Dhtxx/DhtBase.cs
ZhangGaoxing: Because DHT10 does not provide validation, I just let this method not return bool values. `IsLastReadSuccessful` still gives value.

ZhangGaoxing:
reopen

krwq:
@ZhangGaoxing let me know once you finish making changes

ZhangGaoxing:
@krwq I think we can start review😄

src/devices/Dhtxx/Devices/Dht22.cs
krwq: If the code is the same please create some base class for them or just inherit one from another

ZhangGaoxing:
@krwq @joperezr

PR #527, opened 2019-06-25
Added SPI Create method

### LINKED ISSUE #119: [Proposal] SpiController and ISpiController
Similar to the ```GpioController```, there needs to be a mechanism to determine the best ```SpiDevice``` to use.

#### Updated 6/24/2019

# Proposed Changes

## 1. Add New SpiController

The creation of devices with a new static helper method that attempts to pick the best SPI device for OS app is executing on.  This static method will be added to a new `sealed partial class` respective to OS.

```csharp
// SpiController.Linux.cs
public static SpiDevice Create(SpiConnectionSettings settings) => new UnixSpiDevice(settings);
// SpiController.Windows.cs
public static SpiDevice Create(SpiConnectionSettings settings) => new Windows10SpiDevice(settings);
```

shaggygi:
Resolves #119

This adds new Create method to return SpiDevice based on OS app is executing.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Spi/SpiDevice.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Spi/SpiDevice.Linux.cs
@@ -1,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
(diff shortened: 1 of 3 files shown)


joperezr:
We didn't have a successful build last night with the I2c Create methods, so I just queued one now that will contain this change as well and will submit a PR to ingest those changes, so that we can use this new API in our device bindings.

Issue #498, opened 2019-06-13
SenseHat does not create appropriate I2C devices


krwq:
Problem is that we do not have a generic way to create I2C/SPI on Windows and Linux (i.e. I2cDevice.Create(..)).

I.e. https://github.com/krwq/iot/blob/master/src/devices/SenseHat/SenseHatLedMatrixI2c.cs#L142

this is causing that users have to use workarounds in order to use SenseHat on Windows.

cc: @joperezr

shaggygi:
Referencing https://github.com/dotnet/iot/issues/118#issuecomment-496053073

krwq:
I'm seeing 2 options:
- long term: System.Device: SpiDevice.Create/I2cDevice.Create
- short term: Iot.Device.Binding: GenericSpiDevice/GenericI2cDevice (put it somewhere in the common folder)

joperezr:
I agree with @shaggygi, this is a dupe of #118 and honestly I wouldn't rush adding any of the two solutions, and instead implement a proper SpiController and I2cController. I think adding these is not blocking 3.0 as what our bindings currently do is to take a SpiDevice or a I2cDevice in their constructor, but it would be a nice to have. I'm closing this as a dupe, but @krwq feel free to add your suggestions into #118

shaggygi:
This should actually be reopened even though this discussion is related to ideas how to fix the issue.

We can fix/update the SenseHat binding and close this issue after #118 is completed.

PR #528, opened 2019-06-25
Update samples using I2c to work on Windows

krwq:
Fixes: https://github.com/dotnet/iot/issues/498

**This won't build until we get new package version**

This incorporates changes done here: https://github.com/dotnet/iot/pull/505

### DIFF
--- a/src/devices/Ads1115/README.md
+++ b/src/devices/Ads1115/README.md
@@ -9,10 +9,7 @@ ADS1115 is an Analog-to-Digital converter (ADC) with 16 bits of resolution.
 // set I2C bus ID: 1
 // ADS1115 Addr Pin connect to GND
 I2cConnectionSettings settings = new I2cConnectionSettings(1, (int)I2cAddress.GND);
-// get I2cDevice (in Linux)
-UnixI2cDevice device = new UnixI2cDevice(settings);
-// get I2cDevice (in Win10)
-//Windows10I2cDevice device = new Windows10I2cDevice(settings);
+I2cDevice device = I2cDevice.Create(settings);
 
 // pass in I2cDevice
 // measure the voltage AIN0
--- a/src/devices/Ads1115/samples/Program.cs
+++ b/src/devices/Ads1115/samples/Program.cs
@@ -18,9 +18,7 @@ namespace Ads1115.Samples
             // ADS1115 Addr Pin connect to GND
             I2cConnectionSettings settings = new I2cConnectionSettings(1, (int)I2cAddress.GND);
             // get I2cDevice (in Linux)
-            UnixI2cDevice device = new UnixI2cDevice(settings);
-            // get I2cDevice (in Win10)
-            //Windows10I2cDevice device = new Windows10I2cDevice(settings);
+            I2cDevice device = I2cDevice.Create(settings);
 
             // pass in I2cDevice
             // measure the voltage AIN0
--- a/src/devices/Ads1115/samples/README.md
+++ b/src/devices/Ads1115/samples/README.md
@@ -26,8 +26,7 @@ Rotary Potentiometer
 // set I2C bus ID: 1
 // ADS1115 Addr Pin connect to GND
 I2cConnectionSettings settings = new I2cConnectionSettings(1, (int)I2cAddress.GND);
-// get I2cDevice (in Linux)
-UnixI2cDevice device = new UnixI2cDevice(settings);
+I2cDevice device = I2cDevice.Create(settings);
 
 // pass in I2cDevice
 // measure the voltage AIN0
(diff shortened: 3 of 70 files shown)


krwq:
@shaggygi I'll leave you SPI :smile:

src/devices/Ads1115/README.md
```diff
@@ -9,10 +9,7 @@ ADS1115 is an Analog-to-Digital converter (ADC) with 16 bits of resolution.
 // set I2C bus ID: 1
 // ADS1115 Addr Pin connect to GND
 I2cConnectionSettings settings = new I2cConnectionSettings(1, (int)I2cAddress.GND);
-// get I2cDevice (in Linux)
-UnixI2cDevice device = new UnixI2cDevice(settings);
-// get I2cDevice (in Win10)
-//Windows10I2cDevice device = new Windows10I2cDevice(settings);
+I2cDevice device = I2cDevice.Create(settings);
```
joperezr: Just as a comment, now that we have I2cDevice.Create method, we might want to think again how most bindings take a I2cDevice as part of their constructor, such that insted they would just create a new one inside their class. There were only two reasons to have it passed in: support for windows/unix (which we no longer need) and in order to allow passing in a device that is not a Unix or Windows I2c device, but instead something that implementes the II2cDevice interface, like a I2C extender.

shaggygi: Could the guidance be to pass in I2cDevice? or SpiDevice? similar to IGpioController.  As you mentioned, I could see other flavors (i. e. GpioI2cDevice, GpioSpiDevice) in the future. 

Inside the binding constructor, have something like...

```csharp
_i2cDevice = I2cDevice ?? I2cDevice.Create(settings);
_spiDevice = SpiDevice ?? SpiDevice.Create(settings);
```

Also, with guidance showing the nullable types to pass... could hopefully trigger binding devs to code the binding API with the new C# 8 feature.

joperezr:
How did you find all these places? Did you just do a search on all places where we were doing a `new UnixI2cDevice(` or `new WindowsI2cDevice` ?

src/devices/Ads1115/README.md
krwq: I'd refrain from that for now unless it covers 95% of the cases. Example thing what I did for SenseHat:
- Lsm9Ds1AccelerometerAndGyroscope doesn't know anything about I2C addresses
- SenseHatAccelerometerAndGyroscope thin wrapper on top of Lsm9Ds1AccelerometerAndGyroscope which only defaults I2C address since it's constant in context of SenseHat

Ideally I think we should create an issue and set up guidelines for this but I think this is non-issue at this point and regardless of this issue we need at least one constructor which takes already existing I2cDevice object.

I don't personally like default value as the ownership is not immediately obvious. Having two classes one which always default and one which never does makes stuff much clearer

krwq:
@joperezr almost. I searched for `UnixI2cDevice` and `WindowsI2cDevice` (no new) in src/devices and samples and manually reviewed the code if it needs some more adjustments. Find & replace covers 90% though

joperezr:
Why only src/devices? should we do it on the root as well in case we were doing this in the tools or root samples?

src/devices/Ads1115/README.md
krwq: I'd prefer if we made this non-nullable and create second implementation like DefaultXyzDevice which just inherits, creates I2cDevice and disposes of it. This way you got covered 90% of the cases and advanced usage

shaggygi:
Yes, I was just looking at DeviceApiTester and will update those, as well with `var` or `I2cDevice`/`SpiDevice`

krwq:
@joperezr I meant root samples when I said samples.

shaggygi:
@krwq @joperezr I get local build errors now with the following package added in System.Device.Gpio.csproj.  All projects build in iot directory after I remove it.  Thoughts?

```xml
<PackageReference Include="Microsoft.DotNet.GenAPI" Version="$(MicrosoftDotNetGenApiPackageVersion)">
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
```

joperezr:
@shaggygi do you mind sharing a binlog of the error you are seeing so that I can take a better look? In order to produce one, just run `msbuild System.Device.Gpio.csproj /t:rebuild /bl:msbuild.binlog` on a developer command prompt which will generate a file called msbuild.binlog.

shaggygi:
@joperezr  Does this work?  Just remove the .zip.
[msbuild.binlog.zip](https://github.com/dotnet/iot/files/3332123/msbuild.binlog.zip)

joperezr:
>  Does this work? Just remove the .zip.

mmm that is very weird, does this repro if you are on a clean master? try doing:
```
git pull
git reset --hard
git clean -fdx
build.cmd
```
This will get latest changes, remove all pending changes, clean up your enlistment and run build again.

krwq:
also if you cancelled build during restore it might be worth to clean nuget cache (`$HOME/.nuget` or `%userprofile%\.nuget`)

shaggygi:
> git pull
> git reset --hard
> git clean -fdx
> build.cmd

This appears to have fix the error.  One thing I noticed was a version of preview6 of something (in the output text) where this has preview5.  Not sure if was SDK, runtime, etc. 😕 

Thx

PR #530, opened 2019-06-26
Updating dependencies from dotnet/iot

joperezr:
FYI: @shaggygi @krwq 

This will update our iot dependencies so that we get the I2c and Spi Create support.


krwq:
@joperezr how was this PR generated? Could you add some instructions in the readme or somewhere in the repo?

Issue #126, opened 2018-12-29
[Proposal] Move classes and rename to be consistent throughout API


shaggygi:
This proposal is a collection of different modifications and grouped here as there are many breaking changes being proposed.  The intent is to review, decide and update (if needed) namespaces, class names and file locations in a consistent structure between the different types of interfaces within System.Device.Gpio API.

This seems to fit with @terrajobst comments during [this design review](https://youtu.be/UZc3sbJ0-PI?t=4656), as well.

## Rationale and Usage

Since System.Device.Gpio is in early previews, it seems like the opportune time to review and decide how to structure the core components.  This will reduce breaking changes in the future.

## Examples

- GPIO and PWM have abstract classes based on **Drivers** and are located under the *.**Drivers** folder.  This includes RaspberryPi3Driver, UnixDriver.Linux.cs, etc..  Gpio is not in the file names where I2C/Pwm/SPI are included in the file names.
- I2C and SPI are based on **Devices**.  However, there are implementations under the *.**Drivers** folders.

## Proposed Changes

The proposed changes are to rename and move files as shown in the folder structure below.  To simplify, this attempts to point out only the changes for the files being modified.

### GPIO Folder
- [ ] TODO
1. Drivers and related classes renamed to include "Gpio".
2. GpioDriver.cs moved to Drivers folder.
```
Device
  Gpio
    Drivers
      GpioDriver.cs
      HummingBoardGpioDriver.cs
      HummingBoardGpioDriver.Linux.cs
      HummingBoardGpioDriver.Windows.cs
      RaspberryPi3GpioDriver.cs
      RaspberryPi3GpioDriver.Linux.cs
      RaspberryPi3GpioDriver.Windows.cs
      UnixGpioDriver.Linux.cs
      UnixGpioDriver.Windows.cs
      UnixGpioDriverPin.Linux.cs  // Also simplified by removing Device
      Windows10GpioDriver.Linux.cs
      Windows10GpioDriver.Windows.cs
      Windows10GpioDriverPin.Windows.cs
    GpioPinChangeEventHandler.cs
    GpioPinEventTypes.cs
    GpioPinMode.cs
    GpioPinNumberingScheme.cs
    GpioPinValue.cs
    GpioPinValueChangedEventArgs.cs
```

### I2C Folder
- [x] Completed in #543
1. Folder renamed from "Drivers" to "Devices".
2. Namespaces be renamed from "System.Device.I2c.Drivers" to "System.Device.I2c"
[...]

PR #533, opened 2019-06-27
Rename I2C Drivers Folder

shaggygi:
This is part of #126

Renames I2C Drivers folder to Devices.


shaggygi:
@krwq This first change doesn't appear to break anything (yet).  Do you want me to add more commits under this PR for the namespace updates or create another PR specifically for those?

krwq:
@shaggygi let's do it separately since we will likely squash and merge which can break the diff

shaggygi:
@krwq I just pushed the button as I saw your response

joperezr:
> I just pushed the button as I saw your response

That is fine, we will take it as is, I'm just taking a final look and making sure CI is green.

joperezr:
The reason why your PR is red is because you premptively changed samples on bindings that haven't ingested the changes on this PR. Remember that the samples use a package reference to System.Device.Gpio, so it will take us first to merge this PR, and then to ingest those changes in order to be able to use the new namespaces in the samples. For this PR, we probably only want the namespace/file renames plus anything that would break the build, and then in the ingestion PR we will handle the namespace renames on the samples.

shaggygi:
Gotcha.  So I basically need to revert back to b86f15d.

krwq:
so basically:
- do renames only in System.Device.Gpio (no renames in the samples)
- when new package gets produced we will get PR which becomes red and the renames for samples should be added to that PR

shaggygi:
Well snap.  DeviceApiTester is referencing directly to core project unlike bindings. So that's why it failed last time.  Let's try again 😦

src/System.Device.Gpio/System/Device/I2c/Devices/I2cDevice.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-namespace System.Device.I2c
+namespace System.Device.I2c.Devices
```
krwq: Is this change expected as well? I'd expect I2cDevice to live under System.Device.I2c and implementations under .Devices. cc: @joperezr

shaggygi: I was thinking this was part of...

> 3. I2cDevice.cs moved to Devices folder.

krwq: don't have strong opinion, it just feels weird from naming point of view. @joperezr, thoughts?

joperezr: mm I also thought about this, but I think it does make sense to have it moved to this new namespace as well. It makes sense that a Device is on the Devices namespace :)

krwq: @joperezr I'm wondering, perhaps we should make Windows/LinuxI2cDevice internal (I don't see good justification for having them right now) and just don't rename namespace (just add directories)

krwq: We talked offline with @joperezr and we will only have 1 I2cDevice class which collapses the logic of both Linux and Windows and I2cDevice.Create will become `new I2cDevice`. We are also thinking of doing similar thing for board drivers but that one has a bit more justification on having them separate so we need to figure out what should happen.

joperezr: We will do that in a separate PR though, so we can take this in first and we will make the required changes subsequently

shaggygi: Do I need to create the next PR to update the bindings to reflect these changes or wait?  Also, I was going to create another PR similar to this for SPI.  Do I need to hold off on that base on the planned changes?

PR #537, opened 2019-07-01
Update permissions of dotnet-install.sh

krwq:
Every time I locally build iot repo on linux some build script updates permissions of dotnet-install.sh and it shows that my repo is unclean. This submits that change permanently.


krwq:
@joperezr merging this since it seems harmless

joperezr:
Interesting, I'm not sure if this change will be persisted as everything under /eng/common gets overridden by the arcade updates.

krwq:
@joperezr perhaps we should generate PRs on linux then? (or both)

joperezr:
that won't make a difference, what I meant is that darc updates from arcade rewrites these files automatically and submit PRs for the changes, so I wonder if this change will remain, in any case, I don't think that would be an issue. If you see this happen again in the future after this fix we can investigate more.

PR #538, opened 2019-07-01
Re-generate device listing

krwq:


src/devices/README.md
```diff
@@ -8,38 +8,47 @@ Our vision: the majority of .NET bindings are written completely in .NET languag
...
 * [ADS1115 - Analog to Digital Converter](Ads1115/README.md)
 * [AGS01DB - MEMS VOC Gas Sensor](Ags01db/README.md)
 * [Cpu Temperature](CpuTemperature/README.md)
 * [Sense HAT](SenseHat/README.md)
 * [BrickPi3](BrickPi3/README.md)
+* [BH1750FVI - Ambient Light Sensor](Bh1750fvi/README.md)
 * [NXP/TI PCx857x](Pcx857x/README.md)
 * [Character LCD (Liquid Crystal Display)](CharacterLcd/README.md)
 * [HMC5883L - 3 Axis Digital Compass](Hmc5883l/README.md)
+* [Solomon Systech Ssd1306](Ssd13xx/README.md)
```
krwq: seems I need to add sorting logic :P

tools/device-listing/DeviceInfo.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public int CompareTo(DeviceInfo other)
+        {
+            return Title.CompareTo(other.Title);
+        }
+
+        private static string GetTitle(string readmePath)
+        {
+            string[] lines = File.ReadAllLines(readmePath);
+            if (lines[0].StartsWith("# "))
```
joperezr: I guess it's fine since we control this, but it might be better here if we just change this to have the `#` be replaced with a ` ` and then we trim the beginning spaces, that way we will work even when more than one space is left.

tools/device-listing/device-listing.cs
```diff
@@ -35,22 +36,37 @@ static void Main(string[] args)
                 string readme = Path.Combine(directory, "README.md");
                 if (File.Exists(readme))
                 {
-                    string title = GetTitle(readme);
-                    if (title == null)
+                    var device = new DeviceInfo(readme);
+
+                    if (device.Title == null)
```
joperezr: Should we ensure the Title is unique? I suppose that will be the case in most bindings, but we might have the case where the binding differed in the second word so the Title might not be unique.

joperezr:
Left two small comments but this looks good otherwise, thanks for updating this @krwq

tools/device-listing/DeviceInfo.cs
krwq: @joperezr I prefer it this way since you will get warnings when README is misformatted - I believe we should fix misformatting instead of ignoring it here

tools/device-listing/device-listing.cs
krwq: I feel this will be easy to spot on the diff now since I started sorting. I'll leave it as is for now and add extra logic whenever we find a real case like this.

Issue #524, opened 2019-06-25
Raspberry Pi 4 Model B Support


shaggygi:
Most likely is supported, but needs to be verified and added to [list](https://github.com/dotnet/iot/tree/master/samples#requirements).

https://www.raspberrypi.org/blog/raspberry-pi-4-on-sale-now-from-35/

EDIT (@krwq)
Seems Raspberry Pi 4 is supported but we need to update board detection logic and pick LibgpiodDriver automatically

joperezr:
Awesome, yeah I saw the announcement yesterday. I'm almost sure that .net core should run just fine there, so we shouldn't have to worry but it is nice to have this to be able to track the documentation change.

dobova86:
Not working for me under netcore 3.0 preview6. The code is working fine on pi 3/3+/cm3+.
Pi 4 (2GB) with Buster installed. Unless I forgot some dependancy for System.Device.Gpio.Driver on the Pi 4 ...

> Unhandled Exception: System.AggregateException: One or more errors occurred. (Error 0 initializing the Gpio driver.) ---> System.IO.IOException: Error 0 initializing the Gpio driver.
>    at System.Device.Gpio.Drivers.RaspberryPi3Driver.Initialize()
>    at System.Device.Gpio.Drivers.RaspberryPi3Driver.OpenPin(Int32 pinNumber)
>    at System.Device.Gpio.GpioController.OpenPin(Int32 pinNumber)
>    at System.Device.Gpio.GpioController.OpenPin(Int32 pinNumber, PinMode mode)
>    at Device.BG96Iot.BG96Iot..ctor() in H:\PROJECTS.2019\RPITest\RpiBG96Test\ConsoleApp1\BG96Iot.cs:line 61
>    at RpiBG96Test.Program.MainSetup() in H:\PROJECTS.2019\RPITest\RpiBG96Test\ConsoleApp1\Program.cs:line 30
>    at RpiBG96Test.Program.MainLoop() in H:\PROJECTS.2019\RPITest\RpiBG96Test\ConsoleApp1\Program.cs:line 38
>    --- End of inner exception stack trace ---
>    at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
>    at System.Threading.Tasks.Task.Wait()
>    at RpiBG96Test.Program.Main(String[] args) in H:\PROJECTS.2019\RPITest\RpiBG96Test\ConsoleApp1\Program.cs:line 24
[...]

PR #540, opened 2019-07-01
Change to address RaspberryPi4B failing when trying to initialize the…

Frankenslag:
When initializing the GPIO driver on a RaspberyPi 4B then an IO error is produced on some occasions. Please see #524 

The issue appears to be caused by the code testing for a negative return from `mmap` indicating a fault condition whereas `mmap` is documented to return -1 in such cases. This PR changes the tests in the RaspberryPi3Driver to just test for -1.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -363,13 +363,13 @@ namespace System.Device.Gpio.Drivers
                 }
 
                 int fileDescriptor = Interop.open(GpioMemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
-                if (fileDescriptor < 0)
+                if (fileDescriptor != -1)
                 {
                     throw new IOException($"Error {Marshal.GetLastWin32Error()} initializing the Gpio driver.");
                 }
 
                 IntPtr mapPointer = Interop.mmap(IntPtr.Zero, Environment.SystemPageSize, (MemoryMappedProtections.PROT_READ | MemoryMappedProtections.PROT_WRITE), MemoryMappedFlags.MAP_SHARED, fileDescriptor, GpioRegisterOffset);
-                if (mapPointer.ToInt32() < 0)
+                if (mapPointer.ToInt32() != -1)
                 {
                     throw new IOException($"Error {Marshal.GetLastWin32Error()} initializing the Gpio driver.");
                 }
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -363,13 +363,13 @@ namespace System.Device.Gpio.Drivers
                 }
 
                 int fileDescriptor = Interop.open(GpioMemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
-                if (fileDescriptor != -1)
+                if (fileDescriptor == -1)
                 {
                     throw new IOException($"Error {Marshal.GetLastWin32Error()} initializing the Gpio driver.");
                 }
 
                 IntPtr mapPointer = Interop.mmap(IntPtr.Zero, Environment.SystemPageSize, (MemoryMappedProtections.PROT_READ | MemoryMappedProtections.PROT_WRITE), MemoryMappedFlags.MAP_SHARED, fileDescriptor, GpioRegisterOffset);
-                if (mapPointer.ToInt32() != -1)
+                if (mapPointer.ToInt32() == -1)
                 {
                     throw new IOException($"Error {Marshal.GetLastWin32Error()} initializing the Gpio driver.");
                 }


src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -363,13 +363,13 @@ private void Initialize()
                 }
 
                 int fileDescriptor = Interop.open(GpioMemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
-                if (fileDescriptor < 0)
+                if (fileDescriptor != -1)
```
joperezr: are you saying that the result you are getting back from open and mmap is negative but that is still a success case? After making this change, where you able to use this driver to control pins in a raspberry pi 4?

Frankenslag: I am saying that the value I am getting back for a file descriptor is better compared to -1 as that is what the documentation says.

 This wasn't causing the error though … The subsequent mmap  was returning a value of 0xB5302000 when I tested it and it was then pushed through `mapPointer.ToInt32()` thus giving a negative number and failing the test. (hence the change to now test against -1)

I have not yet tested this from the point of view of driving pins I have just tested that the driver gets initialized without an IO error.
…

Frankenslag: **Oops. I see why the tests are failing …. Will send an amended PR**

krwq: Yes, Per http://man7.org/linux/man-pages/man2/open.2.html

> open(), openat(), and creat() return the new file descriptor, or -1
> if an error occurred (in which case, errno is set appropriately).

this seems correct

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -363,13 +363,13 @@ private void Initialize()
...
                 int fileDescriptor = Interop.open(GpioMemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
-                if (fileDescriptor < 0)
+                if (fileDescriptor == -1)
                 {
                     throw new IOException($"Error {Marshal.GetLastWin32Error()} initializing the Gpio driver.");
                 }
 
                 IntPtr mapPointer = Interop.mmap(IntPtr.Zero, Environment.SystemPageSize, (MemoryMappedProtections.PROT_READ | MemoryMappedProtections.PROT_WRITE), MemoryMappedFlags.MAP_SHARED, fileDescriptor, GpioRegisterOffset);
-                if (mapPointer.ToInt32() < 0)
+                if (mapPointer.ToInt32() == -1)
```
krwq: Should you convert to Int64? IntPtr has different size on different platform and I'm not sure this check will be correct for say ARM64

Frankenslag: Well that's a good question.... This is code that is specific to the RaspberryPi3 (and now 4). and they are Arm32 and so I suppose they have a 32bit address space which would indicate that the conversion to Int32 is correct.

PR now amended.

krwq: @Frankenslag actually RPi3 is 64-bit, only Rasbian is 32-bit. I got one RPi3 running on ARM64 Ubuntu 😄
You can use this image: https://ubuntu.com/download/iot/raspberry-pi-2-3 for 64-bit testing (or just try using ToInt64 and make the code work with 32-bit and assume it will work for 64-bit as well - a bit more risky but I guess it's ok considering RPi is mostly used with 32-bit)

Frankenslag: @krwq . I am happy to test for both ubuntu and Raspbian (I have spare SDCards). I can confirm that I have built a test nuget package and have used it with the Ssd1351 sample I was working on and this is working on the Pi4 now. So I guess that covers SPI also.

Do you want to merge the PR now so that others can get working with the Pi4 or do you want to wait until I have tested against Ubuntu (might take a day or 2). I assume you use the Arm64 dotnet core runtime package ?

krwq: @Frankenslag, I'd prefer to wait until the testing is complete unless it makes it easier to test then I'm fine with merging earlier

RheaAyase:
I can confirm that this change fixes the issue [#524](https://github.com/dotnet/iot/issues/524).

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: @Frankenslag merging this, let's do the testing and any extra fixes later

krwq: https://github.com/dotnet/iot/issues/557

Issue #355, opened 2019-04-07
MyGet feeds not updated


techniq:
[NOTE] This is tracking the work of updating our docs to point to the new feeds that contain daily builds of the packages.

I was looking to update the packages to pull in the fixes for [ADS1115](https://github.com/dotnet/iot/commit/ac92a6069024905b3ceb79911bbfba8c11e35d50) which was merged on 3/31 but it appears the last time the MyGet feeds were updated is 3/27.

- https://dotnet.myget.org/feed/dotnet-core/package/nuget/System.Device.Gpio
- https://dotnet.myget.org/feed/dotnet-core/package/nuget/Iot.Device.Bindings

I also saw packages on nuget.org but they are a little older than MyGet

- https://www.nuget.org/packages/System.Device.Gpio/0.1.0-prerelease.19171.3
- https://www.nuget.org/packages/Iot.Device.Bindings/0.1.0-prerelease.19171.3

krwq:
cc: @safern

joshfree:
> the last time the MyGet feeds were updated is 3/27.

I believe @tarekgh and @dagood may have been looking into this

safern:
This is because the MyGet feed is full. @mmitche has been trying to get this sorted out with MyGet people but haven't had response I believe.

dagood:
Yeah, see https://github.com/dotnet/core-eng/issues/5070 for details. The current recommendation is to use the blob feeds like https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json, but I don't know if IoT is configured to publish to a blob feed yet.

safern:
cc: @joperezr

joperezr:
I'll work today on making sure that we publish to blobfeed as well as to myget. That way, even if myget publish fails, there is still a place where the community can get the packages out of.

joperezr:
Just to update this thread, our myget feeds have been having a lot of issues so we haven't been able to push for a while. Instead, we are now pushing all of our official builds to a blob feed: https://dotnetfeed.blob.core.windows.net/dotnet-iot/index.json

We will eventually stop pushing packages to myget and only use the feed instead, so I will go ahead and update our documentation in the repo to point to the blob feed instead of myget in order to consume packages.

MarkCiliaVincenti:
[...]

PR #541, opened 2019-07-01
Update blob feeds

krwq:
Fixes: https://github.com/dotnet/iot/issues/355


krwq:
merging this as there is no point of doing CI on README change

PR #542, opened 2019-07-01
Updating Gpio package and updating I2c namespace

joperezr:
cc: @shaggygi @krwq 

This will get the latest package ingested, plus it will address all of the namespace changes in all the bindings and samples.


joperezr:
Merging in order to anticipate the build break

Issue #534, opened 2019-06-28
Unify platform specific implementations of I2cDevice/SpiDevice into one class


krwq:
- Make WindowsI2cDevice/LinuxI2cDevice single non-abstract class I2cDevice
- `I2cDevice.Create` => `new I2cDevice()`

open questions:
- should we have II2cDevice interface or rather have I2cDevice be unsealed
  - personally leaning toward no interface because we won't be able to add any APIs to I2cDevice after we do that and ship - on the other hand lack of clear abstraction is weird - perhaps we should have I2cDevice as is and have GenericI2cDevice
- should we do similar changes for drivers (libgpiod/sysfs/...)

cc: @joperezr @shaggygi @ZhangGaoxing

shaggygi:
How would we implement a GpioI2cDevice?  Is this where GenericI2cDevice comes into play?

krwq:
GpioI2cDevice/SoftI2cDevice assuming we don't pick the interface would derive from I2cDevice.

The only thing is that without GenericI2cDevice the abstraction for I2cDevice will be kinda weird (software implementation would not use anything from that class - it would just reuse the public surface).

On the other hand GenericI2cDevice is just awkward from naming and discoverability point of view (which feels we should have I2cDevice.Create which creates internal GenericI2cDevice).

shaggygi:
I might be overlooking this new direction, but I'm thinking we keep the `I2cDevice` and update it based on your recent comments where it can determine whether its for Linux or Windows.  This would implement `II2cDevice`.

Then the `GpioI2cDevice/SoftI2cDevice` could implement `II2cDevice`, as well.

We could pass `II2cDevice` to binding constructors similar to how we do `IGpioController`.  We could do in 2 ways.

1) Pass in only the `I2cConnectionSettings` and the default could be used from `I2cDevice` using the provided settings.
2) Pass in `II2cDevice` where it could be non-nullable and caller would have to define or nullable and if null the default could be created.

Again, not completely sure.

shaggygi:
Thinking a little more on this... why not consolidate to the following files...
**NOTE**: I changed all this locally and compiles, but have not tested.

```console
I2c  
  Devices (can possibly get rid of this directory and place I2cDevice files back at I2c folder?)
    I2cDevice.Linux.cs   // Implements II2cDevice. Move all code from UnixI2cDevice to this file.
    I2cDevice.Windows.cs  // Implements II2cDevice. Move all code from Windows10I2cDevice to this file.
[...]

PR #543, opened 2019-07-01
Rename i2c structure again

shaggygi:
This addresses the I2C portion of #534


krwq:
what you have overall looks good. I remember there are couple of XML docs which mention WIndows/LinuxI2c/Spi which we should fix since they are internal now

tools/templates/DeviceBindingTemplate/dotnet_new_device-binding_csharp/samples/_DeviceBinding.Sample.cs
```diff
@@ -5,9 +5,7 @@
 using System;
 using System.Device.Gpio;
 using System.Device.I2c;
-using System.Device.I2c.Devices;
```
krwq: looks much nicer now, two namespaces were kinda confusing

shaggygi:
> what you have overall looks good. I remember there are couple of XML docs which mention WIndows/LinuxI2c/Spi which we should fix since they are internal now

Good point.  Working on it now.

shaggygi:
@krwq I think this finishes up here.  I'll have a PR for SPI soon.

src/devices/Ads1115/Ads1115.cs
```diff
@@ -61,13 +61,13 @@ public DataRate DataRate
         /// <summary>
         /// Initialize a new Ads1115 device connected through I2C
         /// </summary>
-        /// <param name="sensor">I2C Device, like UnixI2cDevice or Windows10I2cDevice</param>
+        /// <param name="i2cDevice">>The I2C device used for communication.</param>
```
krwq: double `>`

src/devices/Ags01db/Ags01db.cs
```diff
@@ -40,19 +40,19 @@ public class Ags01db : IDisposable
         /// <summary>
         /// Creates a new instance of the ASG01DB
         /// </summary>
-        /// <param name="sensor">I2C Device, like UnixI2cDevice or Windows10I2cDevice</param>
-        public Ags01db(I2cDevice sensor)
+        /// <param name="i2cDevice">>The I2C device used for communication.</param>
```
krwq: double `>`

src/devices/Bh1750fvi/Bh1750fvi.cs
```diff
@@ -45,15 +45,15 @@ public double LightTransmittance
         /// <summary>
         /// Creates a new instance of the BH1750FVI
         /// </summary>
-        /// <param name="sensor">I2C Device, like UnixI2cDevice or Windows10I2cDevice</param>
+        /// <param name="i2cDevice">>The I2C device used for communication.</param>
```
krwq: ditto

src/devices/Dhtxx/DhtSensor.cs
```diff
@@ -90,11 +90,11 @@ public DhtSensor(int pin, DhtType dhtType, PinNumberingScheme pinNumberingScheme
         /// <summary>
         /// Create a DHT sensor through I2C (Only DHT12)
         /// </summary>
-        /// <param name="sensor">I2C Device, like UnixI2cDevice or Windows10I2cDevice</param>
-        public DhtSensor(I2cDevice sensor)
+        /// <param name="i2cDevice">>The I2C device used for communication.</param>
```
krwq: ditto

src/devices/Ds1307/Ds1307.cs
```diff
@@ -27,10 +27,10 @@ public class Ds1307 : IDisposable
         /// <summary>
         /// Creates a new instance of the DS1307
         /// </summary>
-        /// <param name="sensor">I2C Device, like UnixI2cDevice or Windows10I2cDevice</param>
-        public Ds1307(I2cDevice sensor)
+        /// <param name="i2cDevice">>The I2C device used for communication.</param>
```
krwq: ditto

src/devices/Bh1750fvi/Bh1750fvi.cs
shaggygi: Hmmmm.  I must of copied paste something common.  Lemme check

src/devices/Ds3231/Ds3231.cs
```diff
@@ -33,10 +33,10 @@ public class Ds3231 : IDisposable
         /// <summary>
         /// Creates a new instance of the DS3231
         /// </summary>
-        /// <param name="sensor">I2C Device, like UnixI2cDevice or Windows10I2cDevice</param>
-        public Ds3231(I2cDevice sensor)
+        /// <param name="i2cDevice">>The I2C device used for communication.</param>
```
krwq: ditto (check everywhere, not commenting anymore)

shaggygi:
@krwq ok.  one more time 😄 I found a few others along the way.

src/devices/Ads1115/Ads1115.cs
```diff
@@ -168,10 +168,10 @@ public double RawToVoltage(short val)
         /// </summary>
         public void Dispose()
         {
-            if (_sensor != null)
+            if (_i2cDevice != null)
             {
-                _sensor.Dispose();
-                _sensor = null;
+                _i2cDevice.Dispose();
```
joperezr: since you are already touching this, can we move this as we have it in the rest of the repo:

```csharp
_i2cDevice??.Dispose();
_i2cDevice = null;
```

krwq:
@shaggygi we will need to fix the build as part of this PR so if needed you'll need to do SPI as well

src/devices/Ads1115/Ads1115.cs
shaggygi: Sure.  Gimme a minute.  I found a couple others.

joperezr:
> @shaggygi we will need to fix the build as part of this PR so if needed you'll need to do SPI as well

The build is not failing because of SPI. It is failing because the bindings package references the live build version of GPIO, I will point out which changes are required in order to fix the CI Build

src/System.Device.Gpio/System/Device/I2c/Devices/UnixI2cDevice.Linux.cs
```diff
@@ -5,12 +5,12 @@
...
 using System.Runtime.InteropServices;
 
-namespace System.Device.I2c.Devices
+namespace System.Device.I2c
 {
     /// <summary>
     /// Represents an I2C communication channel running on Unix.
     /// </summary>
-    public class UnixI2cDevice : I2cDevice
+    internal class UnixI2cDevice : I2cDevice
```
joperezr: We don't need an internal class here. Instead what we need is simply to move all of this code under I2CDevice.Linux.cs. Sorry if that didn't make sense before. Ideally what we want is not to need to have extra internal classes to shell out to, instead we just want to have one I2CDevice class, which in Linux has the copied implementation from the previous UnixI2CDevice, and in windows it uses what used to be Windows10I2cDevice. Does that make sense?

shaggygi: Sure.  Lemme review.  Man, y'all are making sweat here 🏃 😓

krwq: @shaggygi don't do anything yet, we're talking offline about this right now

joperezr: Hey @shaggygi sorry for randomizing you, So I think we want to talk a bit more about the comment I gave and maybe have a small API review for it. In the meantime, I'm fine with taking this as is, except that we should fix the CI Build, I'll let you know how to do that.

shaggygi: @joperezr no prob.  i was expecting some tap dancing with this breaking change 😄 i'll check back in a few.

joperezr:
What you will want to change in order to fix the BuildPackages build, you will need to change these lines:

https://github.com/dotnet/iot/blob/80b8fddadc47491c3fc8aeb260b35281c3bce3cc/src/Iot.Device.Bindings/Iot.Device.Bindings.csproj#L21-L26

to instead look like:

```xml
  <ItemGroup>
    <PackageReference Include="System.Drawing.Common" Version="4.6.0-preview4.19164.7" />
    <PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
  </ItemGroup>
```

As for the other build breaks that you started getting, seems like you want to remove readonly from some of the changes that you updated for the Dispose method. In fact, I'd be fine if you want to just remove e941613 entirely from your PR so that we only deal with the namespace change here.

shaggygi:
@joperezr No problem and was actually fun.  If CI is green the SPI PR is a few seconds away 👍

PR #545, opened 2019-07-01
Rename SPI Structure

### LINKED ISSUE #119: [Proposal] SpiController and ISpiController

### LINKED ISSUE #534: Unify platform specific implementations of I2cDevice/SpiDevice into one class

shaggygi:
This addresses the SPI updates for moving to Devices folder and updating namespaces..

Resolves #534, resolves #119


shaggygi:
@krwq I plan to circle back and update the namespaces for bindings after this PR.

PR #552, opened 2019-07-03
Corrected wording

shaggygi:
Overlooked previously

### DIFF
--- a/src/System.Device.Gpio/System/Device/Spi/Devices/SpiDevice.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Spi/Devices/SpiDevice.Linux.cs
@@ -5,7 +5,7 @@
 namespace System.Device.Spi
 {
     /// <summary>
-    /// Represents an SPI communication channel running on Unix.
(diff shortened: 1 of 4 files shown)


krwq:
@shaggygi wasn't previous version correct? Usually you use `an` when you can hear the vowel and SPI is pronounces as `ES`-`PEE`-`I`

krwq:
Worst is that even wikipedia is not consistent:
https://en.wikipedia.org/wiki/Serial_Peripheral_Interface

> In-system programmable AVR controllers (including blank ones) can be programmed using **a SPI** interface.

> **An SPI** host adapter lets the user play the role of a master on **an SPI** bus directly from a PC.

shaggygi:
Good point.  I guess it depends on how you say the word (or acronym).  I usually spell the letters when saying I2C ( I square C).  For SPI, I say the word (like Spy) instead of letters ( S P O).

Confusing language sometimes 🤕

krwq:
anyway, at least we are consistent now :P

Issue #63, opened 2018-12-02
[Proposal] Move device projects to respective vendor folder


shaggygi:
/cc @joperezr @richlander 

I'm currently reviewing and trying to clean up some of the code for the Mcp3008 project and think it would be a good idea to come up with some better structuring regarding IoT devices folder.  This might be a good topic to add in doc related to #55.

If we begin to have a "rich set" of many component bindings, it could get messy quick.  I propose we at least have a `vendor` (or similar) folder that separate components.  Yes, some manufacturers produce the same component sometimes, but most are differentiated and the links/specs usually come from the creator of said components.

For example, the MCP3008 is a Microchip component.  Therefore, it would be good to place its related projects (API, tests, etc.) under the `dotnet\iot\devices\vendor\Microchip` (or similar) folder.

I've seen a few places where peeps reference components like Adafruit [component] for example.  This is not necessarily correct as Adrafruit is a distributor, but provides [code examples](https://www.adafruit.com/product/856).  There might be times they (Adafruit, Sparkfun, etc.) create their own component (maybe a combination of other components).  In this case, there would be a specific folder for their components.  Using #12 as an example, we could have a  `dotnet\iot\devices\vendor\Adafruit\PiOLED` folder.

To elaborate a little further...  Adrafruit offers a [breakout board](https://www.adafruit.com/product/2717) for the [TCA9548A](https://cdn-shop.adafruit.com/datasheets/tca9548a.pdf).  This component is produced by Texas Instrument.  The breakout board is basically an easy way for hobbyists to connect to controller boards.  So in this case, we could have `dotnet\iot\devices\vendor\TexasInstruments\TCA9548A` (or similar) folder.  This API should be able to work for Adrafruit's breakout board and where others use said component in other hardware boards.

One problem with labeling to specific vendor folders is [acquisitions do occur from time to time](https://www.microchip.com/investor/Announcement/MCHP%20to%20Acquire%20Microsemi%20Corporation.030118.pdf).  However, transitions are usually long and most of the time they keep the branding.  Engineers are usually smart enough to know the difference and how to find specs.  If the vendors' name is not included in project namespace, it shouldn't be a breaking change in code... just a new folder location in repo. 

Does this make sense?  What are thoughts around this?  I know I'm not thinking of all scenarios.  Just trying to collect some ideas to make a better ecosystem related to this repo while it's still young.

Thx

joshfree:
[...]

Issue #418, opened 2019-05-12
Azure DTDL & IoT Plug and Plug Support


shaggygi:
There were recent announcements regarding Azure's new Digital Twin Definition Language (DTDL) and [IoT Plug and Play](https://github.com/Azure/IoTPlugandPlay) at Build 2019.  We should review and collaborate with that team as this repo and APIs could "plug and play" nicely with each other.

There are scenarios that can help with System.Device.* tooling and code generation for device bindings.  This would also be extremely helpful to better define bindings as this has been a huge gap/struggle trying to enforce consistent standards between contributors.  Device GPIO pin assignments, I2C/SPI settings, and register definitions are just a few examples where it would benefit.  In addition, I could also see these types of DTDL docs being used for [external configuration in containers](https://github.com/dotnet/iot/issues/8).

I already created a few issues on the [IoTPlugAndPlay repo](https://github.com/Azure/IoTPlugandPlay/issues) as I see some limitations related to DTDL.  For example, it doesn't currently [support unsigned data types](https://github.com/Azure/IoTPlugandPlay/issues/8).

It might be worth doing an exercise of defining a few or our current bindings using DTDL and see what issues/ideas/innovation can be derived.

Thoughts?

//cc @joperezr @joshfree  @richlander

joperezr:
Thanks for logging this @shaggygi. In fact we have worked closely with the Azure IoT team and I believe that this is definitely something we want to work closely with them. We can use this issue for planning what the work on our side would look like, and also to gather the ideas of what we would need from the Azure team's side.

shaggygi:
@shaggygi Sounds good.  Until then, I'll keep adding issues to their repo on thoughts.  I might create a temp repo under my name to play around with some formats.  If so, I'll post link here for you to chime in.  I'd rather do that to not clutter this repo.

Also, if you get wind of [early bits](https://github.com/Azure/IoTPlugandPlay/issues/6)... send them my way so I can tinker around with them.  Interesting stuff 🎉

shaggygi:
Just a side note... if Microsoft wants to get serious about this topic, it should try to put some weight behind the support for [Common Data Model (CDM)](https://github.com/microsoft/CDM) and [Open Data Initiative (ODI)](https://www.microsoft.com/en-us/open-data-initiative).  It seems like a better bigger picture and appears like the [IoT Plug and Play DTDL](https://github.com/Azure/IoTPlugandPlay) should be based on CDM format instead.

PR #554, opened 2019-07-03
Categorized device listing

krwq:
Contributes to: https://github.com/dotnet/iot/issues/126
Contributes to: https://github.com/dotnet/iot/issues/63

This is a first pass on the categorized device listing. I've quickly added categories to all devices.

We need to figure out which categories do we want and which ones to display.

Please take a look at devices such as BNO055 (it's present in multiple categories)


shaggygi:
Referencing [this issue](https://github.com/dotnet/iot/issues/63#issuecomment-455824649) since it is somewhat related.  Not sure if it helps.

shaggygi:
I like the simplicity with the category.txt for tagging.  I'm wondering if the files should be in the [Digital Twin Definition Language](https://github.com/Azure/IoTPlugandPlay/tree/master/DTDL) format.

While this makes it a little more complicated to structure/read compared to just the category text name, it would allow for more information.  Future tooling/SDKs/APIs provided by [IoT Plug and Play](https://github.com/Azure/IoTPlugandPlay) would help improve this support.

In addition, it might be an introduction to help learn more about that modeling language for doing other advanced features.  For example, let's say I'm building a module that has a Beagle Bone Black connected to a few different binding types... like an I/O expander (via I2C), a motor driver (via PWM) and RTC (via SPI).  We could reference those respective DTDL files as a starting place, update the module's DTDL file with how the devices are connected (i.e. defined pins, comm ports, etc.).  And if I understand how DTDL works with Azure IoT, just upload the module's file to begin monitor the connected sensors.  **Disclaimer**: Yes, I know this is simplified 😄 

Another future goal is to offer code generation for bindings.  While I like having the flexibility to develop functionality around a binding's component (for specific registers and other datasheet stuff), DTDL might offer a way to produce boilerplate code based on the provided metadata.  This would hopefully keep some things standardized between the bindings (e.g. private fields, constructors, property names, disposing, etc.).

#418 is related to this topic.  If you think using DTDL is something worth investigating, I can try to provide some example DTDL files based on a few current bindings.  Thoughts?

shaggygi:
Here's another data point related to structuring devices.  It includes some categories and drivers for components.

https://github.com/torvalds/linux/tree/master/drivers/iio

joperezr:
This looks good except for two comments I have:

1. It would be great if we can move the flat list to some other file(Device-Index.md perhaps) or remove it entirely as I think the categorized list is good enough.
2. It would be great if in the list we expand families, so instead of having one entry for Bmx280, have one entry for Bme280 and one for Bmp280. Also related, can we do different categories on devices that belong to the same family of bindings?

It would be good if @richlander takes a look at this.

krwq:
@shaggygi DTDL looks interesting and I think we should investigate it but I don't think I'll address it in this PR. For linux source code - unfortunatelly it's GPL so can't look at it.
Would be useful to create an issue about this after this PR is merged and you can share all the details you have over there.

@joperezr I'll not address 2. in this PR as I'd need to do more digging which I don't have cycles for and it possibly is already covered by @shaggygi's suggestion.

krwq:
@joperezr seems like everything which can be quickly addressed already is, can we merge this?

PR #558, opened 2019-07-03
Update README.md

krwq:
Fixes: https://github.com/dotnet/iot/issues/524

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -29,7 +29,7 @@ You need to use at least [.NET Core 2.1](https://www.microsoft.com/net/download/
 
 Many of these samples use the [Raspberry Pi](https://www.raspberrypi.org/), however, .NET Core can be used with other devices. A [Raspberry Pi Starter Pack](https://www.adafruit.com/product/3058) contains enough electronics to get started on many projects.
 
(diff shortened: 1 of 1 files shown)


samples/README.md
```diff
@@ -29,7 +29,7 @@ You need to use at least [.NET Core 2.1](https://www.microsoft.com/net/download/
 
 Many of these samples use the [Raspberry Pi](https://www.raspberrypi.org/), however, .NET Core can be used with other devices. A [Raspberry Pi Starter Pack](https://www.adafruit.com/product/3058) contains enough electronics to get started on many projects.
 
-.NET Core is supported on Raspberry Pi 2B and 3A+/B/B+ models (ARMv7/v8). .NET Core is not supported on Raspberry Pi Zero or Arduino.
+.NET Core is supported on Raspberry Pi 2B as well as 3 and 4 models (ARMv7/v8). .NET Core is not supported on Raspberry Pi Zero or Arduino.
```
joperezr: Is there any Model 3 where we are not supported? or are those the only three there are? I thought there was a 3A in which we didn't run

krwq: Model 3A is not listed here: https://www.raspberrypi.org/products/
I believe we work on all of them. cc: @shaggygi, do you know of any 3s which don't work with IoT?

shaggygi: There wasn't a 3A.  Only A+/B/B+.

joperezr: I see, in that case I think this looks good.

Issue #557, opened 2019-07-03
Ensure RaspberryPi4 fix works on ARM64 OS


krwq:
Follow up on https://github.com/dotnet/iot/pull/540#discussion_r299150881

I.e. Ubuntu can be used for ARM64 testing:
https://ubuntu.com/download/iot/raspberry-pi-2-3 

cc: @Frankenslag

Frankenslag:
I have been looking at this already. The short answer is that the fix works on Ubuntu and I have it running the Ssd1351 sample. However the code that was changed for #540 is not used on Ubuntu as the detection logic doesn't recognise the hardware as a Pi because /proc/cpuinfo doesn't show the hardware details.

The testing I did was using ToInt64 rather than ToInt32. I think that you were right and we should use the 64 bit variant as it seems to work for both pointer sizes.

Any thoughts on how I should proceed ?

krwq:
@Frankenslag is there any way you can override the logic to think it's PI? (i.e. use the driver directly)
If you've tested that code path and work fine I'd leave it

If we can't hit this code path and we're not sure I think it might be safer to go with Int64 option

Frankenslag:
Yes... task for. Tomorrow.

Frankenslag:
@krwq I have tested. If we force the use of the RaspberryPi3Driver whilst running Ubuntu/arm64 then `IntPtr.ToInt32()` will throw an overflow exception. Using the `Toint64()` as the basis of the comparison works for Ubuntu on a Pi3B and Raspbian Buster on a Pi3B and a Pi4B. I intend to raise a PR for the use of `ToInt64()`

krwq:
This is already fixed with #560

PR #560, opened 2019-07-04
Ensure RaspberryPi4 fix works on ARM64 OS

Frankenslag:
@krwq, @joperezr  On ensuring that the RaspberryPi4 fix works on ARM64 OS #557 then I found that if the use of the RaspberryPi3Driver was 'forced' then an overflow exception occurred as a 64 bit IntPtr was trying to be converted to  32 bits. (I should have looked in MSDN).

This PR uses 64 bit comparison of the IntPtr. I have tested that the fix works and also that the software fails correctly in the following environments :

- Raspbian Buster on a Pi 4B (arm32)
- Raspbian Buster on a Pi 3B (arm32)
- Ubuntu Server on a Pi 3B (arm64)

The issue I found is unlikely to be encountered in normal operation but is probably worthwhile doing as it feels more correct and is more futureproof.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -365,7 +365,7 @@ namespace System.Device.Gpio.Drivers
                 }
 
                 IntPtr mapPointer = Interop.mmap(IntPtr.Zero, Environment.SystemPageSize, (MemoryMappedProtections.PROT_READ | MemoryMappedProtections.PROT_WRITE), MemoryMappedFlags.MAP_SHARED, fileDescriptor, GpioRegisterOffset);
-                if (mapPointer.ToInt32() == -1)
+                if (mapPointer.ToInt64() == -1)
                 {
                     throw new IOException($"Error {Marshal.GetLastWin32Error()} initializing the Gpio driver.");
                 }


src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -363,13 +363,13 @@ private void Initialize()
                 }
 
                 int fileDescriptor = Interop.open(GpioMemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
-                if (fileDescriptor < 0)
+                if (fileDescriptor == -1)
```
krwq: you seem to have merge conflict with yourself, rebasing should auto-fix it

Frankenslag: @krwq  Rebased... and looks ok. Tests have failed though. I am not convinced the test failure is anything to do with this PR however.

p.s. as a newbie all this git stuff is a black box. is there documentation anywhere on how the project workflows ?

eng/Version.Details.xml
```diff
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="System.Device.Gpio" Version="0.1.0-prerelease.19352.2">
+    <Dependency Name="System.Device.Gpio" Version="0.1.0-prerelease.19353.3">
```
krwq: I don't think you should be changing versions in this PR. cc: @joperezr

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: I think your rebase should be a bit different:
make sure you use your remote name which points to `github.com/dotnet/iot` (`git remote -v`). Assuming you remote name is `origin`:

```
git fetch --all
git rebase origin/master
```

you will probably see some issues at this point, you will need to use `git mergetool` or some notepad to fix the issues. Might be easier to:

```
git clean -fdx
git reset --hard origin/master
# redo your change by hand
git commit -am "Fix ARM64 issue"
git push RemoteNameWithYourFork -f
```

eng/Version.Details.xml
joperezr: Yup this is wrong. You shouldn’t change this in your PR. Looks like you may have includded other changes in your rebase.

krwq: @joperezr interestingly the version is higher

Frankenslag: Well this is a puzzle for me. I synced my local master with a fetch and merge with your master. Then I rebased the change branch onto the end of the master branch. After that I pushed to my forked change.

The changes to the versions already appear in your master and so not sure why they are showing and history shows they came from another PR.

Anyway it might be quicker to just cancel this PR right now as I can create start again with little effort. What do you think ?

krwq: @Frankenslag force pushing to your branch is equivalent to starting new PR

joperezr:
Thanks a lot @Frankenslag!

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -365,7 +365,7 @@ private void Initialize()
                 }
 
                 IntPtr mapPointer = Interop.mmap(IntPtr.Zero, Environment.SystemPageSize, (MemoryMappedProtections.PROT_READ | MemoryMappedProtections.PROT_WRITE), MemoryMappedFlags.MAP_SHARED, fileDescriptor, GpioRegisterOffset);
-                if (mapPointer.ToInt32() == -1)
+                if (mapPointer.ToInt64() == -1)
```
tmds: fwiw, an alternative way to write this would be: `mapPointer == new IntPtr(-1)`

Issue #231, opened 2019-02-14
IOT device tests should use MOQ for testing


JeremyKuhne:
Using a mocking framework will make tests more robust and easier to follow as frameworks allow easier expressing of intent over making manual mocks.

MOQ is the standard: https://github.com/moq/moq4

Starting Guidelines:

1. _Prefer_ to use MOQ to generate mocks
2. _Prefer_ strict mode mocking (calls to non mocked API throw)

@krwq, @joperezr, @shaggygi

JeremyKuhne:
Note: In my first attempt to use this I can't seem to get MOQ to work with `Span<>`. I'm not finding any info on how to set that up. The generated proxy throws an `InvalidProgramException` with `Cannot create boxed ByRef-like values.` when trying to mock `SpiDevice` and call `Write()` on it.

joshfree:
@ahsonkhan

Ellerbach:
We have quite some examples in the tests where I2C, SPI devices are mocked. That may be enough as, correct, Span<> can't be mocked.

krwq:
[Triage] We should talk about this more and decide on the direction

krwq:
[Triage] let's close this - it's not actionable for now

PR #561, opened 2019-07-04
Restructure GPIO Tests

shaggygi:
These changes attempt to mimic the core API structure for better test management as future sections are added.  This included a few new folders, file moves and namespace updates.


shaggygi:
@joperezr  Thanks for clarifying related info.  What if we left same namespaces, but stored in related folders?  Would that break some of the conventions, as well?  If so, I can go ahead and close out this PR.

joperezr:
Storing in different folders should be fine. Currently we only have gpio driver tests so moving them might be overkill, but I’m fine with moving them as well.

shaggygi:
@joperezr I think with just the GPIO isn't enough to justify for now.  I'll go ahead and close for now and maybe separate out when adding future tests.

Related topic... it would be beneficial to gather thoughts for a strategy on testing in various ways.  Just a few thoughts....

1. #231
2. Document explaining how things are testing with CI/Helix.
3. A simple test fixture for dev boards users could build on their own (if motivated) to test locally.
4. Some pointers for binding devs when creating/updating bindings.
5. Other?

joperezr:
Great feedback! Regarding your points:

1. Having MOQ tests for device bindings would be great, and perhaps even some of the Gpio library tests could also use it but we do also want to make sure that we keep running hardware tests as in the end we are still a library very tightly-coupled with the hardware.
2. Great point, I can certainly add more info on how CI/Helix is hooked up and shed some light into how the whole process works.
3. Great point! Today the boards only have literally two wires on them because we are only doing GPIO testing, but we will eventually want to do I2c and Spi testing as well so the wiring of the board will start getting more complex.
4. I agree there is not much info here, we only have very few bindings that actually added tests. Incorporating MOQ (or a similar framework) to the repo would be great progress in order to encourage binding authors to add tests.

PR #564, opened 2019-07-04
Random cleaning of SpiConnectionSettings

shaggygi:
Removed private fields & moved to default property values, added spaces, and a misspelling in a pear tree.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
+++ b/src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
@@ -42,7 +42,7 @@ namespace System.Device.Spi
         /// <summary>
         /// The length of the data to be transfered.
         /// </summary>
-        public int DataBitLength { get; set; } = 8;
+        public int DataBitLength { get; set; } = 8;  // 1 byte
 
         /// <summary>
         /// The frequency in which the data will be transferred.
--- a/src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
+++ b/src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
@@ -11,10 +11,6 @@ namespace System.Device.Spi
     /// </summary>
     public sealed class SpiConnectionSettings
     {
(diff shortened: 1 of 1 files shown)


src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
```diff
@@ -11,10 +11,6 @@ namespace System.Device.Spi
     /// </summary>
     public sealed class SpiConnectionSettings
```
krwq: I wonder if we should rename this to `HardwareSpiSettings`, some of these options don't apply for SoftwareSpi

shaggygi: They are kind of a mixture.  Better yet... how about just SpiSettings (also I2cSettings)?

src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
```diff
@@ -26,35 +22,38 @@ public SpiConnectionSettings(int busId, int chipSelectLine)
...
+        /// <summary>
+        /// The SPI mode being used.
+        /// </summary>
+        public SpiMode Mode { get; set; } = SpiMode.Mode0;
+
         /// <summary>
         /// The length of the data to be transfered.
         /// </summary>
-        public int DataBitLength { get; set; }
+        public int DataBitLength { get; set; } = 8;
```
joperezr: NIT: can we also have the comment moved mentioning that this represents 1 byte?

src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
joperezr: I don't think we should rename this. At least not to HardwareSpiSettings, SpiSettings might be an option, but honestly I don't think that the current names are wrong, plus a good thing is that they are the same in WinRT, so having parity may be better.

joperezr:
Thanks @shaggygi

PR #568, opened 2019-07-05
Switching symbols to be generated as snupkg to be uploaded to NuGet

joperezr:
Adding the generation of .snupkg to the library so that they get pushed to NuGet's symbol server.

cc: @ericstj @richlander

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -17,6 +17,10 @@
 
   <!-- Packaging Targets -->
   <PropertyGroup>
+    <!-- Generate snupkg package -->
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+    <!-- Collect package assets from builds -->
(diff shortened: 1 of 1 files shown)


Directory.Build.targets
```diff
@@ -17,6 +17,10 @@
 
   <!-- Packaging Targets -->
   <PropertyGroup>
+    <!-- Generate snupkg package -->
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
krwq: is this normal extension, not something like `symbols.nupkg`?

joperezr: We currently have `symbols.nupkg` but NuGet recently added support for portable pdbs via new packages called `.snupkg`s. Here is some additional info: https://devblogs.microsoft.com/nuget/improved-package-debugging-experience-with-the-nuget-org-symbol-server/

Basically the current guidance is that if you have a library that ships as a package in NuGet.org, you would want to also use NuGet's new .snupkg feature and push the packages there as well, for a better debugging experience.

Issue #463, opened 2019-05-30
Is the ICommand interface necessary for display bindings.


Frankenslag:
I am writing a binding for the Ssd1351 display controller and have recently submitted a PR #442 for the same that is being reviewed by @joperezr  and @krwq  so far.

My intent was to follow the way that @shaggygi engineered the Ssd1306 binding where an ICommand interface was used and then classes created for each command to be submitted like below

`device.SendCommand(new SetContrastABC(0xC8, 0x80, 0xC8));`

I have had feedback from @krwq with the opinion that we would be better using just methods and properties to give the following calling convention

`device.SetContrastABC(0xC8, 0x80, 0xC8);`

I do prefer the second example but as I don't know the reasoning behind using the first way of invoking a command on the device then I would like to open this up for discussion.

What opinions do others have ?

EDIT by @krwq:
- [ ] For 3.0: clarify guidelines
- [ ] Future: refactoring

shaggygi:
I'm on the fence for this approach.  I thought it was better at the time for separation, testing, and possible future simulations.  Although, we could get around this when we focus on #231.

My vote is to add the preferred method either in [Contributing to Binding](https://github.com/dotnet/iot/tree/master/src/devices#contributing-a-binding) or [Coding Guidelines](https://github.com/dotnet/iot/blob/master/Documentation/Coding-guidelines.md) sooner than later as it will create inconsistencies or refactoring work.

shaggygi:
Mcp25xxx is another binding where we used interfaces.  See [IRegister](https://github.com/dotnet/iot/blob/26e1b07557ce650d3110769262a27b50f38650be/src/devices/Mcp25xxx/Register/IRegister.cs).

In the case with this binding, there were dozens of registers.  Many registers were structured differently making up the bits of a byte that represented the register.   It seemed more appropriate to pass as an interface with a method signature of ToByte() so the [WriteByte method](https://github.com/dotnet/iot/blob/26e1b07557ce650d3110769262a27b50f38650be/src/devices/Mcp25xxx/Mcp25xxx.cs#L266) could easily convert/send to binding via protocol.

I'm open/flexible to ideas to help future bindings.

krwq:
I'm marking this as 3.0 issue but scope for 3.0 is only clarifying the guidelines and not the refactoring.

Frankenslag:
[...]

PR #575, opened 2019-07-06
Add the code and assets for the Ssd1351 Binding.

Frankenslag:
Here is the PR for the proposed Ssd1351 binding.

This was previously PR #442 but got lost in in the change for arm64.

It is based on the comments against #442.

image


krwq:
@Frankenslag, is this device by any chance any similar to https://github.com/dotnet/iot/tree/master/src/devices/Ssd13xx ? Is it possible to merge them?

Frankenslag:
> @Frankenslag, is this device by any chance any similar to https://github.com/dotnet/iot/tree/master/src/devices/Ssd13xx ? Is it possible to merge them?

@krwq 
There are a number of significant differences....

- This is an SPI device but the others are I2C.
- This device uses a write command to address the display memory whereas the other SSD devices stuff the image data into registers (the image field is much smaller)

My estimation is that the complexity of having a single device with logic to determine what functions to call etc. Is much greater than creating two devices...

src/devices/Ssd1351/Ssd1351.csproj
```diff
@@ -0,0 +1,17 @@
...
+    <!--Disabling default items so samples source won't get build by the main library-->
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="Command\*.cs" />
+    <Compile Include="*.cs" />
+    <None Include="README.md" />
+    <PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
+    <PackageReference Include="System.Drawing.Common" Version="4.6.0-preview4.19164.7" />
```
joperezr: NIT: you don't have to do this now, but I will later make sure that we move the version of System.Drawing.Common into Versions.props so that all of our bindings refer to the same version of it.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,308 @@
...
+
+using Iot.Device.Ssd1351.Command;
+
+namespace Iot.Device.Ssd1351
+{
+    /// <summary>
+    /// A single-chip CMOS OLED/PLED driver with controller for organic/polymer
+    /// light emitting diode dot-matrix graphic display system. 
+    /// </summary>
+    public partial class Ssd1351 : IDisposable
```
joperezr: Is there a particular reason why you decided to separate the implementation of the driver into two different files?

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,308 @@
...
+        /// <summary>
+        /// Initializes new instance of Ssd1351 device that will communicate using SPI bus.
+        /// A single-chip CMOS OLED/PLED driver with controller for organic/polymer
+        /// light emitting diode dot-matrix graphic display system. 
+        /// </summary>
+        /// <param name="spiDevice">>The SPI device used for communication.</param>
+        /// <param name="dcPinId">>The id of the GPIO pin used to control the DC line (data/command).</param>
+        /// <param name="resetPinId">>The id of the GPIO pin used to control the /RESET line (data/command).</param>
+        /// <param name="spiBufferSize">>The size of the SPI buffer. If data larger than the buffer is sent then it is split up into multiple transmissions. The default value is 4096.</param>
+        public Ssd1351(SpiDevice spiDevice, int dcPinId, int resetPinId, int spiBufferSize = DefaultSPIBufferSize)
```
joperezr: NIT: can we call them `dataCommandPin` and `resetPin` instead?

joperezr: Would it make sense to also take in a IGpioController? That way if dcPin and resetPin are located in a GpioExpander device this binding will still support it. If no IGpioController is provided, then you can initialize a new GpioController.

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: Yes... There was a discussion somewhere on what image functions and imaging libraries should be used as part of a device writing out graphics. I don't think that this has been grounded out but I have seen others use System.Drawing and so I have used it here.

Given this I wanted to separate the code that writes to the device from anything that deals with bitmaps etc and hence the two files.

src/devices/Ssd1351/Ssd1351.csproj
Frankenslag: OK. Please be aware the other bindings also use System.Common.Drawing.

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: I have made this change and an updated and others from the comments above. PR amended.

Frankenslag: Done ...

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,326 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.Gpio;
+using System.Device.Spi;
+using System.Drawing;
+using System.Threading.Tasks;
+
+using Iot.Device.Ssd1351.Command;
```
joperezr: NIT: Extra line above.

Frankenslag: Line removed in latest PR update

src/devices/Ssd1351/ssd1351Imaging.cs
```diff
@@ -0,0 +1,145 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
```
joperezr: NIT: looks like this file name needs the first letter capitalized.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+        /// <summary>
+        /// Initializes new instance of Ssd1351 device that will communicate using SPI bus.
+        /// A single-chip CMOS OLED/PLED driver with controller for organic/polymer
+        /// light emitting diode dot-matrix graphic display system. 
+        /// </summary>
+        /// <param name="spiDevice">The SPI device used for communication.</param>
+        /// <param name="dataCommandPin">The id of the GPIO pin used to control the DC line (data/command).</param>
+        /// <param name="resetPin">The id of the GPIO pin used to control the /RESET line (data/command).</param>
+        /// <param name="spiBufferSize">The size of the SPI buffer. If data larger than the buffer is sent then it is split up into multiple transmissions. The default value is 4096.</param>
+        public Ssd1351(SpiDevice spiDevice, int dataCommandPin, int resetPin, int spiBufferSize = DefaultSPIBufferSize) : this(spiDevice, new GpioController(), dataCommandPin, resetPin, spiBufferSize)
```
joperezr: Instead of having two constructors, just keep one and move the IGpioController parameter to the end of the parameter list in the other constructor and default it to null. Inside the constructor, when this param is null then initialize a new GpioController and set _disposeGpioController to true.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+        private const byte SCREEN_HEIGHT_PX = 128;
+        private const int DefaultSPIBufferSize = 0x1000;
+
+        /// <summary>
+        /// Initializes new instance of Ssd1351 device that will communicate using SPI bus.
+        /// A single-chip CMOS OLED/PLED driver with controller for organic/polymer
+        /// light emitting diode dot-matrix graphic display system. 
+        /// </summary>
+        /// <param name="spiDevice">The SPI device used for communication.</param>
+        /// <param name="gpioController">The GPIO controller used for communication.</param>
```
joperezr: I would actually mention in the description that this controller is the one that provides dataCommandPin and resetPin with their appropiate paramrefs so that it is obvious what this parameter is.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+        {
+            //as a default GPIO controller has been used then dispose of it later..
+            _disposeGpioController = true;
+        }
+
+        /// <summary>
+        /// Convert a color structure to a byte tuple represening the colour in 565 format.
+        /// </summary>
+        /// <param name="color">The color to be converted.</param>
+        private (byte, byte) Color565(Color color)
```
joperezr: You don't have to do it now, but it bight be worth it to move this method in common as a helper, since I know that several bindings would benefit from this. I know that at least the SenseHat binding is also doing this.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+            // combine with the 6 MSB if the red or blue value
+            retval |= (UInt16)((_colorSequence == ColorSequence.ABC ? color.R : color.B) >> 3);
+            return ((byte)(retval >> 8), (byte)(retval & 0xFF));
+        }
+
+        /// <summary>
+        /// Send filled rectangle to the ssd1351 display.
+        /// </summary>
+        /// <param name="color">The color to fill the rectangle with.</param>
+        /// <param name="x">The x co-ordinate of the point to start the rectangle at.</param>
```
joperezr: Is there a range for both the x and y coordinate possible values so that they still fit on the screen? Also, what is the unit of width and height, pixels? Could we add a <remarks> section on the xml docs here to answer this two questions?

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+        /// <param name="y">The y co-ordinate of the point to start the rectangle at.</param>
+        /// <param name="w">The width of the rectangle.</param>
+        /// <param name="h">The height of the rectangle.</param>
+        public void FillRect(Color color, byte x, byte y, byte w, byte h)
+        {
+
+            byte[] colourBytes = new byte[_colorDepth == ColorDepth.ColourDepth65K ? 2 : 3]; // create a short array that holds the colour data to be sent to the display
```
joperezr: Instead of all of the allocations for byte arrays that we are doing here, we should instead just use Span<byte> in order to not stress out the GC if it is not necessary. For that I mean
```c#
Span<byte> colourBytes = stackalloc byte[_colorDepth == ColorDepth.ColourDepth65K ? 2 : 3];
```

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+            byte[] aryTmp = new byte[w * h * (_colorDepth == ColorDepth.ColourDepth65K ? 2 : 3)]; // array used to form the data to be written out to the SPI interface
```
joperezr: NIT: It is a bit weird here that aryTmp variable is the one we pass in to SendCommand. Not a big deal but it should probably be renamed.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+
+        }
+
+        /// <summary>
+        /// Send a command to the the display controller.
+        /// </summary>
+        /// <param name="command">Command to send.</param>
+        public void SendCommand(Ssd1351Command command)
+        {
+            SendSPI(new byte[] { (byte)command }, true);
```
joperezr: same comment about using Span<byte> instead of byte arrays.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        /// <summary>
+        /// Send a command to the the display controller.
+        /// </summary>
+        /// <param name="command">Command to send.</param>
+        /// <param name="data">Byte array to send as parameters for the command.</param>
+        public void SendCommand(Ssd1351Command command, byte[] data)
```
joperezr: In order to be able to apply the feedback I gave about the use of Spans, you should change this method to take in Span<byte> instead of byte[].

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+            }
+
+        }
+
+        /// <summary>
+        /// Send a command to the the display controller.
+        /// </summary>
+        /// <param name="data">Byte array to send as a combination of command in the first byte
+        /// and following with parameters for the command.</param>
+        public void SendCommand(byte[] data)
```
joperezr: Change to Span.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+            {
+                throw new ArgumentNullException(nameof(data));
+            }
+
+            if (data.Length == 0)
+            {
+                throw new ArgumentException("Data array is empty", nameof(data));
+            }
+
+            SendCommand((Ssd1351Command)data[0], buffer.Slice(1).ToArray());
```
joperezr: When you move to take in spans, you won't need the ToArray() call here which is just boxing and allocating an extra copy of the array.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        /// <summary>
+        /// Send data to the display controller.
+        /// </summary>
+        /// <param name="data">The data to send to the display controller.</param>
+        public void SendData(byte[] data)
+        {
+            const int StackThreshold = 512;
```
joperezr: Do you mind explaining why did we pick this constant as the stack threshold?

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,325 @@
...
+
+            if (data == null)
+            {
+                throw new ArgumentNullException(nameof(data));
+            }
+
+            Span<byte> buffer = data.Length < StackThreshold ? stackalloc byte[data.Length + 1] : new byte[data.Length + 1];
+
+            buffer[0] = 0x40; // Control byte.
+            data.CopyTo(buffer.Slice(1));
```
joperezr: using spans everywhere will also save you from having to do this call.

src/devices/Ssd1351/Command/Commands.cs
```diff
@@ -0,0 +1,62 @@
...
+        SetDisplayEnhancement = 0xB2,
+        SetGreyLevels = 0xB8,
+        SetDefaultGreyLevels = 0xB9,
+        SetPrechargeLevel = 0xBB,
+        SetHorizontalScroll = 0x96,
+        StopScroll = 0x9E,
+        StartScroll = 0x9F,
+    }
+
+    public interface ICommand
```
joperezr: There is already an issue about using ICommands [https://github.com/dotnet/iot/issues/463](https://github.com/dotnet/iot/issues/463) (In fact looks like you logged it :P)

I actually agree with you and @krwq in there of not using this as I don't like all of the extra boxing that this entails, plus other concerns like versioning the ICommand interface or having to new up a command every time you want to send one command. Perhaps all commands should just be structs(so that they don't have to allocate) and just return value tuples which are the bytes. What do you think?

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: @joperezr 

Well... I have not done this as the Ssd1351 device can either use RGB565 or BGR565 and so this function is more complex than the one used in the sense hat. Perhaps this should be done when we have more display examples.

src/devices/Ssd1351/ssd1351Imaging.cs
Frankenslag: @joperezr 
Renamed in the latest PR update

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: @joperezr 
Uggh. I always like to have the controller close to the pins it controls in the parameter list... Anyway I have made the change you suggest in the lates PR update

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: @joperezr 
Done in the latest PR update

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: @joperezr 
Done in the latest PR update. Also had a look at most of the code for issues like this

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: @joperezr 
Done in the latest PR update

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: @joperezr 
Done in the latest PR update

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: @joperezr 

I can't remember right now but it I cant think of why this should be needed after reading one of the MS blogs so I have removed the conditional stackalloc and now just use it in all cases in the latest PR updated

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: @joperezr 

Not sure about this ? I am essentially adding a byte to the beginning of a byte[] How else could I do this ?

src/devices/Ssd1351/Command/Commands.cs
Frankenslag: @joperezr 
I have replaced the ICommand methodology with method calls on the binding object for the latest PR update. This makes more sense to me right now

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: @joperezr 

Done in the latest PR

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: Have added the information to the updated PR

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: Made the change to the updated PR

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: Changed in the latest PR update

krwq:
@Frankenslag please let us know once this is ready for re-review (and ping if we don't do it)

Frankenslag:
> @Frankenslag please let us know once this is ready for re-review (and ping if we don't do it)

@krwq @joperezr 

Can you please review. I have gone through your remarks and comments and would hope that it's coming closer to complete. I probably need to squash the commits though.

src/devices/Ssd1351/Ssd1351.cs
joperezr: I see you are right, yeah this is not super trivial. I guess this extra copy is not ideal but it is fine for now.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,295 @@
...
+            Span<byte> displayBytes = stackalloc byte[w * h * (_colorDepth == ColorDepth.ColourDepth65K ? 2 : 3)]; // span used to form the data to be written out to the SPI interface
+
+            // set the colourbyte array to represent the fill colour 
+            if (_colorDepth == ColorDepth.ColourDepth65K )
+            {
+                (colourBytes[0], colourBytes[1]) = Color565(color);
+            }
+            else
+            {
+                colourBytes[0] = (byte)((_colorSequence == ColorSequence.ABC ? color.B : color.R) >> 2);
```
joperezr: NIT: Should ColorSequence.ABC really be BGR?

src/devices/Ssd1351/Ssd1351Commands.cs
```diff
@@ -0,0 +1,427 @@
...
+
+using System;
+using System.Drawing;
+using System.Drawing.Imaging;
+using System.Runtime.InteropServices;
+
+namespace Iot.Device.Ssd1351
+{
+
+    public enum Ssd1351Command : byte
```
joperezr: Should this enum be public? You seem to have a wrapper function for each one of this which also include param validation some times so I'm not sure if it makes sense to expose a different way of doing it as well by calling SendCommand and passing one of these.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,295 @@
...
+        {
+            return (value - start) <= (end - start);
+        }
+
+        /// <summary>
+        /// Send a command to the the display controller along with associated parameters.
+        /// </summary>
+        /// <param name="command">Command to send.</param>
+        /// <param name="commandParameters">parameteters for the command to be sent</param>
+        public void SendCommand(Ssd1351Command command, params byte[] commandParameters)
```
joperezr: Should SendCommand methods be public? Is there a lot that you can do there that you can't do with the wrappers that you provided in the Commands file?

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: Well it was ABC/CBA because that is what the datasheet described but as the API is user facing I think that BGR/RGB is correct so I have updated the PR with this.

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: I hadn't done all the commands but now I have. and so the methods are now private in the latest update.

src/devices/Ssd1351/Ssd1351Commands.cs
Frankenslag: I hadn't wrapped all the commands but now I have. and so the enum is now internal in the latest update.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,275 @@
...
+                }
+            }
+
+            // specifiy a location for the rows and columns on the display where the data is to be written
+            SetColumnAddress(x, (byte)(x + w - 1));
+            SetRowAddress(y, (byte)(y + h - 1));
+
+            // write out the pixel data
+            SendCommand(Ssd1351Command.WriteRam, displayBytes);
+
```
krwq: nit: extra enter

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,275 @@
...
+                // SPI device cannot read registers from the ssd1351 and so changes need to be captured
+                switch (command)
+                {
+                    // capture changes to the colour depth and colour sequence
+                    case Ssd1351Command.SetRemap:
+                        _colorSequence = (ColorSequence)((data[0] >> 2) & 0x01);
+                        _colorDepth = (ColorDepth)((data[0] >> 6) & 0x03);
+                        break;
+                }
+
```
krwq: nit: extra enter

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,275 @@
...
+            }
+
```
krwq: nit: extra enter

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,275 @@
...
+            // set the DC pin to indicate if the data being sent to the display is DATA or COMMAND bytes.
+            _gpioDevice.Write(_dcPinId, blnIsCommand ? PinValue.Low : PinValue.High);
+
+            // write the array of bytes to the display. (in chunks of SPI Buffer Size)
+            do
+            {
+                // calculate the amount of spi data to send in this chunk
+                len = Math.Min(data.Length - index, _spiBufferSize);
+                // send the slice of data off set by the index and of length len.
+                _spiDevice.Write(data.Slice(index, len));
```
krwq: when will this be more than one chunk?

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,275 @@
...
+
+            _spiBufferSize = spiBufferSize;
+
+            SetSegmentReMapColorDepth();
+        }
+
+        /// <summary>
+        /// Convert a color structure to a byte tuple represening the colour in 565 format.
+        /// </summary>
+        /// <param name="color">The color to be converted.</param>
```
joperezr: NIT: Do you mind adding the `<returns>`  comment to explain which order are the bytes returned as? I just ask this because I know that other bindings will want to reuse this method.

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,275 @@
...
+        private const byte SCREEN_HEIGHT_PX = 128;
+        private const int DefaultSPIBufferSize = 0x1000;
+
+        /// <summary>
+        /// Initializes new instance of Ssd1351 device that will communicate using SPI bus.
+        /// A single-chip CMOS OLED/PLED driver with controller for organic/polymer
+        /// light emitting diode dot-matrix graphic display system. 
+        /// </summary>
+        /// <param name="spiDevice">The SPI device used for communication.</param>
+        /// <param name="gpioController">The GPIO controller used for communication and controls the the <paramref name="resetPin"/> and the <paramref name="dataCommandPin"/>.</param>
```
joperezr: It would be nice to mention either here or in the remarks that if a controller is passed in, we will dispose it on Dispose()

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,275 @@
...
+            // combine with the 6 MSB if the red or blue value
+            retval |= (UInt16)((_colorSequence == ColorSequence.BGR ? color.R : color.B) >> 3);
+            return ((byte)(retval >> 8), (byte)(retval & 0xFF));
+        }
+
+        /// <summary>
+        /// Send filled rectangle to the ssd1351 display.
+        /// </summary>
+        /// <param name="color">The color to fill the rectangle with.</param>
+        /// <param name="x">The x co-ordinate of the point to start the rectangle at in pixels (0..126 for the Ssd1351).</param>
```
joperezr: Can we add a quick remarks comment that mentions which corner is x = 0 and y =0 ? I assume bottom left?

src/devices/Ssd1351/Ssd1351Imaging.cs
```diff
@@ -0,0 +1,139 @@
...
+            SendBitmapPixelData(GetBitmapPixelData(bm, new Rectangle(sourcePoint.X, sourcePoint.Y, destinationRect.Width, destinationRect.Height)), destinationRect);
+
+        }
+
+        /// <summary>
+        /// Convert a bitmap into an array of pixel data suitable for sending to the display
+        /// </summary>
+        /// <param name="bm">The bitmap to be sent to the display controller note that only Pixel Format Format32bppArgb is supported.</param>
+        /// <param name="sourceRect">A rectangle that defines where in the bitmap data is to be converted from.</param>
+        public Span<byte> GetBitmapPixelData(Bitmap bm, Rectangle sourceRect)
```
joperezr: Should this method be public? Seems like people should just call SendBitmap instead right? Or is there a scenario where it would be useful to call this method directly?

src/devices/Ssd1351/Ssd1351.cs
```diff
@@ -0,0 +1,275 @@
...
+        /// <summary>
+        /// Verifies value is within a specific range.
+        /// </summary>
+        /// <param name="value">Value to check.</param>
+        /// <param name="start">Starting value of range.</param>
+        /// <param name="end">Ending value of range.</param>
+        /// <returns>Determines if value is within range.</returns>
+        internal static bool InRange(uint value, uint start, uint end)
+        {
+            return (value - start) <= (end - start);
```
joperezr: Should this just be:
```c#
return (value >= start && value <= end);
```

src/devices/Ssd1351/Ssd1351Commands.cs
```diff
@@ -0,0 +1,541 @@
...
+        /// increment mode is enabled by command 20h, after finishing read/write one column data, it is
+        /// incremented automatically to the next column address. Whenever the column address pointer finishes
+        /// accessing the end column address, it is reset back to start column address and the row address
+        /// is incremented to the next row.  This command is only for horizontal or vertical addressing modes.
+        /// </summary>
+        /// <param name="startColumn">Column start address with a range of 0-127.</param>
+        /// <param name="endColumn">Column end address with a range of 0-127.</param>
+        public void SetColumnAddress(byte startColumn = 0x00, byte endColumn = 0x7F)
+        {
+            if (startColumn > 0x7F)
```
joperezr: You are only checking for upper bound not for lower bounds. Same goes for endColumn.

src/devices/Ssd1351/Ssd1351Commands.cs
```diff
@@ -0,0 +1,541 @@
...
+        /// The chip has 256 contrast steps from 00h to FFh.  The segment
+        /// output current ISEG increases linearly with the contrast step,
+        /// which results in brighter display.
+        /// </summary>
+        /// <param name="colorAContrast">Contrast level for color A.</param>
+        /// <param name="colorBContrast">Contrast level for color B.</param>
+        /// <param name="colorCContrast">Contrast level for color C.</param>
+        public void SetContrastABC(byte colorAContrast = 0x86, byte colorBContrast = 0x51, byte colorCContrast = 0x86)
+        {
+            SendCommand(Ssd1351Command.SetContrastABC, colorAContrast, colorBContrast, colorCContrast);
```
joperezr: Shouldn't we be doing parameter checks here?

src/devices/Ssd1351/Ssd1351Commands.cs
```diff
@@ -0,0 +1,541 @@
...
+
+        /// <summary>
+        /// This command sets the divide ratio to generate DCLK (Display Clock) from CLK and
+        /// programs the oscillator frequency Fosc that is the source of CLK if CLS pin is pulled high.
+        /// </summary>
+        /// <param name="displayClockDivideRatio">Display clock divide ratio with a range of 0-15.</param>
+        /// <param name="oscillatorFrequency">Oscillator frequency with a range of 0-15.</param>
+        public void SetDisplayClockDivideRatioOscillatorFrequency(byte displayClockDivideRatio = 0x01, byte oscillatorFrequency = 0x0D)
+        {
+            if (displayClockDivideRatio > 0x0F)
```
joperezr: Same comment about lower bound checks.

src/devices/Ssd1351/Ssd1351Commands.cs
```diff
@@ -0,0 +1,541 @@
...
+        }
+
+        /// <summary>
+        /// This command specifies the mapping of the display start line to one of COM0-COM127
+        /// (assuming that COM0 is the display start line then the display start line register is equal to 0).
+        /// </summary>
+        /// <param name="displayOffset">Display offset with a range of 0-127.</param>
+        public void SetDisplayOffset(byte displayOffset = 0x60)
+        {
+            if (displayOffset > 0x7F)
```
joperezr: lower bound check. If you want to keep it as you have it and still check lower bound for all of these, you can just cast the value you are checking to a uint. this will work as long as the lower bound is 0, because all negative numbers will overflow and become very large numbers, all larger than your upper bounds so you only need one check.

src/devices/Ssd1351/Ssd1351Commands.cs
```diff
@@ -0,0 +1,541 @@
...
+
+        /// <summary>
+        /// This command sets the Display Start Line register to determine starting address of display RAM,
+        /// by selecting a value from 0 to 127. With value equal to 0, RAM row 0 is mapped to COM0.
+        /// With value equal to 1, RAM row 1 is mapped to COM0 and so on.
+        /// </summary>
+        /// <param name="displayStartLine">Display start line with a range of 0-127.</param>
+        public void SetDisplayStartLine(byte displayStartLine = 0x00)
+        {
+            if (displayStartLine > 0x7F)
```
joperezr: I won't comment on the lower band check anymore but assume we have to fix this on all subsequent cases.

src/devices/Ssd1351/Ssd1351Commands.cs
```diff
@@ -0,0 +1,541 @@
...
+            SendCommand(Ssd1351Command.SetDisplayOffset, displayOffset);
+        }
+
+        /// <summary>
+        /// This command sets the Display Start Line register to determine starting address of display RAM,
+        /// by selecting a value from 0 to 127. With value equal to 0, RAM row 0 is mapped to COM0.
+        /// With value equal to 1, RAM row 1 is mapped to COM0 and so on.
+        /// </summary>
+        /// <param name="displayStartLine">Display start line with a range of 0-127.</param>
+        public void SetDisplayStartLine(byte displayStartLine = 0x00)
```
joperezr: For all these Set Commands, do we really want to have a default for the parameters to be set? I suppose most people that call this will want to provide a value right?

src/devices/Ssd1351/Ssd1351Commands.cs
```diff
@@ -0,0 +1,541 @@
...
+            if (displayStartLine > 0x7F)
+            {
+                throw new ArgumentException("The display start line is invalid.", nameof(displayStartLine));
+            }
+
+            SendCommand(Ssd1351Command.SetStartLine, displayStartLine);
+        }
+
+        /// <summary>
+        /// This double byte command is used to set the states of GPIO0 and GPIO1 pins
```
joperezr: which ones are pin0 and pin1? Are they DCPin and reset pin?

Frankenslag: No they are some general IO that the chipset provides for 'free'. The datasheet mentions that you should initialize these pins after a reset and so I provided a function.

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: Much of the time. Full image would be 32K (128*128*2 bytes) bytes of data ignoring the command bytes. The default for the raspbian SPI buffer is 4K, I believe, and so unless you use a non standard buffer size then chunking the output is inevitable.

src/devices/Ssd1351/Ssd1351Imaging.cs
Frankenslag: This is public as it is useful to be able to cache the bitmap data returned. This is used in the sample where a number of bitmaps are read, converted and then displayed sequentially.

src/devices/Ssd1351/Ssd1351Commands.cs
Frankenslag: The contrast values are 0->255 and are the range of the byte parameters.

src/devices/Ssd1351/Ssd1351Commands.cs
Frankenslag: The default value in the method call is the default value from the datasheet and so calling the method without a parameter becomes a way of just ensuring that the default value is set without having to actually read the datasheet.

src/devices/Ssd1351/Ssd1351Commands.cs
Frankenslag: @joperezr 
Am I misunderstanding something... these parameters are bytes and hence cannot go below 0 and so I have not checked for lower bounds. I believe I do need to check that start values are <= end values though.

src/devices/Ssd1351/Ssd1351Commands.cs
krwq: if the lower boundary is 0 then anything below will underflow and will satisfy the check

src/devices/Ssd1351/Ssd1351Commands.cs
joperezr: you are right, that was my bad I didn't see that all these were bytes. I originaly thought they were ints. You can disregard all related comments.

src/devices/Ssd1351/Ssd1351Commands.cs
joperezr: I see. Do we expect to have cases where the default won't already be set without having called any of these methods earlier? Naming might be the part that is a bit confusing, perhaps they should be named something like `SetDisplayStartLineOrDefault` or something like that.

src/devices/Ssd1351/Ssd1351.cs
Frankenslag: actually the origin is top left. I have put this information in the param tag.

src/devices/Ssd1351/Ssd1351Commands.cs
Frankenslag: I think that adding OrDefault would make the names too wordy. I have added descriptions of the default values to the param tags

Frankenslag:
> Left a few comments but they are all minor, the rest LGTM. Once last round of feedback is addressed, we can go ahead and merge. Thanks @Frankenslag

I have made the minor changes and had to force push them. It looks like the helix test is failing. Don't know why.

Frankenslag:
> So that problem has been fixed now, but your build is failing because we recently just merged a PR where IGpioController got removed from our library. In order to fix your PR, simply change the places where you took an optional IGpioController, and use a GpioController instead.

Changed in the latest PR Update.

Issue #100, opened 2018-12-15
Create System.Device.Gpio Overview doc


shaggygi:
It would be beneficial to create a **System.Device.Gpio Overview** markdown to help newbies get started along with more technical details on infrastructure.  It could include the following topics:

**NOTE**: The intent of this overview is to focus on the following:
1. Intro examples of creating a protocol connection (e.g. I2C, SPI) and instantiating the respective device (e.g. I2cDevice, SpiDevice).  This would be more focused on how it relates to interface topics (e.g. buses, addressing, etc.);  not so much on particular binding.
2. Understanding how the OS and/or specific device is determined.
3. Understanding of external bindings.
4. Other topics?

⚠️  It might be more appropriate to document certain topics in [Microsoft Docs](https://docs.microsoft.com) to keep consistent with the overall .NET ecosystem.

- [ ] Create/add diagram helper showing the structure along with sections for future work.
- [ ] Add System.Device.Gpio API folder and placeholder markdown files under [Documentation](https://github.com/dotnet/iot/tree/master/Documentation).
- [ ] Update [Documentation](https://github.com/dotnet/iot/tree/master/Documentation) with other types of interfaces.
- [ ] Update [Documentation System.Device.* APIs section](https://github.com/dotnet/iot/tree/master/Documentation#systemdevice-apis) with link to System.Device.Gpio API README markdown.
- [ ] Add link to System.Device.Gpio API README in [main repo README markdown](https://github.com/dotnet/iot/blob/master/README.md).

shaggygi:
@joperezr @joshfree another one you can assign to me.  I have some early thoughts started for this.  The plan is to have multiple PRs over time since this will cover lots of material and changes are needed as progress is made on API.

If possible, the idea is to stub out beginning markdowns and such with the intent to add content as time allows.  I'll try to add some action items above while reviewing.

joperezr:
I'm not sure of concrete workitems that can come out of this. I know that you logged this some time ago and a bunch of changes to documentation have happened since, do you think this is still an issue? Is there a specific thing we should add to our existing docs?

shaggygi:
I was originally thinking it would be good to have an overall doc that explained the basics down to more technical detail.  Sort of explaining with diagrams and such for peeps to understand structure of things like device/controller and how it fits with with the different OS/hardware boards.  I'm not sure the best format (markdown, slides, etc.), but the dust needs to settle (new talks of controllers, etc. before attempting to do so.

[...]

PR #576, opened 2019-07-07
Better abstraction for PWM: PwmChannel

### LINKED ISSUE #204: PwmController/PwmDevice should be bound to specific pin

shaggygi:
Fixes #204 

This PR is based on the new `PwmChannel` concept.  See #204 for more details.

cc:  @krwq @joperezr This only includes the code for Linux.  I wanted to make sure the feature fits what was expected before adding files for Windows.  I started some review comments to point out a few thoughts/issues found along the way.


src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,181 @@
...
+            int pwmChannel,
+            double frequency,
+            double dutyCyclePercentage = 0.5,
+            PwmPolarity polarity = PwmPolarity.ActiveHigh)
+        {
+            _pwmChip = pwmChip;
+            _pwmChannel = pwmChannel;
+            ValidatePwmChannel();
+            OpenChannel();
+            Thread.Sleep(100);  // TODO: Need a better solution to delay for available file.
```
shaggygi: I noticed the PWMx file isn't created fast enough.  I added this delay, but there needs to be a better way to check when file is available.  Thoughts?

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,181 @@
...
+            {
+                return _dutyCyclePercentage;
+            }
+            set
+            {
+                SetDutyCyclePercentage(value);
+            }
+        }
+
+        // Helpers for file paths.
```
shaggygi: These paths are called in various areas.  Is this the preferred/best approach?  In not, please provide example and will update.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,181 @@
...
+            if (!Directory.Exists(GetChannelPath()))
+            {
+                File.WriteAllText($"{GetChipPath()}/export", Convert.ToString(_pwmChannel));
+            }
+        }
+
+        /// <summary>
+        /// Starts writing to the channel.
+        /// </summary>
+        public override void Start()
```
shaggygi: The original plan was to have StartWriting/StopWriting, but it seems like Start/Stop sounds better.  It is also the same method names in the WinRT PWM API.  Linux uses Enable/Disable.  I like any of them, so let me know the names and will update.

@krwq You mentioned not having Start/Stop in the end.  We can still remove, but I thought of a scenario where a user might want to stop the PWM (without disposing) while making DC adjustments (especially if there were multiple channels) and then start back.  The channel is opened when it is instantiated and closed when disposed.

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+using System.Threading;
+
+namespace System.Device.Pwm
+{
+    internal class Windows10PwmChannel : PwmChannel
```
shaggygi: This is just a placeholder for now.  Supporting code will be added after Linux flavor has been determined.

src/System.Device.Gpio/System/Device/Pwm/PwmPolarity.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Device.Pwm
+{
+    /// <summary>
+    /// The polarity of the PWM signal.
+    /// </summary>
+    public enum PwmPolarity
```
shaggygi: This was not in the initial concept, but came across it while coding the feature..  It seems like it would be a nice options to have and there is support in both Linux/Windows.  I can remove if decided there is no use.

src/System.Device.Gpio/System/Device/Pwm/Channels/PwmChannel.Linux.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <param name="frequency">The frequency in hertz.</param>
+        /// <param name="dutyCyclePercentage">The duty cycle percentage represented as a value between 0.0 and 1.0.</param>
+        /// <param name="polarity">The polarity the channel should start in.</param>
+        /// <returns>A PWM channel running on Unix.</returns>
+        public static PwmChannel Create(
+            int pwmChip,
+            int pwmChannel,
+            double frequency,
+            double dutyCyclePercentage = 0.5,
+            PwmPolarity polarity = PwmPolarity.ActiveHigh) =>
```
krwq: what's polarity? is it just remapping duty cycle into (1-duty_cycle)?

shaggygi: Hopefully the attached image can explain.

Active High is the normal (majority use-case) where it is idled with a Low value and the pulse has a High value.

**Active High (DC = 25%)** and **Active Low (DC = 75%)** are essentially the same, but the High pulse comes at different time in the signal.

The idle state is where polarity becomes more helpful as there could be different circuit configurations.  For example, one could have a RGB LED tied to Vcc.  If the channel was stopped (or set to Active High DC = 0%), the LED would be on.  Same case for type of transistors used to drive motors and such.

PwmPolarityExample

src/System.Device.Gpio/System/Device/Pwm/Channels/PwmChannel.Linux.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace System.Device.Pwm
+{
+    public partial class PwmChannel
+    {
+        /// <summary>
+        /// Creates a PWM channel running on Unix.
+        /// </summary>
+        /// <param name="pwmChip">The PWM chip number.</param>
+        /// <param name="pwmChannel">The PWM channel number.</param>
```
shaggygi: Should `pwmChip` and `pwmChannel` just be named `chip` and `channel`?  I recall @JeremyKuhne and @bartonjs suggesting `pinNumber` for GPIO pin be renamed to just `pin` for simplicity.

src/System.Device.Gpio/System/Device/Pwm/Channels/PwmChannel.Windows.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <param name="frequency">The frequency in hertz.</param>
+        /// <param name="dutyCyclePercentage">The duty cycle percentage represented as a value between 0.0 and 1.0.</param>
+        /// <param name="polarity">The polarity the channel should start in.</param>
+        /// <returns>A PWM channel running on Windows 10 IoT.</returns>
+        public static PwmChannel Create(
+            int pwmChip,
+            int pwmChannel,
+            double frequency,
+            double dutyCyclePercentage = 0.5,
+            PwmPolarity polarity = PwmPolarity.ActiveHigh) =>
```
shaggygi: Since we are changing `PwmChannel` to be similar to `I2cDevice`/`SpiDevice`, I wonder if we should create `PwmConnectionSettings`.

```csharp
public sealed class PwmConnectionSettings(
            int pwmChip,
            int pwmChannel,
            double frequency,
            double dutyCyclePercentage = 0.5,
            PwmPolarity polarity = PwmPolarity.ActiveHigh)
```

Therefore, you could create default channels like...

```csharp
PwmConnectionSettings settings = new PwmConnectionSettings(0, 0, 2000);
var pwmChannel = PwmChannel.Create(settings);
```

Makes it a little cleaner when you have more than one PWM channel for a binding, as well.

```csharp
// RGB LED.
var rgbLed = RgbLed(
    new PwmConnectionSettings(0, 0, 2000),   // Red
    new PwmConnectionSettings(0, 1, 2000),   // Green
    new PwmConnectionSettings(0, 2, 2000));  // Blue
rgbLed.SetColor(...);

// Since it creates default channels, they would dispose with binding.
rgbLed.Dispose();

// Could also have overload for PwmChannel objects if needed.  Channels would be disposed outside of binding.
var rgbLed = RgbLed(redPwmChannel, greenPwmChannel, bluePwmChannel);
```

shaggygi:
@krwq @joperezr I'm trying to reinstall Win10 IoT on one of my RPi 3B+ and not having any luck.  You have any narrowed down steps on what to download, install, and copy to SD card?  Thx

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: Try calling https://linux.die.net/man/2/sync instead of delay and see if you can see any difference. Assuming it work `sync` might be generally to slow, so you might want to expand File.WriteAllText into FileStream+StreamWriter and call `sysfs` (or perhaps even just explicit Flush will work).

krwq: Generally to explain my theory:
- sync normally forces buffer cache to be written into disk - I suspect it might also work on virtual fs - that would force driver to process the virtual fs operation

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: Fine, although if the file handle would be open all the time (as an optimization) it would be as good if path was hard-coded in the only place where it would be used.
The only time I tried using hardware PWM I've experienced lag whenever I tried changing speed quickly - perhaps this was an issue here. (because of that I've switched to @Ellerbach's SoftPwm implementation which worked much better perhaps I'd have used hardware if this worked faster or at least the same)

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,181 @@
...
+                }
+            }
+            else
+            {
+                throw new IOException($"Unable to parse the number of supported channels at {npwmPath}.");
+            }
+        }
+
+        // In Linux, the period needs to be a whole number and can't have a decimal point.
+        private int GetPeriodInNanoSeconds(double frequency) => (int)((1.0 / frequency) * 1_000_000_000);
```
krwq: would it worked if we changed code to use Period instead of frequency to avoid using fractions? Or perhaps just change Frequency to be integer Hz - is there any advantage of using double?

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        /// <summary>
+        /// Sets the polarity for the channel to start in.
+        /// </summary>
+        /// <param name="polarity">The polarity to set.</param>
+        private void SetPolarity(PwmPolarity polarity)
```
krwq: I feel like this is something we should not have. What's the difference between this and and setting DutyCycle to `DutyCycle = 1 - DutyCycle`?

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <param name="pwmChannel">The PWM channel number.</param>
+        /// <param name="frequency">The frequency in hertz.</param>
+        /// <param name="dutyCyclePercentage">The duty cycle percentage represented as a value between 0.0 and 1.0.</param>
+        /// <param name="polarity">The polarity the channel should start in.</param>
+        public Windows10PwmChannel(
+            int pwmChip,
+            int pwmChannel,
+            double frequency,
+            double dutyCyclePercentage = 0.5,
+            PwmPolarity polarity = PwmPolarity.ActiveHigh)
```
krwq: i feel like this property is useless, it's less confusing to use `1 - DutyCycle`

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
```
krwq: Can we rename Windows10PwmChannel to GenericPwmChannel or DefaultPwmChannel and use one code on one platform and the other on other? (this way we have 2 files less)

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
krwq: not sure I follow

src/System.Device.Gpio/System/Device/Pwm/PwmPolarity.cs
krwq: I feel using `1 - DutyCycle` is less confusing, I didn't know you added, I aleady expressed my not liking of this above :P (sorry, hopefully no hard feeling)

tools/DevicesApiTester/Commands/Pwm/PwmCommand.cs
```diff
@@ -10,28 +10,19 @@ namespace DeviceApiTester.Commands.Pwm
...
-        {
-            PwmDriver pwmDriver = DriverFactory.CreateFromEnum<PwmDriver, PwmDriverType>(this.Driver);
+        [Option('d', "dutycycle", HelpText = "The duty cycle percentage for PWM output from 0.0 - 1.0.", Required = false, Default = 0.5)]
+        public double DutyCyclePercentage { get; set; }
 
-            return pwmDriver != null
-                ? new PwmController(pwmDriver)
-                : new PwmController();
-        }
+        [Option('p', "polarity", HelpText = "The polarity the channel should start in.", Required = false, Default = PwmPolarity.ActiveLow)]
```
krwq: Polarity is fine to add in the API tester if you find it useful

krwq: you can implement it as `1-DutyCycle`

src/System.Device.Gpio/System/Device/Pwm/Channels/PwmChannel.Linux.cs
krwq: how is that different from using `1-DutyCycle` except everything being shifted by 1 cycle on the chart?

src/System.Device.Gpio/System/Device/Pwm/Channels/PwmChannel.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Stops the PWM channel.
+        /// </summary>
+        public abstract void Stop();
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
```
krwq: not needed since you don't have a finalizer

bartonjs: Presumably the type is unsealed (otherwise it wouldn't need the pattern at all), so the SuppressFinalize here doesn't hurt... and does catch if a derived type adds one.

krwq: perhaps we should ask a language team to inject it in every Dispose team then

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
shaggygi: > not sure I follow

Mainly just a note-to-self and is far I got over the weekend.  Trying to setup Win10 on another RPi and also wanted to make sure the new structure looked okay before porting code for this flavor.

src/System.Device.Gpio/System/Device/Pwm/PwmPolarity.cs
shaggygi: > (sorry, hopefully no hard feelings)

Not the case at all.  That's what design/development is all about 😄.  Let me try another way to explain why I think this is useful.

I see your point on looking at it as `1 - Duty Cycle`, but don't think of this related to timing or the pulse.  Polarity is all about the state the logic levels for the signal.  You can verify by doing the following:

### Active High (default; also known as normal)

src/System.Device.Gpio/System/Device/Pwm/Channels/PwmChannel.Linux.cs
shaggygi: > how is that different from using 1-DutyCycle except everything being shifted by 1 cycle on the chart?

More details https://github.com/dotnet/iot/pull/576#discussion_r301008653

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
shaggygi: More details https://github.com/dotnet/iot/pull/576#discussion_r301008653

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
shaggygi: > is there any advantage of using double?

No, I used since it was used in original code.  I can review/change.  I like `frequency` better as most peeps will understand it versus `period`.

src/System.Device.Gpio/System/Device/Pwm/Channels/PwmChannel.cs
bartonjs: They don't like injecting things :smile:. But the pattern is

```C#
public void Dispose()
{
    Dispose(true);
    GC.SuppressFinalize(this);
}

{
    ...
}
```

src/System.Device.Gpio/System/Device/Pwm/Channels/PwmChannel.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Device.Pwm
+{
+    /// <summary>
+    /// Represents a single PWM channel.
+    /// </summary>
+    public abstract partial class PwmChannel : IDisposable
```
joperezr: NIT: This type should be under System\Device\Pwm, since we usually echo the namespace structure with the folder structure for the implementation.

shaggygi: @joperezr Shouldn't this follow similar to I2C/SPI Devices' folders?  Instead of Devices, it's Channels. I left the namespaces the same just as we did for Devices (System.Device.I2c/System.Device.Spi), as well.  I thought of placing under the Pwm folder, but Channels folder is the grouping of the different channels (and future home of SoftPwmChannel).  Plus, if we ever had to expand in future (the infamous controller, chip, or such 😄), we would be ahead of the game on separating the channels.  If this is the direction, we might need to fix I2C/SPI to be same?

joperezr: but I2cDevice and SpiDevice are still in the I2c (and Spi) root namespace, and they are the types that will be most likely used most of the time, since people will do I2cDevice.Create. Same goes here, I'm fine with having the platform specific channels inside the channels namespace, but PwmChannel should be in the root of Pwm as that one will be the one used by consumers.

shaggygi: I'm still scratching my head on this.  See attachment to compare PWM with SPI.

PwmChannel, along with all flavors of Channels, are under the Channels folder.
All flavors of Channels are under the namespace System.Device.Pwm including PwmChannel.

SpiDevice, along with all flavors of Devices, are under the Devices folder.
All flavors of Devices are under the namespace System.Device.Spi including SpiDevice.

You can call SpiDevice.Create(...).
You can call PwmChannel.Create(...).

Am I missing something?  I could be staring at the screen too long 😄 

PwmVsSpi

joperezr: I see the issue then 😄. For folder structure, we always try to mimic the namespace, so in this case, SpiDevice.cs, and PwmChannel.cs, would need to go one level up in order to be placed where they are supposed to be. I thought PwmChannel was under a Channels namespace, but I just saw that it is directly under Pwm. I'm fine if we want to keep the UnixPwmChannel and the Windows10PwmChannel under the Channels folder (as long as they are also on the Channels namespace) but the idea here is that unless this is the test project, we want the folder structure to match the namespace, does that make sense?

shaggygi: We did discuss that for I2c/Spi.  I believe it was [decided to move I2cDevice/SpiDevice to the Devices folder](https://github.com/dotnet/iot/pull/533#discussion_r298669724).

Yes, all channels are in the System.Device.Pwm namespace, but located under the Channels folder.

shaggygi: > I'm fine if we want to keep the UnixPwmChannel and the Windows10PwmChannel under the Channels folder (as long as they are also on the Channels namespace)

Okay.  I'll update PWM related in this PR, and will create another PR to move the I2C/SPI related later.

shaggygi: Just an opinion... I don't really like this file move.  PwmChannel.Linux.cs and PwmChannel.Windows.cs are partials of PwmChannel.  This means all 3 files have to go up one level.  We get back to mixture of Channel files located in different folders.  This is the same for I2cDevice.cs/I2cDevice.Linux.cs/I2cDevice.Windows.cs and Spi files, as well.

I guess having them separate isn't a biggie, just seems like all flavors would be located together.

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        public abstract void Stop();
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        public virtual void Dispose(bool disposing)
```
joperezr: This method should be protected.

src/System.Device.Gpio/System/Device/Pwm/PwmPolarity.cs
joperezr: Are there specific bindings in mind that require this sort of concept? There doesn't seem to be a concept of this in most libraries, although I do see a few that have it, like WinRT Apis for example. If we can make a case that some bindings won't work without this, I think it's fine to add but in general I think I agree with @krwq in which I don't see much point in adding this.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,201 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+using System.Threading;
+
+namespace System.Device.Pwm
```
joperezr: Given the folder structure, didn't we agree that this would be under Channels namespace as well? Not super important as it is not public facing but still might be worth it in order to follow convention.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,201 @@
...
+            int channel,
+            double frequency,
+            double dutyCyclePercentage = 0.5,
+            PwmPolarity polarity = PwmPolarity.ActiveHigh)
+        {
+            _chip = chip;
+            _channel = channel;
+            Validate();
+            Open();
+            Thread.Sleep(100);  // TODO: Need a better solution to delay for available file.
```
joperezr: How was this not a problem with the previous implementation?

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            SetFrequency(frequency);
+            DutyCyclePercentage = dutyCyclePercentage;
+
+            // NOTE: There appears to be a lower-level bug that won't allow the
+            // polarity to be changed until a valid period has been written.
```
joperezr: When you say period, I assume you mean DutyCycle? In that case, should Validate also ensure that frequency and dutycycle are valid?

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// <param name="frequency">The frequency in hertz.</param>
+        /// <returns>The frequency period in nanoseconds.</returns>
+        private static int GetPeriodInNanoSeconds(double frequency)
+        {
+            // In Linux, the period needs to be a whole number and can't have a decimal point.
+            return (int)((1.0 / frequency) * 1_000_000_000);
+        }
+
+        // Helpers for file paths.
+        private string GetChipPath() => $"/sys/class/pwm/pwmchip{_chip}";
```
joperezr: Should we just make these 3 private fields instead of methods? I don't really see the point of having them be methods.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,201 @@
...
+        private void SetPolarity(PwmPolarity polarity)
+        {
+            string polarityString = polarity switch
+            {
+                PwmPolarity.ActiveHigh => "normal",
+                PwmPolarity.ActiveLow => "inversed",
+                _ => throw new ArgumentException($"Invalid PWM polarity {polarity}.")
+            };
+
+            File.WriteAllText($"{GetChannelPath()}/polarity", Convert.ToString(polarityString));
```
joperezr: Why do you have to call Convert.ToString if polarityString is already a string?

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
shaggygi: It is updated in last commit

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
shaggygi: Yes, this was updated in last commit

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
shaggygi: oops.  copy/paste from other types.  will update

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// <summary>
+        /// Stops writing to the channel.
+        /// </summary>
+        public override void Stop()
+        {
+            string enablePath = $"{_channelPath}/enable";
+            File.WriteAllText(enablePath, "0");
+        }
+
+        public override void Dispose(bool disposing)
```
joperezr: This should be protected.

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
```diff
@@ -0,0 +1,47 @@
...
+            int chip,
+            int channel,
+            double frequency,
+            double dutyCyclePercentage = 0.5,
+            PwmPolarity polarity = PwmPolarity.ActiveHigh)
+        {
+            // TODO: This is just a placeholder to complete later.
+        }
+
+        public override double Frequency => throw new NotImplementedException();
```
joperezr: Most of the requested comments on the Unix side are NITs so I think we should be ok to start adding implementaiton of the Windows side. My only big question that could alter API would be whether or not we should remove PwmPolarity class.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
shaggygi: No.  When the RPi boots and you open/export the pwm channel (at least for the first time), the period is set to '0'.  For some reason you must change the period (of frequency) to a valid value before changing the polarity.  It should be okay since the period is set to a valid value in the SetFrequency(...) method a few lines above and is the only place polarity should be changed.  I noticed others were seeing this issue in other IoT libraries while researching this.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,201 @@
...
+
+using System.IO;
+using System.Threading;
+
+namespace System.Device.Pwm.Channels
+{
+    /// <summary>
+    /// Represents a PWM channel running on Unix.
+    /// </summary>
+    internal class UnixPwmChannel : PwmChannel
```
joperezr: @krwq do you think we should make this and WindowsPwmChannel sealed?

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
shaggygi: Thanks for the review/feedback.  I'm still digesting and will tidy up a bit before moving to the Windows flavor.

When it comes to the PwmPolarity enum, I understand it wasn't part of agreed work.  However, after diving into the code/feature, I hope my explanations (https://github.com/dotnet/iot/pull/576#discussion_r300865292 and https://github.com/dotnet/iot/pull/576#discussion_r301008653) would justify on why this is important when it comes to basic electronics.  I can remove if that is what is decided.  It shouldn't take much to remove.

src/System.Device.Gpio/System/Device/Pwm/PwmPolarity.cs
shaggygi: > Are there specific bindings in mind that require this sort of concept

This is simply a basic electronics concept for hardware configurations.  Some devices require active low or high based on design functionality or design constraints.  This works for both channels that are located on dev boards and binding components.

Here are just a few links I Googled that have the PWM polarity topic.

- https://python-periphery.readthedocs.io/en/latest/pwm.html
- https://docs.microsoft.com/en-us/uwp/api/windows.devices.pwm.pwmpulsepolarity
- https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52832.ps.v1.1%2Fpwm.html
- https://lore.kernel.org/patchwork/patch/612159/

As mentioned, it can be easily removed if decided.  If that is the path, I think we at least owe it to users when [documenting the feature](https://github.com/dotnet/iot/issues/100) not having optional polarity as it would help them determine specific types of hardware to purchase and how they configure it when using the core API.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
shaggygi: Updated in next commit.

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
shaggygi: Updated in next commit

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
shaggygi: Updated in next commit

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
shaggygi: I'm not sure.  I went back and tried with original and appears to work.  The cold is very similar, but I could be overlooking something.  I can reproduce in PwmChannel and get the following if it doesn't wait long enough.

> Error: Access to the path '/sys/class/pwm/pwmchip0/pwm0/period' is denied.

shaggygi: Note: I came across this and appears to be similar issue.

https://github.com/processing/processing/blob/37108add372272d7b1fc23d2500dce911c4d1098/java/libraries/io/src/processing/io/PWM.java#L81

joperezr: mm we should look into this a bit more, it makes me a bit uncomfortable to have a Thread.Sleep without really knowing if it is required. I suppose you are running as sudo given that this seems to be a permission issue? Maybe the file gets created right away but permissions aren't set so if not running as sudo then it doesn't work at first?

src/System.Device.Gpio/System/Device/Pwm/PwmPolarity.cs
joperezr: Looks like both in Linux and Windows this functionality is supported natively by the driver, so in that case I'm fine with keeping it.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: @shaggygi have you tried the sync solution I've written in one of the other comments?

shaggygi: @krwq no. i briefly read over, but was working on the other stuff.  i'll try to review tonight.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: i don't care, it's internal

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
krwq: @shaggygi I personally think we should not have this API and especially not as part of this work, please create an issue with all the details you put here and then we can decide on that (I'd not add it unless it gets heavily upvoted or have some good justification for it)

shaggygi: @krwq understood.  issue #581 created for future discussion.  latest commit removed the PwmPolarity related code.

joperezr:
Build is failing because of the chicken an egg problem again. To fix it, you could either do the temporary fix that we did in SPI and I2C with making the bindings package use a package reference to the main library like here:

https://github.com/dotnet/iot/blob/d6274e441c0040a70275b5acff4c3bbb625053ce/src/Iot.Device.Bindings/Iot.Device.Bindings.csproj#L21-L24

Or alternatively you could have the projects of the bindings that are failing to use the live version of System.Device.Gpio, and fix all the namespace/class issues upfront.

shaggygi:
I will be off the grid most of the next 2 weeks.  I'll try to check in every now and then if time allows.

If you can send me some docs on how to prepare and install the lastest Win10 IoT for RPi 3B+, I will try to at least prepare the device to work on the Windows10PwmChannel for when I return.
For some reason I'm having troubles getting mine configured and have not been able to troubleshoot the issue.  I've had no problems setting up the Linux flavor.

krwq:
@joperezr do you have instructions for Win10 IoT Core? I've only used Raspbian/Ubuntu with RPi

joperezr:
If that's ok with you @shaggygi I will finish this PR since we do want to take it in sooner rather than later :). As to instructions on how to set up Windows IoT core, I have usually just downloaded the [Windows IoT Core Dashboard](https://docs.microsoft.com/en-us/windows/iot-core/downloads) and follow the steps from that in order to flash my SD card.

shaggygi:
@joperezr that is fine.  Sorry on delay

shaggygi:
@joperezr that is fine.  Sorry on delay

joperezr:
> Sorry on delay

@shaggygi no worries at all, it wasn't a delay, you got us in a very great place with a head start 😉. We just want to finish this in less than two weeks and don't want to interfere at all with your plans 😄

src/devices/Buzzer/Buzzer.csproj
```diff
@@ -9,11 +9,13 @@
   <ItemGroup>
     <Compile Include="Buzzer.cs" />
     <None Include="README.md" />
-    <PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
+    <ProjectReference Include="$(MSBuildThisFileDirectory)../../System.Device.Gpio/System.Device.Gpio.csproj">
```
krwq: why is MSBuildThisFileDirectory needed?
Does `<AdditionalProperties>RuntimeIdentifier=linux</AdditionalProperties>` have to be passed? What about Windows IoT Core?

joperezr: Additional properties is required, and it will work even with Windows IoT Core. Basically this means that the surface area that we want to use is the linux one, as the Windows one has some extra base classes for its implementation. MSBuildThisFileDirectory is not strictly required, but it is convention to try to have full paths instead of relative ones. We do this in the rest of our repos as well when possible.

src/devices/DCMotor/samples/DCMotor.sample.csproj
```diff
@@ -6,9 +6,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
-    <!-- Normally you will use following package reference instead  -->
-    <!-- <PackageReference Include="IoT.Device.Bindings" Version="$(IotDeviceBindingsPackageVersion)" /> -->
+    <ProjectReference Include="..\..\..\System.Device.Gpio/System.Device.Gpio.csproj">
```
krwq: can you fix `/` and `\` here @joperezr since you're fixing merge conflicts?

joperezr: Yup I'll take care of it.

joperezr:
Code is ready after my last change, I just want to run some more tests to make sure everything is behaving as expected in both Unix and Windows.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -0,0 +1,188 @@
...
+            double dutyCyclePercentage = 0.5)
+        {
+            _chipPath = $"/sys/class/pwm/pwmchip{_chip}";
+            _channelPath = $"{_chipPath}/pwm{_channel}";
+            _chip = chip;
+            _channel = channel;
+            Validate();
+            Open();
+            //Thread.Sleep(100);  // TODO: Need a better solution to delay for available file.
+            Console.WriteLine("Set Frequency");
```
krwq: leftover Console.WriteLine

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <param name="frequency">The frequency in hertz.</param>
+        /// <param name="dutyCyclePercentage">The duty cycle percentage represented as a value between 0.0 and 1.0.</param>
+        public Windows10PwmChannel(
+            int chip,
+            int channel,
+            int frequency = 400,
+            double dutyCyclePercentage = 0.5)
+        {
+            // When running on Hummingboard we require to use the default chip.
+            var deviceInfo = new EasClientDeviceInformation();
```
krwq: `EasClientDeviceInformation` this looks like it has typo in the name

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
```diff
@@ -0,0 +1,125 @@
...
+                useDefaultChip = true;
+            }
+
+            // Open the Windows PWM controller for the specified PWM chip.
+            string deviceSelector = useDefaultChip ? WinPwm.PwmController.GetDeviceSelector() : WinPwm.PwmController.GetDeviceSelector($"PWM{chip}");
+
+            DeviceInformationCollection deviceInformationCollection = DeviceInformation.FindAllAsync(deviceSelector).WaitForCompletion();
+            if (deviceInformationCollection.Count == 0)
+            {
+                throw new ArgumentException($"No PWM device exists for PWM chip at index {chip}.", $"{nameof(chip)}");
```
krwq: `$"{nameof(chip)}")` => `nameof(chip)`

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.Linux.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <param name="chip">The PWM chip number.</param>
+        /// <param name="channel">The PWM channel number.</param>
+        /// <param name="frequency">The frequency in hertz.</param>
+        /// <param name="dutyCyclePercentage">The duty cycle percentage represented as a value between 0.0 and 1.0.</param>
+        /// <returns>A PWM channel running on Unix.</returns>
+        public static PwmChannel Create(
+            int chip,
+            int channel,
+            int frequency = 400,
+            double dutyCyclePercentage = 0.5) =>
```
krwq: I think we should have Create and CreateInternal, Create should have defaults and CreateInternal should be explicit so that we don't accidentally have different defaults on both platforms

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
joperezr: Nope, this is a WinRT class. I will adress your feedback in another change I’m about to push which also contains updates to Readme.md files and xml comments for the bindings we changed.

krwq: I've already pushed xml change too since it was breaking the build. add whatever else is needed (if possible try to make it so that we have only 3 commits in the end but honestly we can as well just merge this branch as is)

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.Linux.cs
joperezr: I think having Create is fine for now since that is what we have in other similar APIs (like I2cDevice.Create or how the GpioController picks a driver), I'll keep it as is and we can discuss later if we want to change to CreateInternal

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -90,6 +89,10 @@ private static int GetPeriodInNanoSeconds(int frequency)
         /// <param name="frequency">The frequency in hertz to set.</param>
         private void SetFrequency(int frequency)
         {
+            if (frequency < 0)
```
krwq: I think we should throw also for 0, unless that means "disabled"

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -90,6 +89,10 @@ private static int GetPeriodInNanoSeconds(int frequency)
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(frequency), frequency, "Value must not be negative.");
+            }
```
krwq: nit: enter

src/devices/SoftPwm/README.md
```diff
@@ -4,78 +4,51 @@ Software PWM is necessary when you cannot use hardware PWM. PWM is used for exam
...
+You can then create a software PWM like this:
 
 ```csharp
-var PwmController = new PwmController(new SoftPwm());
+// Creates a Software Pwm channel on GPIO pin 17 with a frequency of 200 hertz an initial duty cycle of 0%
+var softwarePwmChannel = new SoftwarePwmChannel(17, 200, 0);
 ```
 
-By default SoftPwm is using a low priority clock to emulate the PWM. It is ok if you have leds and other non time sensitive elements attached.
+By default SoftarePwm is using a low priority clock to emulate the PWM. It is ok if you have leds and other non time sensitive elements attached.
```
krwq: typo

krwq: also it's SoftwarePwmChannel

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -33,27 +33,49 @@ public class SoftwarePwmChannel : PwmChannel
...
+        /// </summary>
         public override int Frequency
         {
             get => _frequency;
             set
             {
+                if (value < 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(value), value, "Value must note be negative.");
+                }
```
krwq: ditto for both value <= 0 and enter

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
joperezr: by default in Linux the value is 0 so I don't want to throw in case of 0.

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
joperezr: Given they are internal we don't have to have the extra files so I don't mind having one called Windows and the other Unix.

joperezr:
Thanks @krwq and @shaggygi for the group effort on this 😄

shaggygi:
@joperezr @krwq thanks for finishing this up while I was on vacation.  I have a few devices on order and will try to tinker/test around with results when time allows. 👍

microhobby:
Great work guys, by the way this solve a multi channel access issue that was in the old code  ... thanks

`if (pwmChip < 0 || pwmChip >= numSupportedChannels)
                {
                    throw new ArgumentException($"The PWM chip {pwmChip} does not support the channel {pwmChannel}.");
                }
`

PR #579, opened 2019-07-09
initial commit TM1637-segment display

Ellerbach:
Adding TM1637
* Docuementation
* Code to display characters, raw data
* Adjusting brightness, display on/off
* Changing orders of segments


src/devices/Tm1637/Character.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Pre build character to be displayed including the potential dot
+    /// Please note that depending on your LCD display you may not have the dot present
+    /// </summary>
+    public class Character
+    {
+        public Display Char { get; set; }
+        public bool Dot { get; set; }
+
```
krwq: nit: extra enter

src/devices/Tm1637/DelayHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using System.Diagnostics;
+using System.Threading;
+
+namespace System.Device
+{
+    /// <summary>
+    /// Helpers for short waits.
+    /// </summary>
+    internal static class DelayHelper
```
krwq: this should already exist somewhere in the Common folder, is there any diff between this and that version?

src/devices/Tm1637/Display.cs
```diff
@@ -0,0 +1,35 @@
...
+        Car7 = 7,
+        Car8 = 8,
+        Car9 = 9,
+        CarA = 10,
+        CarB = 11,
+        CarC = 12,
+        CarD = 13,
+        CarE = 14,
+        CarF = 15,
+        CarNothin = 16,
```
krwq: Nothing

src/devices/Tm1637/Display.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Text;
+
+namespace Iot.Device.Tm1637
+{
+    /// <summary>
+    /// Pre build characters that can be displayed
+    /// </summary>
+    public enum Display
+    {
+        Car0 = 0,
```
krwq: Should this be `Char0` (and similarly below?)

src/devices/Tm1637/README.md
```diff
@@ -0,0 +1,105 @@
...
+## Device family
+
+You can find this display as [Grove](http://wiki.seeedstudio.com/Grove-4-Digit_Display/) elements as well as very cheap with no brand. A search on your favorite online shop will give you lots of options. Those simple displays are used a lot for simple clock for example. 
+
+## Usage
+
+You need to create a ```Tm1637``` class with 2 pins, the clock one and the data one. 
+
+```csharp
+Tm1637 tm1637 = new Tm1637(21, 20);
```
krwq: there's non-zero chance this will work with SPI (DIN=>MISO, MOSI unconnected)

src/devices/Tm1637/samples/Tm1637.sample.cs
```diff
@@ -0,0 +1,81 @@
...
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine("Hello Tm1637!");
+            Tm1637 tm1637 = new Tm1637(20, 21);
+            tm1637.Brightness = 7;
+            tm1637.ScreenOn = true;
+            tm1637.ClearDisplay();
+            // Displays 4 haracters
```
krwq: nit: characters

src/devices/Tm1637/DelayHelper.cs
Ellerbach: Absolutely, forgot to remove the file and add ```< Compile Include="../Common/Device/DelayHelpler.cs">``` into the project file

src/devices/Tm1637/Display.cs
Ellerbach: yes, will fix.

src/devices/Tm1637/README.md
Ellerbach: Will try once the PR will go thru. Only thing that may not work with SPI is that there is a read of the status so at some point, the state of the DIO pin is changed to input to read the status
```chsarp
// Wait 1 µs, it's the waiting time between clk up and down
// That's according to the documentation
DelayHelper.DelayMicroseconds(ClockWidthMicroseconds, true);

var ack = _controller.Read(_pinDio);
if (ack == PinValue.Low)
{
        _controller.SetPinMode(_pinDio, PinMode.Output);
        _controller.Write(_pinDio, PinValue.Low);
}
```

src/devices/Tm1637/SegmentDisplay.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// bit 2 = c      f   b
+    /// bit 3 = d      |_g_|
+    /// bit 4 = e      |   |
+    /// bit 5 = f      e   c
+    /// bit 6 = g      |_d_|  .dp
+    /// bit 7 = dp
+    /// 
+    /// </summary>
+    [Flags]
+    public enum SegmentDisplay : byte
```
krwq: This should be called Character

src/devices/Tm1637/SegmentDisplay.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// </summary>
+        SegmentMiddle = 0b0100_0000,
+        /// <summary>
+        /// Segment dp
+        /// </summary>
+        Dot = 0b1000_0000,
+        /// <summary>
+        /// Character 0
+        /// </summary>
+        Char0 = SegmentTop | SegmentTopRight | SegmentBottomRight | SegmentBottom | SegmentBottomLeft | SegmentTopLeft,
```
krwq: Digit0, Digit1, ...

src/devices/Tm1637/SegmentDisplay.cs
```diff
@@ -0,0 +1,129 @@
...
+        E = SegmentMiddle | SegmentTopLeft | SegmentBottomLeft | SegmentBottom | SegmentTop,
+        /// <summary>
+        /// Character F
+        /// </summary>
+        F = SegmentMiddle | SegmentTopLeft | SegmentBottomLeft | SegmentTop,
+        /// <summary>
+        /// Character -
+        /// </summary>
+        Minus = SegmentMiddle,
+
```
krwq: nit: extra enter

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,305 @@
...
+                    throw new ArgumentException($"Size of {nameof(SegmentOrder)} can only be 6 length");
+                // Check if we have all values from 0 to 5
+                bool allExist = true;
+                for (int i = 0; i < MaxSegments; i++)
+                {
+                    allExist &= Array.Exists(value, e => e == i);
+                }
+                if (!allExist)
+                    throw new ArgumentException($"{nameof(SegmentOrder)} needs to have all existing segments from 0 to 5");
+                _segmentOrder = value;
```
krwq: this should copy, otherwise changing the array after setting will affect how this device works

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,305 @@
...
+        /// <summary>
+        /// Set the screen on or off
+        /// </summary>
+        public bool ScreenOn
+        {
+            get { return _secrrenOn; }
+
+            set
+            {
+                _secrrenOn = value;
```
krwq: typo

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,305 @@
...
+                _controller.SetPinMode(_pinDio, PinMode.Output);
+                _controller.Write(_pinDio, PinValue.Low);
+            }
+
+            _controller.Write(_pinClk, PinValue.High);
+            DelayHelper.DelayMicroseconds(ClockWidthMicroseconds, true);
+            _controller.Write(_pinClk, PinValue.Low);
+            DelayHelper.DelayMicroseconds(ClockWidthMicroseconds, true);
+
+            // Wait 50 Microseconds
```
krwq: commented out code, here and function below

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,299 @@
...
+        /// 0 to 5, any order. Most of the time 4 segments do
+        /// not need to be changed but the 6 ones may be in different order
+        /// like 0 1 2 5 4 3. In this case, this byte array has be be in this order
+        /// </summary>
+        public byte[] SegmentOrder
+        {
+            get { return _segmentOrder; }
+            set
+            {
+                if (value.Length != MaxSegments)
```
joperezr: does it have to be exactly 6? Shouldn't I be allowed to pass a byte[] of size 3 for example if I only really want to print 3 digits?

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,299 @@
...
+        private byte[] _lastDisplay = new byte[6] { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
+        private bool _screenOn;
+
+        /// <summary>
+        /// Initialize a TM1637
+        /// </summary>
+        /// <param name="pinClk">The clock pin</param>
+        /// <param name="pinDio">The data pin</param>
+        /// <param name="pinNumberingScheme">Use the logical or physical pin layout</param>
+        public Tm1637(int pinClk, int pinDio, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical)
```
joperezr: This should also take in an optional IGpioController, so that if one gets passed in, then that one is used for locating the clock and data pin.

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,299 @@
...
+            }
+        }
+
+        private PinValue WriteByte(byte data)
+        {
+            // We send data by 8 bits
+            for (byte i = 0; i < 8; i++)
+            {
+
+                _controller.Write(_pinClk, PinValue.Low);
```
joperezr: NIT: Extra empty line above.

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,299 @@
...
+            _controller.Write(_pinClk, PinValue.High);
+            DelayHelper.DelayMicroseconds(ClockWidthMicroseconds, true);
+            _controller.SetPinMode(_pinDio, PinMode.Input);
+
+            // Wait 1 µs, it's the waiting time between clk up and down
+            // That's according to the documentation
+            DelayHelper.DelayMicroseconds(ClockWidthMicroseconds, true);
+
+            var ack = _controller.Read(_pinDio);
+            if (ack == PinValue.Low)
```
joperezr: Can you add a comment of what is going on here? If we get a PinValue.High does that mean we didn't get an acknowledge from the device?

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,299 @@
...
+using System.Device;
+using System.Device.Gpio;
+using System.Diagnostics;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.WindowsRuntime;
+
+namespace Iot.Device.Tm1637
+{
+    public class Tm1637 : IDisposable
```
joperezr: Class should probably be sealed.

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,299 @@
...
+        /// bit 4 = e      |   |
+        /// bit 5 = f      e   c
+        /// bit 6 = g      |_d_|  .dp
+        /// bit 7 = dp
+        /// 
+        /// Representation of the number 0 so lighting segments a, b, c, d, e and F is then 0x3f
+        /// </summary>
+        /// <param name="segmentAddress">The segment address from 0 to 6</param>
+        /// <param name="rawData">The rsegemnets to display</param>
+        public void Display(byte segmentAddress, Character rawData)
```
joperezr: Given that segmentAddress refers to a numbered position, I would change this so that it takes an int instead (and add validation so that the number is from 0 to 6) and change the name of the parameter to be something like segmentPosition or something. The remarks should also say that this segmentPosition starts at index 0.

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,299 @@
...
+        /// bit 2 = c      f   b
+        /// bit 3 = d      |_g_|
+        /// bit 4 = e      |   |
+        /// bit 5 = f      e   c
+        /// bit 6 = g      |_d_|  .dp
+        /// bit 7 = dp
+        /// 
+        /// Representation of the number 0 so lighting segments a, b, c, d, e and F is then 0x3f
+        /// </summary>
+        /// <param name="segmentAddress">The segment address from 0 to 6</param>
```
joperezr: typo: the address would really be from 0 to 5 right?

src/devices/Tm1637/Tm1637.cs
```diff
@@ -0,0 +1,299 @@
...
+            _controller.Write(_pinDio, PinValue.Low);
+            DelayHelper.DelayMicroseconds(ClockWidthMicroseconds, true);
+            _controller.Write(_pinClk, PinValue.High);
+            _controller.Write(_pinDio, PinValue.High);
+        }
+
+        /// <summary>
+        /// Displays segments starting at first segment with byte array containing raw data for each segment including the dot
+        /// 
+        /// bit 0 = a       _a_
```
joperezr: Perhaps this table should be in the <remarks> section of the comment?

src/devices/Tm1637/Tm1637.cs
Ellerbach: Yes, for this one. Because you can have hardware where only the segments 3 to 5 are connected. So if you want them to be visible, you'll have to setup the full table anyway. 
What I've seen buying couple of those ones is that the ones with 4 segments used for clocks are usually all ok, 0 to 3 are usualy corresponding. So in those cases, you just Don't have to use this at all.
For the 6 segments ones, it's either all ok either order is 2 1 0 5 4 3.
And finally when you have only 3 segements, they are mostly ok but I've seen as well 5 4 3 as the first order. 
Finally, if in the case you have 3 segments, and in the case they are inversted in some ways, you'll pass Something like 5 4 3 0 1 2 where actually you Don't care or the real order of the last 3 ones.

src/devices/Tm1637/Tm1637.cs
Ellerbach: Yes, that is what is happening. So concretely, you will miss a character or have a character shift. 
The device can act as a keyboard reader as well. This feature is mainly used when you are trying to read the keys. I did not implement this feature as it's mostly not used in hardware setup I found.

src/devices/Tm1637/Tm1637.cs
Ellerbach: right, totally missed this one!

src/devices/Tm1637/Tm1637.cs
Ellerbach: ok to rename but what is the advantage of changing the type from byte to int in this case? With byte, you're at least sure it starts at 0. And I'm anyway testing if we have more than the maximum number of 6 segments.

src/devices/Tm1637/Tm1637.cs
Ellerbach: What do you mean by:

> Perhaps this table should be in the  section of the comment?

joperezr: sorry I meant the `<remarks>` section of the comment.

src/devices/Tm1637/Tm1637.csproj
```diff
@@ -0,0 +1,14 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <Compile Include="..\Common\System\Device\DelayHelper.cs" />
+    <PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
```
krwq: this (and sample proj) needs to be adjusted to use project to project reference (copy from different project)

PR #587, opened 2019-07-10
Adding System.Device.Gpio Symbols to the Snupkg

joperezr:
We are now generating .Snupkgs that will be pushed to NuGet along with our libraries but they didn't contain any pdbs yet given our custom build. This target will add them to the package.

cc: @ericstj

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -20,6 +20,7 @@
     <!-- Generate snupkg package -->
     <IncludeSymbols>true</IncludeSymbols>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+    <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);_GetSymbolsForSnupkg</GenerateNuspecDependsOn>
     <!-- Collect package assets from builds -->
     <IncludeRIDSpecificBuildOutput Condition="'$(IncludeRIDSpecificBuildOutput)' == '' And '$(RuntimeIdentifiers)' != ''">true</IncludeRIDSpecificBuildOutput>
     <TargetsForTfmSpecificContentInPackage Condition="'$(IncludeRIDSpecificBuildOutput)' == 'true'">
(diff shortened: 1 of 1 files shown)


Directory.Build.targets
```diff
@@ -33,6 +34,28 @@
...
+      <Output
+          TaskParameter="TargetOutputs"
+          ItemName="_DebugSymbolsWithTfm" />
+    </MSBuild>
+    <PropertyGroup>
+      <!-- Need to set this back to True so that the snupkg will contain symbols. -->
+      <IncludeBuildOutput>true</IncludeBuildOutput>
+    </PropertyGroup>
+    <ItemGroup>
+      <_TargetPathsToSymbols Include="@(_DebugSymbolsWithTfm)">
```
ericstj: Could you have instead just added these to the `RIDSpecificPackageFile` item?

joperezr: I believe that wouldn't work because if I did that then they would be packaged in the .nupkg and not in the .snupkg. `_TargetPathsToSymbols` is the Item that gets populated with the pdbs that you want to go to the .snupkg but not the .nupkg. Another annoying thing is that with this Item, you can't really specify a `PackagePath`, so basically this item only hast two metadatas that will decide where will the pdb be placed in the .snupkg. Pdbs will be placed at lib/%(_TargetPathsToSymbols.TargetFramework)/%(_TargetPathsToSymbols.Identity) and there doesn't seem to be  a way of overriding this. That is why for now I'm using the hack of appending the RID into the TargetFramework on the next line, so that the Windows and Linux pdbs won't collide.

ericstj: This is pretty ugly, can you file an issue on nuget to enable a better story?

joperezr: Will do.

joperezr:
Looks like .snupkg is not the right way to approach this. Instead we will push to symbol servers as part of official build using our .symbols.nupkg packages as corefx does. I will submit a PR soon with those changes.

Issue #580, opened 2019-07-09
Add support for /dev/mem to the RaspberryPiDriver


joperezr:
When testing our support for arm64, I discovered that the Debian available for arm64 doesn't have a /dev/gpiomem memory mapped file which is what the RaspberryPiDriver uses today in order to get access to the registers. Instead, it only has the /dev/mem memory mapped file, which should provide access to such registers. We should add the capability to the RaspberryPi Driver that in case we can't find /dev/gpiomem/ we should be able to just use the /dev/mem file instead. That would make the RaspberryPiDriver work for arm64 on this specific distro.

Frankenslag:
@joperezr  I would like to have an informal look at this. Did you get the image from https://wiki.debian.org/RaspberryPi3 ?

joperezr:
@Frankenslag Awesome! yes, more specifically the image I got was from: `https://people.debian.org/~gwolf/raspberrypi/20190628/20190628_raspberry-pi-3_buster_PREVIEW.img.xz`. That said, you shouldn't need that to prototype this since /dev/mem is also available in regular Raspbian arm32 images. We initially picked /dev/gpiomem because of two reasons: 1) You only get access to the gpio related registers, which contained all we need at the time, and mor importantly 2) It didn't require to run as sudo, whereas /dev/mem does. This means that when using the RaspberryPi drivier today, your app doesn't necessarily need to run as sudo which would change with /dev/mem. The ideal implementation, would use /dev/gpiomem when available, and fallback to /dev/mem if not found. I did start adding this support some time ago but never finished, here is a pointer to my partial work in case it helps: https://github.com/joperezr/corefxlab/blob/970f4a57047e53ab22ec972e303307c0afca745d/src/System.Devices.Gpio/RaspberryPiDriver.cs

Frankenslag:
@joperezr OK. Will look into it. I have just created a Debian installation to play with so I'll probably start with that. As a side issue (and I have only spent 5 minutes on it) but it seems that the I2C devices are not working for whatever reason and I can't get an SPI device to show in /dev/spi*. Having said all that I want to finish off the preliminary work on #559

joperezr:
It does work, you only need to make sure you enable both Spi and I2c protocols. In Raspbian doing this is very straight forward as you can just do `sudo raspi-config` and do it from there, but Debian doesn't have such an easy access. I just did this for I2c, so I know that all you have to do is to add 
```
i2c-bcm2708
i2c-dev
```

[...]

PR #595, opened 2019-07-14
Addition of allowing writes to /dev/dem if /dev/gpiomem is not available

Frankenslag:
This is a [WIP] PR for comments for issue #580 

I have tested this on a PI4b and  when I force it to use /dev/mem it seems to work. I am having problems with 64bit Debian right now as it doesn't seem to work (I get EPERM) when I try and do the mmap to /dev/mem.

I have added detection of the physical peripheral address and also amended the detection of a RaspberryPi as Debian (and ubuntu I think) on a pi would be seen as standard linux otherwise

Please let me know your comments


joperezr:
This is awesome @Frankenslag I will take a look at this very soon. Thanks for putting it up.

Frankenslag:
Thanks @joperezr It's worthwhile adding some comments...

I have relaxed the RaspberryPi detection logic to read the model from the /proc/device-tree/model in addition to the current logic as Pis running Debian and Ubuntu were not being detected. 

I have had a bit of a think and now only fail over to /dev/mem when /dev/gpiomem doesn't exist. If there is a permissions issue then I believe that an exception should be thrown as normal.

Status of testing:-

Pi4B, Pi3B using Raspbian Buster O.K. uses /dev/gpiomem. I used the Ssd1351 sample. If I take the user out of the gpio group then it fails on permissions.
Pi3B Ubuntu 64Bit O.K. and uses /dev/mem when forced to by misnaming /dev/gpiomem in the source. I used the Ssd1351 sample. I want to do a little more testing of this though.
Pi3B Debian 64Bit O.K.  and uses /dev/mem but couldn't get SPI working and so just used a blinking light test.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -30,6 +31,17 @@ public GpioController(PinNumberingScheme numberingScheme)
         /// <returns>A driver that works with the board the program is executing on.</returns>
         private static GpioDriver GetBestDriverForBoard()
         {
+            // a first chance of finding a raspberry pi..... perhaps not on Raspbian note 
+            // that we want this to fail quietly if something goes wrong with the test
+            try
+            {
+                if (File.ReadAllText(ModelInfoPath).StartsWith("Raspberry Pi"))
```
joperezr: Why did you have to add this? was the current way of getting the right device not picking the Raspberry Pi Driver for the Pi 4? I would have expect the logic that we have after this with the Regex to have worked unless they changed their hardware which is very possible.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -13,11 +13,17 @@ namespace System.Device.Gpio.Drivers
...
     {
+
+        // error indicates that an entity doesn't exist
+        private const int Linux_Enoent = 2;
+
         private RegisterView* _registerViewPointer = null;
-        private const int GpioRegisterOffset = 0x00;
         private static readonly object s_initializationLock = new object();
         private static readonly object s_sysFsInitializationLock = new object();
+        private const uint GpioPeripheralOffset = 0x00200000; // offset from the peripheral base address of the GPIO registers
```
joperezr: NIT: can you use digit separators like `0x200_000`

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,51 @@ protected ulong ClearRegister
             set {  *(ulong*)(_registerViewPointer->GPCLR) = value; }
         }
 
+        /// <summary>
+        /// Returns the pperipheral base address of the raspberry pi based on the ranges set within the device tree.
+        /// The return value is 32bit even on 64 bit operating systems (debian / ubuntu tested) but may change in the future
+        /// 
+        /// Based on file at https://github.com/RPi-Distro/raspi-gpio/blob/master/raspi-gpio.c
```
joperezr: I'm not an expert on this, but looks like this License is not exactly MIT so I would wait for @RichardLander to confirm if we can use this as a reference or not.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,51 @@ protected ulong ClearRegister
...
+                // if the next range is 0 then assume that the base address is in the next 32 bits so read again.
+                if ((buffer[0] | buffer[1] | buffer[2] | buffer[3]) == 0)
+                {
+                    buffer = rdr.ReadBytes(4);
+                }
+
+                // convert the read data into big endian
+                retval = (uint)(buffer[0] << 24) | (uint)(buffer[1] << 16) | (uint)(buffer[2] << 8) | buffer[3];
+
+                rdr.Close();
```
krwq: why close it just before it's disposed?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -358,13 +412,40 @@ private void Initialize()
...
+                        }
+
+                        // add on the offset from the peripheral base address to point to the gpio registers
+                        gpioRegisterOffset += GpioPeripheralOffset;
+                    }
 
-                IntPtr mapPointer = Interop.mmap(IntPtr.Zero, Environment.SystemPageSize, (MemoryMappedProtections.PROT_READ | MemoryMappedProtections.PROT_WRITE), MemoryMappedFlags.MAP_SHARED, fileDescriptor, GpioRegisterOffset);
+                }
+                    
+                IntPtr mapPointer = Interop.mmap(IntPtr.Zero, Environment.SystemPageSize, (MemoryMappedProtections.PROT_READ | MemoryMappedProtections.PROT_WRITE), MemoryMappedFlags.MAP_SHARED, fileDescriptor, (int)gpioRegisterOffset);
```
krwq: could this be using https://docs.microsoft.com/en-us/dotnet/api/system.io.memorymappedfiles.memorymappedfile?view=netcore-3.0 instead?

Frankenslag: @krwq  @joperezr 

Perhaps but I think that this would be out of scope for this PR as here we are trying to ensure that Gpio works with /dev/men as a fallback but to move away form interop for we would need to ask some wider questions about the merits of both methods.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: @joperezr @richlander 

I haven’t copied the code directly the reference to the code is more to indicate that we are determining the base address in a way that is consistent with the way that raspi-bipolar does it.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
Frankenslag: It would only pick the raspberry pi for Raspbian. Debian does not include the chipset in /proc/cpuinfo (this may also be the case for Ubunto but I cannot remember right now). Reading the device tree allows detection of the pi regardless of Linux distribution. I did decide to leave the original detection code in as I didn't want to break something that was currently working but I was unaware of.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Addressed in latest PR

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Addressed in latest PR

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
joperezr: I see, good to know. Actually it may very well be that your way of reading the device tree is better so we might want to switch our main logic to use this regardless of the /dev/mem work. That said, I'm curious if it would catch the case when running in an old RaspberryPi like the Pi1. That one I believe doesn't work with the RaspberryPi Driver so we needed to fallback to SysFs.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
joperezr: I'm fine with keeping this Interop for now, and then we can move to using memorymappedfiles in the future as a separate issue.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
Frankenslag: Is the Pi 1 supported ? The docs for core seem to indicate that it isn't i.e.

> Note: All models of generation 1 and Pi Zero are not supported because the .NET Core JIT depends on armv7 instructions not available on those versions.

Perhaps I could find a Pi 2

joperezr: oh right that's a good point. I'll do some tests of this in our end as well, but I guess it's fine to keep it in your PR for now in order to get Debian support using this driver. Does Debian have /dev/gpiomem?

Frankenslag: No, Debian appears to just have /dev/mem. At the moment it seems very unfinished but was useful for testing Arm64 et. al.

Frankenslag: I have tested on a Pi 2 and it works as expected using /dev/gpiomem

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -358,13 +410,40 @@ private void Initialize()
...
+                // try and open /dev/gpiomem
+                fileDescriptor = Interop.open(GpioMemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
                 if (fileDescriptor == -1)
                 {
-                    throw new IOException($"Error {Marshal.GetLastWin32Error()} initializing the Gpio driver.");
-                }
+                    // if the failure is NOT because /dev/gpiomem doesn't exist then throw an exception at this point.
+                    // if it were anything else then it is probably best not to try and use /dev/mem on the basis that
+                    // it would be better to solve the issue rather than use a method that requires root privileges
+                    if(Marshal.GetLastWin32Error() != Linux_Enoent)
```
joperezr: We should probably cache the GetLastWin32Error response so that we don't call it again in the next line when formatting the string.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,49 @@ protected ulong ClearRegister
             set {  *(ulong*)(_registerViewPointer->GPCLR) = value; }
         }
 
+        /// <summary>
+        /// Returns the pperipheral base address of the raspberry pi based on the ranges set within the device tree.
```
joperezr: NIT: `pperipheral` -> `peripheral`

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,49 @@ protected ulong ClearRegister
...
+        /// The return value is 32bit even on 64 bit operating systems (debian / ubuntu tested) but may change in the future
+        /// 
+        /// Based on file at https://github.com/RPi-Distro/raspi-gpio/blob/master/raspi-gpio.c
+        /// </summary>
+        /// <returns>This reurns the peripheral base address as a 32 bit address.</returns>
```
joperezr: `reurns` -> `returns`

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,49 @@ protected ulong ClearRegister
...
+        private uint GetPeripheralBaseAddress()
```
joperezr: Given that the driver is really only supose to work with the BCM2835, isn't this address always the same?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,49 @@ protected ulong ClearRegister
...
+                // if the next range is 0 then assume that the base address is in the next 32 bits so read again.
+                if ((buffer[0] | buffer[1] | buffer[2] | buffer[3]) == 0)
+                {
+                    buffer = rdr.ReadBytes(4);
+                }
+
+                // convert the read data into big endian
+                retval = (uint)(buffer[0] << 24) | (uint)(buffer[1] << 16) | (uint)(buffer[2] << 8) | buffer[3];
+
+                // if the values don't look correct then assume an error and return 0
```
joperezr: What do you mean about not looking correct? Was this tested in a few different devices and distros?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,49 @@ protected ulong ClearRegister
...
+                if (firstRange != 0x7E000000 || !(retval == 0x20000000 || retval == 0x3F000000 || retval == 0xFE000000))
+                {
+                    retval = 0xFFFFFFFF;
+                }
+
+            }
+
+            return (retval);
```
joperezr: NIT: remove parenthesis.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -30,6 +31,17 @@ public GpioController(PinNumberingScheme numberingScheme)
         /// <returns>A driver that works with the board the program is executing on.</returns>
         private static GpioDriver GetBestDriverForBoard()
         {
+            // a first chance of finding a raspberry pi..... perhaps not on Raspbian note 
+            // that we want this to fail quietly if something goes wrong with the test
+            try
```
joperezr: I think we will want to take the PR changes in for 3.0, but we don't want to take this particular change here as it may be risky. I think that if somebody wants to force the use of the RaspberryPiDriver if it is not picked by default they have a way of doing it, so it is probably dangerous to change this logic now as we don't know how other boards will behave with this change.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Well I believe that the BCM2835 was used in the Pi 1, Pi 2 has the 2836, Pi 3 2837 and the Pi4 2711. For software sake the system just reports BCM2835 but they really are different models with IO mapped to different addresses. So to determine where to write the IO data for different models we need to read the info from the device tree. /dev/gpiomem does this for use but for /dev/mem we have to work it out.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Well I am checking that the value being returned is consistent with a Pi1,2,3,4 and if not then return a value indicating an error and throw an exception later. It is a double sanity check to ensure that we do not accidentally write to non io space. Tested on Pi 2,3,4 Raspbian, Debian and Ubuntu.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Well I am checking that the value being returned is consistent with a Pi1,2,3,4 and if not then return a value indicating an error and throw an exception later. It is a double sanity check to ensure that we do not accidentally write to non io space. Tested on Pi 2,3,4 Raspbian, Debian and Ubuntu.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
Frankenslag: @joperezr 

Do you really want to do this ? The implication of doing this test is that, by default, Debian and Ubuntu on a Raspberry Pi will end up using the Unix driver as only Raspbian has the SOC identifier in /proc/cpuinfo.

This does feel a bit of a contradiction as the some of idea of using /dev/mem is that other distributions may not implement /dev/gpiomem and without this change you would have to force the 

I also think that the risk is low as the chances of a non-raspberry pi having the text in the "/proc/device-tree/model" node is unlikely and in case of any exception then we just continue with the existing code.

Given this do you still want me to remove this area of the PR ?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Changed.... Will be in the next PR Update

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Changed.... Will be in the next PR Update

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Changed.... Will be in the next PR Update

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Changed.... Will be in the next PR Update

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
joperezr: The thing that worries me is that we have done some extensive board testing in the past but we don't yet have all of these different boards in the lab, so we don't have automation against many boards yet. In our initial test run, we would have chosen a specific driver (like the SysFS or Libgpiod) in one board but if we add this logic we would be changing that and we are getting very close to release so I'm afraid this would be a risky change. The other thing that makes me feel like we may not want to take this very specific part of the change in, is that there is a workaroud, that I agree is not ideal, but the consumer can force the use of RaspberryPiDriver if they wanted to. Most of our device bindings that use GpioControllers take in a IGpioController in their constructor, so they could  have code in their end to detect a RaspberryPi, and if they did then they could pass in a controller that forced this driver.

I'm not saying that we will never take this change, I'm just saying that it may be risky to take this at this stage of the release where we will only have about 2 more previews before release. I actually think that we should have used your check in the beginning instead of using /proc/cpuinfo, but given that we have done extensive testing with that logic, I would rather not change this now. 

This is my current thinking, but I'm totally up for discussing this more if you feel strongly about this.

Frankenslag: @joperezr 
I have no issues with anything from above .... I will remove the detection logic and then let you know. Perhaps I should raise an issue that could be marked for 3.1

joperezr: Sounds like a plan, I would actually mention in the issue that what we want is to replace the existing logic that uses Regex and /proc/cpuinfo and use yours instead.

Frankenslag: @joperezr 
I have removed the Pi Detection logic... I have also tested that it still works no a Pi 4B on Raspbian by renaming /dev/gpiomem and seeing it accessing /dev/mem with the debugger. Hope that is o.k.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -10,6 +10,7 @@ namespace System.Device.Gpio
 {
     public sealed partial class GpioController
     {
+
```
joperezr: NIT: remove edits on this file.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -13,11 +13,17 @@ namespace System.Device.Gpio.Drivers
 {
     public unsafe partial class RaspberryPi3Driver : GpioDriver
     {
+
```
joperezr: NIT: extra line here and after your constant.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,49 @@ protected ulong ClearRegister
...
+        /// The return value is 32bit even on 64 bit operating systems (debian / ubuntu tested) but may change in the future
+        /// 
+        /// Based on file at https://github.com/RPi-Distro/raspi-gpio/blob/master/raspi-gpio.c
+        /// </summary>
+        /// <returns>This returns the peripheral base address as a 32 bit address.</returns>
+        private uint GetPeripheralBaseAddress()
+        {
+            uint retval = 0xFFFFFFF;
+            byte[] buffer;
+
```
joperezr: NIT: extra line.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,49 @@ protected ulong ClearRegister
...
+            uint firstRange;
+
+            using (BinaryReader rdr = new BinaryReader(File.Open(DeviceTreeRanges, FileMode.Open, FileAccess.Read)))
+            {
+                // get the first bigendian range... this seems to allways be 0x7E000000
+                buffer = rdr.ReadBytes(4);
+                firstRange = ((uint)buffer[0] << 24) | (uint)(buffer[1] << 16) | (uint)(buffer[2] << 8) | buffer[3];
```
joperezr: NIT: Instead of calculating this yourself, you should probably use
```c#
firstRange = BinaryPrimitives.ReadUInt32BigEndian(buffer);
```

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,49 @@ protected ulong ClearRegister
...
+                buffer = rdr.ReadBytes(4);
+
+                // if the next range is 0 then assume that the base address is in the next 32 bits so read again.
+                if ((buffer[0] | buffer[1] | buffer[2] | buffer[3]) == 0)
+                {
+                    buffer = rdr.ReadBytes(4);
+                }
+
+                // convert the read data into big endian
+                retval = (uint)(buffer[0] << 24) | (uint)(buffer[1] << 16) | (uint)(buffer[2] << 8) | buffer[3];
```
joperezr: Same here, use BinaryPrimitves instead.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -346,6 +395,10 @@ private void InitializeSysFS()
 
         private void Initialize()
         {
+            uint gpioRegisterOffset = 0;
```
joperezr: NIT: no need to do this assignment, 0 is the default value for uint.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -13,11 +13,17 @@ namespace System.Device.Gpio.Drivers
...
     {
+
+        // error indicates that an entity doesn't exist
+        private const int Linux_Enoent = 2;
+
         private RegisterView* _registerViewPointer = null;
-        private const int GpioRegisterOffset = 0x00;
         private static readonly object s_initializationLock = new object();
         private static readonly object s_sysFsInitializationLock = new object();
+        private const uint GpioPeripheralOffset = 0x0020_0000; // offset from the peripheral base address of the GPIO registers
```
krwq: @buyaa-n does this look correct to you? asking because you mentioned once something about address being different on x64 vs x86 or something along these lines

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: I think is fine as long as we do whatever this mentions: https://github.com/RPi-Distro/raspi-gpio/blob/master/LICENSE

You might need to edit https://github.com/dotnet/iot/blob/master/THIRD-PARTY-NOTICES.TXT (use similar file in corefx for reference)

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +334,49 @@ protected ulong ClearRegister
...
+            {
+                // get the first bigendian range... this seems to allways be 0x7E000000
+                buffer = rdr.ReadBytes(4);
+                firstRange = ((uint)buffer[0] << 24) | (uint)(buffer[1] << 16) | (uint)(buffer[2] << 8) | buffer[3];
+
+                // read the next range into a buffer....
+                buffer = rdr.ReadBytes(4);
+
+                // if the next range is 0 then assume that the base address is in the next 32 bits so read again.
+                if ((buffer[0] | buffer[1] | buffer[2] | buffer[3]) == 0)
```
krwq: where is this assumption coming from? Is this documented somewhere?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: might be also worth to check 64 vs 32 bit - there might be some differences there

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: please file an issue on this

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
buyaa-n: Have no idea about this address, that was about pin numbering offset address only for SysFsDriver, didn't tried RaspberryPiDriver for 64 bit Linux. Read the offset from the system https://github.com/dotnet/iot/pull/386/files#diff-f9b9a68ada4ceabc2a9b052f1c52a2cbR43
  
I guess he get this from device reference or something like that, so maybe better add link to reference

joperezr: This looks correct to me. This is in the BCM chipset datasheet and it is standard for all of those chips so it is accurate.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: @krwq 
There is... For a 64 bit range then the peripheral base address will still be somewhere < 4GB. and so the upper 32 bits will be zero.  The lines before the snippet above test for this zero and then read the next 32 bits to get the base address. I haven't commented this very well though. and so will either revisit the method or comment more clearly.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: See previous response regarding 64bit Ranges

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: @joperezr @krwq 
I am intending to investigate using a function in libbcm_host to determine the Peripheral Base Address via interrop. This is described in the PI documentation as the correct way to go. If it works then the conversations about the home grown method will be unnecessary I guess.

joperezr: > update: It looks like the userland library libbcm_host is not installed by default in Debian and ubuntu. Given this perhaps we should continue with the GetPeripheralBaseAddress method.

I see, yeah or another think we could try is to PInvoke and see if the method is there and if we fail then we fallback to this method

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: The compiler complains if it is not set somewhere.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Issue #686 filed

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: @joperezr @krwq The latest PR update calls libbcm_host for the base address and fails over to the GetPeripheralBaseAddress method if the library isn't installed.

I have also done more commenting around this method.

The PR update seems to file on timeout for helix tests

krwq:
@Frankenslag [my PR](https://github.com/dotnet/iot/pull/685) could have caused conflicts here. Please rebase your PR and make sure to fix all of the docs issues (previously they might not have been reported and PR might have stayed green)

Frankenslag:
> @Frankenslag my PR could have caused conflicts here. Please rebase your PR and make sure to fix all of the docs issues (previously they might not have been reported and PR might have stayed green)

@krwq
Rebased - no errors found - Pull updated,

src/System.Device.Gpio/Interop/Unix/libbcm_host/Interop.libbcmhost.cs
```diff
@@ -0,0 +1,20 @@
...
+// See the LICENSE file in the project root for more information
+
+using System;
+using System.Runtime.InteropServices;
+
+internal partial class Interop
+{
+    internal partial class libbcmhost
+    {
+        private const string LibbcmhostLibrary = "libbcm_host";
```
krwq: is this available on every PI? (I can see it on my Pi3 under `/opt/vc/lib/libbcm_host.so` but not sure about Pi2/Pi4). Also does this need `.so` extension? (If you hit this code path and works fine then presumably works)

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -2,22 +2,26 @@
...
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Runtime.CompilerServices;
 
 namespace System.Device.Gpio.Drivers
 {
     public unsafe partial class RaspberryPi3Driver : GpioDriver
     {
+        private const int Linux_Enoent = 2; // error indicates that an entity doesn't exist
```
krwq: it's ok to use original casing (`ENOENT`) and remove `Linux` prefix considering this is in file with `Linux` suffix and const is private

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -358,13 +405,60 @@ private void Initialize()
...
+                            throw new Exception($"Error - Unable to determine peripheral base address.");
+                        }
+
+                        // add on the offset from the peripheral base address to point to the gpio registers
+                        gpioRegisterOffset += GpioPeripheralOffset;
+                    }
                 }
-
-                IntPtr mapPointer = Interop.mmap(IntPtr.Zero, Environment.SystemPageSize, (MemoryMappedProtections.PROT_READ | MemoryMappedProtections.PROT_WRITE), MemoryMappedFlags.MAP_SHARED, fileDescriptor, GpioRegisterOffset);
+
```
krwq: nit: random spaces

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -358,13 +405,60 @@ private void Initialize()
...
+                            gpioRegisterOffset = Interop.libbcmhost.bcm_host_get_peripheral_address();
+
+                            // if we get zero back then we use our own internal method. This can happen 
+                            // on a Pi4 if the userland libraries haven't been updated and was fixed in Jul/Aug 2019.
+                            if(gpioRegisterOffset == 0)
+                            {
+                                gpioRegisterOffset = GetPeripheralBaseAddress();
+                            }
+                        }
+                        catch(DllNotFoundException)
```
krwq: nit: space after `catch`

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -358,13 +405,60 @@ private void Initialize()
...
+
+                        try
+                        {
+                            // get the periphal base address from the libbcm_host library which is the reccomended way
+                            // according to the RasperryPi website
+                            gpioRegisterOffset = Interop.libbcmhost.bcm_host_get_peripheral_address();
+
+                            // if we get zero back then we use our own internal method. This can happen 
+                            // on a Pi4 if the userland libraries haven't been updated and was fixed in Jul/Aug 2019.
+                            if(gpioRegisterOffset == 0)
```
krwq: nit: space after `if`

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: is this exception expected on some PIs? if not I'd rather just let it bubble up

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -358,13 +405,60 @@ private void Initialize()
...
+
+                    // if /dev/gpiomem doesn't seem to be available then let's try /dev/mem
+                    fileDescriptor = Interop.open(MemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
+                    if (fileDescriptor == -1)
+                    {
+                        throw new IOException($"Error {Marshal.GetLastWin32Error()} initializing the Gpio driver.");
+                    }
+                    else // success so set the offset into memory of the gpio registers
+                    {
+                        gpioRegisterOffset = ~0U;
```
krwq: use explicit value

Frankenslag: Done in the PR Update

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Done in the PR Update

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Done in the PR Update

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Done in the PR Update I hope.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Done in the PR Update

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: On some Pi's that don't use Raspbian then I have found this library missing in the default install. I can't remember of it was the Arm64 Debian or Ubuntu.

src/System.Device.Gpio/Interop/Unix/libbcm_host/Interop.libbcmhost.cs
Frankenslag: No it's not available on every pi (see comment later on).  but when it is available then It is reccomended that it is used by the Pi folk. As you say it works without the .so suffix and I note that the other interrop libraries don't use a suffix either.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: still here (I think you only updated the one below)

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -358,13 +405,60 @@ private void Initialize()
...
+                        }
+                        catch (DllNotFoundException)
+                        {
+                            // if the code gets here then then use our internal method as libbcm_host isn't available.
+                            gpioRegisterOffset = GetPeripheralBaseAddress();
+                        }
+
+                        if (gpioRegisterOffset == 0xFFFFFFFF)
+                        {
+                            throw new Exception($"Error - Unable to determine peripheral base address.");
```
krwq: We should throw some more meaningful exception type here. Perhaps `InvalidOperationException`

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +332,45 @@ protected ulong ClearRegister
...
+                // if the CPU bus Peripheral Base Address is 0 then assume that this is a 64 bit address and so read the next 32 bits.
+                if (cpuBusPeripheralBaseAddress == 0)
+                {
+                    cpuBusPeripheralBaseAddress = BinaryPrimitives.ReadUInt32BigEndian(rdr.ReadBytes(4));
+                }
+
+                // if the address values don't fall withing known values then assume an error
+                if (vcBusPeripheralBaseAddress != 0x7E000000 || !(cpuBusPeripheralBaseAddress == 0x20000000 || cpuBusPeripheralBaseAddress == 0x3F000000 || cpuBusPeripheralBaseAddress == 0xFE000000))
+                {
+                    cpuBusPeripheralBaseAddress = 0xFFFFFFFF;
```
krwq: please add constants for all values, i.e.:

```csharp
const uint InvalidAddress = 0xFFFFFFFF;
// etc
```

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +332,45 @@ protected ulong ClearRegister
...
+                cpuBusPeripheralBaseAddress = BinaryPrimitives.ReadUInt32BigEndian(rdr.ReadBytes(4));
+
+                // if the CPU bus Peripheral Base Address is 0 then assume that this is a 64 bit address and so read the next 32 bits.
+                if (cpuBusPeripheralBaseAddress == 0)
+                {
+                    cpuBusPeripheralBaseAddress = BinaryPrimitives.ReadUInt32BigEndian(rdr.ReadBytes(4));
+                }
+
+                // if the address values don't fall withing known values then assume an error
+                if (vcBusPeripheralBaseAddress != 0x7E000000 || !(cpuBusPeripheralBaseAddress == 0x20000000 || cpuBusPeripheralBaseAddress == 0x3F000000 || cpuBusPeripheralBaseAddress == 0xFE000000))
```
krwq: would be useful to add some comment to where those constants come from (i.e. spec page or some link)

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: I think I have them all in the next PR Update

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Changed Exception in the next PR Update

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: Created a constant and also created constants for the expected values of PBA,

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
Frankenslag: added a link to the DTS repository in GitHub. Now have named constants that indicate which chipset the values are associated with,

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -328,6 +337,46 @@ protected ulong ClearRegister
...
+            using (BinaryReader rdr = new BinaryReader(File.Open(DeviceTreeRanges, FileMode.Open, FileAccess.Read)))
+            {
+                // get the Peripheral Base Address on the VC bus from the device tree this is to be used to verify that
+                // the right thing is being read and should always be 0x7E000000
+                vcBusPeripheralBaseAddress = BinaryPrimitives.ReadUInt32BigEndian(rdr.ReadBytes(4));
+
+                // get the Peripheral Base Address on the CPU bus from the device tree.
+                cpuBusPeripheralBaseAddress = BinaryPrimitives.ReadUInt32BigEndian(rdr.ReadBytes(4));
+
+                // if the CPU bus Peripheral Base Address is 0 then assume that this is a 64 bit address and so read the next 32 bits.
```
krwq: double checking - should you be reading next 32 or 64 bits?

Frankenslag: @krwq 

The range for the Pi4 has the address defined as a big-endian 64 bit value with the first 32bits set to zero is it will allways be in the first 4GB.

`ranges = <0x7e000000  0x0 0xfe000000  0x01800000>,`

This is a 32bit address for the videocore pba, then a 64bit address for the userland pba and then a 32bit value for the pba length.

Whereas the Pi3 has a 32 bit value  for the userland PBA  (and similarly for the Pi2)

`ranges = <0x7e000000 0x3f000000 0x1000000>,`

So this code and also the libbcm_host userland code tests for 64 bit by reading the first 32bit value of the pba and then if it is zero knows that it is 64 bits wide and so reads the next 32 bit value for the pba.

 Note that early versions of Raspbian buster have the correct device tree but outdated libbcm_host and so a zero is returned. This code detects that and uses the internal method to determine the pba which is correct.

joperezr:
Thanks a lot for getting this through and for being patient with reviews @Frankenslag! I think this will help us run in a bunch of new places which is amazing!

PR #606, opened 2019-07-18
Fixing SetLastError and remove HasFlag to use bitwise instead

### LINKED ISSUE #146: Apply comments  from PR #133 related/can be found in SysFs driver implementation
1. Refactor Sysfs event handling logic, make all tests pass. 
2. Some comments for LibGpiod driver implementation PR https://github.com/dotnet/iot/pull/133 related to SysFs driver too. Apply those comments to SysFs driver and test accordingly such as:

https://github.com/dotnet/iot/pull/133#discussion_r244890760
https://github.com/dotnet/iot/pull/133#discussion_r244891801
https://github.com/dotnet/iot/pull/133#discussion_r245413628
https://github.com/dotnet/iot/pull/133#discussion_r245459454

joperezr:
Fix #146 

cc: @buyaa-n @krwq 

Addressing comments made on PR #133 on all of our drivers.

### DIFF
--- a/src/System.Device.Gpio/Interop/Unix/Libc/Interop.lseek.cs
+++ b/src/System.Device.Gpio/Interop/Unix/Libc/Interop.lseek.cs
@@ -6,7 +6,7 @@ using System.Runtime.InteropServices;
 
 internal partial class Interop
 {
-    [DllImport(LibcLibrary, SetLastError = true)]
+    [DllImport(LibcLibrary)]
     internal static extern int lseek(int fd, int offset, SeekFlags whence);
 }
 
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -16,32 +16,11 @@ internal partial class Interop
         internal static IntPtr InvalidHandleValue = new IntPtr(-1);
 
         /// <summary>
-        /// Create a new gpiochip iterator.
-        /// </summary>
-        /// <returns>Pointer to a new chip iterator object or a SafeChipIteratorHandle pointing to 0 which will return <see langword="true" /> for IsInvalid.</returns>
-        [DllImport(LibgpiodLibrary)]
-        internal static extern SafeChipIteratorHandle gpiod_chip_iter_new();
-
-        /// <summary>
         /// Release all resources allocated for the gpiochip iterator and close the most recently opened gpiochip(if any).
         /// </summary>
         /// <param name="iter">The gpiochip iterator object</param>
         [DllImport(LibgpiodLibrary)]
         internal static extern void gpiod_chip_iter_free(IntPtr iter);
-
-        /// <summary>
-        /// Release all resources allocated for the gpiochip iterator but don't close the most recently opened gpiochip (if any).
-        /// </summary>
-        /// <param name="iter">The gpiochip iterator object</param>
-        [DllImport(LibgpiodLibrary)]
-        internal static extern void gpiod_chip_iter_free_noclose(SafeChipIteratorHandle iter);
-
-        /// <summary>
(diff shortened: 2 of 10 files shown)


src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -38,15 +38,15 @@ public LibGpiodDriver(int gpioChip = 0)
 
         protected internal override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEventTypes eventTypes, PinChangeEventHandler callback)
         {
-            if (eventTypes.HasFlag(PinEventTypes.Rising) || eventTypes.HasFlag(PinEventTypes.Falling))
+            if ((eventTypes & PinEventTypes.Rising) != 0 || (eventTypes & PinEventTypes.Falling) != 0)
```
krwq: I don't think we should change this - instead we should make HasFlag faster - HasFlag is much cleaner than this

joperezr: This was a suggestion by @danmosemsft on PR #133 since just doing a bitwise operation would be faster. I know that HasFlag is a bit cleaner since it validates that the enum types specified are the same, but given that we control all the places where we do these checks I believe that we should be fine to just do the bitwise operation to have a better perf.

krwq: I feel like this is a bit microptimization since we don't expect people to register events too frequently (expecting once per runtime of the app?)

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -40,7 +40,7 @@ internal partial class libgpiod
         /// Get the next gpiochip handle.
         /// </summary>
         /// <param name="iter">The gpiochip iterator object</param>
-        [DllImport(LibgpiodLibrary, SetLastError = true)]
+        [DllImport(LibgpiodLibrary)]
         internal static extern SafeChipHandle gpiod_chip_iter_next(SafeChipIteratorHandle iter);
```
buyaa-n: Seems this and above methods not used anymore, should we remove it?

src/System.Device.Gpio/Interop/Unix/Libc/Interop.lseek.cs
```diff
@@ -6,7 +6,7 @@
 
 internal partial class Interop
 {
-    [DllImport(LibcLibrary, SetLastError = true)]
```
buyaa-n: Seems we are using LastError only for Interop open() mmap() methods, shouldn't we remove it from all other remaining methods? like read(), write() methods ...

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
joperezr: Agreed but that doesn't mean we shouldn't optimize where we can, specially if it is as easily as it is here where we aren't increasing complexity in the code.

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
joperezr: Yup you are right, I'll remove the unused ones.

src/System.Device.Gpio/Interop/Unix/Libc/Interop.lseek.cs
joperezr: I did go and made sure that all the remaining SetLastError PInvokes actually call Marshal.GetLastWin32Error after it, which means that the ones that still have it, is because they need it.

PR #607, opened 2019-07-19
Add simple SoftwareSpi implementation (no timing control) and extract it from MCP3008

krwq:
I've initially wanted to just extract the code but I ended up doing most of the implementation. The only thing to do left here is timing control (match the frequency). We should probably also have a way of having untimed version as well when doing that (so as is right now).

Supports all 4 modes (only tested Mode0 though as I don't have an oscilloscope or devices to test with) data bit length and other SPI settings


src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+            _controller.OpenPin(_clk, PinMode.Output);
+            _controller.OpenPin(_miso, PinMode.Input);
+            _controller.OpenPin(_mosi, PinMode.Output);
+            _controller.OpenPin(_cs, PinMode.Output);
+
+            // aka. CPOL
+            bool idle = ((int)_settings.Mode & 1) == 1;
+
+            // aka. CPHA
+            bool cpha = ((int)_settings.Mode & 0b10) == 0b10;
```
krwq: rename this to `onPulseEnd`

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            if (dataToRead.Length != dataToRead.Length)
+            {
+                throw new ArgumentException(nameof(dataToWrite));
+            }
+
+            int bitLen = _settings.DataBitLength;
+            int lastBit = bitLen - 1;
+
+            using (ChipSelect)
```
krwq: this isn't horribly fast and can be theoretically optimized - but considering we actually plan to add additional delay to conform to timing restrictions it actually might not make sense to make this any faster

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+    public class SoftwareSpi : SpiDevice
+    {
+        private GpioController _controller;
+        private readonly int _clk;
+        private readonly int _miso;
+        private readonly int _mosi;
+        private readonly int _cs;
+        private readonly SpiConnectionSettings _settings;
+
+        public SoftwareSpi(int clk, int miso, int mosi, int cs, SpiConnectionSettings settings = null, GpioController controller = null)
```
joperezr: Can you please add XML docs to all of these? We are now packing the docs in the package so it will be very useful for people developing on VSCode or Visual Studio to get info on what each of this parameters should be.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            _controller = controller ?? new GpioController();
+
+            _settings = settings ?? new SpiConnectionSettings(-1, -1);
+
+            _clk = clk;
+            _miso = miso;
+            _mosi = mosi;
+            _cs = cs;
```
joperezr: Can we rename this to chipSelectPin? the rest of the names might be more intuitive, but `cs` might be confusing.

krwq: it's called like that pretty much everywhere including drawn on the board on multiple devices - I don't think this is confusing but can rename

joperezr: I guess it's fine as long as you add the xml docs since you can have more verbose info there.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+            }
+        }
+
+        private bool ReadWriteBit(bool bitToWrite)
+        {
+            _controller.Write(_mosi, bitToWrite);
+            return (bool)_controller.Read(_miso);
+        }
+
+        public override void Read(Span<byte> data)
```
krwq: I feel like the boilerplate should be moved as default implementation for SpiDevice (so change them from abstract to virtual there)

src/devices/SoftwareSpi/samples/README.md
```diff
@@ -0,0 +1,12 @@
+﻿# Software implementation of SPI
+
+Usage is very simple:
+
+```csharp
```
krwq: the READMEs are wrong, this should go under SoftwareSpi/ and this file should only point to Mcp3008 sample and all .cs files in this folder removed

src/devices/SoftwareSpi/samples/SoftwareSpi.Sample.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
krwq: remove me

src/devices/SoftwareSpi/samples/SoftwareSpi.Samples.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
krwq: remove me

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+            _mosi = mosi;
+            _cs = cs;
+
+            _controller.OpenPin(_clk, PinMode.Output);
+            _controller.OpenPin(_miso, PinMode.Input);
+            _controller.OpenPin(_mosi, PinMode.Output);
+            _controller.OpenPin(_cs, PinMode.Output);
+
+            // aka. CPOL
+            bool idle = ((int)_settings.Mode & 1) == 1;
```
joperezr: I think you got these backwards. CPOL is the msb CPHA is the lsb

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            public void Exit()
+            {
+                if (_exit != null)
+                {
+                    _exit();
+                }
+            }
+        }
+        private struct Scope : IDisposable
```
joperezr: NIT: missing new line

joperezr: Why is this not a class as well? Or why is ScopeData not a struct?

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+                    },
+                    exit: () => {
+                        controller.Write(_clk, idle);
+                    });
+            }
+            else
+            {
+                _bitTransfer = new ScopeData(
+                    exit: () => {
+                        _controller.Write(_clk, !idle);
```
joperezr: Why don't you need an enter here?

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            _controller.Write(_cs,!(bool)_settings.ChipSelectLineActiveState);
+            _controller.Write(_clk, idle);
+
+            // TODO: To respect ClockFrequency we need to inject the right delays here
+            //       and have some very accurate way to measure time.
+            //       Ideally we should verify the output with an oscilloscope.
+            if (cpha)
+            {
+                _bitTransfer = new ScopeData(
```
joperezr: I think this whole section with cpha should be commented much more so that it is more readable. CPOL idle is very straight forward, but the CPHA shiftness isn't so it would help if you add verbose comments in here describing the process in case there is a bug here in the future and somebody is trying to read/modify this code.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+        private bool ReadWriteBit(bool bitToWrite)
+        {
+            _controller.Write(_mosi, bitToWrite);
+            return (bool)_controller.Read(_miso);
+        }
+
+        public override void Read(Span<byte> data)
+        {
+            Span<byte> dataToWrite = stackalloc byte[data.Length];
+            dataToWrite.Fill(0);
```
joperezr: Why do you need to fill the span with 0? Presumably all the data will be filled in any case.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+            TransferFullDuplex(dataToWrite, data);
+        }
+
+        public override void Write(ReadOnlySpan<byte> data)
+        {
+            Span<byte> dataToRead = stackalloc byte[data.Length];
```
joperezr: If you only want to write, then you don't need this extra allocation, why not just send null and handle it in your TrasnferFullDuplex method?

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+        public override byte ReadByte()
+        {
+            Span<byte> data = stackalloc byte[1];
+            Read(data);
+            return data[0];
+        }
+
+        public new void Dispose()
+        {
+            _controller?.Dispose();
```
joperezr: doesn't SpiDevice implement the dispose pattern? If it doesn't I think it should

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+            _controller = null;
+            base.Dispose();
+        }
+
+        private ScopeData _bitTransfer;
+        private ScopeData _chipSelect;
+        private Scope BitTransfer => new Scope(_bitTransfer);
```
joperezr: Why not just do this in the constructor and that way Scope doesn't need to be a struct?

src/devices/SoftwareSpi/SoftwareSpi.cs
joperezr: I get where exit will be called, but when will enter be called?

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,199 @@
...
+            int lastBit = bitLen - 1;
+
+            using (ChipSelect)
+            {
+                for (int i = 0; i < dataToRead.Length; i++)
+                {
+                    byte readByte = 0;
+                    for (int j = 0; j < bitLen; j++)
+                    {
+                        using (BitTransfer)
```
joperezr: >BitTransfer [](start = 31, length = 11)

Same question here, where will enter be called?

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: Scope is constructed frequently so it's a struct so that the operation is cheap. ScopeData doesn't need to be fast as it's only constructed once per usage.

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: on enter you are supposed to be in the idle state which you already are

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: @joperezr wasn't sure if that's the case, will double check if we can safely remove it

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: you still should be writing zeros when you read and this is almost no op

src/devices/SoftwareSpi/SoftwareSpi.cs
joperezr: The way the hardware version works is that when you call this method, we should expect exactly `data.Length` bytes to be read, so all the buffer will be filled. That is why in hardware version we don't have to initialize the data to 0

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: well it's not virtual so you cannot override. I got mixed feeling if it should. I will check what we do elsewhere and follow that pattern

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: this needs to be constructed every time you access this so that Enter and Exit gets called

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: ChipSelect constructs the struct every time you access the property - constructor will call it

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: ditto

src/devices/SoftwareSpi/SoftwareSpi.cs
joperezr: Ideally we would want SpiDevice to be like:

```c#
public class SpiDevice : IDisposable
{
    public void Dispose()
    {
        Dispose(true);
        GC.SuppressFinalize(this);
    }

    {
        //Dispose stuff that is specific to SpiDevice and let subclasses dispose their own stuff.
    }
}
```

src/devices/SoftwareSpi/SoftwareSpi.cs
joperezr: oh I missed the call in the constructor

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: not addressing this here

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: changing this to func so it's more explicit

Frankenslag:
@krwq , @joperezr 

I have been testing this out and it looks to work ok for mode 3 also.

Just for interest I looked at the clock being generated and found the following

- Pi4b Raspbian using /dev/gpiomem 420khz
- Pi3b Ubuntu (arm64) using /dev/gpiomem 250khz
- Pi3b Debian (arm64) using /dev/mem 125khz
- Pi3b Ubuntu (arm64) using libgpiod 420hz (yes hz)

So the change to use /dev/mem (if appropiate) rather than falling back to libgpiod seems well worth it and given that I have not been able to get hardware spi working in raw Debian then this PR really helps in some circumstances.

But shouldn't it be in the System.Device.Spi namespace and package somewhere ?

krwq:
@Frankenslag cool, thanks for measuring - how did you measure? Did you count it with software or with oscilloscope? We can probably get a bit better perf assuming linux driver is not a limit

I'm kinda surprised libgpiod didn't do too well. @buyaa-n do you perhaps know how fast it should be?

> But shouldn't it be in the System.Device.Spi namespace and package somewhere?

It should be in Iot.Device.Bindings couple of days after it's merged. We are planning to eventually add it to System.Device.Spi (probably not public and some kind of SpiDevice.Create overload) but since this isn't production ready (not well tested, perf improvements need to be investigated) it will be better to first let it be tried out, polished up and then we can think about moving it to System.Device namespace.

Also one way to optimize on RPi would be to use direct register access similarly as LED matrix does (https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs#L313 and ClearRegister next to it)

Frankenslag:
@krwq @buyaa-n 
I measured using a cheap logic analyzerr but it seems to measure with reasonable accuracy.

Rather than libgpiod I should have said that it was using the unix driver and so that could be sysfs.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,229 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Threading;
+using System.Device.Gpio;
+using System.Device.Spi;
+
+namespace Iot.Device.Spi
+{
+    public class SoftwareSpi : SpiDevice
```
shaggygi: We currently have [SoftPwm](https://github.com/dotnet/iot/tree/master/src/devices/SoftPwm).  This needs to be renamed `SoftSpi` or `SoftPwm` needs to be renamed to `SoftwarePwm` to be consistent.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,229 @@
...
+            _controller.OpenPin(_mosi, PinMode.Output);
+            _controller.OpenPin(_cs, PinMode.Output);
+
+            // aka. CPOL - tells us which state of the clock means idle (false means 'low' or 'ground' or '0')
+            bool idle = ((int)_settings.Mode & 0b10) == 0b10;
+
+            // aka. CPHA - tells us when read/write is 'captured'
+            bool onPulseEnd = ((int)_settings.Mode & 1) == 1;
+
+            _controller.Write(_cs,!(bool)_settings.ChipSelectLineActiveState);
```
shaggygi: NIT: Space between parameters.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,229 @@
...
+            _controller = null;
+            base.Dispose();
+        }
+
+        private ScopeData _bitTransfer;
+        private ScopeData _chipSelect;
+        private Scope StartBitTransfer() => new Scope(_bitTransfer);
+        private Scope StartChipSelect() => new Scope(_chipSelect);
+
+        private class ScopeData
```
shaggygi: NIT: Would be nice to have in separate file if it stays under device binding.  I guess it would be okay to keep within if moving to core API later.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,229 @@
...
+    {
+        private GpioController _controller;
+        private readonly int _clk;
+        private readonly int _miso;
+        private readonly int _mosi;
+        private readonly int _cs;
+        private readonly SpiConnectionSettings _settings;
+
+        /// <summary>
+        /// Software implementation of the SPI
```
shaggygi: NIT: Periods at end of summary and parameters.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,229 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Threading;
+using System.Device.Gpio;
+using System.Device.Spi;
+
+namespace Iot.Device.Spi
```
shaggygi: Should the namespace be `Iot.Device.Spi.Devices`?  I guess it fits and not redundant like we kept main `SpiDevice`.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,229 @@
...
+                            }
+                        }
+                    }
+
+                    dataToRead[i] = readByte;
+                }
+            }
+        }
+
+        /// <inheritdoc />
```
shaggygi: How does this work since it doesn't override like the others?

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,229 @@
...
+        private bool ReadWriteBit(bool bitToWrite)
```
shaggygi: Should this be `WriteReadBit`?  What you have sounds better, but the logic is the opposite.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -0,0 +1,229 @@
...
+        private ScopeData _bitTransfer;
+        private ScopeData _chipSelect;
+        private Scope StartBitTransfer() => new Scope(_bitTransfer);
+        private Scope StartChipSelect() => new Scope(_chipSelect);
+
+        private class ScopeData
+        {
+            Action _enter, _exit;
+
+            public ScopeData(Action enter = null, Action exit = null)
```
shaggygi: Name is okay, but I'm wondering if `ScopeBit` would be better.

krwq: this is unrelated to bits except in this context

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: it happens at the same time considering time is "stopped" here :smile:

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: copy & paste fail :P

krwq:
```suggestion
```

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: actually SoftPwm is since couple of days ago SoftwarePwmChannel, we just didn't rename the folder

src/devices/SoftwareSpi/SoftwareSpi.cs
shaggygi: I wasn't actually sure as I've never used **inheritdoc** before and just curious 😄

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq:
```suggestion
            _controller.Write(_cs, !(bool)_settings.ChipSelectLineActiveState);
```

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: it's currently a nested private class - I don't think our guidelines require moving it (at least I've seen similar patterns elsewhere and noone ever complained)

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: I'm expecting this type will eventually get moved to SpiDevice.Create and became internal. I'd just leave this as is

shaggygi: Oh that's right.  I recall similar chat for PWM.

krwq:
@Frankenslag could you perhaps run the same test forcing libgpiod? I'm actually curious how it would perform. ~400Hz for SysFs is fine I guess since it's considered the slowest implementation

buyaa-n:
@Frankenslag in order to use libgpiod driver its need to be installed on your Linux platform, if its installed GpioController should load libpgiod driver automatically. 
```
wget https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/snapshot/libgpiod-1.4.tar.gz
tar xvf libgpiod-1.4.tar.gz
cd libgpiod-1.4.tar.gz
./autogen.sh \
   --enable-tools=yes       \
   --prefix=/usr/local
make
sudo make install
```

Reference: https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/about/

Frankenslag:
@buyaa-n @krwq 

I forced libgpiod (note that I used apt-get install libgpiod-dev to install it)

You can see that it is running at approx. 100khz. It does look like I was testing sysfs when it was really slow.

image

krwq:
@Frankenslag cool, 100kHz sounds pretty good comparing to others

src/devices/SoftwareSpi/SoftwareSpi.cs
joperezr: I like this being SoftwareSpi and the other one being SoftwarePwmChannel, I agree we should just rename the folder of SoftPwm to match the binding.

src/devices/SoftwareSpi/SoftwareSpi.cs
joperezr: There still seems to be a problem here since this Dispose shouldn't be public, but I assume that the problem lies on the fact that SpiDevice doesn't implement the dispose pattern correctly.

shaggygi: > SpiDevice doesn't implement the dispose pattern correctly.

We recently updated this.  Did I do something incorrectly here... https://github.com/dotnet/iot/pull/614/commits/6d4b8718b0da000b92156c18e9519372ff4a6f1e

src/devices/SoftwareSpi/SoftwareSpi.cs
shaggygi: >  I agree we should just rename the folder of SoftPwm to match the binding.

Let's hold off on the SoftPwm rename.  I wanted to test some and I'm assuming it will be moved to core API and hidden with its own Create(... pin params ...) based on previous talks.

shaggygi: > I like this being SoftwareSpi...

Since this will also eventually be hidden and placed in core API with its own Create(... pin params ...), what would be the harm of naming `SoftwareSpiDevice` as it will implement SpiDevice?  Just curious.

src/devices/SoftwareSpi/SoftwareSpi.cs
joperezr: mm might be that we just haven't updated the version we ingest here of System.Device.Gpio. @krwq might be worth it to just wait and rebase tomorrow once we have ingested the new changes so this one is a protected override instead of a public override.

src/devices/SoftwareSpi/SoftwareSpi.cs
joperezr: > I wanted to test some and I'm assuming it will be moved to core API and hidden with its own Create(... pin params ...) based on previous talks.

This is in the future plans but not for 3.0, so What I was referring above was simply renaming the folder under devices. Same goes for SoftwareSpi, we don't want to move that under System.Device.Gpio library just yet but do think that we will do so in the future.

shaggygi: @joperezr gotcha.  i'll try to rename the SoftPwm > SoftwarePwm later in the week then.

PR #614, opened 2019-07-22
Update dispose access modifier

shaggygi:
Updated Dispose method based on recent `PwmChannel`.


krwq:
Matches guidelines from here https://docs.microsoft.com/en-us/dotnet/api/system.idisposable?view=netcore-3.0

PR #615, opened 2019-07-22
Adding MLX90614 Binding

ZhangGaoxing:
Infra Red Thermometer MLX90614


src/devices/Mlx90614/Register.cs
```diff
@@ -0,0 +1,13 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Mlx90614
+{
+    internal enum Register
+    {
+        MLX_AMBIENT_TEMP = 0x06,
+        MLX_OBJECT1_TEMP = 0x07,
+        MLX_OBJECT2_TEMP = 0x08
```
krwq: what is MLX_OBJECT2_TEMP? It doesn't seem to be used anywhere?

src/devices/Mlx90614/samples/Program.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Device.I2c;
+using System.Threading;
+
+namespace Iot.Device.Mlx90614.Sample
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine("Hello MLX90614!");
```
krwq: please remove this, it's not strictly part of the sample

src/devices/Mlx90614/Mlx90614.cs
```diff
@@ -0,0 +1,84 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Buffers.Binary;
+using System.Device.I2c;
+using Iot.Units;
+
+namespace Iot.Device.Mlx90614
```
krwq: another sensor with literally `X` in the name. cc: @joperezr

src/devices/Mlx90614/Register.cs
ZhangGaoxing: In the application guide, the specific version of the sensors read the 0x08 register effectively, while the 0x07 register is universal. The functions of these two registers are the same. I will remove 0x08

ZhangGaoxing:
According to #626, I removed properties and read temperature through methods. Recently, I have free time, If necessary, I can revise my previous bindings. I will not submit new bindings lately, and I will check the previous bindings to make sure there are no obvious bugs. @krwq @joperezr

joperezr:
Thanks a lot for all your work @ZhangGaoxing I'll take a look at this today

src/devices/Mlx90614/Mlx90614.cs
```diff
@@ -0,0 +1,74 @@
...
+            return Temperature.FromCelsius(Math.Round(temp, 2));
+        }
+
+        /// <summary>
+        /// Read object temperature from MLX90614
+        /// </summary>
+        /// <returns>Temperature</returns>
+        public Temperature ReadObjectTemperature()
+        {
+            Span<byte> writeBuffer = stackalloc byte[] { (byte)Register.MLX_OBJECT1_TEMP };
```
joperezr: Both of your Read methods look exactly the same except that they read different registers. I believe you should instead just have an InternalRead method that takes in a byte as a parameter containing the register that you want to read, and returns the temperature.

src/devices/Mlx90614/Mlx90614.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            return Temperature.FromCelsius(Math.Round(temp, 2));
+        }
+
+        /// <summary>
+        /// Cleanup
+        /// </summary>
+        public void Dispose()
+        {
+            _i2cDevice?.Dispose();
```
joperezr: you should also set this to null after calling dispose, so that we don't call dispose twice on the i2cdevice in case that we accidentally call dispose twice on this binding.

src/devices/Mlx90614/Mlx90614.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Buffers.Binary;
+using System.Device.I2c;
+using Iot.Units;
+
+namespace Iot.Device.Mlx90614
+{
+    /// <summary>
+    /// Infra Red Thermometer MLX90614
+    /// </summary>
+    public class Mlx90614 : IDisposable
```
joperezr: This should probably be a sealed class, which might help when looking that the x in the name is actually part of the name and not meaning that it is a base class for a device family.

src/devices/Mlx90614/Mlx90614.csproj
```diff
@@ -0,0 +1,20 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="Mlx90614.cs" />
+    <Compile Include="Register.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
+	    <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
```
joperezr: NIT: you have a couple of tabs here. Please change them to be spaces instead.

src/devices/Mlx90614/Mlx90614.cs
```diff
@@ -0,0 +1,74 @@
...
+            _i2cDevice.WriteRead(writeBuffer, readBuffer);
+
+            // The formula is on the datasheet P30.
+            double temp = BinaryPrimitives.ReadInt16LittleEndian(readBuffer) * 0.02 - 273.15;
+
+            return Temperature.FromCelsius(Math.Round(temp, 2));
+        }
+
+        /// <summary>
+        /// Read object temperature from MLX90614
```
joperezr: I'm not sure what "object temperature" means. What object is this refering to? what is the difference between this and the ambient temperature?

joperezr: oh nvm just realized this is a infra red temperature sensor so this now makes sense. Could we fix the comment though in order to make it more clear that this is the method that most people will want to call? perhaps mentioning in the comment that object is the actual thing you are measuring the temperature of when pointing the laser?

src/devices/Mlx90614/samples/Mlx90614.Sample.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
+	    <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
+		</ProjectReference>
```
joperezr: there are some tabs in here as well, please change them to be spaces instead.

joperezr:
> using System;

NIT: missing licensing headers.

---
Refers to: src/devices/Mlx90614/samples/Program.cs:1 in c842796. [](commit_id = c842796a6744fd9b29e66d8eb9cb9b295195730e, deletion_comment = False)

joperezr:
Added some comments but this looks great in general, thanks a lot again @ZhangGaoxing

joperezr:
Thanks a lot for your contribution @ZhangGaoxing!

PR #618, opened 2019-07-22
Fixed by setting usePrecisionTimer in constructor

shaggygi:
The precision timer field wasn't being set in constructor and always false.  Updated to better naming and description, as well.

I have some PWM-related devices on order and plan to try out with new PwmChannel updates soon.  There are some other areas in this binding I want to refactor, but wanted to go ahead and get this bug fixed as peeps might expect the correct timer in the meantime.


src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -74,9 +75,9 @@ public override double DutyCyclePercentage
         /// <param name="pinNumber">The GPIO pin number to be used</param>
         /// <param name="frequency">The frequency in hertz. Defaults to 400</param>
         /// <param name="dutyCyclePercentage">The duty cycle percentage represented as a value between 0.0 and 1.0</param>
-        /// <param name="precisionTimer"><see langword="true"/> to use a precision timer. <see langword="false"/> otherwise</param>
+        /// <param name="usePrecisionTimer">Determines if a high precision timer is used.</param>
```
joperezr: small NIT:
`Determines if a high precision timer is used.` -> `Determines if a high precision timer should be used.`?

krwq:
@shaggygi thanks, also for Linux you might wanna add the same fix for the thread priority as in the LED matrix:
https://github.com/dotnet/iot/blob/master/src/devices/RGBLedMatrix/RgbLedMatrix.cs#L553

note: currently `Thread.CurrentThread.Priority = ThreadPriority.Highest;` doesn't do anything and it's a bug in corefx which needs to be fixed

shaggygi:
@krwq i believe i have it added now in last commit.  there were a build errors after trying to update the csproj similar to RgbMatrix.  Lemme know if this was how you expected code to be added for the thread issue.  thx

src/devices/SoftPwm/SoftwarePwm.csproj
```diff
@@ -2,12 +2,14 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <EnableDefaultItems>false</EnableDefaultItems>
-    <!--Disabling default items so samples source won't get build by the main library-->
+    <RuntimeIdentifier>linux-arm</RuntimeIdentifier>
```
joperezr: Why do you need to specify  a RuntimeIdentifier here, I think that is wrong.

shaggygi: > Why do you need to specify a RuntimeIdentifier here, I think that is wrong.

I thought it looked strange, as well.  I removed and appeared to build locally for binding, but had failures on entire iot build.  Seem to fix it when adding back.  Based it on the RGBLedMatrix project.

shaggygi: @joperezr so i just removed and local build was successful.  It must be the following I removed and reason it failed.  I'm assuming that is needed for the Linux thread issue?

I also removed the RuntimeIdentifier from the RGBLedMatrix project (even though this PR was unrelated).

```
<ProjectReference Include="$(MSBuildThisFileDirectory)../../System.Device.Gpio/System.Device.Gpio.csproj">
      <AdditionalProperties>RuntimeIdentifier=linux</AdditionalProperties>
    </ProjectReference>
```

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -103,9 +105,10 @@ private void UpdateRange()
 
         private void RunSoftPWM()
         {
-            if (_precisionPWM)
+            if (_usePrecisionTimer)
             {
                 Thread.CurrentThread.Priority = ThreadPriority.Highest;
+                System.Interop.ThreadHelper.SetCurrentThreadHighPriority();
```
joperezr: This won't work on Windows right? If we want we can check here where are we running and if running in Linux we can then call this method. This is actually probably the reason why you needed to set a runtimeIdentifier before so let's just put it inside an if check that checks the runtimeinformation to see if we are in Linux.

krwq: other options is perhaps just remove this for now and we can address this separately (we might wanna do similar fix for LED matrix as well)

shaggygi: Removed in last commit

PR #622, opened 2019-07-23
Moving all of our projects to use Project-to-Project references instead of package references.

### LINKED ISSUE #621: Move projects to use P2P instead of package reference
Currently doing any changes to the product is mildly saying a bit annoying:
- make changes to a product without changing anything which depends on it
- merge, wait for package to get published
- edit and auto-pr which is red to fix build errors

joperezr:
fixes #621 

cc: @krwq 

FYI: @shaggygi @Ellerbach @ZhangGaoxing @Frankenslag @RobinTTY @HumJ0218

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,6 +6,7 @@
   <PropertyGroup>
     <Copyright>$(CopyrightNetFoundation)</Copyright>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <MainLibraryPath>$(MSBuildThisFileDirectory)src/System.Device.Gpio/</MainLibraryPath>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/src/devices/Ds1307/samples/Ds1307.Samples.csproj
+++ b/src/devices/Ds1307/samples/Ds1307.Samples.csproj
@@ -6,7 +6,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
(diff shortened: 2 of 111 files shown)


Directory.Build.props
```diff
@@ -6,6 +6,7 @@
   <PropertyGroup>
     <Copyright>$(CopyrightNetFoundation)</Copyright>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <MainLibraryPath>$(MSBuildThisFileDirectory)src/System.Device.Gpio/</MainLibraryPath>
```
krwq: why not define path directly to csproj?

joperezr: We don't want to keep a copy around on every single .csproj we have, since in case we ever decide to restructure the repo, it would mean we would have to edit this property everywhere. This way, we only have it declared in a central place, so modifying it in the future is trivial.

src/devices/Ds1307/samples/Ds1307.Samples.csproj
```diff
@@ -6,7 +6,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Device.Gpio" Version="$(SystemDeviceGpioPackageVersion)" />
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
+      <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
```
krwq: I'm still confused why is `RuntimeIdentifier=linux` needed

krwq:
solves the problem so fine to merge. Any feedback and improvements we can merge later.

src/devices/Ds1307/samples/Ds1307.Samples.csproj
joperezr: I logged an issue to fix that in the future: #623

PR #624, opened 2019-07-23
Fixing symbol packages so that they contain the expected pdbs in the right locations

joperezr:
This replaces #587 

This is now producing the right symbol packages with the pdbs located where they should be. Next step will be to also add publishing to symbol servers.

cc: @safern @ericstj


joperezr:
cc: @safern and @riarenas for Yml changes.

joperezr:
I'll go ahead and merge this in order to fix other intermittent build failures. Will address any issue in a subsequent PR.

Issue #632, opened 2019-07-26
RTC devices not setup for device-listing


shaggygi:
I haven't looked in any detail, but it seems like the recent changes/addition to [Rtc binding folder](https://github.com/dotnet/iot/tree/master/src/devices/Rtc) doesn't work with [device-listing](https://github.com/dotnet/iot/tree/master/tools/device-listing) generation now.

krwq:
let me re-gen and see what's going on

krwq:
@shaggygi device-listing is not being auto-generated - you have to manually run the tool (I've found manual better so far since it gives me a chance to review the changes). When I run `dotnet run` on device-listing dir I'm seeing:

```
Warning: /home/krwq/work/iot/src/devices/SoftwareSpi/category.txt is missing
Warning: /home/krwq/work/iot/src/devices/Rtc/category.txt is missing
```

so that seems to be a culprit - I'll fix it and send PR with the changes

shaggygi:
Yeah, i noticed that, which is a nifty utility.  It seemed like it wasn’t generating for Clocks section.

PR #633, opened 2019-07-26
Regenerate device listing

krwq:
Fixes: https://github.com/dotnet/iot/issues/632

cc: @shaggygi


tools/device-listing/device-listing.cs
```diff
@@ -46,6 +46,7 @@ class Program
             "joystick",
             "color",
             "led",
+            "spi",
```
krwq: note that this is not for devices using SPI, rather for stuff providing SPI (not sure if this category is useful for anything else than software SPI implementation but since it's not in System.Device.Gpio probably better to put it there for better visibility)

joperezr: It might be useful as well for spi bus expanders that work over i2c. I believe that is a thing although I don't have an example of such devices.

krwq: we'll figure it out over time :smile:

Issue #626, opened 2019-07-24
Property vs Function for reading: Update guideliness and device bindings


krwq:
One of the offline feedback I got recently is that property when accessed twice in the row it should always produce the same value. We currently use properties for reading (i.e. Temperature) and that is not ture.

We should convert those to `ReadTemperature` - that should also make it easier to create async version of the code since the naming will be more consistent.

- [ ] Change all properties to `Read*` methods
  - [ ] Temperature
  - [ ] (placeholder for remainder: Pressure, ...)
- [ ] Make sure all async methods follow same naming convention (`Read*Async`)
- [ ] Update guideliness docs and all samples/readmes

FYI: @shaggygi @ZhangGaoxing @Ellerbach @richlander 
cc: @bartonjs for guideliness comments

bartonjs:
Yeah, the guidelines are there for it, they're just not cleanly in a bullet, so they didn't get transcribed.

Framework Design Guidelines, 2nd Edition, 5.1 p135-136:

* DO use a method, rather than a property, in the following situations:
  * The operation is orders of magnitude slower than a field access would be.
  * If you are even considering providing an asynchronous version of an operation to avoid blocking the thread [...].
...
  * The operation is a conversion, such as Object.ToString method.
  * **The operation returns a different result each time it is called, even if the parameters don’t change.** [...]
  * The operation has a significant and observable side effect. [...]
...
  * The operation returns a copy of an internal state (this does not include copies of value type objects returned on the stack).
  * The operation returns an array.


(and other reasons later, like "if it can throw an exception", but that's stated inversely as "AVOID throwing exceptions from property getters.")

shaggygi:
I'm assuming our strategy on [getting GpioPin values from properties](https://github.com/dotnet/iot/blob/26e1b07557ce650d3110769262a27b50f38650be/src/devices/Mcp25xxx/Mcp25xxx.cs#L168) needs to be changed, as well?

krwq:
@shaggygi yes but we can change that separately

Ellerbach:
I understand the rational for traditional elements but for sensors, I don't really see a value of having both a function and a property. 
In the case of having both, if I understan,d correctly, a property to be a property should be Something like "LastTemperature" which will always return the last read temperature and it will be updated by a "ReadTemperature" async function. am I correct?
Also with all the sensors we currently have, it will be quitea lot of work to change in all the sensors.

bartonjs:
[...]

PR #635, opened 2019-07-29
Update Servo Motor Binding

shaggygi:
Updated and simplified the Servo Motor binding.  Also added testing.


src/devices/ServoMotor/README.md
```diff
@@ -0,0 +1,69 @@
+﻿# Servo Motor
+
+## Summary
+This is a generic binding to control many servo motors using a hardware or software `PwmChannel`.  Servo motors are usually based on a signal frequency of 50Hz.  They also require a minimum/maximum pulse width to determine the position.  The pulse width is generally between 1 and 2 milliseconds, where 1ms is approximately 0 degrees, 1.5ms is the rest position, and 2ms is 180 degrees.  This information can be found in each servo motor's datasheet.
```
krwq: add enter after `## Summary` - I found markdown often breaks if you don't put enter before and after certain things like headers, tables and code blocks

src/devices/ServoMotor/README.md
```diff
@@ -0,0 +1,69 @@
...
+[Sparkfun Servo Motor Accessories](https://www.sparkfun.com/categories/245)  
+
+## Binding Notes
+
+The `ServoMotor` binding offers an easy way to begin controlling a servo motor.  The quickest approach is to provide the `ServoMotor` object a `PwmChannel` using the default values for other optional arguments.
+
+```csharp
+// Example of hardware PWM using chip 0 and channel 0 on a dev board.
+ServoMotor servoMotor = new ServoMotor(PwmChannel.Create(0, 0, 50));
+servoMotor.Start();  // Enable control signal.
```
krwq: what is this actually starting? will servo motor do anything after this call?

src/devices/ServoMotor/README.md
```diff
@@ -0,0 +1,69 @@
...
+servoMotor.WritePulseWidth(2000); // 2ms; Approximately 180 degrees.
+
+servoMotor.Stop(); // Disable control signal.
+```
+
+The position of servo motor can also be adjusted by the angle.  The `ServoMotor` constructor's optional arguments must be set according to device's specs.  NOTE: These are usually an approximation, so you may need to manually tweak to determine exact values.
+
+For example, the [Hitec HS-300BB](https://servodatabase.com/servo/hitec/hs-300bb) servo has the following specifications:
+- MaximumAngle = 180
+- MinimumPulseWidthMicroseconds = 900
```
krwq: How can pulse width length be less than PWM period (`1/frequency=1/50=200ms`)? I find this name rather confusing

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,133 @@
...
+    /// <summary>
+    /// Represents a rotary actuator or linear actuator that allows for precise control of angular or linear position.
+    /// </summary>
+    public class ServoMotor : IDisposable
+    {
+        private PwmChannel _pwmChannel;
+        private readonly int _maximumAngle;
+        private readonly int _maximumPulseWidthMicroseconds;
+        private readonly int _minimumPulseWidthMicroseconds;
+        private readonly double _servoAngleSlope;
```
krwq: this name is also a bit confusing: In linear equations `slope=tg(alpha)` where alpha is an angle - so I cannot see how angle and slope be in the name. Also what is this slope of? What is the function and variable here?

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,133 @@
...
+
+namespace Iot.Device.ServoMotor
+{
+    /// <summary>
+    /// Represents a rotary actuator or linear actuator that allows for precise control of angular or linear position.
+    /// </summary>
+    public class ServoMotor : IDisposable
+    {
+        private PwmChannel _pwmChannel;
+        private readonly int _maximumAngle;
```
krwq: minimum is zero? is this in degrees, radians, grad, some other unit? Why is this an int and not say double? Some comment would be useful here

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,133 @@
...
+        private readonly int _maximumPulseWidthMicroseconds;
+        private readonly int _minimumPulseWidthMicroseconds;
+        private readonly double _servoAngleSlope;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServoMotor"/> class.
+        /// </summary>
+        /// <param name="maximumAngle">The maximum angle the servo motor can move.</param>
+        /// <param name="minimumPulseWidthMicroseconds">The minimum pulse width, in microseconds, that represent an angle for 0 degrees.</param>
+        /// <param name="maximumPulseWidthMicroseconds">The maxnimum pulse width, in microseconds, that represent an angle for maximum angle.</param>
```
krwq: 1. typo in `maximum`
2. `that represent an angle for maximum angle` - huh?

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        /// <summary>
+        ///  Initializes a new instance of the <see cref="ServoMotor"/> class.
+        /// </summary>
+        /// <param name="chip">The PWM chip number used to control the servo motor.</param>
+        /// <param name="channel">The PWM channel number used to control the servo motor.</param>
+        /// <param name="maximumAngle">The maximum angle the servo motor can move.</param>
+        /// <param name="minimumPulseWidthMicroseconds">The minimum pulse width, in microseconds, that represent an angle for 0 degrees.</param>
+        /// <param name="maximumPulseWidthMicroseconds">The maxnimum pulse width, in microseconds, that represent an angle for maximum angle.</param>
+        public ServoMotor(
```
krwq: I'd drop this overload - we don't have similar convenience methods for SPI/I2C devices (i.e. we take I2cDevice directly and not I2cSettings)

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,133 @@
...
+                throw new ArgumentOutOfRangeException(nameof(minimumPulseWidthMicroseconds), minimumPulseWidthMicroseconds, "Value must not be negative.");
+            }
+
+            if (maximumPulseWidthMicroseconds < minimumPulseWidthMicroseconds)
+            {
+                throw new ArgumentOutOfRangeException(nameof(maximumPulseWidthMicroseconds), maximumPulseWidthMicroseconds, $"Value must be greater than or equal to {minimumPulseWidthMicroseconds}.");
+            }
+
+            // slope = (y2 - y1) / (x2 - x1).  x1 is 0 in this case, so it is not subtracted.
+            double servoAngleSlope = (double)(maximumPulseWidthMicroseconds - minimumPulseWidthMicroseconds) / maximumAngle;
```
krwq: I feel I'd prefer this to be called `dPulseWidthOverDAngle` so at least differential is clear

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,133 @@
...
+            return servoAngleSlope;
+        }
+
+        /// <summary>
+        /// Starts the servo motor.
+        /// </summary>
+        public void Start() => _pwmChannel.Start();
```
krwq: I feel like this should be in the constructor. Same we don't have `Thermometer.Start` which constructs I2cDevice or some other micro optimizations with unclear usage

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,133 @@
...
+                throw new ArgumentOutOfRangeException(nameof(angle), angle, $"Value must be between 0 and {_maximumAngle}.");
+            }
+
+            WritePulseWidth((int)(_servoAngleSlope * angle + _minimumPulseWidthMicroseconds));
+        }
+
+        /// <summary>
+        /// Writes a pulse width to the servo motor.
+        /// </summary>
+        /// <param name="microseconds">The pulse width, in microseconds, to write to the servo motor.</param>
```
krwq: I feel like I don't understand this description even after reading the code - you seem to be changing fill of the PWM, not really writing anything, also not sure why is this public

krwq: I might be missing how this device works, are you settting an angle by setting PWM fill? if that's the case I'd call it `SetAngle`

src/devices/ServoMotor/ServoMotor.cs
krwq: After reading the rest of the code perhaps better name would be `angleToTimeRatio`

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,133 @@
...
+        public void WritePulseWidth(int microseconds)
+        {
+            double dutyCyclePercentage = (double)microseconds / 1_000_000 * _pwmChannel.Frequency; // Convert to seconds 1st.
```
krwq: would it make sense to keep the time in units so that conversion wouldn't be necessary?

src/devices/ServoMotor/samples/README.md
```diff
@@ -0,0 +1,20 @@
...
+**NOTE**: The following image shows the servo is connected to a regular GPIO and not a hardware PWM pin.
+
+![schema](./servomotor.png)
+
+## Sample Code
+
+The sample code provides helpful examples where you can control the servo motor's position by pulse width or angle.
+
+### Writing Pulse Width
+This example allows you to manually view the full left and right positions based on the pulse width signal entered.  Many times, servo motors do not match the exact range based on datasheets and/or have to be calibrated using the screw on side of servo motor.
```
krwq: this should be in the comment above WritePulseWidth (specifically that this method is used for calibration)

src/devices/ServoMotor/ServoMotor.cs
krwq: ok, I think I understand why this is public (calibration) - other part of the comment stays

krwq:
@shaggygi I was trying to review this commit by commit to see what exactly got changed here but I'm still not sure so I did the full review

shaggygi:
> I was trying to review this commit by commit to see what exactly got changed here but I'm still not sure so I did the full review

Sorry about that it.  I was trying to break it down as much as possible, but this got gutted pretty bad throughout changes.

src/devices/ServoMotor/ServoMotor.cs
shaggygi: In next commit

src/devices/ServoMotor/README.md
shaggygi: Gotcha.  I like that better (with space), but did a verify on other bindings.  We'll probably need to do a future scan to update those.

This is updated in next commit.

src/devices/ServoMotor/README.md
shaggygi: Start() method just starts the pulse width on channel that sends signal to servo motor.  The motor will adjust to position based on current frequency and duty cycle.  Once started, the properties can be changed and position will adjust without having to call Start() again.

The signal could automatically start (when instantiating).  However, it is similar to the PwmChannel, where a user might want to create the object first without sending pulses and then start/stop to do things without disposing.  This would be critical in some scenarios and bad things could happen if it was coded using some of the default frequencies & duty cycles.  For example... a model airplane with servo throttle 🚀

src/devices/ServoMotor/README.md
shaggygi: > How can pulse width length be less than PWM period

I'm not sure I understand question.  The PWM period is the entire time it takes for one cycle of frequency.  The pulse width is the duration the pulse is active.  As in your equation, servo frequencies are usually 50Hz so the period is 20ms.

The min/max widths represents the pulse width needed to position.  Properties are in microseconds, where min is 900uS (or 0.9mS) and min is 2100uS (or 2.1mS).  

image

image

src/devices/ServoMotor/ServoMotor.cs
shaggygi: Yeah, I just used name that first came to mind.  This is updated in future commit based on other feedback.

src/devices/ServoMotor/ServoMotor.cs
shaggygi: `maximumAngle` represents how far the position can travel (from 0 - maximum angle).  It will be 180 degrees in most cases, but I've seen 120/270/360 degrees.  This also allows users to set a smaller range within the define spec if they wanted.  Meaning, they might have a servo that can move 180 degrees, but only want to move with 90 degrees.

In addition, they might want the 90 degrees be a certain range within the 180 degrees.  Meaning, the angle could be set to use 0 - 90 values to represent 45 - 135 values.  They would just need to set the min/max pulse parameters accordingly.

src/devices/ServoMotor/ServoMotor.cs
shaggygi: I thought about that when we worked on the new `PwmChannel`, where the signal automatically starts when creating the object.  However, users will want to start/stop on demand without having to dispose and recreate the object.  The signals can still be adjusted while it is enabled/started, but there will be times they need to stop the signal while working on things, make other adjustments (not just for the PWM signal) and start signals back.

Users could damage hardware and what they are controlling if not careful.  For example, they could use the default PwmChannel and not specific the frequency (400Hz) and duty cycle (50%).  Let's say they had it connected to throttle on remote control lawn mower.  If they created servo motor and it automatically started... bad things could happen 😄

src/devices/ServoMotor/ServoMotor.cs
shaggygi: > if that's the case I'd call it `SetAngle` or `SetDesiredAngle`

Yes, this is basically setting the active pulse width for signal.  You change this to change the position of motor.

I thought about using  `Set`, but thought `Write` was more appropriate related to the #626 discussions going on.

To me, `Set`/`Get` should be used for property setters/getters (at least when accessing non-public/writable things) and `Write`/`Read` should be used when updating binding stuff over the wire.  This can be updated based on #626 discussion results.

src/devices/ServoMotor/ServoMotor.cs
shaggygi: As far as how the code works, you pass in a value on how long you want the pulse width to be during the cycle period.  This usually ranges from 0.001s/1ms/1000us to 0.002s/2ms/2000us.

This is sort of a bigger discussion to keep consistent (at least with some of the PWM stuff) as units of time will vary how datasheets reference based on the hardware, lower-level APIs, etc.  For example, Linux uses Nanoseconds, servo motor datasheets reference pulse width ranges in microseconds, servo tutorials explain with milliseconds.  Blame, why can't the world be in sync?!?! 😄 

I thought about using the pulse width parameter in seconds as a `double`, but might confuse some users.  Use 900 microseconds as an example.

```csharp
// In microseconds as int.
public void WritePulseWidth(900) { ... }

// In seconds as double.
public void WritePulseWidth(0.000900) { ... }
```

Using seconds and decimal places might cause a user to miss the correct period location compared to using the value provided by datasheet in microseconds.

Long story short, I can change to use seconds as a double and remove conversion if you think that is a better approach.   Still gives same result.

```console
50Hz; 0.020s or 20000us
Setting pulse width of 0.0015s or 1500us

double dutyCyclePercentage = (double)microseconds / 1_000_000 * _pwmChannel.Frequency;

Seconds
0.075 = 0.0015 * 50

Microseconds
0.075 = (double)1500 / 1_000_000 * 50
```

src/devices/ServoMotor/ServoMotor.cs
shaggygi: The angle is usually represented in degrees, so maybe I don't have the correct naming for variables.  Maybe should change to `angleDegrees` like we did with `pulseWidthMicroseconds`.  I think using degrees (versus 0 - 1 values) is a little better as it gives a better picture of how far you're actually changing the position.

src/devices/ServoMotor/ServoMotor.cs
shaggygi: In next commit

src/devices/ServoMotor/samples/README.md
shaggygi: > this should be in the comment above WritePulseWidth

Do you mean add to remarks of method summary?  That would probably be a good idea.

I don't want to give the impression that `WritePulseWidth` method should only be used for calibration as some users might base their apps on pulse width values instead of angles (in degrees).  Having the two methods offers them options.

src/devices/ServoMotor/README.md
krwq: LOL, seems I had a bad idea on how PWM works - I thought the frequency was relating to granularity of the fill and 50% with 50Hz frequency means sending 10101... over the wire and swap value every 20ms but that doesn't seem to be the case. Now I'm wondering if we should have an API telling us on how accurate can we be with pulse width - so basically clock frequency or whatever it is called.

src/devices/ServoMotor/ServoMotor.cs
krwq: > The angle is usually represented in degrees

@shaggygi I tend to disagree - I don't think we ever used degrees when I was studying - it was always radians. Same in graphics - only radians (except maybe UI). The reason is that it combines really well with the rest of the math: i.e you can pass it directly to Math.Sin or calculate physical distance traversed without doing extra conversions.

I guess I don't care that much either way but for public APIs units should be very clear and ideally added in the name (i.e. MoveByDegrees). Other option we can add another struct in the `Units` project then problem will be solved.

src/devices/ServoMotor/README.md
krwq: I was actually trying to get rid of Start/Stop on PWM because PWM with 0% fill is the same as stopped PWM but seems we might be too far in the release cycle for that at the moment.

Presumably when you don't start PWM then PWM fill will be 0 or 1 (or basically all high or low).

Would that have any effect on the servo? I.e. if servo was set at `MinAngle + 30% * AngleRange` before it was powerd up and I powered it up to line with control signal with 0% fill would that have any effect on servo? (i.e. servo going to MinAngle)

If Start/Stop have observable difference for user I'm fine with having them but if no movement is equivalent to some setting I'd rather not have it.

src/devices/ServoMotor/ServoMotor.cs
krwq: either way - if unit is not clear it should be in the name or the comment

src/devices/ServoMotor/ServoMotor.cs
krwq: @shaggygi I feel like we should call it relative or normalized angle at this point and let it be value from 0-1 - it's usually much easier to operate on: i.e. if you are controlling fuel injection then 1 would be 100%, in any other case where you need specific unit you can just add `MinAngle + RelativeAngle * AngleRange` <-- note how this terminology is much easier to understand - I'm fine with having additional Function/Property returning such result as well

src/devices/ServoMotor/ServoMotor.cs
krwq: @shaggygi as I previously wrote: for PWM I don't think Start/Stop are justified because 0% fill is equivalent to not starting. In here 0% fill I think it might cause servo not to move in which case Start/Stop are justified: Start causes servo to start moving into the set position and Stop stops any movement. 0% would mean: move to the left vs Stop means don't move at all in which case this sounds justify - I just need to hear the confirmation that is the case 😄

src/devices/ServoMotor/ServoMotor.cs
krwq: @shaggygi the difference here is that in control engineering `set position` refers to desired angle. Write suggest you transmit some data somewhere which is not the case here. For sensors you usually read the data from the sensor or write some data to a sensor - in here we literally want to set a position of the servo and hence I think it should have Set in the name. Since the action actually starts the movement and not immediately sets the position (even though it might seem it is immediate) I'd prefer to have some words suggesting that you are about to start the movement into a desired position. PulseWidth basically doesn't mean anything for users who don't know internals on how servo works while SetDesiredAngle has correlation with physical properties which are immediately obvious

src/devices/ServoMotor/ServoMotor.cs
krwq: if you call this relativeAngle which is value from 0 to 1 and allow to modify MinPulseWidth and MaxPulseWidth and AngleRange (and possibly MinAngle) then I feel like those helpers won't be longer necessary and everything stays clear on how it works. Make sure to match names with whatever manufacturers use

src/devices/ServoMotor/samples/README.md
krwq: @shaggygi what would be the usage for using pulse width? I don't see that there is any use other than calibration. Worst case scenarios users can use PWM directly if they for whatever reason need to have more control

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// <summary>
+    /// Represents a rotary actuator or linear actuator that allows for precise control of angular or linear position.
+    /// </summary>
+    public class ServoMotor : IDisposable
+    {
+        private PwmChannel _pwmChannel;
+        private readonly int _maximumAngle;
+        private readonly int _maximumPulseWidthMicroseconds;
+        private readonly int _minimumPulseWidthMicroseconds;
+        private readonly double _angleInMicroseconds;
```
krwq: I think you meant angleToMicroseconds - InMicroseconds kinda suggests that angle is measured in microseconds which is not the case

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServoMotor"/> class.
+        /// </summary>
+        /// <param name="maximumAngle">The maximum angle the servo motor can move.</param>
+        /// <param name="minimumPulseWidthMicroseconds">The minimum pulse width, in microseconds, that represent an angle for 0 degrees.</param>
+        /// <param name="maximumPulseWidthMicroseconds">The maxnimum pulse width, in microseconds, that represent an angle for maximum angle.</param>
+        private ServoMotor(int maximumAngle, int minimumPulseWidthMicroseconds, int maximumPulseWidthMicroseconds)
```
krwq: the more I look at it the more I think we should add new unit for angles with internal representation being radians. The only trick part is angles are in modulo space and -180..180 are relatively commonly used so we should probably create an issue on how to design it

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,114 @@
...
+            PwmChannel pwmChannel,
+            int maximumAngle = 180,
+            int minimumPulseWidthMicroseconds = 1_000,
+            int maximumPulseWidthMicroseconds = 2_000)
+            : this(maximumAngle, minimumPulseWidthMicroseconds, maximumPulseWidthMicroseconds)
+        {
+            _pwmChannel = pwmChannel;
+        }
+
+        private static double CalculateAngleToMicroseconds(int maximumAngle, int minimumPulseWidthMicroseconds, int maximumPulseWidthMicroseconds)
```
krwq: you even got `To` in the name here so definitely you meant `To` in variable name as well

src/devices/ServoMotor/README.md
shaggygi: > we should have an API telling us on how accurate can we be with pulse width

That is basically what PWM is all about.  The frequency states how many times a cycle can occur in a second and the duty cycle states how long the pulse is active during each cycle.

So if you have a 1Hz signal (1 cycle a second) and a duty cycle of 50%, the pulse width would be 0.5s/500ms/500000us.

In this Servo Motor binding update, WritePulseWidth is setting how accurate the pulse should be active.  You could technically move this method to PwmChannel classes (and I started to), but thought it would be a longer review/discussion on having it added to core API.

😄 sometimes i hate technology

src/devices/ServoMotor/README.md
shaggygi: > Presumably when you don't start PWM then PWM fill will be 0 or 1 (or basically all high or low).

This would be a bad move for a few reasons.  Below are some I could think of...
- Some PWM devices don't allow for 100% duty cycle.  For example, I've seen some potential binding devices that throw errors when writing 100% DC to registers.  I'm not sure, but might be case for 0%, as well.  Those same bindings also have flags in registers to enable/disable the channels' signals.
- Users would have to remember to write 0/1 to disable and then remember the value they were using beforehand (e.g. 0.452, etc.).  Stop/Start appears better and you don't have to remember as the device memory does it for you automatically.
- Start/Stop seems to be used in other IoT APIs including WinRT.  Maybe we could use Enable/Disable.  I don't want to get hung on a particular name (but be similar across PWM APIs; core, motors, buzzers, LEDs,, etc.).  I'm just more set on having the API ability to create the PWM related objects without starting signal immediately and then being able to enable/disable the signal when ready.

src/devices/ServoMotor/README.md
krwq: yep, unfortunately issue first for System.Device.Gpio at least :smile:

src/devices/ServoMotor/ServoMotor.cs
shaggygi: @krwq The majority of other servo related APIs i came across use angle in the wording and degrees as the units.  [Here](https://wpilib.screenstepslive.com/s/currentCS/m/java/l/599703-repeatable-low-power-movement-controlling-servos-with-wpilib) is an example.

Maybe we include both SetScale (for 0 - 1) and SetAngle (for degrees).

src/devices/ServoMotor/ServoMotor.cs
shaggygi: @krwq  Here are some other library examples that offer similar for start/stop.

https://www.arduino.cc/en/Reference/ServoAttach
https://rpi.science.uoit.ca/lab/servo/#example-of-controlling-a-servo-from-the-raspberry-pi

src/devices/ServoMotor/ServoMotor.cs
shaggygi: maybe we just name parameter degrees and _maximumDegrees

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// <param name="pwmChannel">The PWM channel used to control the servo motor.</param>
+        /// <param name="maximumAngle">The maximum angle the servo motor can move.</param>
+        /// <param name="minimumPulseWidthMicroseconds">The minimum pulse width, in microseconds, that represent an angle for 0 degrees.</param>
+        /// <param name="maximumPulseWidthMicroseconds">The maxnimum pulse width, in microseconds, that represent an angle for maximum angle.</param>
+        public ServoMotor(
+            PwmChannel pwmChannel,
+            int maximumAngle = 180,
+            int minimumPulseWidthMicroseconds = 1_000,
+            int maximumPulseWidthMicroseconds = 2_000)
+            : this(maximumAngle, minimumPulseWidthMicroseconds, maximumPulseWidthMicroseconds)
```
joperezr: You might have answer this already since I'm joining the review a bit late, but why do we even want to have a private constructor that is only called from here? Why not just move the code inside that private constructor to be done in this public one?

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            _pwmChannel = pwmChannel;
+        }
+
+        private static double CalculateAngleToMicroseconds(int maximumAngle, int minimumPulseWidthMicroseconds, int maximumPulseWidthMicroseconds)
+        {
+            if (maximumAngle < 0 || maximumAngle > 360)
```
joperezr: NIT: I know that this method is only called once during construction, so this might be a micro optimization that is not worth it, but you could perform both of this checks in one operation if you did: `if (((uint)maximumAngle) > 360)` basically by casting you would be making maximumAngle to be a very large number so it would be caught by the check if the number is > 360.

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,114 @@
...
+namespace Iot.Device.ServoMotor
+{
+    /// <summary>
+    /// Represents a rotary actuator or linear actuator that allows for precise control of angular or linear position.
+    /// </summary>
+    public class ServoMotor : IDisposable
+    {
+        private PwmChannel _pwmChannel;
+        private readonly int _maximumAngle;
+        private readonly int _maximumPulseWidthMicroseconds;
```
joperezr: I don't think you are using this guy any longer, so it could be removed.

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// </summary>
+        public void Stop() => _pwmChannel.Stop();
+
+        /// <summary>
+        /// Writes an angle to the servo motor.
+        /// </summary>
+        /// <param name="angle">The angle to write to the servo motor.</param>
+        public void WriteAngle(int angle)
+        {
+            if (angle < 0 || angle > _maximumAngle)
```
joperezr: Same comment about casting to uint to make the call faster.

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,114 @@
...
+
+using System;
+using System.Device.Pwm;
+
+namespace Iot.Device.ServoMotor
+{
+    /// <summary>
+    /// Represents a rotary actuator or linear actuator that allows for precise control of angular or linear position.
+    /// </summary>
+    public class ServoMotor : IDisposable
```
joperezr: Should class be sealed? if not, and you expect people to derive from this, then we should probably implement the dispose pattern.

src/devices/ServoMotor/ServoMotor.cs
```diff
@@ -0,0 +1,114 @@
...
+        private PwmChannel _pwmChannel;
+        private readonly int _maximumAngle;
+        private readonly int _maximumPulseWidthMicroseconds;
+        private readonly int _minimumPulseWidthMicroseconds;
+        private readonly double _angleInMicroseconds;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServoMotor"/> class.
+        /// </summary>
+        /// <param name="maximumAngle">The maximum angle the servo motor can move.</param>
```
joperezr: I assume that the range of this is 0-360, but could we call that out explicitly in the docs, same goes for the other params, in case a range applies.

src/devices/ServoMotor/ServoMotor.cs
shaggygi: Originally had 2 public constructors that call this private one.  1 with channel and other with chip/channel.  I'll remove in next commit.

src/devices/ServoMotor/ServoMotor.cs
shaggygi: Will update in next commit

src/devices/ServoMotor/ServoMotor.cs
shaggygi: Will update in next commit

src/devices/ServoMotor/ServoMotor.cs
shaggygi: Correct.  Will update in next commit.

src/devices/ServoMotor/ServoMotor.cs
shaggygi: Will update in next commit

src/devices/ServoMotor/ServoMotor.cs
shaggygi: Will update in next commit.

src/devices/ServoMotor/ServoMotor.cs
shaggygi: Updated to sealed.  This should cover most scenarios for sweep and continuous types.

krwq:
let me re-review

PR #636, opened 2019-07-29
Fixed naming for seconds

shaggygi:
Fixed capitalizing of seconds.


shaggygi:
That's what I thought when originally adding, but noticed the following later...

- https://en.wikipedia.org/wiki/Microsecond
- TimeSpan.FromMilliseconds
- DateTime.Now.Millisecond

Issue #638, opened 2019-07-29
Make hardware PWM more responsive


krwq:
Last time I used hardware PWM it had a bit of a lag when changing duty cycle - that made it basically unusable for many real-time scenarios like controlling the vehicle where such lag was causing wheels to turn (i.e. SoftPWM on one wheel reacted faster so effectively it caused vehicle to turn).

This needs to be investigated in 3.0 timeline.

joperezr:
Just a note, the only thing here that is 3.0 is the investigation. If there are any action items out from that, this won't need to happen necessarily for 3.0.

PR #640, opened 2019-07-29
Avoid openning file handle when changing duty cycle or freq in hardware PWM

krwq:
Fixes: https://github.com/dotnet/iot/issues/638

I noticed that depending on the battery level PWM is more or less responsive in general but this allows to be more consistent since we don't have to open the file on every operation.

Most of the cases only duty cycles will be changed frequently but some other like buzzer the frequency is changed relatively quickly as well

### DIFF
--- a/src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
@@ -41,7 +41,7 @@ namespace System.Device.Pwm.Channels
             Validate();
             Open();
 
-            // avoid openning the file for operations changing relatively frequently
+            // avoid opening the file for operations changing relatively frequently
             _dutyCycleWriter = new StreamWriter(new FileStream($"{_channelPath}/duty_cycle", FileMode.Open, FileAccess.ReadWrite));
             _frequencyWriter = new StreamWriter(new FileStream($"{_channelPath}/period", FileMode.Open, FileAccess.ReadWrite));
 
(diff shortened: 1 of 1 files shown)


src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -38,6 +40,11 @@ internal class UnixPwmChannel : PwmChannel
             _channel = channel;
             Validate();
             Open();
+
+            // avoid openning the file for operations changing relatively frequently
```
shaggygi: NIT: opening

krwq:
```suggestion
            // avoid opening the file for operations changing relatively frequently
```

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -38,6 +40,11 @@ internal class UnixPwmChannel : PwmChannel
             _channel = channel;
             Validate();
             Open();
+
+            // avoid opening the file for operations changing relatively frequently
+            _dutyCycleWriter = new StreamWriter(new FileStream($"{_channelPath}/duty_cycle", FileMode.Open, FileAccess.ReadWrite));
```
joperezr: Does the fileStream need to be disposed as well or does it get disposed by the streamwriter?

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -184,6 +195,10 @@ public override void Stop()
 
         protected override void Dispose(bool disposing)
         {
+            _dutyCycleWriter?.Dispose();
```
joperezr: Will this also call Close on the FileStream?

krwq: yes

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: streamwriter disposes it

Issue #641, opened 2019-07-30
Simplify DCMotor abstraction


krwq:
DCMotor is one of the devices we will declare as more stable for 3.0 and therefore it needs to be cleaned up to align with our guidelines better.

- [X] Name
  - Should this be called DCMotor <s>or DcMotor or differently</s>
- [ ] https://github.com/dotnet/iot/pull/643 All different ways of connecting DCMotor need to be simpler to use`*`

`*` This is a bit longer so let's expand it

DC Motor can be connected using 1, 2 or 3 pins:
## 1 pin mode

One of the motor pins is connected to ground and other pin is PWM which controls the speed. Motor can move only in one direction

## 2 pin mode

One pin is PWM (similarly as in 1 pin mode) and controls the speed.
Second pin is either 0/1 depending on direction (when 1 then PWM duty cycle is negated since polarity is swapped).

1/2 pin modes can be used to connect small (!) motor directly to microcontroller (although it is always recommended to use H-bridge).

## 3 pin mode (or the most correct way)

DC motor has 2 pins but H-bridge which should be used to control the motor.
2 pins of H-bridge map almost directly to output pins (except voltage is higher on High state) and third pin is used as "Enable" which is meant for connecting PWM (using it in 2 pin mode might cause a bit more heat to be produced than enable pin)

This effectively makes it so that first two pins control direction (1/0 or 0/1) and PWM controls the speed.

## Design issue

Above setup is causing a bit of design pain as the abstraction is really awkward.

### Current state

DCMotorSettings represents how things are connected:

- Pin1, Pin2 tells if and how pins are connected
- PwmChannel tells how to control the speed
- UseEnableAsPwm tells if PWM is connected to Enable pin

drawback of that is that valid settings are not clear (and also it's confusing what the options actually mean)

### Proposed state

No DCMotorSettings.

(names might be a bit awkward but not sure how to call them better)

```csharp
class DCMotor
{
  // note that names of pin cannot be any better since the direction is arbitrary depending on how motor got connected
  public static DCMotor Create1Pin(PwmChannel pwmChannel);
  public static DCMotor Create2Pin(PwmChannel pwmChannel, int pin);
  
  // these could be called directionPin1, directionPin2 but then it would be inconsistent with above
  public static DCMotor Create3Pin(PwmChannel enablePinPwmChannel, int pin1, int pin2);
  // other option is to perhaps drop suffixes and just leave Create

  // options to control DC Motor truncated from here
}
```

[...]

Issue #478, opened 2019-06-06
Error when trying to set speed of L298N H-Bridge (question how to)


MaklaCof:
I am trying to do a "HelloWord"  asp.net core application. I would like to control motor speed through web page.

But I don't know how to properly set motor speed. My code is:
```
        public async Task SetSpeed(int motor, double speed)
        {
            var sett = new DCMotorSettings()
            {
                Pin0 = 23,
                Pin1 = 24,
                PwmController = new PwmController(new SoftPwm()),
                PwmChip = 25,
                //PwmChannel = 1, // use for hardware PWM
                PwmFrequency = 50, // optional, defaults to 50
            };
            using (var m = DCMotor.Create(sett))
            { 
                m.Speed = speed / 100;
                Console.WriteLine($"Set speed of motor 1 to {m.Speed}.");
                await Task.Delay(3000);
            }
        }
```
With pin `23`, `24` and `25` I get error:

> Set speed of motor 1 to 0.6.
> dotnet :
>     + CategoryInfo          : NotSpecified: (:String) [], RemoteException
>     + FullyQualifiedErrorId : NativeCommandError
> 
> Unhandled Exception:
> 
> System.InvalidOperationException: Can not write to a pin that is not open.
>    at System.Device.Gpio.GpioController.Write(Int32 pinNumber, PinValue value)
>    at System.Device.Pwm.Drivers.SoftPwm.RunSoftPWM()
>    at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
>    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback,
> Object state)
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback,
> Object state)
>    at System.Threading.ThreadHelper.ThreadStart()
> 

I also try to set pin `33`, `35`, `37`, but I get

> System.Exception: Element not found.
> 
> Pin ' is not available. It is reserved by the system or in use for another function.
>    at Windows.Devices.Gpio.GpioController.OpenPin(Int32 pinNumber, GpioSharingMode sharingMode)
>    at System.Device.Gpio.Drivers.Windows10Driver.OpenPin(Int32 pinNumber)
>    at System.Device.Gpio.GpioController.OpenPin(Int32 pinNumber)
[...]

PR #643, opened 2019-07-30
Simplify DCMotor abstraction

krwq:
Fixes: #641
Fixes: #478

See details in #641

Also fixed the dispose issue where DCMotor in 2/1 pin mode is using software PWM with the same controller and controller gets disposed before software PWM stops writting (which surfaces as exception as described in #478) - noticed when testing.

cc: @shaggygi @MaklaCof


krwq:
@joperezr, is there anything blocking here?

joperezr:
haven't had a chance to look at this yet, I should have some time before EOD

src/devices/DCMotor/DCMotor.cs
```diff
@@ -41,58 +41,130 @@ public virtual void Dispose()
         }
 
         /// <summary>
-        /// Constructs DCMotor class.
-        /// Depending on the settings it will pick implementation suitable for the given connection.
+        /// Creates DCMotor instance which allows to control speed in one direction.
         /// </summary>
-        public static DCMotor Create(DCMotorSettings settings)
+        /// <param name="speedControlChannel">PWM channel used to control the speed of the motor</param>
```
joperezr: NIT: use CRefs in comments.

src/devices/DCMotor/DCMotor.cs
```diff
@@ -41,58 +41,130 @@ public virtual void Dispose()
...
-                        throw new ArgumentException("Pin0 or Pin1 must be set when PWM is not specified.");
-                    }
-
-                    return new DCMotor2PinNoEnable(new SoftwarePwmChannel(pin0.Value, 50, 0.0, controller: settings.Controller), pin1, settings.Controller);
-                }
-            }
+            if (speedControlChannel == null)
+                throw new ArgumentNullException(nameof(speedControlChannel));
+
+            return new DCMotor2PinNoEnable(speedControlChannel, -1, null);
```
joperezr: So if you do this, seems like the DCMotor will have both a PWMChannel and a GpioController right? given that eventually the DCMotor2PinNoEnable class will call base constructor with null which will create a new GpioController. Is this controller used at all? If not then why not change the base constructor such that if null gets passed in then we don't create a new GpioController?

src/devices/DCMotor/DCMotor.cs
```diff
@@ -41,58 +41,130 @@ public virtual void Dispose()
...
+        /// <param name="controller">GPIO controller related to the pin</param>
+        /// <returns>DCMotor instance</returns>
+        public static DCMotor Create(int speedControlPin, GpioController controller = null)
+        {
+            if (speedControlPin == -1)
+                throw new ArgumentOutOfRangeException(nameof(speedControlPin));
+
+            controller = controller ?? new GpioController();
+            return new DCMotor2PinNoEnable(
+                new SoftwarePwmChannel(speedControlPin, 50, 0.0, controller: controller),
```
joperezr: This is a bit confusing as well, so you are creating a DCMotor that is driven by software pwm in a specified controller, but then you are also passing that controller over to the DCMotor base. This means you could have the case where that guy is null, and you initialize two difierent instances of GpioControllers, one in the soft pwm and one in the dcmotor base class constructor.

src/devices/DCMotor/DCMotor.cs
```diff
@@ -41,58 +41,130 @@ public virtual void Dispose()
...
+                -1,
+                controller);
+        }
+
+        /// <summary>
+        /// Creates DCMotor instance which allows to control speed in both directions.
+        /// </summary>
+        /// <param name="speedControlChannel">PWM channel used to control the speed of the motor</param>
+        /// <param name="directionPin">Pin used to control the direction of the motor</param>
+        /// <param name="controller">GPIO controller related to the pin</param>
```
joperezr: NIT: use paramref to denote which pin are you refering to here.

src/devices/DCMotor/DCMotor.cs
joperezr: Why did we pick 50 hertz? Is this documented as the frequency that needs to be used?

src/devices/DCMotor/DCMotor.cs
```diff
@@ -41,58 +41,130 @@ public virtual void Dispose()
...
+        {
+            if (speedControlPin == -1)
+                throw new ArgumentOutOfRangeException(nameof(speedControlPin));
+
+            if (directionPin == -1)
+                throw new ArgumentOutOfRangeException(nameof(directionPin));
+
+            controller = controller ?? new GpioController();
+            return new DCMotor2PinNoEnable(
+                new SoftwarePwmChannel(speedControlPin, 50, 0.0, controller: controller),
```
joperezr: Same question about 50 hertz here. Should that be configurable?

src/devices/DCMotor/DCMotor.cs
```diff
@@ -41,58 +41,130 @@ public virtual void Dispose()
...
+                directionPin,
+                controller);
+        }
+
+        /// <summary>
+        /// Creates DCMotor instance which allows to control speed in both directions.
```
joperezr: From the summary description, this factory method seems to be the exact same thing as the previous factory method. Why would I want to pass in otherDirectionPin? What do I gain with using 3 pin instead of 2 pin?

src/devices/DCMotor/DCMotor2PinNoEnable.cs
```diff
@@ -77,9 +77,9 @@ public override double Speed
         }
         public override void Dispose()
         {
-            base.Dispose();
             _pwm?.Dispose();
             _pwm = null;
+            base.Dispose();
```
joperezr: This is wrong as it is not implementing the dispose pattern. We should instead have a public Dispose() only in the base class, which calls a protected virtual void Dispose(bool disposing) and that should be the one that is overriden in subclasses.

src/devices/DCMotor/README.md
```diff
@@ -9,8 +9,5 @@ DC motors are controlled by simply providing voltage on the inputs (inverted vol
 DC motors can be controlled with 1, 2 or 3 pins.
 Please refer to the [sample](samples/README.md) to see how to connect it.
 
-**Recommended setup is 3-pin mode, 1 and 2 pin mode may cause excessive heat or damage your H-bridge (larger engines are more likely to damage).**
```
joperezr: It would be nice to explain a bit more why is 3 pin better than 2. From what I can tell in the code and on the README it is not clear why I should pick one over the other.

src/devices/DCMotor/samples/Program.cs
```diff
@@ -14,53 +14,19 @@ namespace samples
...
-            DCMotorSettings settings = ThreePinModeHardware();
             Stopwatch sw = Stopwatch.StartNew();
-            using (DCMotor motor = DCMotor.Create(settings))
+            // 1 pin mode
+            // using (DCMotor motor = DCMotor.Create(6))
+            // using (DCMotor motor = DCMotor.Create(PwmChannel.Create(0, 0, frequency: 50)))
+            // 2 pin mode
+            // using (DCMotor motor = DCMotor.Create(27, 22))
+            // using (DCMotor motor = DCMotor.Create(new SoftwarePwmChannel(27, frequency: 50), 22))
+            // 3 pin mode
```
joperezr: perhaps also include an example here to use a provided gpio extender in order to show an example of passing in that IGpioController.

src/devices/DCMotor/DCMotor.cs
```diff
@@ -41,58 +41,130 @@ public virtual void Dispose()
...
+            return new DCMotor2PinNoEnable(speedControlChannel, -1, null);
+        }
+
+        /// <summary>
+        /// Creates DCMotor instance which allows to control speed in one direction.
+        /// </summary>
+        /// <param name="speedControlPin">Pin used to control the speed of the motor with software PWM</param>
+        /// <param name="controller">GPIO controller related to the pin</param>
+        /// <returns>DCMotor instance</returns>
+        public static DCMotor Create(int speedControlPin, GpioController controller = null)
```
joperezr: All of these GpioController parameters should be IGpioController instead.

src/devices/DCMotor/DCMotor.cs
```diff
@@ -41,58 +41,130 @@ public virtual void Dispose()
...
 
         /// <summary>
-        /// Constructs DCMotor class.
-        /// Depending on the settings it will pick implementation suitable for the given connection.
+        /// Creates DCMotor instance which allows to control speed in one direction.
         /// </summary>
-        public static DCMotor Create(DCMotorSettings settings)
+        /// <param name="speedControlChannel">PWM channel used to control the speed of the motor</param>
+        /// <returns>DCMotor instance</returns>
+        public static DCMotor Create(PwmChannel speedControlChannel)
```
joperezr: I suppose the idea of having factory methods is to hide from people which type (how many pins) of DCMotor are they getting?

src/devices/DCMotor/DCMotor.cs
krwq: there is only one, do we need to be that specific?

joperezr: the pwm channel will eventually also be using pins, so it is not really clear that the controller is supposed to be the provider of direction pin, so yes, I think having a paramref helps making that dependency clear.

src/devices/DCMotor/DCMotor.cs
krwq: that's not possible because of `controller = controller ?? new GpioController();` line above

krwq: it's a sane choice. All H-bridges I've got allow max of 50Hz. You never really should need more than that. I'd wait until someone complains before changing this since I expect it won't ever be an issue

joperezr: it is possible, because inside the SoftwarePwm channel you have the exact same line, where it will also initialize a controller if it gets passed a null, so you end up with 2 GpioControllers

src/devices/DCMotor/DCMotor.cs
krwq: I can add a constant but it is configurable if you pass your own SoftwarePwmChannel instance

src/devices/DCMotor/DCMotor.cs
krwq: Read README.md

src/devices/DCMotor/DCMotor.cs
joperezr: Given that this is one of the bindings that we will support, I don't think that we should base a sane choice based on the devices that you've got 😄. This should be configurable with a parameter given that there is no way to access this softwarepwmchannel after initializing from the DCMotor.

src/devices/DCMotor/README.md
krwq: do you mean to expand `1 and 2 pin mode may cause excessive heat or damage your H-bridge`? H-bridges are designed such that inputs should not change that quickly

src/devices/DCMotor/DCMotor.cs
joperezr: I did, it is not clear enough. It just say that we recommend 3-pin but doesn't state why. Also, we shouldn't expect consumers of this API to have to go read a README file in github. Instead they should be able to pick which API to call/use based on the xml docs that the API provides.

src/devices/DCMotor/samples/Program.cs
krwq: I don't have GPIO extender

src/devices/DCMotor/README.md
joperezr: yes, I mean that we should elaborate a bit more on that as in, why is 3 pin better, where are they used, how are the instructions passed to the device different, etc.

src/devices/DCMotor/samples/Program.cs
joperezr: That doesn't mean you can't add an example that passes in a Mcp23008 object just as an example of how that code would look like.

src/devices/DCMotor/DCMotor.cs
shaggygi: Seems like this would be named `DcMotor`

krwq: apparently both forms are acceptable per our guideliness

src/devices/DCMotor/DCMotor.cs
krwq: that's not the case in this case because local `controller` is not null after `controller = controller ?? new GpioController()` and both DCMotor2PinNoEnable and SoftwarePwmChannel will get that non-null value. The static Create doesn't call base class. I'll validate remaining Create methods to make sure that is not the case as well

krwq: Seems this was a case for:

```csharp
public static DCMotor Create(PwmChannel speedControlChannel, int directionPin, int otherDirectionPin, GpioController controller = null)
```

so possibly you commented under wrong overload

krwq: I'll add a documentation for 50Hz but you can override the choice by passing `PwmChannel` directly.

src/devices/DCMotor/samples/Program.cs
krwq: not in this PR

src/devices/DCMotor/DCMotor.cs
krwq: it's an implementation detail so yes

krwq:
testing in progress (not expecting any issues though)

src/devices/DCMotor/DCMotor.cs
```diff
@@ -32,67 +34,185 @@ protected DCMotor(GpioController controller)
...
+                directionPin,
+                controller);
+        }
+
+        /// <summary>
+        /// Creates <see cref="DCMotor"/> instance which allows to control speed in both directions.
+        /// </summary>
+        /// <param name="speedControlChannel"><see cref="PwmChannel"/> used to control the speed of the motor</param>
+        /// <param name="directionPin">Pin used to control the direction of the motor</param>
+        /// <param name="otherDirectionPin">Pin used to control the direction of the motor</param>
```
joperezr: Honestly I still think it is not super clear to me why to pick the dual direction pin vs the single pin direction one until I read in the remarks section. Perhaps we should add some of that info into the Summary? Anyway, this may also be because I haven't connected one of this before so it might just be that I'm not used to using these devices.

krwq: We can improve that once we get someone using it and complaining about summary/remarks - it's hard to pick easier words after you already know how something works

PR #647, opened 2019-07-31
Fix compensate temperature formula of Bmx device family

RobinTTY:


krwq:
@ZhangGaoxing any chance you could validate if this is fixing the issue you mentioned?

RobinTTY:
I agree that a default configuration would be good, even more so for the Bme680 which has more configuration options. Setting those ever time is cumbersome. I'll integrate that in #482 which I will get around to at the end of this week. 

It's also true that the accuracy of the measurements is no where near high enough (+/-0.5°C per datasheet) for a full double to be meaningful. So the result could either be truncated or it's in the hands of the consumer how to display the results.

src/devices/Bmxx80/Bme680.cs
```diff
@@ -222,5 +222,19 @@ private double CompensatePressure(int adcPressure)
...
             return calculatedPressure;
         }
+
+        /// <inheritdoc cref="Bmxx80Base"/>
+        protected override Temperature CompensateTemperature(int adcTemperature)
+        {
+            // Calculation for the Bme680 differs slightly from other sensors of this family!
+            var var1 = (adcTemperature / 16384.0 - _calibrationData.DigT1 / 1024.0) * _calibrationData.DigT2;
+            var var2 = adcTemperature / 131072.0 - _calibrationData.DigT1 / 8192.0;
+            var2 *= var2 * _calibrationData.DigT3 * 16.0;
```
joperezr: Seems like the only difference between Bme680 and the base implementation is the factor for multiplying this. If so, can we instead change it so that we have a protected virtual method that returns this factor method so that we only override that part? That way we don't have much duplicated code.

RobinTTY: Sure, I could just add a field to the base class (e.g. _tempCalibrationFactor) that bme680 can set to 16 and is 1 for the other ones. Would that be fine? Or should it really be a method that only returns this factor?

joperezr: a virtual getter method would take advantage of the polymorphic features of c#. I was thinking something super simple like this:

```c#
// on the base class, you have a virtual method like the following:
protected virtual double _factor => 1.0;

protected Temperature CompensateTemperature(int adcTemperature)
{
//...
    var2 *= var2 * _calibrationData.DigT3 * _factor;
//...
}



// And in your Bme680 subclass you can just do:
protected override double _factor => 16.0;
```

joperezr: also, now that I wrote it I just realized that you are doing `var2 *= var2 * ....`  Is that intended? do you actually want var2 squared? the previwous formula would expand to: `var2 = var2 * var2 * _calibrationData.DigT3 * 16.0;`

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -121,14 +121,16 @@ protected virtual Temperature CompensateTemperature(int adcTemperature)
...
-            var2 *= var2 * _calibrationData.DigT3;
+            var2 *= var2 * _calibrationData.DigT3 * GetTempCalibrationFactor();
 
             TemperatureFine = (int)(var1 + var2);
 
             double temp = (var1 + var2) / 5120.0;
             return Temperature.FromCelsius(temp);
         }
 
+        protected virtual int GetTempCalibrationFactor() => 1;
```
joperezr: NIT: doesn't have to be a method if you don't want to, you can just do a property getter so that you use it as if it were a field like:
```c#
protected override int _factor => 1;
```

It's up to you, whatever you think is cleanest is fine. I just mention this because I saw that you commented on if this could just be a field, so using virtual properties would be the closest thing.

src/devices/Bmxx80/Bme680.cs
RobinTTY: > also, now that I wrote it I just realized that you are doing `var2 *= var2 * ....` Is that intended? do you actually want var2 squared? the previwous formula would expand to: `var2 = var2 * var2 * _calibrationData.DigT3 * 16.0;`

Yeah that's correct, var2 has to be squared.

src/devices/Bmxx80/Bmxx80Base.cs
RobinTTY: Yeah I was thinking of a virtual property I'll change that.

RobinTTY:
Addressed the feedback, also tested it on the Bme680 which gave me correct measurements when comparing to another library.

PR #656, opened 2019-08-05
Clone I2cConnectionSettings and SpiConnectionSettings in order to make them immutable

### LINKED ISSUE #603: SpiConnectionSettings should be immutable or cloned by SpiDevice
Currently properties in SpiConnectionSettings have setters which makes them mutable. Also SpiDevice takes settings and simply assigns them. This has side effect that changing settings after SpiDevice is constructed may have effect on SPI transmission

joperezr:
fixes #603 

cc: @shaggygi @Frankenslag @krwq 

These changes will fix the current issue where you can modify the connection settings of a device after creation. It would still allow people to change settings object themselves in order to allow dynamic instantiation (see comment https://github.com/dotnet/iot/issues/603#issuecomment-513172714) and when the settings are requested form the device, it will just create a clone and pass it to the caller.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Spi/Devices/UnixSpiDevice.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Spi/Devices/UnixSpiDevice.Linux.cs
@@ -37,9 +37,10 @@ namespace System.Device.Spi
         public string DevicePath { get; set; }
 
         /// <summary>
-        /// The connection settings of a device on a SPI bus.
+        /// The connection settings of a device on a SPI bus. The connection settings are immutable after the device is created
+        /// so the object returned will be a clone of the settings object.
         /// </summary>
-        public override SpiConnectionSettings ConnectionSettings => _settings;
+        public override SpiConnectionSettings ConnectionSettings => new SpiConnectionSettings(_settings);
 
         private unsafe void Initialize()
         {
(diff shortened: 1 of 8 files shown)


src/System.Device.Gpio/System/Device/Spi/Devices/UnixSpiDevice.Linux.cs
```diff
@@ -37,9 +37,10 @@ public UnixSpiDevice(SpiConnectionSettings settings)
         public string DevicePath { get; set; }
 
         /// <summary>
-        /// The connection settings of a device on a SPI bus.
+        /// The connection settings of a device on a SPI bus. The connection settings are immutable after the device is created
```
krwq: this doesn't need summary since it's an internal class (also you can use inheritdoc)

krwq: same for Windows10SpiDevice

joperezr: you are right, we should probably apply the inheritdoc over all of the driver classes in iot repo. I will do that in a subsequent PR so that this one doesn't look polluted by all those changes.

Issue #226, opened 2019-02-13
Update Pca95x4 with IGpioController


shaggygi:
@joperezr Go ahead and assign to me.  This will be a good exercise for better understanding on the good work @JeremyKuhne has done.  Most likely, we will have a circle-back to update the I2cController stuff, so this will focus only on the IGpioController for now.

JeremyKuhne:
Note that I've got a new library with Pcx857x support in it that I hope to get a PR open for today.

shaggygi:
Another side note... we might want to review #135 at some point since expanders support other pin modes in addition to what the current core API includes.  For example, a couple of the MCPs have open-drain outputs.  This might be a good exercise to do sooner than later as I noticed there are some places in the core GPIO that would have to be updated, as well.

JeremyKuhne:
https://github.com/dotnet/iot/pull/229 is the initial Pcx857x PR.

Ellerbach:
Rather than IGpioController which won't really happen, this should then implement the board abstraction and propoer a real GpioController.

krwq:
[Triage] We only should implement GpioDriver

Issue #225, opened 2019-02-13
Add CY8C95xx I/O Expander with EEPROM Binding


shaggygi:
Add new binding to list.  https://www.cypress.com/file/37971/download

@joperezr Go ahead and assign to me.

Issue #90, opened 2018-12-10
Add System.Device.Adc.{AdcChannel, AdcChannelMode, AdcController} and port Mcp3008 device binding to new surface area


joshfree:
We need to standardize ADC device binding APIs, using Windows.Devices.Adc as a starting point, and update the 10-bit ADC Mcp3008 device binding accordingly.

https://docs.microsoft.com/en-us/uwp/api/windows.devices.adc.adccontroller

https://github.com/dotnet/iot/tree/master/src/devices/Mcp3008

joshfree:
@joperezr would you be interested in picking this one up next?

joperezr:
I was planning on working on testing infrastructure next, but I can also do this one instead.

shaggygi:
Referencing https://www.linux.com/learn/how-get-analog-input-beaglebone-black

shaggygi:
This might help...  [[Proposal] Provide interfaces for different controllers](https://github.com/dotnet/iot/issues/125)

joperezr:
The thing about this is that technically ADC is not really a communication protocol, so I'm not so sure if the right place for this would be the main library with a namespace like System.Device.Adc. IMO, this is something that should probably live on the device bidnings package instead, perhaps some sort of interface that all ADCs need to implement.

krwq:
@joperezr ADC is a device so System.Device sounds like a good place :smile:

joperezr:
For the namespace we’ve been using System.Device.<CommunicationProtocol> Where Communication protocol is a device protocol. We currently have Gpio, Pwm, I2c, and Spi. That means that this would be the first namespace where the third part of it is not a protocol. For device bindings. We have been using Iot.Device instead, so this would probably be something like Iot.Device.Adc

shaggygi:
@joperezr so are you suggesting not having System.Device.Adc in core API similar to System.Device.Gpio/I2c/Spi/Pwm?

If so, I disagree as there are dev boards that have true ADC pins (like BeagleBoard Black).  We should have similar classes/interfaces for controllers/drivers as we're doing with Gpio/Pwm.  Therefore, dev boards and bindings could implement them.  For example, the Mcp3008 could also implement IAdcController like Mcp23xxx implements IGpioController.

If I misunderstood you.... then disregard 😕

krwq:
[...]

PR #663, opened 2019-08-08
Simplify MCP3008

krwq:
Mainly remove InputConfiguration but also:
- remove unnecessary operations
- no allocations
- error detection in reading code
- add comments to explain bit shifting

Fixes TODO on one of our 3.0 items: https://github.com/dotnet/iot/projects/4#card-24042553

### DIFF
--- a/src/devices/Mcp3008/Mcp3008.cs
+++ b/src/devices/Mcp3008/Mcp3008.cs
@@ -12,7 +12,7 @@ namespace Iot.Device.Mcp3008
     {
         private SpiDevice _spiDevice;
 
-        public enum InputConfiguration
+        internal enum InputConfiguration
         {
             Differential = 0,
             SingleEnded = 1
@@ -29,14 +29,14 @@ namespace Iot.Device.Mcp3008
             _spiDevice = null;
         }
 
-        public int Read(int channel, InputConfiguration inputConfiguration = InputConfiguration.SingleEnded)
+        public int Read(int channel)
         {
             if (channel < 0 || channel > 7)
             {
                 throw new ArgumentException("ADC channel must be within 0-7 range.");
             }
 
-            return ReadSpi(channel, inputConfiguration);
+            return ReadSpi(channel, InputConfiguration.SingleEnded);
         }
 
         private static byte GetConfigurationBits(int channel, InputConfiguration inputConfiguration)
--- a/src/devices/Mcp3008/Mcp3008.cs
+++ b/src/devices/Mcp3008/Mcp3008.cs
@@ -12,12 +12,6 @@ namespace Iot.Device.Mcp3008
     {
         private SpiDevice _spiDevice;
 
-        internal enum InputConfiguration
-        {
-            Differential = 0,
-            SingleEnded = 1
-        }
-
         public Mcp3008(SpiDevice spiDevice)
         {
             _spiDevice = spiDevice;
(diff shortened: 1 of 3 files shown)


krwq:
ping @joperezr

src/devices/Mcp3008/Mcp3008.cs
```diff
@@ -29,41 +23,36 @@ public void Dispose()
...
-        }
+            Span<byte> writeBuffer = stackalloc byte[3];
+            Span<byte> readBuffer = stackalloc byte[3];
 
-        private static byte GetConfigurationBits(int channel, InputConfiguration inputConfiguration)
-        {
-            int configurationBits = (0b0001_1000 | channel) << 3;
+            // first we send 5 bits, rest is ignored
+            // start bit = 1
+            // conversion bit = 1 (single ended)
```
joperezr: Why did we stop allowing the Differential configuration?

src/devices/Mcp3008/Mcp3008.cs
```diff
@@ -29,41 +23,36 @@ public void Dispose()
...
+        public int Read(int channel)
         {
             if (channel < 0 || channel > 7)
             {
                 throw new ArgumentException("ADC channel must be within 0-7 range.");
             }
 
-            return ReadSpi(channel, inputConfiguration);
-        }
+            Span<byte> writeBuffer = stackalloc byte[3];
```
joperezr: question: stackalloc would initialize all bytes to 0 right?

src/devices/Mcp3008/Mcp3008.cs
```diff
@@ -29,41 +23,36 @@ public void Dispose()
...
-
-            _spiDevice.TransferFullDuplex(writeBuffer, readBuffer);
-
-            int result = (readBuffer[0] & 0b0000_0001) << 9;
-            result |= (readBuffer[1] & 0b1111_1111) << 1;
-            result |= (readBuffer[2] & 0b1000_0000) >> 7;
-            result = result & 0b0011_1111_1111;
+            int result = (readBuffer[0] & 1) << 9;
+            result |= readBuffer[1] << 1;
+            result |= readBuffer[2] >> 7;
```
joperezr: This is different than what we were doing before. Before we were doing `(readBuffer[2] & 0b1000_0000) >> 7;` but you are flipping all of readBuffer[2] bits. Also, you seem to have missed the clearing of a couple of bits that we used to do here: `result = result & 0b0011_1111_1111;`

src/devices/Mcp3008/Mcp3008.cs
```diff
@@ -29,41 +23,36 @@ public void Dispose()
             _spiDevice = null;
         }
 
-        public int Read(int channel, InputConfiguration inputConfiguration = InputConfiguration.SingleEnded)
+        public int Read(int channel)
```
joperezr: Given that you are already here and that we want to support this binding officially, can you add xml docs to all of the three public methods and to the class?

src/devices/Mcp3008/Mcp3008.cs
krwq: it's not useful. Previously the support was very limited - it wasn't clear what it was subtracting exactly or how to use it.

The software implementation for the differential configuration between channel 0 and 1 is just:
`adc.Read(1) - adc.Read(0)` so there is no clear usage for that as subtraction is just easier to use.

I'd recommend not having it until there is an actual scenario where you can achieve something you cannot with subtracting (which I doubt there is such)

src/devices/Mcp3008/Mcp3008.cs
krwq: yes, it must, otherwise you could leak some sensitive data by accident and that would be security issue

src/devices/Mcp3008/Mcp3008.cs
krwq: yes, I got read of clearing bits because they effectively didn't do anything.

I.e. masking byte with `0b1111_1111` is the same as not doing anything since it's already fulfilling the mask.
Shifting by 7 to the right - you will lose first 7 LSB bits anyway so there is no point in extra bitmasking

joperezr: that part makes sense, but you did remove the other part `result = result & 0b0011_1111_1111;` which was clearing a couple of bits in the result, that part I believe we should still keep.

krwq: @joperezr the way we currently shift they will always be zero anyway

joperezr: I see you are right.

src/devices/Mcp3008/Mcp3008.cs
shaggygi: > I'd recommend not having it until there is an actual scenario where you can achieve something you cannot with subtracting (which I doubt there is such)

I think this is why there is a need for better guidance/standards on expectations for binding development.  When I work on bindings (especially types with registers or memory locations that can be modified), I look at trying to develop the API based on the manufacture datasheet.  This includes values/properties (usually named based on how the datasheet references/labels them) to allow an easy way to 1) understand/search between the API and datasheet and 2) pass in parameters that will change bits/bytes to control the device down to the lowest level.  There's been previous discussions on not including enums or properties based on registers.  I disagree as it offers clarity and probably easier for the majority of others to get started faster.  But I also agree it can make the API larger and sometimes overwhelming to beginners that don't have experience with circuits and datasheets.  That's why they start with the Samples page.

Of course, the binding API should be convenient and help configure the device for most use-case scenario.  However, the API shouldn't limit functionality as others might need that granularity and now can't use the API for their needs.  In the case of this binding (Mcp3008), @richlander original added a very limited API.  This was understandable, as we were trying to get our minds wrapped around how to build upon the core API and show some samples.  I later updated with fixes including the `InputConfiguration` (as based on the datasheet) to help select type of channel.  I realize this binding was going to change as we would eventually create new [ADC APIs](https://github.com/dotnet/iot/issues/90) and implement like it was done with Mcp23xxx and `GpioController`.

There are cases where one might need a couple single and a couple diff channels.  There are cases where hardware designers use them to measure signals from on-board components and connectors (from external wire harnesses).  Functionality is reduced by removing this parameter.  Yes, the substraction scheme may work in case (I've not taken time to review in detail).  You also lose time as you would have to perform 2 channel reads and calculate where diff channel reads take 1.

Why not just change the parameter to optional so user has the option as it might be a use-case for thier needs?  Afterall, the device offers it and the datasheet helps explain.

BTW, I'm not trying to 💩  on the improvements made in this PR.  Just a little frustrated we may continue to go in circles trying to update what surface area is expected for bindings.  The guidance doc would be a big help here.  That's one reason I've held off on working on current assigned bindings.

For example...
- #225: PwmController was removed and not a good way to manage related channels now.
- #226: Looks like there's talks (from [Project Board](https://github.com/dotnet/iot/projects/4)) on removing IGpioController.

krwq: Current state on [guidelines for device bindings](https://github.com/dotnet/iot/blob/master/Documentation/Devices-conventions.md) is following:

```
Only the most useful APIs should be public, anything else which may be useful but will unlikely get used by most of the people should be protected (inheriting the class allows you to use it but it is not visible by default) or internal/private
```

krwq: I've opened https://github.com/dotnet/iot/pull/673 please let me know what do you think is missing there or we should clarify

PR #664, opened 2019-08-08
Using V4L2 to realize VideoDevice

ZhangGaoxing:
I tested it on Raspberry Pi 3B+ and Orange PI Zero using CSI and USB cameras. This draft version can capture the static image correctly. It can't shoot videos yet because I don't know much about how video streams are handled. But the method to get the raw frame data is done. V4L2 struct and consts are not complete, they are too many, and I just transplanted what the methods needed. Now I have some questions, such as whether it should be placed in `System.Device.xxx` or `Iot.Device.xxx` namespace, and whether there are some extensions to help transform the image format(my USB camera only supports YUV format😄). I hope to receive your reply and help😄. @joperezr @krwq 

A simple example
```C#
{
    VideoConnectionSettings settings = new VideoConnectionSettings(0)
    {
        CaptureSize = (2560, 1920),
        PixelFormat = PixelFormat.JPEG,
        ExposureType = ExposureType.Auto,
    };
    using VideoDevice device = VideoDevice.Create(settings);

    // Get the supported formats of the device
    foreach (var item in device.GetSupportedPixelFormats())
    {
        Console.Write($"{item} ");
    }
    Console.WriteLine();

    // Get the resolutions of the format
    foreach (var item in device.GetPixelFormatResolutions(PixelFormat.JPEG))
    {
        Console.Write($"{item.Width}x{item.Height} ");
    }
    Console.WriteLine();

    // Query v4l2 controls default and current value
    var value = device.GetVideoDeviceValue(VideoDeviceValueType.ExposureTime);
    Console.WriteLine($"{value.Name} Min: {value.Minimum} Max: {value.Maximum} Step: {value.Step} Default: {value.DefaultValue} Current: {value.CurrentValue}");

    // Capture static image
    device.Capture("/home/pi/test1.jpg");

    // Change capture setting
    device.Settings.HorizontalFlip = true;

    // Capture static image
    device.Capture("/home/pi/test2.jpg");
}
```


src/System.Device.Gpio/System/Device/Media/Devices/UnixVideoDevice.Linux.cs
```diff
@@ -0,0 +1,420 @@
...
+            return dataBuffer;
+        }
+
+        private unsafe void SetVideoConnectionSettings()
+        {
+            FillVideoConnectionSettings();
+
+            // Set capture format
+            v4l2_format format;
+            if (Settings.CaptureSize == (0, 0))
```
ZhangGaoxing: My Raspberry Pi CSI camera(unofficial) couldn't get maximum resolution, but the USB camera could get it. I guess it's a driver problem.

src/System.Device.Gpio/System/Device/Media/Devices/Windows10VideoDevice.Windows.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace System.Device.Media
+{
+    internal class Windows10VideoDevice : VideoDevice
```
ZhangGaoxing: Not implemented on Windows🤔

src/System.Device.Gpio/System/Device/Media/VideoConnectionSettings.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// The pixel format of video device captured image.
+        /// </summary>
+        public PixelFormat PixelFormat { get; set; } = PixelFormat.YUYV;
+
+        /// <summary>
+        /// The exposure type of video device.
+        /// </summary>
+        public ExposureType ExposureType { get; set; }
```
ZhangGaoxing: These settings I refer to the settings provided by raspistill. These are commonly used settings. V4L2 can set too many settings😅.

krwq:
That explains why you were not so active recently :smile: Nice work @shaggygi!

src/System.Device.Gpio/Interop/Unix/Libc/Interop.ioctl.cs
```diff
@@ -10,6 +10,10 @@ internal partial class Interop
     [DllImport(LibcLibrary, SetLastError = true)]
     internal static extern int ioctl(int fd, uint request, IntPtr argp);
 
+
+    [DllImport(LibcLibrary, SetLastError = true)]
+    internal static extern int ioctl(int fd, int request, IntPtr argp);
```
krwq: can you remove the previous overload since per `ioctl(2)` it's unsigned long anyway so I presume it will be casted

src/System.Device.Gpio/Interop/Unix/Libc/Interpo.ioctl.h.cs
```diff
@@ -0,0 +1,37 @@
...
+    const int _IOC_NRSHIFT = 0;
+    const int _IOC_TYPESHIFT = _IOC_NRSHIFT + _IOC_NRBITS;
+    const int _IOC_SIZESHIFT = _IOC_TYPESHIFT + _IOC_TYPEBITS;
+    const int _IOC_DIRSHIFT = _IOC_SIZESHIFT + _IOC_SIZEBITS;
+
+    const int _IOC_NONE = 0;
+    const int _IOC_WRITE = 1;
+    const int _IOC_READ = 2;
+
+    internal static int _IOC(int dir, int type, int nr, int size)
```
krwq: is this a port of some C/C++ macro?

src/System.Device.Gpio/System/Device/Media/Devices/UnixVideoDevice.Linux.cs
krwq: Did you try changing other parameters while increasing the size? (i.e. slow down capture rate or something similar - might be worth to check with strace what other software is doing)

src/System.Device.Gpio/System/Device/Media/PixelFormat.cs
```diff
@@ -0,0 +1,463 @@
...
+            Type type = typeof(RawPixelFormat);
+
+            var fields = type.GetFields(BindingFlags.Static | BindingFlags.Public);
+            foreach (var field in fields)
+            {
+                Console.WriteLine($"{field.Name} = {field.GetValue(new RawPixelFormat())},");
+            }
+        }
+
+        public static uint v4l2_fourcc(uint a, uint b, uint c, uint d) => a | (b << 8) | (c << 16) | (d << 24);
```
krwq: if you're passing this to native you might want to consider fixed size buffer of bytes for this

krwq:
Please add some sample code, haven't seen anything blocking!

src/System.Device.Gpio/Interop/Unix/Libc/Interpo.ioctl.h.cs
ZhangGaoxing: Yes, these are from `ioctl.h`

src/System.Device.Gpio/Interop/Unix/Libc/Interop.ioctl.cs
ZhangGaoxing: Request contains negative numbers

krwq: you can use `(uint)-1` - it should have the same effect

ZhangGaoxing:
In the new commit, I added some new v4l2 settings and converters for common pixel formats. For video shooting, there are still many problems to consider. Because the functions is not only for Raspberry Pi CSI cameras, but also for all USB cameras that meet the Linux standard. For example, Raspberry Pi CSI camera can directly output H264 format video. We don't need to do much work, just read and save data from the driver. Some video players can play H264 format video directly. However, many webcams only support YUV format, which requires some transformations to save videos. And the file format saved is avi or MP4 or flv... Implementing this requires a lot of work, and I haven't found the right NuGet package, no one has used C# to implement these functions, and. NET itself lacks some methods of media processing. I also have my own job, and I'm unlikely to concentrate on it... I have the following suggestions:
1. Not adding video method for the time being.
2. I looked at some similar libraries, some of which use the command to perform the function of shooting videos. In C#, we can use `Process` class to run `ffmpeg` to realize video shooting.

krwq:
@ZhangGaoxing thanks, just pictures are fine for now. For H264 there is very likely some C library you can wrap to do it instead of implementing it from the scratch

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -0,0 +1,546 @@
...
+
+        /// <summary>
+        /// Capture a picture from the video device.
+        /// </summary>
+        /// <param name="path">Picture save path</param>
+        public override async Task CaptureAsync(string path)
+        {
+            using (CancellationTokenSource cts = new CancellationTokenSource())
+            {
+                await CaptureAsync(path, cts.Token);
```
krwq: `default(CancellationToken)` or `CancellationToken.None`

krwq: or even better take cancellationToken as an input arg with default value

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -0,0 +1,546 @@
...
+        }
+
+        /// <summary>
+        /// Capture a picture from the video device.
+        /// </summary>
+        /// <param name="path">Picture save path</param>
+        /// <param name="token">A cancellation token that can be used to cancel the work</param>
+        public override async Task CaptureAsync(string path, CancellationToken token)
+        {
+            await Task.Run(() =>
```
krwq: the token is mostly useless here - it won't automagically work so currently it will only have effect if you cancel before Task.Run gets called

krwq: (suggestion is to use synchronous method here since async doesn't buy you anything)

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -0,0 +1,546 @@
...
+                DefaultValue = query.default_value,
+                CurrentValue = ctrl.value
+            };
+        }
+
+        /// <summary>
+        /// Get all the pixel formats supported by the device.
+        /// </summary>
+        /// <returns>Supported pixel formats</returns>
+        public override List<PixelFormat> GetSupportedPixelFormats()
```
krwq: this should return `IEnumerable<PixelFormat>`. List is fine as an implementation detail though

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -0,0 +1,546 @@
...
+
+            return result;
+        }
+
+        /// <summary>
+        /// Get all the resolutions supported by the specified pixel format.
+        /// </summary>
+        /// <param name="format">Pixel format</param>
+        /// <returns>Supported resolution</returns>
+        public override List<(uint Width, uint Height)> GetPixelFormatResolutions(PixelFormat format)
```
krwq: Same: return `IEnumerable<...>`

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -0,0 +1,546 @@
...
+            };
+            var res = V4l2Struct(VideoSettings.VIDIOC_S_FMT, ref format);
+
+            // Set exposure type
+            v4l2_control ctrl = new v4l2_control
+            {
+                id = VideoDeviceValueType.ExposureType,
+                value = (int)Settings.ExposureType
+            };
+            V4l2Struct(VideoSettings.VIDIOC_S_CTRL, ref ctrl);
```
krwq: I feel like V4l2Struct should have an overload which take id and value so that you don't have to touch the struct directly

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -0,0 +1,546 @@
...
+
+            // Set vertical flip
+            ctrl.id = VideoDeviceValueType.VerticalFlip;
+            ctrl.value = Settings.VerticalFlip ? 1 : 0;
+            V4l2Struct(VideoSettings.VIDIOC_S_CTRL, ref ctrl);
+        }
+
+        private void FillVideoConnectionSettings()
+        {
+            if (Settings.CaptureSize.Equals(default))
```
krwq: Should this return settings instead?

src/devices/Media/VideoDevice/PixelFormat.cs
```diff
@@ -0,0 +1,464 @@
...
+
+using System;
+using System.Reflection;
+
+namespace Iot.Device.Media
+{
+    /// <summary>
+    /// The pixel format of a video device.
+    /// </summary>
+    public enum PixelFormat : uint
```
krwq: would be useful to add comment where those values come from

krwq: nvm, I see it's there below

src/devices/Media/VideoDevice/samples/V4l2.Samples.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
```
krwq: any specific reason for 3.0? other projects are fine with 2.1 (it's ok to be this way, just curious)

krwq:
LGTM as first pass on video devices.

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
ZhangGaoxing: This method can operate on all v4l2 structs. When setting these settings, they just used a struct together. I think that's fine now.🤔

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
ZhangGaoxing: The logic now is to modify the property. If the user does not set this property, the method reads the default setting from the v4l2 driver.

src/devices/Media/VideoDevice/samples/V4l2.Samples.csproj
ZhangGaoxing: My fault, I forgot to modify it.😂

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
ZhangGaoxing: Actually, what I want is something like this😂. Also it can be cancelled from other threads
```C#
CancellationTokenSource cts = new CancellationTokenSource(1000);
await device.CaptureAsync($"{path}/jpg_direct_output.jpg", cts.Token);
```

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -67,10 +66,7 @@ public UnixVideoDevice(VideoConnectionSettings settings)
         /// <param name="path">Picture save path</param>
         public override async Task CaptureAsync(string path)
         {
-            using (CancellationTokenSource cts = new CancellationTokenSource())
-            {
-                await CaptureAsync(path, cts.Token);
-            }
+            await CaptureAsync(path, CancellationToken.None);
```
krwq: this could be just `return CaptureAsync(path, CancellationToken.None);`

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -87,24 +83,20 @@ public override async Task CaptureAsync(string path, CancellationToken token)
...
         /// </summary>
         /// <returns>Picture stream</returns>
+        /// <param name="token">A cancellation token that can be used to cancel the work</param>
         public override async Task<MemoryStream> CaptureAsync()
         {
-            using (CancellationTokenSource cts = new CancellationTokenSource())
-            {
-                return  await CaptureAsync(cts.Token);
-            }
+            return await CaptureAsync(CancellationToken.None);
```
krwq: ditto

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
krwq: (sorry didn't caught it on first pass)

src/devices/Media/VideoDevice/samples/Program.cs
```diff
@@ -40,15 +40,19 @@ static async Task Main(string[] args)
             var value = device.GetVideoDeviceValue(VideoDeviceValueType.Rotate);
             Console.WriteLine($"{value.Name} Min: {value.Minimum} Max: {value.Maximum} Step: {value.Step} Default: {value.DefaultValue} Current: {value.CurrentValue}");
 
-            await device.CaptureAsync("/home/pi/jpg_direct_output.jpg");
+            string path = "/home/pi/images";
```
krwq: why not in current directory? it's usually more intuitive

src/devices/Media/VideoDevice/README.md
```diff
@@ -0,0 +1,45 @@
+# VideoDevice
+
+## Getting Started
+Before that, please make sure you install `v4l-utils`.
```
krwq: remove non-actionable parts like `Before that, please make sure you` so basically just:

- Install `v4l-utils`

```
sudo apt-get update
sudo apt-get install v4l-utils
```

src/devices/Media/VideoDevice/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+sudo apt-get update
+sudo apt-get install v4l-utils
+```
+
+1. First you need to create a `VideoConnectionSettings` and set the parameters for capture.
```
krwq: Remove `First you need to`

src/devices/Media/VideoDevice/README.md
```diff
@@ -0,0 +1,45 @@
...
+    ```C#
+    VideoConnectionSettings settings = new VideoConnectionSettings(busId: 0)
+    {
+        CaptureSize = (2560, 1920),
+        PixelFormat = PixelFormat.JPEG,
+        ExposureType = ExposureType.Auto
+    };
+    ```
+2. Creates a communications channel to a video device.
```
krwq: nit: Creates => Create

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
ZhangGaoxing: > CS1997 C# Since is an async method that returns 'Task', a return keyword must not be followed by an object expression.

src/devices/Media/VideoDevice/samples/Program.cs
ZhangGaoxing: I will submit dockerfile later today

krwq: I still think current directory would be a better choice, you can remap it to /home/pi/images when running docker container

ZhangGaoxing: `Directory.GetCurrentDirectory()` does not get the path of the current program. If I want to get the right path, I need to run the `cd` command first, like `cd VideoDevice/sample` `sudo ./VideoDevice.Samples`. When running the docker image, docker does not execute the `cd` command to switch to the program directory. So `Directory.GetCurrentDirectory()` will return "/", and I can't mount the docker's root directory.
If I use the `docker run -w`  parameter to specify the docker directory, the .NET SDK will not be found.

krwq:
@ZhangGaoxing [my PR](https://github.com/dotnet/iot/pull/685) could have caused conflicts here. Please rebase your PR and make sure to fix all of the docs issues (previously they might not have been reported and PR might have stayed green)

ZhangGaoxing:
@krwq

src/devices/Interop/Unix/Libc/Interop.close.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+
+internal partial class Interop
```
krwq: Please move all of those to Interop.Libc.cs since this seems a bit too granular. Also for the common stuff please also move out stuff from https://github.com/dotnet/iot/blob/master/src/devices/SocketCan/Interop.cs with whatever is common (I can see: ioctl and close at minimum but you might want to review that)

src/devices/Interop/Unix/Libc/Interop.open.cs
```diff
@@ -0,0 +1,19 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+
+internal partial class Interop
+{
+    [DllImport(LibcLibrary, SetLastError = true)]
+    internal static extern int open([MarshalAs(UnmanagedType.LPStr)] string pathname, FileOpenFlags flags);
```
krwq: Why can't you use FileStream and just get a [handle](https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream.handle?view=netframework-4.8) out of it?

src/devices/Interop/Unix/Libc/Interpo.ioctl.h.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
krwq: nit: typo in file name

krwq: Optional: I'd prefer this file was called `Interop.ioctl.macros.cs` since I think that's what those are

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -0,0 +1,509 @@
...
+                Settings.HorizontalFlip = Convert.ToBoolean(GetVideoDeviceValue(VideoDeviceValueType.HorizontalFlip).DefaultValue);
+            }
+
+            if (Settings.VerticalFlip.Equals(default))
+            {
+                Settings.VerticalFlip = Convert.ToBoolean(GetVideoDeviceValue(VideoDeviceValueType.VerticalFlip).DefaultValue);
+            }
+        }
+
+        private void Initialize()
```
krwq: can you just always initialize in the constructor and remove the lock at all?

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -0,0 +1,509 @@
...
+                _deviceFileDescriptor = Interop.open(deviceFileName, FileOpenFlags.O_RDWR);
+
+                if (_deviceFileDescriptor < 0)
+                {
+                    throw new IOException($"Error {Marshal.GetLastWin32Error()}. Can not open video device file '{deviceFileName}'.");
+                }
+            }
+        }
+
+        private void Close()
```
krwq: should this be happening only in Dispose?

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -0,0 +1,509 @@
...
+        /// Get and set v4l2 struct.
+        /// </summary>
+        /// <typeparam name="T">V4L2 struct</typeparam>
+        /// <param name="request">V4L2 request value</param>
+        /// <param name="struct">The struct need to be read or set</param>
+        /// <returns>The ioctl result</returns>
+        private int V4l2Struct<T>(VideoSettings request, ref T @struct)
+            where T : struct
+        {
+            IntPtr ptr = Marshal.AllocHGlobal(Marshal.SizeOf(@struct));
```
krwq: Isn't this leaking memory?

src/devices/Media/VideoDevice/VideoDevice.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// Capture a picture from the video device.
+        /// </summary>
+        /// <param name="path">Picture save path.</param>
+        public abstract void Capture(string path);
+
+        /// <summary>
+        /// Capture a picture from the video device.
+        /// </summary>
+        /// <returns>Picture stream.</returns>
+        public abstract MemoryStream Capture();
```
krwq: This should return Stream on the API level. Underneath it can use whatever stream implementation

krwq:
@ZhangGaoxing only treat comment about the leak and returning Stream as blocking. Anything else I'd prefer to be fixed but I'm fine with this being left as is.

src/devices/Interop/Unix/Libc/Interop.open.cs
ZhangGaoxing: I referred to the implementation of `I2C`, and Interop was copied from System.Device.Gpio project.

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
ZhangGaoxing: If I change the shooting parameters after the first capture, then the second capture will throw an error. So the process must be initialization -> shooting -> close.

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
ZhangGaoxing: It happens at the end of each capture.

src/devices/Interop/Unix/Libc/Interop.open.cs
krwq: ok, we should clean duplicate code in the future but fine in this PR

Issue #419, opened 2019-05-13
Reconsider Dispose pattern across devices


krwq:
Multiple devices take SpiDevice/I2cDevice in the constructor and then dispose it later. We should reconsider if this is good approach and if disposing is always expected. Some alternatives:
- do not dispose the device and let the caller dispose it instead
- take bool flag (i.e. shouldDispose) and let the caller pick what they want

joperezr:
Instead of taking a flag, I would rather have a bool shouldDispose property that would be defaulted to true, but you could set otherwise if needed.

Frankenslag:
I like the 'shouldDispose' property... Certainly something is needed as multiple device bindings can use the same GPIO controller for example. I am not sure that it should be defaulted as true as I would generally prefer the 'caller dispose' model but I am undecided if it would be a good idea to change what seems to be happening already.

krwq:
I'm personally in favor of not disposing at all (move the responsibility to the caller) and where Dispose ends up being empty (presumably most of the devices) I'd just drop interface implementation to make it clearer that we do not Dispose anything

Gnbrkm41:
+1 to not disposing at all. Personally I think people will dispose of both the original disposable and the type holding the disposable anyway (that's what I do at least 😛 )
```cs
using (Stream stream = File.Open())
using (StreamReader reader = new StreamReader(stream))
{
    // Some nice codes come here
}
```
Like this. `using`s make them pretty easy to do if one has to dispose of both instances anyway.

> I would rather have a bool shouldDispose property that would be defaulted to true

But I think this is a nice idea as well. Giving them some choice is better than no choice, I guess.

joperezr:
> I think people will dispose of both the original disposable and the type holding the disposable anyway

If they do this, it should be fine with the way we are writing the bindings and our devices, because calling dispose twice should just be a no-op the second time. I'd still argue that disposing them being the default would be better, since I believe that most of the time the I2c or Spi device won't be used other than to initialize the device binding, so it should be ok to assume that we call dispose. I do agree however, that we should provide a way for people to change a setting in order to not dispose when the main device binding gets disposed, for those very few cases where people would want this.

krwq:
> I believe that most of the time the I2c or Spi device won't be used other than to initialize the device binding
@joperezr I'd disagree - for simple tests of devices that will be true but any larger/commercial app will likely use multiple SPI devices and will probably try to reuse the same instance.

[...]

PR #673, opened 2019-08-15
Re-iterate on devices conventions

krwq:
Fixes 3.0 part (docs update) of #419

cc: @joperezr @bartonjs 

@shaggygi please let me know what else should be clarified here (anything in this doc)

### DIFF
--- a/Documentation/Devices-conventions.md
+++ b/Documentation/Devices-conventions.md
@@ -2,9 +2,10 @@
 
 ## Value returned by sensor
 
-- Methods (i.e. `ReadTemperature`) should be used for anything which may change between the calls
+- Methods (i.e. `ReadTemperature`/`GetTemperature`) should be used for anything which may change between the calls
 - Properties (i.e. `SomeRegister`) should be used when value does not change between the calls (except when changed by setter)
 - Use `double` when you need to return any floating point value
+- Async methods can be added but the name of such method should have an `Async` suffix and there should also be synchronous equivalent method
 - Use `Vector2/3/4` for returning vectors`*`
 - Value should conform to units conventions (see below)
 - Only the most useful APIs should be public, anything else which may be useful but will unlikely get used by most of the people should be protected (inheriting the class allows you to use it but it is not visible by default) or internal/private
--- a/Documentation/Devices-conventions.md
+++ b/Documentation/Devices-conventions.md
@@ -1,5 +1,17 @@
 # Conventions for devices APIs
 
+This document is meant to provide guidelines. If you think the guidelines don't make sense in a specifc case (or at all) please file an issue or mention it in the PR description.
+
+## Expose everything device has vs. expose simple functionality
+
+All APIs should be as simple as they can be and simplicity is the most important when desiging APIs.
+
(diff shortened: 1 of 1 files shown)


shaggygi:
> Only the most useful APIs should be public, anything else which may be useful but will unlikely get used by most of the people should be protected (inheriting the class allows you to use it but it is not visible by default) or internal/private

Not part of this PR, but related topic.  This is a bit vague.  Not sure how we could keep consistent.  As stated previously, my opinion on binding APIs is they should be designed around the provided datasheet.  This allows for low-level interaction if users need.  You can then have the higher-level methods provided to perform the simpler calls (that most would use).  For example, `ReadDateTime` might be all that is needed for a RTC.  However, there might be some odd registers/bits that need to be changed to perform certain cases (daylight savings, alerts, exception flags, etc.).

It would be better to offer both scenarios within core binding class as I don't won't to be required to create a separate API that inherits just to offer a public enum/register property.  If I created a PR to add that additional API, there would be many back/forth discussions on why it is needed, etc.  For example, the MCP25xxx CAN device includes many registers to perform many functions related to the protocol.  With all the possible configurations, the API surface should allow to easily access the lower-level and then have higher-lever calls like `SendCanMessage`, etc.

Just my 2 pennies.

krwq:
@shaggygi I think this should be covered here and this kind of stuff is what I meant by `what else should be clarified here`. Let's gather a list of questions you might think of and I'll talk with folks to figure out what the official stance is on that.

cc: @ZhangGaoxing @Ellerbach

shaggygi:
@richlander was recently on an [episode of Azure DevOps Podcast](http://azuredevopspodcast.clear-measure.com/richard-lander-on-net-core-runtime-episode-50) and I believe this comment (around the 50 minute mark) fits my thinking.

When it comes to interfaces, you could have something like `ReadDateTime()` for a binding that implements an `IRealTimeClock`.  However, you could still expose the specifics of each device that implements that interface to allow users to get to specific features (registers, memory, etc.) of the device if they choose to.

krwq:
@shaggygi let's not start the interface conversation in this PR :smile:

My initial thoughts on `everything` vs `only easy to use stuff` is following:
- whatever is easy to use should be public
- whatever is all/full functionality should be protected or be in a separate class (if there is easy to use class it's fine to also have full implementation separately but easy to use stuff must be in all of the samples first)

I haven't talked with other people about this yet but that was our initial thoughts when I did have some conversations months back

joperezr:
I'm fine with both of the changes that you are making here since those we had already agreed upon. As to for the expose all APIs or just the most common stuff, I do agree with @shaggygi where I believe that exposing everything should be the goal in most cases, and only on some very special cases should we instead remove whatever is not necessary(These are just guidelines for most bindings, but we can have exceptions to the rules).

krwq:
@joperezr I've heard exactly opposite opinions (make it as simple to use as possible) in the past so we will need to discuss this - there is value in having both.

joperezr:
at least I believe we do all agree on the changes you are making on this PR, so let's go ahead and merge this and continue the discussion for other things in the already existing issue. That way we won't have stalled PRs for a long time till we come all to consensus.

krwq:
@joperezr one more look please at the latest commit

krwq:
@shaggygi please let me know if your concerns/questions are addressed here

Issue #677, opened 2019-08-20
Add GPS (NEO-M8 / NMEA 0183 protocol)


krwq:
I'm currently working on GPS module bindings and just wanted to make sure no one duplicates the work. 
I already got basic functionality working but there is a lot of clean up left and support for some functionality is missing.

Multiple GPS/navigation devices use NMEA0183 protocol to communicate with the sensor.
NEO-M8 supports multiple sentence types:
- [X] RMC - Required minimum navigation information (#679)
- [ ] GLL - Geographic Position – Latitude/Longitude
- [ ] VTG - Track Made Good and Ground Speed
- [ ] GGA - Global Positioning System Fix Data. Time, Position and fix related data for a GPS receiver
- [ ] GSA - GPS DOP and active satellites
- [ ] GSV - Satellites in view

NMEA supports lots more than that though but will leave it out of scope

ZhangGaoxing:
Maybe we can abstract a base class in this PR. I have a NEO-6M module. After this PR merge, I can help binding it.

krwq:
@ZhangGaoxing I suspect the current class might work as is - it doesn't do almost anything except setting line endings on serial port - we might rename it (and folder) if it will work as is. Rest is basically implementation of NMEA0183.

After that is merged I was planning to add more sentences to NMEA0183 (at least all the device is sending by default) - I'm happy if you're willing to test with your device and fix any issues you might find (or perhaps your device might generate more sentences which might be interesting).

stevehayles:
In the wild I see NMEA0183 being used as a data protocol at much faster rates than the standard of 4800 and also being used over UDP and TCP connections.  My own parser originall used a serial port 'readline' approach but I hit major performance problems running on a lower powered device (RPi3) on a 'busy' stream of data.

I have moved to a System.IO.Pipelines based approach working with the underlying serial port stream and then a simple state machine to parse out incoming data.  It avoids the allocations, copying and manipulation of the various buffers in the serial port read methods and the StringBuilder allocation in the Readline method. 

The 'state machine' parser is much cleaner in my view, is more resilient to corrupt data and well matched to the Pipe approach. I am now working on parsing the data fields as ReadOnlySpan<char> to further avoid allocations. I don't know whether any of this is suitable for this repo. 

[...]

PR #679, opened 2019-08-20
Add basic GPS support: NEO-M8, NMEA0183

krwq:
Contributes to: https://github.com/dotnet/iot/issues/677

Only `RMC` is added here. I'm planning to add more in subsequent PR to avoid unnecessary refactoring in case of PR feedback.

cc: @richlander

### DIFF
--- a/src/devices/Neo-M8/Neo-M8.csproj
+++ b/src/devices/Neo-M8/Neo-M8.csproj
@@ -1,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <!--Disabling default items so samples source won't get build by the main library-->
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <None Include="README.md" />
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
+      <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
+    </ProjectReference>
+    <ProjectReference Include="../Nmea0183/Nmea0183.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.IO.Ports" Version="4.6.0-preview8.19405.3" />
+  </ItemGroup>
+
+</Project>
--- a/src/devices/Neo-M8/NeoM8.cs
+++ b/src/devices/Neo-M8/NeoM8.cs
@@ -1,0 +1,48 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO.Ports;
+using Iot.Device.Nmea0183;
+using Iot.Device.Nmea0183.Sentences;
+
+namespace Iot.Device.Gps
+{
+    public sealed class NeoM8 : IDisposable
+    {
+        private SerialPort _sp;
+
+        /// <summary>
+        /// Constructs object representing Neo-M8
+        /// </summary>
+        /// <param name="serialPortName">Name of the serial port to use for communication</param>
+        public NeoM8(string serialPortName)
+        {
+            _sp = new SerialPort(serialPortName);
+            _sp.NewLine = "\r\n";
+            _sp.Open();
+
+            // Device streams continuously and therefore most of the time we would end up in the middle of the line
+            // therefore ignore first line so that we align correctly
+            _sp.ReadLine();
+        }
+
+        /// <summary>
+        /// Reads NMEA0183 sentence
+        /// </summary>
+        /// <returns>TalkerSentence instance</returns>
+        public TalkerSentence Read()
+        {
+            string sentence = _sp.ReadLine();
+            return TalkerSentence.FromSentenceString(sentence);
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            _sp?.Dispose();
+            _sp = null;
+        }
+    }
+}
--- a/src/devices/Neo-M8/README.md
+++ b/src/devices/Neo-M8/README.md
@@ -1,0 +1,30 @@
+﻿# NEO-M8
+
+## Summary
+
+u-blox M8 is Global Navigation Satellite System (GNSS).
+It is using NMEA0183 protocol for communication.
+
+## Binding Notes
+
+This device uses UART and therefore regular PC with RS232 to TTL converter can be used (i.e. Raspberry PI is not required).
+When using Raspberry PI use `raspi-config` to disable login shell and enable serial port (interfacing options).
+
+### Communication methods
+
+NEO-M8 supports multiple communication methods but only UART is currently supported:
+
+- [X] UART
+- [ ] USB
+- [ ] SPI
+- [ ] DDC (I2C compliant)
+
+### Supported sentences
+
+Current support for sentences is limited to only `RMC` (NMEA0183's Recommended Minimum Navigation Information). Please refer to [sentence directory](../Nmea0183/Sentences) for latest information what is currently supported.
+
+Full list of supported sentences varies by device.
+
+## References 
+
+- https://www.u-blox.com/sites/default/files/NEO-M8_DataSheet_%28UBX-13003366%29.pdf
--- a/src/devices/Neo-M8/samples/NeoM8.Sample.cs
+++ b/src/devices/Neo-M8/samples/NeoM8.Sample.cs
@@ -1,0 +1,49 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Iot.Device.Nmea0183;
+using Iot.Device.Nmea0183.Sentences;
+using Iot.Device.Gps;
+
+namespace Iot.Device.Gps.NeoM8Samples
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            using (NeoM8 neoM8 = new NeoM8("/dev/ttyS0"))
+            {
+                bool gotRmc = false;
+                while (!gotRmc)
+                {
+                    TalkerSentence sentence = neoM8.Read();
+
+                    object typed = sentence.TryGetTypedValue();
+                    if (typed == null)
+                    {
+                        Console.WriteLine($"Sentence identifier `{sentence.Id}` is not known.");
+                    }
+                    else if (typed is RecommendedMinimumNavigationInformation rmc)
+                    {
+                        gotRmc = true;
+
+                        if (rmc.LatitudeDegrees.HasValue && rmc.LongitudeDegrees.HasValue)
+                        {
+                            Console.WriteLine($"Your location: {rmc.LatitudeDegrees.Value:0.00000}, {rmc.LongitudeDegrees.Value:0.00000}");
+                        }
+                        else
+                        {
+                            Console.WriteLine($"You cannot be located.");
+                        }
+                    }
+                    else
+                    {
+                        Console.WriteLine($"Sentence of type `{typed.GetType().FullName}` not handled.");
+                    }
+                }
+            }
+        }
+    }
+}
--- a/src/devices/Neo-M8/samples/README.md
+++ b/src/devices/Neo-M8/samples/README.md
@@ -1,0 +1,61 @@
+﻿# NEO-M8 sample
+
+## Notes
+
+Serial port name will need to be adjusted for this to work correctly.
+You can find list of your ports using `System.IO.Ports.SerialPort.GetPortNames()` in `System.IO.Ports` package.
+Serial port in the sample below is a default UART on Raspberry PI.
+
+Sample might print several messages that identifier is not known, those can be safely ignored but you might be missing out on some of the information your device is providing.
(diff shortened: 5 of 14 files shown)


krwq:
@joperezr ping

src/devices/Neo-M8/Neo-M8.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <Compile Include="*.cs" />
+    <None Include="README.md" />
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
+      <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
+    </ProjectReference>
+    <ProjectReference Include="../Nmea0183/Nmea0183.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.IO.Ports" Version="4.6.0-preview8.19405.3" />
```
joperezr: Do you actually need the latest preview for this? Would it be possible to just use latest stable? I ask because we are not yet sure if we are going to ship at the same time of .NET Core or a little bit earlier, so it would be good if we don't depend on preview versions if possible.

src/devices/Neo-M8/NeoM8.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO.Ports;
+using Iot.Device.Nmea0183;
+using Iot.Device.Nmea0183.Sentences;
```
joperezr: NIT: not used namespace

src/devices/Neo-M8/Neo-M8.csproj
krwq: but this is OOB package, right? we need a version from at least around January this year so I think it will be preview either way

src/devices/Neo-M8/NeoM8.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="serialPortName">Name of the serial port to use for communication</param>
+        public NeoM8(string serialPortName)
+        {
+            _sp = new SerialPort(serialPortName);
+            _sp.NewLine = "\r\n";
+            _sp.Open();
+
+            // Device streams continuously and therefore most of the time we would end up in the middle of the line
+            // therefore ignore first line so that we align correctly
+            _sp.ReadLine();
```
joperezr: Should we have an optional bool in the constructor in order to not do a ReadLine here in case we are not in the middle of the line and we don't want to ignore? Or do you think that won't be useful?

krwq: Currently I don't think it will be useful since you'd have to connect some kind of relay to the NEO-M8 in order to shut it down and otherwise it won't stop talking. We can add it if we find it's useful one day. Also the device repeats measurements every second so there is no loss

src/devices/Neo-M8/samples/NeoM8.Sample.cs
```diff
@@ -0,0 +1,49 @@
...
+                            Console.WriteLine($"Your location: {rmc.LatitudeDegrees.Value:0.00000}, {rmc.LongitudeDegrees.Value:0.00000}");
+                        }
+                        else
+                        {
+                            Console.WriteLine($"You cannot be located.");
+                        }
+                    }
+                    else
+                    {
+                        Console.WriteLine($"Sentence of type `{typed.GetType().FullName}` not handled.");
```
joperezr: I suppose that these will eventually be added right? At least I believe these are the rest of the ones tracked by the issue.

krwq: yes, at least those few streamed by this device - all possible sentences are a lot of work (I think there might be more than a 100 types of sentences)

src/devices/Nmea0183/README.md
```diff
@@ -0,0 +1,34 @@
+﻿# NMEA 0183 Protocol
+
+## Summary
+
+NMEA stands for `National Marine Electronics Associations`.
```
joperezr: So this isn't really a binding then right? this is more of a standard for comunication. I'm fine with the way you have it, but what was the reasoning of having this separate to Neo-M8? Do we expect other bindings to use the same standard?

krwq: @joperezr yes, I know @ZhangGaoxing has one which will likely work with that but some parameters might vary (i.e. NEO-M8 also supports I2c but I wasn't able to get it to work (it just didn't respond nor sent anything, also you have to set line endings on serial port etc)

src/devices/Neo-M8/samples/NeoM8.Sample.cs
```diff
@@ -0,0 +1,49 @@
...
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            using (NeoM8 neoM8 = new NeoM8("/dev/ttyS0"))
+            {
+                bool gotRmc = false;
+                while (!gotRmc)
+                {
+                    TalkerSentence sentence = neoM8.Read();
```
joperezr: Can we also show an example of how you would use a QuerySentence?

src/devices/Nmea0183/QuerySentence.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// Reads NMEA0183 query sentence from provided string
+        /// </summary>
+        /// <param name="sentence">NMEA0183 query sentence</param>
+        /// <returns>QuerySentence instance</returns>
+        /// <remarks><paramref name="sentence"/> does not include new line characters</remarks>
+        public static QuerySentence FromSentenceString(string sentence)
+        {
+            // http://www.tronico.fi/OH6NT/docs/NMEA0183.pdf
+            // page 3
+            // $XXYYY ...
```
joperezr: Should we say here that X means requester and Y represents deviceId? Or perhaps in the doc of this method we should mention how to create query sentences.

src/devices/Neo-M8/samples/NeoM8.Sample.cs
krwq: it's part of the spec, initially device didn't want to send anything (due to my fault, except I2C which I'm not sure why) and then I realized that you don't have to send a query to get anything from the device. I actually haven't tried it but left it since it was in the spec anyway and I've already tried if it constructs correct sentences

src/devices/Nmea0183/SentenceId.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            return Id1 == other.Id1 && Id2 == other.Id2 && Id3 == other.Id3;
+        }
+
+        public static bool operator== (SentenceId obj1, SentenceId obj2)
+        {
+            return obj1.Equals(obj2);
+        }
+
+        public static bool operator!= (SentenceId obj1, SentenceId obj2) => !(obj1 == obj2);
```
joperezr: NIT: just call `!(obj1.Equals(obj2))` since it will be called by the equals operator anyway

src/devices/Nmea0183/SentenceId.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        public bool Equals(SentenceId other)
+        {
+            return Id1 == other.Id1 && Id2 == other.Id2 && Id3 == other.Id3;
+        }
+
+        public static bool operator== (SentenceId obj1, SentenceId obj2)
+        {
+            return obj1.Equals(obj2);
```
joperezr: NIT: you don't have to, but all these simple return methods could use the syntax like:
```c#
public static bool operator== (SentenceId obj1, SentenceId obj2) => obj1.Equals(obj2);
```
In order to be consistent with the not equals one.

src/devices/Nmea0183/TalkerSentence.cs
```diff
@@ -0,0 +1,217 @@
...
+using Iot.Device.Nmea0183.Sentences;
+
+namespace Iot.Device.Nmea0183
+{
+    /// <summary>
+    /// Represents NMEA0183 talker sentence
+    /// </summary>
+    public class TalkerSentence
+    {
+        private static ConcurrentDictionary<SentenceId, Func<TalkerSentence, object>> s_registeredSentences = GetKnownSentences();
```
joperezr: Why does this Dictionary need to be static?

joperezr: Oh I see, it's because of RegisterSentence method.

src/devices/Neo-M8/Neo-M8.csproj
joperezr: I see, bummer. That's fine then, we'll have to ship after .NET Core then.

src/devices/Neo-M8/samples/NeoM8.Sample.cs
joperezr: mm I see, so neoM8 can only read but not request any data?

src/devices/Nmea0183/TalkerSentence.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// NMEA0183 sentence identifier
+        /// </summary>
+        public SentenceId Id { get; private set; }
+
+        private string[] _fields;
+
+        /// <summary>
+        /// Fields of the NMEA0183 sentence
+        /// </summary>
+        public IEnumerable<string> Fields => _fields;
```
joperezr: NIT: Technically this could also just be a `{get; private set;}`

src/devices/Nmea0183/TalkerSentence.cs
```diff
@@ -0,0 +1,217 @@
...
+            int lastStarIdx = lastEntry.LastIndexOf('*');
+
+            if (lastStarIdx == -1 || lastStarIdx != lastEntry.Length - 3)
+            {
+                // there is no checksum, last entry is the last field
+                return (null, lastEntry);
+            }
+
+            int lastIdx = lastEntry.Length - 1;
+            int sum = HexDigitToDecimal(lastEntry[lastIdx - 1]) * 16 + HexDigitToDecimal(lastEntry[lastIdx]);
```
joperezr: NIT: It would be great if you add a small comment here to point to the page in the spec of how the checksum must be calculated in case it ever needs to be changed/fixed.

krwq:
needs rebase to make sure all of the errors [my PR](https://github.com/dotnet/iot/pull/685) could have triggered are fixed

src/devices/Neo-M8/README.md
```diff
@@ -0,0 +1,30 @@
+﻿# NEO-M8
+
+## Summary
+
+u-blox M8 is Global Navigation Satellite System (GNSS).
+It is using NMEA0183 protocol for communication.
+
+## Binding Notes
+
+This device uses UART and therefore regular PC with RS232 to TTL converter can be used (i.e. Raspberry PI is not required).
```
MarkCiliaVincenti:
```suggestion
This device uses UART and therefore regular PC with RS232 to TTL converter can be used (i.e. Raspberry Pi is not required).
```

src/devices/Neo-M8/README.md
```diff
@@ -0,0 +1,30 @@
...
+When using Raspberry PI use `raspi-config` to disable login shell and enable serial port (interfacing options).
```
MarkCiliaVincenti:
```suggestion
When using Raspberry Pi use `raspi-config` to disable login shell and enable serial port (interfacing options).
```

src/devices/Neo-M8/samples/README.md
```diff
@@ -0,0 +1,61 @@
+﻿# NEO-M8 sample
+
+## Notes
+
+Serial port name will need to be adjusted for this to work correctly.
+You can find list of your ports using `System.IO.Ports.SerialPort.GetPortNames()` in `System.IO.Ports` package.
+Serial port in the sample below is a default UART on Raspberry PI.
```
MarkCiliaVincenti:
```suggestion
Serial port in the sample below is a default UART on Raspberry Pi.
```

pgrawehr:
I started picking this up. But I'll wait until the PR backlog is a bit down before proposing a new PR.

joperezr:
@pgrawehr We've definitely been trying to burn some of the PR backlog we have, specially the PRs that are just waiting in some input and feedback, but please if you notice any PRs that are becoming stale, please do ping us on them so we can focus our attention to those. 😄

src/devices/Nmea0183/Sentences/RecommendedMinimumNavigationInformation.cs
```diff
@@ -0,0 +1,300 @@
...
+
+        private static double? Nmea0183ToDegrees(double? degreesMinutes, CardinalDirection? direction)
+        {
+            if (!degreesMinutes.HasValue || !direction.HasValue)
+                return null;
+
+            // ddddmm.mm
+            double degrees = Math.Floor(degreesMinutes.Value / 100);
+            double minutes = degreesMinutes.Value - (degrees * 100);
+            return ((double)degrees + (double)minutes / 60.0) * DirectionToSign(direction.Value);
```
mpalenov, line 254: Why is it required to cast degrees and minutes (which have `double` type) to `double`?

src/devices/Nmea0183/TalkerId.cs
```diff
@@ -0,0 +1,267 @@
...
+        /// <param name="id2">second character identifying the talker</param>
+        public TalkerId(char id1, char id2)
+        {
+            Id1 = id1;
+            Id2 = id2;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj is SentenceId other)
```
mpalenov, line 32: It seems it should be like that `if (obj is TalkerId other)`

joperezr:
[Triage] We are going to close this PR given that there are external libraries that already implement GPS protocol which have a wider support. We will follow up the conversation of which library to use on the issue that this PR references: https://github.com/dotnet/iot/issues/677

PR #683, opened 2019-08-22
Add FM transmitter KT0803

ZhangGaoxing:
This code was almost completed in March this year, and since I2C had no WriteRead() method at that time, it has been shelved until now. I've sorted it out today.


krwq:
@joperezr can we make it so that incremental build also fails with docs issues the same way as full build?

joperezr:
@krwq yeah we could do that, all we need is to set the property `<GenerateDocumentationFile>true</GenerateDocumentationFile>` on binding projects which will generate the warnings for incorrect or missing xml docs. That said, this will only throw a warning, the reason why full build fails is because we do TreatWarningsAsErrors = true.

krwq:
@joperezr is there any props or something which gets picked by all the projects? We should do exactly the same CI is doing on incremental builds and then have a way to override that

joperezr:
>is there any props or something which gets picked by all the projects?

yup. https://github.com/dotnet/iot/blob/master/src/devices/Directory.Build.props

krwq:
@joperezr https://github.com/dotnet/iot/pull/685

src/devices/Kt0803/Country.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Kt0803
+{
+    /// <summary>
+    /// Country
+    /// </summary>
+    public enum Country
```
krwq: Was this meant to be continent/region?

krwq: also there should be option like Other for people from other Asian countries and Africa

src/devices/Kt0803/Kt0803.cs
```diff
@@ -0,0 +1,378 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.I2c;
+
+namespace Iot.Device.Kt0803
```
krwq: Please make the namespace categorize this device (i.e. Iot.Device.Radio or something)

src/devices/Kt0803/Kt0803.cs
```diff
@@ -0,0 +1,378 @@
...
+        public void Dispose()
+        {
+            _i2cDevice?.Dispose();
+            _i2cDevice = null;
+        }
+
+        /// <summary>
+        /// Set Kt0803 FM Channel
+        /// </summary>
+        /// <param name="mhz">MHz ( Range from 70Mhz to 108Mhz )</param>
```
krwq: nit: remove spaces after `(` and before `)`

src/devices/Kt0803/Register.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// <summary>
+    /// Kt0803 Register
+    /// </summary>
+    internal enum Register : byte
+    {
+        KT_CHSEL = 0x00,
+        KT_CONFIG1 = 0x01,
+        KT_CONFIG2 = 0x02,
+        KT_CONFIG3 = 0x04,
+        KT_CONFIG4 = 0x0B,
```
krwq: 4,6,10? :smile:

src/devices/Kt0803/Country.cs
ZhangGaoxing: In fact, this setting is the `Pre-Emphasis Time Constant`. This parameter varies from country to country. Enum values come from the datasheet. I'll rename it to Region and add Other option.

src/devices/Kt0803/Register.cs
ZhangGaoxing: These registers do not have a fixed name. These numbers are the serial numbers of the chapters in the datasheet😂.

krwq:
@ZhangGaoxing [my PR](https://github.com/dotnet/iot/pull/685) could have caused conflicts here. Please rebase your PR and make sure to fix all of the docs issues (previously they might not have been reported and PR might have stayed green)

krwq:
@joperezr no need to rebase? just the re-running is fine? (expect cases with actual merge conflicts)

joperezr:
correct, just re-running will internally do a merge with master before actually running the build, so it will pick up your changes

src/devices/RadioTransmitter/Devices/Kt0803/BassBoost.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// Bass Boost
+    /// </summary>
+    public enum BassBoost : byte
+    {
+        /// <summary>
+        /// Disable
+        /// </summary>
+        Boost_Disable = 0,
```
krwq: please remove underscores since lack of them doesn't impede legibility of the code

src/devices/RadioTransmitter/Devices/Kt0803/PgaGain.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    /// <summary>
+    /// PGA ( Programmable Gain Amplifier ) Gain
+    /// </summary>
+    public enum PgaGain : byte
+    {
+        /// <summary>
+        /// 0 dB
+        /// </summary>
+        PGA_00dB = 4,
```
krwq: same comment about underscores (would be fine for non-public APIs)

src/devices/RadioTransmitter/Devices/Kt0803/PgaGain.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.RadioTransmitter
+{
+    /// <summary>
+    /// PGA ( Programmable Gain Amplifier ) Gain
```
krwq: okease remove spaces after `(` and before `)`

src/devices/RadioTransmitter/Devices/Kt0803/TransmissionPower.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    /// <summary>
+    /// Transmission Power
+    /// </summary>
+    public enum TransmissionPower : byte
+    {
+        /// <summary>
+        /// 95.5 dBuV
+        /// </summary>
+        Power_095_5dBuV = 0,
```
krwq: Please remove first underscore (second one is fine since it's separating numbers)

krwq:
Thanks @ZhangGaoxing!

PR #684, opened 2019-08-22
Removing IGpioController interface and use GpioDriver instead

joperezr:
When we originally added IGpioController, we didn't stop to think if this was the best way to go about solving the problem of having a controller work with different gpio providers (like gpio extenders). 
At the time we went with IGpioController interface as that was the natural place for this, but after several discussions with @shaggygi and @krwq we thought that it was probably not the right choice, which is also why we didn't go through the same route with SPI and I2C. For those, we picked abstract classes instead, which won't have the versioning problem. When thinking how to apply the same concept to Gpio, we realized that GpioDriver was already an abstract class and it was actually performing this specific porpouse, so this PR is removing IGpioController interface and instead moving existing gpio providers to instead be GpioDrivers themselves, so that you could create new GpioControllers that would use these drivers underneath. Here is an example of what is changed:

```c#
// How it worked before with IGpioController. Mcp23008 is an IGpioController.
Mcp23008 mcpDevice = new Mcp23008(...);
mcpDevice.OpenPin(15, PinMode.Output);
mcpDevice.Write(15, true);
```

```c#
// How it will work now that mcp is a Gpio Driver.
Mcp23008 mcpDevice = new Mcp23008(...);
GpioController controller = new GpioController(PinNumberingScheme.Logical, mcpDevice);
controller.OpenPin(15, PinMode.Output);
controller.Write(15, true);
```

This way we are no longer constrained with the problem where interfaces don't version well, plus we have more consistency since everybody that wants to use GPIO would go through a GpioController. One more advantage of this, is that we won't need to reinvent the wheel everytime we want to write a binding for a new gpio extender by adding all the validation that the controller was already doing, and we can instead simply write the actual behavior of the specific driver.


src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -438,19 +435,43 @@ private byte GetMappedAddress(Register register, Port port = Port.PortA, BankSty
...
+
+        protected override void RemoveCallbackForPinValueChangedEvent(int pinNumber, PinChangeEventHandler callback) => throw new NotImplementedException();
+
+        protected override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber) => pinNumber;
 
-        public PinMode GetPinMode(int pinNumber) => throw new NotImplementedException();
+        protected override WaitForEventResult WaitForEvent(int pinNumber, PinEventTypes eventTypes, CancellationToken cancellationToken) => throw new NotImplementedException();
 
-        public bool IsPinModeSupported(int pinNumber, PinMode mode) => throw new NotImplementedException();
+        protected override bool IsPinModeSupported(int pinNumber, PinMode mode) => true;
```
joperezr: Just spotted this is wrong, this should be `(mode == PinMode.Output || mode == PinMode.Input)`

src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -201,11 +197,12 @@ protected void InternalWriteUInt16(Register register, ushort value)
             }
         }
 
-        public virtual void Dispose()
+        protected override void Dispose(bool disposing)
         {
             _masterGpioController?.Dispose();
             _masterGpioController = null;
             _bus?.Dispose();
+            base.Dispose(disposing);
```
joperezr: NoteToSelf: _bus should be assigned to null here

shaggygi:
@joperezr haven't looked into the PR details yet, but my main concern is how will we implement bindings that are based on multiple features?  Meaning, something includes GPIO, ADC (future) and possibly a PwmController (if we brought that back)?

joperezr:
mm not sure I understand your question correctly, but I would imagine you could do something like:

```c#
public class ExampleCode
{
    void Example()
    {
        Mcp23008 driver = new Mcp23008(I2cDevice.Create(new I2cConnectionSettings(0, 0x20)));
        GpioController controller = new GpioController(PinNumberingScheme.Logical, driver);
        SomeADC someAdc = new SomeADC(controller);
        SomeNewDeviceBinding myBinding = new SomeNewDeviceBinding(someAdc, controller);
    }
}

public class SomeNewDeviceBinding
{
    private GpioController _controller;
    private SomeADC _someAdc;

    public SomeNewDeviceBinding(SomeADC adc, GpioController controller = null)
    {
        if (controller == null)
            controller = new GpioController();
        _controller = controller;
        _someAdc = adc;
    }
}

public class SomeADC
{
    private GpioController _controller;
    public SomeADC(GpioController controller = null)
    {
        if (controller == null)
            controller = new GpioController();
        _controller = controller;
    }
}
```

This way both the adc and the new device binding will be using the same controller, also if you wanted them to use different controllers because each one of them is using a different gpio expander you are able to easily do that.
Does that make sense?

shaggygi:
@joperezr Let me try and provide a few more details.

There are devices that can offer multiple features... GPIO, PWM, ADC, etc..  I understand the pros/cons of using interfaces, but I don't see how we can do that with abstract classes.

For example, this PR changes the Mcp23xxx back to inherit the `GpioDriver`.  A multifunctional device couldn't do the following:


```csharp
public class MultiFunctionBinding : GpioDriver, AdcDriver, PwmDriver
{
    // supporting code...
}
```

// And then do...

```csharp
public class ExampleCode
{
    void Example()
    {
        var driver = new MultiFunctionBinding(I2cDevice.Create(new I2cConnectionSettings(0, 0x20)));
        var gpioController = new GpioController(PinNumberingScheme.Logical, driver);
        var adcController = new AdcController(driver);  // Not developed yet.
    }
}
```

Therefore, the binding is going to need Driver properties (I guess).  And then somehow passed to the controller being used.

```csharp
public class MultiFunctionBinding
{
    public GpioDriver GpioDriver { get; set; }
    public AdcDriver AdcDriver { get; set; }
    public PwmDriver PwmDriver { get; set; }
}
```

joperezr:
> A multifunctional device couldn't do the following:

Why? couldn't MultifunciontingBinding class implement all the methods from all three Drivers (assuming there aren't any name collisions on the members) and that way you would still be able to do the code in your ExampleCode class? In the end, you would be calling  the driver through the gpioController and adcController, but other than that, there shouldn't really be a difference right? Am I missing something obvious here 😄 ?

shaggygi:
I just tried to do a very simple example of...

```csharp
public class MultiFunctionBinding : GpioDriver, AdcDriver, PwmDriver
{
    // supporting code...
}
```

And get exception of:

> Error	CS1721	Class 'MultiFunctionBinding' cannot have multiple base classes: 'GpioDriver' and 'AdcDriver'	MultiFunctionBinding	

Maybe I'm the one that is confused 🤕

krwq:
@shaggygi correct way would be i.e.:

```csharp
class MyNewDevice // no base class
{
   public GpioController GetController() { ... }
   public AdcDriver GetAdc() { ... }
}
```

src/devices/Mcp23xxx/BusAdapter.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Iot.Device.Mcp23xxx
 {
-    public abstract partial class Mcp23xxx : IGpioController
+    public abstract partial class Mcp23xxx
```
krwq: Do those need some property/base class to make it possible to use them as GpioController?

src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -5,17 +5,18 @@
 using System;
 using System.Device.Gpio;
 using System.Runtime.CompilerServices;
+using System.Threading;
 
 namespace Iot.Device.Mcp23xxx
 {
-    public abstract partial class Mcp23xxx : IGpioController
+    public abstract partial class Mcp23xxx : GpioDriver
```
krwq: nvm my previous comment

src/devices/Ssd13xx/tests/Commands/SetMultiplexRatioTests.cs
```diff
@@ -30,7 +30,6 @@ public void Get_Bytes(byte multiplexRatio, byte[] expectedBytes)
 
         [Theory]
         [InlineData(0x0E)]
-        [InlineData(0x40)]
```
krwq: ?

src/devices/Mcp23xxx/BusAdapter.cs
krwq: nvm, it's there

src/devices/Ssd13xx/tests/Commands/SetMultiplexRatioTests.cs
joperezr: As part of these changes I also ensured that all bindings that had test projects would run all tests correctly (we currently don't run them as part of CI). I did this because for some bindings (like both of the bindings I changed which were gpio extenders) this was a major change and I wanted to make sure that all tests would pass after the refactoring. That said, while running all bindings tests I found out that this one was failing, and I believe it has failed ever since it was added as I believe that this is a a value which wouldn't be considered out of range by the binding. That is why I removed this test case, in order to have all tests passing given that this one has been failing since day 1

krwq: ok

PR #685, opened 2019-08-23
Add missing XML comments to all Iot.Device.Bindings

krwq:
This is adding docs to many APIs (I didn't count but suspect around 700-1000).

I initially saw around 700 errors but they were occasionally increasing (when all errors in the project were fixed it cascaded to anything which depended on it)

- changing build to make sure it fails when new APIs doesn't contain docs
- change tests template to not generate docs (samples will still generate it but they should not have anything public in the first place and hence I didn't disable it there)


joperezr:
mmm very weird you are seeing this errors in CI:

```
##[error]ComparatorLatching.cs(12,9): error CS1591: Missing XML comment for publicly visible type or member 'ComparatorLatching.NonLatching'
```

I'm not sure why they weren't being catched by the full build before. The full build was definitely includding all of these source files, but for some reason the compiler didn't complain then about all of these.

krwq:
I will take a look, perhaps it's a dead code

joperezr:
I don't think so, also, there is a bunch of them. the log I checked had about 700 errors for missing xml docs.

krwq:
@joperezr please be gentle on this review - we can fix everything incrementally (can't look at the docs anymore)

krwq:
cc: @carlossanlop this is the PR I was talking about

src/devices/Bmp180/Bmp180.cs
```diff
@@ -226,7 +237,8 @@ internal ushort Read16BitsFromRegisterBE(byte register)
 
             return BinaryPrimitives.ReadUInt16BigEndian(bytes);            
         }
-        
+
+        /// <inheritdoc/>
```
carlossanlop: @krwq keep in mind that the porting tool currently does not know how to handle this tag.

samples/led-more-blinking-lights/TimeEnvelope.cs
```diff
@@ -19,10 +19,10 @@ public static void AddTime(IEnumerable<TimeEnvelope> envelopes, int value)
         }
     }
 
-    public TimeEnvelope(int count, bool throwOnOverthrow = true)
+    public TimeEnvelope(int count, bool throwOnOverflow = true)
```
carlossanlop: I know the class is internal, but shouldn't these methods also be documented, for easier maintenance?

carlossanlop: I mention this because I noticed that other internal classes (like Bmxx80Register) do have documentation.

src/devices/BrickPi3/Models/Sensors.cs
```diff
@@ -71,40 +107,42 @@ public enum SensorType : byte
...
     /// </summary>
     public enum SensorCustom
     {
+        /// <summary>Enable 9V out on pin 1 (for LEGO NXT Ultrasonic sensor).</summary>
         Pin1_9V = 0x0002,
+        /// <summary>Set pin 5 state to output.Pin 5 will be set to input if this flag is not set.</summary>
         Pin5_Out = 0x0010,
+        /// <summary>
+        /// If PIN5_OUT is set, this will set the state to output high, otherwise the state will
+        /// be output low.If PIN5_OUT is not set, this flag has no effect.
```
carlossanlop:
```suggestion
        /// be output low. If PIN5_OUT is not set, this flag has no effect.
```

src/devices/BrickPi3/Models/Sensors.cs
```diff
@@ -71,40 +107,42 @@ public enum SensorType : byte
...
+        /// </summary>
         Pin5_State = 0x0020,
+        /// <summary>Set pin 6 state to output.Pin 6 will be set to input if this flag is not set.</summary>
```
carlossanlop:
```suggestion
        /// <summary>Set pin 6 state to output. Pin 6 will be set to input if this flag is not set.</summary>
```

src/devices/BrickPi3/Models/Sensors.cs
```diff
@@ -71,40 +107,42 @@ public enum SensorType : byte
...
         Pin6_Out = 0x0100,
+        /// <summary>
+        /// If PIN6_OUT is set, this will set the state to output high, otherwise the state will
```
carlossanlop: Do we have a rule in our coding guidelines about adding endlines in the middle of sentences? I ask because when we port these multi-line comments, they will have to be manually merged into a single line (unless there is a dot or a colon).

src/devices/BrickPi3/Models/Sensors.cs
```diff
@@ -71,40 +107,42 @@ public enum SensorType : byte
...
+        /// be output low.If PIN6_OUT is not set, this flag has no effect.
```
carlossanlop:
```suggestion
        /// be output low. If PIN6_OUT is not set, this flag has no effect.
```

src/devices/BrickPi3/Movement/Motor.cs
```diff
@@ -153,6 +159,10 @@ public int GetSpeed()
         public int Speed
         { get { return GetSpeed(); } set { SetSpeed(value); } }
 
+        /// <summary>
+        /// Motor port
+        /// </summary>
+        /// <value></value>
```
carlossanlop: Interesting. First time I see someone explicitly add the `<value></value>` tags, which are actually required to get documented in the dotnet-api-docs repo for Properties.

The `<value>` tags is used to describe the Property data type that can the user can get and/or set.

It' clearer to explain with booleans. For example:

```cs
/// <summary>Determines if the instance is enabled.</summary>
/// <value><see langword="true" /> if enabled, <see langword="false" /> if it's disabled.</value>
public bool IsEnabled { get; }
```

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -101,14 +121,42 @@ public virtual bool BacklightOn
...
 
-        protected virtual bool SetTwoLineMode(int rows) => rows > 1;
+        /// <summary>
+        /// Send commands to the device
+        /// </summary>
+        /// <param name="commands">Each byte represents command being sent to the device</param>
+        protected void SendCommands(ReadOnlySpan<byte> commands) => _interface.SendCommands(commands);
 
+        /// <summary>
+        /// Determines if device should use two line mode
```
carlossanlop:
```suggestion
        /// Determines if the device should use two line mode
```

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -368,6 +416,11 @@ public void Write(string value)
             ArrayPool<byte>.Shared.Return(buffer);
         }
 
+        /// <summary>
+        /// Releases unmanaged resources used by Hd44780
+        /// and optionally release managed resources
+        /// </summary>
+        /// <param name="disposing">True to release both managed and unmanaged resources; false to release only unmanaged resources</param>
```
carlossanlop:
```suggestion
        /// <param name="disposing"><see langword="true" /> to release both managed and unmanaged resources; <see langword="false" /> to release only unmanaged resources.</param>
```

src/devices/CharacterLcd/LcdInterface.cs
```diff
@@ -70,10 +92,16 @@ public virtual void WaitForNotBusy(int microseconds)
         /// </summary>
         public abstract bool BacklightOn { get; set; }
 
+        /// <summary>
+        /// Releases unmanaged resources used by LcdInterface
+        /// and optionally release managed resources
+        /// </summary>
+        /// <param name="disposing">True to release both managed and unmanaged resources; false to release only unmanaged resources</param>
```
carlossanlop:
```suggestion
        /// <param name="disposing"><see langword="true" /> to release both managed and unmanaged resources; <see langword="false" /> to release only unmanaged resources.</param>
```

src/devices/Common/Iot/Device/Graphics/BitmapImage.cs
```diff
@@ -7,8 +7,18 @@
 
 namespace Iot.Device.Graphics
 {
+    /// <summary>
+    /// Represents bitmap image
+    /// </summary>
     public abstract class BitmapImage
     {
+        /// <summary>
+        /// Constructs BitmapImage instance
```
carlossanlop: Constructors normally use a sentence like this:

```suggestion
        /// Initializes a <see cref="T:Iot.Device.Graphics.BitmapImage" /> instance with the specified data, width, height and stride.
```

src/devices/Directory.Build.props
```diff
@@ -6,5 +6,7 @@
     <DeterministicSourcePaths>false</DeterministicSourcePaths>
     <IsPackable>false</IsPackable>
     <MainLibraryPath>$(MSBuildThisFileDirectory)../System.Device.Gpio/</MainLibraryPath>
```
carlossanlop: Hey since you're documenting everything in these files, should you also set to true the prop that throws a compiler error when a public API is not documented?

src/devices/GoPiGo3/Models/Motors.cs
```diff
@@ -6,29 +6,32 @@
...
-    /// flags -- 8-bits of bit-flags that indicate motor status:
-    /// bit 0 -- LowVoltageFloat - The motors are automatically disabled because the battery voltage is too low
-    /// bit 1 -- Overloaded - The motors aren't close to the target (applies to position control and dps speed control).
+    /// Flags that indicate motor status
     /// </summary> 
     public enum MotorStatusFlags
     {
+        /// <summary>No problems with motors detected</summary>
         AllOk = 0,
+        /// <summary>LowVoltageFloat - The motors are automatically disabled because the battery voltage is too low</summary>
```
carlossanlop: The name in the summary is not necessary.
```suggestion
        /// <summary>The motors are automatically disabled because the battery voltage is too low.</summary>
```

src/devices/GoPiGo3/Models/Motors.cs
```diff
@@ -6,29 +6,32 @@
...
         LowVoltageFloat = 0x01,
+        /// <summary>Overloaded - The motors aren't close to the target (applies to position control and dps speed control)</summary>
```
carlossanlop: The name in the summary is not necessary:
```suggestion
        /// <summary>The motors aren't close to the target (applies to position control and dps speed control).</summary>
```

src/devices/GoPiGo3/Models/Motors.cs
```diff
@@ -6,29 +6,32 @@
...
         Overloaded = 0x02
     }
 
     /// <summary>
-    /// Set quickly a speed for the motor
+    /// Set speed for the motor
```
carlossanlop:
```suggestion
    /// Sets the speed for the motor.
```

src/devices/Ads1115/ComparatorQueue.cs
```diff
@@ -5,13 +5,17 @@
 namespace Iot.Device.Ads1115
 {
     /// <summary>
-    /// Set Comparator Queue
+    /// Comparator Queue
```
carlossanlop: Here and elsewhere: All sentences should end with a dot.

joperezr:
> if we ask @mairaw Maira Wenzel FTE and @rpetrusha Ron Petrusha FTE to help add proper documentation to these files before we port them all to the Docs repo.

All of this docs won't go in the docs repo as this package is community maintained. This means that the level of support and documentation is not the same as corefx, coreclr, et. al. For this reason, this PR shouldn't be blocked and is why @krwq mentioned that any big changes can be addressed later.

src/devices/Bmp180/Bmp180.cs
joperezr: These won't be ported.

samples/led-more-blinking-lights/TimeEnvelope.cs
joperezr: The bar we have for this package is that you have to document your public API but that's it. Some community members decide to also document their internal APIs but this is optional

src/devices/Directory.Build.props
joperezr: This is doing it. By saying you want to generate a documentation file, compiler will throw warnings if public api is not documented. That is why we are setting the second property to treat warnings as errors.

src/devices/BrickPi3/Movement/Motor.cs
krwq: My editor was adding `<value>` but I kept removing it because I've never seen anyone do that and thought this was the tooling issue.

I'm not sure if I'll address this one everywhere but will consider changing the keywords for now at least

krwq:
@carlossanlop sorry for confusing, I wasn't sure if we were planning to port this - the main reason of this PR was to enforce people adding xml docs but we had to fix all of the gaps first. I wasn't meant to be perfect on the first pass

src/devices/Ads1115/ComparatorQueue.cs
krwq: Actually some of the stuff I added xml docs to I wasn't sure how it was supposed to be used but if I wanted to document everything super accurately I'd have to spend week-month instead of a day. Since the main goal was to enforce xml docs I'd prefer we do first pass to enable global flag and then incrementally improve it from there (so that new PRs will not cause build breaks etc)

src/devices/BrickPi3/Models/Sensors.cs
krwq: I'm generally fine with that as part of this PR. I suspect in the future we can modify the tool to be more noisy and report all of these

krwq:
Merging this to not block the remaining PRs on this

PR #692, opened 2019-08-29
Pca9685/Servo improvements

krwq:
- allow to combine it with other things like i.e. ServoMotor class
- added sample with calibration of the servo
- renamed Pca9685 namespace to `Iot.Device.Pwm` to avoid collision between namespace and class name
- ServoMotor now allows to take fraction angles (there is no point in blocking that)
  - one thing which I also considered is to rename Angle in API names to Value since Angle doesn't makes much sense for linear servos and also some people might prefer to use different units than angle (i.e. 0.0-1.0 or 0-1000) since it might mean something else depending on what you attach to it
  - other thing which I think we should do is to rename ServoMotor namespace to avoid double typing `ServoMotor` when using `ServoMotor` namespace (to disambiguate and quiet down compiler) - I couldn't think of any good name but I think we should do it regardless (perhaps `Actuators` - we could move `DCMotor` in there too)

cc: @HumJ0218 @shaggygi (for tiny adjustments to ServoMotor)

### DIFF
--- a/src/devices/Pca9685/Pca9685.cs
+++ b/src/devices/Pca9685/Pca9685.cs
@@ -3,135 +3,165 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Buffers.Binary;
+using System.Device;
 using System.Device.I2c;
+using System.Device.Pwm;
+using System.Diagnostics;
 using System.Threading;
-using static Iot.Device.Pca9685.Mode1;
-using static Iot.Device.Pca9685.Mode2;
-using static Iot.Device.Pca9685.Register;
 
-namespace Iot.Device.Pca9685
+namespace Iot.Device.Pwm
 {
+    using static Mode1;
+    using static Iot.Device.Pwm.Mode2;
+    using static Iot.Device.Pwm.Register;
+
     /// <summary>
     /// PCA9685 PWM LED/servo controller
     /// </summary>
     public class Pca9685 : IDisposable
     {
         /// <summary>
-        /// I2C Device
+        /// I2C address base. Equal to actual address when all selectable address bits are equal to 0.
         /// </summary>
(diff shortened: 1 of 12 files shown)


src/devices/Pca9685/Pca9685.cs
```diff
@@ -3,186 +3,354 @@
...
         /// </summary>
         /// <param name="i2cDevice">The I2C device to be used</param>
-        public Pca9685(I2cDevice i2cDevice)
+        /// <param name="pwmFrequency">Initial PWM frequency on all channels</param>
+        /// <param name="dutyCycleAllChannels">Duty cycle set on all channels after constructing</param>
+        /// <param name="usingExternalClock">
+        /// When true specifies that external clock is used.
+        /// Use <see cref="ClockFrequency"/> to set frequency of an external clock
+        /// </param>
+        public Pca9685(I2cDevice i2cDevice, double pwmFrequency = 100.0, double dutyCycleAllChannels = 0.0, bool usingExternalClock = false)
```
HumJ0218: PRE_SCALE register has a default value `0x1E` with the PWM frequency is about 196.8876 Hz
Is it good to set `pwmFrequency` to 100.0 ?

HumJ0218: and so do `dutyCycleAllChannels`

krwq:
need to rebase this

src/devices/Pca9685/Pca9685.cs
krwq: Ok, I've just changed the binding to default to -1 for PWM frequency: that will cause prescale to be read from the device and reported back. Same for duty cycle - it won't get changed unless use explicitly changes it

krwq:
@joperezr I'm merging this as is, I'll address any feedback in a separate PR

PR #694, opened 2019-08-30
Fritzing Changed for Mcp3008

Frankenslag:
It looks like the Fritzing diagram and the generated png for the Mcp3008 Hardware SPI sample is incorrect. I believe that it should be as follows.

rpi-trimpot_spi


krwq:
@Frankenslag what's the diff? I think there are two pictures, one in the Mcp3008 root and another in samples. Samples look correct.

Also there is a tiny glitch in the picture above (dead connection on the right side of RPi near the top of the picture)

Frankenslag:
> @Frankenslag what's the diff? I think there are two pictures, one in the Mcp3008 root and another in samples. Samples look correct.
> Also there is a tiny glitch in the picture above (dead connection on the right side of RPi near the top of the picture)

See below note that the tiny glitch is only in the image above. It looks like CS is correct in the original 

In the original it seems that

mosi => clk
miso => dout
clk => din

and so I believe that mosi and clk are incorrect from where I am sitting

image

Issue #681, opened 2019-08-22
Can you please implement the driver for MCP3204?


pankaj-nikam:
Hi team,

Thank you for contributing towards the IoT devices. I would like to have a driver for MCP3204. It would be great if it is implemented. At least a base class from which these devices can inherit and use it. For example a base class of MCP32XX can help us implement the classes for MCP3204, MCP3208 etc. somewhat like what GPIO Zero library.

In case you want the datasheet for MCP3204, it can be found in the link below:

http://ww1.microchip.com/downloads/en/DeviceDoc/21298e.pdf

Thanks in advance.

Frankenslag:
@joperezr 
Happy to look at this if you want to assign me. 

@pankaj-nikam 
Not sure which GPIO Zero library you mean. I believe that we would want to keep bindings for these ADC's looking like bindings for ADC's that we already  have.

krwq:
@Frankenslag yes please try to make it as similar to Mcp3008 as possible (possible even some of the code can be reused)

pankaj-nikam:
Hi @Frankenslag, 

Sorry for not being clear, what I meant to say is the way GPIOZero is implemented, the base class is for MCP3XXX where the derived classes have the correct implementation - this can help code be clean. Thats all what I wanted to convey. In case you are wondering the docs of the same, please check: https://gpiozero.readthedocs.io/en/stable/api_spi.html#base-classes

Thanks.

Frankenslag:
I think that it would be possible to have the ADC's derived as you describe but I think that first I will need to get the Mcp3204 and 3208 working and then discuss if we take the effort to include the others. We would have to be careful as people may be already using the 3008.

pankaj-nikam:
Hi @Frankenslag , I agree with you. We should not break the existing code. What we can do is create a super base class which handles the common things, then a 32xx and a 30xx which inherit from it; so that the 10 bit and 12 bit things can be handled differently. As the new devices get added, it will still be working. What do you think?
Thanks.

krwq:
@pankaj-nikam @Frankenslag I'm closing this since PR is merged, feel free to reopen if you think there is anything left out or any follow up work

PR #696, opened 2019-08-31
WIP Addition of Mcp3Adc bindings for Mcp32xx and Mcp30xx ADC's

Frankenslag:
Applies to #681 @pankaj-nikam

This PR aims to provide bindings for the following 12 bit and 14 bit Micorochip Analog to Digital Converters

- Mcp3001
- Mcp3002
- Mcp3004
- Mcp3008
- Mcp3201
- Mcp3202
- Mcp3204
- Mcp3208

Mcp3008 already exists as a binding and so if this binding is approved I am unsure what should happen to the original.

These bindings have not been fully tested (I need some more of the hardware) and so I am opening this PR for comments rather than expecting the work to be merged.


pankaj-nikam:
Hi @Frankenslag , Thank you. I will have this tested and will update.

krwq:
@Frankenslag We already have binding for Mcp3008, any chance you could merge those together?

src/devices/Mcp3Adc/Mcp3001.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Device.Spi;
+
+namespace Iot.Device.Mcp3Adc
```
krwq: Please use Iot.Device.Adc namespace

src/devices/Mcp3Adc/Mcp3002.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <param name="spiDevice">Device used for SPI communication</param>
+        public Mcp3002(SpiDevice spiDevice) : base(spiDevice, pinCount: 2) { }
+
+        /// <summary>
+        /// Reads 10-bit (0..1023) value from the device
+        /// </summary>
+        /// <param name="channel">Channel which value should be read from (valid values: 0-1 or 0 with differential inputs)</param>
+        /// <param name="inputType">The type of input pin to read</param>
+        /// <returns>10-bit value corresponding to relative voltage level on specified device channel</returns>
+        public int Read(int channel, InputType inputType = InputType.SingleEnded)
```
krwq: Please split differential and single ended reading since channel means something else for differential

krwq: optionally you can remove differential at all as I did in Mcp3008 (you can just read twice and subtract numbers)

krwq:
@Frankenslag please address all 3 comments before I look at the implementation details

src/devices/Mcp3Adc/Mcp3001.cs
Frankenslag: Done in the latest PR update

src/devices/Mcp3Adc/Mcp3002.cs
Frankenslag: I have added ReadPseudoDifferential. The differential implementation on these ADC's is intended for reducing common-mode interference rather than give a true differential value as I understand it. So if you wanted to, for example, compare set-point with process value for a PID controller then you would take two readings and subtract as you describe.

Frankenslag:
> @Frankenslag We already have binding for Mcp3008, any chance you could merge those together?

@krwq

I have left my implemention in as it is in a different namespace than the standard Mcp3008. 

How would you expect that this is merged ie.

- Remove the new implementation of the 3008
- Remove the old one (I assume people have been using it ?)
- Modify the existing 3008 to inherit from the new abstract class ?

Appreciate your views and guidance

krwq:
@Frankenslag - remove/merge current code of Mcp3008 into your binding but review the APIs and see if there is something useful you can adapt (MCP3008 already went through couple of reviews so APIs are mostly stable at this point)

src/devices/Mcp3Adc/Mcp3004.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Reads a 10-bit (0..1023) value from the device using pseudo-differential inputs
+        /// </summary>
+        /// <param name="channel">Channel which value should be read from (valid values: 0-1). Channel 0 is on inputs 0 and 1 and
+        /// Channel 1 is on inputs 2 and 3.</param>
+        /// <param name="polarityInverted">When false then then inputs 0 and 2 are positive and inputs 1 and 3 are negative
+        /// when true the polarity is swapped so that inputs 0 and 2 are negative and inputs 1 and 3 positive.</param>
+        /// <returns>10-bit value corresponding to relative voltage level on specified device channel.</returns>
+        public int ReadPseudoDifferential(int channel, bool polarityInverted = false)
```
krwq: would it make sense to have something like `ReadDifferential(int referenceChannel, int channel)` and then if hardware has a way of representing this combination then use that, if not then just subtract two reads (perhaps add flag like forceHardwareImplementation which would cause it to throw if such is not available)

src/devices/Mcp3Adc/Mcp30xx32xx.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                InputType.Differential => channel * 2,
+                InputType.InvertedDifferential => channel * 2 + 1,
+                _ => channel
+            };
+
+            // create a value to represent the request to the ADC 
+            requestVal = _pinCount switch
+            {
+                var x when (x == 4 || x == 8) => (inputType == InputType.SingleEnded ? 0b1_1000 : 0b1_0000) | channelVal,
```
krwq: wow, I didn't know this syntax

src/devices/Mcp3Adc/Mcp3202.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <param name="spiDevice">Device used for SPI communication</param>
+        public Mcp3202(SpiDevice spiDevice) : base(spiDevice, pinCount: 2) { }
+
+        /// <summary>
+        /// Reads a 12-bit (0..4096) value from the device using pseudo-differential inputs
+        /// </summary>
+        /// <param name="polarityInverted">When false then then input 0 is positive and input 1 is negative
+        /// when true the polarity is swapped so that input 0 is negative and input 1 is positive.</param>
+        /// <returns>12-bit value corresponding to relative voltage level on specified device channel.</returns>
+        public int ReadPseudoDifferential(bool polarityInverted = false)
```
krwq: if this took 2 channels instead then inverting could be achieved by swapping channels (and hardware implementation could figure out the switch might be useful)

src/devices/Mcp3Adc/Mcp3001.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
krwq: I'm unsure I like the folder name... Would something like Mcpx00x make sense?

krwq: i.e. MCP3421 seems to also be ADC (not sure if similar). If there are any exceptions in the future we can figure it out at that point

src/devices/Mcp3Adc/Mcp3004.cs
Frankenslag: @krwq
Makes perfect sense. I don't think that we need to 'force' the matter as the primary classes will know if the hardware supports true differential (which is the 3300 series I believe). Perhaps the channels should be named positiveChannel and negativeChannel as this fits in with the datasheet description. I also think that we should use the same two channel approach for ReadPseudoDifferential. Please give me a few days and I'll make these changes.

src/devices/Mcp3Adc/Mcp30xx32xx.cs
Frankenslag: Neither did I until I looked for it. It is much more readable to me than a series of if/then/else's or ternary expressions.

src/devices/Mcp3Adc/Mcp3001.cs
Frankenslag: The 3421 is I2C and fairly different in how it is controlled. Perhaps for the future. I will change the folder for the next PR.

Frankenslag:
> Hi @Frankenslag , Thank you. I will have this tested and will update.

Please be aware that there are a number of changes happening right now.

src/devices/Mcp3Adc/Mcp3004.cs
Frankenslag: @krwq
OK I have done a good deal of refactoring, added the 33xx family and other work.

 I have added support for using positive and negative channels. On the 32xx and 30xx family then Read differential will always just subtract values as hardware support is not available. On the 33xx family then ReadDifferential  checks  to see if the supplied channels exist and are in a valid pair and if they are then the hardware is used otherwise the values are subtracted like on the 32xx and 30xx series.

I have copied over the sample from the original Mcp3008 binding but I have not yet deleted Mcp3008 as there are other samples that are using that binding and before going ahead I wanted to be sure I am on the right track.

I haven't tested the 33xx code as I don't have any hardware samples but they should be with me real soon.

src/devices/Mcp3xxx/Mcp30xx32xx.cs
```diff
@@ -0,0 +1,205 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Reads a  value from the device using pseudo-differential inputs
+        /// </summary>
+        /// <param name="positiveChannel">Channel which represents the signal (valid values: 0 to pincount - 1).</param>
+        /// <param name="negativeChannel">Channel which represents the signal ground (valid values: 0 to pincount - 1).</param>
+        /// <returns>A value corresponding to relative voltage level on specified device channels</returns>
+        public virtual int ReadPseudoDifferential(int positiveChannel, int negativeChannel)
```
krwq: I'd prefer these to be called `referenceChannel` and `valueChannel` (or something along these lines) as I think this would be more clear what's relative to what

src/devices/Mcp3xxx/Mcp30xx32xx.cs
```diff
@@ -0,0 +1,205 @@
...
+        /// <param name="negativeChannel">Channel which represents the negative signal (valid values: 0 to pincount - 1).</param>
+        /// <returns>A value corresponding to relative voltage level on specified device channels</returns>
+        public virtual int ReadDifferential(int positiveChannel, int negativeChannel)
+        {
+            CheckChannelRange(positiveChannel, PinCount);
+            CheckChannelRange(negativeChannel, PinCount);
+
+            if (positiveChannel == negativeChannel)
+            {
+                throw new ArgumentException($"ADC differential channels must be different.", nameof(positiveChannel) + " " + nameof(negativeChannel));
```
krwq: it would also be correct to return 0 although it's more likely a bug so makes sense to leave as is :)

src/devices/Mcp3xxx/Mcp3xxx.cs
```diff
@@ -0,0 +1,93 @@
...
+            // take the resuest and put it in a byte array
+            for (int i = 0; i < requestBuffer.Length; i++)
+            {
+                requestBuffer[i] = (byte)(adcRequest >> (adcRequestLengthBytes - i - 1) * 8);
+            }
+
+            _spiDevice.TransferFullDuplex(requestBuffer, responseBuffer);
+
+            // transfer the response from the ADC into the return value
+            for (int i=0; i < responseBuffer.Length; i++)
```
krwq: nit: spaces around `=`

src/devices/Mcp3xxx/Mcp3xxx.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        /// <summary>
+        /// Reads a value from the device
+        /// </summary>
+        /// <param name="adcRequest">A bit pattern to be sent to the ADC.</param>
+        /// <param name="adcRequestLengthBytes">The number of bytes to be sent to the ADC containing the request and returning the response.</param>
+        /// <param name="adcResolutionBits">The number of bits in the returned value</param>
+        /// <param name="delayBits">The number of bits to be delayed between the request and the response being read.</param>
+        /// <returns>A value corresponding to a voltage level on the input pin described by the request.</returns>
+        protected int Read(int adcRequest, int adcRequestLengthBytes, int adcResolutionBits, int delayBits)
```
krwq: optional: I think it might make sense to rename this method to something else (i.e. ReadInternal) or something along these lines to make it clear it's only meant to be used by implementation

krwq:
@Frankenslag I haven't looked at any implementation details (only API shape and a bit style) and this mostly looks good to me. Consider removing WIP from name on the next iteration

src/devices/Mcp3xxx/Mcp30xx32xx.cs
Frankenslag: I have done this and I sort of agree that the names are better however on all of the datasheets (not exhaustive but including other ADC manufacturers) they are described in the terms positive and negative.

Frankenslag:
> @Frankenslag I haven't looked at any implementation details (only API shape and a bit style) and this mostly looks good to me. Consider removing WIP from name on the next iteration

O.K. I have actioned the most recent comments and updated the PR. I have also added the sample by mostly copying from Mcp3008. Hope all looks good.

src/devices/Mcp3xxx/Mcp33xx.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using System;
+using System.Device.Spi;
+
+namespace Iot.Device.Adc
+{
+    /// <summary>
+    /// MCP3304 Analog to Digital Converter (ADC)
+    /// </summary>
+    public abstract class Mcp33xx : Mcp30xx32xx
```
krwq: nit: 33xx doesn't match either 30xx or 32xx consider renaming base class to `3xxx`

src/devices/Mcp3xxx/samples/README .md
```diff
@@ -0,0 +1,116 @@
...
+The sample is based on following resources:
+
+* [Analog Inputs for Raspberry Pi Using the MCP3008](https://learn.adafruit.com/reading-a-analog-in-and-controlling-audio-volume-with-the-raspberry-pi) 
+* [Raspberry Pi Analog to Digital Converters](https://learn.adafruit.com/raspberry-pi-analog-to-digital-converters).
+* [Raspbery Pi Analog Input with MCP3008](https://gist.github.com/ladyada/3151375)
+* [MCP3008.py](https://github.com/adafruit/Adafruit_Python_MCP3008/blob/master/Adafruit_MCP3008/MCP3008.py)
+
+Major thanks to [Adafruit](https://adafruit.com) for providing python implementations, which were ported to C# for this sample.
+
+See [Using .NET Core for IoT Scenarios](../README.md) for more samples.
```
krwq: please remove Mcp3008 folder since it's now included in here

src/devices/Mcp3xxx/Mcp30xx32xx.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        private byte _adcResolutionBits;
+
+        /// <summary>
+        /// Constructs Mcp30xx32xx instance
+        /// </summary>
+        /// <param name="spiDevice">Device used for SPI communication</param>
+        /// <param name="pinCount">Value representing the number of pins on the device.</param>
+        /// <param name="adcResolutionBits">The number of bits of resolution for the ADC.</param>
+        public Mcp30xx32xx(SpiDevice spiDevice, byte pinCount, byte adcResolutionBits) : base(spiDevice)
```
krwq: Does pinCount refer to number of ADC channels? If so please rename to `channelCount` (here and everywhere where applies) - if not please clarify docs

src/devices/Mcp3xxx/Mcp30xx32xx.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        /// <summary>
+        /// Checks that the channel is in range of the available input pins and that both pins are part of a valid pairing of
+        /// input pins.
+        /// </summary>
+        /// <param name="positiveChannel">Positive channel to be checked</param>
+        /// <param name="negativeChannel">Negative channel to be checked</param>
+        /// <param name="channelCount">Value representing the number of channels on the device which may vary depending on the configuration.
+        /// If channelCount is not set then the PinCount is used.</param>
+        protected void CheckChannelPairing(int positiveChannel, int negativeChannel, int channelCount = -1)
```
krwq: nit: please rename args to match ReadPseudoDifferential

src/devices/Mcp3xxx/Mcp30xx32xx.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// </summary>
+        /// <param name="valueChannel">Channel which represents the signal (valid values: 0 to pincount - 1).</param>
+        /// <param name="referenceChannel">Channel which represents the signal ground (valid values: 0 to pincount - 1).</param>
+        /// <returns>A value corresponding to relative voltage level on specified device channels</returns>
+        public virtual int ReadPseudoDifferential(int valueChannel, int referenceChannel)
+        {
+            // ensure that the channels are part of the same pairing
+            CheckChannelPairing(valueChannel, referenceChannel, PinCount / 2);
+
+            return ReadInternal(valueChannel / 2, valueChannel > referenceChannel ? InputType.InvertedDifferential : InputType.Differential, _adcResolutionBits);
```
krwq: I'm confused on how this division works, wouldn't passing 2 and 3 produce same result? Please add a comment to clarify assuming current code is correct

src/devices/Mcp3xxx/Mcp30xx32xx.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            // Check that the channel pins are adjacent and part of the same differential pairing
+            if (positiveChannel / 2 != negativeChannel / 2 || positiveChannel == negativeChannel)
+            {
+                throw new ArgumentException($"ADC differential channels must be different and part of the same channel pairing.", nameof(positiveChannel) + " " + nameof(negativeChannel));
+            }
+        }
+
+        /// <summary>
+        /// Reads a  value from the device using pseudo-differential inputs
```
krwq: It would be useful to shortly explain what pseudo-differential means so that people who just need ADC and have first time experience with hardware can tell which method should they use :) (same comment for full differential)

src/devices/Mcp3xxx/Mcp30xx32xx.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <param name="negativeChannel">Negative channel to be checked</param>
+        /// <param name="channelCount">Value representing the number of channels on the device which may vary depending on the configuration.
+        /// If channelCount is not set then the PinCount is used.</param>
+        protected void CheckChannelPairing(int positiveChannel, int negativeChannel, int channelCount = -1)
+        {
+            CheckChannelRange(positiveChannel, (channelCount == -1) ? PinCount : channelCount);
+            CheckChannelRange(negativeChannel, (channelCount == -1) ? PinCount : channelCount);
+
+            // Check that the channel pins are adjacent and part of the same differential pairing
+            if (positiveChannel / 2 != negativeChannel / 2 || positiveChannel == negativeChannel)
```
krwq: please explain the division (or create method to make it simpler to reason about)

src/devices/Mcp3xxx/Mcp30xx32xx.cs
```diff
@@ -0,0 +1,149 @@
...
+            return ReadInternal(valueChannel / 2, valueChannel > referenceChannel ? InputType.InvertedDifferential : InputType.Differential, _adcResolutionBits);
+        }
+
+        /// <summary>
+        /// Reads a value from the device using differential inputs
+        /// </summary>
+        /// <param name="positiveChannel">Channel which represents the positive signal (valid values: 0 to pincount - 1).</param>
+        /// <param name="negativeChannel">Channel which represents the negative signal (valid values: 0 to pincount - 1).</param>
+        /// <returns>A value corresponding to relative voltage level on specified device channels</returns>
+        public virtual int ReadDifferential(int positiveChannel, int negativeChannel)
```
krwq: rename args

src/devices/Mcp3xxx/Mcp30xx32xx.cs
Frankenslag: I wanted to distinguish between the word channel and pins as in the earlier api then channel could refer to a channel pair which there were pins/2 of. Now that the api has changed then this can be changed.

src/devices/Mcp3xxx/Mcp30xx32xx.cs
Frankenslag: I have added a comment. With differential inputs a channel pairing is read and the divide by 2 is to work out which pairing the valueChannel is in.

src/devices/Mcp3xxx/samples/README .md
Frankenslag: @krwq
I believed I have have addressed all the comments and I have posted another PR update,

src/devices/Mcp3xxx/samples/Mcp3008.Sample.cs
```diff
@@ -21,7 +23,7 @@ static void Main(string[] args)
             using (SpiDevice spi = new SoftwareSpi(clk: 6, miso: 23, mosi: 5, cs: 24))
             // For hardware implementation replace it with following
             // using (SpiDevice spi = SpiDevice.Create(hardwareSpiSettings))
-            using (Mcp3008.Mcp3008 mcp = new Mcp3008.Mcp3008(spi))
+            using (Mcp3008 mcp = new Mcp3008(spi))
```
krwq: that's why I asked to change the namespace so that we don't have to double type `Mcp3008` 😄 Thanks!

src/devices/Mcp3xxx/samples/Mcp3008.Sample.cs
```diff
@@ -7,6 +7,8 @@
 using System.Threading;
 using Iot.Device.Spi;
 
+using Iot.Device.Adc;
```
krwq: nit: enter between namespaces not needed (optional)

src/devices/Mcp3xxx/README.md
```diff
@@ -0,0 +1,25 @@
+# MCP3000, MCP3200 and MCP3300 ranges of Analog to Digital Converters
```
krwq: please rename this to something like:

`MCP3xxx family of Analog to Digital Converters`

(this title is picked by the script which generates device listing (src/devices/README.md) so I'm more picky about this)

src/devices/Mcp3xxx/Mcp3xxx.csproj
```diff
@@ -2,15 +2,17 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
-    <EnableDefaultItems>false</EnableDefaultItems> <!--Disabling default items so samples source won't get build by the main library-->
+    <!--Disabling default items so samples source won't get build by the main library-->
+    <EnableDefaultItems>false</EnableDefaultItems>
```
krwq: I think the default items are `<Compile Include="*.cs" />` so you might wanna try removing this as well as `Compile` item

src/devices/Mcp3xxx/Mcp3208.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Iot.Device.Adc
+{
+    /// <summary>
+    /// MCP3208 Analog to Digital Converter (ADC)
+    /// </summary>
+    public class Mcp3208 : Mcp3xxx
+    {
+        /// <summary>
+        /// Constructs Mcp3008 instance
```
krwq: Mcp3208

src/devices/Mcp3xxx/Mcp3301.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// Reads a 13-bit signed value from the device
+        /// </summary>
+        /// <returns>Integer value corresponding to relative voltage level on specified device channel</returns>
+        public int Read()
+        {
+            int retval = ReadInternal(adcRequest: 0, adcRequestLengthBytes: 2, adcResolutionBits: 13, delayBits: 0);
+
+            //convert 13 bit signed to 32 bit signed
+            return (retval >> 12) == 0 ? retval : (int)(0xFFFFE000 | retval);
```
krwq: would it make sense to instead subtract `2^12` (+/- 1)? This is not exactly immediately obvious what's going on over here

src/devices/Mcp3xxx/Mcp3301.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Reads a 13-bit signed value from the device
+        /// </summary>
+        /// <returns>Integer value corresponding to relative voltage level on specified device channel</returns>
+        public int Read()
+        {
+            int retval = ReadInternal(adcRequest: 0, adcRequestLengthBytes: 2, adcResolutionBits: 13, delayBits: 0);
+
+            //convert 13 bit signed to 32 bit signed
```
krwq: nit: space after `//`

src/devices/Mcp3xxx/Mcp3304.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Iot.Device.Adc
+{
+    /// <summary>
+    /// MCP3304 Analog to Digital Converter (ADC)
+    /// </summary>
+    public class Mcp3304 : Mcp33xx
+    {
+        /// <summary>
+        /// Constructs Mcp3302 instance
```
krwq: 3304

src/devices/Mcp3xxx/Mcp3304.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// </summary>
+        /// <param name="spiDevice">Device used for SPI communication</param>
+        public Mcp3304(SpiDevice spiDevice) : base(spiDevice, channelCount: 8, adcResolutionBits: 13) { }
```
krwq: this one is unsigned I presume (curious because previous 13-bit was signed)

src/devices/Mcp3xxx/Mcp33xx.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Reads a  value from the device using pseudo-differential inputs
+        /// </summary>
+        /// <param name="valueChannel">Channel which represents the signal (valid values: 0 to channelcount - 1).</param>
+        /// <param name="referenceChannel">Channel which represents the signal ground (valid values: 0 to channelcount - 1).</param>
+        /// <returns>A value corresponding to relative voltage level on specified device channels</returns>
+        public override int ReadPseudoDifferential(int valueChannel, int referenceChannel)
+        {
+            throw new NotSupportedException("Mcp33xx device does not support ReadPseudoDifferential.");
```
krwq: `$"..... {nameof(ReadPseudoDifferential)} ....."` - this is making it easier to rename this if we decide to do so in the future (editor will do it for you)

src/devices/Mcp3xxx/Mcp33xx.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+            // check if it is possible to use hardware differential because both input channels are in the same differential channel pairing
+            if (valueChannel / 2 == referenceChannel / 2)
+            {
+                // read a value from the ADC where the channel is the channel pairing
+                retval = ReadInternal(channel: valueChannel / 2, valueChannel > referenceChannel ? InputType.InvertedDifferential : InputType.Differential, adcResolutionBits: 13);
+
+                //convert 13 bit signed to 32 bit signed
+                retval = (retval >> 12) == 0 ? retval : (int)(0xFFFFE000 | retval);
```
krwq: would it make sense to add some utility for this? (i.e. ReadSignedInternal or just a static method to convert) - also same comment - might be better to subtract instead to make it clearer

src/devices/Mcp3xxx/Mcp33xx.cs
```diff
@@ -0,0 +1,75 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.Spi;
+
+namespace Iot.Device.Adc
+{
+    /// <summary>
+    /// MCP3304 Analog to Digital Converter (ADC)
```
krwq: MCP33XX family of Analog to Digital Converters

src/devices/Mcp3xxx/Mcp3Base.cs
```diff
@@ -0,0 +1,169 @@
...
+    // M = SingleEnded
+    // D = Delay
+    // N = Null Bit = 0
+    // R = Response
+    // - = Sign Bit
+    // x = Dont Care
+    //
+
+    /// <summary>
+    /// Mcp3Base Abstract class representing the MCP ADC devices.
```
krwq: MCP family of ADC devices

src/devices/Mcp3xxx/Mcp3Base.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+
+            // test the response from the ADC to check that the null bit is actually 0
+            if ((retval & (1 << adcResolutionBits)) != 0)
+            {
+                throw new InvalidOperationException("Invalid data was read from the sensor");
+            }
+
+            // return the ADC response with any possible higer bits masked out
+            return retval & (int)(0xFFFFFFFF >> (32 - adcResolutionBits));
```
krwq: I'm unclear if shifting `0xFFFFFFFF` will work as expected - AFAIR signed integers will preserve sign when shifting (so essentially first bit will stay 1) - might be worth validating this or starting with uint and then casting to int

src/devices/Mcp3xxx/Mcp3Base.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            _spiDevice?.Dispose();
+            _spiDevice = null;
+        }
+
+        /// <summary>
+        /// Reads a value from the device
+        /// </summary>
+        /// <param name="adcRequest">A bit pattern to be sent to the ADC.</param>
+        /// <param name="adcRequestLengthBytes">The number of bytes to be sent to the ADC containing the request and returning the response.</param>
```
krwq: can this be calculated with adcResolutionBits? i.e.: `((adcResolutionBits + 3) / 8)` or something along these lines

src/devices/Mcp3xxx/Mcp3Base.cs
krwq: assuming this is correct please make types more explicit (i.e. explicit cast) or going reverse way `((1 << adcResolutionBits) - 1)` (might be +/- 1 somewhere)

src/devices/Mcp3xxx/Mcp3xxx.cs
```diff
@@ -0,0 +1,158 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.Spi;
+
+namespace Iot.Device.Adc
+{
+    /// <summary>
+    /// Mcp3xxx Abstract class representing the MCP 10 and 12 bit ADC devices.
```
krwq: MCP family of ADC devices (I'd skip number of bits especially some of these are 13-bits from what I've seen)

src/devices/Mcp3xxx/Mcp3xxx.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// <param name="referenceChannel">Reference channel to be checked</param>
+        protected void CheckChannelPairing(int valueChannel, int referenceChannel)
+        {
+            CheckChannelRange(valueChannel, ChannelCount);
+            CheckChannelRange(referenceChannel, ChannelCount);
+
+            // Check that the channels are in the differential pairing.
+            // When using differential inputs then then the single ended inputs are grouped into channel pairs such that for an 8 input deviice then the pairs
+            // would be CH0 and CH1, CH2 and CH3, CH4 and CH5, CH6 and CH7 and thus to work out which channel pairing a channel is in then the channel number can be divided by 2.
+            if (valueChannel / 2 != referenceChannel / 2 || valueChannel == referenceChannel)
```
krwq: would that be equivalent to `Math.Abs(valueChannel  - referenceChannel) != 1`?

src/devices/Mcp3xxx/Mcp3xxx.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// <returns>A value corresponding to relative voltage level on specified device channels</returns>
+        public virtual int ReadDifferential(int valueChannel, int referenceChannel)
+        {
+            CheckChannelRange(valueChannel, ChannelCount);
+            CheckChannelRange(referenceChannel, ChannelCount);
+
+            if (valueChannel == referenceChannel)
+            {
+                throw new ArgumentException($"ADC differential channels must be different.", nameof(valueChannel) + " " + nameof(referenceChannel));
+            }
```
krwq: should there also be equivalent of:

```csharp
if (ChannelsSupportPseudoDifferential(valueChannel, referenceChannel))
{
    return ReadPseudoDifferential(valueChannel, referenceChannel);
}
```

src/devices/Mcp3xxx/Mcp3xxx.cs
```diff
@@ -0,0 +1,158 @@
...
+            // then incorporate the lower bit which indicates if the channel is inverted or not.
+            channelVal = inputType switch
+            {
+                InputType.Differential => channel * 2,
+                InputType.InvertedDifferential => channel * 2 + 1,
+                _ => channel
+            };
+
+            // create a value to represent the request to the ADC 
+            requestVal = ChannelCount switch
```
krwq: please use regular switch syntax, while I like functional programming this is inconsistent with the rest of the code base.

src/devices/Mcp3xxx/Mcp3201.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <param name="spiDevice">Device used for SPI communication</param>
+        public Mcp3201(SpiDevice spiDevice) : base(spiDevice) { }
+
+        /// <summary>
+        /// Reads a 12-bit (0..4096) value from the device
+        /// </summary>
+        /// <returns>12-bit value corresponding to relative voltage level on specified device channel</returns>
+        public int Read()
+        {
+            return ReadInternal(adcRequest: 0, adcRequestLengthBytes: 2, 13 + 1, delayBits: 0) >> 1;
```
krwq: please use named arg for `13 + 1` as well - also curious why this is 14 when reading 12-bit value

krwq: might be worth adding a comment

src/devices/Mcp3xxx/Mcp3001.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <param name="spiDevice">Device used for SPI communication</param>
+        public Mcp3001(SpiDevice spiDevice) : base(spiDevice) { }
+
+        /// <summary>
+        /// Reads a 10-bit (0..1023) value from the device
+        /// </summary>
+        /// <returns>10-bit value corresponding to relative voltage level on specified device channel</returns>
+        public int Read()
+        {
+            return ReadInternal(adcRequest: 0, adcRequestLengthBytes: 2, 10 + 3, delayBits: 0) >> 3;
```
krwq: explicit arg name for `10+3` and please add a comment why reading 13 bits for 10-bit ADC (unless the summary comment is wrong)

krwq:
@Frankenslag this mostly looks good to me. There are couple of typos and some minor things but in general ready to merge after fixing those.

cc: @shaggygi - would you also be interested in reviewing since you've also contributed to MCP3008?

src/devices/Mcp3xxx/Mcp3301.cs
Frankenslag: What I see here is that we are sign extending bit 13. What about if I just say that in the comment.

src/devices/Mcp3xxx/Mcp3304.cs
Frankenslag: With the 33 series if you are reading single ended it’s really only 12 bit positive whic I suppose is the same as 13 bit signed with the sign bit as 0. Is that what you were asking

src/devices/Mcp3xxx/Mcp3xxx.cs
Frankenslag: Nope because if we use The channels 3 and 4 then it will pass the test but that are in different pairings. Another way would be to compare the values with bit 0 masked out.

src/devices/Mcp3xxx/Mcp3001.cs
Frankenslag: The answer to this is that all the xxx1 devices don’t have a start bit being sent so the readout of the data is a certain number of bits after the /cs pin goes low. This means that unlike the xxx8, xxx4 and xxx2 devices where the b0 result is aligned with the b0 of the returned int the result data is potentially somewhere in the middle and has to be shifted. I have documented. The bit patterns in the base class.

src/devices/Mcp3xxx/Mcp3xxx.csproj
Frankenslag: If I remove the `<EnableDefaultItems>` then it tries to build the samples recursively as far as I can tell and thus includes the samples.

src/devices/Mcp3xxx/Mcp3201.cs
Frankenslag: Actually it should be 12 + 1 and is a bug

src/devices/Mcp3xxx/Mcp3xxx.cs
Frankenslag: I don't believe so as you wouldn't want to do a pseudo-differential read in place of a real one. With pseudo-differential on these chipsets the referenceValue is limited to +/- 100mv and so you probably wouldn't get the same answer.

src/devices/Mcp3xxx/Mcp3Base.cs
Frankenslag: I have made a change so that it looks for a start bit in bits 16-23 and if it is found then the buffer size is 3 bytes otherwise 2.

src/devices/Mcp3xxx/Mcp3Base.cs
Frankenslag: It actually works I think because 0xFFFFFFFF cannot be cast to an int and thus the compiler assumes a uint. perhaps to be clear I should use 0xFFFFFFFFU.

Anyway I do like your method but it also has an issue if you ask for 32 bits of resolution as it would return 0. i.e. 1 << 32 == 0. I think 1L << 32 == 0xFFFFFFFF so i'll use that

src/devices/Mcp3xxx/Mcp3301.cs
Frankenslag: O.K. I have created a static method for extending the sign... 

        /// <summary>
        /// Convert a signed value with a sign bit at a particular location to an int.
        /// </summary>
        /// <param name="signedValue">Signed value with a sign bit at a particular location</param>
        /// <param name="signingBit">Bit number that contains the sign bit</param>
        /// <returns>A value corresponding to the signed value sign extended into an int</returns>
        public static int SignExtend(int signedValue, int signingBit)
        {
            // if the sign bit is set then extend the signing bit to create a signed integer
            return (signedValue >> signingBit) == 0 ? signedValue : signedValue - (2 << signingBit);
        }

src/devices/Mcp3xxx/Mcp33xx.cs
Frankenslag: > @Frankenslag this mostly looks good to me. There are couple of typos and some minor things but in general ready to merge after fixing those.

I have made the changes but before updating the PR I want to do some more testing with hardware...

src/devices/Mcp3xxx/Mcp3304.cs
krwq: @Frankenslag not sure - I know Mcp3008 starts sending reading bits backward

Frankenslag: Note... for the 3301 I have changed the name of the function used to read the value to ReadDifferential (13 bit signed) as the 3301 does not have the ability to read single ended.

krwq:
@Frankenslag, I've seen you've marked couple of comments as resolved but have not seen any commits yet - presumably you're marking as resolved are you're working locally?

src/devices/Mcp3xxx/Mcp3001.cs
krwq: sounds good, just double checking this is not accidental

Frankenslag:
> @Frankenslag, I've seen you've marked couple of comments as resolved but have not seen any commits yet - presumably you're marking as resolved are you're working locally? (except the ones I resolved)

Yes that is the case. I don’t want to submit an updated pr until I have received two 3300 series’s chips to test out the changes. 

If you don’t want me to resolve until I submit a PR then happy to oblige

krwq:
@Frankenslag that's fine. Let's do this, push whatever you got, I'll double check Mcp3008 is not regressed, add a note saying "Mcp33xx" are not tested in the README.md, once you receive them you can test them and fix any issues you might find or just remove the note.

(normally I'd ask you to wait but we wanted to create 3.0 branch after merging your PR)

Frankenslag:
> @Frankenslag that's fine. Let's do this, push whatever you got, add a note saying "Mcp33xx" are not tested in the README.md, I'll double check Mcp3008 is not regressed, once you receive them you can test them and fix any issues you might find or just remove the note.
> (normally I'd ask you to wait but we wanted to create 3.0 branch after merging your PR)

@krwq
I have pushed the latest PR updates. I did a final last test of running the Mcp3008.Sample under hardware SPI. I didn't have time for testing under Software SPI.

krwq:
@Frankenslag if hardware SPI works that's fine - you're not changing software implementation so I'm good with that.

Frankenslag:
> @Frankenslag if hardware SPI works that's fine - you're not changing software implementation so I'm good with that.

Now tested Mcp3008 with software SPI.

PR #703, opened 2019-09-07
Rename DutyCyclePercentage -> DutyCycle

krwq:
Fixes: https://github.com/dotnet/iot/projects/4#card-25528890

`Percentage` suffix means that value will be multiplied by 0.01 (since `%=0.01`) which suggest that value is in range 0-100 which is not correct. Removing the suffix for correctness.


joperezr:
We checked for precedence in the framework and found that all APIs in .NET Core that use "Percentage" in the name were ints that go from 0-100, which is not the case for ours.

PR #704, opened 2019-09-08
Support for .netstandard2.0

hugener:
Please see the pull request as a discussion on .netstandard2.0 support. In its current state it cannot be merged and I didn't have any means for testing the changes. However, with those changes I am able to build for .netstandard2.0.
- Changed TargetFramework to netstandard2.0
- Added package references to: System.Buffers, System.Memory, System.Numerics.Vectors, System.Runtime.InteropServices for missing types.
- Added MathEx with Clamp method as fill in for Math.Clamp, which is not available. (Copied from here: https://github.com/dotnet/corefx/blob/d58a51f911efb3c98beca21b6cf08cc703424fdf/src/Common/src/CoreLib/System/Math.cs). Might have to rethink how to do that.
- Replaced MemoryMarshal.CreateReadOnlySpan and MemoryMarshal.CreateSpan with ctor call in unsafe block.
- Replaced ToImmutableDictionary with new ReadOnlyDictionary.
- Used array overloads of BitConverter.ToSingle since the data is already in an array and Span overloads are not available.
- Used WriteByte instead of Write (Span), since it's not available.
- Used AsSpan rather that implicit cast where necessary.


src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
```diff
@@ -99,13 +99,9 @@ private void WriteColor(Color color)
...
             byte b = (byte)(color.B >> 3);
             ushort col = (ushort)((r << 11) | (g << 5) | b);
-            Span<byte> encoded = stackalloc byte[2]
-            {
-                (byte)(col & 0xff),
-                (byte)(col >> 8)
-            };
 
-            _deviceFile.Write(encoded);
+            _deviceFile.WriteByte((byte)(col & 0xff));
```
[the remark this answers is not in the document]
krwq: @hugener I'm fine with not testing this particular fix since this is low risk:
- sysfs is most likely not affected by this
- this is a backup implementation for LED Matrix - I2C is used by default

hugener:
I merged the lastest dotnet/master into my fork to resolve merge conflicts and also added multi targetting for netstandard2.0 and netstandard2.1.
The upside to this is that the netstandard2.1 version can get pretty identical to what it was before.
The downside is that #ifs are required to provide fallbacks for 2.0.
Let me know how you feel about the multitargetting and the use of #ifs?

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
```
krwq: I think 2.1 is implied by 2.0. cc: @joperezr

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -20,8 +20,22 @@
     <Compile Include="$(DeviceRoot)**/*.cs" Exclude="@(_ExcludeFromCompile)" />
   </ItemGroup>
 
+  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard2.0'))">
```
krwq: `'$TargetFramework' == 'netstandard2.0'` (same below)

src/Iot.Device.Bindings/MathExtension.cs
```diff
@@ -0,0 +1,126 @@
+﻿
```
krwq: license header missing

src/Iot.Device.Bindings/MathExtension.cs
```diff
@@ -0,0 +1,129 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+
+namespace System
+{
+    public static class MathExtension
```
joperezr: Definitely don't love the fact of copying code from corefx. How many bindings depend on this? Would it be feasable to just change the implementation of those bindings (assuming there aren't many of them)?

hugener: There are 21 usages spread across 6 bindings (12 classes).
I took another stab at it, renaming MathExtension to MathHelper and limiting the #ifs to inside MathHelper. This significantly reduces the amount of #ifs to one. Bindings would have to use MathHelper rather than Math (with the using alias). That way I could also remove the forwarders for Math.Round and Abs.

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -20,6 +20,17 @@
     <Compile Include="$(DeviceRoot)**/*.cs" Exclude="@(_ExcludeFromCompile)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="MathHelper.cs" />
+  </ItemGroup>
+
+  <ItemGroup Condition="$(TargetFramework) == 'netstandard2.0'">
```
krwq: I think you will need to exclude CreditCard stuff for netstandard2.0 build as well which I'm not sure if it's ok

src/devices/BrickPi3/BrickPi3.csproj
```diff
@@ -13,6 +13,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Compile Include="..\..\Iot.Device.Bindings\MathHelper.cs" Link="MathHelper.cs" />
```
joperezr: Usually files like these would go under src/devices/Common instead of IoT.Device.Bindings folder.

src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
```diff
@@ -108,13 +108,9 @@ private void WriteColor(Color color)
...
+            _deviceFile.WriteByte((byte)(col >> 8));
```
joperezr: If possible, I would rather only calling Write once, as we used to.

src/devices/SocketCan/CanRaw.cs
```diff
@@ -74,13 +74,16 @@ public bool TryReadFrame(Span<byte> data, out int frameLength, out CanId id)
             }
 
             CanFrame frame = new CanFrame();
-            
-            Span<CanFrame> frameSpan = MemoryMarshal.CreateSpan(ref frame, 1);
-            Span<byte> buff = MemoryMarshal.AsBytes(frameSpan);
-            while (buff.Length > 0)
+
+            unsafe
```
joperezr: Is MemoryMarshal not available in netstandard2.0?

src/Iot.Device.Bindings/Iot.Device.Bindings/Iot.Device.Bindings.sln
```diff
@@ -0,0 +1,101 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
joperezr: We tipically don't checkin solutions into the repo with the only exception of the main library.

joperezr:
Honestly now that I look at this I'm a little bit hesitant to let this PR go through. I would have been ok if this was just a straight change of the targetFramework + add some additional package references, but modifying the bindings source and copying code from corefx of missing members in netstandard 2.0 makes me a bit nervous. Mainly because most of the devices in this package we can't really test because we don't own them, and we've relied on people adding the bindings to test them, and now we are changing their code and hoping that it would all work without testing on the hardware. I would be ok to change this to netstandard2.1 and hopefully Mono will eventually support that and would give us some reach on other devices. What do you think @krwq

src/devices/BrickPi3/BrickPi3.csproj
hugener: I moved it there

src/devices/SocketCan/CanRaw.cs
hugener: MemoryMarshal is, but not the CreateReadOnlySpan and CreateSpan

src/Iot.Device.Bindings/Iot.Device.Bindings/Iot.Device.Bindings.sln
hugener: It was a mistake, I removed it again

src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
hugener: It's possible yes, but we would have to allocate a byte[]
I assumed WriteByte would write to the FileStreams buffer, since Flush is always called afterwards.

hugener:
I understand your concern. Would it be an option to do a limited netstandard2.0, that do support those binding that cannot be implemented in a clean way on netstandard2.0?

src/devices/SenseHat/SenseHatLedMatrixSysFs.cs
krwq: we can allocate ahead of time though (i.e. in the constructor) and then only replace the bytes

krwq:
@hugener would netstandard2.1 work for you? I'm slightly curious - is there any particular reason why upgrading your project to 2.1 could be a problem?

hugener:
netstandard2.1 would be fine for me too, although I think if possible we should target the wide support possible and since System.Devices.Gpio is netstandard2.0, I think it would have made sense for Iot.Devices.Bindings too.
I wanted to exhaust the possibilities before I move my code to netstandard2.1 as I need the following devices:
 - Max9744 amplifier (I2c - binding in the works)
 - Ky040 Rotaryencoder (Gpio - binding in the works)
 - Mfrc522 Rfid transceiver (SPI - not sure what do do here)
 - RemotePI power management/Infrared receiver (GPIO and LIRC - not sure what to do about LIRC)
 - CharacterLCD - Already supported and no problems for netstandard2.0
 - Buttons - Already support on netstandard2.0 (System.Devices.Gpio)
My goal was to implement bindings for those that do not yet exist and eventually move to System.Devices.Gpio. Max9744, Ky040 and RemotePI should be fairly easy depending on what I do with LIRC.
Mfrc522 is harder, as the code I based mine on, is very buggy/incomplete but for me it's not a problem because I only need to read IDs (ID reading is incomplete, but good enough for me). So I wouldn't expect that to be accepted here.

As another attempt my latest commit reverts the "unclean" implementations for netstandard2.0 support and removes those bindings for netstandard2.0. That means netstandard2.0 is still supports a lot of devices and there are no hacks to support the other devices.

I could of course also consider to maintain my own CharacterLCD netstandard2.0 version, along with the bindings I plan to add, but I thought it was worth a shut to add support through here.

joperezr:
It is possible and would probably make sense to have Iot.Device.Bindings package to provide two different assets: one for netstandard2.0 and one for netsatndard2.1. In the 2.0 one we will add all of the bindings that would build just fine without the need of if/deffing or new API ported from corefx, and then we could have the 2.1 one which would have all the bindings. That way, if somebody wants to add a reference to this package, they could decide to either A) Run on more platforms but with less bindings available, or B) run in less platforms (since 2.1 is not supported widely yet) but have more bindings available. This way, when a new binding is submitted, can try not limiting it to just netstandard2.0 surface area.

Would that sound reasonable?

hugener:
@joperezr 
This is already implemented on my branch, the check the Iot.Device.Bindings.csproj.
It would be nice to devise a better way of excluding non-supported devices for netstandard2.0 though.

MarkCiliaVincenti:
> Honestly now that I look at this I'm a little bit hesitant to let this PR go through. I would have been ok if this was just a straight change of the targetFramework + add some additional package references, but modifying the bindings source and copying code from corefx of missing members in netstandard 2.0 makes me a bit nervous. Mainly because most of the devices in this package we can't really test because we don't own them, and we've relied on people adding the bindings to test them, and now we are changing their code and hoping that it would all work without testing on the hardware. I would be ok to change this to netstandard2.1 and hopefully Mono will eventually support that and would give us some reach on other devices. What do you think @krwq

I think Microsoft should afford to buy you these devices so that you could run some tests ;)

joperezr:
> Given that this is a project that has gone stable and is basically being tested by hobbyists, the least we could do is implement more Unit tests.

Totally agree, @krwq is working on that with this PR #758, the problem is that I believe we will likely never be on a position where we will have all of the devices we have device bindings for in our test lab, so we could still add some unit test coverage that won't run in the actual hardware but there will always be that small gap for some device bindings.

scout208:
Hey, was just wondering what the status is on this pull request? Are you just trying to implement more tests before you move it in? I have a UWP app on Windows 10 IoT Core and currently it doesn't support referencing .NET Core projects but does support .NET Standard so it would be great to be able to use this for UWP apps.

joperezr:
@scout208 we are still discussing how to move forward with this but doesn't look that it will be defined any time soon. As a workaround, one thing you could do is to reference System.Device.Gpio package which does support .NET Standard 2.0, and then copy paste the specific binding code that you need for your project and hopefully that one you need won't require .NET Core 2.1. Can I get more info on your project? If it is not targeting .NET Core, is it Mono? Where do you plan to run?

scout208:
@joperezr I will try that out. The only device binding I need is for the MCP2515. It's not Mono, it's Universal Windows Platform app and it's running on Windows 10 IoT Core on a Raspberry Pi 3 Model B. Although I'm thinking of switching to an Intel board soon to have access to more drivers.

krwq:
@hugener this has been open for a very long time. Seems the stuff unrelated to re-targetting is already pulled out of here and this seems like we prefer code clarity and moving forward with the features in the repo rather than having netstandard 2.0. IMO this PR should be closed and any conversations about re-targetting should be moved to separate issue if needed.

hugener:
Closed per request

PR #705, opened 2019-09-08
Add RotaryEncoder Binding

Frankenslag:
This PR is to add a binding for an incremental rotary encoder.

I notice that there is a branch for a different version of this but it seems not to be being pursued.


krwq:
@Frankenslag what do you mean by:

> I notice that there is a branch for a different version of this but it seems not to be being pursued.

src/devices/RotaryEncoder/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+From above if we look at Pin B (data) at the time of a falling edge on Pin A (clk) then the if the value of pin P is 1 then the direction is clockwise and if it is 0 then the rotation is counter clockwise.
+
+## Binding Notes
+
+This binding implements a rotary encoder as a generic class `RotaryEncoder<T>`. As it is a generic class then the type T can be whatever is appropiate for the task as long as it is nummeric. For example it could be set up as an integer that has a range of 0..100 and each step incrments or decrements the value by 1. It coould also be set up as a tuning dial for an FM radio that is a decimal with a range of 88.0 to 108.0 with a step of 0.1. It is probably best not to use a float or double value as the lack of absolute mathematical precision can cause issues.
+
+The code below shows an example of using the encoder as an FM tuner control.
+
+    using(GpioController controller = new GpioController())
```
krwq: space after `using`

Frankenslag:
> @Frankenslag what do you mean by:
> 
> > I notice that there is a branch for a different version of this but it seems not to be being pursued.

There is a closed stale branch called rotary encoder and it looks like rich lander was trying to put a binding together

src/devices/RotaryEncoder/RotaryEncoder.cs
```diff
@@ -0,0 +1,190 @@
...
+        {
+            Value = value;
+        }
+    }
+
+    /// <summary>
+    /// Bindinding that exposes a 2 wire incremental rotary encode as a Value that varies as the encoder is turned.
+    /// </summary>
+    /// <typeparam name="T">The element type of the Value controlled by the RotaryEncoder</typeparam>
+    public class RotaryEncoder<T> where T : IComparable<T>
```
krwq: can this just use double instead?

src/devices/RotaryEncoder/RotaryEncoder.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// <summary>
+        /// Initialize a RotaryEncoder
+        /// </summary>
+        /// <param name="controller">GpioController that hosts Pins A and B.</param>
+        /// <param name="pinA">Pin A that is connected to the rotary encoder. Sometimes called clk</param>
+        /// <param name="pinB">Pin B that is connected to the rotary encoder. Sometimes called data</param>
+        /// <param name="pulseIncrement">The amount that the value increases or decreases on each pulse from the rotary encoder</param>
+        /// <param name="rangeMin">Minimum value permitted. The value is clamped to this.</param>
+        /// <param name="rangeMax">Maximum value permitted. The value is clamped to this.</param>
+        private void Initialize(GpioController controller, int pinA, int pinB, T pulseIncrement, T rangeMin, T rangeMax)
```
krwq: would it makes sense to instead have `number of pulses per rotation` then have properties:

- PulseCountPerRotation
- PulseCount
- NumberOfRotations
- NumberOfRotationsPerSecond (equivalent to current acceleration)

all rotary encoders I find on Amazon display `pulses/rotation` so it would be nice if you could just type it in the constructor

Frankenslag: The design was to allow the encoder to provide values that reflected the use to which it was being put. I have used the FM radio scale in the demo 88.0 to 108.0 in 0.1 Mhz steps. I believe that building the scaling and the maths into the control makes sense. Another application could be setting the time for a digital clock with perhaps a with a TimeSpan as T and overriding the OnPulse to add or subtract a certain number of seconds.

I am a bit confused why you are suggesting properties related to rotation as most uses of its use case wouldn't care how far in angular terms it has been turned (unless it's a shaft encoder) . 

I don't think double is a good choice as add 0.1D to 0 three times gives me 0.300000000000004 and this could cause issues and this is why I chose to implement this as a generic as you could use whatever type was appropriate. You are right that the users could do the maths themselves but I am trying not to constrain them to types

Also the properties PulseCount NumberOfRotations do not make any sense to me as they would have to be relative to something would it be the last time the property was read or something else.

Perhaps we have a different view about where these encoders might me used.

krwq: @Frankenslag I don't disagree that builtin math is useful there but my point is that before you add something more advanced you should first provide the basic functionality which is what the basic parameters are.

Let's say I buy rotary encoder now - the only parameter I can find on the website is P/R (pulses per rotation) - this is what I'd expect to pass in to the constructor.

> Why have anything related with rotations?
One of the usages of rotary encoder is to connect it with motor to see how fast it is rotating.
Other reason is it's really easy to think about - you rotate once, you expect 1 to be somewhere in the properties.

> Also the properties PulseCount NumberOfRotations do not make any sense to me as they would have to be relative to something would it be the last time the property was read or something else.
You can ask identical question to your case - what would be the initial number? I'd say 0 is probably expected, optionally you could provide your own number or have a way to reset it

> I don't think double is a good choice as add 0.1D to 0 three times gives me 0.300000000000004

You have identical problem with your current implementation, also if you calculate in terms of pulses/rotations and have pulsesPerRotations then (untested):

```csharp
value = ((double)pulses / pulsesPerRotation) * (maxRange - minRange) + minRange;
```

will add as accurately as it is possible and increment by 0.1 doesn't exist since you calculate on integers (long or int)

My point is to have at minimum `Pulses` and `PulsesPerRotation` exposed, anything else can be calculated from that. If you expose only something like `Value` then it actually makes it harder to go the other way

krwq:
I haven't personally used them yet but I'm wondering if some really small capacitor acting as low pass filter somewhere wouldn't eliminate the problem (it does fix similar issue with regular buttons and many other things).

Frankenslag:
@hugener @krwq
I agree that we should rename and I'm happy to accept suggestions. The device I am working with is supposed to be an EC11 according to eBay and is incremental very much like the ky040 you mention. I am a bit reluctant to call it an EC11 though as there seems to be a large number of variations.

Although I haven't read your source code in detail it does seem a lot more complex than mine and I'm not sure why.

Regarding false detections. I used the PinChanged event from the Gpio controller. I believe that this is polling behind the scenes but it seemed to be much better than anything I put together to detect changes in signal and I found really easy to use. I still got a good deal of bounce and what was puzzling was that I would often get falling edges without corresponding rising ones.  The debounce functionality really helped with that. It is fairly simple and works by only accepting the first event and then will ignore any subsequent ones within the debounce period. To calibrate this I turned the encoder as fast as I could and measured the pulse width and set the debounce interval to half of that.

I should also mention that I only look for falling edges of the clock (pinA) right now. This is because my encoder has detents and you always tend to get a rising and falling edge for each detent as you rotate. I can see that some encoders might not have a detent in this way and I should add the option to detect on both edges as it would increase the resolution of the encoder.

src/devices/RotaryEncoder/RotaryEncoder.cs
Frankenslag: @krwq
I take your point with rotation and pulses and rotations as properties. Perhaps we should have a base class that works that way and create a subclass to deal with values and ranges. What do you think ?

Frankenslag: @krwq @hugener
I have created a base class called `QuadratureRotaryEncoder` which I hope delivers the rotation properties. I have then created a subclass for a `ScaledQuadratureEncoder` to replace the previous `RotaryEncoder`. I hope that this naming describes things a bit better.

I have also added the ability to define which Gpio events the encoder is listening for.

I have updated the PR to reflect this.

Frankenslag:
@krwq 
I see that the check for rotaryencoder have failed. Is there anyway to rerun ?

krwq:
@Frankenslag just retried

src/devices/RotaryEncoder/QuadratureRotaryEncoder.cs
```diff
@@ -0,0 +1,126 @@
...
+            Initialize(controller, pinA, pinB, edges);
+        }
+
+        /// <summary>
+        /// QuadratureRotaryEncoder constructor
+        /// </summary>
+        /// <param name="pinA">Pin A that is connected to the rotary encoder. Sometimes called clk</param>
+        /// <param name="pinB">Pin B that is connected to the rotary encoder. Sometimes called data</param>
+        /// <param name="pulsesPerRotation">The number of pulses to be received for every full rotation of the encoder.</param>
+        public QuadratureRotaryEncoder(int pinA, int pinB, int pulsesPerRotation) : this(new GpioController(), pinA, pinB, PinEventTypes.Falling, pulsesPerRotation) { }
```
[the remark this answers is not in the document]
Frankenslag: @hugener 
I have made this change to my local branch. As this is a minor change will wait before updating the PR.

krwq:
@Frankenslag ready for re-review or are there some pending changes?

Frankenslag:
> @Frankenslag ready for re-review or are there some pending changes?

@krwq 

Happy for a review. There were quite substantial changes.. I haven't made the minor change from @hugener in the remote branch yet as I am waiting for more comments and will bundle the updates.

hugener:
> @hugener @krwq
> I agree that we should rename and I'm happy to accept suggestions. The device I am working with is supposed to be an EC11 according to eBay and is incremental very much like the ky040 you mention. I am a bit reluctant to call it an EC11 though as there seems to be a large number of variations.

I'm sceptic about the name because the implementation does not actually capture the 4 phases between detents. You would have to listen to both rising and falling edges and also use the previous state. There is a pretty god state table here: https://en.wikipedia.org/wiki/Incremental_encoder#Quadrature_outputs
My implementation tries to use AB State, a CE of x1 and DIR. I don't report ERR states, but use it to reset intermediate states, so the encoder doesn't get hung up (Not sure if that's correct though).
The AB states are also what I refer to as Gray codes. Notice how only one bit changes at the time in the valid states.

> Although I haven't read your source code in detail it does seem a lot more complex than mine and I'm not sure why.

I guess it is because I use Gray codes.

> Regarding false detections. I used the PinChanged event from the Gpio controller. I believe that this is polling behind the scenes but it seemed to be much better than anything I put together to detect changes in signal and I found really easy to use. I still got a good deal of bounce and what was puzzling was that I would often get falling edges without corresponding rising ones.  The debounce functionality really helped with that. It is fairly simple and works by only accepting the first event and then will ignore any subsequent ones within the debounce period. To calibrate this I turned the encoder as fast as I could and measured the pulse width and set the debounce interval to half of that.

Do you know what the polling rate is?
I might also give debouncing a try in my implementation. I know the GPIO lib I'm using is polling in C#, which is probably worse that this (GC, context switches, etc.). I remember seeing libgpiod being used here, but I don't know the implementation details.

> I should also mention that I only look for falling edges of the clock (pinA) right now. This is because my encoder has detents and you always tend to get a rising and falling edge for each detent as you rotate. I can see that some encoders might not have a detent in this way and I should add the option to detect on both edges as it would increase the resolution of the encoder.

Maybe instead of inheriting from the QuadratureRotaryEncoder in ScaledQuadratureEncoder, we should agree on an interface. That way your scaled encoder could be used with different implementation.

I would also suggest to keep QuadratureRotaryEncoder purely directional, meaning not counting Pulses and Rotation. Those could overflow (I know they are big, but still) and how that should be handled, would be up to the user. The direction is also everything your scaled encoder depends on, so it shouldn't be a problem. That would also make our two encoders potentially interface compatible.

krwq:
I think it makes sense that you figure out the interface (or even better: abstract class) since you both have internals on how this device works. Perhaps we should open an issue and propose API shape and after that modify this PR accordingly?

Other option is we merge this after review and then @hugener sends another PR with incremental changes fixing/improving specific things.

Or some hybrid of those two perhaps: more or less shape you are both mostly happy with and then incremental fixes for specific issues.

maloo:
Have you considered a binding of kernel rotary encoder drivers?
https://www.kernel.org/doc/html/v4.16/input/devices/rotary-encoder.html
https://blog.ploetzli.ch/2018/ky-040-rotary-encoder-linux-raspberry-pi/

krwq:
@maloo fine to have both implementation since as is will work on Windows as well

krwq:
@Frankenslag got time to finish this up? What's left here?

Frankenslag:
> @Frankenslag got time to finish this up? What's left here?

Will try and revisit soon.

nahueltaibo:
@Frankenslag If it helps to motivate you, I'm waiting for the QuadratureRotaryEncoder to use it in my project 😃

HakanL:
What's the status on this? How can we help?

Frankenslag:
> What's the status on this? How can we help?

I have got a bit stuck really. What I found was that the lack of 'true' interrupts and meant that there were timing issues. It works most of the time but you would then get glitches because of missed pulses. I am coming to the conclusion that it would be better for designs to implement a microcontroller based solution for reading the encoder pulses and then reading the value from that.

nahueltaibo:
> I have got a bit stuck really. What I found was that the lack of 'true' interrupts and meant that there were timing issues. It works most of the time but you would then get glitches because of missed pulses. I am coming to the conclusion that it would be better for designs to implement a microcontroller based solution for reading the encoder pulses and then reading the value from that.

I think the same. I am planning to usa an arduino nano, or something like that for reading real time inputs. Didnt get to that part on my projects yet though

HakanL:
I just saw that there was a PR to improve interrupt speed, maybe it's worth a shot to retry it with latest `master`?

krwq:
@Frankenslag ping, if we can declare this PR abandoned please let us know so someone else can pick it up if needed

Frankenslag:
> @Frankenslag ping, if we can declare this PR abandoned please let us know so someone else can pick it up if needed

Yes, I don't believe that reading the encoders pins using gpio directly is the way to go. to I am happy for someone else to try.

Ellerbach:
> Yes, I don't believe that reading the encoders pins using gpio directly is the way to go

Agree for high speed and precision application. Tests shows that whatever OS, because no interrupt are used, some pulse are missed. Now, for manual encoders like simple gears, this is working perfectly! So if you just add a simple rotary button, that works just perfectly. Those buttons are used to select menus and things like this on small screens or increase/decrease something. 
So, regarding the work done already, I would keep it with a big mention that this implementation is not for real time application but rather for small speed (will be hard to define) or manual rotary.
Also, it seems that the native Linux driver is a bit better but still not real time, so same issue. That would just work better but still not the perfect way.
So we can for example add an example with an Arduino connected in SPI with some Arduino code gathering the real time data and sending it thru SPI to .NET Core IoT with a simple Fritzing schema. Still, that would not solve real time control but that would solve the precision.
@krwq  and @Frankenslag, what do you think?

Frankenslag:
> > Yes, I don't believe that reading the encoders pins using gpio directly is the way to go
> 
> Agree for high speed and precision application. Tests shows that whatever OS, because no interrupt are used, some pulse are missed. Now, for manual encoders like simple gears, this is working perfectly! So if you just add a simple rotary button, that works just perfectly. Those buttons are used to select menus and things like this on small screens or increase/decrease something.
> So, regarding the work done already, I would keep it with a big mention that this implementation is not for real time application but rather for small speed (will be hard to define) or manual rotary.
> Also, it seems that the native Linux driver is a bit better but still not real time, so same issue. That would just work better but still not the perfect way.
> So we can for example add an example with an Arduino connected in SPI with some Arduino code gathering the real time data and sending it thru SPI to .NET Core IoT with a simple Fritzing schema. Still, that would not solve real time control but that would solve the precision.
> @krwq and @Frankenslag, what do you think?

@Ellerbach I agree with this approach. I haven't done any Arduino work for some time though. Having said that perhaps an Arduino is a sledgehammer cracking a nut and perhaps an example using minimal microcontroller implementation would be good also.

Ellerbach:
> I haven't done any Arduino work for some time though

I was interested in this rotary implementation as my youngest son is doing autonomous car racing (the small cars, not human sized). He wanted to have an idea of the speed of the car, so we added a hall effect sensor which outputs A and B signals. The real time management of the car is in an Arduino, so I have some Arduino code already. Either I can reuse, either we can write anything. Core idea is to show something that is working if real time and high precision is needed.

So we can most likely adjust the documentation of this PR first with the disclaimer and the usage for "slow" or "low precision" usage and then add later the native Linux driver and the Arduino part.

HakanL:
Sparkfun has a good implementation of this, which is open-source using an TINY84A: https://www.sparkfun.com/products/15083

krwq:
@Ellerbach might be worth to try this on nano framework in a while and see if we can get it to reliably work, if not perhaps we should add a doc explaining unsupported scenarios and add this as an example

Ellerbach:
> @Ellerbach might be worth to try this on nano framework in a while and see if we can get it to reliably work, if not perhaps we should add a doc explaining unsupported scenarios and add this as an example

On my list! Not for right away but on the list :-)

joperezr:
[Triage] We are ready to take this PR. The only thing that should be done before we do is to make sure we add the right documentation to mention that there are some limitations with its use(non-real time scenarios). @Frankenslag are you interested in finishing this?

joperezr:
[Triage] We'll take the work item of explaining the limitation by documenting this.

src/devices/RotaryEncoder/QuadratureRotaryEncoder.cs
```diff
@@ -35,12 +35,20 @@ public class QuadratureRotaryEncoder
         public float Rotations { get => (float)PulseCount / PulsesPerRotation; }
 
         /// <summary>The DebounceMilliseconds property represents the minimum amount of delay allowed between falling edges of the A (clk) pin.</summary>
-        public uint DebounceMilliseconds { get; set; } = 5;
+        public TimeSpan DebounceMilliseconds
```
krwq, line 38: DebounceTime since unit is part of the type now (make sure to update the doc to mention recommended unit)

src/devices/RotaryEncoder/QuadratureRotaryEncoder.cs
```diff
@@ -50,13 +58,15 @@ public class QuadratureRotaryEncoder
         /// <param name="pinB">Pin B that is connected to the rotary encoder. Sometimes called data</param>
         /// <param name="edges">The pin event types to 'listen' for.</param>
         /// <param name="pulsesPerRotation">The number of pulses to be received for every full rotation of the encoder.</param>
-        public QuadratureRotaryEncoder(GpioController controller, int pinA, int pinB, PinEventTypes edges, int pulsesPerRotation)
+        /// <param name="shouldDispose">True to dispose the controller</param>
+        public QuadratureRotaryEncoder(GpioController? controller, int pinA, int pinB, PinEventTypes edges, int pulsesPerRotation, bool shouldDispose = true)
```
krwq, line 62: in most of the bindings gpioController is after required args and it defaults to null so that you are not required to provide it if not needed.

src/devices/RotaryEncoder/QuadratureRotaryEncoder.cs
```diff
@@ -65,7 +75,10 @@ public QuadratureRotaryEncoder(GpioController controller, int pinA, int pinB, Pi
         /// <param name="pinA">Pin A that is connected to the rotary encoder. Sometimes called clk</param>
         /// <param name="pinB">Pin B that is connected to the rotary encoder. Sometimes called data</param>
         /// <param name="pulsesPerRotation">The number of pulses to be received for every full rotation of the encoder.</param>
-        public QuadratureRotaryEncoder(int pinA, int pinB, int pulsesPerRotation) : this(new GpioController(), pinA, pinB, PinEventTypes.Falling, pulsesPerRotation) { }
+        public QuadratureRotaryEncoder(int pinA, int pinB, int pulsesPerRotation)
```
krwq, line 78: I'd recommend adding default args to other constructor and change order of args and remove this constructor completely

src/devices/RotaryEncoder/README.md
```diff
@@ -1,39 +1,50 @@
...
-        };
+    encoder.ValueChanged += (o, e) =>
+    {
+        Console.WriteLine($"Value {e.Value}");
+    };
 
-        // Do Other Stuff
-    }
+    // Do Other Stuff
+}
```
krwq: I think this event based API is a pitfall for real scenario since ValueChanged lambda will execute in different thread so you may encounter threading issues if you use same i.e. I2C device in both threads

krwq: (I'm fine with creating the issue to think of redesigning this API but might be good to also document this might be an issue at least in the comment)

krwq (approved):
LGTM after renaming DebounceMilliseconds to DebounceTime. Remainder treat as non-blocking suggestions which we can solve in separate PR

src/devices/RotaryEncoder/QuadratureRotaryEncoder.cs
Ellerbach: fixing

src/devices/RotaryEncoder/QuadratureRotaryEncoder.cs
Ellerbach: fixing as well

src/devices/RotaryEncoder/README.md
Ellerbach: Yes, now, we wrote that it's for slow usage, not like a motor

joperezr:
@Ellerbach should be ready to go as soon as the build is fixed.

Ellerbach:
> @Ellerbach should be ready to go as soon as the build is fixed.

hum, that should be straight forward. Don't even know how I managed to miss this one!

joperezr (approved):

PR #723, opened 2019-09-16
Add Max9744 binding

hugener:
The Max 9744 is a 2x20W stereo amplifier that connects using i2c for volume control and two gpio pins for mute and shutdown.
https://www.maximintegrated.com/en/products/analog/audio/MAX9744.html/tb_tab0


joperezr:
Instead of making the TFM change here, can we keep IoT.Device.Bindings package to be netcoreapp2.1 for now and just scope the nestandard2.0 change to the new binding you are adding? We have that other PR where we are still designing the solution for multi-targeting, so lets not base this change on that one just yet until we come up with the best solution for that.

krwq:
@hugener do you think you'll have time to remove non Max9744 related stuff from this PR so that we can merge?

joperezr:
[Triage]We are closing this PR as it seems abandoned for now. If anybody is interested in picking it up where @hugener left it off, feel free to let us know so we can reopen this.

PR #728, opened 2019-09-19
Fix README links

garciaolais:
- wrong slash  **xxx\\README.md**
- Change **Mcp3008** -> **Mcp3xxx**


joperezr:
Thanks for fixing this @garciaolais! great catch. This happened because the last time we updated these lists we did it running the tool on Windows, which produced backslashes instead of forward slashes. I will open an issue to fix the tool so that it produces the same output on Windows and Linux and mark it up-for-grabs in case you are interested in fixing the root cause.

PR #732, opened 2019-09-20
Do not generate backslashes with device-listing on Windows

### LINKED ISSUE #731: [Tools] Fix tool that generates the device lists to output forward slashes instead of backslash
The tool we use for generating the device listings produces the wrong output when running it on a Windows machine, since it uses forward slashes instead of backslashes. Related #728 #729 

FYI: @garciaolais @krwq

garciaolais:
Fix [#731 ](https://github.com/dotnet/iot/issues/731)

- Now the **README** generation works in Windows and Linux 
- Add category file in **BH1745** and **CARD**


tools/device-listing/device-listing.cs
```diff
@@ -47,6 +47,8 @@ class Program
             "color",
             "led",
             "spi",
+            "nfc",
```
krwq: fine to have both categories but since they currently have exactly the same bindings just display one of them

tools/device-listing/device-listing.cs
```diff
@@ -210,13 +214,11 @@ private static string GetRelativePathSimple(string path, string parentPath)
...
             if (path.StartsWith(parentPath))
             {
-                int i = parentPath.Length;
-                if (path[i] == '/' || path[i] == '\\')
-                {
-                    i++;
-                }
+                string fileName = Path.GetFileName(path);
+                string dirName = new DirectoryInfo(path).Parent.Name;                    
+                UriBuilder uriBuilder = new UriBuilder(){Path = Path.Combine(dirName, fileName)};
```
krwq: I'd leave this logic as it was since it worked correctly according to the name but instead change the logic for generating links (inside of `GetDeviceListing`) since that is actually incorrect (I'd just to `Replace("\\", "/")` but Uri is fine too - you might want to add utility like `CreateMarkdownLinkFromPath` or similar)

krwq: also nit: space after and before `{` and `}`

garciaolais: I thought about using `Replace("\\", "/")`, but I needed to add logic to know which operating system is running.

tools/device-listing/device-listing.cs
```diff
@@ -47,6 +47,7 @@ class Program
             "color",
             "led",
             "spi",
+            "nfc",
```
joperezr: NIT: can you remove the extra spaces after `"nfc",`

tools/device-listing/device-listing.cs
```diff
@@ -204,19 +206,16 @@ private static string FindRepoRoot(string dir)
...
             if (path.StartsWith(parentPath))
             {
-                int i = parentPath.Length;
-                if (path[i] == '/' || path[i] == '\\')
-                {
-                    i++;
-                }
+                string fileName = Path.GetFileName(path);
+                string dirName = new DirectoryInfo(path).Parent.Name;                    
+                UriBuilder uriBuilder = new UriBuilder(){ Path = Path.Combine(dirName, fileName) };
```
krwq:
```suggestion
                UriBuilder uriBuilder = new UriBuilder() { Path = Path.Combine(dirName, fileName) };
```

tools/device-listing/device-listing.cs
```diff
@@ -91,6 +92,7 @@ class Program
             { "buzzer", null },
             { "gopigo3", null },
             { "grovepi", null },
+            { "nfc", "RFID/NFC modules" },
```
krwq:
```suggestion
            { "nfc", "RFID/NFC modules" },
```

krwq:
LGTM after the last suggestion but I wanted to run the tool on linux against this changeset and double check there is no diff

krwq:
```diff
diff --git a/src/devices/README.md b/src/devices/README.md
index 3af9b4c..617490b 100644
--- a/src/devices/README.md
+++ b/src/devices/README.md
@@ -196,7 +196,7 @@ Our vision: the majority of .NET bindings are written completely in .NET languag

 * [Software implementation of SPI](SoftwareSpi/README.md)

-### RFID modules
+### RFID/NFC modules

 * [PN532 - RFID and NFC reader](Pn532/README.md)
 * [RFID shared elements](Card/README.md)
```

krwq:
@garciaolais seems it's worse right now. While there is no visual diff the permissions of the file changed and it shows whole file as a diff

krwq:
ok I fixed the permission issue on your branch

joperezr:
Thanks for fixing this @garciaolais!

PR #738, opened 2019-09-23
Add on-board LED controller

### LINKED ISSUE #539: using /sys/class/leds
I have some linux hardware, looking at the code and through some preliminary tests, this nuget/library defaults to the sysfs driver. 

The hardware itself has some normal IO available in /sys/class/gpio, but these are the purest forms of IO. Is there anyway using the library currently, to control the LED devices pointed to in the sys/class/leds folder? 

I am just trying some code out on a yocto linux board but cant see how it is possible at present (without writing extra code)

ZhangGaoxing:
fix #539

### DIFF
--- a/src/devices/BoardLed/BoardLed.cs
+++ b/src/devices/BoardLed/BoardLed.cs
@@ -17,7 +17,7 @@ namespace Iot.Device.BoardLed
         private const string DefaultDevicePath = "/sys/class/leds";
 
         /// <summary>
-        /// The name of LED.
+        /// The name of the LED.
         /// </summary>
         public string Name { get; }
 
--- a/src/devices/BoardLed/BoardLed.cs
+++ b/src/devices/BoardLed/BoardLed.cs
@@ -60,7 +60,7 @@ namespace Iot.Device.BoardLed
         /// <summary>
         /// Get all BoardLed instances of on-board LEDs.
         /// </summary>
-        /// <returns></returns>
+        /// <returns>BoardLed instances.</returns>
         public static IEnumerable<BoardLed> EnumerateLeds()
         {
             var infos = Directory.GetDirectories(DefaultDevicePath)
--- a/src/devices/BoardLed/BoardLed.cs
+++ b/src/devices/BoardLed/BoardLed.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -12,9 +13,15 @@ namespace Iot.Device.BoardLed
     /// <summary>
     /// On-board LED on the device.
     /// </summary>
-    public class BoardLed
+    public class BoardLed : IDisposable
     {
         private const string DefaultDevicePath = "/sys/class/leds";
+
+        private StreamReader _brightnessReader;
+        private StreamWriter _brightnessWriter;
+        private StreamReader _triggerReader;
+        private StreamWriter _triggerWriter;
+        private StreamReader _maxBrightnessReader;
 
         /// <summary>
         /// The name of the LED.
@@ -24,17 +31,22 @@ namespace Iot.Device.BoardLed
         /// <summary>
         /// The kernel controls the trigger of the LED.
         /// </summary>
+        /// <remarks>
+        /// The kernel provides some triggers which let the kernel control the LED.
+        /// For example, the red light of Raspberry Pi, whose trigger is "default-on", which makes it keep lighting up.
+        /// If you want to operate the LED, you need to remove the trigger, which is to set its trigger to none.
+        /// </remarks>
         public string Trigger { get => GetTrigger(); set => SetTrigger(value); }
 
         /// <summary>
         /// The current brightness of the LED.
         /// </summary>
-        public int Brightness { get => GetBrightness("brightness"); set => SetBrightness(value); }
+        public int Brightness { get => GetBrightness(); set => SetBrightness(value); }
 
         /// <summary>
         /// The max brightness of the LED.
         /// </summary>
-        public int MaxBrightness { get => GetBrightness("max_brightness"); }
+        public int MaxBrightness { get => GetMaxBrightness(); }
 
         /// <summary>
         /// Creates a new instance of the BoardLed.
@@ -43,19 +55,16 @@ namespace Iot.Device.BoardLed
         public BoardLed(string name)
         {
             Name = name;
+
+            Initialize();
         }
 
         /// <summary>
         /// Get all triggers of current LED.
         /// </summary>
         /// <returns>The name of triggers.</returns>
-        public IEnumerable<string> EnumerateTriggers()
-        {
-            using StreamReader reader = new StreamReader(File.Open($"{DefaultDevicePath}/{Name}/trigger", FileMode.Open, FileAccess.Read));
-
-            return Regex.Replace(reader.ReadToEnd(), @"\[|\]", "")      // remove selected chars
+        public IEnumerable<string> EnumerateTriggers() => Regex.Replace(_triggerReader.ReadToEnd(), @"\[|\]", "")      // remove selected chars
                 .Split(' ');
-        }
 
         /// <summary>
         /// Get all BoardLed instances of on-board LEDs.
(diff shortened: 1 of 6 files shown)


src/devices/BoardLed/BoardLed.cs
```diff
@@ -0,0 +1,110 @@
...
+            if (value > 255)
+            {
+                value = 255;
+            }
+            else if (value < 0)
+            {
+                value = 0;
+            }
+
+            using StreamWriter writer = new StreamWriter(File.Open($"{DefaultDevicePath}/{Name}/brightness", FileMode.Open));
```
krwq: would it make sense to store 1 instance of StreamWriter per class and just SetLength(0) + WriteWhatever + Flush? For reading just set position to 0 and then read

src/devices/BoardLed/BoardLed.cs
```diff
@@ -0,0 +1,110 @@
...
+    {
+        private const string DefaultDevicePath = "/sys/class/leds";
+
+        /// <summary>
+        /// The name of the LED.
+        /// </summary>
+        public string Name { get; }
+
+        /// <summary>
+        /// The kernel controls the trigger of the LED.
```
krwq: would be useful to add some remarks on what does this mean and how to use it

krwq: also there is an API to enumerate triggers and get 1 trigger? Presumably this api will return some kind of concatenated version of the other one? Does it make sense to hide one or another?

ZhangGaoxing: For example, the power red light of Raspberry Pi, whose trigger is `default-on`(keep lighting up), and the green is `mmc0`(Light up when when writing SD card). If I want to operate the LED, I need to remove the trigger, which is to set its trigger to `none`. In addition, the kernel provides some default triggers that can be changed by setting their triggers.

src/devices/BoardLed/BoardLed.cs
ZhangGaoxing: I don't quite understand what you mean. Do I need to save two instances when I read `brightness` and `trigger` files?

src/devices/BoardLed/BoardLed.cs
ZhangGaoxing: @krwq btw: Please review #664 if you have time. Only by merging can I add `SoundDevice` next.😄

src/devices/BoardLed/BoardLed.cs
krwq: I think for brightness it makes sense to always store it. For trigger its optional since I expect this to be set only once per instance but both options are fine with me. Also I don't feel strongly about fixing this so you can leave as is until someone complains there is a lag or something which I don't expect anyone to

src/devices/BoardLed/BoardLed.cs
krwq: Please add example (if not present already) showing how this should be used since the API is not exactly obvious

src/devices/BoardLed/BoardLed.cs
```diff
@@ -88,23 +112,25 @@ private void SetBrightness(int value)
...
-            using StreamWriter writer = new StreamWriter(File.Open($"{DefaultDevicePath}/{Name}/trigger", FileMode.Open));
+            _triggerWriter.Write(name);
+            _triggerWriter.Flush();
+        }
 
-            writer.Write(name);
+        private void Initialize()
+        {
+            _brightnessReader = new StreamReader(File.Open($"{DefaultDevicePath}/{Name}/brightness", FileMode.Open, FileAccess.Read, FileShare.ReadWrite));
+            _brightnessWriter = new StreamWriter(File.Open($"{DefaultDevicePath}/{Name}/brightness", FileMode.Open, FileAccess.Write, FileShare.ReadWrite));
```
krwq: you can actually re-use the same handle here but `leaveOpen: true` on half of them to not double dispose

src/devices/BoardLed/BoardLed.cs
```diff
@@ -88,23 +112,25 @@ private void SetBrightness(int value)
...
-            using StreamReader reader = new StreamReader(File.Open($"{DefaultDevicePath}/{Name}/trigger", FileMode.Open, FileAccess.Read));
-
-            return Regex.Match(reader.ReadToEnd(), @"(?<=\[)(.*)(?=\])").Value;
-        }
+        private string GetTrigger() => Regex.Match(_triggerReader.ReadToEnd(), @"(?<=\[)(.*)(?=\])").Value;
 
         private void SetTrigger(string name)
         {
-            using StreamWriter writer = new StreamWriter(File.Open($"{DefaultDevicePath}/{Name}/trigger", FileMode.Open));
+            _triggerWriter.Write(name);
```
krwq: do you also need to SetLength(0) or position on the underlying stream?

krwq: (I'm expecting first call to work but not sure about any subsequent calls)

src/devices/BoardLed/BoardLed.cs
```diff
@@ -0,0 +1,170 @@
...
+            Initialize();
+        }
+
+        /// <summary>
+        /// Get all triggers of current LED.
+        /// </summary>
+        /// <returns>The name of triggers.</returns>
+        public IEnumerable<string> EnumerateTriggers()
+        {
+            _triggerReader.DiscardBufferedData();
```
krwq: why not flush? I think the only buffered data will be written by this library

src/devices/BoardLed/BoardLed.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        private void SetBrightness(int value)
+        {
+            if (value > 255)
+            {
+                value = 255;
+            }
+            else if (value < 0)
+            {
+                value = 0;
```
krwq: Math.Clamp?

src/devices/BoardLed/BoardLed.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            _triggerReader.DiscardBufferedData();
+            _triggerReader.BaseStream.Position = 0;
+
+            return Regex.Match(_triggerReader.ReadToEnd(), @"(?<=\[)(.*)(?=\])").Value;
+        }
+
+        private void SetTrigger(string name)
+        {
+            _triggerWriter.BaseStream.Position = 0;
```
krwq: I think this should be SetLength,  otherwise I expect changing 123 to 23 will end up being 233

src/devices/BoardLed/BoardLed.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            _triggerWriter.Write(name);
+            _triggerWriter.Flush();
+        }
+
+        private void Initialize()
+        {
+            FileStream brightnessStream = File.Open($"{DefaultDevicePath}/{Name}/brightness", FileMode.Open);
+            _brightnessReader = new StreamReader(brightnessStream, Encoding.ASCII, false, 2, true);
```
krwq: use named args and if there is some buffering passed use the smallest possible or no buffering option if possible

src/devices/BoardLed/samples/Program.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        static void Main(string[] args)
+        {
+            // Open the green led on Raspberry Pi.
+            using BoardLed led = new BoardLed("led0");
+
+            string defaultTrigger = led.Trigger;
+
+            // LED can be controlled only if the trigger is set to none.
+            led.Trigger = "none";
```
krwq: I'd recommend to use this API for some time and then once we have a better idea on what's useful figure out some better APIs on top of it (i.e. `DisableAllTriggers()`; `RestoreTriggers()`) - I don't expect most of the people to know how to use this unless there is some better option provided. Other option could be something along the lines of:

```csharp
struct Trigger
{
  public static Trigger None => new Trigger("none");

  public string Value { get; private set; }
  public Trigger(string value)
  {
    Value = value;
  }
}
```

and then define couple of other common triggers and leave it as that - otherwise it's really hard to know how this should be used.

krwq: please ping me if you'll skip it or do it so that I know if I should press the merge button

ZhangGaoxing: I don't want to change for the time being, because we provide the `EnumerateTriggers()` method for triggers. Or, we can remove the `brightness` property and provide methods to operate the LED. Set the trigger to none inside the method.

```C#
public void Open(int brightness = 255)
{
    Trigger = "none";
    ......
}
public void Close();
```

There's a problem with that. For example, after the user operates the raspberry pi red LED, it will not always light up(restore default trigger after reboot). But if I restore the default trigger in `Dispose()`, then for the user who intends to set the trigger, his settings will fail after the program is finished.

PR #741, opened 2019-09-24
Add led-matrix-weather sample

krwq:
For the first 30 seconds it displays IP address of the device then it tries to get weather data from openweathermap.org if not available then displays example data.

Videos:
- rain: https://photos.app.goo.gl/1fk6nMuECvpiD7Zr8
- partially clouded: https://photos.app.goo.gl/VBMogeFb67qoYXka6
- sun: https://photos.app.goo.gl/XvsssjiTKtk1xStD7
- clouds: https://photos.app.goo.gl/ofU9mBBmzfQKfavk7
- moon: https://photos.app.goo.gl/HkgD8PFbS6DCuSGJA (no phases)

*Text on the video might be different from what you will see when running this sample*


krwq:
Latest change is adding ticks to the clock:
https://photos.app.goo.gl/86Lj4si4YJ3L9khD7

samples/led-matrix-weather/MathUtils.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Iot.Device.LEDMatrix;
+using Iot.Device.Graphics;
+using System.Drawing;
+using System.Numerics;
+
+namespace LedMatrixWeather
+{
+    static class MathUtils
```
tarekgh: could you please add private/internal specifier?

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+using System.Drawing;
+using System.Numerics;
+using System.Diagnostics;
+using System.Net;
+using System.Net.Sockets;
+using System.Net.NetworkInformation;
+
+namespace LedMatrixWeather
+{
+    partial class Program
```
tarekgh: >     [](start = 0, length = 4)

add private/internal specifier please

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+    {
+        static Action<RGBLedMatrix> s_scenario = WeatherDemo;
+        static Stopwatch s_showLocalIp = null;
+        static string[] s_ips;
+        static bool s_networkAvailable = false;
+        static Weather s_client;
+        static OpenWeatherResponse s_weatherResponse;
+        static TimeZoneInfo s_timeZonePst = TimeZoneInfo.FindSystemTimeZoneById("America/Los_Angeles");
+
```
tarekgh: all should be private. should it be readonly only too?

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+        static void Main(string[] args)
```
tarekgh: >static [](start = 8, length = 6)

private

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+                            case ConsoleKey.Q:
+                            {
+                                s_scenario = null;
+                                break;
+                            }
+                        }
+                    }
+                }
+
+                drawing.Wait();
```
tarekgh: >drawing.Wait(); [](start = 16, length = 15)

would it be better to do await drawing; and convert Main to be async?

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+            }
+            finally
+            {
+                matrix.Dispose();
+            }
+        }
+
+        static void WeatherUpdater()
```
tarekgh: >static [](start = 8, length = 6)

private

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            while (true)
+            {
+                UpdateWeather();
+                Thread.Sleep(5000);
+            }
+        }
+
+        static void IpsGetter()
```
tarekgh: >static [](start = 8, length = 6)

private

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+                matrix.Dispose();
+            }
+        }
+
+        static void WeatherUpdater()
+        {
+            while (true)
+            {
+                UpdateWeather();
+                Thread.Sleep(5000);
```
tarekgh: >Thread.Sleep(5000); [](start = 16, length = 19)

maybe make it every 3 minutes would be more reasonable? 5 seconds is too short I guess

samples/led-matrix-weather/Program.Drawings.cs
```diff
@@ -0,0 +1,222 @@
...
+using Iot.Device.LEDMatrix;
+using Iot.Device.Graphics;
+using System.Drawing;
+using System.Numerics;
+
+namespace LedMatrixWeather
+{
+    using static MathUtils;
+    
+    partial class Program
```
tarekgh: >partial [](start = 4, length = 7)

private

samples/led-matrix-weather/Program.Drawings.cs
```diff
@@ -0,0 +1,222 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Iot.Device.LEDMatrix;
+using Iot.Device.Graphics;
+using System.Drawing;
+using System.Numerics;
+
+namespace LedMatrixWeather
+{
+    using static MathUtils;
```
tarekgh: >using [](start = 4, length = 5)

move this up

samples/led-matrix-weather/Program.Drawings.cs
```diff
@@ -0,0 +1,222 @@
...
+    
+    partial class Program
+    {
+        static float Line(Vector2 uv, float len)
```
tarekgh: >static [](start = 8, length = 6)

private

samples/led-matrix-weather/Program.Drawings.cs
```diff
@@ -0,0 +1,222 @@
...
+            {
+                return Math.Max(uv.Y - len, Math.Abs(uv.X));
+            }
+            else
+            {
+                return Math.Abs(uv.X);
+            }
+        }
+
+        static Vector3 Clock(Vector2 uv, DateTimeOffset time)
```
tarekgh: >static [](start = 8, length = 6)

private

samples/led-matrix-weather/Program.Drawings.cs
```diff
@@ -0,0 +1,222 @@
...
+            float tickCircleDist = Math.Abs(len - 0.4f);
+            float dots = 1.0f - smoothstep(0f, 0.02f, Math.Max(tickDist, tickCircleDist));
+
+            return mix(
+                new Vector3(1, 0, 0),
+                new Vector3(Math.Max(dots, Math.Max(outerCircle, secondsLine)), minutesLine, hoursLine),
+                smoothstep(0f, 0.01f, innerRadiusDist));
+        }
+
+        static Vector3 OpenWeatherIcon(Vector2 uv, string icon, float time)
```
tarekgh: >static [](start = 8, length = 6)

private

samples/led-matrix-weather/Program.Drawings.cs
```diff
@@ -0,0 +1,222 @@
...
+                default:
+                {
+                    // Don't know the icon
+                    // We're in Seattle so let's assume rain
+                    return IconRain(uv, time);
+                }
+            }
+        }
+
+        static Vector3 IconSunny(Vector2 uv, float time)
```
tarekgh: >static [](start = 8, length = 6)

private

samples/led-matrix-weather/Program.Drawings.cs
```diff
@@ -0,0 +1,222 @@
...
+            float rayLen = RayLength * rayShape;
+            float radius = InnerRadius + rayLen;
+            float radiusDist = uv.Length() - radius;
+            return mix(
+                new Vector3(1, 1, 0),
+                new Vector3(0, 0, 0),
+                smoothstep(0f, 0.02f, radiusDist));
+        }
+
+        static float Clouds(Vector2 uv, float time)
```
tarekgh: >static [](start = 8, length = 6)

private

samples/led-matrix-weather/Program.Drawings.cs
```diff
@@ -0,0 +1,222 @@
...
+
+            float r2 = 0.15f;
+            Vector2 p2 = new Vector2(0.2f, 0.05f - 0.05f * (float)Math.Sin(2 * pi * time / 4f + 0.9f));
+            float c2 = (uv - p2).Length() - r2;
+
+            float clouds = Math.Min(Math.Min(c0, c1), c2);
+            return smoothstep(0f, 0.2f, clouds + 0.05f);
+        }
+
+        static readonly Vector3 s_cloudsColor = new Vector3(0.3f, 0.3f, 0.3f);
```
tarekgh: >static read [](start = 8, length = 11)

private

I'll stop adding these comments, so I'll assume you'll fix them all. thanks.

samples/led-matrix-weather/led-matrix-weather.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <RootNamespace>led_matrix_weather</RootNamespace>
```
joperezr: NIT: this isn't really needed so let's get rid of it.

samples/led-matrix-weather/led-matrix-weather.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <RootNamespace>led_matrix_weather</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
```
joperezr: Given that these are sample projects which we ideally want people to copy-paste and have them working, could we change these two references to package references instead?

samples/led-matrix-weather/led-matrix-weather.csproj
```diff
@@ -0,0 +1,26 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
+      <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
+    </ProjectReference>
+    <ProjectReference Include="../../src/Iot.Device.Bindings/Iot.Device.Bindings.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Protobuf Include="weather.proto" />
```
joperezr: mm Didn't you require a package reference to GRPC tools in order for this to work?

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+        static Stopwatch s_showLocalIp = null;
+        static string[] s_ips;
+        static bool s_networkAvailable = false;
+        static Weather s_client;
+        static OpenWeatherResponse s_weatherResponse;
+        static TimeZoneInfo s_timeZonePst = TimeZoneInfo.FindSystemTimeZoneById("America/Los_Angeles");
+
+        static void Main(string[] args)
+        {
+            if (args.Length != 1)
```
joperezr: Shouldn't we just show example data in case no args ar passed in? that seems to be what is suggested by your Usage text. I would just add support for a `/?` or `-?` that would print the Usage.

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+                Console.WriteLine();
+                Console.WriteLine("If wrong key is used or there is no network connection");
+                Console.WriteLine("example data will be displayed.");
+                Console.WriteLine();
+                Console.WriteLine("For the first 30 seconds of execution");
+                Console.WriteLine("IP addresses will be displayed instead of data.");
+                return;
+            }
+
+            s_client = new Weather(args[0]);
```
joperezr: NIT: Should we have this inside a try and fail  if the key was invalid?

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+                // Choke all errors.
+                // This is function is used for diagnostics
+                // we don't want it to interrupt the process in any way
+                s_ips = new string[1] { "Cannot get IPs" };
+            }
+
+            s_showLocalIp = Stopwatch.StartNew();
+        }
+
+        static void UpdateWeather()
```
joperezr: Why make this a void instead of a OpenWeatherResponse?

samples/led-matrix-weather/Program.cs
joperezr: Should this have a cancellationToken? Similar for the  IpsGetter.

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+                Console.WriteLine("example data will be displayed.");
+                Console.WriteLine();
+                Console.WriteLine("For the first 30 seconds of execution");
+                Console.WriteLine("IP addresses will be displayed instead of data.");
+                return;
+            }
+
+            s_client = new Weather(args[0]);
+
+            UpdateWeather();
```
joperezr: Why do you need to call this from the main thread and not just wait till you dispatch to the other thread of WeatherUpdater?

samples/led-matrix-weather/Program.cs
joperezr: We could also probably have this be configurable.

samples/led-matrix-weather/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+                UpdateWeather();
+                Thread.Sleep(5000);
+            }
+        }
+
+        static void IpsGetter()
+        {
+            try
+            {
+                s_ips = GetLocalNetworkIPAddresses().ToArray();
```
joperezr: Why do we want to run this in a separate thread? shouldn't it be possible to just run it on the main thread?

samples/led-matrix-weather/Program.cs
krwq: only some of it

samples/led-matrix-weather/led-matrix-weather.csproj
krwq: This actually should be removed since I don't use this anymore

samples/led-matrix-weather/led-matrix-weather.csproj
krwq: @joperezr this needs relatively recent version of the packages otherwise it won't display correctly. I.e.: https://github.com/dotnet/iot/pull/737
Also I'm planning to use this project for testing in the future

samples/led-matrix-weather/Program.cs
krwq: actually seems only time zone doesn't change

samples/led-matrix-weather/Program.cs
krwq: actually it starts producing warnings for the two threads I'm running but do not want to be awaited, seems like it's cleaner to just `Wait()`

samples/led-matrix-weather/Program.cs
krwq: will change to a minute

samples/led-matrix-weather/Program.cs
krwq: I'll prefer to refrain from adding command line parser here. Once I need more options will use library for that.

Showing example data was meant to be a fallback option since this likely will be used in headless scenario and better to show something than nothing.

samples/led-matrix-weather/Program.cs
krwq: the constructor doesn't validate the key so doesn't matter

samples/led-matrix-weather/Program.cs
krwq: it's hard to name when this returns the value. I.e.:
`bool GetLatestWeatherInfoOrExampleData(out OpenWeatherResponse)` where true would mean that using real data and false example but then bool argument is meaningless unless you read the code so needs docs etc. Current version is just simpler to read since there is only one caller anyway (actually two but used in the same context)

samples/led-matrix-weather/Program.cs
krwq: Don't care for now, the threads will just die with the process anyway and they don't do anything which needs cleanup

samples/led-matrix-weather/Program.cs
krwq: @joperezr we need some kind of data to be displayed and it just avoids unnecessary jerking the screen with info being updated in the first few seconds

samples/led-matrix-weather/Program.cs
krwq: Will change this once we have a need for more option

samples/led-matrix-weather/Program.cs
krwq: That's what I initially did and it works most of the time but when running this when device is booting the network card might still be initializing which is causing an exception or hang causing the updating thread to freeze.

krwq:
merging since we need to link README.md somewhere and this didn't seem to have any blocking comments.

PR #754, opened 2019-10-01
Add unix sound device

### LINKED ISSUE #715: Add unix sound device
I used ALSA to implement `SoundDevice`(some basic methods, such as playback and recording), and I will clean up the code and submit a PR recently. Record it here in case someone does the same job.

ZhangGaoxing:
fix #715 
(After merging, I will update the sample of radio transmitter to make it a "real" transmitter.)


src/devices/Interop/Unix/Interop.Libraries.cs
```diff
@@ -5,4 +5,6 @@
 internal partial class Interop
 {
     private const string LibcLibrary = "libc";
+
```
krwq: nit: remove extra enter

src/devices/Interop/Unix/Libasound/Interop.libasound.cs
```diff
@@ -0,0 +1,143 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal partial class Interop
+{
+    [DllImport(AlsaLibrary)]
+    internal static extern IntPtr snd_strerror(int errnum);
```
krwq: Have you perhaps had a chance to test this on ARM64? IntPtr size depends on the architecture so I'd feel better with this if we did

src/devices/Interop/Unix/Libasound/Interop.libasound.cs
```diff
@@ -0,0 +1,143 @@
...
+
+    // <param name="pcm">snd_pcm_t**</param>
+    [DllImport(AlsaLibrary)]
+    internal static extern int snd_pcm_open(ref IntPtr pcm, string name, snd_pcm_stream_t stream, int mode);
```
krwq: should pcm be out? Also any chance we could wrap the pointer into SafeHandle?

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,551 @@
...
+
+            SetWavHeader(saveStream, header);
+
+            OpenRecordingPcm();
+            PcmInitialize(_recordingPcm, header, ref @params, ref dir);
+            ReadStream(saveStream, header, ref @params, ref dir);
+            CloseRecordingPcm();
+        }
+
+        private void SetWavHeader(Stream wavStream, WavHeader header)
```
krwq: I think this should be called WriteWavHeader

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,551 @@
...
+        {
+            Span<byte> writeBuffer2 = stackalloc byte[2];
+            Span<byte> writeBuffer4 = stackalloc byte[4];
+
+            wavStream.Position = 0;
```
krwq: I'd avoid changing position in case we ever want to change this to stream it over the network or something that doesn't support seeking

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,551 @@
...
+
+                Encoding.ASCII.GetBytes(header.Subchunk2Id, writeBuffer4);
+                wavStream.Write(writeBuffer4);
+
+                BinaryPrimitives.WriteUInt32LittleEndian(writeBuffer4, header.Subchunk2Size);
+                wavStream.Write(writeBuffer4);
+            }
+            catch
+            {
+                throw new Exception("Write WAV header error.");
```
krwq: we should either let the exception bubble up or wrap it (set inner exception to original) with some better typed exception

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,551 @@
...
+            }
+        }
+
+        private WavHeader GetWavHeader(Stream wavStream)
```
krwq: ReadWavHeader

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,551 @@
...
+        {
+            Span<byte> readBuffer2 = stackalloc byte[2];
+            Span<byte> readBuffer4 = stackalloc byte[4];
+
+            wavStream.Position = 0;
```
krwq: ditto

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,551 @@
...
+
+                wavStream.Read(readBuffer4);
+                header.Subchunk2Id = Encoding.ASCII.GetString(readBuffer4).ToCharArray();
+
+                wavStream.Read(readBuffer4);
+                header.Subchunk2Size = BinaryPrimitives.ReadUInt32LittleEndian(readBuffer4);
+            }
+            catch
+            {
+                throw new Exception("Non-standard WAV file.");
```
krwq: ditto

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,551 @@
...
+            {
+                while (wavStream.Read(readBuffer) != 0)
+                {
+                    _errorNum = Interop.snd_pcm_writei(_playbackPcm, (IntPtr)buffer, frames);
+                    ThrowErrorMessage("Can not write data to the device.");
+                }
+            }
+        }
+
+        private unsafe void ReadStream(Stream saveStream, WavHeader header, ref IntPtr @params, ref int dir)
```
krwq: outputStream or just output

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,551 @@
...
+        {
+            long volumeLeft, volumeRight;
+
+            OpenMixer();
+
+            _errorNum = Interop.snd_mixer_selem_get_playback_volume(_elem, snd_mixer_selem_channel_id.SND_MIXER_SCHN_FRONT_LEFT, &volumeLeft);
+            _errorNum = Interop.snd_mixer_selem_get_playback_volume(_elem, snd_mixer_selem_channel_id.SND_MIXER_SCHN_FRONT_RIGHT, &volumeRight);
+            ThrowErrorMessage("Get playback volume error.");
+
+            CloseMixer();
```
krwq: this should be in the `try .. finally` or wrapped by SafeHandle and `using` statement used

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,551 @@
...
+
+            return (volumeLeft + volumeRight) / 2;
```
krwq: I'm not sure this is correct I'd expect this to be closer to `Math.Max(volumeLeft, volumeRight)` or even `Math.Sqrt(volumeLeft * volumeLeft + volumeRight * volumeRight)`. I'm fine with this being as is though

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,551 @@
...
+            ThrowErrorMessage("Set recording volume error.");
+
+            CloseMixer();
+        }
+
+        private unsafe long GetRecordingVolume()
+        {
+            long volumeLeft, volumeRight;
+
+            OpenMixer();
```
krwq: can we open it once and leave it open and close in the Dispose?

src/devices/Media/SoundDevice/SoundConnectionSettings.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Media
+{
+    /// <summary>
+    /// The connection settings of a sound device.
+    /// </summary>
+    public class SoundConnectionSettings
```
krwq: this seems to be mixing recording and playback settings. I'd recommend to either unify options (i.e. just DeviceName) or split classes.

src/devices/Media/SoundDevice/SoundDevice.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// </summary>
+        /// <param name="wavStream">WAV stream.</param>
+        public abstract void Play(Stream wavStream);
+
+        /// <summary>
+        /// Sound recording.
+        /// </summary>
+        /// <param name="second">Recording duration(In seconds).</param>
+        /// <param name="savePath">Recording save path.</param>
+        public abstract void Record(uint second, string savePath);
```
krwq: `outputFilePath`, `recordTimeSeconds` (also I'd make seconds `double` if possible and swap args)

src/devices/Media/SoundDevice/SoundDevice.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// Sound recording.
+        /// </summary>
+        /// <param name="second">Recording duration(In seconds).</param>
+        /// <param name="saveStream">Recording save stream.</param>
+        public abstract void Record(uint second, Stream saveStream);
```
krwq: similar as above

src/devices/Media/SoundDevice/WavHeader.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Media
+{
+    internal struct WavHeader
+    {
+        public char[] ChunkId { get; set; }
```
krwq: Looks like it might be worth to create:

```csharp
struct WavHeaderChunk
{
    public char[] ChunkId { get; set; }
    public uint ChunkSize { get; set; }
}
```

It can be re-used for this chunk and subchunks

src/devices/Interop/Unix/Libasound/Interop.libasound.cs
ZhangGaoxing: the pointer is a `snd_pcm_t` struct. Because I don't need to modify it, I use pointers instead.

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
ZhangGaoxing: I will remove try-catch

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
ZhangGaoxing: How should I modify it? If the method is executed according to the current process, there is no need to change the position of the stream.

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
ZhangGaoxing: Actually, I don't know how to express the total volume. ALSA doesn't provide a total volume `Get` function, so I use the average of left and right channels to represent the total volume.

src/devices/Media/SoundDevice/SoundConnectionSettings.cs
ZhangGaoxing: In the case of multiple sound card, it is possible to set device name. And I don't really want to separate the settings. For example, there's no recording device on the Raspberry Pi, and if I use SoundDevice, it's going to be complicated.

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
krwq: I think you should just remove this line and also make sure that after you're done reading you've read the whole header (position is exactly after the header but without using position explicitly)

src/devices/Interop/Unix/Libasound/Interop.libasound.cs
krwq: I'm not sure I understand. Per https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m.html#ga8340c7dc0ac37f37afe5e7c21d6c528b

the signature of this is:

```c++
int snd_pcm_open (snd_pcm_t** pcmp, const char* name, snd_pcm_stream_t stream, int mode)
```

and per description pcmp is `Returned PCM handle`.

That suggests that this sould be:

```suggestion
    internal static extern int snd_pcm_open(out IntPtr pcm, string name, snd_pcm_stream_t stream, int mode);
```

`in`/`out`/`ref` work equivalent on the interop layer, the only person seeing the difference is caller. I.e. currently if you call `snd_pcm_open` you'll need to first initialize `pcm`, with `out` you don't also it better reflects what actually happens

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
krwq: Assuming those are in dB (which btw should be mentioned in the xml doc) https://en.wikipedia.org/wiki/Sound_pressure#Multiple_sources

src/devices/Interop/Unix/Libasound/Interop.libasound.cs
ZhangGaoxing: Because of the nature of the work, I rarely come into contact with the code in this area. I learned some more knowledge.😁

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
ZhangGaoxing: This value is not in dB. There is special method for dB.

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
ZhangGaoxing: Yes, but I don't think it's a common API.

src/devices/Media/SoundDevice/WavHeader.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Iot.Device.Media
+{
+    internal struct WavHeader
+    {
+        public WavHeaderChunk Chunk { get; set; }
+
+        public char[] Format { get; set; }
+
+        public WavHeaderChunk Subchunk1 { get; set; }
```
MarkCiliaVincenti: Should the 'c' be capitalized here?

src/devices/Media/SoundDevice/WavHeader.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public uint SampleRate { get; set; }
+
+        public uint ByteRate { get; set; }
+
+        public ushort BlockAlign { get; set; }
+
+        public ushort BitsPerSample { get; set; }
+
+        public WavHeaderChunk Subchunk2 { get; set; }
```
MarkCiliaVincenti: Should the 'c' be capitalized here?

src/devices/Media/SoundDevice/samples/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            SoundConnectionSettings settings = new SoundConnectionSettings();
+            using SoundDevice device = SoundDevice.Create(settings);
+
+            Console.WriteLine("Recording...");
+            device.Record(10, "/home/pi/record.wav");
```
krwq: can we use current path? (just file name)

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,553 @@
...
+    /// </summary>
+    internal class UnixSoundDevice : SoundDevice
+    {
+        private IntPtr _playbackPcm;
+        private IntPtr _recordingPcm;
+        private IntPtr _mixer;
+        private IntPtr _elem;
+        private int _errorNum;
+
+        private static readonly object playbackInitializationLock = new object();
```
krwq: Please use `s_playbackInitializationLock` and similar for static fields

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,553 @@
...
+        /// Play WAV file.
+        /// </summary>
+        /// <param name="wavStream">WAV stream.</param>
+        public override void Play(Stream wavStream)
+        {
+            IntPtr @params = new IntPtr();
+            int dir = 0;
+            WavHeader header = ReadWavHeader(wavStream);
+
+            OpenPlaybackPcm();
```
krwq: please put these in `try`..`finally` so that you ensure closing the device in case something goes wrong

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,553 @@
...
+            ReadStream(outputStream, header, ref @params, ref dir);
+            CloseRecordingPcm();
+        }
+
+        private void WriteWavHeader(Stream wavStream, WavHeader header)
+        {
+            Span<byte> writeBuffer2 = stackalloc byte[2];
+            Span<byte> writeBuffer4 = stackalloc byte[4];
+
+            Encoding.ASCII.GetBytes(header.Chunk.ChunkId, writeBuffer4);
```
krwq: nit: these almost asks for method for writing strings :)

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -0,0 +1,553 @@
...
+        protected override void Dispose(bool disposing)
+        {
+            ClosePlaybackPcm();
+            CloseRecordingPcm();
+            CloseMixer();
+
+            base.Dispose(disposing);
+        }
+
+        private void ThrowErrorMessage(string message)
```
krwq: nit: I'd recommend to rename this to `ThrowIfError`

src/devices/Media/SoundDevice/SoundDevice.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.IO;
+
+namespace Iot.Device.Media
+{
+    /// <summary>
+    /// The communications channel to a sound device.
+    /// </summary>
+    public abstract partial class SoundDevice : IDisposable
+    {
+        /// <summary>
```
krwq: I think you're missing private parameterless constructor

src/devices/Interop/Unix/Libasound/Interop.libasound.cs
krwq: fine to just add a comment in the readme that this wasn't tested on 64-bit ARM and to investigate interop types if there are any issue

krwq:
@ZhangGaoxing got time to address remaining comments? I'm fine with merging this as is but would prefer if at least some of the comments were addressed (the comment about ARM64 and potential Close issues)

ZhangGaoxing:
@krwq I have other things to do. I'll make changes at the weekend.

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
ZhangGaoxing: The `ThrowIfError` method will call `Dispose` when an exception occurs internally. I think this step is not necessary. But I will add `try...finally`.

src/devices/Media/SoundDevice/SoundDevice.cs
ZhangGaoxing: 😅I don't quite understand why. It is a `abstract` class

Issue #757, opened 2019-10-01
Linux PWM implementation throws during construction/setting frequency in some cases


krwq:
Repro:

```csharp
using (var pwm = PwmChannel.Create(0, 0, 500))
{
   // duty cycle is 0.5 by default
   pwm.Frequency = 10000; // throws
}
```

from now constructing pwm channel with frequency 10k Hz is not possible without manually editing duty cycle (sysfs).

This is happening because duty cycle still has the old value in the driver (measured in nanosecond) and updating frequency causes that duty cycle gets out of range

shaggygi:
I noticed this when we were updating to the new PwmChannel.  I believe it is because there is not enough time delayed when [opening the port (via file)](https://github.com/dotnet/iot/blob/90f0786d82afbf12357868733ce5ddb2b2b8c8e8/src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs#L170) before interacting with the DC/freq.  I mentioned it then, but we couldn't figure out a good approach on timing.

I could duplicate using the [PwmPinOutput command](https://github.com/dotnet/iot/blob/master/tools/DevicesApiTester/Commands/Pwm/PwmPinOutput.cs) in the DeviceTesterApi.

After the exception occurs, you can still see the channel opened by reading the file.  And writing the DC/freq works by then, as well.

krwq:
@shaggygi, I've added a fix in https://github.com/dotnet/iot/pull/758 - I got 100% repro before and now doesn't repro with the fix.

In some scenarios this is not particularly obvious:
- you create PWM with one frequency and dispose
- you create PWM with different frequency and it throws when constructing

The issue is with duty cycle being out of range (also IO exception suggests that: `Invalid argument`)

joperezr:
Would we want to service 1.0 with this?

krwq:
I don't have strong opinion either way. On one hand this is relatively rare but on the other it's super annoying when it affects you

Issue #753, opened 2019-09-30
BME280 does not measure pressure/altitude correctly when SetPowerMode is not called


krwq:
Test:

```csharp
private static Bme280 CreateBme280()
{
    var settings = new I2cConnectionSettings(1, Bme280.DefaultI2cAddress);
    return new Bme280(I2cDevice.Create(settings));
}

const double SeaLevelPressure = 1013.25;
using (Bme280 bme280 = CreateBme280())
{
    Assert.True(bme280.TryReadTemperature(out Temperature temperature));
    Console.WriteLine($"Temperature: {temperature.Celsius} C");
    Assert.True(bme280.TryReadPressure(out double pressure));
    Console.WriteLine($"Pressure: {pressure} Pa");
    Assert.True(bme280.TryReadAltitude(SeaLevelPressure, out double altitudeMeters));
    Console.WriteLine($"Altitude: {altitudeMeters} meters");
    Assert.True(bme280.TryReadHumidity(out var relativeHumidity));
    Console.WriteLine($"Humidity: {relativeHumidity} %");
}
```

Output:

```
Temperature: 23.0049688216532 C
Pressure: 74116.26953125 Pa
Altitude: 2560.9681612177 meters
Humidity: 62.4485427633319 %
```

definitely Pressure and Altitude don't seem reasonable. I'm currently at around 15-100m above the sea level.

cc: @RobinTTY @georgemathieson

krwq:
Seems this is working around the problem with pressure:

```csharp
bme280.SetPowerMode(Bmx280PowerMode.Forced);
```

The pressure/altitude after this shows us:

```csharp
Pressure: 101477.66015625 Pa
Altitude: -12.7022561440805 meters
```

I'm definitely not under water right now so presumably sea level pressure or some other calculation is wrong as well

RobinTTY:
Yeah you will always need to perform a measurement by calling SetPowerMode (this is by design), I just tested this with a reference python library and basically got the same value for pressure:

![Capture](https://user-images.githubusercontent.com/27669180/65912370-5e842f00-e3ce-11e9-916d-eec6f4dccd48.PNG)

My phones sensor is currently giving me a value of 981.0 so not far off. I'll take a look at the Altitude formula since I didn't tocuh it in my last PR, maybe something is wrong with that.

krwq:
@RobinTTY python's bug doesn't mean we have to repeat the same bug in here.

If pressure cannot be read for whatever reason it should throw an exception or return false.
[...]

PR #758, opened 2019-10-01
Upgrade CI setup (SPI/I2C/PWM/BME280 functional tests)

krwq:
Fixes: https://github.com/dotnet/iot/issues/757

Upgrades CI setup and adds couple of functional and unit tests.

Issues found while testing:
- https://github.com/dotnet/iot/issues/753 (BME280 does not report pressure correctly when no extra settings passed)
- (fixed in the PR) https://github.com/dotnet/iot/issues/757 (PWM frequency changes are not done correctly)
- (fixed in the PR) No ArgumentNullException on null SpiDevice in Mcp3008 (didn't bother to file this one)

<s>**Please make sure to see the [Fritzing diagram](https://raw.githubusercontent.com/dotnet/iot/1937ebb9faf9071e9eb5541ec0e4e67aaf187492/src/System.Device.Gpio.Tests/ci-setup_bb.jpg)**</s> See README.md instead, fritzing was replaced with schematics and gerber

Testing details:
- PWM output is tested using ADC and vice-versa
  - PWM is converted into DAC using low pass RC filter on the output (details on the diagram)
- ADC also has extra "hard wired" inputs: 0.5VCC (voltage divider), VCC, GND
- BME280 checks if it can read values using I2C and checks if values within reasonable ranges
  - this is meant to be I2C test not BME280
- SPI is tested using Mcp3008

Do not merge until we actually plug that into CI.

Please let me know what other tests you think I should add or any recommendation to modify the circuit

cc: @ZhangGaoxing @shaggygi @Ellerbach

There is one modification I'm planning to do: test BME280 with both I2C and SPI - that should test scenario with multiple devices connected on SPI. Currently I don't have any multiplexers available and waiting for them to arrive.


Ellerbach:
On the tests, you can add a led and a photoled connected to face each others. In that case, you'll test simple GPIO both writing and reading.

On this: #753 (BME280 does not report pressure correctly when no extra settings passed)
Giving the fact that yes, you have first to select a mode. The sensor is sleeping by default. ```PowerMode.Sleep``` is the sensor default.
So before reading, you need something like that:
```csharp
i2CBmpe80.SetPowerMode(PowerMode.Forced);
//set samplings
i2CBmpe80.SetTemperatureSampling(Sampling.UltraLowPower);
i2CBmpe80.SetPressureSampling(Sampling.UltraLowPower);
```

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -45,8 +45,9 @@ internal class UnixPwmChannel : PwmChannel
             _dutyCycleWriter = new StreamWriter(new FileStream($"{_channelPath}/duty_cycle", FileMode.Open, FileAccess.ReadWrite));
             _frequencyWriter = new StreamWriter(new FileStream($"{_channelPath}/period", FileMode.Open, FileAccess.ReadWrite));
 
-            SetFrequency(frequency);
-            DutyCycle = dutyCycle;
+            // If duty cycle is set to value 
+            _dutyCycle = dutyCycle;
+            SetFrequency(frequency, forceUpdateDutyCycle: true);
```
joperezr: Did you test this? IIRC by default in the unix driver frequency is set to 0 initially, so it will fail to set duty cycle first as it has to be a number which is lower than the frequency.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -90,18 +91,31 @@ private static int GetPeriodInNanoseconds(int frequency)
...
             }
 
+            double oldDutyCycle = _dutyCycle;
+
+            if (forceUpdateDutyCycle || (oldDutyCycle != 0 && frequency > _frequency))
+            {
+                // Since frequency is higher setting frequency now might
+                // cause errors because old duty cycle value (in nanoseconds)
+                // might now exceed maximum possible range
+                DutyCycle = 0;
```
joperezr: I believe this didn't work either. I tried in the past setting duty cycle to 0 but couldn't make it work.

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -7,30 +7,16 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
-[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
joperezr: Why did you remove this? I believe we still need our tests to not run in parallel.

src/System.Device.Gpio.Tests/Mcp3008Tests.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+using System.Device;
+using System.Device.Pwm;
+using Iot.Device.Adc;
+
+using Xunit;
+
+using static System.Device.Gpio.Tests.SetupHelpers;
```
joperezr: NIT: this makes the code a bit harder to follow. I would just instead call `SetupHelpers.CreateAdc()`

src/System.Device.Gpio.Tests/Mcp3008Tests.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace System.Device.Gpio.Tests
+{
+    public class Mcp3008Tests
```
joperezr: do we also plan to write tests that actually test the device functionality? I ask because technically we are not really testing in here if SPI is working or not.

src/System.Device.Gpio.Tests/ProtocolTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        [Fact]
+        public void I2C_Bme280CanRead()
+        {
+            using (Bme280 bme280 = CreateBme280())
+            {
+                Assert.True(bme280.TryReadTemperature(out Temperature temperature));
+
+                // assuming that tests are run in the room temperature
+                // worst case scenario: it's very hot outside
+                Assert.InRange(temperature.Celsius, 15, 40);
```
joperezr: I would rather not make these kind of assumptions, same goes for the pressure and humidity. The simple fact that we were able to instantiate a bme280 means I2c is working, because as part of the device initialization we check a specific register to make sure we are talking to a bme280 and we throw if that is not the case, so the fact that we were able to create it is proof enough that I2c communication works.

src/System.Device.Gpio.Tests/ProtocolTests.cs
```diff
@@ -0,0 +1,99 @@
...
+                        // Settling time is ~1.1ms (when going from GND to max)
+                        // R=4.7k ohm
+                        // C=0.1uF
+                        // f=10k Hz
+                        // peak to peak is ~0.18V (5.5% VCC)
+                        //   in this scenario is 2 * error
+                        Thread.Sleep(2);
+
+                        int expected = (int)Math.Round(pwm.DutyCycle * 1023.0);
+                        AdcValueAround(expected, adc.Read(0));
```
joperezr: This looks like the sort of test that might be flaky. It's fine to keep it as is for now, but if we start seeing that the test fails some builds we might have to rethink a better way to test pwm.

src/System.Device.Gpio.Tests/RaspberryPiDriverTests.Linux.cs
```diff
@@ -7,6 +7,7 @@
 
 namespace System.Device.Gpio.Tests
 {
+    [Trait("requirement", "root")]
```
joperezr: Currently all tests run as root already

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
joperezr: Nvm I just saw you moved it.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: Yes, I haven't had any issues locally yet. Also I don't set the frequency here, just changing the field - the setting of the duty cycle will actually happen after setting frequency

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: I haven't seen any issues locally yet. I'll try different combinations like restarting my PI and stuff to see if anything breaks

src/System.Device.Gpio.Tests/Mcp3008Tests.cs
krwq: FunctionalTests do test MCP3008

src/System.Device.Gpio.Tests/ProtocolTests.cs
krwq: humidity cannot be outside of 0-100 range since 0 means completely dry and 100 means as humid as it can be. Pressure unless you're running these tests on the moon or during hurricane I don't expect it to ever fail. For temperature I can increase the lower range but I think anything above 40C might suggest that lab is on fire (we can increase it by a a few degrees either way if anyone complains)

src/System.Device.Gpio.Tests/ProtocolTests.cs
krwq: Why do you think it would fail? The only time I expect this might fail is:
- input voltage is not stable
  - if this is the case we should fix the lab
- something got disconnected or got physically broken (including MCP3008 stopping working)
  - legitimate failure which we should fix
- Helix runs multiple tests on the same Pi at the same time

src/System.Device.Gpio.Tests/RaspberryPiDriverTests.Linux.cs
krwq: In the CI yes but not when I test locally.

src/System.Device.Gpio.Tests/ProtocolTests.cs
krwq: + I actually have run this test in a loop on my box for some time and no issues

src/System.Device.Gpio.Tests/Mcp3008Tests.cs
krwq: How is that different from creating a base class and using derived member or using partial classes we use everywhere in corefx?

krwq:
@Ellerbach 

> you can add a led and a photoled connected to face each others. In that case, you'll test simple GPIO both writing and reading.

I was thinking about this but I didn't want to do it for couple of reasons:
- We have multiple of such boards next to each other and lab is fairly well lit so I don't think this would be reliable unless you design some kind of tube to isolate them from the outside
- I eventually would like to convert breadboard setup to PCB - photoresistor will make it harder
- Not sure if this will add more value than current GPIO test (two pins connected to each other through resistor)

src/System.Device.Gpio.Tests/Mcp3008Tests.cs
joperezr: When just reading the code online, first thing people will look for is a method defined in the class, then they will check if this class has a parent class. Both of those checks will fail to tell you where the method is defined, and the `using static` statement is not something that is very intuitive to a lot of devs, so given that all you are really saving here is typing about 10 chars, I would personally just do the full SetupHelpers.CreateAdc() for readability.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: Ok, seems this is actually a problem just after boot but not after you set it the first time. Logic without any issues was a bit tricky though but not seeing any problems on updated version

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: You're right but this is only the case just after the boot

src/System.Device.Gpio.Tests/ProtocolTests.cs
maloo: I would do Math.Max(40, CpuTemp) if sensor is anywhere near the Pi. Heat travels easily on a pcb, like a Pi hat.

krwq: good point, especially we have a binding for the CPU temperature (at least on Linux)

krwq:
I've updated to the working without issues state. Feedback hasn't been addressed yet

krwq:
@joperezr we should figure out what to do with CI on 1.0 branch since it will need to run against the old setup. Perhaps we should leave 1-2 Pis for 1.0 runs?

joperezr:
> we should figure out what to do with CI on 1.0 branch since it will need to run against the old setup. Perhaps we should leave 1-2 Pis for 1.0 runs?

Don't we still plan on having two pins connected to each other via a resistor in the new setup? I thought that was the case, and if so, that would mean that the tests would still be able to run on the new configuration right? (perhaps with some pin numbering changes)

krwq:
fair enough, we can do that

krwq:
This is basically ready, CI setup is upgraded. Left:
- PIs need to be configured to enable all protocols on both Windows and Linux (also docs) - me or Jose will need to SSH to each PI in all 4 queues and upgrade it

krwq:
Ok, Linux setup is working, working on Windows now.

krwq:
@shaggygi @joperezr do you know how to enable PWM on Windows IoT Core? I've tried changing driver to direct memory access but that seems that it might require writing new GPIO/SPI/I2C/PWM controllers and for Inbox Driver I can't find any instructions on how to enable it - some page from couple of years ago claims it's not supported which I'm not sure is still valid info

shaggygi:
@krwq sorry, have not tried on Windows.  I came across https://github.com/lprichar/WindowsIotPwmExample, but you already mentioned Lightning library.

krwq:
@Ellerbach do you perhaps remember what is the max frequency we can reliably get with SoftwarePwmChannel?

krwq:
Ok, seems for regular precision it's around 50Hz, for high precision it seemed 10k Hz was good enough (it was around 9.7kHz)

krwq:
Hopefully this will pass after merging https://github.com/dotnet/iot/issues/927

krwq:
I've disabled PWM on Windows queues and created https://github.com/dotnet/iot/issues/934 to fix that

joperezr:
One small comment is that it would probably be good if we stuff all of the gerber/schematics of the PCB to be in a subfolder inside the tests System.Device.Gpio.Tests folder to not pollute the source files

src/System.Device.Gpio.Tests/README.md
```diff
@@ -1,52 +1,127 @@
...
+
+Our CI is currently using [this schematics](board-schematics.pdf) with Raspberry PI 3.
+
+You can download [gerber files](board-gerber.zip) if you'd like to manufacture PCB yourself (this is a 2 layer board).
+
+Following list of components are needed for PCB:
+
+| Component | Quantity |
+| --- | --- |
+| MCP3008 I/SL | 1 |
```
joperezr: Apart from listing these requirements and showing our PCB, I think it would be useful if we also provided a fritzing diagram since most contributors will probably be using a breadboard to test this. You don't have to block this PR on that, but do at least log an issue so that we add this diagram so people know how to connect each of the components in order for our tests to run without a need to change the code configurations

src/System.Device.Gpio.Tests/README.md
```diff
@@ -1,52 +1,127 @@
...
+| BME280 breakout board | 1 |
+| 2x20 female header (Raspberry PI hat) | 1 |
+| 1x7 female header (for BME280 breakout board) | 1 |
+| 4.7k resistor in 0603 (US) package | 4 |
+| 0.1u (100n) capacitor in 0603 (US) package | 1 |
+| (optional) 1x3 male header | 5 |
+| (optional) 1x4 male header | 4 |
+| (optional) 1x8 male header | 1 |
+| (optional) 200 ohm resistor in 0603 (US) package | 1 |
+| (optional) LED in 0603 (US) package | 1 |
```
pgrawehr: Minor: If connecting an LED an additional resistor (~150 Ohms) is required, or this will be a short-living experience...

krwq: I mentioned this line above: `| (optional) 200 ohm resistor in 0603 (US) package | 1 |` - I usually use 200 ohms for LEDs

src/System.Device.Gpio.Tests/README.md
krwq: there is schematics pdf with description how to connect, I feel fritzing doesn't look any clearer and rather more confusing

src/System.Device.Gpio.Tests/README.md
pgrawehr: Uh sorry, I have overlooked that line.

krwq:
I'll rerun CI couple of times to make sure nothing (new) is flaky

krwq:
Hmm https://github.com/dotnet/iot/issues/832: `System.IO.IOException : Error 121 performing I2C data transfer.`

joperezr:
> Hmm #832: System.IO.IOException : Error 121 performing I2C data transfer.

This usually means one of the sensors is not plugged in correctly. It also makes sense if it fails intermittently because it may not run on the same device next time. We should try to figure out which Pi did that failure happen in in order to make sure we reconnect the sensor.

krwq:
@joperezr failed again same way, ddvsorpil025 again

joperezr:
> @joperezr failed again same way, ddvsorpil025 again

Yup I guess that is expected, I'll try to go to the lab later today or ping the infrastructure people to get it out of the queue.

krwq:
@joperezr I think it might be worth to just try unplug the PI and plug it back in. I recall had something similar locally once - simple reboot wasn't enough but unplugging was good enough. If that helps perhaps there are some ways to fix this in software

krwq:
Ok after sending dozens of jobs I haven't seen any of the I2C issues - only one of the unrelated tests is flaky (testing pull up/down mode). I think we're good with merging this.

Ellerbach:
And congratulations! Very nice work @krwq on this PR! Automating hardware test is far to be easy especially in a real CI/CD pipeline.

PR #766, opened 2019-10-04
Change the Raspberry Pi detection logic.

### LINKED ISSUE #661: Raspberry Pi not detected using Ubuntu and Debian
I believe that I have discovered that the detection logic for determining if a board is a Raspberry Pi only works if running Raspbian and not Ubuntu or Debian.

This is because /proc/cpuinfo doesn't bring back information about the SOC on Debian or Ubuntu.

I believe we should consider improving the detection logic for future releases perhaps using model information from the device tree.

Frankenslag:
Fixes #661

This PR aims to solve an issue with the detection of a raspberry pi within the GpioController.

At the moment the /dev/cpuinfo is read to identify the Pi but it seems that the Pi info is only made available within Raspbian.

To resolve this the detection mechanism is augmented by reading the device tree to find the Raspberry Pi identifier. If there are any issues with reading the device tree or the identifier is not found then it falls back to original detection logic.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
@@ -31,25 +31,27 @@ namespace System.Device.Gpio
         /// <returns>A driver that works with the board the program is executing on.</returns>
         private static GpioDriver GetBestDriverForBoard()
         {
-            string[] cpuInfoLines = File.ReadAllLines(CpuInfoPath);
-            Regex regex = new Regex(@"Hardware\s*:\s*(.*)");
+            string[] cpuInfoLines;
+            Regex regex;
+
+            // a first chance of finding a raspberry pi..... perhaps not on Raspbian. note
+            // that we want this to fail quietly if something goes wrong with the test and continue
+            // with detecting a PI usiing /dev/cpuinfo
+            try
+            {
+                if (File.ReadAllText(ModelInfoPath).StartsWith("Raspberry Pi"))
+                {
+                    return new RaspberryPi3Driver();
+                }
+            }
+            catch { }
+
+            regex = new Regex(@"Hardware\s*:\s*(.*)");
+            cpuInfoLines = File.ReadAllLines(CpuInfoPath);
+
             foreach (string cpuInfoLine in cpuInfoLines)
             {
-                Match match;
-
-                // a first chance of finding a raspberry pi..... perhaps not on Raspbian. note
-                // that we want this to fail quietly if something goes wrong with the test and continue
-                // with detecing a PI usiing /dev/cpuinfo
-                try
-                {
-                    if (File.ReadAllText(ModelInfoPath).StartsWith("Raspberry Pi"))
-                    {
-                        return new RaspberryPi3Driver();
-                    }
-                }
-                catch { }
-
-                match = regex.Match(cpuInfoLine);
+                Match match = regex.Match(cpuInfoLine);
 
                 if (match.Success)
                 {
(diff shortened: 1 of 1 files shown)


src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -34,7 +35,22 @@ private static GpioDriver GetBestDriverForBoard()
...
+                // that we want this to fail quietly if something goes wrong with the test and continue
+                // with detecing a PI usiing /dev/cpuinfo
+                try
+                {
+                    if (File.ReadAllText(ModelInfoPath).StartsWith("Raspberry Pi"))
+                    {
+                        return new RaspberryPi3Driver();
+                    }
+                }
+                catch { }
```
mjjames: shouldn't this be done before the cpuInfoLines are even read and then spun through? Assuming you mean this to be an optimisation so do this first before any other pattern matching

Frankenslag: > shouldn't this be done before the cpuInfoLines are even read and then spun through? Assuming you mean this to be an optimisation so do this first before any other pattern matching

Thanks @mjjames. It must a Friday thing.

It even passed testing.

Correction coming up.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -30,11 +31,28 @@ public GpioController(PinNumberingScheme numberingScheme)
...
+            // that we want this to fail quietly if something goes wrong with the test and continue
+            // with detecting a PI usiing /dev/cpuinfo
+            try
+            {
+                if (File.ReadAllText(ModelInfoPath).StartsWith("Raspberry Pi"))
+                {
+                    return new RaspberryPi3Driver();
+                }
+            }
+            catch { }
```
krwq: when do we expect exceptions here? If this is just reading the file should you instead just call `File.Exists`?

Frankenslag: @krwq my thoughts behind this is that I was trying make it as simple as possible I.e. just a single line to do the testing. You would probably want to ignore errors at this point because you are not sure how future operating Linux distributions will do things so I didn’t see why catching the non existence of a file with an exception would be an issue in this case.

Frankenslag: Non rpi boards may not have the file or rpi distributions that do not use a variant of the standard device tree and firmware.

krwq: actually I think it might make sense to let it throw for new board - this way we can react early to any differences on the new distros - if we choke the error we won't know about it until it bites us

Frankenslag:
@krwq If this detection fails then the original detection code is still in place and that returns a generic driver if a specific board is not found.

krwq:
@Frankenslag @pgrawehr whats left here we can re-use and wasn't covered by @pgrawehr's PR?

pgrawehr:
One thing that should be considered (but likely no longer as part of this PR) is making the detection result public.

Frankenslag:
> @Frankenslag @pgrawehr whats left here we can re-use and wasn't covered by @pgrawehr's PR?

@krwq, @pgrawehr 

The essence of this PR is to allow a Raspberry Pi to be detected even if not running Raspbian (i.e. Debian or Ubuntu). When this PR was put together only Raspbian exposed /proc/cpuinfo. I am not sure if that is still true.

If it is true though and @pgrawehr's PR is the one where cpuinfo is read to determine the model then the problem associated with not using Raspbian would still exist.

pgrawehr:
I haven't considered that. I guess somebody would need to test again with a current ubuntu version.

joperezr:
[Triage] Given that we have already merged @pgrawehr PR and it covers most of what this was trying to cover we'll go ahead and close this. @Frankenslag feel free to reopen if you think this would still add some value to the library.

Issue #752, opened 2019-09-30
Generic calculations taken from temperature, relative humidity and pressure


MarkCiliaVincenti:
There are some formulas that can be done from sensor data, such as taking the temperature, relative humidity and pressure and outputting the absolute humidity.

Or getting the dew point from temperature and relative humidity.

Also a constant for the default sea level pressure, which I believe is incorrect in Bme280.sample.cs (I believe it should read 1013.25 instead of 1033.00).

Where would be the best place to put these? I'd like to make a pull request.

krwq:
@MarkCiliaVincenti you will need to inspect:
https://github.com/dotnet/iot/tree/master/src/devices/Bmxx80

if the calculation/data is BME280 specific (or you don't have knowledge/devices to test) then put it directly under Bme280.cs otherwise upgrade on the base class.

For sea level pressure - I'd recommend creating a separate PR for that since this might trigger some extra conversations. I agree that **mean** sea level pressure should be 1013.25  but **mean** implies that it differs depending on location (if I was living at the sea level and it always read 1033 I'd strongly argue with people claiming 1013.25).
I can currently see only using that value in the samples though - which specific API are you referring to? (if this is just sample then fine with updating in the same PR)

MarkCiliaVincenti:
I'm not talking about BME280-specific calculations but very generic weather formulas :)

Agreed with mean sea level pressure varying, but since 1013.25 is widely considered the global average, shouldn't this meaningful number be put somewhere in the code?

krwq:
@MarkCiliaVincenti in case of SeaLevel I think we should consider adding another unit to:
https://github.com/dotnet/iot/tree/master/src/devices/Units

and add it as a constant (or static readonly field).

Since I was already thinking of proposing Pressure unit (because seems like there are at least 4 different units of pressure) I think we should just go ahead and do it.

Full scope of this feature might require a bit more work to convert all usages of Pressure in the repo but we could help with that.


For the formula since I don't know much about this specific domain my gut feeling tells me to create new project under `devices/` called `WeatherHelpers` or some even more generic names where we could gather all helper functions (project name doesn't matter that much since we don't use it anywhere but namespace and class name does)

krwq:
Interestingly when I calculate altitude with 1013.25hPa it tells me my altitude is negative where 1033 shows 150m which sounds reasonable

krwq:
[...]

PR #768, opened 2019-10-05
New Pressure Unit

MarkCiliaVincenti:
- Created new structure Pressure
- Includes static value for Mean Sea-Level Pressure (MSLP) (1013.25 hPa)
- Updated Bmxx80 (including BMP180), LPS25H and SenseHat to use this new unit, which breaks existing code since the return type is now Pressure.

### DIFF
--- a/src/devices/Bmp180/Bmp180.cs
+++ b/src/devices/Bmp180/Bmp180.cs
@@ -101,9 +101,6 @@ namespace Iot.Device.Bmp180
         /// <summary>
         ///  Calculates the altitude in meters from the mean sea-level pressure.
         /// </summary>
-        /// <param name="seaLevelPressure"> 
-        ///  Sea-level pressure
-        /// </param>
         /// <returns>
         ///  Height in meters from the sensor
         /// </returns>
--- a/src/devices/Bmp180/Bmp180.cs
+++ b/src/devices/Bmp180/Bmp180.cs
@@ -81,7 +81,7 @@ namespace Iot.Device.Bmp180
             int p = (B7 < 0x80000000) ? (int)((B7 * 2) / B4) : (int)((B7 / B4) * 2);
             X1 = (((p * p) / 65536 ) * 3038) / 65536;
             
-            return Pressure.FromPa(p + ( ((((p * p) / 65536 ) * 3038) / 65536) + ((-7357 * p) / 65536) + 3791) / 8);
+            return Pressure.FromPascal(p + ( ((((p * p) / 65536 ) * 3038) / 65536) + ((-7357 * p) / 65536) + 3791) / 8);
         }
 
         /// <summary>
@@ -95,7 +95,7 @@ namespace Iot.Device.Bmp180
         /// </returns>
         public double ReadAltitude(Pressure seaLevelPressure)
         {
-            return 44330.0 * (1.0 - Math.Pow((ReadPressure().Pa / seaLevelPressure.Pa), (1.0 / 5.255)));
+            return 44330.0 * (1.0 - Math.Pow((ReadPressure().Pascal / seaLevelPressure.Pascal), (1.0 / 5.255)));
         }
         
         /// <summary>
@@ -120,7 +120,7 @@ namespace Iot.Device.Bmp180
         /// </returns>
         public Pressure ReadSeaLevelPressure(double altitude = 0.0)
         {
-            return Pressure.FromPa(ReadPressure().Pa / Math.Pow((1.0 - (altitude / 44333.0)), 5.255));
+            return Pressure.FromPascal(ReadPressure().Pascal / Math.Pow((1.0 - (altitude / 44333.0)), 5.255));
         }
 
         /// <summary>
--- a/src/devices/Bmp180/Bmp180.cs
+++ b/src/devices/Bmp180/Bmp180.cs
@@ -93,9 +93,23 @@ namespace Iot.Device.Bmp180
         /// <returns>
         ///  Height in meters from the sensor
         /// </returns>
-        public double ReadAltitude(Pressure seaLevelPressure = Pressure.MeanSeaLevel)
+        public double ReadAltitude(Pressure seaLevelPressure)
         {
             return 44330.0 * (1.0 - Math.Pow((ReadPressure().Pa / seaLevelPressure.Pa), (1.0 / 5.255)));
+        }
+        
+        /// <summary>
+        ///  Calculates the altitude in meters from the mean sea-level pressure.
+        /// </summary>
+        /// <param name="seaLevelPressure"> 
+        ///  Sea-level pressure
+        /// </param>
+        /// <returns>
+        ///  Height in meters from the sensor
+        /// </returns>
+        public double ReadAltitude()
+        {
+            return ReadAltitude(Pressure.MeanSeaLevel);
         }
 
         /// <summary>
--- a/src/devices/Bmp180/Bmp180.cs
+++ b/src/devices/Bmp180/Bmp180.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -64,9 +64,9 @@ namespace Iot.Device.Bmp180
         ///  Reads the pressure from the sensor
         /// </summary>
         /// <returns>
-        ///  Atmospheric pressure in Pa
+        ///  Atmospheric pressure
         /// </returns>
-        public double ReadPressure()
+        public Pressure ReadPressure()
         {
             // Pressure Calculations
             int B6 = CalculateTrueTemperature() - 4000;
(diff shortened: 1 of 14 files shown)


MarkCiliaVincenti:
With reference to https://github.com/dotnet/iot/issues/752

src/devices/Units/Pressure.cs
```diff
@@ -0,0 +1,17 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Units
+{
+    /// <summary>
+    /// Class representing pressure
+    /// </summary>
+    public static class Pressure
```
krwq: What I meant is Pressure unit similar to Temperature unit with this constant in it.

MarkCiliaVincenti: There can be conversions between different units here, eventually. It should be in its own class simply because the mean sea-level pressure has absolutely nothing to do with temperature.

krwq: I meant to add a full support, fine for now I guess though

src/devices/Units/Pressure.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        /// <summary>
+        /// The mean sea-level pressure (MSLP) is the average atmospheric pressure at mean sea level.
+        /// </summary>
+        public const double MeanSeaLevelPressure = 1013.25;
```
krwq: I'd recommend to remove `Pressure` suffix from the name since it's already in the name of the class

src/devices/Units/Pressure.cs
MarkCiliaVincenti: Complete rewrite made.

src/devices/Units/Pressure.cs
MarkCiliaVincenti: Complete rewrite made.

MarkCiliaVincenti:
The changes in this pull request are now breaking.

src/devices/Bmp180/Bmp180.cs
```diff
@@ -81,21 +81,32 @@ public double ReadPressure()
...
         /// </summary>
         /// <param name="seaLevelPressure"> 
-        ///  Sea-level pressure in hPa
+        ///  Sea-level pressure
         /// </param>
         /// <returns>
         ///  Height in meters from the sensor
         /// </returns>
-        public double ReadAltitude(double seaLevelPressure = 101325.0)
+        public double ReadAltitude(Pressure seaLevelPressure)
```
krwq: `Pressure seaLevelPressure = Pressure.MeanSeaLevel`?

src/devices/Units/Pressure.cs
```diff
@@ -0,0 +1,119 @@
...
+        
+        /// <summary>
+        /// Pressure in Pa
+        /// </summary>
+        public double Pa => _pa;
+        
+        /// <summary>
+        /// Pressure in mbar
+        /// </summary>
+        public double Mbar => MbarRatio * _pa;
```
krwq: I think the casing should be `MBar`

src/devices/Units/Pressure.cs
```diff
@@ -0,0 +1,119 @@
...
+        
+        /// <summary>
+        /// Pressure in kPa
+        /// </summary>
+        public double Kpa => KpaRatio * _pa;
+        
+        /// <summary>
+        /// Pressure in hPa
+        /// </summary>
+        public double Hpa => HpaRatio * _pa;
```
krwq: I think the casing should be `HPa`... cc: @terrajobst

src/devices/Units/Pressure.cs
```diff
@@ -0,0 +1,119 @@
...
+        
+        /// <summary>
+        /// Pressure in inHg
+        /// </summary>
+        public double Inhg => InhgRatio * _pa;
```
krwq: InHg

src/devices/Units/Pressure.cs
```diff
@@ -0,0 +1,119 @@
...
+        
+        /// <summary>
+        /// Pressure in mmHg
+        /// </summary>
+        public double Mmhg => MmhgRatio * _pa;
```
krwq: MmHg

krwq: same elsewhere

MarkCiliaVincenti:
Are you sure? It's HttpRequest not HTTPRequest. It's Iot.Units not
IoT.Units. So similarly I use this casing standard.

krwq:
@MarkCiliaVincenti let me check with @terrajobst since I'm not sure. Physics and math has already casing established for years

src/devices/Bmp180/Bmp180.cs
MarkCiliaVincenti: > `Pressure seaLevelPressure = Pressure.MeanSeaLevel`?

You can't use defaults because Pressure.MeanSeaLevel is not a constant value.

src/devices/Units/Pressure.cs
MarkCiliaVincenti: Actually there's a problem. The metric unit for millibar is mbar (everything lowercase); that's why I used Mbar. However Mbar (with the capital M) stands for megabar. Therefore I think I should use Millibar and MillibarRatio, do you agree?

krwq:
@MarkCiliaVincenti ok got naming advices: use full names in singular form (i.e. HectoPascal instead of Pa, I think only `Hg` is fine to leave)

src/devices/Bmxx80/samples/Bme680.sample.cs
```diff
@@ -1,94 +1,94 @@
...
+        /// </summary>
+        static void Main()
+        {
+            Console.WriteLine("Hello BME680!");
+
+            // The I2C bus ID on the Raspberry Pi 3.
+            const int busId = 1;
+
+            var i2cSettings = new I2cConnectionSettings(busId, Bme680.DefaultI2cAddress);
+            var unixI2cDevice = I2cDevice.Create(i2cSettings);
```
krwq: remove unix from the name since it's no longer valid

MarkCiliaVincenti: Done.

MarkCiliaVincenti:
> @MarkCiliaVincenti ok got naming advices: use full names in singular form (i.e. HectoPascal instead of Pa, I think only `Hg` is fine to leave)

Done but I found Hg to be weird on its own.... InchOfHg? Anyway check the names now and let me know what you think please.

krwq:
IMO: `MillimeterOfMercury`/`MillimeterHg`, `InchOfMercury`/`InchHg`

we can change it in the future if people complain

MarkCiliaVincenti:
> IMO: `MillimeterOfMercury`/`MillimeterHg`, `InchOfMercury`/`InchHg`
> 
> we can change it in the future if people complain

So similarly Hectopascal, right?

krwq:
@MarkCiliaVincenti `Hectopascal` sounds correct to me then (in my mind it's still two words though :smile:)

MarkCiliaVincenti:
> @MarkCiliaVincenti `Hectopascal` sounds correct to me then (in my mind it's still two words though 😄)

Great! OK for merge now?

src/devices/Bmp180/Bmp180.cs
```diff
@@ -105,11 +116,11 @@ public double ReadAltitude(double seaLevelPressure = 101325.0)
         ///  altitude in meters
         /// </param>
         /// <returns>
-        ///  Pressure in Pascals
+        ///  Pressure
         /// </returns>
-        public double ReadSeaLevelPressure(double altitude = 0.0)
+        public Pressure ReadSeaLevelPressure(double altitude = 0.0)
```
krwq: We should consider moving this to WeatherHelpers

MarkCiliaVincenti: Yes definitely, I will move them but this will be in a completely different branch. I don't want to mix the two up.

src/devices/Bmxx80/samples/Bme280.sample.cs
```diff
@@ -46,7 +47,7 @@ static void Main(string[] args)
                     i2CBmpe80.TryReadTemperature(out var tempValue);
                     Console.WriteLine($"Temperature: {tempValue.Celsius} C");
                     i2CBmpe80.TryReadPressure(out var preValue);
-                    Console.WriteLine($"Pressure: {preValue} Pa");
+                    Console.WriteLine($"Pressure: {preValue.Pascal} Pa");
```
krwq: optional: change these to show hPa instead since it's more useful

MarkCiliaVincenti: I tried to not change as much from the logic of the samples. I noticed they need to be more aligned.

src/devices/Bmxx80/samples/Bme280.sample.cs
```diff
@@ -44,9 +45,9 @@ static void Main(string[] args)
 
                     //read values
                     i2CBmpe80.TryReadTemperature(out var tempValue);
-                    Console.WriteLine($"Temperature: {tempValue.Celsius} C");
+                    Console.WriteLine($"Temperature: {tempValue.Celsius} °C");
```
krwq: use `\uXXXX` notation instead

src/devices/Bmxx80/samples/Bme280.sample.cs
MarkCiliaVincenti: Done

src/devices/Bmxx80/samples/Bme280.sample.cs
MarkCiliaVincenti: I'm not sure what you mean.

MarkCiliaVincenti: Done.

MarkCiliaVincenti:
Cheers!

MarkCiliaVincenti:
@krwq please note that the tutorial https://msdn.microsoft.com/magazine/mt833493 referenced from https://github.com/dotnet/iot will no longer be working with this merge.

MarkCiliaVincenti:
> @krwq please note that the tutorial https://msdn.microsoft.com/magazine/mt833493 referenced from https://github.com/dotnet/iot will no longer be working with this merge.

Please take note @dawidborycki

krwq:
@MarkCiliaVincenti good point, We should add some comment under the article clarifying the difference or make a tiny change to use older package version which still has the old API

krwq:
FYI: @joperezr I created the label to track these kind of changes in case we need to document them

joperezr:
@krwq I just saw this. Instead of breaking, should we just add a new overload that returns the new type instead of double? I would really like to minimize breaking-changes even for the device bindings when possible, and this case seems like it is something that could easily be avoided with a new overload

krwq:
@joperezr I'm fine with that if the old API has units in the name (which I don't think it does). If not I'd go with the breaking change on this one but I'm ok with doing that if I'm outvoted

MarkCiliaVincenti:
I don't see how you could do this without breaking, otherwise it will be messy. There's a unit for Temperature and all methods return an instance of Temperature. Pressure was missing a unit, which it now has, and is consistent with Temperature.

It's been merged a while ago now, so either way it's going to break either for those who are on stable or the ones who are on prerelease who have already adapted their code for this PR.

Furthermore there are other PRs which are dependent on this, such as the WeatherHelper.

krwq:
I agree with the "will be messy" part. I think this one is worth to make a breaking change for consistency. What we can consider though is perhaps making Pressure be implicitly convertible to double with whatever unit we used before (hPa?)

PR #770, opened 2019-10-06
Implementation of Hx711 device

ibebbs:
Implementation of Hx711 based on Ms-Iot implementation [here](https://github.com/ms-iot/hx711)

Currently missing a Fritzing diagram and not fully tested but thought I'd offer a PR early to get feedback on any other changes that might be required.


src/devices/Hx711/Hx711.cs
```diff
@@ -0,0 +1,181 @@
...
+        private byte ShiftInByte()
+        {
+            byte value = 0x00;
+
+            // Convert "GpioPinValue.High" and "GpioPinValue.Low" to 1 and 0, respectively.
+            // NOTE: Loop is unrolled for performance
+            _gpioController.Write(_clockPinNumber, PinValue.High);
+            value |= (byte)((byte)(_gpioController.Read(_dataPinNumber)) << 7);
+            _gpioController.Write(_clockPinNumber, PinValue.Low);
+            _gpioController.Write(_clockPinNumber, PinValue.High);
```
MarkCiliaVincenti: How about this?

for (int i = 7; i == 0; i--)
{
_gpioController.Write(_clockPinNumber, PinValue.High);
value |= (byte)((byte)(_gpioController.Read(_dataPinNumber)) << i);
 _gpioController.Write(_clockPinNumber, PinValue.Low);
}

ibebbs: Hi, thanks for the comment. This code is as provided by the ms-iot implementation. There is a note above it stating "Loop is unrolled for performance". Didn't really see any particular reason to change it but also have no strong feelings about it either.

MarkCiliaVincenti: Haha, woops! Didn't read the comment! My bad, though I'm intrigued whether it actually matters given that the code is compiled.

MarkCiliaVincenti: After doing some research, my guess is that the C# compiler will do the loop unrolling for you anyway, so you may as well have more maintainable code. This is not a definite answer though; I may be wrong.

ibebbs:
Mmm... testing of this device has revealed an issue. To quote a blog post I found [here](https://hackaday.io/project/25848-a-cheap-24-bit-differential-adc-for-raspberry-pi/details):

> Lastly, the HX711 has a weird serial communication protocol that requires a bit-bang on two GPIO pins. The HX711 is designed to go into a low-power sleep mode if the PD_SCK input goes high for more than 50µs. This is a difficult requirement for the Raspberry Pi because of the Linux housekeeping. There is a HX711 interface for Python, using PIGPIO, but it kept quitting after a few seconds or minutes of operation. I suspect that eventually the Raspian kernel would clobber the code and it stops for some reason. I just implemented a 1µs one-shot to control the PD_SCK input, which prevents the HX711 from sleeping and solved a lot of the interface issues -- sometimes software can’t fix everything. Since the one-shot frees the code from having to keep the PD_SCK high time less than 50us I can limp along with the generic GPIO interface. Every once in a while the data is clobbered by the kernel, but I have a pulse swallower that detects this condition and just sticks the previous “good” data into the current sample and proceeds merrily along. I can get away with this because the system is oversampled by 10x. Since a seismometer requires 24/7 operation the sleep function is not required or desired. 

As the Hx711 device in this PR can't get close to the 60μS pulse rate required to clock data out of this device I'm going to close this PR.

I will however be looking to source an ["Ads1232"](http://www.ti.com/lit/ds/symlink/ads1232.pdf) which has less stringent timing requirements. If I can get this working, I'll submit a new PR.

Ellerbach:
Back to this PR, the best way to operate is to use SPI and not GPIO. Just use MISO, MOSI and CLK and all the magic will happen with the right timing.

Issue #179, opened 2019-01-30
Add bindings for HT16K33


joperezr:
This is a 16*8 LED Controller display with keyscan and supports I2c.

joperezr:
Here is an interesting link with a similar implementation for Python: https://github.com/adafruit/Adafruit_Python_LED_Backpack/blob/master/Adafruit_LED_Backpack/SevenSegment.py

tawalke:
Will work on this one.

krwq:
Hey @tawalke, is the branch ready for review yet? What's the remaining work there? :smile:

tawalke:
Hi Krzysztof:

The base class lib is completed, I need to publish the changes made to the 7Segment class for this lib so that the branch can be reviewed.  Will do later this week.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

gLes:
Hi!

As I see this development seems to be stuck in limbo.

I have a working implementation for this controller and with a little cleanup I can put it into a PR in a day or two if you're interested.

krwq:
@gLes thanks! Please do, also please check @tawalke's implementation and see if there is anything there you can add to your implementation:
https://github.com/dotnet/iot/tree/015f68cdd5b01a3eeb8ee2043382e0723241f246/src/devices/HT16K33

gLes:
Can you give me some preliminary insight on my commit? I tried to follow suite, but it would be nice if someone could confirm I'm on the right path here 🙂

Usage, samples, tests, etc. are coming soon...

krwq:
@gLes I haven't reviewed the whole thing, just quickly skimmed and added couple of comments. I think from API perspective you should look at similar device: https://github.com/dotnet/iot/tree/master/src/devices/Tm1637 and see how it was made as an example and do it as similar as possible (if you find that you copy paste a code consider joining some pieces of the code if possible)

krwq:
This is already done: https://github.com/dotnet/iot/blob/master/src/devices/Display/README.md

PR #772, opened 2019-10-06
Add HT16K33 device binding #179

gLes:


src/devices/Ht16k33/Command.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace Iot.Device.Ht16k33
```
krwq: copyright header missing (see other .cs files for reference)

src/devices/Ht16k33/Command.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    internal class Command
```
krwq: should be an enum

src/devices/Ht16k33/Command.cs
gLes: Done.

src/devices/Ht16k33/Command.cs
gLes: Thought I could spare a few casts, but done 🙂

src/devices/Ht16k33/Enums/Address.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Iot.Device.Ht16k33.Enums
+{
+    /// <summary>
+    /// Digit address
+    /// </summary>
+    /// <remarks>
+    /// Index for the digit value in the display buffer
+    /// </remarks>
+    public enum Address
```
krwq: does this need to be public?

src/devices/Ht16k33/Font.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// Used to mask upper 4 bits of a byte for a single hexadecimal value
+        /// </summary>
+        private const byte HexadecimalMask = 0b0000_1111;
+
+        /// <summary>
+        /// Hexadecimal digit (0..15) fonts
+        /// </summary>
+        private static readonly byte[] s_hexDigitFonts = new byte[]
+        {
+            0b0011_1111, // 0
```
krwq: could these be defined as another enum values, i.e.:

```csharp
enum Segment : ...
{
...
LetterA = LeftBottom | LeftTop | ...,
}
```

similarly as we done in https://github.com/dotnet/iot/blob/master/src/devices/Tm1637/Character.cs

There is also a high chance you can re-use or copy&paste it (or at least most of it)

src/devices/Ht16k33/Ht16k33.cs
```diff
@@ -0,0 +1,162 @@
...
+        #region Protected members
+        /// <summary>
+        /// I2C device interface
+        /// </summary>
+        protected I2cDevice _i2cDevice;
+
+        /// <summary>
+        /// Initial brightness level
+        /// </summary>
+        protected byte brightness;
```
krwq: this should be one of following:
- private field with prefixed with `_`
- protected property with regular capital case

krwq: similarly for other fields

src/devices/Ht16k33/Enums/Address.cs
gLes: It's for the indexer property in SevenSegment.cs:
https://github.com/dotnet/iot/pull/772/files#diff-912ce972ab785da08d13dd57b875ba50R86

src/devices/Ht16k33/Font.cs
gLes: I saw the similar enum in the other device project, but there are slight differences (e.g. dot handling), even though they are very close I wasn't quite sure how to reuse them in both places just yet.

I generated these "fonts" in Excel, I think it's pretty compact this way and it's not public anyway. If you insist I can convert them to enum flag combinations.

src/devices/Ht16k33/Enums/Address.cs
krwq: Would it make sense to just take int as an input and translate it into this internally?

gLes: That could work too, I'll do it in my next amend.

src/devices/Ht16k33/SevenSegment.cs
```diff
@@ -0,0 +1,235 @@
...
+            Write((byte)first, (byte)second, (byte)third, (byte)fourth);
+
+        /// <summary>
+        /// Write 4 hexadecimal digits display (from left to right)
+        /// </summary>
+        /// <param name="first">first digit</param>
+        /// <param name="second">second digit</param>
+        /// <param name="third">third digit</param>
+        /// <param name="fourth">fourth digit</param>
+        public void WriteHex(byte first, byte second, byte third, byte fourth)
```
krwq: instead of this and next overload should there be just 1 which takes single `uint`?

src/devices/Ht16k33/SevenSegment.cs
```diff
@@ -0,0 +1,235 @@
...
+            {
+                case 4:
+                    WriteChars(value[0], value[1], value[2], value[3]);
+                    break;
+                case 5:
+                    if (value[2] != ':')
+                    {
+                        throw new ArgumentException(nameof(value), $"{nameof(value)}[2] must be a ':'");
+                    }
+                    Dots = Dot.CenterColon;
```
krwq: when you write:
"12:34" and then on next call "1234" will center colon be erased? I currently cannot see any logic which would remove it

src/devices/Ht16k33/Font.cs
krwq: @gLes I'm expecting as we add more segment displays in the future that we will eventually figure out how to unify this code a bit more so I'd prefer if the interface was as identical as possible.

I expect you will only need to change names to be the same and then copy & paste remaining definition from the other projects (minus not supported values)

src/devices/Ht16k33/SevenSegment.cs
gLes: It will not, it was a conscious decision, but it could go either way. The whole "12:34" branch is there because I mostly used this code to display time and this was a good shorthand.

src/devices/Ht16k33/Font.cs
gLes: Should I move these to the Common folder then?

src/devices/Ht16k33/SevenSegment.cs
gLes: Although the first if that checks for empty input does clear the dots too. I'll make it right.

src/devices/Ht16k33/SevenSegment.cs
gLes: Good idea, added more overloads.

src/devices/Ht16k33/Font.cs
gLes: Didn't make them a common item yet, but converted the binary values to enum combinations.

src/devices/Ht16k33/Ht16k33.cs
gLes: Done.

src/devices/Ht16k33/SevenSegment.cs
```diff
@@ -0,0 +1,292 @@
...
+                case 5 when value[2] != ':':
+                    throw new ArgumentException(nameof(value), $"{nameof(value)}[2] must be a ':'");
+                case 5:
+                    Dots |= Dot.CenterColon;
+                    WriteChars(value[0], value[1], value[3], value[4]);
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(value), $"{nameof(value)} can contain maximum 5 characters");
+            }
+        }
```
gLes: Might've gotten carried away with this WriteString helper and maybe some other Write* overloads too. Should I move these into an extension class or something?

src/devices/Ht16k33/SevenSegment.cs
```diff
@@ -0,0 +1,292 @@
...
+
+        /// <summary>
+        /// Write string value to display
+        /// </summary>
+        /// <param name="value">value to display, max 4 characters, or 5 characters if the 3rd character is ':' (this also turns on the center colon), or 6 characters if 1st character is also ':'</param>
+        /// <remarks>Unsupported characters will be replaced by whitespace</remarks>
+        /// <param name="alignment">alignment on display (left or right, right is default)</param>
+        /// <exception cref="ArgumentException"><paramref name="value"/>[2] must be a ':'</exception>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="value"/> can contain maximum 5 characters</exception>
+        public void WriteString(string value, Alignment alignment = Alignment.Left)
```
krwq: I'd prefer if this was a separate utility ideally working with all segment displays

krwq:
@gLes yes, I'd add the string utilities as separate PR and ideally add it so that we can re-use it for other segment displays. For now let's focus on adding HT16K33 (have some kind of utility of writing character at a position perhaps) and we can focus on this specific usability problem after this PR

gLes:
As a matter of fact this big WriteString method is pretty specific to this display configuration in regards to the positioning of all the dots:
https://raw.githubusercontent.com/dotnet/iot/c8594623e46fa6d713eb1436b6308a393c96bf35/src/devices/Ht16k33/seven-segment.jpg

krwq:
@gLes I bet we can figure out some abstraction since many segment displays have some extra punctuation

gLes:
@krwq I refactored the code, keeping extensibility in mind, but maintaining the specifics of the devices the binding is for.

I dare not add Tm1637 to this lot myself as I'm not in possession of said device, thus I couldn't really test the changes for regression IRL.

src/devices/LedDisplay/Enums/Alignment.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.LedDisplay.Enums
```
krwq: should this be `Iot.Device.SegmentDisplay`?

src/devices/LedDisplay/Ht16k33.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        /// <summary>
+        /// Initial blinking rate
+        /// </summary>
+        private BlinkRates _blinkRate;
+
+        /// <summary>
+        /// Send turn on oscillator command to device
+        /// </summary>
+        private void TurnOnOscillator() => I2cDevice.WriteByte((byte)Command.TurnOnOscillator);
```
krwq:
```suggestion
        private void TurnOnOscillator()
            => I2cDevice.WriteByte((byte)Command.TurnOnOscillator);
```

krwq: please change similarly below - this is to avoid long lines which make it harder to read

src/devices/LedDisplay/Ht16k33.cs
```diff
@@ -0,0 +1,230 @@
...
+        private void SetBlinkDisplayOn() => I2cDevice.WriteByte((byte)((byte)Command.Blink | (byte)Command.BlinkDisplayOn | ((byte)_blinkRate << 1)));
+        #endregion
+
+        #region Public members
+        #region Enums
+        /// <summary>
+        /// Blink rates
+        /// </summary>
+        /// <remarks>Specific to LED driver</remarks>
+        public enum BlinkRates : byte
```
krwq: this should be private or not nested enum

krwq: reason for not nested is that it's kinda annoying to write a type name for nested types since you always have to write things like:  `Htk16k33.BlinkRates.Off` instead of just `BlinkRates.Off`

src/devices/LedDisplay/Ht16k33.cs
```diff
@@ -0,0 +1,230 @@
...
+            TurnOnOscillator();
+            SetBrightness();
+            SetBlinkDisplayOn();
+        }
+
+        /// <summary>
+        /// Gets or sets level of display brightness brightness 0..15
+        /// </summary>
+        /// <exception cref="ArgumentOutOfRangeException">Brightness must be between 0 and 15</exception>
+        public byte Brightness
```
krwq: this abstraction feels weird because it requires for user to read the docs to use. I'd prefer if we exposed this as float/double from 0..1 which internally gets stored as byte (and reports back rounded value).

Presumably if users will allow to change this value frequently it will most likely be through UI with some kind of slider and I think float will make it easier to use.

Other alternative is to also add MaxBrightness const and make sure to use it in the sample (i.e. MaxBrightness / 2). That will be equally good.

src/devices/LedDisplay/Large4Digit7SegmentDisplay.cs
```diff
@@ -0,0 +1,310 @@
...
+            0x01, 0b0000_0000,
+            0x02, 0b0000_0000,
+            0x03, 0b0000_0000,
+            0x04, 0b0000_0000
+        };
+
+        /// <summary>
+        /// Display buffer
+        /// </summary>
+        private readonly byte[] _displayBuffer = new byte[]
```
krwq: should this one be static too? also since this is the same as `s_clearBuffer` can you assign it such as well? (you can write a comment that this is incidental)

src/devices/LedDisplay/Large4Digit7SegmentDisplay.cs
```diff
@@ -0,0 +1,310 @@
...
+        #endregion
+
+        #region Public members
+        #region Enums
+        /// <summary>
+        /// Dot display setup
+        /// </summary>
+        /// <remarks>Specific to device</remarks>
+        [Flags]
+        public enum Dot : byte
```
krwq: this should not be nested

src/devices/LedDisplay/Large4Digit7SegmentDisplay.cs
```diff
@@ -0,0 +1,310 @@
...
+            }
+
+            if (startAddress < 0 || startAddress >= MaxNumberOfDigits)
+            {
+                throw new ArgumentOutOfRangeException(nameof(startAddress));
+            }
+
+            if (digits.Length + startAddress > MaxNumberOfDigits)
+            {
+                throw new NotSupportedException($"{nameof(Large4Digit7SegmentDisplay)} only supports {MaxNumberOfDigits - startAddress} digits starting from address {startAddress}");
```
krwq: should this be `ArgumentOutOfRangeException` too? If so this and previous ifs can also be joined together

src/devices/LedDisplay/Large4Digit7SegmentDisplay.cs
```diff
@@ -0,0 +1,310 @@
...
+            }
+
+            foreach (byte digit in digits)
+            {
+                _displayBuffer[(int)s_digitAddressList[startAddress++]] = (byte)(digit & SegmentMask); // Mask Dot segment, as this display does not support it
```
krwq: nit: comment above

src/devices/LedDisplay/SevenSegmentFontHelper.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            return hexDigits;
+        }
+
+        /// <summary>
+        /// Character font table
+        /// </summary>
+        private static readonly IReadOnlyDictionary<char, Segment> s_characterFontTable = new Dictionary<char, Segment>
+        {
+            { '0', Segment.Top | Segment.TopRight | Segment.BottomRight | Segment.Bottom | Segment.BottomLeft | Segment.TopLeft },
```
krwq: I really prefer the format the other binding has:

```csharp
[Flags]
enum Character
{
    TopSegment,
    BottomSegment,
    ...
    Zero = Segment.Top | Segment.TopRight | Segment.BottomRight | Segment.Bottom | Segment.BottomLeft | Segment.TopLeft
}
```

then this whole dictionary would be also a bit clearer:
`{ '0', Character.Zero },`

etc.

Also other thing is I think this could be just something like:

```csharp
static Character[] s_digits = { Character.Zero, Character.One, ... };
static Character[] s_lowerCaseLetters = { Character.SmallLetterA };
...
```

and just right math and few ifs should give you the right output but I'm fine with dictionary as well.

src/devices/LedDisplay/SevenSegmentFontHelper.cs
```diff
@@ -0,0 +1,162 @@
...
+        /// <param name="value">input character</param>
+        /// <returns>corresponding font bits</returns>
+        public static Segment GetCharacter(char value) => s_characterFontTable.ContainsKey(value) ? s_characterFontTable[value] : Segment.None;
+
+        /// <summary>
+        /// Convert a string of characters to corresponding font segments
+        /// </summary>
+        /// <param name="value">input string</param>
+        /// <returns>list of corresponding font segments</returns>
+        public static ReadOnlySpan<Segment> GetString(ReadOnlySpan<char> value)
```
krwq: Returning span should only be done when span is subspan of the original span which is not the case here.

When array needs to be returned you have 2 options to do it correctly (or rather in way it makes sense):
- return array directly - wrapping with span doesn't give you anything in terms of allocations and it makes the code a bit harder
- return void and take output span as an input (recommended here):
`public static void GetString(ReadOnlySpan<char>, Span<Segment>);`

and then usage as following:

```csharp
string text = "abc";
Segment[] segments = new Segment[foo.Length];
XXX.GetString(text, segments);
```

src/devices/LedDisplay/category.txt
```diff
@@ -0,0 +1,3 @@
+display
+led
```
krwq: I don't think `led` applies here. LED was meant to be bindings like LED strip or LED driver. While this binding to have LEDs to turn on segments I think it doesn't exactly fit well. Easier way to think about it is to try to predict how changing this will affect https://github.com/dotnet/iot/tree/master/src/devices - In this case it will update [LED drivers](https://github.com/dotnet/iot/tree/master/src/devices#led-drivers) which you can probably see it doesn't fit here.

I'm fine with changing the name of this category if you think it's a bit bad and also fine if you remove this file at all since not having it produces the warning when running the tool which I always fix when updating (I always focus on the output doc making sense which was the main purpose of the categories)

src/devices/LedDisplay/samples/Large4Digit7SegmentDisplay.sample.cs
```diff
@@ -0,0 +1,195 @@
...
+    {
+        const string SupportedCharacters = "0123456789aAbBcCdDeEfFgGhHiIjJlLnNoOpPrRsStuUyYzZ-_|°     ";
+
+        static void Main(string[] args)
+        {
+            // Initialize display (busId = 1 for Raspberry Pi 2 & 3)
+            var display = new Large4Digit7SegmentDisplay(I2cDevice.Create(new I2cConnectionSettings(busId: 1, DefaultI2cAddress)))
+            {
+                // Set max brightness
+                Brightness = 15
```
krwq: that's exactly why I think this should be `Brightness = 1.0` or `Brightness = Large4Digit7SegmentDisplay.MaxBrightness` - otherwise it's hard to reason about this code and why 15 is max

krwq:
@gLes thanks for updating this, looks much better now! I think there are still ways to improve this a bit so let's try to iterate a bit more on this.

src/devices/LedDisplay/Enums/Alignment.cs
gLes: The same driver supports and is sold in configurations with LED matrices: https://www.adafruit.com/?q=mini%208x8%20backpack

src/devices/LedDisplay/Ht16k33.cs
gLes: I know it's annoying, but it's specific to the driver, so I thought it should be namespaced somehow with the driver. I'll try to add some abstraction to this to resolve.

src/devices/LedDisplay/Ht16k33.cs
gLes: I like MaxBrightness better. I feel that using a float could result in rounding errors where you set a certain value and you get something different back if you read it and that might lead to some confusion.

src/devices/LedDisplay/Large4Digit7SegmentDisplay.cs
gLes: If the display buffer's static then you can only control a single device within one process.

src/devices/LedDisplay/Large4Digit7SegmentDisplay.cs
gLes: This Dot configuration is specific to the backpack device. Should I move it out of the class and rename it to `Large4Digit7SegmentDisplayDot` to maintain namespacing?

src/devices/LedDisplay/Large4Digit7SegmentDisplay.cs
gLes: I put it at the end because it relates to the very last part of the expression.

src/devices/LedDisplay/SevenSegmentFontHelper.cs
gLes: I'm still getting used to spans 🙂

From an interface usability standpoint I think returning an array is better.

src/devices/LedDisplay/Enums/Alignment.cs
krwq: sounds good then, didn't know that!

src/devices/LedDisplay/Ht16k33.cs
krwq: Other option is `Htk16k33BlinkRate` (btw this should be singular unless it's `[Flags]`)

src/devices/LedDisplay/Ht16k33.cs
krwq: @gLes I don't mind either way. You pretty much always get different value back when you assign `0.1` to any variable

src/devices/LedDisplay/Large4Digit7SegmentDisplay.cs
krwq: yes, let's move it out

src/devices/LedDisplay/Large4Digit7SegmentDisplay.cs
krwq: I'd create a local for that expression then and put the comment above that. (note verbosity is not an issue since it will make it more readable)

src/devices/LedDisplay/SevenSegmentFontHelper.cs
krwq: I don't mind either way since expected usage is that you update this relatively rarely (no more than few times per second)

gLes:
@krwq Refactored yet again, hope this is it 🙂

src/devices/Display/LedDriver/category.txt
```diff
@@ -0,0 +1,2 @@
+display
+led driver
```
krwq: You can pick only from here: https://github.com/dotnet/iot/blob/master/tools/device-listing/device-listing.cs#L55-L96

display
segment

match for sure. If this is LED driver as well then the original version (with `led`) was correct

src/devices/Display/LedDriver/Ht16k33.cs
```diff
@@ -0,0 +1,213 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.I2c;
+using Iot.Device.Display.LedDriver.Enums;
```
krwq: I'd collapse .Enums subnamespace and put stuff directly under LedDriver

src/devices/Display/LedDriver/Ht16k33.cs
```diff
@@ -0,0 +1,213 @@
...
+
+        /// <summary>
+        /// Display on/off state
+        /// </summary>
+        private bool _displayOn = true;
+
+        /// <summary>
+        /// Initial brightness level
+        /// </summary>
+        private byte _brightness;
```
krwq: does brightness = 0 means you can't see anything or is this background brightness? If you can't see anything I'd consider adding this in the constructor as required argument since by default it won't show anything.

src/devices/Display/SevenSegment/category.txt
```diff
@@ -0,0 +1,2 @@
+display
```
krwq: there should be only one category file per folder under devices. It's fine to collapse categories into a sum of all devices. Same goes for `README.md` - it should describe all devices in the folder (extra README under samples is ok)

krwq: (technically we could allow that but that would require changing tool which generates `src/devices/README.md`)

src/devices/Display/LedDriver/Ht16k33.cs
gLes: 0 is visible. 0 won't burn your eyes out in a dark room while 15 is mostly visible in direct sunlight.

src/devices/Display/LedDriver/Ht16k33.cs
gLes: Just this one under LedDriver? Or all the others too?

gLes:
Rebased onto latest master branch.

src/devices/Display/Display.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <LangVersion>8</LangVersion>
```
krwq: `preview` or remove the line

src/devices/Display/LedDriver/BlinkRate.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Display.LedDriver
```
krwq: should the folder name be Htk16k33? the code doesn't seem to be generic to any display. I think only namespace should be Display

src/devices/Display/LedDriver/Ht16k33.cs
```diff
@@ -0,0 +1,212 @@
...
+        #region Protected members
+        /// <summary>
+        /// I2C device interface
+        /// </summary>
+        protected I2cDevice I2cDevice;
+        #endregion
+
+        #region Private members
+        #region Enums
+        private enum Command : byte
```
krwq: please move the nested classes to the bottom of the file

src/devices/Display/LedDriver/Ht16k33.cs
```diff
@@ -0,0 +1,212 @@
...
+        /// Write raw data to display buffer
+        /// </summary>
+        /// <param name="data">Array of bytes to write to the display</param>
+        /// <param name="startAddress">Address to start writing from</param>
+        public abstract void Write(ReadOnlySpan<byte> data, int startAddress = 0);
+
+        /// <summary>
+        /// Clear display buffer
+        /// </summary>
+        public abstract void Clear();
```
krwq: why are these abstract, Htk16k33 seems pretty specific

src/devices/Display/README.md
```diff
@@ -0,0 +1,50 @@
+# Display
+
+This project contains multipurpose LED display drivers and binding implementations for concrete display configurations.
+
+## Device family
+
+The **HT16K33** is LED display driver that supports multiple LED configurations and I2C communication.
+
+Adafruit sells multiple display backpacks built upon this driver:\
```
krwq: remove extra backslash from the line

src/devices/Display/README.md
```diff
@@ -0,0 +1,50 @@
...
+var display = new Large4Digit7SegmentDisplay(I2cDevice.Create(new I2cConnectionSettings(busId: 1, address: Ht16k33.DefaultI2cAddress));
+
+// Set max brightness (automatically turns on display)
+display.Brightness = display.MaxBrightness;
+
+// Write time to the display
+display.Write(DateTime.Now.ToString("H:mm").PadLeft(5));
+
+// Send buffer to the device
+display.Flush();
```
krwq: curious, considering this device is relatively short would it make sense to always flush after Write and make Flush private?

src/devices/Display/SevenSegment/Enums/Segment.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
```
krwq: namespace is probably unused

src/devices/Display/SevenSegment/ISevenSegmentDisplay.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Iot.Device.Display.SevenSegment.Enums;
```
krwq: since you're trying to generalize this I'd rather create `SevenSegmentDisplay` binding or just `Ht16k33`

src/devices/Display/SevenSegment/ISevenSegmentDisplay.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <param name="digits">a list of digits represented in segments</param>
+        /// <param name="startAddress">Address to start writing from</param>
+        void Write(ReadOnlySpan<Segment> digits, int startAddress = 0);
+
+        /// <summary>
+        /// Write a series of characters to the segmented display
+        /// </summary>
+        /// <param name="characters">a list of characters represented in fonts</param>
+        /// <param name="startAddress">Address to start writing from</param>
+        void Write(ReadOnlySpan<Font> characters, int startAddress = 0);
```
krwq: seems that using the interface will not actually cause any writing to happen because you don't flush automatically

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.cs
```diff
@@ -0,0 +1,201 @@
...
+            // Write "Iot" on the display
+            display.Write("Iot");
+            // Send buffer to the device
+            display.Flush();
+
+            // Wait 2 seconds
+            Thread.Sleep(2000);
+
+            // Change 'o' to '°' in "IoT"
+            display[1] = (Segment)FontHelper.GetCharacter('°');
```
krwq: can `°` be used directly in Write? Assuming this is just showcasing the API please at least mention that in the comment. Reading the comment it slightly implies special characters won't work and I think they do

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.cs
```diff
@@ -0,0 +1,201 @@
...
+            display.Flush();
+
+            // Wait 2 seconds
+            Thread.Sleep(2000);
+
+            // Turn off display
+            display.DisplayOn = false;
+
+            // Dispose display object (the device itself will not be turned off until powered down)
+            display.Dispose();
```
krwq: use `using` statement instead (you can use the new syntax so that you don't need to indent)

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.csproj
```diff
@@ -0,0 +1,22 @@
...
+      <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
+    </ProjectReference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Display.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Properties\" />
```
krwq: I think this might be your editor remainder. Please remove unless this is used anywhere

src/devices/Display/LedDriver/Ht16k33.cs
gLes: As I mentioned before this chip is used in multiple configurations of which I have only one hardware. After examining the available code samples for other hardware using this chipset I've come to the conclusion that the common features consist of the setup commands (turn on oscillator, set brightness, blinking), but the actual display memory layout can vary significantly. It could be a segmented display or an LED matrix that requires different coding altogether imho.

src/devices/Display/README.md
gLes: I think having the Flush method exposed is useful if you want to do multiple buffer modifications and only send them in one go, since every flush operation eventually sends the whole buffer over I2C, which could theoretically lead to intermediate states being displayed, but of course this would be so fast in real life that you wouldn't notice.

What if I added an AutoFlush option to the class?

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.cs
gLes: As a matter of fact it can, it is listed in the `SupportedCharacters` constant in line 18 which is iterated and written to the display between lines 159 and 166.

I added this line because it showcased the address operator and the `FontHelper.GetCharacter` method, and maybe also that the `Font` and `Segment` enums are interchangeable.

src/devices/Display/README.md
krwq: AutoFlush default to true sounds good to me (otherwise I'm expecting we will get an issue soon telling us that this binding doesn't work 😄). Samples can use whichever though

src/devices/Display/FontHelper.cs
```diff
@@ -0,0 +1,203 @@
...
+            Font.Whitespace, // letter M is not supported
+            Font.Letter_N,
+            Font.Letter_O,
+            Font.Letter_P,
+            Font.Whitespace, // letter Q is not supported
+            Font.Letter_R,
+            Font.Letter_S,
+            Font.Whitespace, // letter T is not supported
+            Font.Letter_U,
+            Font.Whitespace, // letter V is not supported
```
krwq: U is close enough :P (better than whitespace)

src/devices/Display/FontHelper.cs
```diff
@@ -0,0 +1,203 @@
...
+            Font.Whitespace, // letter W is not supported
+            Font.Whitespace, // letter X is not supported
```
krwq: possibly `X`? I think X is better than whitespace

krwq: treat as optional

src/devices/Display/FontHelper.cs
krwq: treat as optional

gLes: I used wikipedia mostly for inspiration:
https://en.wikipedia.org/wiki/Seven-segment_display

I deliberately left out fonts that would've been ambiguous from this helper, except for certain obvious ones we can't really live without (1, I, l, |, etc.).

gLes:
I'll change AutoFlush to default to true tonight.

krwq:
@gLes also consider calling it "FlushAfterWrite" or "BufferData" or similar

gLes:
@krwq How about `EnableBuffering`?

krwq:
@gLes `EnableBuffering` suggest it's a method name (has a verb), perhaps `BufferingEnabled`

gLes:
Changed `AutoFlush` to `BufferingEnabled` and reversed default setting.

krwq:
Thanks, please don't rebase next time, we can automatically do it on merge

src/devices/Display/Ht16k33.cs
```diff
@@ -0,0 +1,227 @@
...
+        #region Protected members
+        /// <summary>
+        /// I2C device interface
+        /// </summary>
+        protected I2cDevice I2cDevice;
+
+        /// <summary>
+        /// AutoFlush mode
+        /// </summary>
+        protected bool autoFlush = true;
```
krwq: `_bufferingEnabled`

src/devices/Display/Ht16k33.cs
```diff
@@ -0,0 +1,227 @@
...
+            TurnOnOscillator();
+            SetBrightness();
+            SetBlinkDisplayOn();
+        }
+
+        /// <summary>
+        /// Gets or sets whether buffering is enabled
+        /// </summary>
+        /// <remarks>When set to true the display buffer is only flushed to the device when explicitly calling the <see cref="Flush"/> method</remarks>
+        public bool BufferingEnabled
```
krwq: this can be just `{ get; set; }`

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.cs
```diff
@@ -0,0 +1,176 @@
...
+namespace SevenSegmentSample
+{
+    class Program
+    {
+        const string SupportedCharacters = "0123456789aAbBcCdDeEfFgGhHiIjJlLnNoOpPrRsStuUyYzZ-=_|°[]     ";
+
+        static void Main(string[] args)
+        {
+            // Initialize display (busId = 1 for Raspberry Pi 2 & 3)
+            using var display = new Large4Digit7SegmentDisplay(I2cDevice.Create(new I2cConnectionSettings(busId: 1, Ht16k33.DefaultI2cAddress)))
```
krwq: should this also set `BufferingEnabled = true;`

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.cs
```diff
@@ -0,0 +1,176 @@
...
+            display.Write("Iot");
+
+            // Wait 2 seconds
+            Thread.Sleep(2000);
+
+            // Change 'o' to '°' in "IoT"
+            display[1] = (Segment)FontHelper.GetCharacter('°');
+
+            // Flush display buffer (needs to be done explicitly when not calling Write or Dot)
+            display.Flush();
```
krwq: Considering you enable buffering later this shouldn't be needed, right?

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            // Write -42°C to display using "decimal point" between 3rd and 4th digit as the ° character
+            display.Write("-42C");
+            display.Dots = Dot.DecimalPoint;
+
+            // Send buffer to the device
+            display.Flush();
+
+            // Turn off buffering
+            display.BufferingEnabled = false;
```
krwq: Btw setter for BufferingEnabled should also call Flush when buffering is being disabled

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            // Wait 3 seconds
+            Thread.Sleep(3000);
+
+            // Iterate WriteString branches
+            foreach (var s in new KeyValuePair<string, Alignment>[] {
```
krwq: `{` on new line

krwq: please put the array in the local

gLes:
> Thanks, please don't rebase next time, we can automatically do it on merge

Sorry, force of habit.

src/devices/Display/Ht16k33.cs
gLes: Overoptimized a little bit, but I guess I should leave that to the compiler :)

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.cs
gLes: Didn't we want BufferingEnabled = false to be the default behavior?

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.cs
gLes: I deliberately left the this operator as is not to flush the buffer automatically.

src/devices/Display/samples/Large4Digit7SegmentDisplay.sample.cs
krwq: yes the comment is onlyabout sample since the usage (explicit Flush call) suggests you are using non-buffered

PR #778, opened 2019-10-08
Update device listing

krwq:
cc: @MarkCiliaVincenti @Ellerbach @ZhangGaoxing


src/devices/Media/README.md
```diff
@@ -1,4 +1,5 @@
-# Media
+# Still image recording library
```
ZhangGaoxing: SoundDevice will also be included in this folder

krwq: I was thinking to modify the title once it's added, until then I think this is correct as is. What would you change this to?

krwq: I'll leave this as is for now for the reason above. Feel free to update this in your PR (I don't want to report incomplete state even though it is inaccurate for a short time)

PR #780, opened 2019-10-09
Add a generic GPIO driver for Allwinner SoCs and Orange Pi Zero/Lite2 pin map

### LINKED ISSUE #751: [Question] Is OrangePi supported?
There is a device family based on different arm64 chips (allwinner H3,H5, etc).
They can run on Armbian linux. Also I managed to run net core successfully few months ago.
Are their GPIO supported in anyway?

ZhangGaoxing:
partial fix #751 (Although a general driver was added, I did not map the pins of all boards.)

Test on Armbian(Buster)

# Sunxi GPIO Driver for .NET

**sunxi** represents the family of ARM SoCs from Allwinner Technology. This project contains a **full function(PULL-UP, PULL-DOWN)** generic GPIO driver `SunxiDriver` for Allwinner SoCs and some special GPIO drivers like `OrangePiZeroDriver`, `OrangePiLite2Driver`.

### Generic GPIO driver: `SunxiDriver`
```C#
// Beacuse this is a generic driver, the pin scheme can only be Logical.
// The offset(base address) can be find in the corresponding SoC datasheet.
using GpioController gpio = new GpioController(PinNumberingScheme.Logical, new SunxiDriver(gpioRegisterOffset0: 0x01C20800, gpioRegisterOffset1: 0x01F02C00));

// Convert pin number to logical scheme.
int number = SunxiDriver.MapPinNumber('A', 10);
// Open the GPIO pin.
gpio.OpenPin(number);
// Set the pin mode.
gpio.SetPinMode(number, PinMode.InputPullUp);
// Read current value of the pin.
PinValue value = gpio.Read(number);
// Register a value changed callback.
gpio.RegisterCallbackForPinValueChangedEvent(10, PinEventTypes.Rising, Switch_Pressed_Handler);
```

### Special GPIO driver: `OrangePiZeroDriver`, `OrangePiLite2Driver`
```C#
// The programm get the best applicable driver automatically.
using GpioController gpio = new GpioController(PinNumberingScheme.Board);

gpio.OpenPin(10);
gpio.SetPinMode(10, PinMode.Output);
// Write a value to the pin.
gpio.Write(10, PinValue.High);
```

## Adding new drivers
1. Inheriting `SunxiDriver` Class.
    ```C#
    public class OrangePiZeroDriver : SunxiDriver { }
    ```
2. Overriding the GPIO base addresses.
    ```C#
    protected override int GpioRegisterOffset0 => 0x01C20800;
    protected override int GpioRegisterOffset1 => 0x01F02C00;
    ```
3. Overriding the pin count.
    ```C#
    // Orange Pi Zero has 17 GPIO pins.
    protected internal override int PinCount => 17;
    ```
4. Overriding the mapping method for converting a board pin number to the driver's logical numbering scheme.
   ```C#
    protected internal override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber)
    {
        return pinNumber switch
        {
            3 => MapPinNumber('A', 12),
            5 => MapPinNumber('A', 11),
            // ...
            _ => throw new ArgumentException($"Board (header) pin {pinNumber} is not a GPIO pin on the {GetType().Name} device.", nameof(pinNumber))
        };
    }
   ```

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/OrangePiLite2Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/OrangePiLite2Driver.Linux.cs
@@ -1,0 +1,53 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Device.Gpio.Drivers
+{
+    /// <summary>
+    /// A GPIO driver for the Orange Pi Lite 2.
+    /// </summary>
+    /// <remarks>
+    /// SoC: Allwinner H6 (sun50iw6p1)
+    /// </remarks>
+    public class OrangePiLite2Driver : SunxiDriver
+    {
+        protected override int GpioRegisterOffset0 => 0x0300B000;
+        protected override int GpioRegisterOffset1 => 0x07022000;
+
+        /// <summary>
+        /// Orange Pi Lite 2 has 17 GPIO pins.
+        /// </summary>
+        protected internal override int PinCount => 17;
+
+        /// <summary>
+        /// Converts a board pin number to the driver's logical numbering scheme.
+        /// </summary>
+        /// <param name="pinNumber">The board pin number to convert.</param>
+        /// <returns>The pin number in the driver's logical numbering scheme.</returns>
+        protected internal override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber)
+        {
+            return pinNumber switch
+            {
+                3 => MapPinNumber('H', 6),
+                5 => MapPinNumber('H', 5),
+                7 => MapPinNumber('H', 4),
+                8 => MapPinNumber('D', 21),
+                10 => MapPinNumber('D', 22),
+                11 => MapPinNumber('D', 24),
+                12 => MapPinNumber('C', 9),
+                13 => MapPinNumber('D', 23),
+                15 => MapPinNumber('D', 26),
+                16 => MapPinNumber('C', 8),
+                18 => MapPinNumber('C', 7),
+                19 => MapPinNumber('C', 2),
+                21 => MapPinNumber('C', 3),
+                22 => MapPinNumber('D', 25),
+                23 => MapPinNumber('C', 0),
+                24 => MapPinNumber('C', 5),
+                26 => MapPinNumber('H', 3),
+                _ => throw new ArgumentException($"Board (header) pin {pinNumber} is not a GPIO pin on the {GetType().Name} device.", nameof(pinNumber))
+            };
+        }
+    }
+}
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SunxiDriver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SunxiDriver.Linux.cs
@@ -11,10 +11,10 @@ using System.Threading.Tasks;
 namespace System.Device.Gpio.Drivers
 {
     /// <summary>
-    /// A general GPIO driver for Allwinner SoCs.
+    /// A generic GPIO driver for Allwinner SoCs.
     /// </summary>
     /// <remarks>
-    /// This is a general GPIO driver for Allwinner SoCs.
+    /// This is a generic GPIO driver for Allwinner SoCs.
     /// It can even drive the internal pins that are not drawn out.
     /// Before you operate, you must be clear about what you are doing.
     /// </remarks>
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SunxiDriver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SunxiDriver.Linux.cs
@@ -1,0 +1,470 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.Device.Gpio.Drivers
+{
+    /// <summary>
+    /// A general GPIO driver for Allwinner SoCs.
+    /// </summary>
+    /// <remarks>
+    /// This is a general GPIO driver for Allwinner SoCs.
+    /// It can even drive the internal pins that are not drawn out.
+    /// Before you operate, you must be clear about what you are doing.
+    /// </remarks>
+    public unsafe partial class SunxiDriver : GpioDriver
+    {
+        private IntPtr _gpioPointer0;
+        private IntPtr _gpioPointer1;
+
+        /// <summary>
+        /// CPUX-PORT base address.
+        /// </summary>
+        protected virtual int GpioRegisterOffset0 { get; }
+        /// <summary>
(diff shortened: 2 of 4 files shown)


joperezr:
Thanks a lot for your contribution @ZhangGaoxing I'm sorry that we haven't had time to look into this yet but we have been backlogged a bit with work lately but we will soon take a look. Again thanks a lot for the contribution!

krwq:
@ZhangGaoxing I've talked offline with @joperezr and we think the new drivers should be added to Iot.Device.Bindings (also the namespace) because we currently have no way of testing those boards and System.Device.Gpio should be only for stuff we support. Once we get those boards in the future we can move them to System.Device.Gpio - sorry for late feedback on that. I'll take a look at your PR right now to see if we have any extra feedback.

src/System.Device.Gpio/System/Device/Gpio/Drivers/OrangePiLite2Driver.Linux.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Device.Gpio.Drivers
```
krwq: We can create devices/Drivers directory for all of the community driven drivers - we'll eventually try to catch up with our lab and move them to System.Device.Gpio though

src/System.Device.Gpio/System/Device/Gpio/Drivers/SunxiDriver.Linux.cs
```diff
@@ -0,0 +1,470 @@
...
+        private readonly IDictionary<int, PinMode> _sysFSModes = new Dictionary<int, PinMode>();
+
+        protected SunxiDriver() { }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SunxiDriver"/>.
+        /// </summary>
+        /// <param name="gpioRegisterOffset0">CPUX-PORT base address (This can be find in the corresponding SoC datasheet).</param>
+        /// <param name="gpioRegisterOffset1">CPUS-PORT base address (This can be find in the corresponding SoC datasheet).</param>
+        public SunxiDriver(int gpioRegisterOffset0, int gpioRegisterOffset1)
```
krwq: nit: might be worth matching the name with params description (i.e.: `cpuxPortBaseAddess` etc)

src/System.Device.Gpio/System/Device/Gpio/Drivers/SunxiDriver.Linux.cs
```diff
@@ -0,0 +1,470 @@
...
+            int cfgShift = unmapped.Port % 8;
+            int pulNum = unmapped.Port / 16;
+            int pulShift = unmapped.Port % 16;
+
+            // Get register address, register pointer
+            int cfgAddress, pulAddress;
+            uint* cfgPointer, pulPointer;
+            if (unmapped.PortController < 10)
+            {
+                cfgAddress = (GpioRegisterOffset0 + unmapped.PortController * 0x24 + cfgNum * 0x04) & _mapMask;
```
krwq: nit: might be worth to move `unmapped.PortController * 0x24 + cfgNum * 0x04) & _mapMask` and expression below to a local

src/System.Device.Gpio/System/Device/Gpio/Drivers/SunxiDriver.Linux.cs
```diff
@@ -0,0 +1,470 @@
...
+        }
+
+        /// <summary>
+        /// Writes a value to a pin.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        /// <param name="value">The value to be written to the pin.</param>
+        protected internal override void Write(int pinNumber, PinValue value)
+        {
+            var unmapped = UnmapPinNumber(pinNumber);
```
krwq: nit: since the output type is not obvious please use explicit type

src/System.Device.Gpio/System/Device/Gpio/Drivers/SunxiDriver.Linux.cs
```diff
@@ -0,0 +1,470 @@
...
+        /// <returns>The value of the pin.</returns>
+        protected internal unsafe override PinValue Read(int pinNumber)
+        {
+            var unmapped = UnmapPinNumber(pinNumber);
+
+            int dataAddress;
+            uint* dataPointer;
+            if (unmapped.PortController < 10)
+            {
+                dataAddress = (GpioRegisterOffset0 + unmapped.PortController * 0x24 + 0x10) & _mapMask;
```
krwq: feels like there should be some utility like GetPinMemoryAddress or similar

src/System.Device.Gpio/System/Device/Gpio/Drivers/SunxiDriver.Linux.cs
```diff
@@ -0,0 +1,470 @@
...
+            int portController = (pinNumber - port) / 32;
+
+            return (portController, port);
+        }
+
+        private static int MapPortController(char portController)
+        {
+            return portController switch
+            {
+                'A' => 0,
```
krwq: consider just subtracting:

`portController - 'A'`
and add range check

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -30,6 +31,22 @@ public GpioController(PinNumberingScheme numberingScheme)
         /// <returns>A driver that works with the board the program is executing on.</returns>
         private static GpioDriver GetBestDriverForBoard()
         {
+            try
```
krwq: We cannot have this logic yet, might be worth adding some utility in your driver for that, i.e.:
`SunxiDriver.GetBestDriver()` or possibly take model info as an input.

Some extensibility model for best driver would be a better approach but we should solve this as separate issue

krwq:
Drivers code looks good to me with couple of tiny nits

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -30,6 +31,22 @@ public GpioController(PinNumberingScheme numberingScheme)
...
+            {
+                string modelInfo = File.ReadAllText(ModelInfoPath);
```
joperezr: This probably won't matter since as @krwq suggested we want to move this to IoT.Device.Bindings package for now, but in the future I think we don't want to have the logic of GetBestDriverForBoard to change for each new PCB we add. We only did it for the RPi because that is the most used board, and fall back to the Generic one but in reality what we would want is to keep it just like that, and then for each new PCB support we add we will require people to manually select this as the driver they use, like: `var controller = new GpioController(numberingScheme, new OrangePiZeroDriver());`

krwq: I think we should eventually have some extensibility model there but I wouldn't add it yet until we have more boards because we don't have enough data to design APIs sanely yet

lippertmarkus:
@ZhangGaoxing any chance to finish that PR?

mguinness:
Is the likelihood this will get merged if @ZhangGaoxing is able to make final changes?

ropstah:
@ZhangGaoxing it would be really great if your work can be merged!

Ellerbach:
@nzdorovylo I see you are making changes on this PR. If you want to carry over this work, please feel free!

Issue #767, opened 2019-10-04
Measuring power with GPIO


MaxMommersteeg:
Does dotnet/iot currently support a device binding for measuring power? I'm looking for something that is comparable to INA219. If not I am willing to do some research on how to implement it myself.

krwq:
No, we don't support it at the moment but this device seems to have only 6 registers and uses I2C so I expect adding support should be relatively simple. Please let me know if you need any help.

Frankenslag:
@krwq 
I would be happy to take on the INA219. Would you like to assign me ?

krwq:
Thanks for picking this up @Frankenslag!

Frankenslag:
Closing this as now merged.

PR #781, opened 2019-10-09
Add INA219 Binding

Frankenslag:
Implements #767 

@MaxMommersteeg @krwq 

Here is the proposed binding for the INA219 Voltage/Current/Power sensor.

As it is essentially an ADC I have used the ADC namespace.

Hope it is as expected.


src/devices/Ina219/samples/README.md
```diff
@@ -0,0 +1,30 @@
+﻿# INA219 Sample
+
+This sample connects to a Raspberry Pi via the first I2c device. Please ensure that I2c is enabled by using raspi-config or add the following line to the /boot/Config.txt file
+
```
MarkCiliaVincenti: In the comments, please use the naming convention I2C rather than I2c.

src/devices/Ina219/samples/README.md
```diff
@@ -0,0 +1,30 @@
...
+This sample uses an Adafruit INA219 breakout board with a Raspberry Pi and monitors a LED wired into the 3.3 volts supply with a 150 ohm current limiting resistor. It prints the bus voltage, shunt voltage, current and power every second.
+
+![](Ina219.Sample_bb.png)
+
+    const byte Adafruit_Ina219_I2cAddress = 0x40;
+    const byte Adafruit_Ina219_I2cBus = 0x1;
+
+    // create an Ina219 device on I2c bus 1 addressing channel 64
+    using (Ina219 device = new Ina219(new I2cConnectionSettings(Adafruit_Ina219_I2cBus, Adafruit_Ina219_I2cAddress)))
```
MarkCiliaVincenti: Same thing here, should say I2C and INA219 in the comment.

src/devices/Ina219/samples/Ina219.Sample.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    public class Program
+    {
+        static void Main(string[] args)
+        {
+            const byte Adafruit_Ina219_I2cAddress = 0x40;
+            const byte Adafruit_Ina219_I2cBus = 0x1;
+
+            // create an Ina219 device on I2c bus 1 addressing channel 64
+            using (Ina219 device = new Ina219(new I2cConnectionSettings(Adafruit_Ina219_I2cBus, Adafruit_Ina219_I2cAddress)))
```
MarkCiliaVincenti: Please use I2C and INA219 in the inline comments.

src/devices/Ina219/samples/Ina219.Sample.cs
```diff
@@ -0,0 +1,37 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Device.I2c;
+using Iot.Device;
+
+namespace Iot.Device.Adc.Samples
+{
+    /// <summary>
+    /// Sample program for Ina219
+    /// </summary>
```
MarkCiliaVincenti: Please write INA219 instead of Ina219 in the comments.

src/devices/Ina219/samples/Ina219.Sample.cs
```diff
@@ -0,0 +1,37 @@
...
+            {
+                // reset the device 
+                device.Reset();
+
+                // set the calibration to have a +/- 100mA range
+                device.SetCalibration(Ina219.PgaSensitivity.PlusOrMinus40mv, 0.1F);
+                while (true)
+                {
+                    // write out the current values from the Ina219 device.
+                    System.Console.WriteLine($"Bus Voltage {device.GetBusVoltage()}V Shunt Voltage {device.GetShuntVoltage() * 1000}mV Current {device.GetCurrent() * 1000}mA Power {device.GetPower() * 1000}mW");
```
MarkCiliaVincenti: Please use INA219 in the comments.

src/devices/Ina219/samples/README.md
```diff
@@ -0,0 +1,30 @@
...
+    {
+        // reset the device 
+        device.Reset();
+
+        // set the calibration to have a +/- 100mA range
+        device.SetCalibration(Ina219.PgaSensitivity.PlusOrMinus40mv, 0.1F);
+        while (true)
+        {
+            // write out the current values from the Ina219 device.
+            System.Console.WriteLine($"Bus Voltage {device.GetBusVoltage()}V Shunt Voltage {device.GetShuntVoltage() * 1000}mV Current {device.GetCurrent() * 1000}mA Power {device.GetPower() * 1000}mW");
```
MarkCiliaVincenti: INA219 not Ina219 in the comments please.

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+            PowerLsb = CurrentLsb = 1F;
+            _calibrationValue = 0;
+
+            _disposeI2cDevice = true;
+            _i2cDevice = I2cDevice.Create(settings);
+        }
+
+        /// <summary>
+        /// Reset the Ina219 to default values;
+        /// </summary>
```
MarkCiliaVincenti: Please use INA219 in this comment.

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+            _calibrationValue = 0;
+
+            // Get the current values for the Bus and Shunt ADC settings.
+            _busAdcResSamp = (AdcResolutionOrSamples)((regValue & (ushort)ConfigurationFlags.BadcMask) >> 4);
+            _shuntAdcResSamp = (AdcResolutionOrSamples)(regValue & (ushort)ConfigurationFlags.SadcMask);
+        }
+
+        /// <summary>
+        /// Set the Ina219 operating mode.
+        /// </summary>
```
MarkCiliaVincenti: Please use INA219 in this comment

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+
+            regValue &= (ushort)ConfigurationFlags.ModeMask ^ 0xFFFF;
+            regValue |= (ushort)mode;
+
+            WriteRegister(Register.Configuration, regValue);
+        }
+
+        /// <summary>
+        /// Set the Ina219 Bus voltage range.
+        /// </summary>
```
MarkCiliaVincenti: Please use INA219 in this comment.

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+
+            regValue &= (ushort)ConfigurationFlags.BrngMask ^ 0xFFFF;
+            regValue |= (ushort)range;
+
+            WriteRegister(Register.Configuration, regValue);
+        }
+
+        /// <summary>
+        /// Set the Ina219 Programable Gain Amplifier range.
+        /// </summary>
```
MarkCiliaVincenti: Please use INA219 in this comment.

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+
+            regValue &= (ushort)ConfigurationFlags.PgaMask ^ 0xFFFF;
+            regValue |= (ushort)sensitivity;
+
+            WriteRegister(Register.Configuration, regValue);
+        }
+
+        /// <summary>
+        /// Set the Ina219 ADC resolution or samples to be used when reading the Bus voltage.
+        /// </summary>
```
MarkCiliaVincenti: Please use INA219 in this comment.

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+            regValue |= (ushort)((ushort)resorsamp << 4);
+
+            WriteRegister(Register.Configuration, regValue);
+
+            _busAdcResSamp = resorsamp;
+        }
+
+        /// <summary>
+        /// Set the Ina219 ADC resolution or samples to be used when reading the Shunt voltage.
+        /// </summary>
```
MarkCiliaVincenti: Please use INA219 in this comment.

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+            regValue |= (ushort)resorsamp;
+
+            WriteRegister(Register.Configuration, regValue);
+
+            _shuntAdcResSamp = resorsamp;
+        }
+
+        /// <summary>
+        /// Set the Ina219 calibration value used to scale the Shunt voltage into a Current reading.
+        /// </summary>
```
MarkCiliaVincenti: Please use INA219 in this comment.

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+        /// </remarks>
+        /// <param name="calibrationValue">The calibration value to switch to.</param>
+        public void SetCalibration(ushort calibrationValue)
+        {
+            _calibrationValue = calibrationValue;
+            WriteRegister(Register.Calibration, calibrationValue);
+        }
+
+        /// <summary>
+        /// Set the Ina219 calibration values used to scale the Shunt voltage into a Current reading based on the sensitivity of the
```
MarkCiliaVincenti: Please use INA219 in this comment.

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+        {
+            Span<byte> buffer = stackalloc byte[3];
+
+            // set the first byte of the buffer to the register to be writtent
+            buffer[0] = (byte)register;
+
+            // write the value to be written to the second and third bytes in big-endian order.
+            BinaryPrimitives.WriteUInt16BigEndian(buffer.Slice(1, 2), value);
+
+            // write the value to the register via the I2c Bus.
```
MarkCiliaVincenti: Please use I2C in this comment.

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+    /// 
+    /// The INA219 is a high-side current shunt and power monitor with an I2C interface.
+    /// The INA219 monitors both shunt drop and supply voltage, with programmable conversion
+    /// times and filtering. A programmable calibration value, combined with an internal multiplier,
+    /// enables direct readouts in amperes. An additional multiplying register calculates power in watts.
+    /// <see href="http://www.ti.com/lit/ds/symlink/ina219.pdf"/>
+    /// </summary>
+    public partial class Ina219 : IDisposable
+    {
+        private readonly Dictionary<PgaSensitivity, float> _sensitivityVoltage = new Dictionary<PgaSensitivity, float>() { {PgaSensitivity.PlusOrMinus40mv, 0.04F }, { PgaSensitivity.PlusOrMinus80mv, 0.08F }, { PgaSensitivity.PlusOrMinus160mv, 0.16F }, { PgaSensitivity.PlusOrMinus320mv, 0.32F } };
```
krwq: please split these over multiple lines since it's hard to read

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+        /// The CurrentLsb property specifies the electrical current in amperes respresented by the least significant bit of the INA219 Current register.
+        /// It is used by the GetCurrent method to calculate the current from the value exposed by the current register
+        /// <see cref="GetCurrent"/>
+        /// </summary>
+        /// <remarks>
+        /// This property defaults to 1. It is set by calling SetCalibration with values for the ADC sensitivity, current range and optionally shunt resistance.
+        /// If the value of the calibration register is set manually then this property will also need to be set for GetCurrent to return the correct value.
+        /// <see cref="SetCalibration(PgaSensitivity, float, float)"/> <seealso cref="SetCalibration(ushort)"/>
+        /// </remarks>
+        public float CurrentLsb { get; set; }
```
krwq: Lsb should not be in the name since it's transparent to user. I'd go with name something like `CurrentVoltage`.

Also it probably should not be settable (private set is ok)

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+        /// The PowerLsb property specifies the electrical power in watts respresented by the least significant bit of the INA219 Power register.
+        /// It is used by the GetPower method to calculate the power from the value exposed by the Power register
+        /// <see cref="GetCurrent"/>
+        /// </summary>
+        /// <remarks>
+        /// This property defaults to 1. It is set by calling SetCalibration with values for the ADC sensitivity, current range and optionally shunt resistance.
+        /// If the value of the calibration register is set manually then this property will also need to be set for GetPower to return the correct value.
+        /// <see cref="SetCalibration(PgaSensitivity, float, float)"/> <seealso cref="SetCalibration(ushort)"/>
+        /// </remarks>
+        public float PowerLsb { get; set; }
```
krwq: ditto

src/devices/Ina219/Ina219.cs
krwq: Other thing is that doc should mention the unit, i.e. V / mV / relative to Vcc

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+        }
+
+        /// <summary>
+        /// Construct an Ina219 device using an I2cConnectionSettings.
+        /// </summary>
+        /// <remarks>
+        /// This binding creates an I2cDevice ufor communication with the Ina219. The I2cDevice is disposed when then Ina219 is disposed.
+        /// </remarks>
+        /// <param name="settings">The I2cConnectionSettings object initialized with the appropiate settings to communicate with the Ina219.</param>
+        public Ina219(I2cConnectionSettings settings)
```
krwq: we usually don't add such overloads in other bindings but if you're adding one at least please implement it similarly to following:
```csharp
public Ina219(I2cConnectionSettings settings) : this(I2cDevice.Create(settings)
{
}
```

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+        public Ina219(I2cDevice i2cDevice)
+        {
+            if (i2cDevice == null)
+            {
+                throw new System.ArgumentNullException(nameof(i2cDevice));
+            }
+
+            PowerLsb = CurrentLsb = 1F;
+
+            _i2cDevice = i2cDevice;
```
krwq: `_calibrationValue = 0`?

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+
+        /// <summary>
+        /// Set the Ina219 operating mode.
+        /// </summary>
+        /// <param name="mode">The operating mode to switch to.</param>
+        public void SetOperatingMode(OperatingMode mode)
+        {
+            ushort regValue = ReadRegister(Register.Configuration);
+
+            regValue &= (ushort)ConfigurationFlags.ModeMask ^ 0xFFFF;
```
krwq: double checking: did you mean to flip bits here or did you mean `& 0xFFFF`

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+            regValue |= (ushort)mode;
+
+            WriteRegister(Register.Configuration, regValue);
+        }
+
+        /// <summary>
+        /// Set the Ina219 Bus voltage range.
+        /// </summary>
+        /// <param name="range">The voltage range to switch to.</param>
+        public void SetBusVoltageRange(BusVoltageRange range)
```
krwq: I'd make this a property (or make this private and call it from property)

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+        }
+
+        /// <summary>
+        /// Set the Ina219 Programable Gain Amplifier range.
+        /// </summary>
+        /// <remarks>
+        /// This is the voltage range used to read the shunt voltage. It can be one of +/-40mV, +/-80mV, +/-160mV or +/-320mV.
+        /// </remarks>
+        /// <param name="sensitivity">The voltage range of the Programable Gain Amplifier to switch to.</param>
+        public void SetPGARange(PgaSensitivity sensitivity)
```
krwq: ditto

krwq: same for other stuff which just sets/reads register (method applies if we do any waiting or if value which you read changes between calls)

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This can either by the number of bits used for the ADC conversion (9-12 bits) or the number of samples at 12 bits to be averaged for the result.
+        /// </remarks>
+        /// <param name="resorsamp">The ADC resolution or sample count to switch to.</param>
+        public void SetBusAdcResolutionOrSamples(AdcResolutionOrSamples resorsamp)
+        {
+            ushort regValue = ReadRegister(Register.Configuration);
+
+            regValue &= (ushort)ConfigurationFlags.BadcMask ^ 0xFFFF;
```
krwq: should you ve `&` the mask instead of flipping bits?

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+            _calibrationValue = calibrationValue;
+            WriteRegister(Register.Calibration, calibrationValue);
+        }
+
+        /// <summary>
+        /// Set the Ina219 calibration values used to scale the Shunt voltage into a Current reading based on the sensitivity of the
+        /// Progmamable Gain Amplifier, the required full scale current and the value of the shunt resistor.
+        /// </summary>
+        /// <remarks>
+        /// This method sets the calibration register and also the PowerLsb and CurrentLsb properties.
```
krwq: this should explain what will change from user perspective and when should people use it and how and not implementation detail

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+        /// <see cref="PowerLsb"/><seealso cref="CurrentLsb"/>
+        /// </remarks>
+        /// <param name="sensitivity">The sensitivity of the shunt Programable Gain Amplifier</param>
+        /// <param name="fsCurrent">The required full scale current value</param>
+        /// <param name="shuntResistance">The value of the shunt resistor. This defaults to 0.1 Ohms as used in the Adafruit breakout board.</param>
+        public void SetCalibration(PgaSensitivity sensitivity, float fsCurrent, float shuntResistance = 0.1F)
+        {
+            int i;
+            float minCurrentLsb = fsCurrent / 32767;
```
krwq: where does this math in this function coming from? (please datasheet page and link if it's not there) I can see this is slightly over complicated but I wanted double check this for correctness.

I.e. I know:

```csharp
while (i * currentStep < minCurrentLsb) { i++; };
CurrentLsb = i * currentStep;
```

will be equivalent to something along the lines of (not tested):

```csharp
CurrentLsb = currentStep * Math.Floor(minCurrentLsb / currentStep);
```

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,375 @@
...
+        /// <summary>
+        /// Get the calculated current through the INA219.
+        /// </summary>
+        /// <remarks>
+        /// This value is determined by an internal calculation using the calibration register and the read shunt voltage and then scaled
+        /// using the CurrentLsb property.
+        /// <see cref="CurrentLsb"/>
+        /// </remarks>
+        /// <returns>The calculated current in Amperes</returns>
+        public float GetCurrent()
```
krwq: I think this should be called `ReadCurrent`, property should be removed and `SetCalibration` should be private and likely renamed (considering context this is used at I doubt this is calibrating anything - it's most likely adjusting output to human readable units)

src/devices/Ina219/Ina219Enum.cs
```diff
@@ -0,0 +1,123 @@
...
+using System;
+
+namespace Iot.Device.Adc
+{
+    public partial class Ina219
+    {
+        /// <summary>
+        /// An enumeration representting the registers exposed by the INA219 device via I2c.
+        /// </summary>
+        protected enum Register : byte
```
krwq: 1. Anything which is not in the public APIs should be private - if it needs to be shared with other sensors make it internal if possible, otherwise make it public
2. Do not nest types under the classes

src/devices/Ina219/samples/README.md
```diff
@@ -0,0 +1,30 @@
+﻿# INA219 Sample
+
+This sample connects to a Raspberry Pi via the first I2c device. Please ensure that I2c is enabled by using raspi-config or add the following line to the /boot/Config.txt file
```
krwq: I'd avoid using Raspberry PI specific sentences since they will probably work on any board supporting I2C.

src/devices/Ina219/Ina219.cs
Frankenslag: I’m anding with the inverse of the mask

src/devices/Ina219/Ina219.cs
Frankenslag: And with the inverse of the mask.

src/devices/Ina219/Ina219.cs
Frankenslag: Youre right will fix

src/devices/Ina219/Ina219.cs
Frankenslag: I like the idea of a constructor  where you want the binding to manage the lifetime of the i2cdevice and one where it’s managed internally, but I take your point with the implementation.

src/devices/Ina219/Ina219.cs
Frankenslag: don’t really agree as it is the amount of current represented by the least significant bit of the current register. Perhaps it’s CurrentLsbVoltage. It is not private as the users has the option of using the default calibration by calling set calibration with the full scale current value or setting the values directly to optimal values for the applications and components installed. In essence the current calibration is a combination Of the calibration register and the CurrentLsb property. If you wanted less fine grained control then you could just use the calibration register but the data sheet recommends setting the calibration register to the number which gives the best add resolution and then scaling using CurrentLsb. Hope this makes some sense

src/devices/Ina219/Ina219.cs
Frankenslag: Doesn’t this violate the property vs function guideline. I.e. 
Takes some time and so is more than a field access and you may want an asynchronous version in a future version.
Can throw an exception etc.

src/devices/Ina219/Ina219.cs
Frankenslag: I believe this is ‘waiting’ as it will take time to send the data down the I2c bus at a particular clock speed. And you may want an async version. If it was io mapped io or memory mapped io then I could see why a property would be right.

src/devices/Ina219/Ina219.cs
krwq: If this is just partial sensor reading I'd prefer if we made this private and instead exposed some API which allows you to read the actual current

src/devices/Ina219/Ina219.cs
krwq: If the time is just I2C communication this is fine. If there is some extra delays or some heavier work then this should stay as function

src/devices/Ina219/Ina219.cs
krwq: Here is a good example of method. This is doing an actual operation which might take some longer time in some cases (i.e. sensor is still initializing etc). In other cases you're just changing internal registers. This case also even if timing is small returns potentially different value each time it's called without any state change and this is 100% case where we should use method

src/devices/Ina219/Ina219.cs
Frankenslag: It’s a partial calibration value used alongside the calibration register and so it may need to be user set. I can see how I could make the functionality part of the set calibration method. Sometime I just need to step back from the issue.

src/devices/Ina219/Ina219.cs
krwq: so this will flip instead :smile: i.e. `0b1010 ^ 0b1111` will become `0b0101` instead of zeros

maloo: seems functionally correct, but why not just use ~ if you want to flip all bits?

Frankenslag: @maloo 
You are right of course.

src/devices/Ina219/Ina219.cs
Frankenslag: @krwq 

I had a look at this but because I wanted to add error checking to ensure that the passed in settings were non-null then the device would have been created before the check was performed.

I have added a private pseudo constructor to ensure that code isn't duplicated

src/devices/Ina219/Ina219.cs
Frankenslag: I have removed the auto calibration method and detailed an example of calibration in the sample

src/devices/Ina219/Ina219.cs
Frankenslag: @krwq 

I have pushed a new version of the INA219 binding and hopefully addressed your comments.

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,372 @@
...
+    {
+        private I2cDevice _i2cDevice;
+        private bool _disposeI2cDevice = false;
+        private ushort _calibrationValue;
+        private Ina219AdcResolutionOrSamples _busAdcResSamp;
+        private Ina219AdcResolutionOrSamples _shuntAdcResSamp;
+        private float _currentLsb;
+
+        // These are possitive ranges in volts associated with each gain setting of the Programable Gain Controller.
+        private readonly Dictionary<Ina219PgaSensitivity, float> _sensitivityVoltage = new Dictionary<Ina219PgaSensitivity, float>()
```
krwq: if this is not being modified this can be safely static (if it's modified then it won't be thread safe)

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,372 @@
...
+        { 
+            { Ina219PgaSensitivity.PlusOrMinus40mv, 0.04F },
+            { Ina219PgaSensitivity.PlusOrMinus80mv, 0.08F },
+            { Ina219PgaSensitivity.PlusOrMinus160mv, 0.16F },
+            { Ina219PgaSensitivity.PlusOrMinus320mv, 0.32F }
+        };
+
+        // These values are the datasheet defined delays in micro seconds between requesting a Current or Power value from the INA219 and the ADC sampling having completed
+        // along with any conversions.
+        private readonly Dictionary<Ina219AdcResolutionOrSamples, int> _readDelays = new Dictionary<Ina219AdcResolutionOrSamples, int>()
```
krwq: ditto (note: static variables are names as i.e. `s_readDelays`)

src/devices/Ina219/Ina219.cs
```diff
@@ -0,0 +1,372 @@
...
+            { Ina219AdcResolutionOrSamples.Adc32Sample, 17020 },
+            { Ina219AdcResolutionOrSamples.Adc64Sample, 34050 },
+            { Ina219AdcResolutionOrSamples.Adc128Sample, 68100 }
+        };
+
+        /// <summary>
+        /// Method to initialize values during device construction
+        /// </summary>
+        /// <param name="i2cDevice"></param>        
+        private void _Ina219(I2cDevice i2cDevice)
```
krwq: call it `Initialize`, we usually don't use underscores in the method names

src/devices/Ina219/Ina219.cs
Frankenslag: Actually not being used so removed.

src/devices/Ina219/Ina219.cs
Frankenslag: done

src/devices/Ina219/Ina219.cs
Frankenslag: @krwq 

All suggested changes done and PR updated.

Frankenslag:
@krwq @joperezr 

Looks like the PR for #781 INA219 has failed some checks for reasons that are not obvious to me. How can I re-run the checks

pgrawehr:
Question on the namespace use: This uses Iot.Devices.Adc while the ADS1115, which is a pure ADC, uses Iot.Device.Ads1115 as namespace (same as class name!) is this a bug there or what's the reason for that?

Frankenslag:
> Question on the namespace use: This uses Iot.Devices.Adc while the ADS1115, which is a pure ADC, uses Iot.Device.Ads1115 as namespace (same as class name!) is this a bug there or what's the reason for that?

Whilst I was working on the mcp adc family I was asked to use the ADC namespace. As this binding is really a specialised Adc I figured it should also go into the same namespace.

pgrawehr:
Yes, I fully agree on the use of the ADC namespace. My question was really about whether the _other_ namespace is right or should be changed. I'm asking because I'm preparing a PR for improving the ADS1115 binding.

pgrawehr:
Uh, that doesn't work :-(

joperezr:
Looks like a test failed, which is what caused the build to fail. Doesn't look related to your change. I just restarted the build.

src/devices/Ina219/samples/README.md
```diff
@@ -0,0 +1,36 @@
...
+* The bus voltage range can be either 16v or 32v. As this example uses a 3.3v supply then the 16v bus voltage range is chosen
+* The current through the LED is in the low tens of milliamps. If we take 50mA as a reasonable maximum current that we may want to see then the maximum voltage accross the shunt resistor is 0.1 Ohms x 50mA which works out
+at 5mV. Given this we can use a shunt voltage range of +/- 40mV
+* The maximum possible current would then be 40mV / 0.1 = 400mA
+* With a 400mA maximum current and a range of the ADC of 15bits then the LSB of the current would be 400mA/32767 = 12.2207 microamps. We will chose 12.2uA as a round number.
+* From the [INA219 Datasheet](http://www.ti.com/lit/ds/symlink/ina219.pdf) the calibration register should be set at 0.04096/(currentLSB * shunt resistance) = 33574 = 0x8326
+
+![](Ina219.Sample_bb.png)
+
+    const byte Adafruit_Ina219_I2cAddress = 0x40;
```
krwq: nit: please wrap this in (remove spaces between backticks)

```
` ` `csharp
code
` ` `
```

Frankenslag:
> LGTM with tiny nit

OK attended to the nit in the latest PR update.

krwq:
Thank you @Frankenslag!

PR #783, opened 2019-10-09
Added a weather helper to make temperature, humidity and pressure calculations

MarkCiliaVincenti:
Covers the remainder of #752.

Streamlined multiple samples.


MarkCiliaVincenti:
I also have some Unit tests but I'm not sure where to put them.

````
        [Test]
        public void Test_HeatIndex()
        {
            var heatIndex = WeatherHelper.HeatIndex(Temperature.FromCelsius(30), 70);
            Assert.AreEqual(Math.Round(heatIndex.Celsius), 35);
        }

        [Test]
        public void Test_SummerSimmerIndex()
        {
            var summerSimmerIndex = WeatherHelper.SummerSimmerIndex(Temperature.FromFahrenheit(10), 8);
            Assert.AreEqual(Math.Round(summerSimmerIndex.Fahrenheit, 2), 11.06);
        }

        [Test]
        public void Test_SaturatedVaporPressure()
        {
            var saturatedVaporPressure = WeatherHelper.SaturatedVaporPressure(Temperature.FromCelsius(30));
            Assert.AreEqual(Math.Round(saturatedVaporPressure.Pascal, 0), 4232);
        }

        [Test]
        public void Test_ActualVaporPressure()
        {
            var actualVaporPressure = WeatherHelper.ActualVaporPressure(Temperature.FromCelsius(30), 25);
            Assert.AreEqual(Math.Round(actualVaporPressure.Pascal, 0), 1058);
        }

        [Test]
        public void Test_DewPoint()
        {
            var dewPoint = WeatherHelper.DewPoint(Temperature.FromFahrenheit(100), 50);
            Assert.AreEqual(Math.Round(dewPoint.Fahrenheit, 0), 78);
        }

        [Test]
        public void Test_AbsoluteHumidity()
        {
            var absoluteHumidity = WeatherHelper.AbsoluteHumidity(Temperature.FromFahrenheit(100), 50);
            Assert.AreEqual(Math.Round(absoluteHumidity, 1), 22.7);
        }
````

**Comment by krwq** [comment]
@MarkCiliaVincenti for tests please just create tests folder similar to the one here:
https://github.com/dotnet/iot/tree/master/src/devices/Units

File: src/devices/Common/Iot/Device/Common/WeatherHelper.cs [hunk 1]
```diff
@@ -0,0 +1,102 @@
...
+        /// <returns>The heat index, also known as the apparent temperature</returns>
+        public static Temperature HeatIndex(Temperature airTemperature, double relativeHumidity)
+        {
+            double tf = airTemperature.Fahrenheit;
+            double rh = relativeHumidity;
+            return Temperature.FromFahrenheit((-42.379)
+                + (2.04901523 * tf)
+                + (10.14333127 * rh)
+                - (0.22475541 * tf * rh)
+                - (6.83783 * Math.Pow(10, -3) * Math.Pow(tf, 2))
```
krwq: Some of these seem to be repeating (i.e. `Math.Pow(tf, 2)`) - you might want to create local variables to store those

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        /// <summary>
+        /// The summer simmer index is used to measure the amount of discomfort during the summer months
+        /// when heat and humidity often combine to make it feel hotter than it actually is. The summer
+        /// simmer index is usually used for overnight low temperatures. Use the heat index instead for
+        /// afternoon high temperatures.
+        /// </summary>
+        /// <param name="airTemperature">The dry air temperature</param>
+        /// <param name="relativeHumidity">The relative humidity (RH) expressed as a percentage</param>
```
krwq:
```suggestion
        /// <param name="relativeHumidity">The relative humidity (rH%) expressed as a percentage</param>
```

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// The heat index (or apparent temperature) is used to measure the amount of discomfort
+        /// during the summer months when heat and humidity often combine to make it feel hotter
+        /// than it actually is. The heat index is usually used for afternoon high temperatures.
+        /// Use the summer simmer index instead for overnight low temperatures.
+        /// </summary>
+        /// <param name="airTemperature">The dry air temperature</param>
+        /// <param name="relativeHumidity">The relative humidity (RH) expressed as a percentage</param>
+        /// <returns>The heat index, also known as the apparent temperature</returns>
+        public static Temperature HeatIndex(Temperature airTemperature, double relativeHumidity)
+        {
```
krwq: for all of these, please add the sources telling where the formulas come from

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: The compiler will take care of that, I believe. I'm following formulas I found and are generally accepted. The more I deviate from the standard, the harder it becomes for people to follow.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
krwq: per offline conversation RH is actually correct

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: Done

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: Done

MarkCiliaVincenti:
> @MarkCiliaVincenti for tests please just create tests folder similar to the one here:
> https://github.com/dotnet/iot/tree/master/src/devices/Units

I'm a bit confused about what the .csproj should look like exactly.

README.md
```diff
@@ -4,26 +4,26 @@
 
 This repository contains the [System.Device.Gpio](https://www.nuget.org/packages/System.Device.Gpio) library and implementations for various boards like [Raspberry Pi](https://www.raspberrypi.org/) and [Hummingboard](https://www.solid-run.com/nxp-family/hummingboard/).
 
-The repository also contains [IoT.Device.Bindings](https://www.nuget.org/packages/Iot.Device.Bindings), a growing set of community-maintained [device bindings](src/devices/README.md) for IoT components.
+The repository also contains [Iot.Device.Bindings](https://www.nuget.org/packages/Iot.Device.Bindings), a growing set of community-maintained [device bindings](src/devices/README.md) for IoT components.
```
krwq: this is fine in this PR but in the future try to avoid changing stuff unrelated to the issue you're fixing for couple of reasons:
- it's much easier to port/revert changes to stable branches
- it makes it easier to review since changeset is isolated
- you're more likely to get better review since it's easier to see the problem you're solving

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Iot.Units;
+using System;
+
+namespace Iot.Device.Common
+{
+    /// <summary>
+    /// Helpers for weather. Formulas taken from http://www.reahvac.com/tools/humidity-formulas/.
```
krwq: Link should not be in the xml docs since this will be eventually displayed on the docs page and it's not exactly interesting for users and rather for maintainers. Move the link anywhere but xml docs :smile:

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// <summary>
+        /// The heat index (or apparent temperature) is used to measure the amount of discomfort
+        /// during the summer months when heat and humidity often combine to make it feel hotter
+        /// than it actually is. The heat index is usually used for afternoon high temperatures.
+        /// Use the summer simmer index instead for overnight low temperatures.
+        /// </summary>
+        /// <param name="airTemperature">The dry air temperature</param>
+        /// <param name="relativeHumidity">The relative humidity (RH) expressed as a percentage</param>
+        /// <returns>The heat index, also known as the apparent temperature</returns>
+        public static Temperature HeatIndex(Temperature airTemperature, double relativeHumidity)
```
krwq: Since we talked about discoverability offline it might make sense to create a second class and use this there as an extension method, i.e.:

```csharp
public static  class WeatherExtensions
{
    public static Temperature HeatIndex(this Temperature airTemperature, double relativeHumidity)
        => WeatherHelpers.HeatIndex(airTemperature, relativeHumidity);
}
```

this should improve discoverability

krwq: although you can keep this as is since I'm not sure it actually will improve it considering you have to use the namespace and therefore already know about it

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+            double rh = relativeHumidity;
+            double tf2 = Math.Pow(tf, 2);
+            double rh2 = Math.Pow(rh, 2);
+            return Temperature.FromFahrenheit((-42.379)
+                + (2.04901523 * tf)
+                + (10.14333127 * rh)
+                - (0.22475541 * tf * rh)
+                - (6.83783 * Math.Pow(10, -3) * tf2)
+                - (5.481717 * Math.Pow(10, -2) * rh2)
+                + (1.22874 * Math.Pow(10, -3) * tf2 * (rh))
```
krwq: use 0.001 directly since it's also shorter and clearer (or `1e-3` which is equivalent to above notation)

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// afternoon high temperatures.
+        /// </summary>
+        /// <param name="airTemperature">The dry air temperature</param>
+        /// <param name="relativeHumidity">The relative humidity (RH) expressed as a percentage</param>
+        /// <returns>The summer simmer index</returns>
+        public static Temperature SummerSimmerIndex(Temperature airTemperature, double relativeHumidity)
+        {
+            double tf = airTemperature.Fahrenheit;
+            double rh = relativeHumidity;
+            return Temperature.FromFahrenheit(1.98 * (tf - ((0.55 - (0.0055 * rh)) * (tf - 58.0))) - 56.83);
```
krwq: I'm surprised this is calculated from Fahrenheit since most of the scientific articles use Kelvin or Celsius. Please double check this is correct.

src/devices/Common/Iot/Device/Common/tests/WeatherTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Iot.Device.Common.Tests
+{
+    public class WeatherTests
+    {
+        [Fact]
+        void HeatIndex()
```
krwq: 1. This (and other) tests will not run because they're not public
2. Please use name explaining what the test is testing, i.e. `HeatIndexIsCalculatedCorrectly`
3. Consider changing this specific test into Theory so you can test multiple values, i.e.:

[Theory]
[InlineData(35, 30, 70)] // repeat for each test case, add min 3 samples - if there are no exact values just check output within +/- 0.001 range
public void HeatIndexIsCalculatedCorrectly(double expected, double celsius, double relativeHumidity)
{
  ...
}

src/devices/Common/Iot/Device/Common/tests/WeatherTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            var heatIndex = WeatherHelper.HeatIndex(Temperature.FromCelsius(30), 70);
+            Assert.AreEqual(Math.Round(heatIndex.Celsius), 35);
+        }
+
+        [Fact]
+        void SummerSimmerIndex()
```
krwq: similar comment for this and other tests

krwq:
@MarkCiliaVincenti LGTM after fixing test comments

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: Done.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: Done.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: When I created and used the unit tests, I put the values from online calculators before writing the formulas (TDD). The tests passed.

src/devices/Common/Iot/Device/Common/tests/WeatherTests.cs
MarkCiliaVincenti: Done

MarkCiliaVincenti:
Note to self: I need to add methods to get the pressure from given sea-level pressure, temperature and altitude with overloaded methods (mean sea-level pressure of 1013.25 hPa and temperature of 15 C)

MarkCiliaVincenti:
> Note to self: I need to add methods to get the pressure from given sea-level pressure, temperature and altitude with overloaded methods (mean sea-level pressure of 1013.25 hPa and temperature of 15 C). This is used in Bmp180. Similarly we can get temperature from known altitude, pressure and sea-level pressure.

Done.

src/devices/Common/tests/Common.Tests.csproj
```diff
@@ -14,6 +14,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Units\Units.csproj" />
+    <Compile Include="WeatherTests.cs" />
```
krwq: this shouldn't be needed since "*.cs" will be included by default I think

MarkCiliaVincenti:
Depends on #789. I made a test that should fail, but it's not failing. Until that fails we can't be sure that all unit tests are working.

MarkCiliaVincenti:
I confirm all tests are passing now. From my end you can go ahead and merge to master, @krwq

krwq:
(restarted CI)

src/devices/Bmp180/Bmp180.cs
```diff
@@ -91,36 +92,39 @@ public Pressure ReadPressure()
         ///  Sea-level pressure
         /// </param>
         /// <returns>
-        ///  Height in meters from the sensor
+        ///  Height in metres from the sensor
```
krwq: it was correct before. We use US spelling in the docs

src/devices/Bmp180/Bmp180.cs
```diff
@@ -91,36 +92,39 @@ public Pressure ReadPressure()
...
-            return 44330.0 * (1.0 - Math.Pow((ReadPressure().Pascal / seaLevelPressure.Pascal), (1.0 / 5.255)));
+            return WeatherHelper.CalculateAltitude(ReadPressure(), seaLevelPressure, ReadTemperature());
         }
         
         /// <summary>
         ///  Calculates the altitude in meters from the mean sea-level pressure.
         /// </summary>
         /// <returns>
-        ///  Height in meters from the sensor
+        ///  Height in metres from the sensor
```
krwq: ditto

src/devices/Bmp180/Bmp180.cs
MarkCiliaVincenti: You're right. Fixed.

src/devices/Bmp180/Bmp180.cs
MarkCiliaVincenti: Fixed.

src/devices/Bmxx80/README.md
```diff
@@ -17,7 +17,7 @@ The implementation supports the following devices:
 
 3 examples on how to use this device binding are available in the [samples](samples) folder.
 
-The following fritzing diagram illustrates one way to wire up the BMP280 with a Raspberry Pi using I2C:
+The following fritzing diagram illustrates one way to wire up the BMP280 with a Raspberry Pi using I²C:
```
krwq: FYI: I'm fine with this in the docs but will push back on similar changes to `.cs`

src/devices/Bmxx80/samples/Bme280.sample.cs
```diff
@@ -69,13 +77,20 @@ static void Main(string[] args)
...
-                    Console.WriteLine($"Humidity: {humValue} %");
+                    altValue = WeatherHelper.CalculateAltitude(preValue, defaultSeaLevelPressure, tempValue);
+                    
+                    Console.WriteLine($"Temperature: {tempValue.Celsius:0.#}\u00B0C");                    
+                    Console.WriteLine($"Pressure: {preValue.Hectopascal:0.##}hPa");                    
+                    Console.WriteLine($"Altitude: {altValue:0.##}m");                    
+                    Console.WriteLine($"Relative humidity: {humValue:0.#}%");
+                    Console.WriteLine($"Heat index: {WeatherHelper.CalculateHeatIndex(tempValue, humValue).Celsius} \u00B0C");
+                    Console.WriteLine($"Summer simmer index: {WeatherHelper.CalculateSummerSimmerIndex(tempValue, humValue).Celsius} \u00B0C");
+                    Console.WriteLine($"Saturated vapor pressure: {WeatherHelper.CalculateSaturatedVaporPressure(tempValue).Hectopascal} hPa");
```
krwq:
```suggestion
                    Console.WriteLine($"Saturated vapor pressure: {WeatherHelper.CalculateSaturatedVaporPressure(tempValue).Hectopascal}hPa");
```

for consistency

src/devices/Bmxx80/samples/Bme280.sample.cs
```diff
@@ -69,13 +77,20 @@ static void Main(string[] args)
...
+                    Console.WriteLine($"Actual vapor pressure: {WeatherHelper.CalculateActualVaporPressure(tempValue, humValue).Hectopascal} hPa");
```
krwq:
```suggestion
                    Console.WriteLine($"Actual vapor pressure: {WeatherHelper.CalculateActualVaporPressure(tempValue, humValue).Hectopascal}hPa");
```

src/devices/Bmxx80/samples/Bme280.sample.cs
```diff
@@ -69,13 +77,20 @@ static void Main(string[] args)
...
+                    Console.WriteLine($"Dew point: {WeatherHelper.CalculateDewPoint(tempValue, humValue).Celsius} \u00B0C");
```
krwq: can't tell looking at github diff but if there is a space before unit please remove (same below)

src/devices/Bmxx80/samples/Bme680.sample.cs
```diff
@@ -24,6 +26,8 @@ static void Main()
 
             // The I2C bus ID on the Raspberry Pi 3.
             const int busId = 1;
+            //set this to the current sea level pressure in the area for correct altitude readings
```
krwq:
```suggestion
            // set this to the current sea level pressure in the area for correct altitude readings
```

src/devices/Bmxx80/samples/Bme680.sample.cs
```diff
@@ -45,12 +49,23 @@ static void Main()
...
+                        var altValue = WeatherHelper.CalculateAltitude(preValue, defaultSeaLevelPressure, tempValue);
+                    
+                        Console.WriteLine($"Gas resistance: {gasResistance:0.##}Ohm");
+                        Console.WriteLine($"Temperature: {tempValue.Celsius:0.#}\u00B0C");
+                        Console.WriteLine($"Pressure: {preValue.Hectopascal:0.##}hPa");                    
+                        Console.WriteLine($"Altitude: {altValue:0.##}m");                    
+                        Console.WriteLine($"Relative humidity: {humValue:0.#}%");
+                        Console.WriteLine($"Heat index: {WeatherHelper.CalculateHeatIndex(tempValue, humValue).Celsius} \u00B0C");
+                        Console.WriteLine($"Summer simmer index: {WeatherHelper.CalculateSummerSimmerIndex(tempValue, humValue).Celsius} \u00B0C");
+                        Console.WriteLine($"Saturated vapor pressure: {WeatherHelper.CalculateSaturatedVaporPressure(tempValue).Hectopascal} hPa");
```
krwq: similar comment about spaces before unit (here and above and below)

src/devices/Bmxx80/README.md
MarkCiliaVincenti: Had missed that while reverting the squared symbol. Thanks.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// <param name="relativeHumidity">The relative humidity (RH) expressed as a percentage</param>
+        /// <returns>The heat index, also known as the apparent temperature</returns>
+        public static Temperature CalculateHeatIndex(Temperature airTemperature, double relativeHumidity)
+        {
+            double tf = airTemperature.Fahrenheit;
+            double rh = relativeHumidity;
+            double tf2 = Math.Pow(tf, 2);
+            double rh2 = Math.Pow(rh, 2);
+
+            var steadman = 0.5 * (tf + 61 + ((tf - 68) * 1.2) + (rh * 0.094));
```
krwq: 1. use `double` instead of var
2. This formula looks like it could be simplified. If it's the form used in the sources leave this as is (for clarity), if not then please simplify. If you decide to simplify but this form is in the sources please add a comment that this is simplified

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            if (steadman < 80)
```
krwq: out of curiosity: is this function continuous?

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+                return Temperature.FromFahrenheit(steadman);
+
+            var rothfuszRegression = (-42.379)
+                + (2.04901523 * tf)
+                + (10.14333127 * rh)
+                - (0.22475541 * tf * rh)
+                - (6.83783 * 0.001 * tf2)
+                - (5.481717 * 0.01 * rh2)
+                + (1.22874 * 0.001 * tf2 * (rh))
```
krwq:
```suggestion
                + (1.22874 * 0.001 * tf2 * rh)
```

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+            double tf2 = Math.Pow(tf, 2);
+            double rh2 = Math.Pow(rh, 2);
+
+            var steadman = 0.5 * (tf + 61 + ((tf - 68) * 1.2) + (rh * 0.094));
+
+            if (steadman < 80)
+                return Temperature.FromFahrenheit(steadman);
+
+            var rothfuszRegression = (-42.379)
+                + (2.04901523 * tf)
```
krwq: can this formula be simplified?

src/devices/Bmxx80/samples/Bme680.sample.cs
MarkCiliaVincenti: Fixed all.

joperezr:
Hey @MarkCiliaVincenti, I haven't yet taken a look into your PR yet, but is there a specific reason why you add so many commits to the PRs? This makes it such that it is really hard to review, plus it is actually costing money since we are triggering builds for every commit in your PR. Is there any chance that you can try batching your changes so that you aim to have less than 10 commits per PR?

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: Changed to use `double`.
As for part 2, it's in the form used in the sources.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: It's explained in https://www.wpc.ncep.noaa.gov/html/heatindex_equation.shtml.

> In practice, the simple formula is computed first and the result averaged with the temperature. If this heat index value is 80 degrees F or higher, the full regression equation along with any adjustment as described above is applied.
> The Rothfusz regression is not valid for extreme temperature and relative humidity conditions beyond the range of data considered by Steadman.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: Am following the actual linked formulas; it should be better to leave as is for clarity.

MarkCiliaVincenti:
> Hey @MarkCiliaVincenti, I haven't yet taken a look into your PR yet, but is there a specific reason why you add so many commits to the PRs? This makes it such that it is really hard to review, plus it is actually costing money since we are triggering builds for every commit in your PR. Is there any chance that you can try batching your changes so that you aim to have less than 10 commits per PR?

I'm editing online from the browser for the most part, that's why. Apologies will try to use VS more often :)

src/devices/Bmxx80/samples/Bme280.sample.cs
MarkCiliaVincenti: Done.

src/devices/Bmxx80/samples/Bme280.sample.cs
MarkCiliaVincenti: Done.

krwq:
@MarkCiliaVincenti you can create a second branch in your fork and open a PR against your other branch on your fork and then squash your changes to this branch whenever the batch is complete. This way you can still edit in the browser but we will see only 1 commit

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// Calculates the dew point.
+        /// </summary>
+        /// <param name="airTemperature">The dry air temperature</param>
+        /// <param name="relativeHumidity">The relative humidity (RH) expressed as a percentage</param>
+        /// <returns>The dew point</returns>
+        public static Temperature CalculateDewPoint(Temperature airTemperature, double relativeHumidity)
+        {
+            var avp = CalculateActualVaporPressure(airTemperature, relativeHumidity).Hectopascal;
+            var lavp = Math.Log(avp);
+            return Temperature.FromCelsius(((-430.22 + (237.7 * lavp)) / (19.08 - lavp)));
```
krwq:
```suggestion
            return Temperature.FromCelsius((-430.22 + (237.7 * lavp)) / (19.08 - lavp));
```

MarkCiliaVincenti: Good catch.

src/devices/Hts221/Hts221.csproj
```diff
@@ -12,6 +12,7 @@
       <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
     </ProjectReference>
     <ProjectReference Include="..\Units\Units.csproj" />
+    <ProjectReference Include="..\Common\Iot\Device\Common\CommonHelpers.csproj" />
```
krwq: Is this actually using CommonHelpers or is only sample project using it?

src/devices/Dhtxx/Dhtxx.csproj
```diff
@@ -22,7 +22,8 @@
       <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
     </ProjectReference>
     <ProjectReference Include="..\Units\Units.csproj" />
+    <ProjectReference Include="..\Common\Iot\Device\Common\CommonHelpers.csproj" />
```
krwq: Should this be in the sample project?

src/devices/Lps25h/Lps25h.csproj
```diff
@@ -12,6 +12,7 @@
       <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
     </ProjectReference>
     <ProjectReference Include="..\Units\Units.csproj" />
+    <ProjectReference Include="..\Common\Iot\Device\Common\CommonHelpers.csproj" />
```
krwq: should this be in the sample?

src/devices/Lps25h/samples/Lps25h.Sample.cs
```diff
@@ -12,14 +14,23 @@ internal class Program
     {
         // I2C address on SenseHat board
         public const int I2cAddress = 0x5c;
-
+
```
krwq: remove extra spaces

src/devices/Hts221/Hts221.csproj
MarkCiliaVincenti: The sample project, but it might use it in future. Is this a problem?

src/devices/Rtc/Rtc.csproj
```diff
@@ -10,7 +10,7 @@
       <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
     </ProjectReference>
     <ProjectReference Include="..\Units\Units.csproj" />
-    <Compile Include="..\Common\Iot\Device\Common\NumberHelper.cs" />
+    <ProjectReference Include="..\Common\Iot\Device\Common\CommonHelpers.csproj" />
```
krwq: should this be in the sample?

src/devices/Rtc/Devices/Pcf8563/Pcf8563.cs
```diff
@@ -2,9 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Iot.Device.Common;
```
krwq: unnecessary change. This is only ok if you're doing any other functional changes

src/devices/Rtc/Devices/Ds3231/Ds3231.cs
```diff
@@ -2,10 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System;
-using System.Device.I2c;
 using Iot.Device.Common;
 using Iot.Units;
+using System;
```
krwq: please revert

src/devices/Rtc/Devices/Ds1307/Ds1307.cs
```diff
@@ -2,9 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Iot.Device.Common;
```
krwq: please revert

src/devices/Rtc/Rtc.csproj
MarkCiliaVincenti: No, the Rtc was actually using NumberHelper but now I made it use the ProjectReference instead.

src/devices/SenseHat/samples/SenseHat.Sample.cs
```diff
@@ -40,12 +45,29 @@ public static void Main(string[] args)
                     sh.Fill(n % 2 == 0 ? Color.DarkBlue : Color.DarkRed);
                     sh.SetPixel(x, y, Color.Yellow);
 
-                    Console.WriteLine($"Temperature: Sensor1: {sh.Temperature.Celsius} °C   Sensor2: {sh.Temperature2.Celsius} °C");
-                    Console.WriteLine($"Humidity: {sh.Humidity} %rH");
-                    Console.WriteLine($"Pressure: {sh.Pressure} hPa");
-                    Console.WriteLine($"Acceleration: {sh.Acceleration} g");
-                    Console.WriteLine($"Angular rate: {sh.AngularRate} DPS");
-                    Console.WriteLine($"Magnetic induction: {sh.MagneticInduction} gauss");
+                    var tempValue = sh.Temperature;
```
krwq: I'd revert most of the changes from here. This sample is showcasing SenseHat mainly and I feel like most of those weather helpers here are obscuring the usage a bit

src/devices/SenseHat/samples/SenseHat.Sample.cs
```diff
@@ -40,12 +45,29 @@ public static void Main(string[] args)
...
+                    Console.WriteLine($"Temperature Sensor 1: {tempValue.Celsius:0.#}\u00B0C");
+                    Console.WriteLine($"Temperature Sensor 2: {temp2Value.Celsius:0.#}\u00B0C");
+                    Console.WriteLine($"Pressure: {preValue:0.##}hPa");
+                    Console.WriteLine($"Altitude: {altValue:0.##}m");
+                    Console.WriteLine($"Acceleration: {sh.Acceleration}g");
+                    Console.WriteLine($"Angular rate: {sh.AngularRate}DPS");
+                    Console.WriteLine($"Magnetic induction: {sh.MagneticInduction}gauss");
+                    Console.WriteLine($"Relative humidity: {humValue:0.#}%");
+                    Console.WriteLine($"Heat index: {WeatherHelper.CalculateHeatIndex(tempValue, humValue).Celsius:0.#}\u00B0C");
+                    Console.WriteLine($"Summer simmer index: {WeatherHelper.CalculateSummerSimmerIndex(tempValue, humValue).Celsius:0.#}\u00B0C");
```
krwq: I'd show just 1 parameter from weather helpers to showcase the usage but not more to not obscure the usage

src/devices/SenseHat/samples/SenseHat.Sample.cs
MarkCiliaVincenti: Maybe we can separate them instead of eliminate them?

src/devices/Si7021/samples/Program.cs
```diff
@@ -19,8 +21,17 @@ static void Main(string[] args)
...
                 while (true)
                 {
-                    Console.WriteLine($"Temperature: {sensor.Temperature.Celsius}℃");
-                    Console.WriteLine($"Humidity: {sensor.Humidity}%");
+                    var tempValue = sensor.Temperature;
+                    var humValue = sensor.Humidity;
+                    
+                    Console.WriteLine($"Temperature: {tempValue.Celsius:0.#}\u00B0C");                    
+                    Console.WriteLine($"Relative humidity: {humValue:0.#}%");
+                    Console.WriteLine($"Heat index: {WeatherHelper.CalculateHeatIndex(tempValue, humValue).Celsius:0.#}\u00B0C");
```
krwq: Pick only 1 or 2 parameter per sample (for temperature and pressure sensors few more are fine but for the rest let's focus on showing the device and not everything)

src/devices/SenseHat/samples/SenseHat.Sample.cs
MarkCiliaVincenti: Again, maybe I should introduce a separator?

src/devices/Dhtxx/Dhtxx.csproj
MarkCiliaVincenti: I personally prefer it where it is, but will wait for your advice.

krwq: You should only reference what you use. I don't think this uses it directly. Samples are

src/devices/Rtc/Rtc.csproj
krwq: right

src/devices/SenseHat/samples/SenseHat.Sample.cs
krwq: I'd make it as crisp as possible so rather keeping just 1 so that people will explore the type if they need it but no more in case they are not interested to not distract from what we are trying to show

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
krwq: sounds good

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+                + (10.14333127 * rh)
+                - (0.22475541 * tf * rh)
+                - (6.83783 * 0.001 * tf2)
+                - (5.481717 * 0.01 * rh2)
+                + (1.22874 * 0.001 * tf2 * rh)
+                + (8.5282 * 0.0001 * tf * rh2)
+                - (1.99 * 0.000001 * tf2 * rh2);
+
+            if (rh < 13 && tf >= 80 && tf <= 112)
+            { // adjustment
```
krwq: this comment is not exactly helpful, I'd remove it and the one below (and any similar)

src/devices/Bmp180/Bmp180.cs
```diff
@@ -93,9 +94,10 @@ public Pressure ReadPressure()
         /// <returns>
         ///  Height in meters from the sensor
         /// </returns>
+        [ObsoleteAttribute("ReadAltitude is obsolete. Please use Iot.Device.Common.WeatherHelper instead.")]
```
joperezr: I wouldn't mark these guys as obsolete. It is fine if internally we call WeatherHelper ourselves, but it is still way easier to call `bmp180.ReadAltitud(SeaLevelPressure)` than calling `WeatherHelper.CalculateAltitude(ReadPressure(), seaLevelPressure, ReadTemperature());`

The same goes for all of the rest of the Obsolete attributes you added. The implementation change is great, but obsoleting all these members I don't agree with.

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -169,6 +170,7 @@ public override bool TryReadPressure(out Pressure pressure)
         /// Contains <see cref="double.NaN"/> otherwise.
         /// </param>
         /// <returns><code>true</code> if pressure measurement was not skipped, otherwise <code>false</code>.</returns>
+        [ObsoleteAttribute("TryReadAltitude is obsolete. Please use Iot.Device.Common.WeatherHelper instead.")]
```
joperezr: Same comment goes to here and all other of your Obsolete Attributes. The main idea of the device bindings is to make the api surface easy to use, and using the WeatherHelper requires some deeper knowledge and some extra codes from the consuming side in order to get right. Plus it looks like WeatherHelper doesn't have TryRead equivalents which you would want in case you wanted to obsolete this API. Same comment going forward, it's fine to change our inner implementation to use WeatherHelper, but don't obsolete the current API, and keep the samples using the device bindings' API.

src/devices/Common/Iot/Device/Common/NumberHelper.cs
```diff
@@ -9,7 +9,7 @@ namespace Iot.Device.Common
     /// <summary>
     /// Helpers for number.
     /// </summary>
-    internal static class NumberHelper
+    public class NumberHelper
```
joperezr: Why did we remove static? Why did we change to public?

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+
+using Iot.Units;
+using System;
+
+namespace Iot.Device.Common
+{
+    /// <summary>
+    /// Helpers for weather
+    /// </summary>
+    public class WeatherHelper
```
joperezr: This should be a static class

src/devices/Bmp180/Bmp180.cs
MarkCiliaVincenti: The idea is that `ReadAltitude` will force reading the temperature and pressure from your sensors, the values of which you may already have available at hand. That is, if you want to display the temperature, pressure and altitude, using the WeatherHelper IS the way to go to avoid polling the sensors twice. The only time using ReadAltitude directly makes sense is when you only want to get the altitude and don't care for the actual readings. This is why I marked them as obsolete. Please let me know what you think now that I've explained the why.

src/devices/Bmp180/samples/Bmp180.Sample.cs
```diff
@@ -30,23 +31,23 @@ static void Main(string[] args)
...
                 //read values
                 Temperature tempValue = i2cBmp280.ReadTemperature();
-                Console.WriteLine($"Temperature {tempValue.Celsius} \u00B0C");
+                Console.WriteLine($"Temperature: {tempValue.Celsius:0.#}\u00B0C");
                 var preValue = i2cBmp280.ReadPressure();
-                Console.WriteLine($"Pressure {preValue.Hectopascal} hPa");
-                double altValue = i2cBmp280.ReadAltitude();
-                Console.WriteLine($"Altitude {altValue:0.##} m");
+                Console.WriteLine($"Pressure: {preValue.Hectopascal:0.##}hPa");
+                var altValue = WeatherHelper.CalculateAltitude(preValue, Pressure.MeanSeaLevel, tempValue);
```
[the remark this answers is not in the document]
MarkCiliaVincenti: Why would you want the sample to read the temperature and pressure twice instead of once? These sensors were intended for mobile application where energy usage is required to be very low.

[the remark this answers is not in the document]
joperezr: The samples are made to showcase the usage of the API of that device binding, which is why I want to restore the binding specific methods instead of using a completely different class here since that would defeat the purpose of the samples.

src/devices/Bmp180/Bmp180.cs
joperezr: I get that, and I agree that the most optimum code in most cases would be to get the Altitude through the WeatherHelper, but my point is that performance and CPU is not the ONLY goal of our APIs. One of the other goals (and just as important as perf) is the ease of use, which is why I think that we should provide that option as well. We have a lot of examples in the framework as well where there are two ways of doing the exact same thing, One is very easy to use, and the other is harder but more performant, and we care about both. One example is writing text to a file, you could do it through File.WriteAllText(path, text) which is super easy to use but not super performant, or you could also do it through opening a stream and managing the writing and closing of the stream yourself. That would be more performant, but that doesn't mean that we want to remove or obsolete the WriteAllText API.

src/devices/Bmp180/samples/Bmp180.Sample.cs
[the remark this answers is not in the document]
MarkCiliaVincenti: You're asking to showcase a very inefficient way to achieve the same result though. I know this is not in line with general Microsoft sample applications I've seen. Up to you though, but I strongly disagree with your suggestion.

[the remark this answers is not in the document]
joperezr: That is why I thought to have the code for running it using the the bme sensor, but adding a comment above it with something like
```c#
// Note that if you already have the pressure value and the temperature, you could also calculate altitude by 
// calling WeatherHelper.CalculateAltitude(preValue, Pressure.MeanSeaLevel, tempValue) which would be more performant.
double altValue = i2cBmp280.ReadAltitude();
```

src/devices/Bmp180/Bmp180.cs
MarkCiliaVincenti: OK I appreciate this (though the example you brought is unfair since this is essentially a 1:1 line of code substitution) and understand your fair comments. In actuality I hoped the tag would serve a second purpose which is to make developers already using it aware that there's now a more efficient alternative. I wish C# had a [Suggestion] attribute; it'd make more sense than an obsolete attribute.

src/devices/Bmp180/samples/Bmp180.Sample.cs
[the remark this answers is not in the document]
joperezr: That way, the samples are still showing how to consume this binding to perform operations, but if somebody cares about performance, they will read the comment and figure out that there is a faster way of doing it.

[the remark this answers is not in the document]
MarkCiliaVincenti: Fair enough :-)

src/devices/Bmp180/Bmp180.cs
joperezr: Right, I agree suggestion would be great, the problem with obsolete is that many people compile with `TreatWarningsAsErrors` so this would mean that they would have to use #pragma disable on their code in order to be able to call this method, and they might not really care about performance, so that is the part I don't like about obsolete.

src/devices/Common/Iot/Device/Common/NumberHelper.cs
MarkCiliaVincenti: I had problems with the `internal` and just copied and pasted what I had in `WeatherHelper`. No problems, I'll change it to static too.

RobinTTY:
@joperezr Regarding the ReadAltitude() method:
In #797 I changed the API surface as discussed in #753 so you only need to call Read() to get all measured values which makes it a lot easier to use (see updated samples). I didn't add altitude to the returned measured values since it's not something the sensor measures and just something you can derive from the pressure/temperature value. I also removed the old API surface (ReadTemperature(), ReadHumidity(), etc.).

Would you want to keep the old API surface (ReadTemperature(), ReadHumidity(), etc.) and just mark it as obsolete? Also I guess altitude should be one of the default "measured" values you get from calling Read()?

MarkCiliaVincenti:
Hmm, I would lazily compute anything not read by a sensor, and that holds for all that WeatherHelper currently does and will do in future.

RobinTTY:
@MarkCiliaVincenti yeah that's what I would have expected. Just using the WeatherHelper to calculate all other values the sensor doesn't directly measure. I was just going of your conversation with @joperezr.

I see some value in directly returning values you often use in applications like altitude but at the same time just calling WeatherHelper seems easy enough to me.

krwq:
I'd expect tiny math operations not matter that much vs i.e. allocation or transmission over the wire. I'd personally not cache it for simplicity and add cache when we find it being a bottle neck. As for the:

Read + Temperature/Pressure
vs
ReadTemperature/ReadPressure

I personally prefer the first option since you don't need to read twice from the sensor but you need to make sure to throw a reasonable exception (i.e. InvalidOperationException with some error) if Temperature/Pressure is accessed before first time reading.

RobinTTY:
@krwq Yeah the way I have it now is that you do:

```C#
// perform measurement
Bme280ReadResult measurement = bme280.Read();

// results are available after Read() completed
Console.WriteLine($"Temperature: {measurement.Temperature.Celsius} °C");
Console.WriteLine($"Pressure: {measurement.Pressure.Hectopascal} hPa");
Console.WriteLine($"Humidity: {measurement.Humidity} %");
```

So through this approach you can't even read the register before the reading is complete (still need to test this).

src/devices/Hts221/Hts221.csproj
MarkCiliaVincenti: Done (need to commit)

src/devices/Dhtxx/Dhtxx.csproj
MarkCiliaVincenti: Done (need to commit)

src/devices/Lps25h/Lps25h.csproj
MarkCiliaVincenti: Done (need to commit)

src/devices/Lps25h/samples/Lps25h.Sample.cs
MarkCiliaVincenti: I don't understand what you're seeing. As much as I tried to remove these invisible changes, VS2019 can't see any changes. This is not important, I'm resolving this.

src/devices/Rtc/Devices/Pcf8563/Pcf8563.cs
MarkCiliaVincenti: It's using NumberHelper, which is in Iot.Device.Common.

src/devices/Rtc/Devices/Ds3231/Ds3231.cs
MarkCiliaVincenti: Why? Without it, it can't resolve `DateTime`.

src/devices/Rtc/Devices/Ds1307/Ds1307.cs
MarkCiliaVincenti: This is needed for NumberHelper.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: Done (pending commit)

src/devices/Bmp180/Bmp180.cs
MarkCiliaVincenti: Done (pending commit)

src/devices/Bmxx80/Bmx280Base.cs
MarkCiliaVincenti: Done (pending commit)

src/devices/Bmp180/samples/Bmp180.Sample.cs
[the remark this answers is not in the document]
MarkCiliaVincenti: Done (pending commit)

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: Done (pending commit)

src/devices/Common/Iot/Device/Common/NumberHelper.cs
MarkCiliaVincenti: Done (pending commit)

src/devices/SenseHat/samples/SenseHat.Sample.cs
MarkCiliaVincenti: Done (pending commit)

src/devices/SenseHat/samples/SenseHat.Sample.cs
MarkCiliaVincenti: Done (pending commit)

src/devices/Si7021/samples/Program.cs
MarkCiliaVincenti: Done (pending commit)

MarkCiliaVincenti:
> I'd personally not cache it for simplicity and add cache when we find it being a bottle neck.

@krwq It's not exactly rocket science, but yes no harm in having it in a separate pull request in a future phase once @RobinTTY merge both of our pull requests.

krwq:
@joperezr want to take a look?

MarkCiliaVincenti:
> @joperezr want to take a look?

Ping @joperezr

src/devices/Bmp180/samples/Bmp180.Sample.cs
```diff
@@ -30,23 +31,27 @@ static void Main(string[] args)
...
                 //read values
                 tempValue = i2cBmp280.ReadTemperature();
-                Console.WriteLine($"Temperature {tempValue.Celsius} \u00B0C");
+                Console.WriteLine($"Temperature: {tempValue.Celsius:0.#}\u00B0C");
                 preValue = i2cBmp280.ReadPressure();
-                Console.WriteLine($"Pressure {preValue.Hectopascal} hPa");
-                altValue = i2cBmp280.ReadAltitude();
-                Console.WriteLine($"Altitude {altValue:0.##} m");
+                Console.WriteLine($"Pressure: {preValue.Hectopascal:0.##}hPa");
+                altValue = WeatherHelper.CalculateAltitude(preValue, Pressure.MeanSeaLevel, tempValue);
```
joperezr: NIT: seems like you forgot to change this one back to use i2cBmp280.ReadAltitude()

src/devices/Bmxx80/samples/Bme680.sample.cs
```diff
@@ -45,12 +49,21 @@ static void Main()
...
-                        bme680.TryReadTemperature(out var temperature);
-                        bme680.TryReadPressure(out var pressure);
-                        bme680.TryReadHumidity(out var humidity);
+                        bme680.TryReadTemperature(out var tempValue);
+                        bme680.TryReadPressure(out var preValue);
+                        bme680.TryReadHumidity(out var humValue);
                         bme680.TryReadGasResistance(out var gasResistance);
-
-                        Console.WriteLine($"{temperature.Celsius:N2} \u00B0C | {pressure.Hectopascal:N2} hPa | {humidity:N2} %rH | {gasResistance:N2} Ohm");
+                        var altValue = WeatherHelper.CalculateAltitude(preValue, defaultSeaLevelPressure, tempValue);
```
joperezr: does bme680 have a TryReadAltitude as well? If so I believe we should use that instead. If not, then I'm fine with keeping this as is

src/devices/Common/Iot/Device/Common/NumberHelper.cs
```diff
@@ -9,7 +9,7 @@ namespace Iot.Device.Common
     /// <summary>
     /// Helpers for number.
     /// </summary>
-    internal static class NumberHelper
+    public static class NumberHelper
```
joperezr: I think we should try to keep this internal if possible, since the name is something very generic and might clash with people that are trying to use similar types in their apps. Perhaps keep it as internal, and whoever requires it could still compile the file directly instead of referencing the project.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+            return Pressure.FromHectopascal(6.11 * Math.Pow(10, ((7.5 * tc) / (237.7 + tc))));
+        }
+
+        /// <summary>
+        /// Calculates the actual vapor pressure.
+        /// </summary>
+        /// <param name="airTemperature">The dry air temperature</param>
+        /// <param name="relativeHumidity">The relative humidity (RH) expressed as a percentage</param>
+        /// <returns>The actual vapor pressure</returns>
+        public static Pressure CalculateActualVaporPressure(Temperature airTemperature, double relativeHumidity)
```
joperezr: What does actual mean here? Does it mean current? or Relative? Not sure I understand the difference between this API and the previous one.

src/devices/Rtc/Rtc.csproj
joperezr: I would keep the Compile item here, and similar on the CreditCard project, that way you won't have to change the API from internal to public.

src/devices/Bmp180/samples/Bmp180.Sample.cs
MarkCiliaVincenti: Done (pending commit)

src/devices/Bmxx80/samples/Bme680.sample.cs
MarkCiliaVincenti: It doesn't. Will keep as is then. I believe #797 will take care of it.

src/devices/Common/Iot/Device/Common/NumberHelper.cs
MarkCiliaVincenti: I tried going for consistency, especially since it's possible for some projects to use both.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
MarkCiliaVincenti: "Actual vapor pressure" is a term. http://ww2010.atmos.uiuc.edu/(Gh)/guides/mtr/cld/dvlp/rh.rxml#targetText=Actual%20vapor%20pressure%20is%20a,amount%20of%20water%20vapor%20increases.&targetText=That%20means%2C%20an%20equal%20number,air%20back%20into%20the%20water.

src/devices/Rtc/Rtc.csproj
MarkCiliaVincenti: Done (pending commit)

MarkCiliaVincenti:
> Left a few comments but nothing blocking other than changing the NumberHelper class to be internal again. Once that is done, this LGTM, thanks a lot for doing all this work @MarkCiliaVincenti

Done :)

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
joperezr: I see, I didn't know that 😄

NIT: Any chance we can add that description in the summary so that it is more clear for people? so from that link you sent, it would be something like:
```c#
/// Calculates the actual vapor pressure, which is a measurement of the amount of water vapor in a volume of air and increases as the amount of water vapor increases.
```

If you think that most people using this method will know what actual vapor pressure it means then I'm totally fine keeping this as is, but if not I guess we could use the extra description here (I'm sorry for being picky 😞)

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+
+namespace Iot.Device.Common
+{
+    /// <summary>
+    /// Helpers for weather
+    /// </summary>
+    public static class WeatherHelper
+    {
+        #region TemperatureAndRelativeHumidity
+        // Formulas taken from http://www.reahvac.com/tools/humidity-formulas/
```
krwq: per offline conversation we cannot use this as a source because of its non-permissive license. The other link is ok to use. We need to find other source with permissive license in order to use it :disappointed:

MarkCiliaVincenti: Done.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -13,7 +13,7 @@ namespace Iot.Device.Common
     public static class WeatherHelper
     {
         #region TemperatureAndRelativeHumidity
-        // Formulas taken from https://www.aprweather.com/pages/calc.htm, https://www.wpc.ncep.noaa.gov/html/heatindex_equation.shtml and https://ncalculators.com/meteorology/summer-simmer-index-calculator.htm
+        // Formulas taken from https://www.aprweather.com/pages/calc.htm and https://www.wpc.ncep.noaa.gov/html/heatindex_equation.shtml
```
krwq: I'm waiting for the response from https://www.aprweather.com/pages/calc.htm owners with clarification on the license

MarkCiliaVincenti: Are we going to halt a pull request that took a lot of effort over a line of inline documentation? Can I take it off completely and then decide whether or not to add it in later? @krwq @joperezr

joperezr: I am sorry for the experience and really understand what you are going through, but unfortunately this is not just about the comment, it is about the code using the formulas taken from those sources. For cases like this we really have to wait for a green light from experts in order to be able to go ahead and merge your change. We really appreciate all of the work that has gone into this, and @krwq and I will work hard on trying to figure this out as soon as we can.

MarkCiliaVincenti: *bump* :)

terrajobst: @joperezr and @krwq are correct. Open source isn't a free for all. You can't just copy & paste code from random projects. The licenses have be to compatible. I'll check what our legal obligations are with respect to math formulas.

MarkCiliaVincenti: *bump* @krwq @joperezr @terrajobst

joperezr:
I have now pushed the changes on PR #1041 by @pgrawehr

MarkCiliaVincenti:
> I have now pushed the changes on PR #1041 by @pgrawehr

Great, are we ready to merge to master now?
I saw that the changes by @pgrawehr has removed the Summer Simmer Index. I have privately messaged @krwq with the LinkedIn account of the author of this formula, hoping that you guys could get his official permission to include this formula here. But let it not slow down the process; we could add it in as a new PR.

src/devices/Common/Iot/Device/Common/CommonHelpers.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
joperezr, line 1: I can't remember if we discussed the fact of not having this project here and instead just have the *cs files and compile them in the projects we need them in, like the rest of Common. That said, I can address that in a separate PR to no longer block this one.

joperezr:
Thanks a lot to both @MarkCiliaVincenti and @pgrawehr for your hard work and getting this done.

PR #785, opened 2019-10-09
Create Documentation for running the tests

pgrawehr:
- There was no documentation on what needs to be done to successfully run the unit tests. 
- The tests should be extended (i.e. to test the pull up resistor settings or the I2C interface)

### DIFF
--- a/src/System.Device.Gpio.Tests/README.md
+++ b/src/System.Device.Gpio.Tests/README.md
@@ -22,7 +22,7 @@ After that, you can run the tests with the RaspberryPiDriver (which is the defau
 ```
 dotnet test --filter RaspberryPiDriverTests System.Device.Gpio.sln 
 ```
-Depending on the version of the Pi and the installed linux version, it may be required to run the tests as root:
+Depending on the version of the Pi and the installed Linux distribution, it may be required to run the tests as root:
 ```
 sudo dotnet test --filter RaspberryPiDriverTests System.Device.Gpio.sln 
 ```
--- a/src/System.Device.Gpio.Tests/README.md
+++ b/src/System.Device.Gpio.Tests/README.md
@@ -10,15 +10,13 @@ cd src/System.Device.Gpio
 dotnet build System.Device.Gpio.sln
 ```
 This builds the main System.Device.Gpio assembly together with its test assembly. Before running the tests, you need to:
-- Connect an LED in series with a matching resistor to BCM Pin 18 (physical Pin 12) 
-- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable
+- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a 10kΩ resistor between them. 
 
 ## Raspberry Pi driver tests
 After that, you can run the tests with the RaspberryPiDriver (which is the default low-level driver for the Raspberry Pi) like:
 ```
 dotnet test --filter RaspberryPiDriverTests System.Device.Gpio.sln 
 ```
-During the tests, you should see the LED light up for a second. 
 
 If everything went smoothly, the output should end with a success message. 
 
--- a/src/System.Device.Gpio.Tests/README.md
+++ b/src/System.Device.Gpio.Tests/README.md
@@ -1,0 +1,39 @@
+# How to run tests on the Raspberry Pi
+
+Prerequisites: 
+- Installed .NET SDK with the "dotnet" executable in the path
+- Locally cloned repository
+
+Build System.Device.Gpio.dll:
+```
+cd src/System.Device.Gpio
+dotnet build System.Device.Gpio.sln
+```
+This builds the main System.Device.Gpio assembly together with its test assembly. Before running the tests, you need to:
+- Connect an LED in series with a matching resistor to BCM Pin 18 (physical Pin 12) 
+- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable
+
+## Raspberry Pi driver tests
+After that, you can run the tests with the RaspberryPiDriver (which is the default low-level driver for the Raspberry Pi) like:
+```
+dotnet test --filter RaspberryPiDriverTests System.Device.Gpio.sln 
+```
+During the tests, you should see the LED light up for a second. 
+
+If everything went smoothly, the output should end with a success message. 
+
+## SysFsDriver Tests
+You can also run the SysFsDriver Tests (this uses a more generic approach). This driver requires root permissions to run, so you need to `sudo` the command:
+```
+sudo dotnet test --filter SysFsDriverTests System.Device.Gpio.sln 
+```
+
+## LibgpiodDriver Tests
+To run the Libgpiod Driver test (a Linux Kernel Driver for the GPIO device), you need to first install the libgpiod package:
+```
+sudo apt install -y libgpiod-dev
+```
+These tests do not require root permissions to run:
+```
+dotnet test --filter LibGpiodDriverTests System.Device.Gpio.sln 
+```
--- a/src/System.Device.Gpio.Tests/README.md
+++ b/src/System.Device.Gpio.Tests/README.md
@@ -1,5 +1,10 @@
-# How to run tests on the Raspberry Pi
+# How to run the tests
+This shows how to run the tests on a Raspberry Pi. On other platforms, things should be similar. 
 
+## Building on the desktop PC
+(to be extended)
+
+## Building directly on the Pi
 Prerequisites: 
 - Installed .NET SDK with the "dotnet" executable in the path
 - Locally cloned repository
@@ -10,12 +15,16 @@ cd src/System.Device.Gpio
 dotnet build System.Device.Gpio.sln
 ```
 This builds the main System.Device.Gpio assembly together with its test assembly. Before running the tests, you need to:
-- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a 10kΩ resistor between them. 
+- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable. It is suggested to add a resistor between 1kΩ and 10kΩ between the pins, this protects the PI in the case of a missconfiguration (i.e both pins set to Out, one high, the other low). 
 
 ## Raspberry Pi driver tests
 After that, you can run the tests with the RaspberryPiDriver (which is the default low-level driver for the Raspberry Pi) like:
 ```
 dotnet test --filter RaspberryPiDriverTests System.Device.Gpio.sln 
+```
+Depending on the version of the Pi and the installed linux version, it may be required to run the tests as root:
+```
+sudo dotnet test --filter RaspberryPiDriverTests System.Device.Gpio.sln 
 ```
 
 If everything went smoothly, the output should end with a success message. 
--- a/src/System.Device.Gpio.Tests/README.md
+++ b/src/System.Device.Gpio.Tests/README.md
@@ -15,7 +15,7 @@ cd src/System.Device.Gpio
 dotnet build System.Device.Gpio.sln
 ```
 This builds the main System.Device.Gpio assembly together with its test assembly. Before running the tests, you need to:
-- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable. It is suggested to add a resistor between 1kΩ and 10kΩ between the pins, this protects the PI in the case of a missconfiguration (i.e both pins set to Out, one high, the other low). 
+- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable. It is suggested to add a 1kΩ to 10kΩ resistor between the pins; this protects the Pi in the case of a misconfiguration (i.e both pins set to Out, one high, the other low). 
 
 ## Raspberry Pi driver tests
 After that, you can run the tests with the RaspberryPiDriver (which is the default low-level driver for the Raspberry Pi) like:
--- a/src/System.Device.Gpio.Tests/README.md
+++ b/src/System.Device.Gpio.Tests/README.md
@@ -5,9 +5,14 @@ This shows how to run the tests on a Raspberry Pi. On other platforms, things sh
 (to be extended)
 
 ## Building directly on the Pi
-Prerequisites: 
-- Installed .NET SDK with the "dotnet" executable in the path
-- Locally cloned repository
+First, clone the repository on the pi: 
+```
+git clone https://github.com/dotnet/iot
+cd iot
+```
+Now you can 
+- Either download and install the NET Core SDK from https://get.dot.net/ (the Raspberry Pi with the default 32 Bit Raspbian Linux requires the ARM32 version) or 
+- run `./build.sh` in the checkout directory. This will install the correct NET Core SDK in the .dotnet subfolder of the working copy.
 
 Build System.Device.Gpio.dll:
 ```


src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,39 @@
+# How to run tests on the Raspberry Pi
+
+Prerequisites: 
+- Installed .NET SDK with the "dotnet" executable in the path
+- Locally cloned repository
+
+Build System.Device.Gpio.dll:
+```
+cd src/System.Device.Gpio
```
joperezr: Instead of this, could you simply tell people to build the whole repo by `build.cmd`/`build.sh` from the root? That will install all SDK dependencies needed in order to be able to run the tests.

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,39 @@
...
+dotnet build System.Device.Gpio.sln
+```
+This builds the main System.Device.Gpio assembly together with its test assembly. Before running the tests, you need to:
+- Connect an LED in series with a matching resistor to BCM Pin 18 (physical Pin 12) 
+- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable
```
joperezr: We also connect a resistor between the two pins to ensure no short circuiting and protect your board.

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Raspberry Pi driver tests
+After that, you can run the tests with the RaspberryPiDriver (which is the default low-level driver for the Raspberry Pi) like:
+```
+dotnet test --filter RaspberryPiDriverTests System.Device.Gpio.sln
```
joperezr: So these instructions are for  doing everything on the Pi? That might be a bit costly as building in the Pi is very slow. When we do local testing usually we just build in either Windows or Linux dev machine, and then we copy the build output into the Pi.

src/System.Device.Gpio.Tests/README.md
pgrawehr: This doesn't work for me, at least not on the Pi itself (On Windows, I wouldn't use it either, there I just fire up Visual Studio and run the build).


```
pi@raspberrypi:~/projects/iot $ ./build.sh 
 Wiederherstellung in "444.81 ms" für "/home/pi/.nuget/packages/microsoft.dotnet.arcade.sdk/5.0.0-beta.19501.4/tools/Tools.proj" abgeschlossen.
dotnet-install: .NET Core Runtime version 2.1.11 is already installed.
dotnet-install: Adding to current process PATH: `/home/pi/projects/iot/.dotnet`. Note: This change will be visible only when sourcing script.
dotnet-install: Installation finished successfully.
  Wiederherstellung in "30.04 ms" für "/home/pi/projects/iot/eng/common/internal/Tools.csproj" abgeschlossen.
  Wiederherstellung in "7.14 ms" für "/home/pi/projects/iot/samples/force-sensitive-resistor/force-sensitive-resistor.csproj" abgeschlossen.
  Wiederherstellung in "381.32 ms" für "/home/pi/projects/iot/src/Iot.Device.Bindings/Iot.Device.Bindings.csproj" abgeschlossen.
  Wiederherstellung in "513.39 ms" für "/home/pi/projects/iot/src/System.Device.Gpio/System.Device.Gpio.csproj" abgeschlossen.
  <snip many more package restore lines>
Wiederherstellung in "2.6 ms" für "/home/pi/projects/iot/src/System.Device.Gpio/System.Device.Gpio.csproj" abgeschlossen.
/home/pi/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error : Failed to load �/CD���, error: /home/pi/projects/iot/.dotnet/shared/Microsoft.NETCore.App/2.1.11/libhostpolicy.so: wrong ELF class: ELFCLASS64 [/home/pi/projects/iot/src/System.Device.Gpio/System.Device.Gpio.csproj]
/home/pi/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error : An error occurred while loading required library libhostpolicy.so from [/home/pi/projects/iot/.dotnet/shared/Microsoft.NETCore.App/2.1.11] [/home/pi/projects/iot/src/System.Device.Gpio/System.Device.Gpio.csproj]
/home/pi/.nuget/packages/microsoft.net.compilers.toolset/3.1.0-beta1-19127-06/tasks/netcoreapp2.1/Microsoft.CSharp.Core.targets(58,5): error : Failed to load ��, error: /home/pi/projects/iot/.dotnet/shared/Microsoft.NETCore.App/2.1.11/libhostpolicy.so: wrong ELF class: ELFCLASS64 [/home/pi/projects/iot/src/devices/Units/Units.csproj]
<snip many more similar error messages>
Not sure what's happening here, the Pi4 is running an ordinary Raspbian 32 Bit version. For some reason, the build script tries to link 64 Bit libraries.

File: src/System.Device.Gpio.Tests/README.md [hunk 4]
```diff
@@ -0,0 +1,39 @@
...
+- Installed .NET SDK with the "dotnet" executable in the path
+- Locally cloned repository
+
+Build System.Device.Gpio.dll:
+```
+cd src/System.Device.Gpio
+dotnet build System.Device.Gpio.sln
+```
+This builds the main System.Device.Gpio assembly together with its test assembly. Before running the tests, you need to:
+- Connect an LED in series with a matching resistor to BCM Pin 18 (physical Pin 12)
```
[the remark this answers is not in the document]
pgrawehr: I was wondering about that, too. It's nice to see that something is going on, but I agree that this is not a good test. I'll remove that, then.

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,39 @@
+# How to run tests on the Raspberry Pi
```
krwq: Overall comment:
- I'm currently in the process of changing these tests and the new version will include diagram what needs to be connected: https://github.com/dotnet/iot/pull/758
- We can merge these but expect I'll modify these to adjust for changes (I might add something to make it easier to run when people have incomplete setup)

src/System.Device.Gpio.Tests/README.md
pgrawehr: Yes, these instructions are for building directly on the Pi. Since running the tests needs to be done on the Pi anyway (for obvious reasons), I showed the most straight-forward way of getting there. There seem to be no documentation on how to build and deploy from Visual Studio as well (and that involves more steps and should be documented at the root level). 

Or did I miss a way of remotely executing the tests from the Desktop?

src/System.Device.Gpio.Tests/README.md
pgrawehr: I did this as a preparation for another commit, so I guess this can be merged and used as a starting point. 
Since not everybody will have all the hardware to run all the tests (i.e. the ADC converter), selecting tests will be important. Note also that the current input/output tests have no timeout defined, so they will block if the pins are not connected.

joperezr: > Note also that the current input/output tests have no timeout defined, so they will block if the pins are not connected.

For this we were relying on our testing environment which limits the time a test will be executed, and kill it (and mark it as failed obviously) if it is just hanging.

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,37 @@
...
```
krwq:
```suggestion
# How to run tests
```

I think this is not RPi specific

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Prerequisites: 
+- Installed .NET SDK with the "dotnet" executable in the path
```
krwq: I believe only runtime is required if you're not building on RPi directly

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,37 @@
...
+- Locally cloned repository
+
+Build System.Device.Gpio.dll:
+```
+cd src/System.Device.Gpio
+dotnet build System.Device.Gpio.sln
+```
+This builds the main System.Device.Gpio assembly together with its test assembly. Before running the tests, you need to:
+- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a 10kΩ resistor between them.
```
krwq: Technically speaking no resistor is required. Resistor is added so that it's not possible to short-circuit. Any resistor should be fine but I'd recommend anything higher than 1k and not too high so that internal pull up/down doesn't win

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,37 @@
...
+Build System.Device.Gpio.dll:
+```
+cd src/System.Device.Gpio
+dotnet build System.Device.Gpio.sln
+```
+This builds the main System.Device.Gpio assembly together with its test assembly. Before running the tests, you need to:
+- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a 10kΩ resistor between them. 
+
+## Raspberry Pi driver tests
+After that, you can run the tests with the RaspberryPiDriver (which is the default low-level driver for the Raspberry Pi) like:
```
krwq: this is the only part RPi specific

src/System.Device.Gpio.Tests/README.md
krwq: On any platform: `dotnet publish -r linux-arm`
On Pi or wherever you run you need to have runtime and you can: `dotnet xunit.console.runner.dll test.dll`
We currently don't have xunit runner dependency in the project but I'll add it when I'll be upgrading the setup

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+dotnet test --filter RaspberryPiDriverTests System.Device.Gpio.sln 
+```
+
+If everything went smoothly, the output should end with a success message. 
+
+## SysFsDriver Tests
+You can also run the SysFsDriver Tests (this uses a more generic approach). This driver requires root permissions to run, so you need to `sudo` the command:
+```
+sudo dotnet test --filter SysFsDriverTests System.Device.Gpio.sln
```
krwq: actually RPi driver also requires root, only libgpiod doesn't (at least that's what I was seeing)

krwq: (I'll add some xunit traits in my PR to make it easier to run without sudo)

src/System.Device.Gpio.Tests/README.md
pgrawehr: Relying on an external process to terminate the tests is not really a good design for errors that you expect to happen, I think. Especially since here it's easy to be fixed, as there are already CancellationTokens in place. I can have a look into that.

src/System.Device.Gpio.Tests/README.md
pgrawehr: Maybe stupid question, but which other platforms are supported? You need hardware with a GPIO header.

src/System.Device.Gpio.Tests/README.md
pgrawehr: Probably, but I'm showing how to build on the Pi a few lines down. And those who really build this project on their own will want the SDK installed.

src/System.Device.Gpio.Tests/README.md
pgrawehr: That's weird. After a few retries, it _sometimes_ works without root rights and sometimes I get an exception.

src/System.Device.Gpio.Tests/README.md
pgrawehr: What should I add? @joperezr above suggested to include the Resistor, I originally skipped it (which worked as well). Should I write something like: 
Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable. To protect from shortening Pins, a resistor between 1kΩ and 10kΩ should be added in the line.

pgrawehr: Note: See also https://raspberrypi.stackexchange.com/questions/104462/does-directly-connecting-pins-fry-the-pi which gives a good answer on what could happen.

src/System.Device.Gpio.Tests/README.md
pgrawehr: I guess we can update/extend the documentation later, then?

src/System.Device.Gpio.Tests/README.md
pgrawehr: I think we should provide the publish profiles for the most common build configurations. This simplifies building and deploying directly from with VS. I'd rather put that on a separate PR, though.

src/System.Device.Gpio.Tests/README.md
pgrawehr: That should be fixed on a separate PR, I think.

src/System.Device.Gpio.Tests/README.md
pgrawehr: @krwq: It works without root rights each time except the very first time after bootup. If I run it twice (without sudo) it works, and after that it always works. Any idea?

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,46 @@
...
+- Installed .NET SDK with the "dotnet" executable in the path
+- Locally cloned repository
+
+Build System.Device.Gpio.dll:
+```
+cd src/System.Device.Gpio
+dotnet build System.Device.Gpio.sln
+```
+This builds the main System.Device.Gpio assembly together with its test assembly. Before running the tests, you need to:
+- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable. It is suggested to add a resistor between 1kΩ and 10kΩ between the pins, this protects the PI in the case of a missconfiguration (i.e both pins set to Out, one high, the other low).
```
MarkCiliaVincenti:
```suggestion
- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable. It is suggested to add a resistor between 1kΩ and 10kΩ between the pins, this protects the Pi in the case of a missconfiguration (i.e both pins set to Out, one high, the other low). 
```

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+This builds the main System.Device.Gpio assembly together with its test assembly. Before running the tests, you need to:
+- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable. It is suggested to add a resistor between 1kΩ and 10kΩ between the pins, this protects the PI in the case of a missconfiguration (i.e both pins set to Out, one high, the other low). 
+
+## Raspberry Pi driver tests
+After that, you can run the tests with the RaspberryPiDriver (which is the default low-level driver for the Raspberry Pi) like:
+```
+dotnet test --filter RaspberryPiDriverTests System.Device.Gpio.sln 
+```
+Depending on the version of the Pi and the installed linux version, it may be required to run the tests as root:
```
MarkCiliaVincenti:
```suggestion
Depending on the version of the Pi and the installed Linux distribution, it may be required to run the tests as root:
```

src/System.Device.Gpio.Tests/README.md
MarkCiliaVincenti:
```suggestion
- Connect BCM Pins 12 and 16 (physical Pins 32 and 36) with a cable. It is suggested to add a 1kΩ to 10kΩ resistor between the pins; this protects the Pi in the case of a misconfiguration (i.e both pins set to Out, one high, the other low). 
```

pgrawehr:
I guess you squash anyway upon merge, so no need to combine the commits, right?

src/System.Device.Gpio.Tests/README.md
krwq: anything ARM32 (v7 I believe) / ARM64. I.e. Hummingboard and there is plenty other but they are not as popular as RPi

krwq:
@pgrawehr yes, we will squash

src/System.Device.Gpio.Tests/README.md
krwq: fair enough

src/System.Device.Gpio.Tests/README.md
krwq: @pgrawehr this is good info. It might be a driver bug though but can't tell without investigating deeper

src/System.Device.Gpio.Tests/README.md
krwq: Actually seems you've already done that

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,46 @@
+# How to run the tests
+This shows how to run the tests on a Raspberry Pi. On other platforms, things should be similar. 
+
+## Building on the desktop PC
+(to be extended)
+
+## Building directly on the Pi
+Prerequisites: 
+- Installed .NET SDK with the "dotnet" executable in the path
```
joperezr: NIT: Can we just add a link here to where to get dotnet  in order to install it? For example, maybe pointing people here: https://dotnet.microsoft.com/download/thank-you/dotnet-sdk-3.0.100-linux-arm32-binaries

src/System.Device.Gpio.Tests/README.md
```diff
@@ -0,0 +1,46 @@
...
+- Locally cloned repository
+
+Build System.Device.Gpio.dll:
+```
+cd src/System.Device.Gpio
+dotnet build System.Device.Gpio.sln
```
joperezr: I'm not 100% sure that this will always work, It might be that it is working now by chance, but the ideal workflow should be to just call /build.sh, or in the worst case, ./build.sh -restore which would install dotnet for you along with all of the dependencies that we require for building. Anyways, I'm fine with keeping this as is for now and just fix whatever issue you were hitting above when calling ./build.sh and once we do change these instructions.

src/System.Device.Gpio.Tests/README.md
joperezr: I agree that this might just be a bug in our side. RPi driver shouldn't need sudo because it uses /dev/gpiomem/ which only exposes the gpio registers so it doesn't require sudo.

src/System.Device.Gpio.Tests/README.md
krwq: @pgrawehr can you file an issue on that? (if possible can you temporarily switch language to english to make it easier for us to see the errors?)

src/System.Device.Gpio.Tests/README.md
krwq: I was also seeing locally it required sudo but I thought this was by design

src/System.Device.Gpio.Tests/README.md
krwq: Use: https://get.dot.net/ link

src/System.Device.Gpio.Tests/README.md
krwq: I always use dotnet build as well but I first do full build and force using dotnet from the `.dotnet` directory which got created

joperezr: yeah, that is supposed to work as well. But never calling build.sh, and just using a random dotnet cli most likely not work in the future, so I'd be happier if we could use the one that gets restored by our build instead.

src/System.Device.Gpio.Tests/README.md
pgrawehr: Yea, will do. I'll try to get a log in english.

src/System.Device.Gpio.Tests/README.md
pgrawehr: I think you get an error if the correct CLI version is not installed. I personally prefer having the CLI installed globally once instead of into each working copy. Maybe both methods should be described (once they work...)

krwq: @pgrawehr, in order to use the CLI which got created by the build when you also have a global installation you should also set environment variable: `export DOTNET_MULTILEVEL_LOOKUP=0` (or `set ...` on Windows) to use only framework next to the dotnet.exe and not try to search for other frameworks

src/System.Device.Gpio.Tests/README.md
pgrawehr: Added the link and changed documentation slightly to also include the preferred way of build (even though for now, that doesn't seem to work, see #805)

src/System.Device.Gpio.Tests/README.md
pgrawehr: @krwq : I've created #805 for this. An english log has been added there.

krwq:
Thanks @pgrawehr!

PR #788, opened 2019-10-11
Support setting the Pull up/Pull down resistor mode on the Pi4

pgrawehr:
The Pi4's CPU uses alternate registers to set the pull up/pull down mode than the Pi3 did. This adds a separate driver class for the Pi4 with (for now) one extra method to configure the registers for setting the pull mode of input registers. The required code was adapted from https://github.com/joan2937/pigpio/blob/master/pigpio.c#L8785. 

Unit test provided as well. 

The detection of the Pi4 is quite a hack, this should be improved later (see also PR #766). It would probably be better to either create a separate class for detecting the best driver or to have the drivers detect themselves whether they can run on the given hardware.

### DIFF
--- a/src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
+++ b/src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
@@ -1,4 +1,8 @@
-﻿using System.Device.Gpio.Drivers;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Device.Gpio.Drivers;
 using System.Threading;
 using Xunit;
 
@@ -31,7 +35,6 @@ namespace System.Device.Gpio.Tests
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), bestDriver))
             {
                 controller.OpenPin(OpenPin, PinMode.InputPullUp);
-                Thread.Sleep(10);
                 Assert.Equal(PinValue.High, controller.Read(OpenPin));
                 controller.SetPinMode(OpenPin, PinMode.InputPullDown);
                 Assert.Equal(PinValue.Low, controller.Read(OpenPin));
--- a/src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
+++ b/src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
@@ -1,39 +1,43 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Device.Gpio.Drivers;
-using System.Text;
+﻿using System.Device.Gpio.Drivers;
 using System.Threading;
 using Xunit;
 
 namespace System.Device.Gpio.Tests
 {
-	public class RaspberryPi4DriverTests : GpioControllerTestBase
-	{
+    public class RaspberryPi4DriverTests : GpioControllerTestBase
+    {
         /// <summary>
-		/// Leave this pin open (unconnected) for the tests
-		/// </summary>
-		private const int OpenPin = 23;
+        /// Leave this pin open (unconnected) for the tests
+        /// </summary>
+        private const int OpenPin = 23;
 
-		protected override GpioDriver GetTestDriver()
-		{
-			return new RaspberryPi4Driver();
-		}
+        protected override GpioDriver GetTestDriver()
+        {
+            return new RaspberryPi4Driver();
+        }
 
-		protected override PinNumberingScheme GetTestNumberingScheme() => PinNumberingScheme.Logical;
+        protected override PinNumberingScheme GetTestNumberingScheme() => PinNumberingScheme.Logical;
 
         [Fact]
         public void InputPullResistorsWork()
-		{
-			using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
-			{
-				controller.OpenPin(OpenPin, PinMode.InputPullUp);
-				Thread.Sleep(10);
-				Assert.Equal(PinValue.High, controller.Read(OpenPin));
-				controller.SetPinMode(OpenPin, PinMode.InputPullDown);
-				Assert.Equal(PinValue.Low, controller.Read(OpenPin));
-				controller.SetPinMode(OpenPin, PinMode.InputPullUp);
-				Assert.Equal(PinValue.High, controller.Read(OpenPin));
-			}
-		}
-	}
+        {
+            // This is a hack to test whether we're running on a Pi4 without duplicating that detection logic. 
+            var bestDriver = GpioController.GetBestDriverForBoard();
+            if (!(bestDriver is RaspberryPi4Driver))
+            {
+                bestDriver.Dispose();
+                return;
+            }
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), bestDriver))
+            {
+                controller.OpenPin(OpenPin, PinMode.InputPullUp);
+                Thread.Sleep(10);
+                Assert.Equal(PinValue.High, controller.Read(OpenPin));
+                controller.SetPinMode(OpenPin, PinMode.InputPullDown);
+                Assert.Equal(PinValue.Low, controller.Read(OpenPin));
+                controller.SetPinMode(OpenPin, PinMode.InputPullUp);
+                Assert.Equal(PinValue.High, controller.Read(OpenPin));
+            }
+        }
+    }
 }
--- a/src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
+++ b/src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Device.Gpio.Drivers;
+using System.Text;
+using System.Threading;
+using Xunit;
+
+namespace System.Device.Gpio.Tests
+{
+	public class RaspberryPi4DriverTests : GpioControllerTestBase
+	{
+        /// <summary>
+		/// Leave this pin open (unconnected) for the tests
+		/// </summary>
+		private const int OpenPin = 23;
+
+		protected override GpioDriver GetTestDriver()
+		{
+			return new RaspberryPi4Driver();
+		}
+
+		protected override PinNumberingScheme GetTestNumberingScheme() => PinNumberingScheme.Logical;
+
+        [Fact]
+        public void InputPullResistorsWork()
+		{
+			using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+			{
+				controller.OpenPin(OpenPin, PinMode.InputPullUp);
+				Thread.Sleep(10);
+				Assert.Equal(PinValue.High, controller.Read(OpenPin));
+				controller.SetPinMode(OpenPin, PinMode.InputPullDown);
+				Assert.Equal(PinValue.Low, controller.Read(OpenPin));
+				controller.SetPinMode(OpenPin, PinMode.InputPullUp);
+				Assert.Equal(PinValue.High, controller.Read(OpenPin));
+			}
+		}
+	}
+}
--- a/src/System.Device.Gpio/Interop/Unix/Libc/Interop.mmap.cs
+++ b/src/System.Device.Gpio/Interop/Unix/Libc/Interop.mmap.cs
@@ -88,4 +88,10 @@ internal unsafe struct RegisterView
     ///<summary>GPIO Pin Pull-up/down Enable Clock, 2x32 bits, R/W.</summary>
     [FieldOffset(0x98)]
     public fixed uint GPPUDCLK[2];
+
+    /// <summary>
+    /// Pull-up/down Enable on the BCM2711, 4x32 bits, R/W (2 bits per Pin)
+    /// </summary>
+    [FieldOffset(0xE4)]
+    public fixed uint GPPUPPDN[4];
 }
--- a/src/System.Device.Gpio/Interop/Unix/Libc/Interop.mmap.cs
+++ b/src/System.Device.Gpio/Interop/Unix/Libc/Interop.mmap.cs
@@ -89,9 +89,7 @@ internal unsafe struct RegisterView
     [FieldOffset(0x98)]
     public fixed uint GPPUDCLK[2];
 
-    /// <summary>
-    /// Pull-up/down Enable on the BCM2711, 4x32 bits, R/W (2 bits per Pin)
-    /// </summary>
+    /// <summary>Pull-up/down Enable on the BCM2711, 4x32 bits, R/W (2 bits per Pin)</summary>
     [FieldOffset(0xE4)]
     public fixed uint GPPUPPDN[4];
 }
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/GenericRaspberryPiDriver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/GenericRaspberryPiDriver.Linux.cs
@@ -12,7 +12,7 @@ using System.Threading.Tasks;
 
 namespace System.Device.Gpio.Drivers
 {
-    public unsafe partial class RaspberryPi3Driver : GpioDriver
+    public unsafe partial class GenericRaspberryPiDriver : GpioDriver
     {
         private const int ENOENT = 2; // error indicates that an entity doesn't exist
         private RegisterView* _registerViewPointer = null;
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi4Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi4Driver.Linux.cs
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
 using System.Collections.Generic;
 using System.Text;
 
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi4Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi4Driver.Linux.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace System.Device.Gpio.Drivers
+{
+    public unsafe partial class RaspberryPi4Driver
+    {
+        /// <summary>
+        /// Sets the resistor pull up/down mode for an input pin.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        /// <param name="mode">The mode of a pin to set the resistor pull up/down mode.</param>
+        protected override void SetInputPullMode(int pinNumber, PinMode mode)
+        {
+            // Doesn't hurt to use the old (Pi3) code as well
+            base.SetInputPullMode(pinNumber, mode);
+
+            int shift = (pinNumber & 0xf) << 1;
+            UInt32 pull = 0;
+            UInt32 bits = 0;
+            switch (mode)
+            {
+                case PinMode.Input: pull = 0; break;
+                case PinMode.InputPullUp: pull = 1; break;
+                case PinMode.InputPullDown: pull = 2; break;
+            }
+
+            var gpioReg = RegisterViewPointer;
+            bits = (gpioReg->GPPUPPDN[(pinNumber >> 4)]);
+            bits &= ~(3u << shift);
+            bits |= (pull << shift);
+            gpioReg->GPPUPPDN[(pinNumber >> 4)] = bits;
+        }
+    }
+}
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi4Driver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi4Driver.cs
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
 using System.Collections.Generic;
 using System.Text;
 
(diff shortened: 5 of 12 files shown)


src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
```diff
@@ -0,0 +1,43 @@
...
+            var bestDriver = GpioController.GetBestDriverForBoard();
+            if (!(bestDriver is RaspberryPi4Driver))
+            {
+                bestDriver.Dispose();
+                return;
+            }
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), bestDriver))
+            {
+                controller.OpenPin(OpenPin, PinMode.InputPullUp);
+                Thread.Sleep(10);
```
MarkCiliaVincenti: Can you please explain why you are pausing for 10ms?

Frankenslag:
@pgrawehr 

Some initial thoughts

Are you saying that the pi4 isn’t working with regards to seeing the pull-ups ?

Anyway I think that this or suffers the same issue as #766 tries to solve. Because the hardware and model bits of /proc/cpuinfo are Raspbian and so debian and Ubuntu wouldn’t have the info that you’re looking for. One way might be to look for the value of peripheral base address from the device tree which seems to be there in all cases I have looked at. This value from memory would be 0xFE000000 for the Pi 4.

I do think you’re right about having some cook way for these detections. How about as a static method of the RaspberryPi3Device.

Would this work better are a modification of the standard pi device rather than having a new device?

src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
pgrawehr: Not really. I was first expecting it to take a moment to settle, but (as the remainder of the test shows), the change happens instantly. So this is not needed.

pgrawehr:
> 
> 
> @pgrawehr
> 
> Some initial thoughts
> 
> Are you saying that the pi4 isn’t working with regards to seeing the pull-ups ?

Yes. When attempting to set the pull up mode, the existing Pi3 driver just doesn't change the mode from whatever it is now to the new mode. After some investigations, I found that this requires writing different hardware registers than previously. Also other low-level GPIO implementations (such as wiringpi) have the same problem, some have been updated meanwhile. See my question [here](https://raspberrypi.stackexchange.com/questions/104312/setting-pull-up-pull-down-mode-doesnt-work-mostly) for a confirmation of my observation and the hint to the correct solution (which I implemented now).  

> 
> Anyway I think that this or suffers the same issue as #766 tries to solve. Because the hardware and model bits of /proc/cpuinfo are Raspbian and so debian and Ubuntu wouldn’t have the info that you’re looking for. One way might be to look for the value of peripheral base address from the device tree which seems to be there in all cases I have looked at. This value from memory would be 0xFE000000 for the Pi 4.
I agree, but this should be solved in one place, therefore I just added a hack for this PR. 
> 
> I do think you’re right about having some cook way for these detections. How about as a static method of the RaspberryPi3Device.
Or even as a normal method. One would then create all the devices in order, and each of them would then magically detect whether they work on the current platform (+ eventually whether they're the optimal match for that platform). Now one can manually create all Device implementations, but whether they're any good is just left to the caller. 
> 
> Would this work better are a modification of the standard pi device rather than having a new device?
Not sure about this. For this single change maybe, but I don't know whether there's more to come.

src/System.Device.Gpio/Interop/Unix/Libc/Interop.mmap.cs
```diff
@@ -88,4 +88,10 @@ internal unsafe struct RegisterView
     ///<summary>GPIO Pin Pull-up/down Enable Clock, 2x32 bits, R/W.</summary>
     [FieldOffset(0x98)]
     public fixed uint GPPUDCLK[2];
+
+    /// <summary>
+    /// Pull-up/down Enable on the BCM2711, 4x32 bits, R/W (2 bits per Pin)
```
krwq: please use one liner summary similarly as other fields

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi4Driver.Linux.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
```
krwq: copyright header here and any other files it might be missing

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi4Driver.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace System.Device.Gpio.Drivers
+{
+    public unsafe partial class RaspberryPi4Driver : RaspberryPi3Driver
```
krwq: is this file actually needed? Not sure what's the value in having empty class

krwq: I mean we can leave it for compat but then I'd change RaspberryPi3Driver to be the same (and not partial) and create GenericRaspberryPiDriver or something similar (or ..Pi3Or4Driver)

pgrawehr: You mean rename the current RaspberryPi3Driver to GenericRaspberryPiDriver and then create 2 derived classes?

pgrawehr: Done.

pgrawehr:
This PR is ready for a final review, I think.

pgrawehr:
@Frankenslag : When will PR #766 be ready? I guess it would be easier to combine that directly, otherwise we'll get merge conflicts anyway.

Frankenslag:
> @Frankenslag : When will PR #766 be ready? I guess it would be easier to combine that directly, otherwise we'll get merge conflicts anyway.

@pgrawehr 
Well I have no further work to do on this at the moment I am not sure why it hasn't yet been reviewed and merged. Do you want to combine with your work ? If so let me know and I'll cancel the PR.

pgrawehr:
Leave it open for now, if it gets merged I can rebase to that; Otherwise I'll merge that on my branch, but I'll have to see whether I get to that (I have some other open ends).

src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
```diff
@@ -0,0 +1,46 @@
...
+            return new RaspberryPi4Driver();
+        }
+
+        protected override PinNumberingScheme GetTestNumberingScheme() => PinNumberingScheme.Logical;
+
+        [Fact]
+        public void InputPullResistorsWork()
+        {
+            // This is a hack to test whether we're running on a Pi4 without duplicating that detection logic. 
+            var bestDriver = GpioController.GetBestDriverForBoard();
```
krwq: I'm slightly skeptical about this check. If our detection logic is broken this will pass because it will not test anything. Can you at least make it so that test is skipped instead so it's more likely to be spotted?

In corefx we have [ConditionalFact](https://github.com/dotnet/corefx/search?q=ConditionalFact&unscoped_q=ConditionalFact) and [SkipTestException](https://github.com/dotnet/corefx/search?q=SkipTestException&unscoped_q=SkipTestException) for that but you can also use xunit Trait if not sure how to use that (i.e. `[Trait("SkipOnCi", "true")]`). For traits you will need to pass something similar to let our CI know to not run it, so i.e. https://github.com/dotnet/iot/blob/master/eng/sendToHelix.proj#L48 except do it unconditionally

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi4Driver.cs
```diff
@@ -0,0 +1,14 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace System.Device.Gpio.Drivers
+{
+    public unsafe partial class RaspberryPi4Driver : RaspberryPi3Driver
```
krwq: should this inherit from GenericRaspberryPiDriver?

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -28,19 +28,37 @@ public GpioController(PinNumberingScheme numberingScheme)
...
+                            {
+                                string revisionLine = cpuInfoLines[i + 1];
+                                match = regexRev.Match(revisionLine);
+                                if (match.Success && match.Groups.Count > 1)
+                                {
+                                    string rev = match.Groups[1].Value;
+                                    if (rev.EndsWith("03111"))
+                                    {
+                                        // The Pi4 reports the same CPU (although that's technically wrong) but different board revisions.
+                                        // The currently known revisions are 0xa03111, 0xb0311, 0xc03111 for the models with 1GB, 2GB and 4GB of ram, respectively.
```
krwq: Do you mean this won't work for 2GB or should this be `0xb03111`? Are the revisions supposed to be the same here?

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -28,19 +28,37 @@ public GpioController(PinNumberingScheme numberingScheme)
...
+                                    string rev = match.Groups[1].Value;
+                                    if (rev.EndsWith("03111"))
+                                    {
+                                        // The Pi4 reports the same CPU (although that's technically wrong) but different board revisions.
+                                        // The currently known revisions are 0xa03111, 0xb0311, 0xc03111 for the models with 1GB, 2GB and 4GB of ram, respectively. 
+                                        return new RaspberryPi4Driver();
+                                    }
+                                }
+                            }
+                            // Assume Pi3 otherwise
```
krwq: nit: enter after `}`

krwq:
Hey, sorry for the delay, this mostly looks good to me but @joperezr is more familiar with this code so he'll need to review this as well

src/System.Device.Gpio/Interop/Unix/Libc/Interop.mmap.cs
```diff
@@ -88,4 +88,8 @@ internal unsafe struct RegisterView
     ///<summary>GPIO Pin Pull-up/down Enable Clock, 2x32 bits, R/W.</summary>
     [FieldOffset(0x98)]
     public fixed uint GPPUDCLK[2];
+
+    /// <summary>Pull-up/down Enable on the BCM2711, 4x32 bits, R/W (2 bits per Pin)</summary>
+    [FieldOffset(0xE4)]
+    public fixed uint GPPUPPDN[4];
```
joperezr: Is this available on other Pis or only on Pi4? Also, do you need access to '/dev/mem' memory mapped file or is this register available via '/dev/gpiomem'?

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -28,19 +28,37 @@ public GpioController(PinNumberingScheme numberingScheme)
         /// Attempt to get the best applicable driver for the board the program is executing on.
         /// </summary>
         /// <returns>A driver that works with the board the program is executing on.</returns>
-        private static GpioDriver GetBestDriverForBoard()
+        public static GpioDriver GetBestDriverForBoard()
```
joperezr: Why do we want to change this to public? The main idea here is that you either want to get the driver automatic selection done for you which the default constructor is already doing, or you want to pass it in yourself, so I don't see why we would ever want to make this API public

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -28,19 +28,37 @@ public GpioController(PinNumberingScheme numberingScheme)
...
         {
             string[] cpuInfoLines = File.ReadAllLines(CpuInfoPath);
-            Regex regex = new Regex(@"Hardware\s*:\s*(.*)");
-            foreach (string cpuInfoLine in cpuInfoLines)
+            Regex regexHw = new Regex(@"Hardware\s*:\s*(.*)");
+            Regex regexRev = new Regex(@"Revision\s*:\s*(.*)");
```
joperezr: We are trying to move away from using Regex, which is not performant and can sometimes be unsafe to use. Is there a way to get the fact that we are running on a Pi4 via a different mechanism?

src/System.Device.Gpio/System/Device/Gpio/GpioController.Windows.cs
```diff
@@ -34,7 +34,7 @@ public GpioController(PinNumberingScheme numberingScheme)
         ///     The GpioController could use reflection to find all GpioDriver-derived classes and call this
         ///     static method to determine if the driver considers itself to be the best match for the environment.
         /// </remarks>
-        private static GpioDriver GetBestDriverForBoard()
+        public static GpioDriver GetBestDriverForBoard()
```
joperezr: We don't want to make this public, so please revert back.

src/System.Device.Gpio/System/Device/Gpio/Drivers/GenericRaspberryPiDriver.Linux.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace System.Device.Gpio.Drivers
 {
-    public unsafe partial class RaspberryPi3Driver : GpioDriver
+    public unsafe partial class GenericRaspberryPiDriver : GpioDriver
```
joperezr: This will be an implementation detail so we shouldn't make this class public. Is it not possible to simply have the RaspberryPi4Driver to inherit from RaspberryPi3Driver and override the desired functionality?

src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
```diff
@@ -0,0 +1,46 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Device.Gpio.Drivers;
+using System.Threading;
+using Xunit;
+
+namespace System.Device.Gpio.Tests
+{
+    public class RaspberryPi4DriverTests : GpioControllerTestBase
```
joperezr: This won't work until we actually have physical Raspberry Pi 4 devices on the lab in a queue. We will have to figure out how to test this before merging.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
joperezr: see @Frankenslag comment above. He has been working on getting new hardware detection logic which is moving away from Regex.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
krwq: what about making driver which extends/augments driver?

src/System.Device.Gpio/System/Device/Gpio/Drivers/GenericRaspberryPiDriver.Linux.cs
pgrawehr: I had it that way in the beginning, but @krwq suggested to change it this way.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
pgrawehr: Yes, I know. I'm actually waiting for that branch to be merged so this can be combined. However, I think that we need a more sophisticated concept for detecting compatibility between drivers and hardware.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
pgrawehr: Or (as in the test) check whether the current driver is the suggested one?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi4Driver.cs
pgrawehr: Yes, of course. Fixed.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
pgrawehr: That was a simple typo in the comment. I don't understand your second question.

src/System.Device.Gpio.Tests/RaspberryPi4DriverTests.Linux.cs
krwq: I got one Pi4 (1GB) but I haven't had time to set it up yet

src/System.Device.Gpio/Interop/Unix/Libc/Interop.mmap.cs
pgrawehr: I do not know what happens when you use that register on an older Pi, probably nothing you want. 
I did not change the register map start detection logic, so it uses /dev/gpiomem by default (which works). When forcing it to use /dev/mem, it seems to work as well (at least when sudo'ing).

src/System.Device.Gpio/System/Device/Gpio/Drivers/GenericRaspberryPiDriver.Linux.cs
pgrawehr: I agree on the visibility, though.

src/System.Device.Gpio/Interop/Unix/Libc/Interop.mmap.cs
joperezr: It is definitely something we will want to check as we don't want to map stuff to memory which doesn't exist. In general I would probably propose having another registerView for Pi4

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
joperezr: For both of those cases, I believe we should have user-defined logic for calculating that. This is an API which by definition will have breaking changes for every new board we add, so we definitely don't want to make it public and take breaking change after breaking change. Keeping it private and changing the logic is ok, and for augmenting driver we should define a different function, and for tests we might consider using InternalsVisibleTo, which is ugly, but better than exposing this API.

src/System.Device.Gpio/System/Device/Gpio/Drivers/GenericRaspberryPiDriver.Linux.cs
joperezr: This is kind of why I suggested having an API proposal for review first, so that the design is decided before implementation so that we can first have these discussions.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
krwq: @pgrawehr I didn't notice `a`, `b` and `c` changing and thought those are three the same numbers

src/System.Device.Gpio/System/Device/Gpio/Drivers/GenericRaspberryPiDriver.Linux.cs
krwq: @pgrawehr I don't recall suggesting making `GenericRaspberryPiDriver` public, only to make it a base class

src/System.Device.Gpio/Interop/Unix/Libc/Interop.mmap.cs
pgrawehr: Yes, agree. I finally found the Pi3 register map specification: https://www.raspberrypi.org/app/uploads/2012/02/BCM2835-ARM-Peripherals.pdf (pages 89ff). This ends at address B0. 
The corresponding specification document for the BCM2711 still was not published and I took the addresses for the additional registers from https://github.com/joan2937/pigpio/blob/master/pigpio.c (at the top of the file).

src/System.Device.Gpio/System/Device/Gpio/Drivers/GenericRaspberryPiDriver.Linux.cs
pgrawehr: I can make it abstract, but not internal. It's not possible to have a public class derive from an internal class.

pgrawehr: Note: The current pullup/pulldown change code for the Pi3 should be moved to the Pi3 class. It doesn't seem to hurt using that on the Pi4, but its a waste of time, since that requires some delays to be introduced.

Frankenslag: @pgrawehr @krwq @joperezr 

Perhaps we should just fix the bug and at the moment I don't believe that this should involve anything more than something like the pseudocode below in the raspberrypi3 Linux driver

```
if is raspberrypi4 then
 use new pi 4 code for SetInputPullMode
else
 use existing code for SetInputPullMode
```

I think I have suggested that the test for pi4 could be based on the peripheral base address which could be cached during initialization.

krwq:
@pgrawehr there are some merge conflicts here (probably with your other PR)

pgrawehr:
@krwq Yes, that was to be expected. I don't know how to continue here, actually. We didn't quite get to a conclusion on #809, which I originally opened for a design discussion on this, but meanwhile went into a much broader discussion related to the "board" design topic. 

So I think until we come to a conclusion on that issue, and the result involves creating separate classes for Pi3 and Pi4, we can close this PR. The main problem is solved for now.

pgrawehr:
Closing as per previous comment (for now)

dontspeak:
Hello
why is this issue closed? the InputPullUp and -PullDown Resistors do not work in any way on a RP4. At least not in current version 1.0.0. Also, a `RaspberryPi4Driver` does not exist yet.
Regards Vasilios.

joperezr:
This has been fixed in master branch, but we don't have a package in NuGet with this so it will ship with vNext. We plan on shipping some preview packages for vNext to NuGet in the next couple of weeks, but if you need this in the meantime you can just add our nightly feed to your project and restore the latest package. To do this, just add afile next to your project called NuGet.config with the following contents:
```xml
<configuration>
  <packageSources>
    <add key="dotnet5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
  </packageSources>
</configuration>
```

The latest version as of today, is `1.1.0-prerelease.20078.2`

dontspeak:
I did that and its working now.
Thank you so much!

krwq:
@joperezr this nuget feed is different than the one on the README.md, should it be updated?

joperezr:
Yes it should. I'll take care of that 😄

PR #789, opened 2019-10-11
Run unit tests as part of the build (and CI)

krwq:
Noticed that unit tests don't get run as part of the CI (only System.Device.Gpio.Tests do).

Marked as draft until I finish testing

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -4,13 +4,13 @@ trigger:
     include:
     - master
     - release/3.0
-    
+
 pr:
   branches:
     include:
     - master
     - release/3.0
-    
+
 variables:
   # Cannot use key:value syntax in root defined variables
   - name: Build.Repository.Clean
@@ -111,7 +111,7 @@ stages:
 
         - script: powershell -ExecutionPolicy ByPass -NoProfile eng\common\msbuild.ps1 -warnaserror:0 -ci
             eng/sendToHelix.proj
-            /t:Test
+            /t:TestToMultipleQueues
             /p:TestOS=Windows_NT
             /p:Configuration=$(_BuildConfig)
             /p:HelixBuild=$(Build.BuildNumber)
@@ -128,7 +128,7 @@ stages:
             artifactName: BuiltPackageOutputs
             artifactType: container
           condition: eq(variables['_BuildConfig'], 'Release')
-      
+
       - job: Linux
         container: LinuxContainer
         pool:
@@ -159,7 +159,7 @@ stages:
           displayName: Unix Build
         - script: ./eng/common/msbuild.sh --warnaserror false --ci
             eng/sendToHelix.proj
-            /t:Test
+            /t:TestToMultipleQueues
             /p:TestOS=Unix
             /p:Configuration=$(_BuildConfig)
             /p:HelixBuild=$(Build.BuildNumber)
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -4,13 +4,13 @@ trigger:
     include:
     - master
     - release/3.0
-
+    
 pr:
   branches:
     include:
     - master
     - release/3.0
-
+    
 variables:
   # Cannot use key:value syntax in root defined variables
   - name: Build.Repository.Clean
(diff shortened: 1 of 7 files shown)


.vsts-ci.yml
```diff
@@ -165,7 +166,20 @@ stages:
             /p:HelixBuild=$(Build.BuildNumber)
             /bl:$(Build.SourcesDirectory)/artifacts/log/$(_BuildConfig)/SendToHelix.binlog
             /p:HelixAccessToken=$(HelixApiAccessToken)
-          displayName: Run Helix Tests
+          displayName: Run Helix Unit Tests
+          env:
+            SYSTEM_ACCESSTOKEN: $(System.AccessToken)
+            HelixAccessToken: $(HelixApiAccessToken)
+        - script: ./eng/common/msbuild.sh --warnaserror false --ci
```
joperezr: Why do you need to add another step, Ideally we would run both with just the one call.

joperezr:
How many device binding unit test projects do we have? If there aren't many, it is probably easier for now to just run these tests as part of the build. That way you won't really have to deal with Helix or build infrastructure, and you will get coverage in Windows, Linux, and OSX. Because they are unit tests, they shouldn't need to run on any specific hardware.

.vsts-ci.yml
krwq: You can't unless we run it on PIs as well since there is no way specify which project runs on which queue. If I create a target which call sendToHelix twice then I hit this issue: https://github.com/dotnet/arcade/pull/4107

Other thing is that it makes sense to first run unit tests before running any hardware tests so that should be fine

krwq:
@joperezr we have 6 unit tests projects not including System.Device.Gpio. I prefer if they got run as part of CI since that's what the unit tests are for

krwq:
seems something got stuck in helix. The tests took less than 30s on my box

krwq:
@joperezr @safern let me know if you have any clues why is helix job getting stuck or how do you check what went wrong

.vsts-ci.yml
```diff
@@ -112,11 +112,12 @@ stages:
         - script: powershell -ExecutionPolicy ByPass -NoProfile eng\common\msbuild.ps1 -warnaserror:0 -ci
             eng/sendToHelix.proj
             /t:Test
+            /p:HardwareDependentProjects=true
             /p:TestOS=Windows_NT
             /p:Configuration=$(_BuildConfig)
             /p:HelixBuild=$(Build.BuildNumber)
             /bl:$(Build.SourcesDirectory)/artifacts/log/$(_BuildConfig)/SendToHelix.binlog
```
krwq: binlog name should be different for hardware dependent

MarkCiliaVincenti:
What's wrong with GitHub? My checks are also failing with strange Azure DevOps problems.

krwq:
@MarkCiliaVincenti not sure, haven't seen any e-mails internally yet

krwq:
Rebased to restart CI since it was showing green when clicking on link but red on github

MarkCiliaVincenti:
> Rebased to restart CI since it was showing green when clicking on link but red on github

Is it still WIP? I'm depending on this for #783.

joperezr:
> @MarkCiliaVincenti not sure, haven't seen any e-mails internally yet

There is an internal thread about problems with Azure Devops that we are seeing, this is very likely the cause to issues you were seeing.

build.proj
```diff
@@ -56,7 +60,13 @@
     <MSBuild Projects="@(_ProjectsToRestore)" Targets="Restore" />
   </Target>
 
-  <Target Name="Test" />
+  <Target Name="Test">
+    <MSBuild Projects="@(UnitTestProjects)" Targets="VSTest"
+        ContinueOnError="ErrorAndContinue"
+        Properties="Verbosity=normal" />
```
joperezr: Let's just check to see if this is the best way to print out failures from running the tests. In general we try to avoid passing down global properties as they leak down into the the projects and mess up incremental build. cc: @ericstj

safern: Also if you're running tests as part of the build, you want to publish the xunit results in the yaml file as part of the build in CI so that test results show in the test explorer: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=yaml#yaml-snippet

krwq:
Ok lack of output seems to be VSTest issue: https://github.com/microsoft/vstest/issues/2224 I'll remove the hack and let's wait until they fix it

PR #793, opened 2019-10-12
1-wire controller and thermometer binding

maloo:
Implementation of proposed API in #197.

Code sample to scan for devices, enumerate buses and devices and print temperature of all thermometer family devices found.


src/devices/OneWire/OneWireController.cs
```diff
@@ -0,0 +1,56 @@
...
+        internal static OneWireDevice CreateDeviceByFamily(OneWireBus bus, string deviceId, OneWireBus.DeviceFamily family)
+        {
+            switch ((int)family)
+            {
+                case 0x10:
+                    goto case 0x42;
+                case 0x28:
+                    goto case 0x42;
+                case 0x3B:
+                    goto case 0x42;
```
krwq: This can be replaced with

```csharp
                case 0x10:
                case 0x28:
                case 0x3B:
                case 0x42:
                    return new OneWireThermometerDevice(bus, deviceId, family);
```

src/devices/OneWire/OneWireController.cs
```diff
@@ -0,0 +1,56 @@
...
+                foreach (var dev in bus.EnumerateDevices(family))
+                {
+                    yield return dev;
+                }
+            }
+        }
+
+        internal static OneWireDevice CreateDeviceByFamily(OneWireBus bus, string deviceId, OneWireBus.DeviceFamily family)
+        {
+            switch ((int)family)
```
krwq:
```csharp
switch (family)
{
    case OneWireBus.DeviceFamily.XYZ:
   ...
}
```

src/devices/OneWire/samples/OneWire.Sample.cs
```diff
@@ -0,0 +1,31 @@
...
+    class Program
+    {
+        static async Task Main(string[] args)
+        {
+            // Make sure you can access the bus device before requesting a device scan
+            // $ sudo chmod a+rw /sys/bus/w1/devices/w1_bus_master1/w1_master_*
+            foreach (var bus in OneWireController.EnumerateBuses())
+            {
+                Console.WriteLine($"Found bus '{bus.DeviceId}', scanning for devices ...");
+                await bus.ScanForDevicesAsync();
```
krwq: should EnumerateDevices call the scanner?

src/devices/OneWire/samples/OneWire.Sample.cs
```diff
@@ -0,0 +1,31 @@
...
+                foreach (var dev in bus.EnumerateDevices())
+                {
+                    Console.WriteLine($"Found family '{(int)dev.Family:x2}' device '{dev.DeviceId}' on master '{bus.DeviceId}'");
+                    if (dev is OneWireThermometerDevice devTemp)
+                    {
+                        Console.WriteLine("Temperature reported by device: " + (await devTemp.ReadTemperatureAsync()).Celsius.ToString("F2") + "°C");
```
krwq: use `\uXYZV` for degrees (°) character

src/devices/OneWire/OneWireController.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            return EnumerateBusesInternal();
+        }
+
+        /// <summary>
+        /// Enumerate all devices found on 1-wire busses in this system.
+        /// </summary>
+        /// <param name="family">Family id used to filter devices.</param>
+        /// <returns>A list of devices found.</returns>
+        public static IEnumerable<OneWireDevice> EnumerateDevices(OneWireBus.DeviceFamily family = OneWireBus.DeviceFamily.Any)
```
krwq: DeviceFamily should not be nested class

src/devices/OneWire/OneWireController.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// <summary>
+    /// Represents a 1-wire controller.
+    /// </summary>
+    public static partial class OneWireController
+    {
+        /// <summary>
+        /// Enumerate all 1-wire busses in the system.
+        /// </summary>
+        /// <returns>A list of discovered busses.</returns>
+        public static IEnumerable<OneWireBus> EnumerateBuses()
```
krwq: I got mixed feeling on creating OneWireBus/OneWireDevice objects when enumerating. Especially I'd expect OneWireDevice to open the file in the constructor and keep it open until it is disposed.

I think I'd go with enumerating names/ids of the buses instead of the object and for enumerating devices either return a tuple or create a OneWireDeviceInfo struct which contains bus id/name and device name/id

src/devices/OneWire/OneWireBus.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        /// <summary>
+        /// The 1-wire device id.
+        /// </summary>
+        public string DeviceId { get; }
+
+        /// <summary>
+        /// Different device families supported by the driver.
+        /// </summary>
+        public enum DeviceFamily : int
```
krwq: this should not be nested

src/devices/OneWire/OneWireController.Linux.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        private static IEnumerable<OneWireBus> EnumerateBusesInternal()
+        {
+            foreach (var entry in Directory.EnumerateDirectories(SysfsBusDevicesPath, "w1_bus_master*"))
+            {
+                yield return new OneWireBus(Path.GetFileName(entry));
+            }
+        }
+
+        internal static IEnumerable<OneWireDevice> EnumerateDevices(OneWireBus bus, OneWireBus.DeviceFamily family)
```
krwq: I think this should be moved as a static method of OneWireDevice, enumerating buses to OneWireBus and this class should be removed

src/devices/OneWire/OneWireController.cs
maloo: There are no names except hex code for each family, but I will add each device I find as family name. You will get more device families back when specifying a device family with multiple devices in it.

src/devices/OneWire/samples/OneWire.Sample.cs
maloo: No, scan is an expensive operation. If you just want to enumerate devices already found, this is fast. Scan is when you hot plug devices.

src/devices/OneWire/OneWireController.cs
maloo: We are talking one small object (bus instance) extra in normal case. I can make it a struct BusInfo. But I don't see any big gain over keeping it easy to understand/use. Device should probably be an object since user will hold on to it for longer. So I prefer them to be similar Bus+Device instad of BusInfo+Device.
And I don't like holding a system resource longer than necessary. IDispose is easy to miss and when a read is async and take quite some time there is no need to cache the file handle. I go for ease of use over micro optimization.

maloo:
Should fix #172

src/devices/OneWire/OneWireController.cs
krwq: for Buses I'd return string. I think it's generally better to hold system resources for things you're gonna use for longer period of time (thermometer)

src/devices/OneWire/OneWireController.cs
krwq: my comment is that enum is equivalent to int and there is no point in obscuring what the hex means by casting it to int

maloo: My point was that it would no longer be a family id, but instead a pseudo device id and not possible to enumerate for example temperature devices. But I have changed all hex to enum values and added special out of spec values to simulate family ids.

src/devices/OneWire/OneWireController.cs
maloo: Why, then you would have to make all bus operations static and take a bus id as input.

maloo: Or you want the user to create a OneWireBus from a string/busid?

src/devices/OneWire/DeviceFamily.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </summary>
+        Ds18B20 = 0x28,
+        /// <summary>
+        /// Family id of a MAX31820 temperature sensor.
+        /// </summary>
+        Max31820 = 0x28,
+        /// <summary>
+        /// Family id of a DS1825 temperature sensor.
+        /// </summary>
+        DS1825 = 0x3B,
```
krwq: inconsistent casing. I.e. this should be `Ds1825`

src/devices/OneWire/OneWireBus.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// </summary>
+    public partial class OneWireBus
+    {
+        internal OneWireBus(string deviceId)
+        {
+            DeviceId = deviceId;
+        }
+
+        /// <summary>
+        /// The 1-wire device id.
```
krwq: should this be bus id?

src/devices/OneWire/OneWireBus.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <summary>
+        /// Enumerate all 1-wire busses in the system.
+        /// </summary>
+        /// <returns>A list of discovered busses.</returns>
+        public static IEnumerable<OneWireBus> EnumerateBuses()
+        {
+            return EnumerateBusesInternal();
+        }
+
+        internal static OneWireDevice CreateDeviceByFamily(OneWireBus bus, string deviceId, DeviceFamily family)
```
krwq: 1. Why is this static if first if of the same type and null doesn't make sense in here?
2. Consider changing to `T CreateDevice<T>(string deviceId) where T : OneWireDevice`
3. Is deviceId actually a string or is this an `int`?

src/devices/OneWire/OneWireBus.cs
```diff
@@ -0,0 +1,70 @@
...
+            return EnumerateDevicesInternal(family);
+        }
+
+        /// <summary>
+        /// Start a new scan for devices on the bus.
+        /// </summary>
+        /// <param name="numDevices">Max number of devices to scan for before finishing.</param>
+        /// <param name="numScans">Number of scans to do to find numDevices devices.</param>
+        /// <returns>Task representing the async work.</returns>
+        public Task ScanForDevicesAsync(int numDevices = 5, int numScans = -1)
```
krwq: why 5?

src/devices/OneWire/DeviceFamily.cs
maloo: Will fix

src/devices/OneWire/OneWireBus.cs
maloo: Sure, will change, thought of bus as a device too like in Linux drivers. But agree bus id probably better here.

src/devices/OneWire/OneWireBus.cs
maloo: 1) leftover from controller, will fix
2) ok
3) it's 8bit family, 48bit id and 8bit crc

src/devices/OneWire/OneWireBus.cs
maloo: I think it is the default value in Linux, but could be any value I guess

src/devices/OneWire/OneWireBus.cs
krwq: string sounds good in that case

src/devices/OneWire/OneWireBus.cs
krwq: Fine if it's a default value on Linux. Please add a tiny comment inside of the method with link pointing to where 5 is coming from

maloo: It was 64, probably changed it to test it, don't have 64 sensors ;)
https://github.com/torvalds/linux/blob/v5.3/drivers/w1/w1.c#L46

krwq: Linux source is GPL2 so we can't look at that code to avoid any legal issues. Unless there is some public API description saying 64 I'd make those required arguments and make the names a bit more explicit: `maxDevices`, `maxScans`. I find this (Linux) API kinda weird though, I'd personally expect all devices to be detected and the second argument is particularly confusing: i.e. when would I want scanning more than once? Is once not enough?

src/devices/OneWire/OneWireController.cs
krwq: I actually didn't notice there is no string constructor but yes, I'd prefer bus was created from string and this API enumerated names.

krwq: Out of curiosity: does the bus id stay the same when rebooting OS?

src/devices/OneWire/OneWireBus.cs
```diff
@@ -62,8 +62,11 @@ public IEnumerable<OneWireDevice> EnumerateDevices(DeviceFamily family = DeviceF
         /// <param name="numDevices">Max number of devices to scan for before finishing.</param>
         /// <param name="numScans">Number of scans to do to find numDevices devices.</param>
         /// <returns>Task representing the async work.</returns>
-        public Task ScanForDevicesAsync(int numDevices = 5, int numScans = -1)
+        public Task ScanForDevicesAsync(int numDevices = 64, int numScans = -1)
         {
+            // Default 64 used to align with Linux driver
+            // https://github.com/torvalds/linux/blob/v5.3/drivers/w1/w1.c#L46
```
krwq: with this link now we essentially must remove this reference unless you find public API mention about 64 👀 (sorry, didn't invent these rules and also not a fan of it)

krwq:
Generally this looks mostly good to me, the bus enumeration and `64` are two things we need to fix up.

maloo:
I could set the default to -1 instead of 64 and interpret that as don't change it from default.

maloo:
As for bus I didn't see any response to my comments/questions. Maybe you could elaborate about what it is exactly that you don't like. Is it the use of an object for bus (which I need to do operations on like scan/search etc)? Or is it the fact that I allocate bus objects in OneWireDevice.EnumerateDevices() (which could be optimized to not using EnumerateBuses()+bus.EnumerateDevices())?

krwq:
@maloo I've already responded above but seems something bugged there and doesn't display correctly.

For 1st question: -1 sounds good but we cannot use value 64 without any non-GPL2 licensed sources (i.e. if you find it in the public docs then it's fine) and we should also remove a link to GPL2 code. Arbitrary constant is fine (except 64) but please document it is arbitrary so that we don't need to scratch our head about it in 5 years in case we need to change it 😄 Biggest concern about this Scan API is that I'm not sure how Windows will or is implementing it and I feel like it's leaking implementation detail a bit. I'd rather have us call it automatically before enumerating even if it's heavy and slow. User can store the output if they find it too slow and for whatever reason they need to enumerate it more than once.
For 2nd question:
- enumerating buses: I don't like allocating object which might potentially grow in the future (i.e. because of Windows implementation). The way I see it you enumerate bus names as strings and then construct bus from the string and store it in the field. This method I think should be static,
- enumerating devices: similar comment as above except it should live as an instance method of the bus (note it can also enumerate strings if it is an instance method)

src/devices/OneWire/OneWireController.cs
maloo: yes, it is the Linux device name

maloo:
I can't find a nice solution with bus as id during enumeration. I keep having to new up the same amount or more OneWireBus instances. I see two use cases (see 1-wire sample).
1) Simple, I just want a list of e.g. temp devices. So I do OneWireDevice.EnumerateDevices(). This will have to return the device instances, because user obviously want to interact with the devices when using this API. Then there are two options for OneWireDevice. It could have the property Bus or BusId. If bus, enumerating the devices means I have to create the bus instances. If BusId, the user would have to new up a bus instance every time it want to do a bus operation like start a scan for device changes. In the second case you might get away with not allocating the bus object, but you loose the control of how/when the bus objects are created. If Windows would come up with a solution where a 1-wire bus instance is expensive to create, I would like to hold on to it for as long as possible just like you said in previous comments when discussing file handles. Best way to do that is to have 1-wire bus control when to create the instances and how to cache them. I would even want the Controller we removed, to allow implementation control over the "session" so it could cahce any expensive objects. Removing the controller forces EnumerateX() to create new instances or have a static cache that is never cleaned (like it would be in OneWireController.Dispose()).
2) Advanced usage, enum buses and then devices. Here user will obviously need the bus device, otherwise he would use the simple device enumeration.

maloo:
Or, we make bus/device structs OWBusId/OWDeviceId w/ same operations as OWBus/OWDevice. That is technically the same as string id and static operations taking id param. But then I would like to get back OWController w/ Dispose to control caching and state lifetime for future platforms with heavyweight device state.
And could make ReadTemp an extension method on devid and skip family type instances to make common use case easy like now.

src/devices/OneWire/OneWireBus.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Threading.Tasks;
+
+namespace Iot.Device.OneWire
+{
+    /// <summary>
+    /// Represents a 1-wire bus.
+    /// </summary>
+    public partial class OneWireBus
+    {
+        internal OneWireBus(string busId)
```
krwq: make this public

src/devices/OneWire/OneWireBus.Linux.cs
```diff
@@ -0,0 +1,51 @@
...
+        internal const string SysfsBusDevicesPath = "/sys/bus/w1/devices";
+        internal const string SysfsDevicesPath = "/sys/devices";
+
+        private static IEnumerable<OneWireBus> EnumerateBusesInternal()
+        {
+            foreach (var entry in Directory.EnumerateDirectories(SysfsBusDevicesPath, "w1_bus_master*"))
+            {
+                yield return new OneWireBus(Path.GetFileName(entry));
+            }
+        }
```
krwq:
```csharp
        private static IEnumerable<string> EnumerateBusesInternal()
        {
            foreach (var entry in Directory.EnumerateDirectories(SysfsBusDevicesPath, "w1_bus_master*"))
            {
                yield return Path.GetFileName(entry);
            }
        }
```

src/devices/OneWire/OneWireBus.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <summary>
+        /// The 1-wire device id.
+        /// </summary>
+        public string BusId { get; }
+
+        /// <summary>
+        /// Enumerate all 1-wire busses in the system.
+        /// </summary>
+        /// <returns>A list of discovered busses.</returns>
+        public static IEnumerable<OneWireBus> EnumerateBuses()
```
krwq:
```suggestion
        public static IEnumerable<string> EnumerateBuses()
```

krwq:
The code suggestions are what I meant

src/devices/OneWire/DeviceFamily.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </summary>
+        Max31826 = 0x3B,
+        /// <summary>
+        /// Family id of a MAX31850 temperature sensor.
+        /// </summary>
+        Max31850 = 0x3B,
+        /// <summary>
+        /// Family id of a DS28EA00 temperature sensor.
+        /// </summary>
+        Ds28EA00 = 0x42,
```
krwq: There still seem to be some capital letters here and in `Ds18S20`

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.Linux.cs
```diff
@@ -0,0 +1,28 @@
...
+        private async Task<Temperature> ReadTemperatureInternalAsync()
+        {
+            // Expected data format:
+            // 7f 01 4b 46 7f ff 01 10 33 : crc=33 YES
+            // 7f 01 4b 46 7f ff 01 10 33 t=23937
+            var data = await File.ReadAllTextAsync(Path.Combine(OneWireBus.SysfsDevicesPath, Bus.BusId, DeviceId, "w1_slave"));
+            if (!data.Contains("YES"))
+                throw new IOException("Unable to read temperature from device.");
+            var tempIdx = data.LastIndexOf("t=");
+            if (tempIdx == -1 || tempIdx + 2 >= data.Length || !int.TryParse(data.AsSpan(tempIdx + 2), out var temp))
```
krwq: nit: enter before and after if statement

maloo:
Yes, that far I got it :)
I might be slow, but I don't understand why? Just to get me on board with the change ...

1) Which use case benefit from this change in the public API?
  a) The basic OneWireDevice.EnumerateDevices()
  b) More advanced OneWireBus.EnumerateBuses() w/ bus.ScanForDeviceChanges/EnumerateDevice()?
  c) Other? ...
2) What exactly is the problem with the current one?
  a) Is it the number of allocations of OneWireBus? Which I believe will be the same in a real world use case.
  b) Is it that an eventual Windows OneWireBus object could be heavy weight? Which I believe should be cached and have the dev/bus instances to be references instead.
  c) Is it the OneWireDevice.EnumerateDevices() that allocate and throw away OneWireBus instances during enumeration? That is internal and could easily be optimized, but for no real gain in the Linux version.
  d) Other? ...

I believe I have listed the pros/cons with the string only solution above, and it in not the API change itself, but rather the changes it force in the other part of the API that I don't like. And the use cases listed in the sample code that get messier.

src/devices/OneWire/DeviceFamily.cs
maloo: This one I assumed to be a Hexcode, and thought it looked better as EA, do you want it to be Ea because it is a name or ea because it is hex code?

maloo: In Ds18S20, I went with what I though was the standard, Pascal case, that is first Upper and rest lower, but since only one letter - Upper case S.

krwq:
@maloo do bus id / device id change when rebooting? If they don't then you can avoid enumerating and allocating all together if user stores ID - if it's not possible on linux it might be possible on different OSes we might support in the future.

Other reason is that it follow patterns we do in other places:
- enumerating files enumerates names and does not create FileStreams
- SerialPort.GetPortNames also enumerates strings and not SerialPort instances

also enumerating string allows users do display something sensible in the UI

maloo:
Ok, so the main reason is SerialPort/File enumeration style.

IDs don't change, but devices come and go dynamically so background scans or user requested bus scans can change What devices are enumerated.

My main objection to the id only solution is the user's code:

    foreach (var dev in OneWireDevice.EnumerateDevices(DeviceFamily.Temp).Cast<OneWireTempDevice>()) {
        var temp = await dev.ReadTemperatureAsync();
        ReportTemp(dev, temp);
    }

would now be:

    foreach (var busId in OneWireBus.EnumerateBusNames())
    {
        var bus = new OneWireBus(busId);
        foreach (var devId in bus.EnumerateDeviceNames(DeviceFamily.Temp))
        {
            var dev = OneWireDevice(bus.BusId, devId);
            var temp = await dev.ReadTemperatureAsync();
            ReportTemp(dev, temp);
        }
    }

The second one is just a bit too complex for me just to keep style with File/SerialPort ;) It is also forcing user to instantiate a OneWireBus, in the other it is not needed (internal details that can be fixed if needed).
Maybe you have another picture of how to do a temperature sampling of available sensors using ids in the API?

Common ground? What about supporting both? EnumerateBusNames/EnumerateDeviceNames() and EnumerateDevices()

maloo:
Looking back in the comments, maybe we have common ground. Do you want the id solution for bus only or bus+device? I first interpreted your comments to want both, but in later comments I see you only mention the bus. So OneWireDevice.EnumerateDevices() would be fine?

src/devices/OneWire/OneWireBus.cs
maloo: done

src/devices/OneWire/OneWireBus.Linux.cs
maloo: done

src/devices/OneWire/OneWireBus.cs
maloo: done

krwq:
@maloo just 1. I initially thought both because it was static but I think I actually prefer to have just 1

src/devices/OneWire/OneWireBus.cs
```diff
@@ -57,17 +62,14 @@ public IEnumerable<OneWireDevice> EnumerateDevices(DeviceFamily family = DeviceF
         }
 
         /// <summary>
-        /// Start a new scan for devices on the bus.
+        /// Start a new scan for device changes on the bus.
         /// </summary>
-        /// <param name="numDevices">Max number of devices to scan for before finishing.</param>
-        /// <param name="numScans">Number of scans to do to find numDevices devices.</param>
+        /// <param name="numDevices">Max number of devices to scan for before finishing. Use -1 for pltform default.</param>
```
krwq: typo: `pltform` -> `platform`

krwq:
@maloo IMO EnumerateDevices (assuming it's an instance method on OneWireBus) should enumerate only ids since bus id is implied by the instance.

For the concern about casting, if you enumerate strings then you could potentially do:
`bus.EnumerateDeviceIds(DeviceFamily.Thermometer).Select((id) => new OneWireThermometerDevice(id))` which doesn't seem bad

src/devices/OneWire/DeviceFamily.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <summary>
+        /// Special family id used to enumerate all temperature sensors.
+        /// </summary>
+        DigitalThermometer = 0x100,
```
krwq: I'd make it just `Thermometer`

src/devices/OneWire/DeviceFamily.cs
maloo: @krwq I didn't get any feedback on these for next update. Can you just give me the names you want?

krwq: I'm not sure, I think if it's one symbol then only first letter should be uppercase

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
```diff
@@ -15,23 +14,50 @@ namespace Iot.Device.OneWire
...
+        /// Check if family is compatible with this type of devices.
+        /// </summary>
+        /// <param name="family">The family to check.</param>
+        /// <returns>Returns true if device is compatible.</returns>
+        public static bool IsCompatible(DeviceFamily family)
+        {
+            return family == DeviceFamily.Ds18S20 ||
+                   family == DeviceFamily.Ds18B20 ||
+                   family == DeviceFamily.Ds1825 ||
+                   family == DeviceFamily.Ds28EA00;
```
krwq: `|| family == DeviceFamily.Thermometer`?

maloo: That is a made up family id for filtering. It is not a valid id, no device should be created with it.

src/devices/OneWire/DeviceFamily.cs
maloo: Ok, I can leave the one letter ones and change EA to Ea in next push.

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
maloo: One solution is to create two enum, add DeviceFamilyFilter. But one is a superset so seems overkill.

src/devices/OneWire/DeviceFamily.cs
krwq: IMO this should be `Ds28ea00` unless `EA00` has any meaning then as you suggested

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.Linux.cs
```diff
@@ -0,0 +1,30 @@
...
+            // Expected data format:
+            // 7f 01 4b 46 7f ff 01 10 33 : crc=33 YES
+            // 7f 01 4b 46 7f ff 01 10 33 t=23937
+            var data = await File.ReadAllTextAsync(Path.Combine(OneWireBus.SysfsDevicesPath, BusId, DeviceId, "w1_slave"));
+            if (!data.Contains("YES"))
+                throw new IOException("Unable to read temperature from device.");
+
+            var tempIdx = data.LastIndexOf("t=");
+            if (tempIdx == -1 || tempIdx + 2 >= data.Length || !int.TryParse(data.AsSpan(tempIdx + 2), out var temp))
+                throw new FormatException("Invalid sensor data format.");
```
krwq: I think this should be `InvalidOperationException` (also the error above). FormatException is when user passed in a format (i.e. to string.Format) and that format is incorrect

src/devices/OneWire/OneWireBus.Linux.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
```
krwq: license header

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
```diff
@@ -0,0 +1,73 @@
...
+                yield return new OneWireThermometerDevice(busId, devId);
+            }
+        }
+
+        /// <summary>
+        /// Reads the current temperature of the device.
+        /// Expect this function to be slow (about one second).
+        /// </summary>
+        /// <returns>The read temperature value.</returns>
+        public Task<Temperature> ReadTemperatureAsync()
```
krwq: there should also be synchronous overload

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
```diff
@@ -0,0 +1,73 @@
...
+                   family == DeviceFamily.Ds28EA00;
+        }
+
+        /// <summary>
+        /// Check if device is compatible with this family of devices.
+        /// </summary>
+        /// <param name="busId">The 1-wire bus the device is found on.</param>
+        /// <param name="devId">The id of the device.</param>
+        /// <returns>Returns true if device is compatible.</returns>
+        public static bool IsCompatible(string busId, string devId)
```
krwq: This should be private I believe (not seeing any reason on making it public)

src/devices/OneWire/OneWireBus.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Start a new scan for device changes on the bus.
+        /// </summary>
+        /// <param name="numDevices">Max number of devices to scan for before finishing. Use -1 for platform default.</param>
+        /// <param name="numScans">Number of scans to do to find numDevices devices. Use -1 for platform default.</param>
+        /// <returns>Task representing the async work.</returns>
+        public Task ScanForDeviceChangesAsync(int numDevices = -1, int numScans = -1)
```
krwq: please add sync overload

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
krwq: I'd make this method private/internal. I'm not sure what it's for (from user perspective)

krwq:
LGTM after fixing comments

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
maloo: See sample, I'll keep it public.

krwq:
Thank you @maloo! I'm still slightly skeptical about the IsCompatible (perhaps the name) but we can re-iterate on that if we get some issues about it

PR #797, opened 2019-10-13
Bmxx80 API modification

RobinTTY:
Addresses #753 

Still need to rebase (as soon as #783 is merged) and make some other changes.
<s>This will be a breaking change.</s>


src/devices/Bmxx80/ReadResult/Bme280ReadResult.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// Collected temperature measurement.
+        /// </summary>
+        public Temperature Temperature;
+
+        /// <summary>
+        /// Collected pressure measurement.
+        /// </summary>
+        public double Pressure;
```
MarkCiliaVincenti: You should use the new Pressure unit here (refer to #768), same as it is with Temperature (property above it).

src/devices/Bmxx80/ReadResult/Bme680ReadResult.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Collected temperature measurement.
+        /// </summary>
+        public Temperature Temperature;
+
+        /// <summary>
+        /// Collected pressure measurement.
+        /// </summary>
+        public double Pressure;
```
MarkCiliaVincenti: Same here, please use the Pressure unit.

src/devices/Bmxx80/ReadResult/Bmp280ReadResult.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// Collected temperature measurement.
+        /// </summary>
+        public Temperature Temperature;
+
+        /// <summary>
+        /// Collected pressure measurement.
+        /// </summary>
+        public double Pressure;
```
MarkCiliaVincenti: And here... Pressure unit.

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -135,24 +135,16 @@ public void Reset()
...
+        /// Reads the pressure from register.
         /// </summary>
-        /// <param name="pressure">
-        /// Contains the measured pressure in Pa if the <see cref="PressureSampling"/> was not set to <see cref="Sampling.Skipped"/>.
-        /// Contains <see cref="double.NaN"/> otherwise.
-        /// </param>
-        /// <returns><code>true</code> if measurement was not skipped, otherwise <code>false</code>.</returns>
-        public abstract bool TryReadPressure(out double pressure);
+        /// <returns>The measured pressure.</returns>
+        protected abstract double ReadPressureRegister();
```
MarkCiliaVincenti: You should use the new Pressure unit here (refer to #768), same as it is with Temperature (property above it).

RobinTTY: Yeah I still need to rebase onto master, which I will do as soon as #783 is merged.

MarkCiliaVincenti: But #768 is merged and is separate from #783.

RobinTTY: I see, will rebase onto current master then for now and use the new pressure unit of course.

MarkCiliaVincenti:
@RobinTTY, #783 was merged at last. How can we incorporate this now?
Also, should we be providing async methods?

MarkCiliaVincenti:
@RobinTTY, #783 was merged at last. How can we incorporate this now?
Also, should we be providing async methods? @krwq @joperezr

RobinTTY:
@MarkCiliaVincenti ah great, I will take a look at it and try to finish this pull request finally in the next few days. I will have to take a look again but I think we agreed in #753 to implement a new read method (synchronous and async implementation) while keeping the old implementation where you still have to set the powerMode manually. If the plan to keep the current methods hasn't changed I would try to incorporate that.

MarkCiliaVincenti:
> @MarkCiliaVincenti ah great, I will take a look at it and try to finish this pull request finally in the next few days. I will have to take a look again but I think we agreed in #753 to implement a new read method (synchronous and async implementation) while keeping the old implementation where you still have to set the powerMode manually. If the plan to keep the current methods hasn't changed I would try to incorporate that.
> 
> Edit: @krwq @joperezr Since last working on this the project states now "This repository is still in experimental stage and all APIs are subject to changes.". Does this mean you would be fine with removing the old public facing methods e.g.:
> 
> ```cs
> i2CBmpe80.SetPowerMode(Bmx280PowerMode.Forced);
> i2CBmpe80.TryReadTemperature(out tempValue);
> i2CBmpe80.TryReadPressure(out preValue);
> i2CBmpe80.TryReadHumidity(out humValue);
> ```
> 
> for one read method as mentioned in #753? It would certainly make the implementation more clean.

My opinion would be to leave them there, call the read method and mark them as obsolete.

RobinTTY:
I implemented the Read/ReadAsync methods but discovered a bug that was introduced when the switch to UnitsNet happened.
Since the Bmxx80 devices have an option to disable a certain reading like this:

```C#
bme680.PressureSampling = Sampling.Skipped;
```

and UnitsNet does not handle NaN values, all sensors will currently throw an exception when a sampling is skipped since the current implementation will try this:

```C#
if (PressureSampling == Sampling.Skipped)
{
    pressure = Pressure.FromPascals(double.NaN);
}
```

Is there already some guidance on how to handle this? All solutions I could think of are not that clean. The easiest and "cleanest" probably being setting those values to the maximum possible value if they are not measured and letting the user handle it. Nullable values don't make much sense since they would complicate using the driver a lot (and would also make the driver code itself messy):

```C#
public static async Task Main()
        {
            Console.WriteLine("Hello BME680!");

            // The I2C bus ID on the Raspberry Pi 3.
            const int busId = 1;
            // set this to the current sea level pressure in the area for correct altitude readings
            var defaultSeaLevelPressure = WeatherHelper.MeanSeaLevel;

            var i2cSettings = new I2cConnectionSettings(busId, Bme680.DefaultI2cAddress);
            var i2cDevice = I2cDevice.Create(i2cSettings);

            using (var bme680 = new Bme680(i2cDevice))
            {
                while (true)
                {
                    // 10 consecutive measurement with default settings
                    for (var i = 0; i < 10; i++)
                    {
                        // Perform the measurement
                        var result = bme680.Read();

                        // Print out the measured data
                        double altValue = 0;
                        if (result.Pressure != null && result.Temperature != null)
                        {
                            altValue = WeatherHelper.CalculateAltitude((Pressure)result.Pressure, defaultSeaLevelPressure, (Temperature)result.Temperature);
                        }

                        Console.WriteLine($"Gas resistance: {result.GasResistance:0.##}Ohm");
                        Console.WriteLine($"Temperature: {result.Temperature?.DegreesCelsius:0.#}\u00B0C");
                        Console.WriteLine($"Pressure: {result.Pressure?.Hectopascals:0.##}hPa");
                        Console.WriteLine($"Altitude: {altValue:0.##}m");
                        Console.WriteLine($"Relative humidity: {result.Humidity:0.#}%");

                        // WeatherHelper supports more calculations, such as saturated vapor pressure, actual vapor pressure and absolute humidity.
                        if (result.Temperature != null && result.Humidity != null)
                        {
                            Console.WriteLine($"Heat index: {WeatherHelper.CalculateHeatIndex((Temperature)result.Temperature, (double)result.Humidity).DegreesCelsius:0.#}\u00B0C");
                            Console.WriteLine($"Dew point: {WeatherHelper.CalculateDewPoint((Temperature)result.Temperature, (double)result.Humidity).DegreesCelsius:0.#}\u00B0C");
                        }

                        // when measuring the gas resistance on each cycle it is important to wait a certain interval
                        // because a heating plate is activated which will heat up the sensor without sleep, this can
                        // falsify all readings coming from the sensor
                        Thread.Sleep(1000);
                    }
                }
            }
        }
```

@MarkCiliaVincenti @krwq @joperezr

RobinTTY:
Also as a note: Users that choose to skip samplings are pobably experienced enough to remember that those values are not "usable". So setting it to the maximum value is probably safe.

MarkCiliaVincenti:
In reality, if you choose to skip pressure, will it actually do something less with the new methods? If it involves the same communication with the sensor, we may want to consider not allowing the developer to skip sampling.

RobinTTY:
Yeah, the measurement will be faster if samplings are skipped, with the new methods even more so because we don't always have to do an additional temperature measurement to update the "tfine" value.

MarkCiliaVincenti:
Man, I'm really looking forward to this being merged. A bit of a bummer regarding the NaN. I know it's not ideal, but I'd personally prefer using nullable rather than a maxvalue. I understand it makes it a bit more tedious to use, but it's definitely neater. @krwq @joperezr

RobinTTY:
Will finish this as soon as we get a response from @krwq or @joperezr. 😄

MarkCiliaVincenti:
If one would still want to keep on using NaN, you'd need code like:

````
var temp = readings.Temperature?.Value.DegreesCelsius ?? double.NaN;

File: src/devices/Bmxx80/ReadResult/Bmp280ReadResult.cs [hunk 5]
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Contains a measurement result of a Bmp280 sensor.
+    /// </summary>
+    public class Bmp280ReadResult
+    {
+        /// <summary>
+        /// Collected temperature measurement. NaN if no measurement was performed.
+        /// </summary>
+        public Temperature Temperature;
```
krwq, line 17: since this is public API this should be a property (`Temperature Temperature { get; set; }`) - here and in all similar places

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -157,6 +159,7 @@ public void Reset()
         /// Contains <see cref="double.NaN"/> otherwise.
         /// </param>
         /// <returns><code>true</code> if measurement was not skipped, otherwise <code>false</code>.</returns>
+        [Obsolete("TryReadTemperature is deprecated, please use Read or ReadAsync instead.")]
```
krwq, line 162: IMO you should remove Obsolete attribute, there is nothing wrong with this method as is.

krwq: same for other places

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -271,6 +241,53 @@ protected override void SetDefaultConfiguration()
             StandbyTime = StandbyTime.Ms125;
         }
 
+        /// <summary>
+        /// Performs a temperature reading.
+        /// </summary>
+        /// <returns><see cref="Temperature"/></returns>
+        protected bool TryReadTemperatureInternal(out Temperature temperature)
```
krwq, line 248: Our design guidelines say to name similar protected methods as `TryReadTemperatureCore`

src/devices/Bmxx80/Bme680.cs
```diff
@@ -583,5 +583,65 @@ private byte CalculateHeaterDuration(ushort duration)
...
+                humidity = double.NaN;
+                return false;
+            }
+
+            // Read humidity data.
+            var hum = Read16BitsFromRegister((byte)Bme680Register.HUMIDITYDATA, Endianness.BigEndian);
+
+            if (!skipTempFineRead)
+            {
+                TryReadTemperatureInternal(out _);
```
krwq, line 600: should you return false when this returns false?

src/devices/Bmxx80/Bme680.cs
```diff
@@ -341,6 +344,40 @@ public int GetMeasurementDuration(Bme680HeaterProfile profile)
...
+        /// Performs a synchronous reading.
+        /// </summary>
+        /// <returns><see cref="Bme680ReadResult"/></returns>
+        public Bme680ReadResult Read()
+        {
+            SetPowerMode(Bme680PowerMode.Forced);
+            Thread.Sleep(GetMeasurementDuration(HeaterProfile));
+
+            TryReadTemperatureInternal(out var temperature);
+            TryReadPressureInternal(out var pressure, true);
```
krwq, line 357: use explicit arg name for all calls which pass true/false since it's not obvious what it means

src/devices/Bmxx80/Bme280.cs
```diff
@@ -141,5 +169,25 @@ private double CompensateHumidity(int adcHumidity)
...
+            if (HumiditySampling == Sampling.Skipped)
+            {
+                humidity = double.NaN;
+                return false;
+            }
+
+            // Read the temperature first to load the t_fine value for compensation.
+            if (!skipTempFineRead)
+            {
+                TryReadTemperatureInternal(out _);
```
krwq, line 184: should you not be ignoring result?

src/devices/Bmxx80/Bme280.cs
```diff
@@ -81,30 +85,54 @@ public Sampling HumiditySampling
...
+        public async Task<Bme280ReadResult> ReadAsync()
         {
-            return s_osToMeasCycles[(int)PressureSampling] + s_osToMeasCycles[(int)TemperatureSampling] + s_osToMeasCycles[(int)HumiditySampling];
+            if (ReadPowerMode() != Bmx280PowerMode.Normal)
+            {
+                SetPowerMode(Bmx280PowerMode.Forced);
+                await Task.Delay(GetMeasurementDuration());
+            }
+
+            TryReadTemperatureInternal(out var temperature);
```
krwq, line 131: should you return false if any of these fails?

src/devices/Bmxx80/Bmxx80Base.cs
RobinTTY: Yeah that's fine by me. My only concern was that the new Read method can achieve the same result as these methods and often do it faster so I would discourage using them.

src/devices/Bmxx80/Bme680.cs
RobinTTY: No, this method is called to update a value wich is used for internal calculation purposes. If it returns false the value won't be updated but that's expected and is not necessarily bad. Per default sensor configuration this will always return true, only if the user chooses to skip temperature measurements will this value not get updated and this is often a conscious decission to get results faster or if the temperature doesn't change a lot.

src/devices/Bmxx80/Bme680.cs
RobinTTY: Might be good to add that to the StyleCop analyzer if possible.

src/devices/Bmxx80/Bme280.cs
RobinTTY: See https://github.com/dotnet/iot/pull/797#discussion_r432845583

src/devices/Bmxx80/Bme280.cs
RobinTTY: That's actually an open question I had. See comment https://github.com/dotnet/iot/pull/797#issuecomment-632354524. In the past if one of the measurements was skipped by the user the value would be NaN. Since UnitsNet doesn't handle NaN we need another solution (null or max value).

krwq:
did you try using `default(Unit)` instead of NaN? I think that will give you invalid value. All structs will allow to do that and since they generate the unit it will produce something unusable, Looking at the code the enum they have inside is Nullable which mean default will produce null:
https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/GeneratedCode/Quantities/Temperature.g.cs#L47

src/devices/Bmxx80/Bme280.cs
krwq: If you use `default(Temperature)` as result then you can check i.e. `temp.Unit == TemperatureUnit.Undefined`

krwq: the old methods returning double can check for that and return NaN instead

RobinTTY: Using `default` would result in:

Capture

(Temperature + Humidity + Pressure measurement skipped)

krwq: I'm using

```csharp
            Temperature temp = default;
            Console.WriteLine(temp);
            Console.WriteLine(temp.Unit);
            Console.WriteLine(temp.DegreesCelsius);
```

and it prints:

```
0 K
Kelvin
-273.15
```

unfortunately it's not throwing any errors or return bogus units but you can tell immediately the value is wrong.

For Try* methods user is responsible for checking the output (similarly as if you used Dictionary with Value being int you'd get 0). For existing methods if possible we should hack it so they still work or just remove them and make the breaking change.

When returning structs where one of the fields can be null use `Temperature?`

krwq: Another option is to use: `Temperature temp = new Temperature(0, (UnitsNet.Units.TemperatureUnit)(-1));` - that is actually causing failure on any conversion or ToString:

```csharp
Temperature temp = new Temperature(0, (UnitsNet.Units.TemperatureUnit)(-1));
            Console.WriteLine(temp); // Throws NotImplementedException
            Console.WriteLine(temp.Unit); // Prints -1
            Console.WriteLine(temp.DegreesCelsius); // Throws NotImplementedException
```

another option is to create a single Invalid struct with reflection and then copy it in all places which need it

MarkCiliaVincenti:
There's another way you can do it too. So you'll have private nullable variables like private Temperature? _temperature;

And you would have properties reading those variables. If the value is null, they would throw an Exception. Eg:

public Temperature Temperature {
get { if (_temperature.HasValue) return _temperature.Value;
throw new Exception("Temperature was not read.");}}

MarkCiliaVincenti:
There's another way you can do it too. So you'll have private nullable variables like:
````
private Temperature? _temperature;
````

If the temperature reading is skipped, you would set the value of its corresponding value to null.
Then you would have public properties reading those variables. If the value is null, it throws an exception. Eg:

````
public Temperature Temperature
{
   get
   {
      if (_temperature.HasValue) return _temperature.Value;
      throw new DataException("Temperature was not read.");
   }
}
````

And here we would be able to decide on what type of exception would be the best to use. I went for DataException, but we could decide on NotSupportedException or TypeAccessException.

@krwq @RobinTTY

**Comment by RobinTTY** [comment]
> There's another way you can do it too. So you'll have private nullable variables like:
> 
> ```
> private Temperature? _temperature;
> ```
> 
> If the temperature reading is skipped, you would set the value of its corresponding value to null.
> Then you would have public properties reading those variables. If the value is null, it throws an exception. Eg:
> 
> ```
> public Temperature Temperature
> {
>    get
>    {
>       if (_temperature.HasValue) return _temperature.Value;
>       throw new DataException("Temperature was not read.");
>    }
> }
> ```
> 
> And here we would be able to decide on what type of exception would be the best to use. I went for DataException, but we could decide on NotSupportedException or TypeAccessException.
> 
> @krwq @RobinTTY

I like this approach we could do this for the double values as well, so instead of returning double.NaN we throw an exception. The exception message is also a lot more helpful than when you see that something went wrong in the UnitsNet conversion.

I also noticed that UnitsNet has units for Distance and Resistance which could have been used for the bmxx devices. (altitude -> meters, GasResistance -> resistance). Was there a reason this was not done or should we make that change as well?

**Comment by krwq** [comment]
When throwing exception from properties we should give user a way to check if it will throw, perhaps we should just expose nullable fields and perhaps make Read return some kind of status like: Success, PartialRead, Error.

**Comment by RobinTTY** [comment]
> When throwing exception from properties we should give user a way to check if it will throw, perhaps we should just expose nullable properties and perhaps make Read return some kind of status like: Success, PartialRead, Error.

@krwq  So would you be in favor of nullable properties for both old methods and new? Or leave it at non nullable and add a return code to the new Read method while leaving the old ones as is. Both approaches are fine by me.

**Comment by MarkCiliaVincenti** [comment]
Yes sure, we should have TryRead. I don't get what PartialRead means though. Is it because you specified something to be skipped or is it because it failed to get, say, the humidity (is this even possible that it gets some but not all values?)

**Comment by RobinTTY** [comment]
> Yes sure, we should have TryRead. I don't get what PartialRead means though. Is it because you specified something to be skipped or is it because it failed to get, say, the humidity (is this even possible that it gets some but not all values?) 

The sensor will only not return a certain measurement if the user specifies it should be skipped. There is no "failing" to get a measurement otherwise.

**Comment by MarkCiliaVincenti** [comment]
Well, that's what we need to handle. Let's say your sensor is unplugged. If you try to read measurements, then we should ideally have a TryRead that would return false in such a case (sensor is unplugged).

**Comment by RobinTTY** [comment]
Well if you try to create an instance of the bmxx class it will check if the sensor is present (checking i2c address). I don't know if it is helpful to check this every time you make a measurement, I don't think we do this for other devices?! I think it's ok to get a communication related exception in this case if thats the behavior for other devices as well.

**Comment by krwq** [comment]
I thought about partial read because I thought it is possible for i.e. humidity read to fail while temperature works. If it's 100% user choice that they don't want a reading then only TryRead method returning bool is enough. With properties throwing user must have a way of telling if that will fail or not (if someone unplugs the sensor while you're reading that is ok but we should then have a way of detecting that as well separately with something like Status)

**Comment by RobinTTY** [comment]
@krwq since we return the read results inside a class with the read method like this:

```C#
public Bme680ReadResult Read();

public class Bme680ReadResult
{
    public Temperature Temperature { get; }
    public Pressure Pressure { get; }
    ....
}
```

we could either just add a Status property to that class which would look something like this:

```C#
class Bme680ReadStatus
{
    bool TemperatureSampled { get; }
    bool PressureSampled { get; }
    ....
}
```

So this way you can individually check if the required sampling was performed. Or we could return the status and have the result as out parameter:

```C#
public Bme680ReadStatus Read(out Bme680ReadResult result);
```

This way it would be similar to the old methods. Would either of those options be fine with you? Aside from that I'll add a method to query if the device is still available.

krwq:
@RobinTTY would it make sense to just make fields nullable? that would make it much simpler to use

RobinTTY:
Sorry for the delay, will get to this in the next few days 😄

joperezr:
[Triage] @RobinTTY still interested in doing this? Do note that given this PR is a bit old a lot of things have changed underneath, including the fact that the bindings code is now marked as nullable so that will require some rebasing. Let us know if you are not interested in doing this so that we can close this in our next triage meeting.

joperezr:
[Triage] Friendly Ping 😄

RobinTTY:
Apologies, I will get to this today.

RobinTTY:
@joperezr Alright this is ready for review.

So what this PR achieves is, instead of doing:

```C#
bme680.SetPowerMode(Bme680PowerMode.Forced);

Thread.Sleep(measurementDuration.ToTimeSpan());

bme680.TryReadTemperature(out var tempValue);
bme680.TryReadPressure(out var preValue);
bme680.TryReadHumidity(out var humValue);
bme680.TryReadGasResistance(out var gasResistance);
```

you can now simply use:

```C#
var readResult = bme680.Read();
// or
var readResult = await bme680.ReadAsync();
```

What you get is an object which looks something like this (depending on the sensor):

```C#
    public class Bme680ReadResult : Bme280ReadResult
    {
        public Temperature? Temperature { get; }
        public Pressure? Pressure { get; }
        public Ratio? Humidity { get; }
        public ElectricResistance? GasResistance { get; }
    }
```

Values are null if the user decides he wants to skip a particular sampling.

joperezr:
Awesome Thanks for getting to it @RobinTTY, we will take a look at this soon.

src/devices/Bmxx80/Bme680.cs
```diff
@@ -435,24 +430,7 @@ public override bool TryReadTemperature(out Temperature temperature)
...
-            var gasResRaw = Read8BitsFromRegister((byte)Bme680Register.GAS_RES);
-            var gasRange = Read8BitsFromRegister((byte)Bme680Register.GAS_RANGE);
-
-            var gasRes = (ushort)((ushort)(gasResRaw << 2) + (byte)(gasRange >> 6));
-            gasRange &= (byte)Bme680Mask.GAS_RANGE;
-
-            gasResistance = CalculateGasResistance(gasRes, gasRange);
-            return true;
-        }
+        public bool TryReadGasResistance(out ElectricResistance gasResistance) => TryReadGasResistanceCore(out gasResistance);
```
krwq, line 433: What do you gain from having Core meethod? wouldn't it be better to directly call TryReadGasResistance?

krwq:
@RobinTTY does this actually fix https://github.com/dotnet/iot/issues/753 ? I didn't see any modifications or calls into SetPowerMode

krwq (approved):
LGTM, thanks, sorry for this being so stretched out in time

PR #522, opened 2019-06-24
Added support for Matrix Keyboard

HumJ0218:

PR #803, opened 2019-10-16
add key matrix (restart #522)

HumJ0218:


src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+        /// Keyboard event
+        /// </summary>
+        public event PinChangeEventHandler PinChangeEvent;
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            ClosePins();
+
+            _masterGpioController.Dispose();
```
HumJ0218: I have a question. Should I really do this when `_masterGpioController` is a common controller?

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+            _masterGpioController = null;
+
+            _outputPins = null;
+            _inputPins = null;
+            _buttonValues = null;
+        }
+
+        private void OpenPins()
```
HumJ0218: I'm not sure it's correct not to do any checks here. So do `ClosePins()`

src/devices/KeyMatrix/KeyMatrix.cs
krwq: actually as long as you don't use the same pins it's fine to create multiple GpioControllers and dispose each one of them independently

HumJ0218: If dispose here, will it affect other devices that are using this controller?

HumJ0218: sorry for deleted the comment. Does that mean, if I use different pins, I can create multiple GPIO controllers?

krwq: yes, think of GpioController as something which groups pins for usage in a specific device. I.e. LCD character display uses 8 pins then LED uses 1 etc. The recommended usage is to create one GpioController per device.

HumJ0218:
With the bad network I can't build on my computer
these code was build correctly in VS2019 but went wrong here
@krwq

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+        private bool _pinsOpened;
+
+        /// <summary>
+        /// Initialize key matrix
+        /// </summary>
+        /// <param name="masterController">GPIO controller</param>
+        /// <param name="outputPins">Output pins</param>
+        /// <param name="inputPins">Input pins</param>
+        /// <param name="scanInterval">Scanning interval in milliseconds</param>
+        public KeyMatrix(GpioController masterController, IEnumerable<int> outputPins, IEnumerable<int> inputPins, int scanInterval)
```
krwq: assuming this is the only controller being used I'd name this `gpioController` or just `controller`

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            _masterGpioController = masterController ?? throw new ArgumentNullException(nameof(masterController));
+
+            if (outputPins == null)
+            {
+                throw new ArgumentNullException(nameof(outputPins));
+            }
+            if (!outputPins.Any())
+            {
+                throw new ArgumentOutOfRangeException(nameof(outputPins), "The number of outputs is at least 1");
```
krwq: `must be at least 1`

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+        /// <param name="scanInterval">Scanning interval in milliseconds</param>
+        public KeyMatrix(GpioController masterController, IEnumerable<int> outputPins, IEnumerable<int> inputPins, int scanInterval)
+        {
+            _masterGpioController = masterController ?? throw new ArgumentNullException(nameof(masterController));
+
+            if (outputPins == null)
+            {
+                throw new ArgumentNullException(nameof(outputPins));
+            }
+            if (!outputPins.Any())
```
krwq: enters after `}` (here and below)

src/devices/KeyMatrix/KeyMatrix.cs
krwq: same comment for inputPins

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+            ScanInterval = scanInterval;
+
+            OpenPins();
+        }
+
+        /// <summary>
+        /// Start a task to open GPIO pins then start keyboard scanning. Raises KeyMatrixEventArgs.
+        /// </summary>
+        /// <param name="token">A cancellation token that can be used to cancel the work</param>
+        public Task ScanAsync(CancellationToken token)
```
krwq: 1. Please provide only sync version of this. Task.Run doesn't buy much in terms of async and user can do it themselves.
2. What's ScanAsync actually doing? Isn't scanInterval causing scans to happen automatically? When would you want to call it rather than waiting for auto scan?

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+                KeyMatrixEventArgs args = Scan(token, false);
+                return args;
+            }, token);
+        }
+
+        /// <summary>
+        /// Start a task to open GPIO pins then start keyboard scanning. End with a KeyMatrixEventArgs raises.
+        /// </summary>
+        /// <param name="token">A cancellation token that can be used to cancel the work</param>
+        public Task<KeyMatrixEventArgs> ReadKeyAsync(CancellationToken token)
```
krwq: only sync version please

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+        /// <summary>
+        /// Start a task to open GPIO pins then start keyboard scanning. Raises KeyMatrixEventArgs.
+        /// </summary>
+        /// <param name="token">A cancellation token that can be used to cancel the work</param>
+        public Task ScanAsync(CancellationToken token)
+        {
+            return Task.Run(() =>
+            {
+                KeyMatrixEventArgs args = Scan(token, false);
+                return args;
```
krwq: what's happening with these args? This method doesn't return anything

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+
+        /// <summary>
+        /// Start a task to open GPIO pins then start keyboard scanning. End with a KeyMatrixEventArgs raises.
+        /// </summary>
+        /// <param name="token">A cancellation token that can be used to cancel the work</param>
+        public Task<KeyMatrixEventArgs> ReadKeyAsync(CancellationToken token)
+        {
+            return Task.Run(() =>
+            {
+                KeyMatrixEventArgs args = Scan(token, true);
```
krwq: considering you're auto scanning (at least that's what scanInterval suggests) why do you need to manually scan?

src/devices/KeyMatrix/KeyMatrix.cs
krwq: this should be `_masterGpioController?.Dispose();` considering you're nulling it out

src/devices/KeyMatrix/KeyMatrix.cs
krwq: fine for private methods assuming you do checks earlier

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+                _masterGpioController.OpenPin(_outputPins[i], PinMode.Output);
+            }
+            for (int i = 0; i < _inputPins.Length; i++)
+            {
+                _masterGpioController.OpenPin(_inputPins[i], PinMode.Input);
+            }
+
+            _pinsOpened = true;
+        }
+        private void ClosePins()
```
krwq: enter after `}`

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,211 @@
...
+                            return args;
+                        }
+                    }
+                }
+
+                _masterGpioController.Write(_outputPins[currentOutput], PinValue.Low);
+                currentOutput = (currentOutput + 1) % _outputPins.Length;
+                _masterGpioController.Write(_outputPins[currentOutput], PinValue.High);
+
+                // Console.Title = string.Join(" ", _buttonValues.Select(m => (int)m));
```
krwq: remove

src/devices/KeyMatrix/KeyMatrix.cs
krwq: reading code further, should you just call Scan in the constructor? If not you should probably call this `StartScanning` and move the `scanInterval` to this method instead of having it in the constructor

src/devices/KeyMatrix/samples/KeyMatrix.Sample.cs
```diff
@@ -0,0 +1,109 @@
...
+            Console.WriteLine("KeyMatrix.Dispose() ");
+
+            // quit
+            Console.ReadKey();
+        }
+
+        /// <summary>
+        /// Keyboard event
+        /// </summary>
+        private static void Mk_PinChangeEvent(object sender, KeyMatrixEventArgs pinValueChangedEventArgs)
```
krwq: what does `Mk` stand for? do you actually need the prefix?

krwq:
@HumJ0218 not trying to hurry or anything just wanted to know if you will have time to address feedback anytime soon

HumJ0218:
@krwq sorry for that. I'm busy for my job and I will be back ASAP

krwq:
No problem @HumJ0218

krwq:
@HumJ0218 do you think you'll have time to pick this up anytime soon?

krwq:
@HumJ0218 friendly ping

HumJ0218:
I'm back! Sorry for left behind

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,179 @@
...
+                    int index = _currentOutput * _inputPins.Length + i;
+
+                    PinValue oldValue = _buttonValues[index];
+                    PinValue newValue = _gpioController.Read(_inputPins[i]);
+                    _buttonValues[index] = newValue;
+
+                    if (newValue != oldValue)
+                    {
+                        args = new KeyMatrixEvent(newValue == PinValue.High ? PinEventTypes.Rising : PinEventTypes.Falling, _currentOutput, i);
+                        return args;
```
krwq, line 124: should you also call `_gpioController.Write(_outputPins[_currentOutput], PinValue.Low);` here?

src/devices/KeyMatrix/samples/KeyMatrix.Sample.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// This sample is for Raspberry Pi Model 3B+
+    /// </summary>
+    internal class Program
+    {
+        public static void Main(string[] args)
+        {
+            // get arguments
+            Console.Write("input output pins(eg. 26,19,13,6) ");
+            var line = Console.ReadLine();
+            IEnumerable<int> outputs = (string.IsNullOrEmpty(line) ? "26,19,13,6" : line).Split(',').Select(m => int.Parse(m));
```
krwq, line 23: (not tested)

```suggestion
            IEnumerable<int> outputs = string.IsNullOrEmpty(line) ? new int[] { 26, 19, 13, 6 }: line.Split(',').Select(m => int.Parse(m));
```

src/devices/KeyMatrix/samples/KeyMatrix.Sample.cs
```diff
@@ -0,0 +1,98 @@
...
+            Console.Write("input input pins(eg. 21,20,16,12) ");
+            line = Console.ReadLine();
+            IEnumerable<int> inputs = (string.IsNullOrEmpty(line) ? "21,20,16,12" : line).Split(',').Select(m => int.Parse(m));
```
krwq, line 26: similar suggestion as for outputs

src/devices/KeyMatrix/samples/KeyMatrix.Sample.cs
krwq: also consider m.Trim() in case someone puts spaces

src/devices/KeyMatrix/samples/KeyMatrix.Sample.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    /// <summary>
+    /// This sample is for Raspberry Pi Model 3B+
+    /// </summary>
+    internal class Program
+    {
+        public static void Main(string[] args)
+        {
+            // get arguments
+            Console.Write("input output pins(eg. 26,19,13,6) ");
```
krwq, line 21: not sure I understand what this message says. Perhaps "Enter output pins (default: 26, 19, 13, 6)", same comment below for inputs

src/devices/KeyMatrix/samples/KeyMatrix.Sample.cs
```diff
@@ -0,0 +1,98 @@
...
+namespace Iot.Device.KeyMatrix.Samples
+{
+    /// <summary>
+    /// This sample is for Raspberry Pi Model 3B+
+    /// </summary>
+    internal class Program
+    {
+        public static void Main(string[] args)
+        {
+            // get arguments
```
krwq, line 20: remove comment as it doesn't add any value

src/devices/KeyMatrix/samples/KeyMatrix.Sample.cs
```diff
@@ -0,0 +1,98 @@
...
+            int interval = int.TryParse(line, out int i) ? i : 15;
+            Console.Write("input read key event count(eg. 20) ");
+            line = Console.ReadLine();
+            int count = int.TryParse(line, out int c) ? c : 20;
+
+            // get GPIO controller
+            GpioController gpio = new GpioController();
+
+            // you can also use other GPIO controller
+            /*
```
krwq, line 38: I don't think this is that useful here, you can put it as a sample on the README page though

src/devices/KeyMatrix/README.md
```diff
@@ -0,0 +1,29 @@
+# Key Matrix
```
krwq, line 1: you should also create another README.md in the samples folder and put some tiny snippets on how to use this

src/devices/KeyMatrix/samples/README.md
```diff
@@ -0,0 +1,9 @@
+# Key Matrix Samples
+
+## Connection using on-board GPIO
```
krwq, line 3: You should add tiny snippets of code here explaining the usage

krwq:
@HumJ0218 sorry for late review, overall looks good to me, my main piece of feedback is to add snippets to the samples/README.md

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,179 @@
...
+        private int _currentOutput = 0;
+
+        /// <summary>
+        /// Initialize key matrix
+        /// </summary>
+        /// <param name="gpioController">GPIO controller</param>
+        /// <param name="outputPins">Output pins</param>
+        /// <param name="inputPins">Input pins</param>
+        /// <param name="scanInterval">Scanning interval in milliseconds</param>
+        public KeyMatrix(GpioController gpioController, IEnumerable<int> outputPins, IEnumerable<int> inputPins, int scanInterval)
```
Ellerbach, line 59: Maybe TimeSpan() for the scanInterval?

Ellerbach, line 59: Wondering if you want as well to implement a constructor with creating automatically a GpioController? See here https://github.com/dotnet/iot/tree/master/src/devices#gpiocontroller

src/devices/KeyMatrix/KeyMatrix.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        /// <summary>
+        /// Get all buttons' values
+        /// </summary>
+        public ReadOnlySpan<PinValue> Values => _buttonValues.AsSpan();
+
+        /// <summary>
+        /// Get or set interval in milliseconds
+        /// </summary>
+        public int ScanInterval { get; set; }
```
Ellerbach, line 37: I would really go for a TimeSpan, you can have an internal millisecond int if you want

krwq: good catch!

HumJ0218:
@Ellerbach My intention was to slow down the scan interval appropriately, and the interval in milliseconds was sufficient and did not require more precision. Because if the scan interval is too short, the CPU will be under a heavy load. Based on my actual tests, for a 4x4 key matrix, the interval of about 15 milliseconds does not affect usage, and reducing the interval only increases CPU load.
So, do I need to change the input parameter type to `TimeSpan` while the internal value still uses `int` millisecond value?

Ellerbach:
> @Ellerbach My intention was to slow down the scan interval appropriately, and the interval in milliseconds was sufficient and did not require more precision. Because if the scan interval is too short, the CPU will be under a heavy load. Based on my actual tests, for a 4x4 key matrix, the interval of about 15 milliseconds does not affect usage, and reducing the interval only increases CPU load.
> So, do I need to change the input parameter type to `TimeSpan` while the internal value still uses `int` millisecond value?

Yes, please, replace for TimeStpan and please add your finding about the right timing to wait in the description and initialize it with a default working condition. So it will be simple and efficient. We really try to use TimeSpan for all timings when they are in a public interface.

joperezr:
[Triage] @HumJ0218 still interested in doing this. If not, let us know so that we can close it in our next triage meeting.

HumJ0218:
@joperezr working on it, and new posts this weekend

joperezr:
[Triage] Sounds good, make sure you rebase since we have made changes on Nullability that may very well have broken the current state of your code.

krwq:
Hey @HumJ0218 there is still couple of comments not resolved (i.e. TimeSpan comments from @Ellerbach), please make sure to fix/respond to all of them (do not resolve ones where you didn't fix, just respond and let us resolve or respond back)

joperezr:
[Triage] Friendly ping 😄

joperezr:
[Triage] Hey @HumJ0218 let us know if you are still interested in continuing the work on this.

joperezr:
[Triage] Hey @HumJ0218 please let us know if you are interested in finishing this, since we'll close it if we don't hear anything from you in the next few weeks. Also if we do that, you could always reopen once you are ready to continue working on it.

src/devices/KeyMatrix/samples/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+        continue;
+    }
+
+    ShowKeyMatrixEvent(mk, key);
+}
+
+Console.WriteLine("This will now start listening to events and display them. Press a key to finish.");
+mk.KeyEvent += KeyMatrixEventReceived;
+mk.StartListeningKeyEvent();
+while (!Console.KeyAvailable)
```
krwq, line 49: I'd prefer the key matrix APIs looked similar to console, so i.e.:

```csharp
while (keyMatrix.KeyAvailable)
{
  char key = keyMatrix.ReadKey();
}
```

treat as optional but I think having consistent APIs would be ideal

krwq (approved):
see my comment about APIs resembling console (optional to fix)

src/devices/KeyMatrix/samples/Program.cs
Ellerbach: it's not that easy to do. ReadKey() return a `KeyMatrixEvent` object which contains the coordinate of the key and the timeline. There are few differences with a keyboard which makes it complicated to implement:

- There is no key mapping available
- There is not key buffer, so you can't really detect if a key is pressed or not and then read the various keys

That would make the matrix class a bit more complex. Why not in the future if there is traction.

krwq: sounds good

Issue #776, opened 2019-10-08
IsPinOpen() doesn't work for output pins


pawelpetruch:
When using System.Device.Gpio(1.0.0) on Raspbian Buster (RPi 2) in the following way:

```
private GpioController controller = new GpioController();
               
controller.OpenPin(pin, PinMode.Output);

```
and when I check the pin's status:
`controller.IsPinOpen(pin)`
it returns false when the pin is actually working fine. 

Input pins are showed correctly when IsPinOpen() is used.

- dotnet 3.0.100
- Version of System.Device.Gpio package 1.0.0

krwq:
Thanks for reporting @pawelpetruch! Quickly looking at the code I'm not sure how would that be possible. Can you show some small sample code? Are you sure you're not accidentally creating new instance of GpioController and disposing the old one?

pawelpetruch:
Thanks for the response @krwq!

A sample code:
```
class XXX
{
        private GpioController controller = new GpioController();
        
        public XXX(int pin)
        {
            controller.OpenPin(pin, PinMode.Output);
            if (controller.IsPinOpen(pin))
            {
                ...
            }
            else
            {
                ...
            }
        }
}
```

krwq:
@pawelpetruch thanks, I'll have this scenario tested either as part of https://github.com/dotnet/iot/pull/758 or sometime afterwards.

joperezr:
@krwq so is there really a problem here or what is it that we want to track for vNext? From the code that I see above, I can't seem to find a reason why this would be happening.

krwq:
At minimum we should add a test verifying this scenario works. If it's broken as suggested per issue we should fix it for vNext.

pgrawehr:
Uploaded a testcase of this Bug, but was unable to reproduce it.

krwq:
@pgrawehr do you still have any project where you can reproduce? Is it possible you're referencing some older version of System.Device.Gpio?

pgrawehr:
This was tested with the latest master build. So far, I was unable to reproduce the bug with anything I tried.

krwq:
@pawelpetruch can you help us write a test case for this issue? PR with test case is in here https://github.com/dotnet/iot/pull/806

pawelpetruch:
@krwq I will have a look over the weekend if that's fine.

krwq:
@pawelpetruch weekend sounds good, thank you!

krwq:
[...]

PR #806, opened 2019-10-17
Test for Bug #776

pgrawehr:
This is a test case for Bug #776, but it did not show the described buggy behavior. The test passes just fine for me.

### DIFF
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -89,8 +89,9 @@ namespace System.Device.Gpio.Tests
         {
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
+                // Open pin in input mode (default)
                 Assert.False(controller.IsPinOpen(LedPin));
-                controller.OpenPin(LedPin);
+                controller.OpenPin(LedPin, PinMode.Input);
                 Assert.True(controller.IsPinOpen(LedPin));
                 controller.ClosePin(LedPin);
                 Assert.False(controller.IsPinOpen(LedPin));
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -85,7 +85,7 @@ namespace System.Device.Gpio.Tests
         }
 
         [Fact]
-        public void IsPinOpenTest()
+        public void IsPinOpenOnInputTest()
         {
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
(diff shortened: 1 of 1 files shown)


src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -95,7 +95,22 @@ public void IsPinOpenTest()
                 controller.ClosePin(LedPin);
                 Assert.False(controller.IsPinOpen(LedPin));
             }
-        }
+			using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
```
krwq: can you create a separate test case? If using `using` statement twice is needed to repro please still do it separately

krwq:
@pgrawehr what do you mean it doesn't repro, are you doing anything different than in the test? Do you have a full project/cs file which you can repro? (even if it requires running in a loop or anything)

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -95,7 +95,22 @@ public void IsPinOpenTest()
...
+			{
+				Assert.False(controller.IsPinOpen(LedPin));
+
+				controller.OpenPin(LedPin, PinMode.Output);
+				Assert.True(controller.IsPinOpen(LedPin));
```
joperezr: Also please make sure that you use spaces instead of tabs.

pgrawehr: Yea, sorry. My VS is set to tabs by default, but for some reason the .editorConfig setting is ignored.

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
pgrawehr: Created as a separate test case.

pgrawehr:
I just tried to reproduce the issue described in the ticket. It said that the IsPinOpen function would not work when the Pin was in Output mode (which is the difference to the existing test). 

I therefore suggest the ticket be closed as "Not reproducible". The test may make sense nevertheless, I think. Maybe we do get the problem sometime.

krwq:
Yes, I agree we should add explicit test case. Btw. initially when I reviewed I mistakenly thought you were the author of the issue which was confusing

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -97,6 +97,27 @@ public void IsPinOpenTest()
             }
         }
 
+        [Fact]
+        public void IsPinOpenTest2()
```
krwq: I'd rename this and previous test so that it's more obvious what's the difference between this and previous test

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -85,7 +85,7 @@ public void ThrowsInvalidOperationExceptionWhenPinIsNotOpened()
         }
 
         [Fact]
-        public void IsPinOpenTest()
+        public void IsPinOpenOnInputTest()
```
joperezr: One last ask, can you also please change line 93 on this file so that we explicitly open the pin using Input mode? Right now it is not setting any mode which is defaulting to Input, but given the change on the test name, it would be good if we explicitly set the mode when opening.

krwq: especially this is opening LED which looks like it's opening for output

PR #812, opened 2019-10-19
Fix GPIO access issue with sysfs caused by udev setting permissions.

### LINKED ISSUE #811: Issue with using gpio with SYSFS with non-root privileges.
**Describe the bug**

This bug is where OpenPin is called with a pin number and a PinMode with the SYSFS driver using a non-root account but one that is in the gpio group.

The first time that the program is executed there will be a permissions exception but subsequent attempts to execute will work.

This issue was described in #459 and it seems that it is because udev takes a finite amount of time between the device being created and the correct permissions being applied. It was closed as people were reluctant to implement the proposed solution. This applies to Raspbian. 

I think, however that this issue should be re-examined because.

* I believe that the implied contract for OpenPin would meant that it should fail if subsequent valid operations on the pin would be failing.
* I think that SYSFS should behave in a consistent manner.
* Similar issues can occur with the default RaspberryPi driver as SYSFS is used when events are attached to a pin.

It is my intention to create a PR that delays OpenPin long enough for subsequent operations to work but only does this if necessary and will only delay for a limited amount of time.

**Steps to reproduce**

Execute the code below twice..

```csharp
            GpioController c = new GpioController(PinNumberingScheme.Logical, SysFsDriver.Create());
            c.OpenPin(16, PinMode.Input);
```

**Expected behavior**

I expect that the pin would be opened.

**Actual behavior**

Describe what actually happened

A System.IO.IOException Permission denied was thrown the first time.

Add following information:

 Version:   3.0.100
 Commit:    04339c3a26

Runtime Environment:
 OS Name:     raspbian
 OS Version:  10
 OS Platform: Linux
 RID:         linux-arm
 Base Path:   /opt/dotnet/sdk/3.0.100/

Host (useful for support):
  Version: 3.0.0
  Commit:  7d57652f33

.NET Core SDKs installed:
  2.1.802 [/opt/dotnet/sdk]
  3.0.100 [/opt/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.13 [/opt/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.13 [/opt/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.0.0 [/opt/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.13 [/opt/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.0.0 [/opt/dotnet/shared/Microsoft.NETCore.App]

- `dotnet --info` on the machine where app is being run (not applicable for self-contained apps)

 Version:   3.0.100
 Commit:    04339c3a26

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18363
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.0.100\

Host (useful for support):
  Version: 3.0.0
  Commit:  7d57652f33

.NET Core SDKs installed:
  1.0.0 [C:\Program Files\dotnet\sdk]
  2.0.2 [C:\Program Files\dotnet\sdk]
  2.1.200 [C:\Program Files\dotnet\sdk]
  2.1.201 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.400 [C:\Program Files\dotnet\sdk]
  2.1.401 [C:\Program Files\dotnet\sdk]
  2.1.402 [C:\Program Files\dotnet\sdk]
  2.1.403 [C:\Program Files\dotnet\sdk]
  2.1.500 [C:\Program Files\dotnet\sdk]
  2.1.502 [C:\Program Files\dotnet\sdk]
  2.1.503 [C:\Program Files\dotnet\sdk]
  2.1.504 [C:\Program Files\dotnet\sdk]
  2.1.505 [C:\Program Files\dotnet\sdk]
  2.1.507 [C:\Program Files\dotnet\sdk]
  2.1.508 [C:\Program Files\dotnet\sdk]
  2.1.509 [C:\Program Files\dotnet\sdk]
  2.1.600-preview-009426 [C:\Program Files\dotnet\sdk]
  2.1.600-preview-009472 [C:\Program Files\dotnet\sdk]
  2.1.600-preview-009497 [C:\Program Files\dotnet\sdk]
  2.1.600 [C:\Program Files\dotnet\sdk]
  2.1.601 [C:\Program Files\dotnet\sdk]
  2.1.602 [C:\Program Files\dotnet\sdk]
  2.1.604 [C:\Program Files\dotnet\sdk]
  2.1.700-preview-009618 [C:\Program Files\dotnet\sdk]
  2.1.700 [C:\Program Files\dotnet\sdk]
  2.1.701 [C:\Program Files\dotnet\sdk]
  2.1.800-preview-009696 [C:\Program Files\dotnet\sdk]
  2.1.801 [C:\Program Files\dotnet\sdk]
  2.1.802 [C:\Program Files\dotnet\sdk]
  2.2.204 [C:\Program Files\dotnet\sdk]
  3.0.100-preview7-012821 [C:\Program Files\dotnet\sdk]
  3.0.100-preview8-013656 [C:\Program Files\dotnet\sdk]
  3.0.100 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.8 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.9 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.11 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.13 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.8 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.9 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.11 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.13 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.2.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.0.0-preview8.19405.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.0.0 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 1.0.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 1.1.1 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.3-servicing-26724-03 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.8 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.11 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.12 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.13 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.2.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.0.0-preview8-28405-07 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.0.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.0.0-preview8-28405-07 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.0.0 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

- Version of `System.Device.Gpio` 1.0.0

Frankenslag:
Fixes #811 

This PR aims to resolve issues when OpenPin is called as a non-root user on a Raspberry Pi running Raspian whilst using the SYSFS driver. The user is correctly in the gpio group.

The first time OpenPin is called it creates a node in the sysfs file system for the pin but the permissions are not set for the gpio group. 

Udev is used to set the permissions but those doesn't happen instantly.

 The implication of this is that often the first operation on the pin will fail and if retried for a second time will succeed

This PR tests if the access is correct to the node representing the pin has the correct permissions and if it hasn't retries a limited number of times until all is well.

Users with root privileges should pass the access test right away and not have to wait for the permissions to be applied.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
@@ -78,6 +78,8 @@ namespace System.Device.Gpio.Drivers
         /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
         protected internal override void OpenPin(int pinNumber)
         {
+            const int numOpenRetries = 100; // wait 1 second max for udev to set the permissions
+            int openRetries;
             int pinOffset = pinNumber + s_pinOffset;
             string pinPath = $"{GpioBasePath}/gpio{pinOffset}";
             // If the directory exists, this becomes a no-op since the pin might have been opened already by the some controller or somebody else.
@@ -86,6 +88,27 @@ namespace System.Device.Gpio.Drivers
                 try
                 {
                     File.WriteAllText(Path.Combine(GpioBasePath, "export"), pinOffset.ToString(CultureInfo.InvariantCulture));
+
+                    // test to see if we have rw access to the GPIO pin.
+                    if ( Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0 )
+                    {
+                        openRetries = 0;
+
+                        // repeatedly check for access to the GPIO pin until access is granted or the number of tries is exceeded
+                        do
+                        {
+                            // delay a little to give udev potentially time to respond
+                            Thread.Sleep(TimeSpan.FromMilliseconds(10));
+                            openRetries++;
+                        } while ( Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0 & openRetries < numOpenRetries );
+
+                        // if we have run out of retries then throw an exception. 
+                        if( openRetries >= numOpenRetries )
+                        {
+                            throw new UnauthorizedAccessException("Opening pins requires root permissions.");
+                        }
+                    }
+
                     _exportedPins.Add(pinNumber);
                 }
                 catch (UnauthorizedAccessException e)
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
@@ -78,8 +78,8 @@ namespace System.Device.Gpio.Drivers
         /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
         protected internal override void OpenPin(int pinNumber)
         {
-            const int numOpenRetries = 100; // wait 1 second max for udev to set the permissions
-            int openRetries;
+            const int retryDelayMaximumMillis = 1000; // wait 1 second max for udev to set the permissions
+            Diagnostics.Stopwatch retryStopwatch;
             int pinOffset = pinNumber + s_pinOffset;
             string pinPath = $"{GpioBasePath}/gpio{pinOffset}";
             // If the directory exists, this becomes a no-op since the pin might have been opened already by the some controller or somebody else.
@@ -90,20 +90,22 @@ namespace System.Device.Gpio.Drivers
                     File.WriteAllText(Path.Combine(GpioBasePath, "export"), pinOffset.ToString(CultureInfo.InvariantCulture));
 
                     // test to see if we have rw access to the GPIO pin.
-                    if ( Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0 )
+                    if (Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0)
                     {
-                        openRetries = 0;
+                        retryStopwatch = new Diagnostics.Stopwatch();
+                        retryStopwatch.Start();
 
-                        // repeatedly check for access to the GPIO pin until access is granted or the number of tries is exceeded
+                        // repeatedly check for access to the GPIO pin until access is granted or the timeout is exceeded
                         do
                         {
                             // delay a little to give udev potentially time to respond
                             Thread.Sleep(TimeSpan.FromMilliseconds(10));
-                            openRetries++;
-                        } while ( Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0 & openRetries < numOpenRetries );
+                        } while (Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0 & retryStopwatch.ElapsedMilliseconds < retryDelayMaximumMillis);
 
-                        // if we have run out of retries then throw an exception. 
-                        if( openRetries >= numOpenRetries )
+                        retryStopwatch.Stop();
+
+                        // if we have run out of time for the permissions to be set then throw an exception. 
+                        if (retryStopwatch.ElapsedMilliseconds >= retryDelayMaximumMillis)
                         {
                             throw new UnauthorizedAccessException("Opening pins requires root permissions.");
                         }
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
@@ -78,8 +79,7 @@ namespace System.Device.Gpio.Drivers
         /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
         protected internal override void OpenPin(int pinNumber)
         {
-            const int retryDelayMaximumMillis = 1000; // wait 1 second max for udev to set the permissions
-            Diagnostics.Stopwatch retryStopwatch;
+            const int RetryDelayMaximumMillis = 1000; // wait 1 second max for udev to set the permissions
             int pinOffset = pinNumber + s_pinOffset;
             string pinPath = $"{GpioBasePath}/gpio{pinOffset}";
             // If the directory exists, this becomes a no-op since the pin might have been opened already by the some controller or somebody else.
@@ -92,22 +92,21 @@ namespace System.Device.Gpio.Drivers
                     // test to see if we have rw access to the GPIO pin.
                     if (Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0)
                     {
-                        retryStopwatch = new Diagnostics.Stopwatch();
-                        retryStopwatch.Start();
+                        Stopwatch retryStopwatch = Stopwatch.StartNew();
 
                         // repeatedly check for access to the GPIO pin until access is granted or the timeout is exceeded
                         do
                         {
                             // delay a little to give udev potentially time to respond
                             Thread.Sleep(TimeSpan.FromMilliseconds(10));
-                        } while (Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0 & retryStopwatch.ElapsedMilliseconds < retryDelayMaximumMillis);
+                        } while (Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0 & retryStopwatch.ElapsedMilliseconds < RetryDelayMaximumMillis);
 
                         retryStopwatch.Stop();
 
                         // if we have run out of time for the permissions to be set then throw an exception. 
-                        if (retryStopwatch.ElapsedMilliseconds >= retryDelayMaximumMillis)
+                        if (retryStopwatch.ElapsedMilliseconds >= RetryDelayMaximumMillis)
                         {
-                            throw new UnauthorizedAccessException("Opening pins requires root permissions.");
+                            throw new UnauthorizedAccessException("Timeout exceeded waiting for access to the pin.");
                         }
                     }
 
--- a/src/System.Device.Gpio/Interop/Unix/Libc/Interop.access.cs
+++ b/src/System.Device.Gpio/Interop/Unix/Libc/Interop.access.cs
@@ -1,0 +1,19 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+
+internal partial class Interop
+{
+    [DllImport(LibcLibrary, SetLastError = true)]
+    internal static extern int access([MarshalAs(UnmanagedType.LPStr)] string pathname, AccessModeFlags flags);
+}
+
+internal enum AccessModeFlags
+{
+    F_OK = 0,
+    X_OK = 1,
+    W_OK = 2,
+    R_OK = 4
+}


joperezr:
@Frankenslag can a user on the gpio group perform all the rest of the controller operations when they are not root using SysFS? So far, we have documented that we require root permissions in order for your app to work as expected, which is the same as most of the other libraries for controlling gpio. I would like to understand how common it is that people run into this because the use a user on the gpio group to run their apps instead of root, since I wouldn't want to optimize for this scenario if it isn't common, since adding this will potentially slow down opening pins for the rest of the people using root.

joperezr:
Also, if we were to take this, we would have to validate if this will work in OSes other than Raspbian, since the SysFS driver is meant to be used across different Linux distros, which may not have access export or may behave differently.

Frankenslag:
> Also, if we were to take this, we would have to validate if this will work in OSes other than Raspbian, since the SysFS driver is meant to be used across different Linux distros, which may not have access export or may behave differently.

I would be happy to test where I can on a pi but before spending the time I would like get general agreement that the change is a good idea. I hope I have designed it so that it *would* work on other OS's with sysfs without problems because the sysfs seems to be the same it is just that Raspbian is providing this ability to use without root privs.

Frankenslag:
> @Frankenslag can a user on the gpio group perform all the rest of the controller operations when they are not root using SysFS? So far, we have documented that we require root permissions in order for your app to work as expected, which is the same as most of the other libraries for controlling gpio.

The best I can tell they can. Below are the udev rules that implement that and you can see that this certainly appears to be intended. There are quite a few mentions of this on the internet. I believe this rule gets set up the first time that you use raspi-config.

```
SUBSYSTEM=="gpio", GROUP="gpio", MODE="0660"
SUBSYSTEM=="gpio*", PROGRAM="/bin/sh -c '\
        chown -R root:gpio /sys/class/gpio && chmod -R 770 /sys/class/gpio;\
        chown -R root:gpio /sys/devices/virtual/gpio && chmod -R 770 /sys/devices/virtual/gpio;\
        chown -R root:gpio /sys$devpath && chmod -R 770 /sys$devpath\
'"

```

> I would like to understand how common it is that people run into this because the use a user on the gpio group to run their apps instead of root, since I wouldn't want to optimize for this scenario if it isn't common, since adding this will potentially slow down opening pins for the rest of the people using root.

I have certainly run into this issue on a number of occasions but just put it down to a glitch as it worked the second time. It can appear when not expected though as the standard driver falls back to sysfs if you ask for a rising or falling event and that can be really confusing because you have not asked to use the sysfs driver. 

I don't think it will slow down (much). If users have root access then it's a simple permissions check and that's all. The delays happen only if the correct access is not detected.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -78,6 +78,8 @@ public SysFsDriver()
         /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
         protected internal override void OpenPin(int pinNumber)
         {
+            const int numOpenRetries = 100; // wait 1 second max for udev to set the permissions
```
krwq: Can we instead define max time this will wait? Thread.Sleep is not really good way to measure time and I think this will sometimes take 1s and other time 15s or more.

(Check permissions immediately and then if you need to wait then create a stopwatch and check time once in a while).

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -86,6 +88,27 @@ protected internal override void OpenPin(int pinNumber)
...
+                        // repeatedly check for access to the GPIO pin until access is granted or the number of tries is exceeded
+                        do
+                        {
+                            // delay a little to give udev potentially time to respond
+                            Thread.Sleep(TimeSpan.FromMilliseconds(10));
+                            openRetries++;
+                        } while ( Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0 & openRetries < numOpenRetries );
+
+                        // if we have run out of retries then throw an exception. 
+                        if( openRetries >= numOpenRetries )
```
krwq:
```suggestion
                        if (openRetries >= numOpenRetries)
```

krwq: please remove extra spaces everywhere

Frankenslag: done

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
Frankenslag: OK I have done this. On my PI 3b it takes only 1 retry so < 10ms for udev to do its thing PR updated

@joperezr @krwq 

I seem to be having repeated problems with the automated checks where everything other than the build release passes.

joperezr: yeah, if you follow the link of the failed build into Azure DevOps, and click on the Tests Tab you'll see that a test failed. We run our tests in actual hardware which causes sometimes for the tests to be a bit flaky, which we are planning to fix. I have restarted the build in the meantime.

pgrawehr:
@krwq, @joperezr Are there still objections against merging this? This would especially remove the requirement to run the Raspi driver as root when using interrupt-driven pin inputs. To make the SysFs driver work nicely together with the Raspi driver, this change is important. 
I've done some tests, and with some small updates, the Raspi driver (with the help of the SysFs driver) will be able to handle pin interrupts at frequencies of several khz, without root rights.

joperezr:
This change is not really fixing the access on the when not root right? This seems like it is just adding some retries to try to get a setting to kick in. And also it looks like this will only work if you do copy the file in order to set the udev setting. Because of this, I would rather not take this and instead document the behavior and provide the workaround. If this were on its own actually fixing the issue and allowing non-root access then that would be amazing, but that doesn't seem to be the case. Please correct me if my understanding is wrong though.

Frankenslag:
> get a setting to kick in. And also it looks like this will only work if you do copy the file in order to set the udev setting. Because of this, I would rather not take this and instead document the behavior and provide

@joperezr 

This change will work on Raspbian with just a normal configuration.. you don't have to copy files. - not sure where that came from.

What happens is that when a pin is opened on raspbian then *as standard* UDEV changes the permissions to allow non root access - it just takes a few milliseconds to do this but can break any code that uses the pin right away.

This is why the PR waits for the permissions to be altered before continuing.

I hope this clears things up.

krwq:
After reading various threads seems like retrying is a standard practice in case of udev controlled permissions so I'm generally fine with the fix

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -78,6 +78,8 @@ public SysFsDriver()
         /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
         protected internal override void OpenPin(int pinNumber)
         {
+            const int retryDelayMaximumMillis = 1000; // wait 1 second max for udev to set the permissions
```
krwq: `RetryDelayMaximumMilliseconds`

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -78,6 +78,8 @@ public SysFsDriver()
...
+            Diagnostics.Stopwatch retryStopwatch;
```
krwq: this should be declared as late as possible

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -86,6 +88,29 @@ protected internal override void OpenPin(int pinNumber)
                 try
                 {
                     File.WriteAllText(Path.Combine(GpioBasePath, "export"), pinOffset.ToString(CultureInfo.InvariantCulture));
+
+                    // test to see if we have rw access to the GPIO pin.
+                    if (Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0)
+                    {
+                        retryStopwatch = new Diagnostics.Stopwatch();
```
krwq: `using System.Diagnostics` and use just type name

krwq: Also you can use Stopwatch.StartNew();

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -86,6 +88,29 @@ protected internal override void OpenPin(int pinNumber)
...
+                            // delay a little to give udev potentially time to respond
+                            Thread.Sleep(TimeSpan.FromMilliseconds(10));
+                        } while (Interop.access(pinPath, AccessModeFlags.R_OK | AccessModeFlags.W_OK) != 0 & retryStopwatch.ElapsedMilliseconds < retryDelayMaximumMillis);
+
+                        retryStopwatch.Stop();
+
+                        // if we have run out of time for the permissions to be set then throw an exception. 
+                        if (retryStopwatch.ElapsedMilliseconds >= retryDelayMaximumMillis)
+                        {
+                            throw new UnauthorizedAccessException("Opening pins requires root permissions.");
```
krwq: this exception will be caught by the try catch and wrapped with exactly the same exception which feels kinda weird...

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
Frankenslag: Done in latest PR.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
Frankenslag: Done in latest PR update

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
Frankenslag: Changed to indicate that the wait for permissions has timed out which will then be combined with the outer exception in the PR update.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
Frankenslag: Changed in the latest update for the PR

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
Frankenslag: @krwq 

Please feel free to review again.

krwq:
@joperezr this is just fixing udev permission race - I couldn't find any better way to check whether udev is done or not. Any objections merging this as is?

pgrawehr:
> This change will work on Raspbian with just a normal configuration.. you don't have to copy files. - not sure where that came from.

I guess the confusion came a bit from #813, where the udev rule needs to be manually installed (for now). Otherwise, it's exactly the same thing there.

krwq:
@Frankenslag I'll merge this as is since I'll refactor this a bit so I can reuse it in https://github.com/dotnet/iot/pull/876

joperezr:
I'm fine with the fix as well, the only part I would like to change is how we use stopwatch here to control the retries, which will just spin up a new thread and over complicate the retry logic. Instead, I would just have a maxNumberOfRetries integer and just have a `while (maxNumberOfRetries-- > 0)` or something like that, and maybe perform a Thread.Sleep between each retry or something like that.

krwq:
@joperezr see my latest modification to this PR, instead of number of retries we went with maximum time we will spend retrying but we won't spin up new thread for stopwatch (using `Stopwatch.GetTimestamp()` and will Thread.Sleep to avoid 100% CPU usage)

PR #815, opened 2019-10-21
Do not dispose GPIO Controller if it's not our own

pgrawehr:
If the reference to the GPIO Controller was provided from outside, we must not dispose it in the Dispose method, otherwise this closes any other pins we might have open at the same time.

### DIFF
--- a/src/devices/SoftPwm/SoftwarePwmChannel.cs
+++ b/src/devices/SoftPwm/SoftwarePwmChannel.cs
@@ -35,8 +35,8 @@ namespace System.Device.Pwm.Drivers
 
         private Thread _runningThread;
         private GpioController _controller;
-        private bool _runThread = true;
-        private bool _shouldDispose;
+        private readonly bool _shouldDispose;
+		private bool _runThread = true;
 
         /// <summary>
         /// The frequency in hertz.
--- a/src/devices/SoftPwm/SoftwarePwmChannel.cs
+++ b/src/devices/SoftPwm/SoftwarePwmChannel.cs
@@ -36,6 +36,7 @@ namespace System.Device.Pwm.Drivers
         private Thread _runningThread;
         private GpioController _controller;
         private bool _runThread = true;
+        private bool _ownController;
 
         /// <summary>
         /// The frequency in hertz.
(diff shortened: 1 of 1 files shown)


krwq:
@pgrawehr actually I believe you should be always creating new instance of GpioController and any GpioController instance passed in to any device should be assumed to be owned by the device and should always be disposed (similar to I2cDevice and SpiDevice).

This might be a bit of misdesign on our part since we weren't sure how we would want it to work at first and we didn't clarify the ownership until the very end of the release cycle and since this was already similar to how I2cDevice, SpiDevice worked we had decided to ship like that and now it's only for back compat.

See https://github.com/dotnet/iot/blob/master/Documentation/Devices-conventions.md#lifetime-management-or-what-should-be-disposed-by-what

cc: @joperezr

pgrawehr:
Although it's a bit confusing (I was assuming that GpioController should basically be considered a singleton for the lifetime of the application), that's an approach I can live with. 
Downside is that you can now create multiple GpioController instances which handle the same pin. This may lead to subtle bugs.

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -197,8 +209,11 @@ protected override void Dispose(bool disposing)
             _runThread = false;
             _runningThread?.Join();
             _runningThread = null;
-            _controller?.Dispose();
-            _controller = null;
+            if (_shouldDispose)
+            {
+                _controller?.Dispose();
+                _controller = null;
+            }
```
krwq: nit: enter after `}` and before `if` statement (here and below)

pgrawehr: Done, but should add StyleCop for that (I guess that's why you created #816).

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -36,6 +36,7 @@ public class SoftwarePwmChannel : PwmChannel
         private Thread _runningThread;
         private GpioController _controller;
         private bool _runThread = true;
+        private bool _shouldDispose;
```
joperezr: NIT: This should be readonly.

joperezr:
Thanks a bunch for the contribution @pgrawehr!

Issue #809, opened 2019-10-19
API Suggestion for Low-Level GPIO Drivers


pgrawehr:
Since the issue addressed in #788 seems to require wider discussion, let's talk about how it should look like to the user in the end (and how that concept should be implemented).

# Overview of the Problems
The primary problem is that the Raspberry Pi4 requires a few things (in particular: Setting the pull up/pull down mode of input pins) to be done differently than on the Pi3. (Not sure about even older models, and whether we support them.)

This leads to two main topics: 
1. How do we organize the two drivers? Which classes/methods should be exposed to the user?
2. How do we detect the correct driver to use? Should we allow the user to manually select a driver? Should there be a way of finding out whether a driver matches a given platform?

# Problem 1: Driver class structure
The current proposal is
```
public abstract class GpioDriver;
    public abstract class GenericRaspberryPiDriver : GpioDriver;*
        public class RaspberryPi3Driver: GenericRaspberryPiDriver;
        public class RaspberryPi4Driver: GenericRaspberryPiDriver;
   public class SysFsDriver: GpioDriver;
   ....

* Note: Additional intermediate class on Windows: Windows10Driver
```
This would expose a new abstract GenericRaspberryPiDriver, as well as the new RaspberryPi4Driver. 

# Problem 2: Detect the correct driver to use / select the desired driver
## Situation
Currently, GpioController has an internal static method `GetBestDriverForBoard()` which returns (after some quite ugly magic, see also #766) a reference to the best matching `GpioDriver` for the current hardware/operating system. `GetBestDriverForBoard()` is called when `GpioController` is created without explicitly specifiying a driver. The user can, however, create instances of any of the Driver classes and use them as argument to the GpioController constructor. 

## Issues with the current solution
The downside of this approach is that the user doesn't know what he actually gets when he uses the BestDriver approach. When he wants to use the "second best" driver (for whatever reason, maybe for testing), he doesn't know which to use. 
If a selected driver doesn't work (be it automatically selected or manually), there's no defined way of knowing, unless probing and hoping not to get an exception (or worse, an access violation). 

## Suggestion
My suggestion would be to add a `CheckDriverIsCompatible()` method to each driver, indicating whether this driver is supported on this hardware (and with the currently installed software components). Then, we only need to specify somewhere a preference order in which `GetBestDriverForBoard()` probes them. In the concrete example, the Pi3Driver would only report true on the Pi3, the Pi4 driver on the Pi4. This could also allow a cleaner solution to #766. 

[...]

PR #823, opened 2019-10-24
Support Pull-Up and Pull-Down resistor setting for the Pi4

pgrawehr:
This is a quick patch for the issue already addressed in #788, but comes without any API changes.
Should be correctly done after the result of discussion #809.

### DIFF
--- a/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.Linux.cs
+++ b/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.Linux.cs
@@ -9,8 +9,30 @@ namespace System.Device.Gpio.Tests
 {
     public class RaspberryPiDriverTests : GpioControllerTestBase
     {
+        /// <summary>
+        /// Leave this pin open (unconnected) for the tests
+        /// </summary>
+        private const int OpenPin = 23;
+
         protected override GpioDriver GetTestDriver() => new RaspberryPi3Driver();
 
         protected override PinNumberingScheme GetTestNumberingScheme() => PinNumberingScheme.Logical;
+
+        /// <summary>
+        /// Tests for setting the pull up/pull down resistors on the Raspberry Pi (supported on Pi3 and Pi4, but with different techniques)
+        /// </summary>
+        [Fact]
+        public void InputPullResistorsWork()
+        {
+            using (GpioController controller = new GpioController(GetTestNumberingScheme()))
+            {
+                controller.OpenPin(OpenPin, PinMode.InputPullUp);
+                Assert.Equal(PinValue.High, controller.Read(OpenPin));
+                controller.SetPinMode(OpenPin, PinMode.InputPullDown);
+                Assert.Equal(PinValue.Low, controller.Read(OpenPin));
+                controller.SetPinMode(OpenPin, PinMode.InputPullUp);
+                Assert.Equal(PinValue.High, controller.Read(OpenPin));
+            }
+        }
     }
 }
--- a/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.Linux.cs
+++ b/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.Linux.cs
@@ -24,7 +24,7 @@ namespace System.Device.Gpio.Tests
         [Fact]
         public void InputPullResistorsWork()
         {
-            using (GpioController controller = new GpioController(GetTestNumberingScheme()))
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 controller.OpenPin(OpenPin, PinMode.InputPullUp);
                 Assert.Equal(PinValue.High, controller.Read(OpenPin));
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -30,6 +30,15 @@ namespace System.Device.Gpio.Drivers
         private UnixDriver _sysFSDriver = null;
         private readonly IDictionary<int, PinMode> _sysFSModes = new Dictionary<int, PinMode>();
 
+        /// <summary>
+        /// Returns true if this is a Raspberry Pi4
+        /// </summary>
+        protected bool IsPi4
+        {
+            get;
+            private set;
+        }
+
         protected override void Dispose(bool disposing)
         {
             if (_registerViewPointer != null)
@@ -214,6 +223,12 @@ namespace System.Device.Gpio.Drivers
         /// <param name="mode">The mode of a pin to set the resistor pull up/down mode.</param>
         private void SetInputPullMode(int pinNumber, PinMode mode)
         {
+            if (IsPi4)
+            {
+                SetInputPullModePi4(pinNumber, mode);
+                return;
+            }
+
             byte modeToPullMode = mode switch
             {
                 PinMode.Input => (byte)0,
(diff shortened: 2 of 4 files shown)


src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -30,6 +30,15 @@ public unsafe partial class RaspberryPi3Driver : GpioDriver
         private UnixDriver _sysFSDriver = null;
         private readonly IDictionary<int, PinMode> _sysFSModes = new Dictionary<int, PinMode>();
 
+        /// <summary>
+        /// Returns true if this is a Raspberry Pi4
+        /// </summary>
+        protected bool IsPi4
```
krwq: I'd rather create an enum for this, I'm expecting there will be new version eventually and I don't necessarily want us to have `IsPi4` .. `IsPi15`. Also might be easier to first discuss these changes in the issue

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -472,6 +512,21 @@ private void Initialize()
...
+                IsPi4 = false;
+                try
+                {
+                    string model = File.ReadAllText("/proc/device-tree/model", Text.Encoding.ASCII);
+                    if (model.Contains("Raspberry Pi 4"))
+                    {
+                        IsPi4 = true;
+                    }
+                }
+                catch
```
krwq: what exceptions do we expect? Can we prevent them with i.e. File.Exists?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: This PR is considered a quick fix, because I consider it a quite critical issue, making the whole library use problematic on the Pi4, so I just used the fastest approach.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: Maybe we can (File.Exists + permission check maybe) but I wouldn't know how to test the behavior on all possible other systems.

krwq: Let's do following:
- guard against anything we know might go wrong (at minimum File.Exists)
- leave try catch but add Debug.Assert in there so it at least fails in debug (make sure to add a comment)

pgrawehr:
@krwq Review issues addressed, I think.

src/System.Device.Gpio.Tests/RaspberryPiDriverTests.Linux.cs
```diff
@@ -9,8 +9,30 @@ namespace System.Device.Gpio.Tests
...
 
         protected override PinNumberingScheme GetTestNumberingScheme() => PinNumberingScheme.Logical;
+
+        /// <summary>
+        /// Tests for setting the pull up/pull down resistors on the Raspberry Pi (supported on Pi3 and Pi4, but with different techniques)
+        /// </summary>
+        [Fact]
+        public void InputPullResistorsWork()
+        {
+            using (GpioController controller = new GpioController(GetTestNumberingScheme()))
```
krwq: should you also pass `GetTestDriver()`?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: public/protected members in this library will need to be supported forever. This will likely get forgotten if we don't fix right now

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -27,9 +27,20 @@ public unsafe partial class RaspberryPi3Driver : GpioDriver
         private const string GpioMemoryFilePath = "/dev/gpiomem";
         private const string MemoryFilePath = "/dev/mem";
         private const string DeviceTreeRanges = "/proc/device-tree/soc/ranges";
-        private UnixDriver _sysFSDriver = null;
+		private const string ModelFilePath = "/proc/device-tree/model";
```
krwq: indentation is off

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -262,6 +279,31 @@ private void SetInputPullMode(int pinNumber, PinMode mode)
...
+        /// Sets the resistor pull up/down mode for an input pin on the Raspberry Pi4. 
+        /// The above, complex method doesn't do anything on a Pi4 (it doesn't cause any harm, though)
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        /// <param name="mode">The mode of a pin to set the resistor pull up/down mode.</param>
+        protected void SetInputPullModePi4(int pinNumber, PinMode mode)
+        {
+            int shift = (pinNumber & 0xf) << 1;
+            UInt32 pull = 0;
+            UInt32 bits = 0;
```
krwq: uint

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -472,7 +514,28 @@ private void Initialize()
...
+							IsPi4 = true;
+						}
+					}
+                }
+                catch (Exception x)
+				{
+                    // This should not normally fail, but we currently don't know how this behaves on different operating systems. Therefore, we ignore
+					// any exceptions in release and just continue as Pi3 if something fails. 
+                    // If in debug mode, we might want to check what happened here (i.e unsupported OS, incorrect permissions)
+					System.Diagnostics.Debug.Fail($"Unexpected exception: {x}");
```
krwq: `using System.Diagnostics`

pgrawehr: @krwq I just saw that `Debug.Fail` has no effect, since DEBUG is not defined. Do we need to add something like 
```
<PropertyGroup Condition="$(Configuration.EndsWith('Debug'))=='true'">
    <DefineConstants>DEBUG;TRACE</DefineConstants>
</PropertyGroup>
```
Or should this be provided by the build scripts?

krwq: `dotnet build -c Debug` should define it I believe. Create an issue if it still doesn't work, let's refrain from these kind of changes here

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -262,6 +280,31 @@ private void SetInputPullMode(int pinNumber, PinMode mode)
             *gppudclkPointer = 0;
         }
 
+        /// <summary>
+        /// Sets the resistor pull up/down mode for an input pin on the Raspberry Pi4. 
+        /// The above, complex method doesn't do anything on a Pi4 (it doesn't cause any harm, though)
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        /// <param name="mode">The mode of a pin to set the resistor pull up/down mode.</param>
+        protected void SetInputPullModePi4(int pinNumber, PinMode mode)
```
krwq: private/internal

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: Created #881. Something is fishy there.

Issue #784, opened 2019-10-09
VS2019 .sln


MarkCiliaVincenti:
How possible would it be to have a VS2019 .sln including Iot.Device.Bindings and all the samples so folks using VS2019 on Windows don't need to go through hurdles to contribute to this project? It'd be nice if we could just fork, branch, clone and open in VS, do changes, right click and 'Commit'.

Frankenslag:
@MarkCiliaVincenti 
What problems are you facing. I open the folder. Right click on the root project and build the forked folder. It’s not as easy as using a full solution but seems ok.

MarkCiliaVincenti:
Well let's see. They say a picture is worth a thousand words.

First I fork dotnet/iot and create my own branch WeatherHelper.
![step1](https://user-images.githubusercontent.com/25120736/66542910-24383180-eb34-11e9-88e0-11585df3ba0a.PNG)
And click clone & download and "open in Visual Studio"

![step2](https://user-images.githubusercontent.com/25120736/66542939-3914c500-eb34-11e9-84a2-9c53c61cd0f8.PNG)
Then I click Clone.

Then I have to switch to team explorer
select branches
expand remotes/origin and double click on my branch WeatherHelper
Back to solution explorer
Then from the startup item I choose "Iot.Device.Bindings.csproj"
Then what? If I open files, I won't get syntax errors and I can't F12.

![step3](https://user-images.githubusercontent.com/25120736/66543144-d7088f80-eb34-11e9-9cbd-e8fa91545d98.png)

Frankenslag:
@MarkCiliaVincenti 
OK. I do it slightly differently. I don't really know if it is correct as I found conflicting discussions on the web.

* Fork the dotnet/iot into your github area i.e. Frankenslag/iot
* From VS2019 then Clone or Checkout code from Frankenslag/iot to a local folder
* Go into settings for the local repositary and add a remote called upstream with the url https://github.com/dotnet/iot.git. This is useful if you want to keep your local master up to date
* Now from within VS create a new local branch for your current work. Ensure that this new branch is now the current one.
* From within the source explorer right click on the root folder and open a developer command prompt.
* type build.cmd This will make sure all the tooling is in place and up to date within your new branch. This may take some minutes. I see a bunch of errors during the build but I think the important thing is that the tooling has been installed
* right click on the root build.proj and select rebuild to clean and build the projects.

Now when I go into the source code I get syntax highlighting and pressing F12 takes me to a definition etc.

[...]

PR #825, opened 2019-10-27
4 Bit LCD Display support

pgrawehr:
This adds support for character LCD displays that are only connected trough 4 of the 8 data lines. 

This was tested with this display: https://www.joy-it.net/de/products/SBC-LCD20x4 which uses an I2C interface, but works only with 4-bit commands.

### DIFF
--- a/src/devices/CharacterLcd/CharacterLcd.sln
+++ b/src/devices/CharacterLcd/CharacterLcd.sln
@@ -1,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29409.12
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CharacterLcd", "CharacterLcd.csproj", "{A455537F-ECF1-4280-98CD-FFC6F027EA16}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CharacterLcd.Samples", "samples\CharacterLcd.Samples.csproj", "{A642926F-A33B-4BC7-9976-8A7754DF3375}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{A455537F-ECF1-4280-98CD-FFC6F027EA16}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{A455537F-ECF1-4280-98CD-FFC6F027EA16}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{A455537F-ECF1-4280-98CD-FFC6F027EA16}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{A455537F-ECF1-4280-98CD-FFC6F027EA16}.Release|Any CPU.Build.0 = Release|Any CPU
+		{A642926F-A33B-4BC7-9976-8A7754DF3375}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{A642926F-A33B-4BC7-9976-8A7754DF3375}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{A642926F-A33B-4BC7-9976-8A7754DF3375}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{A642926F-A33B-4BC7-9976-8A7754DF3375}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {BF8A0758-8ADC-47DC-A172-AC986F303A94}
+	EndGlobalSection
+EndGlobal
--- a/src/devices/CharacterLcd/Lcd1602.cs
+++ b/src/devices/CharacterLcd/Lcd1602.cs
@@ -35,8 +35,18 @@ namespace Iot.Device.CharacterLcd
         /// This is for on-chip I2c support. For connecting via I2c GPIO expanders, use the GPIO constructor <see cref="Lcd1602(int, int, int[], int, float, int, GpioController)"/>.
         /// </remarks>
         /// <param name="device">The I2c device for the LCD.</param>
-        public Lcd1602(I2cDevice device)
-            : base(new Size(16, 2), LcdInterface.CreateI2c(device))
+        /// <param name="uses8Bit">True if the device uses 8 Bit commands, false if it handles only 4 bit commands.</param>
+        public Lcd1602(I2cDevice device, bool uses8Bit)
+            : base(new Size(16, 2), LcdInterface.CreateI2c(device, uses8Bit))
+        {
+        }
+
+        /// <summary>
+        /// Constructs a new 16x2 LCD controller with the given interface
+        /// </summary>
+        /// <param name="interface">The LCD Interface</param>
+        public Lcd1602(LcdInterface @interface)
+            : base(new Size(16, 2), @interface)
         {
         }
     }
--- a/src/devices/CharacterLcd/Lcd1602.cs
+++ b/src/devices/CharacterLcd/Lcd1602.cs
@@ -44,9 +44,9 @@ namespace Iot.Device.CharacterLcd
         /// <summary>
         /// Constructs a new 16x2 LCD controller with the given interface
         /// </summary>
-        /// <param name="interface">The LCD Interface</param>
-        public Lcd1602(LcdInterface @interface)
-            : base(new Size(16, 2), @interface)
+        /// <param name="lcdInterface">The LCD Interface</param>
+        public Lcd1602(LcdInterface lcdInterface)
+            : base(new Size(16, 2), lcdInterface)
         {
         }
     }
--- a/src/devices/CharacterLcd/Lcd1602.cs
+++ b/src/devices/CharacterLcd/Lcd1602.cs
@@ -36,7 +36,7 @@ namespace Iot.Device.CharacterLcd
         /// </remarks>
         /// <param name="device">The I2c device for the LCD.</param>
         /// <param name="uses8Bit">True if the device uses 8 Bit commands, false if it handles only 4 bit commands.</param>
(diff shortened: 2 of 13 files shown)


src/devices/CharacterLcd/CharacterLcd.sln
```diff
@@ -0,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
krwq: please remove sln file

src/devices/CharacterLcd/Lcd1602.cs
```diff
@@ -35,8 +35,18 @@ public Lcd1602(int registerSelectPin, int enablePin, int[] dataPins, int backlig
...
+        public Lcd1602(I2cDevice device, bool uses8Bit)
+            : base(new Size(16, 2), LcdInterface.CreateI2c(device, uses8Bit))
+        {
+        }
+
+        /// <summary>
+        /// Constructs a new 16x2 LCD controller with the given interface
+        /// </summary>
+        /// <param name="interface">The LCD Interface</param>
+        public Lcd1602(LcdInterface @interface)
```
krwq: can you use `lcdInterface` to avoid `@`

src/devices/CharacterLcd/LcdInterface.I2c.cs
```diff
@@ -59,7 +59,17 @@ public I2c(I2cDevice device)
...
+            public override bool BacklightOn 
+            {
+                get
+                {
+                    // Not implemented / Not supported, but don't throw exceptions here
+                    return true;
+                }
+                set
+                {
+                }
```
krwq: I'd leave this as it was instead override this in the new LCD and make getter return true/false and setter throw `NotSupportedException`

src/devices/CharacterLcd/CharacterLcd.sln
pgrawehr: I actually left it in due to #784, but I guess the decission on where to proceed there is still open.

src/devices/CharacterLcd/Lcd1602.cs
pgrawehr: Ok. The existing code was like this as well, but I can change that.

src/devices/CharacterLcd/LcdInterface.I2c.cs
pgrawehr: That's a bit problematic, because then the client(s) always have to know which underlying interface they have. If you want to initialize the display (which usually means enabling the backlight), you don't want to test this at a higher level, I think.

src/devices/CharacterLcd/Lcd1602.cs
krwq: usually it's fine to leave if it matches the existing style (you don't need to fix all the comments I write if you feel it doesn't make sense - I might miss some details or lack some context when reviewing). Since you've already changed I like the new version better :)

src/devices/CharacterLcd/LcdInterface.I2c.cs
krwq: I don't feel strongly about it, fine to leave as is. We can re-iterate on that if we find someone with opposite opinion

src/devices/CharacterLcd/CharacterLcd.sln
krwq: ok, fine to leave since we have a tracking issue to clarify this

src/devices/CharacterLcd/LcdInterface.I2c.cs
```diff
@@ -59,7 +59,17 @@ public I2c(I2cDevice device)
 
             public override bool EightBitMode => true;
 
-            public override bool BacklightOn { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+            public override bool BacklightOn 
+            {
+                get
+                {
+                    // Not implemented / Not supported, but don't throw exceptions here
```
krwq: please explain why not in the comment (i.e. `we currently don't have a way of checking if backlight is supported and therefore exceptions will not be expected by the user code`)

src/devices/CharacterLcd/LcdInterface.I2c4Bit.cs
```diff
@@ -0,0 +1,138 @@
...
+            private void Write4Bits(byte command)
+            {
+                // _device.WriteByte((byte)(command | BacklightFlag));
+                _device.WriteByte((byte)(command | ENABLE | BacklightFlag));
+                _device.WriteByte((byte)((command & ~ENABLE) | BacklightFlag));
+            }
+
+            public override void SendCommands(ReadOnlySpan<byte> commands)
+            {
+                foreach(var c in commands)
```
krwq: nit space after `foreach`

src/devices/CharacterLcd/LcdInterface.I2c4Bit.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            public override void SendCommand(byte command)
+            {
+                Write4Bits((byte)(0x00 | (command & 0xF0)));
+                Write4Bits((byte)(0x00 | (command << 4 & 0xF0)));
+            }
+
+            private void Write4Bits(byte command)
+            {
+                // _device.WriteByte((byte)(command | BacklightFlag));
```
krwq: remove commented out code

src/devices/CharacterLcd/LcdInterface.cs
```diff
@@ -133,9 +133,17 @@ public static LcdInterface CreateGpio(int registerSelectPin, int enablePin, int[
         /// This is for on-chip I2c support. For connecting via I2c GPIO expanders, use the GPIO interface <see cref="CreateGpio(int, int, int[], int, float, int, GpioController)"/>.
         /// </remarks>
         /// <param name="device">The I2c device for the LCD.</param>
-        public static LcdInterface CreateI2c(I2cDevice device)
+        /// <param name="uses8Bit">True if the device uses 8 Bit commands, false if it handles only 4 bit commands.</param>
+        public static LcdInterface CreateI2c(I2cDevice device, bool uses8Bit)
```
krwq: I'd make this default to true since 4 bit is not usual

krwq: (also for back compat)

src/devices/CharacterLcd/samples/CharacterLcd.Sample.cs
```diff
@@ -17,11 +20,12 @@ class Program
         static void Main(string[] args)
         {
             // Sets up a 16x2 character LCD with a hardwired or no backlight.
-            using (Lcd1602 lcd = new Lcd1602(registerSelectPin: 22, enablePin: 17, dataPins: new int[] { 25, 24, 23, 18 }))
-            {
-                lcd.Clear();
-                lcd.Write("Hello World");
-            }
+            //using (Lcd1602 lcd = new Lcd1602(registerSelectPin: 22, enablePin: 17, dataPins: new int[] { 25, 24, 23, 18 }))
```
krwq: why is this commented out?

src/devices/CharacterLcd/samples/Hd44780.ExtendedSample.cs
```diff
@@ -19,87 +21,73 @@ class ExtendedSample
...
-                TestPrompt("CharacterSet", lcd, CharacterSet);
+            TestPrompt("DisplayEnable", lcd, DisplayAndBackLightOnOff);
 
-                // Shifting
-                TestPrompt("Autoshift", lcd, AutoShift);
-                TestPrompt("DisplayLeft", lcd, l => ShiftDisplayTest(l, a => a.ShiftDisplayLeft()));
-                TestPrompt("DisplayRight", lcd, l => ShiftDisplayTest(l, a => a.ShiftDisplayRight()));
-                TestPrompt("CursorLeft", lcd, l => ShiftCursorTest(l, a => a.ShiftCursorLeft()));
-                TestPrompt("CursorRight", lcd, l => ShiftCursorTest(l, a => a.ShiftCursorRight()));
+            // Shifting
```
krwq: why are the changes here needed? Would it make sense to instead add separate sample?

src/devices/CharacterLcd/samples/CharacterLcd.Sample.cs
pgrawehr: Only one example can be used at a time (it's highly unlikely one has all possible variants of an LCD display attached at the same time) and I changed this to work with mine... 
I'll put this into a functions, so that people testing (or playing with) their displays just need to switch a single line.

src/devices/CharacterLcd/samples/Hd44780.ExtendedSample.cs
pgrawehr: Part of it is just whitespace, but the main change is that I moved the display initialization code outside, so that the tests can be run against any interface (8 bit, 4 bit, direct or I2C). But I think you are right, I should split this long method up a bit.

src/devices/CharacterLcd/LcdInterface.I2c.cs
pgrawehr: Updated the comment. It seems that with an 8 Bit I2C driver, this bit cannot be set anywhere, so the backlight is supposed to be on all the time (or is controlled externally).

src/devices/CharacterLcd/LcdInterface.I2c4Bit.cs
pgrawehr: Should have StyleCop for that...

src/devices/CharacterLcd/samples/CharacterLcd.Sample.cs
pgrawehr: Cleaned up, I hope this is more clear now.

src/devices/CharacterLcd/LcdInterface.I2c4Bit.cs
```diff
@@ -0,0 +1,137 @@
...
+                Write4Bits((byte)(REGISTERSELECT | (value & 0xF0)));
+                Write4Bits((byte)(REGISTERSELECT | (value << 4 & 0xF0)));
+            }
+
+            public override void SendData(ReadOnlySpan<byte> values)
+            {
+                foreach (var c in values)
+                {
+                    Write4Bits((byte)(REGISTERSELECT | (c & 0xF0)));
+                    Write4Bits((byte)(REGISTERSELECT | (c << 4 & 0xF0)));
```
krwq: Please add brackets here: `c << 4 & 0xF0` and above - otherwise it's unclear what order this will execute (I think `&` has higher priority)

pgrawehr: You are right, I'll clarify that. However << comes before & (see i.e. https://www.programiz.com/csharp-programming/operator-precedence-associativity); otherwise this code wouldn't work...

pgrawehr:
@krwq : Can you please restart the build? This is very unlikely related to my changes.

krwq:
@pgrawehr yes, I don't believe it's related.

PR #839, opened 2019-11-06
Add Lidar Lite V3 Distance Sensor

vornet:
This is a basic i2c implementation for the LIDAR Lite V3 sensor by Garmin.

Technical spec available here: http://static.garmin.com/pumac/LIDAR_Lite_v3_Operation_Manual_and_Technical_Specifications.pdf


src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+            _i2cDevice = i2cDevice;
+            Reset();            
+        }
+
+        /// <summary>
+        /// Reset FPGA, all registers return to default values
+        /// </summary>
+        public void Reset() 
+        {
+            try {
```
krwq: brace on new line

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+                WriteRegister(Register.ACQ_COMMAND, 0x00);
+            } catch (System.IO.IOException)
```
krwq: is there a way to make it not throw? (i.e. send byte later or something? I haven't seen any other devices needing this)

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Buffers.Binary;
+using System.Collections.Generic;
+using System.Device;
+using System.Device.I2c;
+using System.IO;
+using System.Numerics;
+using System.Text;
+using System.Threading;
+
+namespace Iot.Device.LidarLiteV3
```
krwq: Please change namespace so that it matches for both HC-SR04 and this sensor (HC-SR04 we should change it too but not necessarily in this PR). The problem with having the same class name as namespace name is that it forces users to type both class name and namespace name since name becomes ambiguous which makes it kinda annoying to use

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+                // This exception is expected due to the reset.                
+            }
+        }
+
+        /// <summary>
+        /// Measure distance
+        /// </summary>
+        public ushort MeasureDistance(bool withReceiverBiasCorrection = true) 
+        {
+            if(withReceiverBiasCorrection) {
```
krwq: space after `if`, brace on new line (same below, `else` should be on new line etc)

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+                WriteRegister(Register.ACQ_COMMAND, 0x04);
+            } else {
+                WriteRegister(Register.ACQ_COMMAND, 0x03);
+            }
+
+            SystemStatus status = GetStatus();                
+            while(status.HasFlag(SystemStatus.BusyFlag))
```
krwq: space after `while`

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+            try {
+                WriteRegister(Register.ACQ_COMMAND, 0x00);
+            } catch (System.IO.IOException)
+            {
+                // This exception is expected due to the reset.                
+            }
+        }
+
+        /// <summary>
+        /// Measure distance
```
krwq: please mention units

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// </summary>
+        public ushort MeasureDistance(bool withReceiverBiasCorrection = true) 
+        {
+            if(withReceiverBiasCorrection) {
+                WriteRegister(Register.ACQ_COMMAND, 0x04);
+            } else {
+                WriteRegister(Register.ACQ_COMMAND, 0x03);
+            }
+
+            SystemStatus status = GetStatus();
```
krwq: what's the expected time to wait here? If less than 12ms please consider using SpinWait for waiting instead (anywhere 1-12ms Thread.Sleep is acceptable, for less I'd go with SpinWait)

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Measure distance
+        /// </summary>
+        public ushort MeasureDistance(bool withReceiverBiasCorrection = true) 
+        {
+            if(withReceiverBiasCorrection) {
+                WriteRegister(Register.ACQ_COMMAND, 0x04);
```
krwq: This can be written as:
`WriteRegister(Register.ACQ_COMMAND, withReceiverBiasCorrection ? 0x04 : 0x03);`

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+                Thread.Sleep(1);
+            }
+
+            return GetDistanceMeasurement();
+        }
+
+        /// <summary>
+        /// Get the distance measurement.
+        /// </summary>
+        public ushort GetDistanceMeasurement() {
```
krwq: how is this different from MeasureDistance? Btw both names should be just `GetDistance` - `measurement` is implied

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+            Span<byte> rawData = stackalloc byte[2] { 0, 0 };
+            ReadBytes(Register.FULL_DELAY, rawData);
+            return BinaryPrimitives.ReadUInt16BigEndian(rawData);
+        }
+
+        /// <summary>
+        /// Get the system status
+        /// </summary>
+        public SystemStatus GetStatus() {
```
krwq: - is there any value of this being public? If not consider changing to protected/private/internal
- brace on new line

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        internal void ReadBytes(Register reg, Span<byte> readBytes)
+        {
+            _i2cDevice.WriteByte((byte)reg);
+            _i2cDevice.Read(readBytes);
+        }
+
+        /// <summary>
+        /// Cleanup everything
```
krwq: instead of summary use `</inheritdoc>`

src/devices/LidarLiteV3/LidarLiteV3.csproj
```diff
@@ -0,0 +1,22 @@
...
+      <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
+    </ProjectReference>
+    <PackageReference Include="System.Memory" Version="4.5.3" />
+    <Compile Include="*.cs" />
+    <None Include="README.md" />
+    <Compile Include="..\Common\System\Device\DelayHelper.cs" Link="DelayHelper.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="System.Numerics.Vectors" />
```
krwq: are you using this?

src/devices/LidarLiteV3/LidarLiteV3.csproj
```diff
@@ -0,0 +1,22 @@
...
+  </ItemGroup>
+
+</Project>
```
krwq: can't put comment there. Please fill the `src/devices/LidarLiteV3/README.md` with at minimum a reasonable title (see other devices) and a tiny description. README.md is used for generating the [device listing](https://github.com/dotnet/iot/blob/master/src/devices/README.md)

src/devices/LidarLiteV3/SystemStatus.cs
```diff
@@ -0,0 +1,55 @@
+using System;
+
+namespace Iot.Device.LidarLiteV3
+{
+    /// <summary>
+    /// System Status
+    /// </summary>
+    [Flags]
+    public enum SystemStatus : byte {
```
krwq: I don't think there is any value with this being public. If there is some utility here I'm missing I'd rather expose it as a separate method than making this public

src/devices/LidarLiteV3/samples/LidarLiteV3.Sample.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        // I2C address on SenseHat board
+        public const int I2cAddress = 0x5F;
+
+        public static void Main(string[] args)
+        {
+            using (var llv3 = new LidarLiteV3(CreateI2cDevice()))
+            {
+                // Take 10 measurements, each one second apart.
+                for(int i = 0; i < 10; i++) {
```
krwq: space after `for`, brace on new line

vornet:
@krwq Thanks for the feedback!  I'm going to do the PR fixes soon along with some additional features of the sensor.  Sorry for not getting back sooner, I've been busy making sure all the hardware can work together.  I'm building (hopefully world's first) autonomous sail boat to circumvent the world and this is one of several sensors on-board. I'll share additional devices back to this repo too (working on the Adafruit Rfm95 LoRa transceiver right now).  Project repo @ https://github.com/naovictoria/naovictoria

krwq:
@vornet, this is awesome idea! No worries about the delay 😄

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: This and other instances of this issue should be addressed in the new push.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: I went with `TimeOfFlight` which seems like a common name among these devices, but open for suggestions. https://en.wikipedia.org/wiki/Time_of_flight

I also ended struggling a bit with namespacing issues with all the supporting classes.  Went with a `...TimeOfFlight.Models.LidarLiteV3` namespace to avoid conflict with the other devices, but not sure if I really like this.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: this and other instances of this should be fixed.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: this and other instances of this should be fixed.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: it's a whole number in cm, so I changed the type to `int`.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: The device is busy when initially powered on for 22ms, and each measurement takes ~ 4 ms.  My thought is to add `Thread.Sleep(22)` after a PowerOn(), and use `SpinWait()` here for the 4ms.  I never used `SpinWait()` before though, how do I wait 4ms?

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Fixed!

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: I changed all registry retrival to properties, so this is just `Distance`, the current distance measurement in the device.  `MeasureDistance` is higher level, and does several registry operations to get a measurement.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Yeah, there's other flags in the system status that would be useful to the user.  For example, if the device is unable to make a measurement (completely blocking the transmitter), it would set bit 6 (Process Error).

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: fixed.

src/devices/LidarLiteV3/LidarLiteV3.csproj
vornet: removed this and other unused references (copy/pasted project file from another device)

src/devices/LidarLiteV3/LidarLiteV3.csproj
vornet: Left a more detailed README in the sample folder, and added a summary README for the device

src/devices/LidarLiteV3/SystemStatus.cs
vornet: As mention earlier, I can at least see `ProcessError` error being useful, but I see your point in making the device simplier to use.  By not exposing it, the only way for the user to access it is by modifying the code.  I think what we really need is a higher-level device class, and a lower level one that exposes more of the device if they need it?

src/devices/LidarLiteV3/samples/LidarLiteV3.Sample.cs
vornet: fixed.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: hmm, not sure why I was getting an exception before.  It now works consistently without needing this try/catch.  I'm going to remove it.

src/devices/LidarLiteV3/LidarLiteV3.cs
krwq: DistanceSensor perhaps, no strong opinion

src/devices/LidarLiteV3/LidarLiteV3.cs
krwq: We have DelayHelpers class in a repo for that although with 4ms I'm fine with Thread.Sleep (search for that for usage examples)

src/devices/LidarLiteV3/SystemStatus.cs
krwq: We don't have clear guidelines about this yet so I'm fine with both approaches, I usually prefer simpler to use since for me hardware itself is a low level abstraction

src/devices/LidarLiteV3/AcquistionModeFlag.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace Iot.Device.TimeOfFlight.Models.LidarLiteV3
+{
+    /// <summary>
+    /// Acquisition Mode control
+    /// </summary>
+    [Flags]
+    public enum AcquistionModeFlag
+    {
+        /// <summary>
+        /// 0 - Disable reference process during measurement
```
krwq: I'm not sure what the 0 and 1 refer to by looking at the description, I'd prefer the descriptions be a bit simplified, i.e.:
`Enables reference process during measurement`

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,319 @@
...
+    public class LidarLiteV3 : IDisposable
+    {
+        /// <summary>
+        /// Default address for LidarLiteV3
+        /// </summary>
+        public const byte DefaultI2cAddress = 0x62;
+
+        private GpioController _gpioController;
+        private I2cDevice _i2cDevice;
+        private int? _powerEnablePin;
```
krwq: use `-1` for `no pin` and regular `int` (we've put that in our [guidelines](https://github.com/dotnet/iot/blob/master/Documentation/Devices-conventions.md#value-returned-by-sensor) - second bullet from the last)

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,319 @@
...
+
+        /// <summary>
+        /// Initialize the LidarLiteV3
+        /// </summary>
+        /// <param name="i2cDevice">I2C device</param>
+        /// <param name="gpioController">GPIO controller</param>
+        /// <param name="powerEnablePin">The pin number used to control power to the device</param>
+        public LidarLiteV3(I2cDevice i2cDevice, GpioController gpioController = null, int? powerEnablePin = null)
```
krwq: same comment about `int?`

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,319 @@
...
+        {
+            _gpioController = gpioController;
```
krwq: we usually do:
`_gpioController = gpioController ?? new GpioController();` and remove all the null checks later (except in Dispose method)

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,319 @@
...
+        /// <summary>
+        /// Measure distance in cm
+        /// </summary>
+        /// <remarks>
+        /// Note: Do not call if running while in repetition mode.  It will block until 
+        /// repetition finishes (forever if infinite).
+        /// </remarks>
+        /// <param name="withReceiverBiasCorrection">Faster without bias correction, but more prone to errors if condition changes.</param>
+        /// <returns>Distance in cm</returns>
+        public int MeasureDistance(bool withReceiverBiasCorrection = true)
```
krwq: This could actually be an Async method (note: you should also provide synchronous version which can call async).

You can use `await Task.Delay(1)` instead of sleep and rest would remain the same (except signature)

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,319 @@
...
+            // Disable the default address
+            WriteRegister(Register.I2C_CONFIG, 0x08);
+        }
+
+        #region Device Registers
+
+        /// <summary>
+        /// Get the distance measurement in cm.
+        /// </summary>
+        public int Distance {
```
krwq: 1. Braces on new line (here and everywhere)
2. Since this is not doing the measurement and just reads back the last measurement (if I understand correctly) this should be called LastDistance to make it clear.

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,319 @@
...
+        public int Velocity {
+            get {
+                Span<byte> rawData = stackalloc byte[1] { 0 };
+                ReadBytes(Register.VELOCITY, rawData);
+                return (int)(sbyte)rawData[0];
+            }
+        }
+
+        /// <summary>
+        /// Get or set the acquistion mode control
```
krwq: Can you describe more what this means?

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,319 @@
...
+                ReadBytes(Register.STATUS, rawData);
+                return (SystemStatusFlag)rawData[0];
+            }
+        }
+
+        #endregion
+
+        #region I2C
+
+        internal void WriteRegister(Register reg, byte data)
```
krwq: if possible make those private

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,319 @@
...
+        internal void ReadBytes(Register reg, Span<byte> readBytes)
+        {
+            _i2cDevice.WriteByte((byte)reg);
+            _i2cDevice.Read(readBytes);
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            if (_gpioController != null && _powerEnablePin.HasValue)
```
krwq: gpioController will dispose pins in its Dispose method

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,319 @@
...
+            {
+                _gpioController.ClosePin(_powerEnablePin.Value);
+            }
+
+            _gpioController?.Dispose();
+            _i2cDevice?.Dispose();
```
krwq: also `_gpioController = null;`

src/devices/LidarLiteV3/MeasurementRepetitionMode.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. 
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.TimeOfFlight.Models.LidarLiteV3
+{
+    /// <summary>
+    /// Measurement repetition modes
+    /// </summary>
+    public enum MeasurementRepetitionMode
```
krwq: I'd remove the `Mode` prefix since it doesn't make name any clearer/more specific

src/devices/LidarLiteV3/MeasurementRepetitionMode.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// </summary>
+        Off,
+
+        /// <summary>
+        /// Measurements are done repetitively per command defined in OUTER_LOOP_COUNT.
+        /// </summary>
+        Repeat,
+
+        /// <summary>
+        /// Measurements are done infinitely.
```
krwq: Can you explain also on example or when you should use it? Is this equivalent to: Start measuring and finish measuring when on next read?

src/devices/LidarLiteV3/PowerOptionFlag.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+
+namespace Iot.Device.TimeOfFlight.Models.LidarLiteV3
+{
+    /// <summary>
+    /// Power option flags
+    /// </summary>
+    [Flags]
+    public enum PowerOptionFlag
```
krwq: I think most of the devices call this PowerMode or similar (at least remove `Flag` suffix if you prefer to keep this name)

src/devices/LidarLiteV3/SystemStatusFlag.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    /// <summary>
+    /// System status flags
+    /// </summary>
+    [Flags]
+    public enum SystemStatusFlag
+    {
+        /// <summary>
+        /// Process Error Flag
+        /// 0 - No error detected
```
krwq: if possible remove 0s and 1s from the summary

src/devices/LidarLiteV3/samples/LidarLiteV3.Sample.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Iot.Device.TimeOfFlight.LidarLiteV3Samples
+{
+    internal class Program
+    {
+        public static void Main(string[] args)
+        {
+            using (var llv3 = new LidarLiteV3(CreateI2cDevice()))
+            {
+                // Take 10 measurements, each one second apart.
+                for (int i = 0; i < 10; i++) {
```
krwq: braces on new line

src/devices/LidarLiteV3/README.md
```diff
@@ -0,0 +1,12 @@
+﻿# LidarLiteV3  LIDAR ToF Sensor by Garmin
```
krwq:
```suggestion
# LidarLiteV3 - LIDAR Time of Flight Sensor
```

src/devices/LidarLiteV3/category.txt
```diff
@@ -0,0 +1,2 @@
+lidar
```
krwq: note: we'll ignore `lidar` category anyway for now since there is just 1 device there

src/devices/LidarLiteV3/samples/README.md
```diff
@@ -0,0 +1,195 @@
...
+advantage is longer distances (up to 40m) but more prone to errors when the subject is reflective. 
+
+# Raspberry PI Wiring
+
+The device communicates over I2C and while there's a PWN mode supported by the device, it is not 
+implemented.
+
+I2C Wiring on the Raspberry PI:
+
+| LidarLiteV3 Wire      | Raspberry PI GPIO Pin                   |
```
krwq: add a note that this is physical numbering (most of the people use logical, logical is whatever BCM chip spec calls it)

src/devices/LidarLiteV3/samples/README.md
```diff
@@ -0,0 +1,195 @@
...
+|-----------------------|-----------------------------------------|
+| 5V (red)              | 5V (pin 2)                              |
+| Ground (black)        | Ground (pin 6)                          |
+| Power enable (orange) | Optional, an available IO pin (pin 33)  |
+| Mode control (yellow) | Not used                                |
+| I2C SCL (green)       | SCL (pin 5)                             |
+| I2C SDA (blue)        | SDA (pin 3)                             |
+
+> Important! Don't forget to enable I2C via `raspi-config`.
```
krwq: :) surprisingly many people forget about it - I hope to eventually add some generated part of the doc which will add some more comprehensive note on at least every I2C/SPI device

src/devices/LidarLiteV3/samples/README.md
```diff
@@ -0,0 +1,195 @@
...
+    while(true)
+    {
+        Thread.Sleep(5);
+        ushort currentDistanceInCm = llv3.Distance;
+        int currentVelocityInCm = llv3.Velocity;
+    }
+}
+```
+
+# Change the IC2 Address
```
krwq:
```suggestion
# Change the I2C Address
```

src/devices/LidarLiteV3/samples/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+By default, the device has an address of `0x62`.  It's possible to change this address to 
+resolve a conflict with another device or to run multiple devices.
+
+Available addresses are 7-bit values with a 0 in the lsb.
```
krwq:
```suggestion
Available addresses are 7-bit values with a 0 in the LSB order.
```

src/devices/LidarLiteV3/samples/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+```csharp
+using (var llv3 = new LidarLiteV3(CreateI2cDevice()))
+{
+    // Set device from default `0x62` to `0x68`
+   llv3.SetI2CAddress(0x68);
```
krwq: just curious: will the device still respond on the default address after this?

src/devices/LidarLiteV3/README.md
krwq: similarly on the sample/README.md

src/devices/LidarLiteV3/samples/README.md
vornet: That function does explicitly disable the default address (see https://github.com/dotnet/iot/pull/839/files#diff-f8efedcd82e70ed6dbf0925a2a397575R176).  Could make it configurable, but can't really think of a use case where you want to change it but also keep the old address.

If you have multiple devices, I think you use the enable pin to enable only one of the devices, set the address, enable one more, set the next address, and so on.  I haven't proven this approach works yet since I only have one of the devices.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: No strong opinion either.  That works, will change.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Aren't you suppose to have a single shared instance of `GPIOController` for your application, so it can manage all pin access?

If that's the case, wouldn't you want each disposing devices to release pins it no longer uses, so that other devices in the same application can reuse those pins?  Maybe this is an unrealistic scenario since only one device is physically connected to that pin.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Will fix.

src/devices/LidarLiteV3/AcquistionModeFlag.cs
vornet: will clean this up

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: will fix -- also will try to address anything already mentioned in the guideline

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: will fix

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: will fix

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: will fix

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: updating with a better description.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: made them private

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: ended up removing this, make sense to have controller dipose them.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: fixed

src/devices/LidarLiteV3/MeasurementRepetitionMode.cs
vornet: fixed

src/devices/LidarLiteV3/MeasurementRepetitionMode.cs
vornet: Actually briefly convered it here, https://github.com/vornet/iot/tree/LidarLiteV3/src/devices/LidarLiteV3/samples#repetition-mode.  It can internally trigger measurements, either in a finite loop or infinitely.  So it's the equivalent of doing the measurement in a c# loop, but on the device itself with more predicable timing since there's less overhead.

src/devices/LidarLiteV3/PowerOptionFlag.cs
vornet: will fix

src/devices/LidarLiteV3/SystemStatusFlag.cs
vornet: fixed

src/devices/LidarLiteV3/samples/LidarLiteV3.Sample.cs
vornet: fixed

src/devices/LidarLiteV3/samples/README.md
vornet: makes sense

src/devices/LidarLiteV3/LidarLiteV3.cs
krwq: You should create one GpioController per device although I think it doesn't matter that much if you have one or multi because of the static nature of the physical setup

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,337 @@
...
+
+        /// <summary>
+        /// Power off the device if GPIO controller and power enable pin is provided.
+        /// </summary>
+        public void PowerOff()
+        {
+            if (_powerEnablePin != -1)
+            {
+                _gpioController.Write(_powerEnablePin, PinValue.Low);
+            } else
```
krwq: `else` on new line

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,337 @@
...
+            return MeasureDistanceAsync(withReceiverBiasCorrection).Result;
+        }
+
+        /// <summary>
+        /// Set the repetition mode to enable automatic measurement.
+        /// </summary>
+        /// <param name="measurementRepetition">Repetition mode, either Off, Repeat, or RepeatInfinitely.</param>
+        /// <param name="count">If Repeat, the number of times to repeat the measurement.</param>
+        /// <param name="delay">The delay between each measurements. Note the unit does not directly to hz, a value of 20 maps to about 100 hz.</param>
+        public void SetMeasurementRepetitionMode(MeasurementRepetition measurementRepetition, int? count = null, int? delay = null)
```
krwq: our guidelines recommend to use -1 instead of null whenever it can be ambiguously used

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,337 @@
...
+        {
+            if (count.HasValue && (count < 2 || count > 254))
+            {
+                throw new ArgumentOutOfRangeException("Count must be between 2 and 254");
+            };
```
krwq:
```suggestion
            }
```

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,337 @@
...
+
+            if (delay != null)
+            {
+                WriteRegister(Register.MEASURE_DELAY, (byte)delay.Value);
+
+                // Set mode to use custom delay.
+                var currentAcqSettings = AcquistionSettings;
+                currentAcqSettings |= AcquistionSettings.UseDefaultDelay;
+                AcquistionSettings = currentAcqSettings;
+            } else
```
krwq:
```suggestion
            }
            else
```

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,337 @@
...
+
+        /// <summary>
+        /// Set a new I2C address.
+        /// </summary>
+        /// <param name="address">new address, valid values are 7-bit values with 0 in the LSB.</param>
+        public void SetI2CAddress(byte address)
+        {
+            if ((address & 1) == 1)
+            {
+                throw new ArgumentOutOfRangeException("Address must have 0-bit in the LSB.");
```
krwq: `must be and odd number`?

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,337 @@
...
+        }
+
+        #region Device Registers
+
+        /// <summary>
+        /// Get the last distance measurement in cm.
+        /// </summary>
+        public int LastDistance 
+        {
+            get {
```
krwq:
```suggestion
            get
            {
```

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,337 @@
...
+                Span<byte> rawData = stackalloc byte[2] { 0, 0 };
+                ReadBytes(Register.FULL_DELAY, rawData);
+                return BinaryPrimitives.ReadUInt16BigEndian(rawData);
+            }
+        }
+
+        /// <summary>
+        /// Get the difference between the current and last measurement resulting
```
krwq: There is something off about this description. Velocity should be in time domain so:
- should time between measurements be accounted for here?
- should the unit be cm/s cm/h?

If this is just pure difference then it should not be called Velocity...

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,337 @@
...
+        /// in a signed (2's complement) 8-bit number in cm.
+        /// Positive is away from the device.
+        /// </summary>
+        public int Velocity 
+        {
+            get {
```
krwq:
```suggestion
            get
            {
```

please fix everywhere

src/devices/LidarLiteV3/samples/README.md
krwq: Will the address be preserved after reboot? Can you please update SetI2CAddress xml doc to answer these questions?

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: According to the doc, "The I2C Address can be changed from its default value. Available addresses are 7-bit values with a '0' in the LSB." This validation check makes sure the LSB is 0.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: This one is tricky, it's the difference calculated between the last two distance reading, which you're right, isn't velocity.  

However, when used in conjection with MeasurementRepeititon.RepeatIndefinitely and a delay, it can be used to reliably calculate a velocity from the device.  The default delay equals velocity measurements in 0.1 m/s.  Since the delay is configurable, the unit isn't defined.  I believe this is the intented use for this register (which is why it's called velocity).

While you can do this calculation outside of the device in C#, I persume it would be more precise to do it on the device itself.

I'm going to rename the property to `DifferenceBetweenLastTwoDistances`.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Will fix.  I'm using VS Code which doesn't seem to auto-format these for me. I can't seem get the project to compile in VS 2019, it only works if I use `dotnet build .` in the project directory.

src/devices/LidarLiteV3/samples/README.md
vornet: No, a reset or power off will clear all registers including custom I2C addressses.  Will update doc!

vornet:
@krwq i think I addressed any remaining comments.  Let me know if I missed anything.  Thanks! Will also do another round of testing with the device on the pi tonight or tomorrow.

krwq:
@vornet thanks, please also address the style cop issues (see PR failures) - to see those locally you might need to rebase. We have recently enabled it that's why you probably haven't seen any errors before the last iteration

vornet:
@krwq looks good now!

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+        {
+            return MeasureDistanceAsync(withReceiverBiasCorrection).Result;
+        }
+
+        /// <summary>
+        /// Set the repetition mode to enable automatic measurement.
+        /// </summary>
+        /// <param name="measurementRepetition">Repetition mode, either Off, Repeat, or RepeatInfinitely.</param>
+        /// <param name="count">If Repeat, the number of times to repeat the measurement.</param>
+        /// <param name="delay">The delay between each measurements. Note the unit does not directly to hz, a value of 20 maps to about 100 hz.</param>
```
krwq: can you clarify the description here? What's the unit of this? Hz is 1/s which is not a time unit.

If the unit is not clear please add a conversion routine so that you can use frequency or time as an input.

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+                return BinaryPrimitives.ReadUInt16BigEndian(rawData);
+            }
+        }
+
+        /// <summary>
+        /// Get the difference between the current and last measurement resulting
+        /// in a signed (2's complement) 8-bit number in cm.
+        /// Positive is away from the device.
+        /// </summary>
+        public int DifferenceBetweenLastTwoDistances
```
krwq: `DistanceDelta`?

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+            }
+            set
+            {
+                WriteRegister(Register.ACQ_CONFIG_REG, (byte)value);
+            }
+        }
+
+        /// <summary>
+        /// Get or set the maximum aquisition count limits the number of times
+        /// the device will integrate acquistions to find a correlation
```
krwq: If there are easier words to explain this please do.. This should answer the question: "when and why would user would want to use this". If this is some harder concept please at least add some section explaining that in the sample

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+                return rawData[0];
+            }
+            set
+            {
+                WriteRegister(Register.SIG_COUNT_VAL, (byte)value);
+            }
+        }
+
+        /// <summary>
+        /// Get or set the threshold of peak value that bypasses the internal algorithm.
```
krwq: what internal algorithm will be bypassed? what will exactly happen when it's higher?

src/devices/LidarLiteV3/samples/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+# Optimization
+
+The default settings should work well, but several tweaks can be made to adjust the device.
+See the manual for more details.
+
+## Change Acquistion Count
+
+To isolate the signal from the noise, the device performs a series of acquisitions and sums up the 
+result until a peak is found.  The number of acquistions can be configured via
```
krwq: - remove double spaces
- when you say "peak" do you mean the highest/lowest value or something else? I.e. if I had a distance sensor in front of my car and something was found there for a millisecond I'd prefer "low" to be on the safe side while in any non-sensitive context I'd prefer "high"

src/devices/LidarLiteV3/AcquistionSettings.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System;
+
+namespace Iot.Device.DistanceSensor.Models.LidarLiteV3
+{
+    /// <summary>
+    /// Various acquisition behavior settings
+    /// </summary>
+    [Flags]
+    public enum AcquistionSettings
```
danmosemsft: typo-- Acquisition

src/devices/LidarLiteV3/AcquistionSettings.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// Enable measurement quick termination
+        /// </summary>
+        EnableQuickTermination = 0x08,
+
+        /// <summary>
+        /// Enable default reference acquisition count (5) instead of reference acquisition
+        /// count set in REF_COUNT_VAL (0x12).
+        /// </summary>
+        UseDefaultReferenceAcq = 0x04
```
danmosemsft: If "Acq" is not the standard accepted term, then .NET guidelines would prefer it spelled out.

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+            }
+
+            if (delay != -1)
+            {
+                WriteRegister(Register.MEASURE_DELAY, (byte)delay);
+
+                // Set mode to use custom delay.
+                var currentAcqSettings = AcquistionSettings;
+                currentAcqSettings |= AcquistionSettings.UseDefaultDelay;
+                AcquistionSettings = currentAcqSettings;
```
danmosemsft: "Acquistion" --> "Acquisition" throughout

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+            else
+            {
+                throw new InvalidOperationException("Cannot power off without providing GPIO controller and power enable pin.");
+            }
+        }
+
+        /// <summary>
+        /// Power on the device if GPIO controller and power enable pin is provided.
+        /// </summary>
+        public void PowerOn()
```
danmosemsft: Seems shorter, more readable and more idiomatic to write
```c#
        public void PowerOn()
        {
            if (_powerEnablePin == -1)
            {
                throw new InvalidOperationException("Cannot power off without providing GPIO controller and power enable pin.");
            }

            _gpioController.Write(_powerEnablePin, PinValue.High);
        }
```
And similar elsewhere

vornet:
@krwq @danmosemsft Will be a while before i get back to these PRs comments.  My boat (it uses this sensor) is in a 1000 pieces right now.  Need to put it together first to test some of the changes.

krwq:
hey @vornet, how is your boat? Do you think you will have time to address the feedback?

vornet:
@krwq whoops, completely forgot about this.  I had to put the project on hold for a bit, but looking to jump back in soon.  Will address the feedback this weekend.

src/devices/LidarLiteV3/AcquistionSettings.cs
krwq: @vornet, I think this got missed

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+        private I2cDevice _i2cDevice;
+        private int _powerEnablePin;
+
+        /// <summary>
+        /// Initialize the LidarLiteV3
+        /// </summary>
+        /// <param name="i2cDevice">I2C device</param>
+        /// <param name="gpioController">GPIO controller</param>
+        /// <param name="powerEnablePin">The pin number used to control power to the device</param>
+        public LidarLiteV3(I2cDevice i2cDevice, GpioController gpioController = null, int powerEnablePin = -1)
```
krwq, line 35: there should also be `shouldDispose` parameter (note that if gpioController is null it should always be overriden to "true", see i.e. https://github.com/dotnet/iot/blob/master/src/devices/README.md#gpiocontroller)

src/devices/LidarLiteV3/LidarLiteV3.cs
krwq: this seems it got missed

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+        {
+            _gpioController = gpioController ?? new GpioController();
+            _i2cDevice = i2cDevice;
+            _powerEnablePin = powerEnablePin;
+
+            if (_powerEnablePin != -1)
+            {
+                _gpioController.OpenPin(_powerEnablePin, PinMode.Output);
```
krwq, line 43: consider adding parameter `ensurePoweredOn` (optional)

krwq: (if you are caching state then you'd need to read it so you can skip this comment if you don't want to do that)

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+        /// <summary>
+        /// Reset FPGA, all registers return to default values
+        /// </summary>
+        public void Reset()
+        {
+            WriteRegister(Register.ACQ_COMMAND, 0x00);
+        }
+
+        /// <summary>
+        /// Measure distance in cm
```
krwq, line 89: this method should return Distance and comment should be unitless, you should add expected range in the summary

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Note: Do not call if running while in repetition mode.  It will block until
```
krwq, line 92: consider detecting "forever" state and throwing an exception

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+        /// repetition finishes (forever if infinite).
+        /// </remarks>
+        /// <param name="withReceiverBiasCorrection">Faster without bias correction, but more prone to errors if condition changes.</param>
+        /// <returns>Distance in cm</returns>
+        public int MeasureDistance(bool withReceiverBiasCorrection = true)
```
krwq, line 118: same comment for `Distance`

src/devices/LidarLiteV3/LidarLiteV3.cs
krwq: this seems it got missed

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        /// <summary>
+        /// Set a new I2C address.
+        /// </summary>
+        /// <remarks>
+        /// Note, if the device is powered off or reset, the IC2 address will reset to the default address.
+        /// </remarks>
+        /// <param name="address">new address, valid values are 7-bit values with 0 in the LSB.</param>
+        public void SetI2CAddress(byte address)
```
krwq, line 177: If this device is unusable after changing the address (because communication will be disrupted) then it should also Dispose (SetI2cAddressAndDispose)

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+            // Disable the default address
+            WriteRegister(Register.I2C_CONFIG, 0x08);
+        }
+
+        #region Device Registers
+
+        /// <summary>
+        /// Get the last distance measurement in cm.
+        /// </summary>
+        public int LastDistance
```
krwq, line 205: same comment about `Distance`

src/devices/LidarLiteV3/LidarLiteV3.cs
krwq, line 220: Unit should be `Distance` or `Velocity`

src/devices/LidarLiteV3/SystemStatus.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        /// <summary>
+        /// System error detected during measurement
+        /// </summary>
+        ProcessError = 0x40,
+
+        /// <summary>
+        /// Health status, indicating reference and receiver bias are operational
+        /// </summary>
+        Health = 0x20,
```
krwq, line 23: Should this be "Healthy"?

src/devices/LidarLiteV3/samples/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+# Change the I2C Address
+
+By default, the device has an address of `0x62`.  It's possible to change this address to 
+resolve a conflict with another device or to run multiple devices.
+
+Available addresses are 7-bit values with a 0 in the LSB order.
+
+```csharp
+using (var llv3 = new LidarLiteV3(CreateI2cDevice()))
```
krwq, line 129: So as mentioned before this should be something like:
`new LidarLiteV3(CreateI2cDevice()).SetI2cAddressAndDispose()`

src/devices/LidarLiteV3/samples/README.md
```diff
@@ -0,0 +1,195 @@
...
+}
+
+```
+
+# Optimization
+
+The default settings should work well, but several tweaks can be made to adjust the device.
+See the manual for more details.
+
+## Change Acquistion Count
```
krwq, line 151: Acquisition

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Makes sense.  Will add that.

krwq:
Hey @vornet, there are still some comment which haven't been fixed and I've also added new series of comments (some related to the fact we now use UnitsNet or have some more clarified guidelines)

src/devices/LidarLiteV3/SystemStatus.cs
vornet: These are the direct name of the flags from the device documentation. Not saying we shouldn't change it, but one of the design idea I've been thinking about is to expose a higher-level API (different class, possibly different namespace) for this device, and while also maintaining a bare metal low level version that maps one-to-one to the device.  These flags would remain the same and kept in the low level API.

So I see two classes of users:

Typical User --> High-Level API --> Low-Level API --> Device
Advanced User --> Low-Level API --> Device

Thoughts?

vornet (commented):
I still have to put the device together again to test my changes, hoping to get an update to the PR this week.

krwq:
@vornet, I think there are things user (regardless of knowledge) might want to achieve with the device. I think we should expose the missing functionalities rather than exposing the registers directly. If advanced user needs to achieve something out of ordinary I suggest to fork and edit the code (and ideally contribute back) rather than making it more complicated for everyone else so my vote is to keep registers internal. The device is already bare metal so advanced user can use I2C directly to speak to it...

joperezr:
[Triage] @vornet still interested in working on this? Do you mind answering to @krwq concern above? Let us know if you are not interested in doing this any more so that we can close the PR in our next triage meeting.

vornet:
@joperezr @krwq yes, just running against some work-work deadlines.  I'll try to wrap this up this early next week.

joperezr:
Oh that's totally ok, we just wanted to make sure the PR wasn't abandoned 😄 Take your time and thanks again for contributing!

joperezr:
[Triage] Make sure that you rebase on latest master since we made some changes with nullability that very likely broke part of your code, so it will need attention.

joperezr:
[Triage] Friendly ping 😄

src/devices/LidarLiteV3/AcquistionSettings.cs
joperezr, line 13:
```suggestion
    public enum AcquisitionSettings
```

src/devices/LidarLiteV3/AcquistionSettings.cs
joperezr, line 39:
```suggestion
        UseDefaultReferenceAcquisition = 0x04
```

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+        private I2cDevice _i2cDevice;
+        private int _powerEnablePin;
+
+        /// <summary>
+        /// Initialize the LidarLiteV3
+        /// </summary>
+        /// <param name="i2cDevice">I2C device</param>
+        /// <param name="gpioController">GPIO controller</param>
+        /// <param name="powerEnablePin">The pin number used to control power to the device</param>
+        public LidarLiteV3(I2cDevice i2cDevice, GpioController? gpioController = null, int powerEnablePin = -1)
```
joperezr, line 35: Seems like you missed a comment that @krwq made here around also adding a shouldDispose which is our convention when we take a gpiocontroller. Here is his original comment:

> there should also be shouldDispose parameter (note that if gpioController is null it should always be overriden to "true", see i.e. https://github.com/dotnet/iot/blob/master/src/devices/README.md#gpiocontroller)

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+            else
+            {
+                throw new InvalidOperationException("Cannot power off without providing GPIO controller and power enable pin.");
+            }
+        }
+
+        /// <summary>
+        /// Power on the device if GPIO controller and power enable pin is provided.
+        /// </summary>
+        public void PowerOn()
```
joperezr, line 68: Seems like you are missing to address one comment from @danmosemsft here:
> Seems shorter, more readable and more idiomatic to write
public void PowerOn()
{
    if (_powerEnablePin == -1)
    {
        throw new InvalidOperationException("Cannot power off without providing GPIO controller and power enable pin.");
    }
    _gpioController.Write(_powerEnablePin, PinValue.High);
}
And similar elsewhere

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+        /// <summary>
+        /// Reset FPGA, all registers return to default values
+        /// </summary>
+        public void Reset()
+        {
+            WriteRegister(Register.ACQ_COMMAND, 0x00);
+        }
+
+        /// <summary>
+        /// Measure distance in cm
```
joperezr, line 89: Won't comment on all of these, but all of the methods you have that are returning an int in centimeters, should instead be returning `UnitsNet.Length` types instead so that consumer can decide if they want it on cm or some other unit instead.

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Note: Do not call if running while in repetition mode.  It will block until
+        /// repetition finishes (forever if infinite).
+        /// </remarks>
+        /// <param name="withReceiverBiasCorrection">Faster without bias correction, but more prone to errors if condition changes.</param>
+        /// <returns>Distance in cm</returns>
+        public int MeasureDistance(bool withReceiverBiasCorrection = true)
+        {
+            return MeasureDistanceAsync(withReceiverBiasCorrection).Result;
```
joperezr, line 120: Instead of calling .Result, you should be calling `.GetAwaiter().GetResult()` or else you could run into deadlocks when the task is not finished.

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,358 @@
...
+                Span<byte> rawData = stackalloc byte[1];
+                ReadBytes(Register.VELOCITY, rawData);
+                return (int)(sbyte)rawData[0];
+            }
+        }
+
+        /// <summary>
+        /// Get or set the various settings to control the acquistion behavior.
+        /// </summary>
+        public AcquistionSettings AcquistionSettings
```
joperezr, line 233: Looks like you also have a typo on the word Acquisition here, it would be probably good to just do a find all instances of where you have it misspelled and fix them.

src/devices/LidarLiteV3/samples/LidarLiteV3.Sample.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+using System.Threading;
+using System.Device.I2c;
+using Iot.Device.DistanceSensor;
+
+namespace Iot.Device.DistanceSensor.LidarLiteV3Samples
+{
+    internal class Program
+    {
+        public static void Main(string[] args)
```
joperezr, line 14: We now have most of our samples being top-level programs, which really requires minimum number of changes here in order to have it like that. You don't have to do this but it would be good to have consistency across the repo. You can check any other sample in the repo for examples.

joperezr:
I pushed a change to fix the build that was having nullability and target framework issues, and left a few small set of comments but this looks like it is almost ready to go 😄 Thanks for keep working on this @vornet and now we are finally on the last stretch.

vornet:
@joperezr Amazing, thank you!  I've been struggling to find time to work on this even though there's so little left to do, feel free to wrap it up and get it merged.  Otherwise, I'll try to take a look at it this weekend.

src/devices/LidarLiteV3/samples/LidarLiteV3.Sample.cs
vornet: Made the sample top-level, much more concise.

krwq:
@vornet, I'm still seeing lots of comments here not addressed, i.e.:
- typos
- 3 line copyright header should be 2-line

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Updated to use DelayHelpers.

src/devices/LidarLiteV3/AcquistionSettings.cs
vornet: Fixed.

src/devices/LidarLiteV3/AcquistionSettings.cs
vornet: Fixed

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Added shouldDispose support.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Addressed.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Updated to use UnitsNet.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Ended up getting rid of the async method because DelayHelper wasn't async and that replaced the need for async.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Fixed all instances of this.

vornet:
@krwq sorry didn't realized I missed so much.  I'm waiting on the CI build to finish to see if it's broken, but everything should be addressed at this point.

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,363 @@
...
+            return LastDistance;
+        }
+
+        /// <summary>
+        /// Set the repetition mode to enable automatic measurement.
+        /// </summary>
+        /// <param name="measurementRepetition">Repetition mode, either Off, Repeat, or RepeatInfinitely.</param>
+        /// <param name="count">If Repeat, the number of times to repeat the measurement.</param>
+        /// <param name="delay">The delay between each measurements. Note the unit does not directly to hz, a value of 20 maps to about 100 hz.</param>
+        public void SetMeasurementRepetitionMode(MeasurementRepetition measurementRepetition, int count = -1, int delay = -1)
```
krwq, line 130: ideally `delay` should be of type `TimeSpan?`

src/devices/LidarLiteV3/LidarLiteV3.cs
```diff
@@ -0,0 +1,363 @@
...
+                    break;
+            }
+
+            if (delay != -1)
+            {
+                WriteRegister(Register.MEASURE_DELAY, (byte)delay);
+
+                // Set mode to use custom delay.
+                var currentAcqSettings = AcquisitionSettings;
+                currentAcqSettings |= AcquisitionSettings.UseDefaultDelay;
```
krwq, line 156: this seems quite reverse to what you want: isn't UseDefaultDelay ignore `MEASURE_DELAY` setting?

src/devices/LidarLiteV3/samples/README.md
```diff
@@ -0,0 +1,195 @@
...
+        using (var llv3 = new LidarLiteV3(CreateI2cDevice()))
+        {
+            // Take 10 measurements, each one second apart.
+            for (int i = 0; i < 10; i++) {
+                ushort currentDistance = llv3.MeasureDistance();
+                Console.WriteLine($"Current Distance: {currentDistance} cm");
+                Thread.Sleep(1000);
+            }
+        }
+    }
```
krwq, line 44: samples need to be updated (you return Length now)

krwq (approved):
LGTM after fixing 3 things:
- SetI2cAddress => SetI2cAddressAndDispose comment got missed
- samples in the README.md need to be updated to what you have in the samples folder
- potential problem with setting delay

krwq:
It's fine @vornet 😄 Thanks for updating!

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: There isn't a direct correlation between delay and real-world time.  Documentation only states that a value of 20 is about 100 hz, but 40 is not necessarily 200 hz.  Possible future enhancement is to build a mapping table, but I don't have equipment to accurately measure how often it does the measurement.

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Good point!  Logic as written is correct, the flag name is mislabeled.  Documentaiton states that:

```
0: Use default delay for burst and free running mode.
1: Use delay from MEASURE_DELAY (0x45) for burst and free running mode
```
So `UseDefaultDelay` should be relabeled to `UseCustomDelay`.  I'm going to do that.

src/devices/LidarLiteV3/samples/README.md
vornet: Updated readme to reflect the latest state of the code (including Length).

src/devices/LidarLiteV3/AcquisitionSettings.cs
```diff
@@ -0,0 +1,40 @@
...
+    [Flags]
+    public enum AcquisitionSettings
+    {
+        /// <summary>
+        /// Enables reference process during measurement
+        /// </summary>
+        EnableReferenceProcess = 0x40,
+
+        /// <summary>
+        /// Use default delay (10 hz) instead of delay configured in MEASURE_DELAY (0x45) for repetition mode
```
krwq, line 20:
```suggestion
        /// Use custom delay instead of default delay
```

I'm not sure what 10Hz delay means? 0.1s?

vornet: Not certain, but as I understand it, it's how often the sensor fires in a second.  The default delays slows it down enough to have the sensor fire 10 times a second.  A custom delay of 20 would result in 100 sensor readings a second.  I'm fine with the simplified description.

krwq:
`SetI2cAddress => SetI2cAddressAndDispose` comment got missed

krwq:
@vornet do not worry about the CI, the failures are unrelated

src/devices/LidarLiteV3/LidarLiteV3.cs
vornet: Just pushed up a fix for this, let me know if it makes sense.

krwq (approved):

krwq:
Thanks @vornet!

Issue #813, opened 2019-10-20
Hardware PWM without setup/sudo


pgrawehr:
I cannot find any documentation on how to use hardware PWM on the Raspberry Pi, and I cannot get it to run. 

Enabling PWM requires setting the Pin Mode for (BCM) Pins 12 or 13 to ALT0 mode, for which the library does not offer a possibility, it seems. (Note: Unlike most other alternate functions, PWM mode does not require a special driver and therefore also not a reboot to activate). 

I'm also not sure whether the currently implemented method of using /sys/class/pwm is the right one to access the PWM device. This directory is always empty, even if I enable PWM using `gpio -g mode 13 pwm`. 

Note: The following works to test the PWM Channel (attach an LED to BCM Pin 13):
```
gpio -g mode 13 pwm
gpio -g pwm 13 <value between 0 and 1024>
```

krwq:
I'm closing this as a dup of https://github.com/dotnet/iot/issues/802

Since the other issue doesn't have the answer.

To enable PWM use:



I recommend to read /boot/overlays/README since it describes pretty much how to enable anything on the PI.

pgrawehr:
@krwq: Ok, I'll try this approach, but I'm not happy with it. Some investigation showed that it's possible to drive the PWM mode from memory mapped registers. This, as said, has the advantage that it does not require an overlay module to be loaded and pin assignments can change without reboot. So we should have a driver for that. 
In either case, we need a way of setting the GPIO pins to alternate functions (and back). Even if the overlay mode was properly loaded, some (other) application might have set pin modes in such a way that the requested function (i.e. I2C) is not currently available.

joperezr:
@pgrawehr We could do that but the problem is that this would only work for one board. You are correct in the fact that using registers and setting alternate functions on pins you are able to control PWM as well, but we thought that if we could come up with a non-board-specific implementation of PWM would be ideal. Is there a specific reason other than requiring another overlay module loaded for which you prefer the register way?

krwq:
@joperezr we could potentially expose it directly from RaspberryPiDriver (i.e. CreatePwmChannel)

joperezr:
If we find a good use-case for it, for which the generic driver is not sufficient then yes this is one option. Before adding extra code we would have to maintain and wouldn't be portable for the rest of the boards, I would like to make sure that there is a good reason why we it.

pgrawehr:
[...]

PR #854, opened 2019-11-16
Hardware PWM Fixes / Improvements

pgrawehr:
- This fixes a crash when changing the frequency while the duty-cycle is high (check this command-line sequence):
```
oot@raspberrypi:/sys/class/pwm/pwmchip0/pwm0# echo 1 > enable
root@raspberrypi:/sys/class/pwm/pwmchip0/pwm0# echo 25000 > period
root@raspberrypi:/sys/class/pwm/pwmchip0/pwm0# echo 25000 > duty_cycle # the LED goes on now
root@raspberrypi:/sys/class/pwm/pwmchip0/pwm0# echo 10000 > period # fails, because duty_cycle > period
bash: echo: write error: Invalid argument
root@raspberrypi:/sys/class/pwm/pwmchip0/pwm0# echo 0 > duty_cycle # the LED goes off
root@raspberrypi:/sys/class/pwm/pwmchip0/pwm0# echo 10000 > period # period can be set
root@raspberrypi:/sys/class/pwm/pwmchip0/pwm0# echo 5000 > duty_cycle # and enable LED again
root@raspberrypi:/sys/class/pwm/pwmchip0/pwm0# 
```
- Adds the udev rule to access /sys/class/pwm as non-root (copy the file 55-pwm.rules to /etc/udev/rules.d and reboot) See also #813 
- To make the above really work, a few retries are necessary while opening the hardware PWM device, to make sure the permissions are updated. See also #811

### DIFF
--- a/src/System.Device.Gpio/System/Device/Pwm/55-pwm.rules
+++ b/src/System.Device.Gpio/System/Device/Pwm/55-pwm.rules
@@ -1,0 +1,4 @@
+SUBSYSTEM=="pwm*", PROGRAM="/bin/sh -c '\
+        chown -R root:gpio /sys/class/pwm && chmod -R 770 /sys/class/pwm;\
+        chown -R root:gpio /sys/devices/platform/soc/*.pwm/pwm/pwmchip* && chmod -R 770 /sys/devices/platform/soc/*.pwm/pwm/pwmchip*\
+'"
--- a/src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
@@ -88,6 +88,7 @@ namespace System.Device.Pwm.Channels
 
         /// <summary>
         /// Sets the frequency for the channel.
+        /// This will quickly set the Duty Cycle to 0 before re-enabling it on the previous value.
         /// </summary>
         /// <param name="frequency">The frequency in hertz to set.</param>
         private void SetFrequency(int frequency)
@@ -96,12 +97,16 @@ namespace System.Device.Pwm.Channels
             {
                 throw new ArgumentOutOfRangeException(nameof(frequency), frequency, "Value must not be negative.");
             }
-
+            double oldDutyCycle = _dutyCycle;
+            // We have to set the duty cycle to 0 before changing the frequency, otherwise an InvalidArgumentException occurs when writing the period file and the 
+            // current duty cycle is longer than the period. We have to recompute the duty cycle (in nanoseconds) after this, anyway
+            SetDutyCycle(0);
             int periodInNanoseconds = GetPeriodInNanoseconds(frequency);
             _frequencyWriter.BaseStream.SetLength(0);
             _frequencyWriter.Write(periodInNanoseconds);
             _frequencyWriter.Flush();
             _frequency = frequency;
+            SetDutyCycle(oldDutyCycle);
         }
 
         /// <summary>
--- a/src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
@@ -12,6 +12,7 @@ namespace System.Device.Pwm.Channels
     /// </summary>
     internal class UnixPwmChannel : PwmChannel
     {
+        private readonly TimeSpan OpenTimeout = TimeSpan.FromSeconds(1);
         private readonly int _chip;
         private readonly int _channel;
         private int _frequency;
@@ -41,12 +42,27 @@ namespace System.Device.Pwm.Channels
             Validate();
             Open();
 
-            // avoid opening the file for operations changing relatively frequently
-            _dutyCycleWriter = new StreamWriter(new FileStream($"{_channelPath}/duty_cycle", FileMode.Open, FileAccess.ReadWrite));
-            _frequencyWriter = new StreamWriter(new FileStream($"{_channelPath}/period", FileMode.Open, FileAccess.ReadWrite));
+            var timeout = OpenTimeout;
+            // If we're running as non-root (and the udev rule needs to kick in here) we 
+            // might need a few retries until we can open the channel files. 
+            while (timeout > TimeSpan.Zero)
+            {
+                try
+                {
+                    // avoid opening the file for operations changing relatively frequently
+                    _dutyCycleWriter = new StreamWriter(new FileStream($"{_channelPath}/duty_cycle", FileMode.Open, FileAccess.ReadWrite));
+                    _frequencyWriter = new StreamWriter(new FileStream($"{_channelPath}/period", FileMode.Open, FileAccess.ReadWrite));
+                    break;
+                }
+                catch (Exception x) when (x is UnauthorizedAccessException || x is IOException)
+                {
+                    Thread.Sleep(1);
+                    timeout -= TimeSpan.FromMilliseconds(1);
+                }
+            }
 
             SetFrequency(frequency);
-            DutyCycle = dutyCycle;
+            SetDutyCycle(dutyCycle);
         }
 
         /// <inheritdoc/>


src/System.Device.Gpio/System/Device/Pwm/55-pwm.rules
```diff
@@ -0,0 +1,4 @@
+SUBSYSTEM=="pwm*", PROGRAM="/bin/sh -c '\
```
joperezr: I don't really like this part, I don't believe that we want our libraries copying files around that will change ownership of files. I much rather require root permissions in order to run our library (which is kind of already a requirement for gpio and other parts of the library anyways) than trying to bypass security that will affect the state of your PI/device aftewards.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -12,6 +12,7 @@ namespace System.Device.Pwm.Channels
     /// </summary>
     internal class UnixPwmChannel : PwmChannel
     {
+        private readonly TimeSpan OpenTimeout = TimeSpan.FromSeconds(1);
```
joperezr: We use cammel-case for variables like this, so it would be _openTimeout instead.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -41,12 +42,27 @@ internal class UnixPwmChannel : PwmChannel
             Validate();
             Open();
 
-            // avoid opening the file for operations changing relatively frequently
-            _dutyCycleWriter = new StreamWriter(new FileStream($"{_channelPath}/duty_cycle", FileMode.Open, FileAccess.ReadWrite));
-            _frequencyWriter = new StreamWriter(new FileStream($"{_channelPath}/period", FileMode.Open, FileAccess.ReadWrite));
+            var timeout = OpenTimeout;
+            // If we're running as non-root (and the udev rule needs to kick in here) we 
+            // might need a few retries until we can open the channel files. 
+            while (timeout > TimeSpan.Zero)
```
joperezr: Not really sure why you used a TimeSpan here, seems like what you are trying to do would be more appropiate to use with a `numberOfRetries` or something like that, where we would loop over a specified number of retries with a sleep in between each retry.

Is this only needed for non-root access? If so, I wouldn't add this code to optimize for that as I have already pointed out since we rather require root access for running, so I would just remove this whole logic.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -41,12 +42,27 @@ internal class UnixPwmChannel : PwmChannel
...
+                catch (Exception x) when (x is UnauthorizedAccessException || x is IOException)
+                {
+                    Thread.Sleep(1);
+                    timeout -= TimeSpan.FromMilliseconds(1);
+                }
+            }
 
             SetFrequency(frequency);
-            DutyCycle = dutyCycle;
+            SetDutyCycle(dutyCycle);
```
joperezr: These two calls are essentially the same thing, so I don't see much value but it does make sense to keep consistency and we are calling SetFrequency so I'm fine with that.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -96,12 +113,16 @@ private void SetFrequency(int frequency)
             {
                 throw new ArgumentOutOfRangeException(nameof(frequency), frequency, "Value must not be negative.");
             }
-
+            double oldDutyCycle = _dutyCycle;
+            // We have to set the duty cycle to 0 before changing the frequency, otherwise an InvalidArgumentException occurs when writing the period file and the 
+            // current duty cycle is longer than the period. We have to recompute the duty cycle (in nanoseconds) after this, anyway
+            SetDutyCycle(0);
```
joperezr: This is only the case when frequency will be less than dutycycle, so we don't want to set DutyCycle to 0 every time. What needs to be fixed here that we are not doing is to recalculate the DutyCycle after changing the frequency. This is a bug since today what we do is to set a DutyCycle which is a percentage of the frequency, but when we update the frequency we don't seem to update the DutyCycle so that is a bug.

src/System.Device.Gpio/System/Device/Pwm/55-pwm.rules
krwq: I think this should be described only in the README.md. I don't want to get into changing OS configuration path

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -96,12 +113,16 @@ private void SetFrequency(int frequency)
             {
                 throw new ArgumentOutOfRangeException(nameof(frequency), frequency, "Value must not be negative.");
             }
-
+            double oldDutyCycle = _dutyCycle;
```
krwq: This specific fix is already included in https://github.com/dotnet/iot/pull/758 and there are some edge cases around it which I already fixed.

There is couple of more issues with PWM in general which I'm still investigating (some of them make the driver not usable until RPi is rebooted).

src/System.Device.Gpio/System/Device/Pwm/55-pwm.rules
pgrawehr: I just provided the file, the code does nothing else with it. You have to copy it manually (usually to /etc/udev/rules.d). Should be described in either case, though. 
I think that's the last part that currently require root rights on the Pi. Gpio access no longer does.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
pgrawehr: Unfortunately, non-root access doesn't work without this (because it fails all the time). 
I think waiting for a certain time is better than a number of retries, since it takes a certain time for the rights to be set up. I don't have a big preference there, though.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
pgrawehr: That's why I changed it.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
pgrawehr: That part is fixed with my change as well (as I'm re-setting the existing percentage afterwards). Since the duty cycle is incorrect when changing the freqency in either case and I'm assuming that one wouldn't normally change the frequency all the time, I don't think we should complicate the logic much further and distinguish whether we increase or decrease the frequency or the duty cycle.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
pgrawehr: @krwq Oh... I didn't see that.
The only strange issue I had once was that the PWM channel would behave as if I had set the frequency to 1Hz. Couldn't reproduce this after a reboot. Maybe we need support for register based PWM access anyway, if this is so instable?

Any idea when that PR is ready for merge?

src/System.Device.Gpio/System/Device/Pwm/55-pwm.rules
joperezr: Depends on what part of Gpio and which driver you are using. I agree with @krwq that it is fine to have this on the README or a wiki, but not checked-in as an extra file to the code.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
joperezr: I think I didn't fully explained my point, what I meant is that you don't want to use TimeSpan as the condition in your while, instead you would probably want to do a retry number as there is no reason why to use TimeSpan. Like:

```c#
int numRetries = 5;
while (numRetries--)
{
  try
  {
    // do stuff.
  }
  catch (Exception e)
  {
    if (numRetries == 0)
    {
      throw new AggregateException("Exceeded the number of retries", e)
    }
    Thread.Sleep(someReasonableTimeLikeAMillisecondOrSo);
  }
}
```

Again, I don't think we want to optimize for non-root access, so if all of this code is just for enabling that, then I would just remove it.

src/System.Device.Gpio/System/Device/Pwm/55-pwm.rules
krwq: also keep in mind this particular file was copied from GPL v2 repo (at least per your comment in the issue) which means we cannot use it directly - although you can probably use something similar based on what I've pasted in the issue

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: @pgrawehr cannot merge until I upgrade the lab, currently we are waiting for PCBs to arrive.

I'll isolate that change from my PR so that we can merge this sooner but I'd prefer to do more testing first since I really want to figure out that issue which causes PWM to be unusable until reboot (currently I'm a bit busy with some non IoT related stuff)

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
pgrawehr: Both loops do about the same on the bottom line, but if you think your's is easier to read, I can change that. 
I do think that supporting non-root access is an important feature that will be requested by many users (in general of course, not only for this part).

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
pgrawehr: That shipment is seemingly taking ages, then... 
I have no hurry getting this to work, as I probably won't need it right now. I also just got a PCA9685 delivered, which is probably the better fit for running a number of (RGB) Leds. 
@krwq Question on a side issue: The library currently doesn't support setting the GPIO Pins to something else than IN or OUT. Namely, it is not possible to set any pin to their ALTn modes after they have been set to in or out. Are there already plans to implement this somehow? I think this is required as part of bringing hardware into the CI process, otherwise any wrong pin setup in a build would require a reboot afterwards.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
joperezr: I totally agree that non-root access would be awesome to have, but the current way we have for doing it is not good. Specially since we already said that we won't even copy the file over so that means that users would really still have to do some manual work in order to enable this. If that's the case, I much rather have a readme somewhere in our repo indicating which file you have to copy, and how to add the retries on your code in order to enable non-root access. If we were to have a great solution for this, then I would most definitely not be pushing back here.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
krwq: Yes, I agree that shipment is kinda slow (it's second version of the prototype though). @pgrawehr can you give a specific example of the scenario which requires reboot? (please file a separate issue so it doesn't get lost)

pgrawehr: Created #857 on this (sorry for the lengthy explanation)

pgrawehr:
I'll close this for the moment until #758 is merged. When we're lucky, the required udev rule _will_ even be added to the default installation (at least one of the responsible developers said he would pick this up).

Issue #816, opened 2019-10-21
Enable style checks during the build


krwq:
We should enable at least some style checks during the build.

This will avoid people to have to come back and fort with the PR iterations as they can be spotted by the build.

cc: @joperezr

pgrawehr:
@krwq: I would like to pick this up and enable it at least for core. (I'll see how far I get). I'll try to come up with a reasonable rule set which requires only minimal changes (and no api changes of course) for now. 

Some default rules we obviously break (such as SA1309 - Field names should not begin with underscores - where we go for SX1309 instead, which is just the opposite), but for some I'm not sure. 

- SA1311 Static readonly fields should begin with a capital letter: We now have a mixture of s_ prefix and normal CamelCasing as for properties. I'd personally prefer the later (so enabling the rule) but what would you suggest? - I think the s_ prefix should be for static fields (not readonly). 
- may come up with more

krwq:
@pgrawehr yes, you can go ahead with that. Please split rules into 3 parts (add a comment or something so we can pick them up whenever someone feels like doing the work):
- rules we can enable right now with minimal effort
- rules we eventually want
- rules we don't want

pgrawehr:
A solution with all the projects can be created automatically using:
```
dotnet new sln -n Allprojects
for /R %r in (*.csproj) do dotnet sln AllProjects.sln add %r
```
Some further modifications may be required to configure the correct build target afterwards.

pgrawehr:
With PRs #892, #894, #895, #897, #898, #900, #901, #902, #903 and #904, a basic set of style cop rules was defined and the files fixed (with some exceptions due to potentially breaking API changes). 

The rules enabled in https://github.com/dotnet/runtime/blob/master/eng/CodeAnalysis.ruleset are also enabled here (plus some more), but with a few exceptions (note that the ruleset there only mentions the rules explicitly disabled):
- SA1121 Use Built-In type aliases. [I would leave this disabled for this project, as sometimes it's more clear to write i.e. Int16 instead of short when interfacing with hardware]
- SA1129 Should not use value type default .ctors. [Same here, low-level structures usually are not declared with constructors, and whether I write `var a = new SomeStruct()` or `var a = default(SomeStruct)` (the later being allowed by the rule) doesn't really make a difference. 
- SA1217 Using statements should be ordered by alphabet. [No real preference on this one]
- SA1411 Attribute ctor must not use unnecessary parameters. [Could be done IMHO]
[...]

PR #859, opened 2019-11-21
API Suggestion: High Level Character LCD support

pgrawehr:
This is an API suggestion and a request for comments. It provides a high-level API for writing to character displays, the most important feature being the character translation required if non-english texts (or special characters) are to be printed. It's obvious that supporting all languages will not be possible, but many can be done. 
Suggestions for API improvements/extensions are welcome.


src/devices/CharacterLcd/samples/LcdConsole.Samples.cs
```diff
@@ -0,0 +1,142 @@
+﻿using System;
```
krwq: nit: copyright header (here and everywhere where missing)

src/devices/CharacterLcd/samples/CharacterLcd.Sample.cs
```diff
@@ -75,6 +75,10 @@ static void UsingHd44780OverI2C()
                     hd44780.BacklightOn = true;
                     hd44780.DisplayOn = true;
                     hd44780.Clear();
+                    Console.WriteLine("Starting test");
```
krwq: Whenever waiting for ENTER please write a message saying so

src/devices/CharacterLcd/LineFeedMode.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// </summary>
+    [Flags]
+    public enum LineFeedMode
+    {
+        /// <summary>
+        /// The cursor will stay beyond the edge of the screen. 
+        /// Any further write attempts will not do anything until the cursor is moved or the display cleared. 
+		/// A newline will move to the next line, unless already on the last line of the display. 
+        /// </summary>
+        Disable = 0,
```
krwq: I think this is usually called `Truncate`

src/devices/CharacterLcd/LineFeedMode.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Intelligently wraps to the next line, if needed. This tries to wrap at word borders.
+        /// </summary>
+        Intelligent = 0x6,
+
+        /// <summary>
+        /// If this flag is also set, the display scrolls up when the last line is full.
+        /// </summary>
+        ScrollUp = 0x10,
```
krwq: This should be separate flag somewhere in the class since it looks like separate mode

src/devices/CharacterLcd/LineFeedMode.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Iot.Device.CharacterLcd
+{
+    /// <summary>
+    /// Controls what happens when the cursor moves beyond the end of the display (both horizontally or vertically). 
+    /// </summary>
+    [Flags]
```
krwq: this shouldn't be Flags

src/devices/CharacterLcd/LineFeedMode.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// The cursor will stay beyond the edge of the screen. 
+        /// Any further write attempts will not do anything until the cursor is moved or the display cleared. 
+		/// A newline will move to the next line, unless already on the last line of the display. 
+        /// </summary>
+        Disable = 0, 
+
+        /// <summary>
+        /// Wraps to the next line if the end of the line is reached.
+        /// </summary>
+        UntilFull = 0x2,
```
krwq: `Wrap`?

src/devices/CharacterLcd/LineFeedMode.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Intelligently wraps to the next line, if needed. This tries to wrap at word borders.
+        /// </summary>
+        Intelligent = 0x6,
```
krwq: `WordWrap` (comment it as `Attempts to word wrap`) - I'd be wary of using word `Intelligent` in the API name

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+        private Encoding _characterEncoding;
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="LcdConsole"/> class using the specified LCD low-level interface. 
+        /// This class automatically configures the low-level interface. Do not use the low-level interface at the same time. 
+        /// </summary>
+        /// <param name="lcd">The low-level LCD interface.</param>
+        /// <param name="romType">Name of character ROM of display. Currently supported types: A00 and A02.</param>
+        /// <param name="shouldDispose">If the class should dispose the LCD driver when it is disposed. Defaults to true</param>
+        public LcdConsole(ICharacterLcd lcd, String romType, bool shouldDispose = true)
```
krwq: `string`. Can you clarify what `ROM` means? Read-Only Memory? If so I cannot correlate how it fits in this API

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// Position of the cursor, from top
+        /// Note: May be outside the bounds of the display.
+        /// </summary>
+        public int CursorTop
```
krwq: I think this should be `Point Cursor { get; set; }` (with implementation of set)

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// If larger than Zero, an extra wait is introduced each time a newline is written. Can be used to render entire paragraphs to the display, 
+        /// but keeping it readable. 
+        /// </summary>
+        public TimeSpan NewLineSleepTime
```
krwq: Is this used only with `ScrollUp` setting? If so I'd call it `MinimumLineDisplayTime` and clarify the summary

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+        }
+
+        /// <summary>
+        /// Enables or disables the backlight
+        /// </summary>
+        public bool BacklightOn
+        {
+            get
+            {
+                lock (_lock)
```
krwq: clarify in the summary that this class is thread-safe

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+                {
+                    return _lcd.BacklightOn;
```
krwq: does reading need a lock?

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+                    _lcd.DisplayOn = value;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Sets the Line Feed Mode. 
+        /// This defines what happens when writting past the end of the line/screen.
+        /// </summary>
+        public LineFeedMode LineFeedMode
```
krwq: I'd call this and enum `WordWrap` or `WordWrapOption`  (ing suffix fine, whichever sounds better) and move ScrollUp as a separate property

src/devices/CharacterLcd/LcdConsole.cs
krwq: Also assuming you make it settable please clarify what happens when you send a long text with ScrollUp option and set cursor while it's still displaying

krwq: Another alternative is match the APIs to be the same as Console (SetCursorPosition etc)

krwq: Hmm you already have that so this might be just getter in that case (sorry for messy comments, I comment as I review and might be missing context before I read the whole thing)

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+
+        /// <summary>
+        /// Asynchronously writes text to the display.
+        /// See <see cref="Write(string)"/> for limitations.
+        /// Only one write operation (synchronous or asynchronous) should be executed at once, otherwise the execution order 
+        /// is undefined. 
+        /// </summary>
+        /// <param name="text">Text to write</param>
+        /// <returns>A task object</returns>
+        public Task WriteAsync(string text)
```
krwq: don't add Async version if all it's doing is starting new task

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+        }
+
+        /// <summary>
+        /// Writes text, starting at the given line number. 
+        /// This will overwrite the text in the given line, filling up with spaces, if needed. Line wrapping is disabled, so this method will write exactly one line. 
+        /// </summary>
+        /// <param name="lineNumber">0-based index of the line to start</param>
+        /// <param name="text">Text to insert. No newlines supported.</param>
+        /// <exception cref="ArgumentException">The string contains newlines.</exception>
+        public void Write(int lineNumber, string text)
```
krwq: I'm not sure this should be called `Write` - before I read the description I thought this will insert on that line and wasn't sure if it will insert at start or end but it actually replaces which suggest the name should be clarified (i.e. ReplaceLine)

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+        {
+            lock (_lock)
```
krwq: please do as much work as possible outside of Lock (here and everywhere)

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+        public void WriteLine()
+        {
+            NewLine();
+        }
+
+        /// <summary>
+        /// Writes the given text to the current position, then wraps to the next line
+        /// </summary>
+        /// <param name="text">Text to draw.</param>
+        public void WriteLine(string text)
```
krwq: would it make sense to remove the no parmater overload and make this one default to empty? (you can check for string.IsNullEmpty in Write)

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+
+        /// <summary>
+        /// Loads the character set for the given culture, if available
+        /// </summary>
+        /// <param name="culture">Culture to load</param>
+        /// <param name="unknownCharacter">The character to print for unknown letters, default: ?</param>
+        /// <param name="factory">Character encoding factory that delivers the mapping of the Char type to the hardware ROM character codes. May add special characters into 
+        /// the character ROM. Default: Null (Use internal factory)</param>
+        /// <returns>True if all important characters for that culture are available and loaded.</returns>
+        public bool LoadCulture(CultureInfo culture, char unknownCharacter = '?', LcdCharacterEncodingFactory factory = null)
```
krwq: I think this should just take Encoding directly and so should method name be called (at least not seeing culture being used anywhere)

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,537 @@
...
+
+                _characterEncoding = encoding;
+                return supported;
+            }
+        }
+
+        /// <summary>
+        /// Resets the character mapping to hardware defaults
+        /// </summary>
+        public void ResetCulture()
```
krwq: Not getting this. UTF-8 should cover all characters and this seem somewhat related to old code pages

krwq:
This is really nice addition to LcdCharacter. Thanks! Is it possible to convert this code so that it's not CharacterLcd dependent and instead could be used for other things like segment displays or others? (we can leave as is for now but at least change abstraction a bit so that it it's not tied so much)

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: @krwq Yes, it means "Read only memory". Its the fixed part of the character memory of the display. For the HD44780, there are two documented ROM types, "A00", which contains japanese characters in the upper 127 bits of the character table, and "A02" which contains european dyacritics (äàÄ etc) there. I used a string instead of an enum because other LCD displays might contain other character maps not known here and this class should work against any implementation of ICharcterLcd.

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: Yes, if this is set to something like 1s, a long text scrolls trough the display in readable speed. The name is not so good, you are right there.

pgrawehr:
> 
> 
> This is really nice addition to LcdCharacter. Thanks! Is it possible to convert this code so that it's not CharacterLcd dependent and instead could be used for other things like segment displays or others? (we can leave as is for now but at least change abstraction a bit so that it it's not tied so much)

Thanks, and thanks for the review. 
I actually added ICharacterLcd interface just to make it more independent. The class is independent of the actual implementation of the display. Maybe that interface can still be tidied up a bit, but it should be implementable also by a 7 or 16-segment display (even though of course the 7-segment display will be very limited in displaying characters).

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: See a few lines down. LcdCharacterEncodingFactory.Create() takes the culture to obtain the language name.

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: This removes the custom encoding, returning back to 1:1 mapping between the Char type and the 256-byte character map of the LCD display.

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: The factory parameter is optional and allows specifying user-defined custom mappings.

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: Not sure. I once heard that overloads are preferrable over default arguments (I think that is/was even an FxCop rule), but I'm not sure that still holds true today.

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: I don't understand. The Write operation can take very long (especially if the NewLineSleepTime is big), so why would an async version not be useful?

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: Basically correct, but we shouldn't make any assumptions on the underlying interface here. This might as well trigger some device access.

krwq:
@pgrawehr I'd not refactor it at this point then (unless you already started). I'll keep these utilities in mind when someone tries to add similar thing elsewhere and we should do the merge/generalization of the code when we're ready for it (also makes reviewing easier).

src/devices/CharacterLcd/LineFeedMode.cs
pgrawehr: Yes, I've simplified that. Now this has only three distinct modes: 
- Truncate (no line wrapping at all)
- Wrap (wrapping when line is full)
- IntelligenWrap (wrap at word border)
In both wrapping modes, the display scrolls when the last line is reached. Scroll up is no longer an individual mode, as it complicates the interface and the feature to wrap up only until the display is full is not so very useful, I think.

src/devices/CharacterLcd/LineFeedMode.cs
pgrawehr: Yea, see above.

src/devices/CharacterLcd/LineFeedMode.cs
pgrawehr: It's a hard task to always come up with _intelligent_ function names. :smile:

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: Yes, it's already there.

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: You are right, this behaves differently to the Write methods. Changed to ReplaceLine.

src/devices/CharacterLcd/LcdConsole.cs
krwq: fair enough

src/devices/CharacterLcd/LcdConsole.cs
krwq: @pgrawehr the only time you should be doing Async version if you actually are trully async (i.e. have a dedicated thread always running in the background and picking up jobs and later marking task as completed) or you are trying to match existing interface. In here user can call Task.Run themselves so I believe we should not provide such overload.

src/devices/CharacterLcd/LcdConsole.cs
krwq: I don't think anyone follows that, I'd prefer we kept just 1 method until someone complains during API review in the future

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -403,19 +411,29 @@ public void CreateCustomCharacter(byte location, ReadOnlySpan<byte> characterMap
         /// '}' are usually the same with the exception of '\', which is a
         /// yen symbol on some chips '¥'.
         /// </remarks>
-        /// <param name="value">Text to be displayed.</param>
-        public void Write(string value)
+        /// <param name="text">Text to be displayed.</param>
```
krwq: nit: I think params should go before `<remarks>`

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,929 @@
...
+using System.Text;
+
+namespace Iot.Device.CharacterLcd
+{
+    /// <summary>
+    /// Factory for creating Encodings that support different cultures on different LCD Displays. 
+    /// </summary>
+    public class LcdCharacterEncodingFactory
+    {
+        private static Dictionary<char, byte> DefaultA00Map;
```
krwq: I'm assuming those dictionaries are not being anywhere further except construction? Otherwise this will cause threading issues

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,929 @@
...
+        private static Dictionary<char, byte> DefaultA02Map;
```
krwq: static fields should be called like: `s_defaultA02Map`

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,929 @@
...
+                    break;
+                case "sv":
+                    specialLetters = "åÅöÖüÜ";
+                    break;
+                case "es":
+                    specialLetters = "ñ€¿¡";
+                    break;
+                case "uk":
+                case "ru":
+                    specialLetters = "АБВГДЕЖЗИЙКЛМНОПРСТУФХЦЧШЩЪЫЬЭЮЯ"; // cycrillic script used for russian (only works on ROM A02, capital letters only)
```
krwq: cyryllic

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,929 @@
...
+
+        /// <summary>
+        /// Creates the given letter from a pixel map for Rom Type A00 (7-pixel high letters, bottom row empty)
+        /// </summary>
+        /// <param name="character">Character to create</param>
+        /// <returns>An 8-Byte array of the pixel map for the created letter.</returns>
+        /// <remarks>
+        /// Currently requires the characters to be hardcoded here. Would be nice if we could generate the pixel maps from an existing font, such as Consolas
+        /// </remarks>
+        protected virtual byte[] CreateLetterA00(Char character)
```
krwq: please use aliased type names like `char` (not capital), `string` etc

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,929 @@
...
+        {
+            Span<byte> map = stackalloc byte[8];
```
krwq: is the map used anywhere?

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,929 @@
...
+                    return CreateCustomCharacter(
+                    0b_01010,
+                    0b_00000,
+                    0b_01110,
+                    0b_10001,
+                    0b_11111,
+                    0b_10000,
+                    0b_01111,
+                    0b_00000);
+
```
krwq: nit: extra enter

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,929 @@
...
+        /// Creates the given letter from a pixel map for Rom Type A02 (7 or 8 Pixel high letters, bottom row filled)
+        /// </summary>
+        /// <param name="character">Character to create</param>
+        /// <returns>An 8-Byte array of the pixel map for the created letter.</returns>
+        /// <remarks>
+        /// Currently requires the characters to be hardcoded here. Would be nice if we could generate the pixel maps from an existing font, such as Consolas
+        /// </remarks>
+        protected virtual byte[] CreateLetterA02(Char character)
+        {
+			// TODO: Create letters for A02 map, but that one is a lot better equipped for european languages, so nothing to do for the currently supported languages
```
krwq: nit: indentation is off

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,583 @@
...
+            NewLine();
+        }
+
+        /// <summary>
+        /// Blinks the display text (and the backlight, if available).
+        /// Can be used to get user attention. 
+        /// Operation is synchronous.
+        /// </summary>
+        /// <param name="times">Number of times to blink. The blink rate is 1 Hz</param>
+        public void BlinkDisplay(int times)
```
krwq: this one makes sense to make async version directly and for sync version simply wait on the async

src/devices/CharacterLcd/LineFeedMode.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Controls what happens when the cursor moves beyond the end of the display (both horizontally or vertically). 
+    /// </summary>
+    public enum LineFeedMode
+    {
+        /// <summary>
+        /// The cursor will stay beyond the edge of the screen. 
+        /// Any further write attempts will not do anything until the cursor is moved or the display cleared. 
+		/// A newline will move to the next line, unless already on the last line of the display.
```
krwq: nit: indent is off

src/devices/CharacterLcd/samples/LcdConsole.Samples.cs
krwq: unresolving since it's still missing

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
pgrawehr: They're initialized in the static ctor and shouldn't be touched later on. As you say it... I think I need to make sure they're copied upon usage. Good point.

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
pgrawehr: They should be static readonly, so I think that's right like it is?

src/devices/CharacterLcd/samples/LcdConsole.Samples.cs
pgrawehr: Sorry, missed that one.

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: Ok, done.

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: Called it LineWrapMode now.

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
pgrawehr: It's all greek to me anyway 😄

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
pgrawehr: Nope, was a leftover.

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
krwq: I think fine to keep a name with static readonly

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -383,11 +391,11 @@ public void CreateCustomCharacter(byte location, params byte[] characterMap)
         /// <param name="characterMap">Provide an array of 8 bytes containing the pattern</param>
         public void CreateCustomCharacter(byte location, ReadOnlySpan<byte> characterMap)
         {
-            if (location > 7)
+            if (location > NumberOfCustomCharactersSupported - 1)
```
krwq: `>=` and no subtraction

krwq: `>=` and no subtraction?

src/devices/CharacterLcd/Hd44780.cs
```diff
@@ -383,11 +391,11 @@ public void CreateCustomCharacter(byte location, params byte[] characterMap)
...
                 throw new ArgumentOutOfRangeException(nameof(location));
 
             if (characterMap.Length != 8)
-                throw new ArgumentException(nameof(characterMap));
+                throw new ArgumentException("The character map must be exactly 8 bytes long", nameof(characterMap));
```
krwq:
```suggestion
                throw new ArgumentException($"The character map must be exactly {characterMap.Length} bytes long", nameof(characterMap));
```

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,927 @@
...
+
+            // The character map A00 contains the most used european letters, some greek math symbola plus japanese letters. 
+            // Compare with the HD44780 specification sheet, page 17
+            DefaultA00Map = new Dictionary<char, byte>();
+            // Inserts ASCII characters ' ' to 'z', which are common to most known character sets
+            for (char c = ' '; c <= 'z'; c++)
+            {
+                DefaultA00Map.Add(c, (byte)c);
+            }
+            DefaultA00Map.Remove('\\'); // Instead of the backspace, the Yen letter is in the map, but we can use char 164 instead
```
krwq: instead of removing can you just not add it?

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,927 @@
...
+            DefaultA00Map.Add('・', 0b1010_0101);
+            DefaultA00Map.Add('ァ', 0b1010_0111);
+            DefaultA00Map.Add('ィ', 0b1010_1000);
+            DefaultA00Map.Add('ゥ', 0b1010_1001);
+            DefaultA00Map.Add('ェ', 0b1010_1010);
+            DefaultA00Map.Add('ォ', 0b1010_1011);
+            DefaultA00Map.Add('ャ', 0b1010_1100);
+            DefaultA00Map.Add('ュ', 0b1010_1101);
+            DefaultA00Map.Add('ョ', 0b1010_1110);
+            DefaultA00Map.Add('ヮ', 0b1010_1111); // Not sure on this one
```
krwq: remove the comment or character

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,927 @@
...
+            DefaultA02Map.Add('÷', 0b1111_0111);
+            DefaultA02Map.Add('ø', 0b1111_1000);
+            DefaultA02Map.Add('ù', 0b1111_1001);
+            DefaultA02Map.Add('ú', 0b1111_1010);
+            DefaultA02Map.Add('û', 0b1111_1011);
+            DefaultA02Map.Add('ü', 0b1111_1100);
+            DefaultA02Map.Add('ý', 0b1111_1101);
+            DefaultA02Map.Add('þ', 0b1111_1110);
+            DefaultA02Map.Add('ÿ', 0b1111_1111);
+
```
krwq: remove extra enter

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,927 @@
...
+        /// <remarks>
+        /// Currently requires the characters to be hardcoded here. Would be nice if we could generate the pixel maps from an existing font, such as Consolas
+        /// </remarks>
+        protected virtual byte[] CreateLetterA00(char character)
+        {
+            switch (character)
+            {
+                case '€':
+                    return CreateCustomCharacter(
+                    0b_00111,
```
krwq: these should be indented (`+4` spaces)

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,927 @@
...
+        }
+
+        /// <summary>
+        /// Combines a set of bytes into a pixel map
+        /// </summary>
+        /// <example>
+        /// Use as follows to create letter 'ü':
+        /// <code>
+        /// case 'ü':
+        ///            return CreateCustomCharacter(
```
krwq: remove the `return` and semicolon

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,569 @@
...
+                    }
+                }
+
+                if (i != lines.Count - 1)
+                {
+                    // Insert newline except after the last text segment.
+                    NewLine();
+                }
+            }
+
```
krwq: remove extra enter

src/devices/CharacterLcd/LcdConsole.cs
pgrawehr: Separated `LoadEncoding` from `CreateEncoding` now.

src/devices/CharacterLcd/Hd44780.cs
pgrawehr: Almost... Your suggestion would always print the actual value, not the expected. 
Added a constant.

krwq:
@pgrawehr we should: https://github.com/dotnet/iot/issues/816

pgrawehr:
I had to rebase this

pgrawehr:
@krwq: Is there any reason we have so many PR's open (ie this one) with seemingly no work to be done? Is this lack of time or is it just that it's not clear whose task it is next?

krwq:
@pgrawehr I'm on parental leave, cc: @joperezr

src/devices/CharacterLcd/Hd44780.cs
krwq: wasn't fixed

src/devices/CharacterLcd/ICharacterLcd.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// <summary>
+        /// Enable/disable the display.
+        /// </summary>
+        bool DisplayOn { get; set; }
+
+        /// <summary>
+        /// Enable/disable the underline cursor.
+        /// </summary>
+        bool UnderlineCursorVisible { get; set; }
```
krwq: What's this? Should this be joined with BlinkingCursorVisible (i.e. CursorType enum)

krwq:
@buyaa-n @tarekgh can you take a look please?

src/devices/CharacterLcd/ICharacterLcd.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <summary>
+        /// Write text to the display, without any character translation.
+        /// </summary>
+        /// <remarks>
+        /// There are only 256 characters available. Different chip variants
+        /// have different character sets. Characters from space ' ' (32) to
+        /// '}' are usually the same with the exception of '\', which is a
+        /// yen symbol on some chips '¥'.
+        /// </remarks>
+        /// <param name="text">Text to be displayed.</param>
```
buyaa-n: Please move `param` section before `remark`

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,934 @@
...
+        {
+            // Default map which is used for unknown ROMs
+            DefaultCustomMap = new Dictionary<char, byte>();
+            // Inserts ASCII characters ' ' to 'z', which are common to most known character sets
+            for (char c = ' '; c <= 'z'; c++)
+            {
+                DefaultCustomMap.Add(c, (byte)c);
+            }
+
+            // The character map A00 contains the most used european letters, some greek math symbola plus japanese letters.
```
buyaa-n:
```suggestion
            // The character map A00 contains the most used european letters, some greek math symbols plus japanese letters.
```

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -0,0 +1,934 @@
...
+            // Compare with the HD44780 specification sheet, page 17
+            DefaultA00Map = new Dictionary<char, byte>();
+            // Inserts ASCII characters ' ' to 'z', which are common to most known character sets
+            for (char c = ' '; c <= 'z'; c++)
+            {
+                DefaultA00Map.Add(c, (byte)c);
```
buyaa-n: Can't we add ASCII characters for all maps using only one loop?

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
buyaa-n: Unresolving as it was not removed

src/devices/CharacterLcd/ICharacterLcd.cs
pgrawehr: This could be done, of course. There exist two types of cursors: One underline cursor (using the last row of a character cell) and one block cursor, that fills the whole character cell. I can create an enum with the possible modes.

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
pgrawehr: Possible, but not sure this increases readability.

src/devices/CharacterLcd/ICharacterLcd.cs
```diff
@@ -0,0 +1,122 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Iot.Device.CharacterLcd
+{
+    /// <summary>
+    /// Interface for character LCD Low-Level handler
+    /// </summary>
+    public interface ICharacterLcd : IDisposable
```
joperezr: If we only have one CharacterLcd, it might not be worth it to add an interface abstraction just yet, but instead wait till we have more. I specifically think of this because interfaces don't version very well in case we need to add stuff to the abstraction later

src/devices/CharacterLcd/ICharacterLcd.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        /// <summary>
+        /// Enable/disable the backlight. (Will always return false if no backlight pin was provided.)
+        /// </summary>
+        bool BacklightOn { get; set; }
```
joperezr: NIT: for naming, should we name these EnableBacklight and EnableDisplay?

src/devices/CharacterLcd/ICharacterLcd.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <summary>
+        /// Returns the size of the display.
+        /// </summary>
+        System.Drawing.Size Size { get; }
+
+        /// <summary>
+        /// Returns the number of custom characters for this display.
+        /// A custom character is one that can be user-defined and assigned to a slot using <see cref="CreateCustomCharacter(byte, byte[])"/>
+        /// </summary>
+        int NumberOfCustomCharactersSupported { get; }
```
joperezr: Will all character LCDs suuport custom characters?

src/devices/CharacterLcd/ICharacterLcd.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// book 5x7 and 5x10 is used (Advance Copy #AP4 from July 1985). Perhaps it was a
+        /// marketing change?
+        ///
+        /// As only 11 bytes are used in 5x10 mode, but 16 bytes are reserved, the last 5 bytes
+        /// are useless. The datasheet helpfully suggests that you can store your own data there.
+        /// The same would be true for bits 5-7 of lines that matter for both 5x8 and 5x10.
+        /// </remarks>
+        /// <param name="location">Should be between 0 and <see cref="NumberOfCustomCharactersSupported"/>.</param>
+        /// <param name="characterMap">Provide an array of 8 bytes containing the pattern</param>
+        void CreateCustomCharacter(byte location, params byte[] characterMap);
```
joperezr: why params? Do you expect people to use this to declare more than 1 custom character? In my mind this parameter should really only be a Span<byte>

joperezr: How do you write a custom character after saving it? There doesn't seem to be a Write method that supports this from the interface.

src/devices/CharacterLcd/LcdCharacterEncoding.cs
```diff
@@ -0,0 +1,155 @@
...
+    /// Provides the character encoding for an LCD display.
+    /// Instances of this class are generated by <see cref="LcdCharacterEncodingFactory"/>.
+    /// </summary>
+    public class LcdCharacterEncoding : Encoding
+    {
+        private Dictionary<char, byte> _characterMapping;
+        private char _unknownLetter;
+        // Character pixel maps for characters that will need to be written to the character RAM for this code page
+        private List<byte[]> _extraCharacters;
+        private String _cultureName;
```
joperezr: NIT: lower-case string

src/devices/CharacterLcd/LcdCharacterEncoding.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        /// <summary>
+        /// Creates an instance of <see cref="LcdCharacterEncoding"/>.
+        /// </summary>
+        /// <param name="cultureName">Culture name for this encoding (informational only)</param>
+        /// <param name="romName">Name of the ROM on the display</param>
```
joperezr: Do you mind using full name instead of using ROM acronym?

src/devices/CharacterLcd/ICharacterLcd.cs
joperezr: Should there be a DeleteCustomCharacter from the custom character slots? or is the idea to just replace the lcoation with empty span? Does it even make sense to clear the current custom characters

pgrawehr: `params byte[]` means "any number of bytes" (and not "a list of byte arrays"), which is almost the same as a Span<byte>, but allows someone to write
```
CreateCustomCharacter(0, 0b0000_0001,
                         0b0000_0001,
                         0b0000_0001,
                         0b0000_0001,
                         0b0000_0001,
                         0b0000_0001,
                         0b0000_0001,
                         0b0000_0001);
```
Which is easier to read than when needing an indirection with creating a Span there.

pgrawehr: The `location` parameter is the character code for the newly created char. So writing a binary 0 char will display the first custom character, etc.

pgrawehr: I think it's not required. You can always just overwrite the given character with something else (or also an empty character, if you really want to). The memory is reserved anyway, so you don't get anything back by "deleting" a character.

src/devices/CharacterLcd/ICharacterLcd.cs
pgrawehr: The ones I saw do, otherwise just set this to 0 in the implementation.

src/devices/CharacterLcd/ICharacterLcd.cs
pgrawehr: There's currently only one implementation, but users are free to implement their own. (I.e it will be rather simple to also implement this interface for 7-Segment or 16-segment displays). Also, the only current implementation is named Hd44780, not a very intuitive name to use as handle name in other places. 
Then, this is used by LcdConsole, which is independent of the actual implementation of the Hd44780 and should not be hardlinked to that. 
Therefore I do think that the interface is the right choice here.

joperezr: Is multi-inheritance a concern for implementations of this interface like Hd44780? In case it isn't, I would suggest to instead use abstract classes so that you can version them better. I know that the new csharp features include interface default implementations, but I'm not sure if those work everywhere we want to run, since we currently target netcoreapp2.1

src/devices/CharacterLcd/ICharacterLcd.cs
joperezr: I get the level of inderection, but params usually means that you can take zero, one or any number of bytes which I believe is not the case here, so I think it would be a bit confusing on how to use it. Also, the problem with params is that it is hard to add method overloads in the future if needed. I think that if you want people to be able to write the above, you should probably just have an overload that takes in a location, and then 8 byte parameters. You don't need to change this yet, I just wanted to s hare my thoughts.

joperezr: I see. Do you mind adding one example of this on the samples project? I was looking for that and couldn't find one.

src/devices/CharacterLcd/ICharacterLcd.cs
pgrawehr: In the case we have so far, a base class would do, but I can imagine cases where this would be a limitation, like when you derive from some binding to add the capability of the `ICharacterLcd`. There might be devices where it doesn't make sense to support the capabilities in the base implementation, but in derived classes only. 
I actually don't quite understand why you think that interfaces don't version well. Can you explain that?

src/devices/CharacterLcd/ICharacterLcd.cs
pgrawehr: I see your point. The problem is that other displays might take a different number of bytes (different character size). I'll think of it again.

pgrawehr: The basic example on the low-level interface is the `WalkerTest` example. (Note: Character codes 0-7 and 8-15 are identical, so if you load char 0, you actually change chars 0 and 8)

With the High-Level LcdConsole, the magic is hidden: The method `LcdCharacterEncodingFactory.AssignLettersForCurrentCulture` uses `LcdCharacterEncodingFactory.CreateLetter` to create a set of up to 8 special letters for a culture (ie german needs ÄÖÜ, which may not be part of the ROM). It then stores the index of the letter in the character encoding. 
In `LcdConsoleSamples.WriteTest`, starting from line 109, you'll find texts in different languages. With each call of the form 
```
encoding = LcdConsole.CreateEncoding(CultureInfo.CreateSpecificCulture("fr-fr"), "A00", '?', 8);
console.LoadEncoding(encoding);
```
the character set for that culture is loaded and activated. Therefore, we can print german, french and danish texts properly (of course, if you were to print the french text with the german characters loaded or vice-versa, it would not work)

pgrawehr: I could use Span<byte> in the interface and provide an extension method that converts from byte,byte,byte... to a Span.

src/devices/CharacterLcd/ICharacterLcd.cs
pgrawehr: @krwq If I change this, it would likely imply a breaking change on Hd44780, because I don't think we should use an enum in the interface but keep the bool's in the implementation. Or what do you think? 
Same applies for comment https://github.com/dotnet/iot/pull/859#discussion_r381588322

src/devices/CharacterLcd/ICharacterLcd.cs
pgrawehr: See above https://github.com/dotnet/iot/pull/859#discussion_r381024735. Should we do a breaking change for this, or just keep them both in the current implementation?

src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
pgrawehr: done

pgrawehr:
@joperezr This has some open questions/discussions for you, but could be closed soon as well, I hope.

src/devices/CharacterLcd/ICharacterLcd.cs
joperezr: > Can you explain that?

Sure, what I mean is that (at least before C# added the feature for default implementations) once you ship an interface you should never add/remove or modify members in the future because this would be a big breaking change for consumers that have already implemented the interface or taken a dependency on one of the members. Because of this, we are very careful when introducing interfaces on our API, as you basically need to make sure that you get the API shape right the first time you ship it and won't want to add anything in the future. This is why if you actually check in System.Device.Gpio, we actually don't have that many interfaces, and instead use abstract classes for polymorphism.

Anyways, for this interface in particular I'm probably not super worried, because I expect most of the classes that implement it to to actually live on this project as well, which means that taking an update on one should update the implementing classes as well, that is why I asked if multi-inheritance was a concern, if it isn't then just use abstract classes and won't have this problem, but if you are concerned about it (which seems like you are) we can keep it an interface and document the breaking change.

src/devices/CharacterLcd/ICharacterLcd.cs
joperezr: I see, didn't realize it would be breaking, I'm fine with keeping as is.

src/devices/CharacterLcd/ICharacterLcd.cs
joperezr: that sounds good to me

src/devices/CharacterLcd/LcdConsole.cs
```diff
@@ -0,0 +1,585 @@
...
+        private Encoding _characterEncoding;
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="LcdConsole"/> class using the specified LCD low-level interface.
+        /// This class automatically configures the low-level interface. Do not use the low-level interface at the same time.
+        /// </summary>
+        /// <param name="lcd">The low-level LCD interface.</param>
+        /// <param name="romType">Name of character ROM of display. Currently supported types: A00 and A02.</param>
+        /// <param name="shouldDispose">If the class should dispose the LCD driver when it is disposed. Defaults to true</param>
+        public LcdConsole(ICharacterLcd lcd, string romType, bool shouldDispose = true)
```
joperezr: Correct me if I'm wrong, but seems like you should still be able to use the lcd1602 object to perform most of the operations r ight? so this just seems to be a wrapper on top of the Lcds. If so, I wonder if we should be forcing people to just use LcdConsole directly and not allow people to instantiate the low-level lcd interfaces.

pgrawehr: It's correct that you should not be using the underlying lcd1602 class at the same time as the LcdConsole class, because that would get the internal state of the LcdConsole class out of sync with the hardware. 
However, there are cases where you want to use the underlying class explicitly, i.e. if you want to do some more graphical approaches, or use it's hardware character scrolling capabilities. 
I've got another extension ready (https://github.com/pgrawehr/iot/blob/LCDValueUnitDisplay/src/devices/CharacterLcd/LcdValueUnitDisplay.cs - I'll prepare the PR once this one is merged) that uses block graphics to draw big numbers on a 20x4 display. This would be used instead of the LcdConsole class as alternate high-level api. 
So I think that the low-level interface is still required, even though you would normally just create it to feed it to (one of) the high-level apis. In the case I'm testing now, I am also able to use both those High-Level APIs interchangeably against the same low-level instance (display measurements in big, switch to normal font for menus)

joperezr: I see, that's ok, I guess what I meant my concern was is that we usually want to avoid having two ways of doing the same thing when possible, as that might confuse consumers which one they should use, but I guess most people trying to use this API will look into samples to check different usages so that will be fine. With that cleared up then this PR is mostly good to go, I think we are just waiting on this https://github.com/dotnet/iot/pull/859#discussion_r382423975 to be addressed as well as finalize what to do about https://github.com/dotnet/iot/pull/859#discussion_r391899360

src/devices/CharacterLcd/ICharacterLcd.cs
pgrawehr: Done like that.

src/devices/CharacterLcd/ICharacterLcd.cs
pgrawehr: There's no binding for a 7-segment or a 16-segment display right now (I think), otherwise I could test whether the interface works on those, at least conceptionally. So we would know that it should hold at least for its basic purpose.

pgrawehr: @joperezr : After addressing https://github.com/dotnet/iot/pull/859#discussion_r394592045, I think it's still the most flexible approach to leave this as interface. The interface has basically the same methods as the Hd44780 implementation had previously, so I guess it should be rather stable. Or do you think there's anything missing or not generic enough? 
Otherwise, this would be ready for merge.

Issue #850, opened 2019-11-14
DhtSensor DhtType.Dht11 always returns NaN on .net core running a GrovePi+ on a Rasbperry Pi 2B+ with Rasbpian OS


JodyPetroni:
I am trying to get the temperature and humidity sensor working on .net core running a GrovePi+ on a Rasbperry Pi with Rasbpian OS. This code always produces NaN output.

```
Console.WriteLine("Hello GrovePi!");
            
I2cConnectionSettings i2CConnectionSettings = new I2cConnectionSettings(1, GrovePi.DefaultI2cAddress);
grovePi = new GrovePi(I2cDevice.Create(i2CConnectionSettings));

Console.WriteLine($"Manufacturer :{grovePi.GrovePiInfo.Manufacturer}");
Console.WriteLine($"Board: {grovePi.GrovePiInfo.Board}");
Console.WriteLine($"Firmware version: {grovePi.GrovePiInfo.SoftwareVersion}");
DhtSensor dhtSensor = new DhtSensor(grovePi, GrovePort.DigitalPin4, DhtType.Dht11);
while (true)
 {
       dhtSensor.Read();
       Console.WriteLine($"{dhtSensor.DhtType}: {dhtSensor}");

       Thread.Sleep(2000);
 }
```

output is 

```
Hello GrovePi!
Manufacturer :Dexter Industries
Board: GrovePi+
Firmware version: 1.4.0
Dht11: Temperature: NaN °C, Humidity: NaN %
Dht11: Temperature: NaN °C, Humidity: NaN %
Dht11: Temperature: NaN °C, Humidity: NaN %
Dht11: Temperature: NaN °C, Humidity: NaN %
Dht11: Temperature: NaN °C, Humidity: NaN %
Dht11: Temperature: NaN °C, Humidity: NaN %
```

kurianov-pavel:
@JodyPetroni Try to use something like that
`
var dht11 = new Dht11(YOUR_PIN, PinNumberingScheme.Logical);

                if (!dht11.IsLastReadSuccessful)
                    return result;

                result.Data = new DthData
                {
                    Humidity = dht11.Humidity,
                    Temperature = dht11.Temperature.Celsius
                };
                result.IsSuccess = true;
`

and you can add delay time. Sometimes it helps (like one read data every/per 2 seconds)

JodyPetroni:
alas same result

kurianov-pavel:
> alas same result

Do you use it?

[...]

PR #873, opened 2019-12-01
fixing DHT sensor issue with GrovePi firmware 1.4

Ellerbach:
Fixing issue "DhtSensor DhtType.Dht11 always returns NaN on .net core running a GrovePi+": https://github.com/dotnet/iot/issues/850
Tested with firmware 1.2.2 and 1.4.0

### DIFF
--- a/src/devices/GrovePi/Sensors/DhtSensor.cs
+++ b/src/devices/GrovePi/Sensors/DhtSensor.cs
@@ -69,6 +69,7 @@ namespace Iot.Device.GrovePiDevice.Sensors
             {
                 _grovePi.WriteCommand(GrovePiCommand.DhtTemp, _port, (byte)DhtType, 0);
             }
+            
             // Wait a little bit to read the result
             // Delay from source code, line 90 in the Grove Pi firmware repository
             // This is needed for firmware 1.4.0 and also working for previous versions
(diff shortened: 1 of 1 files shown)


src/devices/GrovePi/Sensors/DhtSensor.cs
```diff
@@ -63,9 +63,16 @@ public double[] Value
         /// </summary>
         public void Read()
         {
-            _grovePi.WriteCommand(GrovePiCommand.DhtTemp, _port, (byte)DhtType, 0);
+            // Fix for the firmware 1.4.0, you can request a new measurement only if you got data
+            // from the previous one
+            if (!(double.IsNaN(_lastTemHum[0]) && double.IsNaN(_lastTemHum[1])))
+            {
+                _grovePi.WriteCommand(GrovePiCommand.DhtTemp, _port, (byte)DhtType, 0);
+            }
```
krwq: nit: enter after `}`

src/devices/GrovePi/Sensors/DhtSensor.cs
```diff
@@ -63,9 +63,16 @@ public double[] Value
...
             // Wait a little bit to read the result
-            Thread.Sleep(50);
+            // Delay from source code, line 90 in the Grove Pi firmware repository
+            // This is needed for firmware 1.4.0 and also working for previous versions
+            Thread.Sleep(300);
```
krwq: is this large delay always needed or only the first time when both values are NaN?

Ellerbach: Yes, it is a large delay. If I don't do a delay like this, you'll get multiple NaN after reading. I know most people will prefer to read a correct value even if it takes long time.
In the previous version of the firmware, the old value was returned so you always had a value. With this new version, you get a NaN if the value is not present.

PR #876, opened 2019-12-02
PWM fixes from #758

krwq:
Fixes https://github.com/dotnet/iot/issues/757
Also added waiting after PWM channel is exported in case driver didn't show directory in time (happened at least once during testing)

This is extracting PWM fixes from #758

cc: @pgrawehr @joperezr

### DIFF
--- a/src/System.Device.Gpio/System.Device.Gpio.csproj
+++ b/src/System.Device.Gpio/System.Device.Gpio.csproj
@@ -34,6 +34,7 @@
   <ItemGroup Condition="'$(TargetsWindows)' == 'true'">
     <!--Excluding Linux implementations-->
     <Compile Remove="**\*.Linux.cs" />
+    <Compile Remove="**\*.Unix.cs" />
     <Compile Remove="Interop\Unix\**\*.cs" />
     <Reference Include="Windows.Devices.DevicesLowLevelContract">
       <HintPath>winmd\Windows.Devices.DevicesLowLevelContract.winmd</HintPath>
--- a/src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
@@ -209,10 +209,7 @@ namespace System.Device.Pwm.Channels
                 File.WriteAllText($"{_chipPath}/export", Convert.ToString(_channel));
             }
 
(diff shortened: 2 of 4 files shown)


src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.Linux.cs
```diff
@@ -169,6 +208,11 @@ private void Open()
             {
                 File.WriteAllText($"{_chipPath}/export", Convert.ToString(_channel));
             }
+
+            while (!Directory.Exists(_channelPath))
```
pgrawehr: Add timeout?

pgrawehr: When we're waiting anyway, we could also wait for permissions.

krwq: We should unify those permissions utilities from both PRs

krwq:
@Frankenslag any chance you can review since this unifies the permission logic into the PWM as well?

src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -34,6 +34,7 @@
   <ItemGroup Condition="'$(TargetsWindows)' == 'true'">
     <!--Excluding Linux implementations-->
     <Compile Remove="**\*.Linux.cs" />
+    <Compile Remove="**\*.Unix.cs" />
```
joperezr: Why do we need to add the Unix suffix? I would either use the existing Linux one, or move all the Linux stuff to have the Linux suffix but not add a new one. The suffix is meant to represent the target OS platform, and today we still only target Linux and Windows, we don't have a separate target for Unix so I would remove this.

krwq: This file is Unix because it uses Unix specific interop file coming from `Unix` folder. Just followed convention from Interop. You can use any Unix stuff on any Linux so don't see a problem

joperezr: the suffix is not intended to relate to the code used in the file, but instead to show which build configuration will the file be included in, and we just have a Linux and a win build, not a Unix one, so I would change it back to be Linux in order to be consistent and simple.

PR #877, opened 2019-12-02
Add check for null SpiDevice in Mcp3xxx

krwq:
This is extracting Mcp3xxx fix from #758 so that the other PR has only lab upgrade changes


krwq:
@joperezr merging this since there is very little room for error here

PR #879, opened 2019-12-06
Adding support for Win32 and Win64 SPI and I2C over FT4222

Ellerbach:
This initial PR offers a full SPI and I2C support on Win64 and Win32 over the FTDI Chip FT4222.
Implemented:
- Full SPI driver System.Device.Spi.SpiDevice
- Full I2C driver as System.Device.I2c.I2cDevice
- Both drivers tested with different devices already implemented like the BNO055
- Documentation and simple example

Notes:
- This implementation is more for developement than production as SPI and I2C are over USB so this can involve additional delays
- This does require to download an additional library from FTDI Chip (added into the docuementation)
- GPIO support can be added but facing an issue with marshaling on the FT4222_GPIO_Init function
- MacOS and Linux 64 bits can be added easilly as the FTI Chip library is compiled as well for those environement
- The project is marked as "win" in the project settings:
```xml
    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
      <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=win</AdditionalProperties>
    </ProjectReference>
```
Not sure about the real impact. So far the files are not names with ```.windows``` in the name.


src/devices/Ft4222/DeviceInformation.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    /// <summary>
+    /// FT4222 device information    
+    /// </summary>
+    public class DeviceInformation
+    {
+        /// <summary>
+        /// Indicates device state.  Can be any combination of the following: FT_FLAGS_OPENED, FT_FLAGS_HISPEED
+        /// </summary>
+        public FtFlag Flags;
```
krwq: make these properties unless there is some reason to keep them fields

src/devices/Ft4222/Ft4222.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+ <ItemGroup>
+    <Compile Include="*.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
+      <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=win</AdditionalProperties>
```
krwq: @joperezr how should we correctly add also linux in the next PR? Package reference instead for this specific project?

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -0,0 +1,145 @@
...
+    /// </summary>
+    public class Ft4222I2c : I2cDevice
+    {
+        private I2cConnectionSettings _settings;
+        private IntPtr _ftHandle = new IntPtr();
+
+        /// <summary>
+        /// Store the FTDI Device Information
+        /// </summary>
+        public DeviceInformation DeviceInformation { get; internal set; }
```
krwq: wouldn't `private set` suffice?

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            // Set the clock
+            FtClockRate ft4222Clock = FtClockRate.Clock24MHz;
+
+            ftStatus = FtFunction.FT4222_SetClock(_ftHandle, ft4222Clock);
+            if (ftStatus != FtStatus.Ok)
+                throw new IOException($"Failed set clock rate {ft4222Clock} on device: {DeviceInformation.Description}, status: {ftStatus}");
+
+            // Set the device as I2C Master
+            ftStatus = FtFunction.FT4222_I2CMaster_Init(_ftHandle, 400);
```
krwq: what's 400?

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// </summary>
+        /// <param name="buffer">
+        /// The buffer to read the data from the I2C device.
+        /// The length of the buffer determines how much data to read from the I2C device.
+        /// </param>
+        public override void Read(Span<byte> buffer)
+        {
+            byte[] buff = new byte[buffer.Length];
+            ushort byteRead = 0;
+            var ftStatus = FtFunction.FT4222_I2CMaster_Read(_ftHandle, (ushort)_settings.DeviceAddress, buff, (ushort)buff.Length, ref byteRead);
```
krwq: can't span be passed directly here? You can convert it to pointer with fixed statement or use `ref byte` in the signature  and pass in `ref MemoryMarshal.GetReference(buffer)`

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -0,0 +1,145 @@
...
+            buff.CopyTo(buffer);
+        }
+
+        /// <summary>
+        /// Reads a byte from the I2C device.
+        /// </summary>
+        /// <returns>A byte read from the I2C device.</returns>
+        public override byte ReadByte()
+        {
+            byte[] toRead = new byte[1];
```
krwq: stackalloc and span?

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// Writes data to the I2C device.
+        /// </summary>
+        /// <param name="buffer">
+        /// The buffer that contains the data to be written to the I2C device.
+        /// The data should not include the I2C device address.
+        /// </param>
+        public override void Write(ReadOnlySpan<byte> buffer)
+        {
+            ushort byteSent = 0;
+            var ftStatus = FtFunction.FT4222_I2CMaster_Write(_ftHandle, (ushort)_settings.DeviceAddress, buffer.ToArray(), (ushort)buffer.Length, ref byteSent);
```
krwq: span should just work

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -0,0 +1,145 @@
...
+                throw new IOException($"{nameof(Write)} failed to write, error: {ftStatus}");
+        }
+
+        /// <summary>
+        /// Writes a byte to the I2C device.
+        /// </summary>
+        /// <param name="value">The byte to be written to the I2C device.</param>
+        public override void WriteByte(byte value)
+        {
+            Write(new byte[] { value });
```
krwq: stackalloc + span

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -0,0 +1,145 @@
...
+        public override void WriteRead(ReadOnlySpan<byte> writeBuffer, Span<byte> readBuffer)
+        {
+            Write(writeBuffer);
+            Read(readBuffer);
+        }
+
+        /// <summary>
+        /// Dispose the class
+        /// </summary>
+        /// <param name="disposing"></param>
```
krwq: description is missing (and you can just use `/// </inheritdoc>`)

src/devices/Ft4222/DeviceInformation.cs
Ellerbach: Yes, there was a good reason why I didn't use property is because I'm using the members in a call like this:
```csharp
ftStatus = FtFunction.FT_GetDeviceInfoDetail(0, ref flags, ref devInfo.Type, ref devInfo.Id, ref devInfo.LocId, sernum, desc, ref devInfo.FtHandle);
```
Giving the fact that the FT_GetDeviceInfoDetail function is from the external dll, the ref would not work with properties. Now if you have a good way to do this, except creating temporary variables, I'm taking it.
If it's necessary to create properties, I'll go for the temporary variables if nothing better.

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -0,0 +1,215 @@
...
+
+namespace System.Device.Spi
+{
+    /// <summary>
+    /// Create a SPI Device based on FT4222 chipset
+    /// </summary>
+    public class Ft4222Spi : SpiDevice
+    {
+        private readonly SpiConnectionSettings _settings;
+        private IntPtr _ftHandle = new IntPtr();
```
krwq: if not too much work it would be beneficial to wrap this handle with SafeHandle (anywhere where it's used) - this make it much harder to leak

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        /// <summary>
+        /// The connection settings of a device on a SPI bus. The connection settings are immutable after the device is created
+        /// so the object returned will be a clone of the settings object.
+        /// </summary>
```
krwq: inheritdoc

krwq: here and everywhere where overriding

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -0,0 +1,215 @@
...
+        public Ft4222Spi(SpiConnectionSettings settings)
+        {
+            _settings = settings;
+            // Check device
+            var devInfos = FtCommon.GetDevices();
+            // Select the one from bus Id
+            DeviceInformation = devInfos[_settings.BusId];
+
+            // Open device
+            var ftStatus = FtFunction.FT_OpenEx(DeviceInformation.LocId, FtOpenType.OpenByLocation, ref _ftHandle);
```
krwq: this suggests the handle should probably be `out` (then you don't need to initialize the handle before calling this)

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -0,0 +1,215 @@
...
+
+            // Set the clock but we need some math
+            var (ft4222Clock, tfSpiDiv) = CalculateBestClockRate();
+
+            ftStatus = FtFunction.FT4222_SetClock(_ftHandle, ft4222Clock);
+            if (ftStatus != FtStatus.Ok)
+                throw new IOException($"Failed set clock rate {ft4222Clock} on device: {DeviceInformation.Description}with error: {ftStatus}");
+
+
+            SpiClockPolarity pol = SpiClockPolarity.ClockIdleLow;
```
krwq: I called it completely differently in SoftwareSpi :smile: Presumably those are correct names for those settings

src/devices/Ft4222/Ft4222I2c.cs
Ellerbach: Good catch, the I2C Frequency in kbps. Added as a constant. Btw, I didn't find a way to pass this in the I2cConnectionSettings, is there somewhere a way to change it while constructing the I2C class?

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -0,0 +1,215 @@
...
+                throw new IOException($"Failed setup SPI on device: {DeviceInformation.Description} with error: {ftStatus}");
+        }
+
+        private (FtClockRate clk, SpiClock spiClk) CalculateBestClockRate()
+        {
+            // Maximum is the System Clock / 1 = 80 MHz
+            // Minimum is the System Clock / 512 = 24 / 256 = 93.75 KHz
+            // Always take the below frequency to avoid over clocking
+            if (_settings.ClockFrequency < 187500)
+                return (FtClockRate.Clock24MHz, SpiClock.DivideBy256);
```
krwq: should this be 48MHz?

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -0,0 +1,215 @@
...
+            if (_settings.ClockFrequency < 234375)
+                return (FtClockRate.Clock24MHz, SpiClock.DivideBy256);
```
krwq: Can these be calculated in more automated way so it's less error prone?

I.e. you can calculate all possible frequencies by foreaching each clock freq and divider, then store them as:
freq => (clock, divider) and then binary search (or just iterate over them since there is only few possible values)

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -0,0 +1,215 @@
...
+            if (_settings.ClockFrequency < 80000000)
+                return (FtClockRate.Clock60MHz, SpiClock.DivideBy1);
+            // Anything else will be 80 MHz
+            return (FtClockRate.Clock80MHz, SpiClock.DivideBy1);
+        }
+
+        /// <summary>
+        /// Reads data from the SPI device.
+        /// </summary>
+        /// <param name="buffer">
```
krwq: inheritdoc

src/devices/Ft4222/Ft4222I2c.cs
Ellerbach: I confirm, it's working. I'll adjust all including the SPI functions as well.

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -0,0 +1,215 @@
...
+            readBuff.CopyTo(buffer);
+        }
+
+        /// <summary>
+        /// Reads a byte from the SPI device.
+        /// </summary>
+        /// <returns>A byte read from the SPI device.</returns>
+        public override byte ReadByte()
+        {
+            byte[] toRead = new byte[1];
```
krwq: similar comments regarding to span usage as with I2C

src/devices/Ft4222/FtClockRate.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+
```
krwq: nit: extra enter

src/devices/Ft4222/FtClockRate.cs
```diff
@@ -0,0 +1,31 @@
...
+        Clock24MHz,
+        /// <summary>
+        /// 48 MHz
+        /// </summary>
+        Clock48MHz,
+        /// <summary>
+        /// 80 MHz
+        /// </summary>
+        Clock80MHz,
+
```
krwq: nit: extra enter

src/devices/Ft4222/FtClockRate.cs
```diff
@@ -0,0 +1,31 @@
...
+    };
```
krwq: semicolon not needed (unlike C/C++ :smiling_imp:)

src/devices/Ft4222/Ft4222I2c.cs
Ellerbach: I changed as well the signature for the ref as above. So I'm passing the pointer directly

src/devices/Ft4222/FtCommon.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// </summary>
+        /// <returns>A list of devices connected</returns>
+        public static List<DeviceInformation> GetDevices()
+        {
+            List<DeviceInformation> devInfos = new List<DeviceInformation>(); ;
+            FtStatus ftStatus = 0;
+
+            // Check device
+            uint numOfDevices = 0;
+            ftStatus = FtFunction.FT_CreateDeviceInfoList(ref numOfDevices);
```
krwq: should this be out?

src/devices/Ft4222/FtCommon.cs
```diff
@@ -0,0 +1,92 @@
...
+            if (numOfDevices == 0)
+                throw new IOException($"No device found");
+
+            byte[] sernum = new byte[16];
+            byte[] desc = new byte[64];
+            uint flags = 0;
+            for (int i = 0; i < numOfDevices; i++)
+            {
+                var devInfo = new DeviceInformation();
+                ftStatus = FtFunction.FT_GetDeviceInfoDetail(0, ref flags, ref devInfo.Type, ref devInfo.Id, ref devInfo.LocId, sernum, desc, ref devInfo.FtHandle);
```
krwq: `out`?

src/devices/Ft4222/FtCommon.cs
```diff
@@ -0,0 +1,92 @@
...
+                return (null, null);
+            // Check if the first not open device
+            int idx = 0;
+            for (idx = 0; idx < devices.Count; idx++)
+            {
+                if ((devices[idx].Flags & FtFlag.PortOpened) != FtFlag.PortOpened)
+                    break;
+            }
+            if (idx == devices.Count)
+                throw new Exception($"Can't find any closed device to check the versions");
```
krwq: InvalidOperationException (I think)

src/devices/Ft4222/FtDevice.cs
```diff
@@ -0,0 +1,70 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+
+namespace System.Device.Ft4222
+{
+    /// <summary>
+    /// List of FTDI device types
+    /// </summary>
+    public enum FtDevice
```
krwq: since this supports the whole family should you rename the namespace/directory?

src/devices/Ft4222/FtDevice.cs
```diff
@@ -0,0 +1,70 @@
...
+        Ft4222HMode1or2With4Interfaces,
+        /// <summary>
+        /// FT4222 hi-speed device Mode 3 - 1 interface
+        /// </summary>
+        Ft4222HMode3With1Interface,
+        /// <summary>
+        /// OTP programmer board for the FT4222.
+        /// </summary>
+        Ft4222OtpProgrammerBoard,
+    };
```
krwq:
```suggestion
    }
```

src/devices/Ft4222/FtFunction.cs
```diff
@@ -0,0 +1,561 @@
...
+        /// <returns>The status</returns>
+        [DllImport("LibFT4222.dll", CallingConvention = CallingConvention.Cdecl)]
+        public static extern FtStatus FT4222_I2CSlave_SetRespWord(IntPtr ftHandle, byte responseWord);
+
+        #endregion
+
+        #region GPIO
+
+        /// <summary>
+        /// Initialize the chi as GPIO
```
krwq:
```suggestion
        /// Initialize the chip as GPIO
```

Although I wish it was chi :smile:

src/devices/Ft4222/FtFunction.cs
```diff
@@ -0,0 +1,561 @@
...
+        /// </summary>
+        /// <param name="ftHandle">The handle of the open device</param>
+        /// <param name="gpioDir">Array of pin configuration</param>
+        /// <returns>The status</returns>
+        [DllImport("LibFT4222.dll", CallingConvention = CallingConvention.Cdecl)]
+        public static extern FtStatus FT4222_GPIO_Init(IntPtr ftHandle, [MarshalAs(UnmanagedType.LPArray, SizeConst = 4)] GpioPinMode[] gpioDir);
```
krwq: IIRC this should ByValArray

src/devices/Ft4222/FtFunction.cs
```diff
@@ -0,0 +1,561 @@
...
+
+        /// <summary>
+        /// Operate a GPIO read
+        /// </summary>
+        /// <param name="ftHandle">The handle of the open device</param>
+        /// <param name="portNum">The pin port</param>
+        /// <param name="value">True if high, false if low</param>
+        /// <returns>The status</returns>
+        [DllImport("LibFT4222.dll", CallingConvention = CallingConvention.Cdecl)]
+        public static extern FtStatus FT4222_GPIO_Read(IntPtr ftHandle, GpioPort portNum, ref GpioPinValue value);
```
krwq: `out` and similarly for other APIs which don't care about the input

src/devices/Ft4222/Ft4222Spi.cs
Ellerbach: Do you have a good example of SafeHandle? Something I can get the best practice out.

src/devices/Ft4222/FtStatus.cs
```diff
@@ -0,0 +1,180 @@
...
+        InterruptNotSupported,
+        /// <summary>
+        /// GPIO input not supported
+        /// </summary>
+        GpioInputNotSupported,
+        /// <summary>
+        /// Event not supported
+        /// </summary>
+        EventNotSupported,
+    };
```
krwq:
```suggestion
    }
```

src/devices/Ft4222/FtVersion.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// Contains the chip and Dll versions
+    /// </summary>
+    public struct FtVersion
+    {
+        /// <summary>
+        /// Chip version
+        /// </summary>
+        public uint ChipVersion;
```
krwq: properties

src/devices/Ft4222/PinDrivingStrength.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Intensity for SPI in Milli Amperes
+    /// </summary>
+    internal enum PinDrivingStrength
+    {
+        /// <summary>
+        /// 4 Milli Amperes
+        /// </summary>
+        Intensity4Ma = 0,
```
krwq: I'd go with regular casing here since this is just confusing (Intensity4mA etc) - I feel like this is one of the places where guidelines makes it confusing

src/devices/Ft4222/PinDrivingStrength.cs
```diff
@@ -0,0 +1,29 @@
...
+        Intensity8Ma,
+        /// <summary>
+        /// 12 Milli Amperes
+        /// </summary>
+        Intensity12Ma,
+        /// <summary>
+        /// 16 Milli Amperes
+        /// </summary>
+        Intensity16Ma,
+    };
```
krwq:
```suggestion
    }
```

src/devices/Ft4222/README.md
```diff
@@ -0,0 +1,102 @@
+# SPI and I2C
+
+This project support SPI and I2C into a normal Windows 64 bits or Windows 32 bits environment thru FT4222 chipset. MacOS and Linux 64 bits can be added as well.
```
krwq:
```suggestion
This project supports SPI and I2C into a normal Windows 64 bits or Windows 32 bits environment thru FT4222 chipset. MacOS and Linux 64 are not supported yet.
```

src/devices/Ft4222/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Device family
+
+This device supports multiple SPI as well as GPIO and I2C. It is a [FT4222](https://www.ftdichip.com/Products/ICs/FT4222H.html) from FTDI Chip.
```
krwq: GPIO is not supported yet?

src/devices/Ft4222/README.md
```diff
@@ -0,0 +1,102 @@
...
+In order to have this FTDI board working and getting support for SPI and I2C, you need to install in a path the ```LibFT4222.dll``` provided by FTDI Chip. You can find the latest version [here](https://www.ftdichip.com/Products/ICs/FT4222H.html).
+The version used to build this project is 1.4.2 and you can download it directly from FTDI [here](https://www.ftdichip.com/Support/SoftwareExamples/LibFT4222-v1.4.2.zip).
+
+### Running it on a Windows 64 bit version
+
+You will need to unzip the file and go to ```LibFT4222-v1.4.2\imports\LibFT4222\lib\amd64```, copy ```LibFT4222-64.dll``` to ```LibFT4222.dll``` into your path or in the same directory as the executable you are launching.
+
+### Running it on a Windows 32 bit version
+
+You will need to unzip the file and go to ```LibFT4222-v1.4.2\imports\LibFT4222\lib\i386```, copy ```LibFT4222.dll``` to your path or in the same directory as the executable you are launching.
```
krwq: I think you can also register DLL globally with REGSVR32

src/devices/Ft4222/README.md
```diff
@@ -0,0 +1,102 @@
...
+```Ft4222I2c``` is the I2C driver which you can pass later to any device requiring I2C or directly use it to send I2C commands. The I2C implementation is fully compatible with ```System.Device.I2c.I2cDevice```.
+
+Form the ```I2cConnectionSettings``` class that you are passing, the ```BusId``` is the FTDI device index you want to use. 
+
+The example below shows how to create the I2C device and pass it to a BNO055 sensor. This sensor is the one which has been used to stress test the implementation.
+
+```csharp
+var winFtdiI2C = new Ft4222I2c(new I2cConnectionSettings(0, Bno055Sensor.DefaultI2cAddress));
+
+Bno055Sensor bno055Sensor = new Bno055Sensor(winFtdiI2C);
```
krwq:
```suggestion
var bno055Sensor = new Bno055Sensor(winFtdiI2C);
```

src/devices/Ft4222/Ft4222Spi.cs
Ellerbach: No, this one is correct. But, thanks for pointing out, the next one has to be 48.

src/devices/Ft4222/README.md
```diff
@@ -0,0 +1,102 @@
...
+### SPI
+
+```Ft4222I2c``` is the SPI driver which you can pass later to any device requiring SPI or directly use it to send SPI commands. The SPI implementation is fully compatible with ```System.Device.Spi.SpiDevice```.
+
+From the ```SpiConnectionSettings``` class that you are passing, the ```BusId``` is the FTDI device index you want to use.
+
+The example below shows how to blink leds out of a HC595 connected to the SPI outpout using chipselect 1.
+
+```csharp
+Ft4222Spi ftSpi = new Ft4222Spi(new SpiConnectionSettings(0, 1) { ClockFrequency = 1_000_000, Mode = SpiMode.Mode0 });
```
krwq:
```suggestion
var ftSpi = new Ft4222Spi(new SpiConnectionSettings(0, 1) { ClockFrequency = 1_000_000, Mode = SpiMode.Mode0 });
```

src/devices/Ft4222/README.md
```diff
@@ -0,0 +1,102 @@
...
+while (!Console.KeyAvailable)
+{
+    ftSpi.WriteByte(0xFF);
+    Thread.Sleep(500);
+    ftSpi.WriteByte(0x00);
+    Thread.Sleep(500);
+}
+```
+
+## Know limitations
```
krwq:
```suggestion
## Known limitations
```

src/devices/Ft4222/SpiClock.cs
```diff
@@ -0,0 +1,53 @@
...
+        DivideBy128,
+        /// <summary>
+        /// 1/256 System Clock (ClockRate)
+        /// </summary>
+        DivideBy256,
+        /// <summary>
+        /// 1/512 System Clock (ClockRate)
+        /// </summary>
+        DivideBy512,
+    };
```
krwq:
```suggestion
    }
```

src/devices/Ft4222/SpiClockPhase.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace System.Device.Ft4222
+{
+    /// <summary>
+    /// SPI Clock Phase
+    /// </summary>
+    internal enum SpiClockPhase
+    {
+        ClockLeading = 0,
+        ClockTailing = 1,
+    };
```
krwq:
```suggestion
    }
```

src/devices/Ft4222/SpiClockPolarity.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace System.Device.Ft4222
+{
+    /// <summary>
+    /// SPI Clock Polarity
+    /// </summary>
+    internal enum SpiClockPolarity
+    {
+        ClockIdleLow = 0,
+        ClockIdelHigh = 1,
+    };
```
krwq:
```suggestion
    }
```

src/devices/Ft4222/SpiDrivingStrength.cs
```diff
@@ -0,0 +1,29 @@
...
+        Intensity8Ma,
+        /// <summary>
+        /// 12 Milli Amperes
+        /// </summary>
+        Intensity12Ma,
+        /// <summary>
+        /// 16 Milli Amperes
+        /// </summary>
+        Intensity16Ma,
+    };
```
krwq:
```suggestion
    }
```

src/devices/Ft4222/SpiDrivingStrength.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Intensity for SPI in Milli Amperes
+    /// </summary>
+    internal enum SpiDrivingStrength
+    {
+        /// <summary>
+        /// 4 Milli Amperes
+        /// </summary>
+        Intensity4Ma = 0,
```
krwq: use regular casing for mA, as is suggests different unit

src/devices/Ft4222/SpiOperatingMode.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// <summary>
+    /// SPI Operation mode as single, dual or quad SPI
+    /// </summary>
+    internal enum SpiOperatingMode
+    {
+        None = 0,
+        Single = 1,
+        Dual = 2,
+        Quad = 4,
+    };
```
krwq:
```suggestion
    }
```

src/devices/Ft4222/Ft4222Spi.cs
Ellerbach: I tried that and it was even more messy. Reason is as we have 24 and 48, we have which are not unique. So it adds a bit of complexity when reading the values. Now, if you want, I can do it, just not sure it will be more clear.

src/devices/Ft4222/SpiSlaveProtocol.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// </summary>
+        WithProtocol = 0,
+        /// <summary>
+        /// Without Protocol
+        /// </summary>
+        WithoutProtocol,
+        /// <summary>
+        /// Never send acknowledge
+        /// </summary>
+        NeverSendAcknoledge,
```
krwq:
```suggestion
        NeverSendAcknowledge,
```

src/devices/Ft4222/category.txt
```diff
@@ -0,0 +1,2 @@
+spi
+i2c
```
krwq: also add `usb`

src/devices/Ft4222/FtFunction.cs
Ellerbach: I'm still trying to figure out what is the right way of marshaling this one. I've tried many things and it seems that it's not working. What puzzle me is that when calling this function, it returns that all is initialized but writing or reading doesn't change the status of the pins. So I'll change this one most likely

src/devices/Ft4222/README.md
Ellerbach: It is now in this new commit

src/devices/Ft4222/Ft4222Gpio.cs
```diff
@@ -0,0 +1,184 @@
...
+namespace System.Device.Gpio
+{
+    /// <summary>
+    /// GPIO driver for the FT4222
+    /// </summary>
+    public class Ft4222Gpio : GpioDriver
+    {
+        private const int PinCountConst = 4;
+
+        private SafeFtHandle _ftHandle = new SafeFtHandle();
```
krwq: I don't think the initializer should be required since you initialize it in the constructor

src/devices/Ft4222/Ft4222Gpio.cs
```diff
@@ -0,0 +1,184 @@
...
+        [MarshalAs(UnmanagedType.ByValArray, SizeConst = 4)]
```
krwq: does this attribute have any effect? I would imagine that would only matter if Ft4222Gpio is being passed to the native code (i.e. on the interop method) but that's not happening here

src/devices/Ft4222/Ft4222Gpio.cs
```diff
@@ -0,0 +1,184 @@
...
+            ftStatus = FtFunction.FT4222_GPIO_Init(_ftHandle, _gpioDirections);
+
+            if (ftStatus != FtStatus.Ok)
+                throw new IOException($"Can't initialize GPIO for device {DeviceInformation.Description}, status: {ftStatus}");
+        }
+
+        /// <inheritdoc/>
+        protected override void ClosePin(int pinNumber)
+        {
+            if ((pinNumber < 0) || (pinNumber >= PinCountConst))
```
krwq: I don't believe you need to check ranges here, @joperezr can you confirm?

src/devices/Ft4222/Ft4222Gpio.cs
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        /// <inheritdoc/>
+        protected override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEventTypes eventTypes, PinChangeEventHandler callback)
+        {
+            // TODO: implement callback
+            // This can be done using the function FT4222_GPIO_SetInputTrigger to set them up
+            // Then calling in an infinite loop FT4222_GPIO_GetTriggerStatus
+            // and finally FT4222_GPIO_ReadTriggerQueue to understand what has trigged
+            throw new PlatformNotSupportedException();
```
krwq: NotImplementedException for unfinished things

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -12,13 +14,15 @@ namespace System.Device.I2c
     /// </summary>
     public class Ft4222I2c : I2cDevice
     {
+        const uint I2cMAsterFrequencyKbps = 400;
+
         private I2cConnectionSettings _settings;
-        private IntPtr _ftHandle = new IntPtr();
+        private SafeFtHandle _ftHandle = new SafeFtHandle();
```
krwq: I don't believe the initializer should be required

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -12,13 +14,15 @@ namespace System.Device.I2c
     /// </summary>
     public class Ft4222I2c : I2cDevice
     {
+        const uint I2cMAsterFrequencyKbps = 400;
```
krwq: `I2cMAsterFrequencyKbps` => `I2cMasterFrequencyKbps` (note: mAster)

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -46,98 +56,59 @@ public Ft4222I2c(I2cConnectionSettings settings)
...
-        /// The buffer that contains the data to be written to the I2C device.
-        /// The data should not include the I2C device address.
-        /// </param>
+        /// <inheritdoc/>
         public override void Write(ReadOnlySpan<byte> buffer)
         {
-            ushort byteSent = 0;
-            var ftStatus = FtFunction.FT4222_I2CMaster_Write(_ftHandle, (ushort)_settings.DeviceAddress, buffer.ToArray(), (ushort)buffer.Length, ref byteSent);
+            ushort byteSent;
+            var ftStatus = FtFunction.FT4222_I2CMaster_Write(_ftHandle, (ushort)_settings.DeviceAddress, in MemoryMarshal.GetReference(buffer), (ushort)buffer.Length, out byteSent);
```
krwq: I haven't seen anyone ever using `in` for buffers but it's actually "correcter"

src/devices/Ft4222/Ft4222I2c.cs
```diff
@@ -46,98 +56,59 @@ public Ft4222I2c(I2cConnectionSettings settings)
...
-        /// <param name="disposing"></param>
+        /// <inheritdoc/>
         protected override void Dispose(bool disposing)
         {
-            if (_ftHandle != IntPtr.Zero)
-            {
-                FtFunction.FT4222_UnInitialize(_ftHandle);
-                FtFunction.FT_Close(DeviceInformation.FtHandle);
-            }
+            if (!_ftHandle.IsClosed)
```
krwq: I think if you just call Dispose this exact logic will happen internally:
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/SafeHandle.cs#L174

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -13,12 +15,9 @@ namespace System.Device.Spi
     public class Ft4222Spi : SpiDevice
     {
         private readonly SpiConnectionSettings _settings;
-        private IntPtr _ftHandle = new IntPtr();
+        private SafeFtHandle _ftHandle = new SafeFtHandle();
```
krwq: ditto

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -35,14 +34,20 @@ public Ft4222Spi(SpiConnectionSettings settings)
             _settings = settings;
             // Check device
             var devInfos = FtCommon.GetDevices();
+            if (devInfos.Count == 0)
+                throw new IOException("No FTDI device available");
+
             // Select the one from bus Id
-            DeviceInformation = devInfos[_settings.BusId];
+            var devInfo = devInfos.Where(m => m.SerialNumber == "A" || string.IsNullOrEmpty(m.SerialNumber)).Where(m => !string.IsNullOrEmpty(m.Description)).ToArray();
```
krwq: Can you add comment on `"A"`?

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -129,85 +134,53 @@ public Ft4222Spi(SpiConnectionSettings settings)
...
-        /// The length of the buffer determines how much data to read from the SPI device.
-        /// </param>
+        /// <inheritdoc/>
         public override void Read(Span<byte> buffer)
         {
-            byte[] readBuff = new byte[buffer.Length];
-            ushort readBytes = 0;
-            var ftStatus = FtFunction.FT4222_SPIMaster_SingleRead(_ftHandle, readBuff, (ushort)buffer.Length, ref readBytes, true);
+            ushort readBytes;
+            var ftStatus = FtFunction.FT4222_SPIMaster_SingleRead(_ftHandle, out MemoryMarshal.GetReference(buffer), (ushort)buffer.Length, out readBytes, true);
```
krwq: buffer should be in/ref, I suspect out might work by accident too though

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -129,85 +134,53 @@ public Ft4222Spi(SpiConnectionSettings settings)
...
-        /// <param name="writeBuffer">The buffer that contains the data to be written to the SPI device.</param>
-        /// <param name="readBuffer">The buffer to read the data from the SPI device.</param>
+        /// <inheritdoc/>
         public override void TransferFullDuplex(ReadOnlySpan<byte> writeBuffer, Span<byte> readBuffer)
         {
-            byte[] readBuff = new byte[readBuffer.Length];
-            ushort readBytes = 0;
-            var ftStatus = FtFunction.FT4222_SPIMaster_SingleReadWrite(_ftHandle, readBuff, writeBuffer.ToArray(), (ushort)writeBuffer.Length, ref readBytes, true);
+            ushort readBytes;
+            var ftStatus = FtFunction.FT4222_SPIMaster_SingleReadWrite(_ftHandle, out MemoryMarshal.GetReference(readBuffer), in MemoryMarshal.GetReference(writeBuffer), (ushort)writeBuffer.Length, out readBytes, true);
```
krwq: both buffers should be `in` I think

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -129,85 +134,53 @@ public Ft4222Spi(SpiConnectionSettings settings)
...
-        /// <param name="buffer">
-        /// The buffer that contains the data to be written to the SPI device.
-        /// </param>
+        /// <inheritdoc/>
         public override void Write(ReadOnlySpan<byte> buffer)
         {
-            ushort writeBytes = 0;
-            var ftStatus = FtFunction.FT4222_SPIMaster_SingleWrite(_ftHandle, buffer.ToArray(), (ushort)buffer.Length, ref writeBytes, true);
+            ushort writeBytes;
+            var ftStatus = FtFunction.FT4222_SPIMaster_SingleWrite(_ftHandle, in MemoryMarshal.GetReference(buffer), (ushort)buffer.Length, out writeBytes, true);
```
krwq: use name: `bytesWritten` which we use everywhere in the framework (if needed also modify the interop with similar conventions)

src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -129,85 +134,53 @@ public Ft4222Spi(SpiConnectionSettings settings)
...
-        /// <param name="disposing"></param>
+        /// <inheritdoc/>
         protected override void Dispose(bool disposing)
         {
-            if (_ftHandle != IntPtr.Zero)
-            {
-                FtFunction.FT4222_UnInitialize(_ftHandle);
-                FtFunction.FT_Close(DeviceInformation.FtHandle);
-            }
+            if (!_ftHandle.IsClosed)
```
krwq: I think just Dispose should work

src/devices/Ft4222/FtCommon.cs
```diff
@@ -24,23 +25,33 @@ public static List<DeviceInformation> GetDevices()
...
 
             Debug.WriteLine($"Number of devices: {numOfDevices}");
             if (numOfDevices == 0)
                 throw new IOException($"No device found");
 
-            byte[] sernum = new byte[16];
-            byte[] desc = new byte[64];
-            uint flags = 0;
-            for (int i = 0; i < numOfDevices; i++)
+            for (uint i = 0; i < numOfDevices; i++)
```
krwq: unless there is any reason for this being `uint` I'd move this back to `int`

src/devices/Ft4222/FtCommon.cs
```diff
@@ -24,23 +25,33 @@ public static List<DeviceInformation> GetDevices()
...
+                Span<byte> sernum = stackalloc byte[16];
+                Span<byte> desc = stackalloc byte[64];
+                uint flags = 0;
                 var devInfo = new DeviceInformation();
-                ftStatus = FtFunction.FT_GetDeviceInfoDetail(0, ref flags, ref devInfo.Type, ref devInfo.Id, ref devInfo.LocId, sernum, desc, ref devInfo.FtHandle);
+                FtDevice ftDevice;
+                uint id;
+                uint locId;
+                IntPtr handle;
+                ftStatus = FtFunction.FT_GetDeviceInfoDetail(i, out flags, out ftDevice, out id, out locId , out MemoryMarshal.GetReference(sernum), out MemoryMarshal.GetReference(desc), out handle);
```
krwq: `in MemoryMarshal.GetReference(sernum)`

src/devices/Ft4222/FtCommon.cs
```diff
@@ -24,23 +25,33 @@ public static List<DeviceInformation> GetDevices()
...
+                if (ftStatus != FtStatus.Ok)
+                    throw new IOException($"Can't read device information on device index {i}, error {ftStatus}");
+                devInfo.Type = ftDevice;
+                devInfo.Id = id;
+                devInfo.LocId = locId;
+                devInfo.FtHandle = handle;
                 devInfo.Flags = (FtFlag)flags;
-                devInfo.SerialNumber = Encoding.ASCII.GetString(sernum, 0, 16);
-                devInfo.Description = Encoding.ASCII.GetString(desc, 0, 64);
+                devInfo.SerialNumber = Encoding.ASCII.GetString(sernum.ToArray(), 0, 16);
```
krwq: @GrabYourPitchforks do you perhaps if there is any way to do that without going through array?

Also since this is returning string I'm unsure if this will take care of the null terminator correctly

src/devices/Ft4222/FtCommon.cs
```diff
@@ -84,7 +95,7 @@ public static (Version chip, Version dll) GetVersions()
 
 
             Version chip = new Version((int)((ftVersion.ChipVersion >> 24) & 0xFF), (int)((ftVersion.ChipVersion >> 16) & 0xFF), (int)((ftVersion.ChipVersion >> 8) & 0xFF), (int)(ftVersion.ChipVersion & 0xFF));
-            Version dll = new Version((int)((ftVersion.DllVersion >> 24) & 0xFF), (int)((ftVersion.DllVersion >> 16) & 0xFF), (int)((ftVersion.DllVersion >> 8) & 0xFF), (int)(ftVersion.DllVersion & 0xFF));
+            Version dll = new Version((int)((ftVersion.dllVersion >> 24) & 0xFF), (int)((ftVersion.dllVersion >> 16) & 0xFF), (int)((ftVersion.dllVersion >> 8) & 0xFF), (int)(ftVersion.dllVersion & 0xFF));
```
krwq: is dllVersion an int/uint? If so then I don't think the mask is required since you will not have any more bits on the left anyway

src/devices/Ft4222/FtFunction.cs
```diff
@@ -162,7 +162,7 @@ internal class FtFunction
         /// <param name="ssoMap">The chip select starting by 0x01</param>
         /// <returns>The status</returns>
         [DllImport("LibFT4222.dll", CallingConvention = CallingConvention.Cdecl)]
-        public static extern FtStatus FT4222_SPIMaster_Init(IntPtr ftHandle, SpiOperatingMode ioLine, SpiClock clock, SpiClockPolarity cpol, SpiClockPhase cpha, byte ssoMap);
+        public static extern FtStatus FT4222_SPIMaster_Init(SafeFtHandle ftHandle, SpiOperatingMode ioLine, SpiClock clock, SpiClockPolarity cpol, SpiClockPhase cpha, byte ssoMap);
```
krwq: so those are called cpol and cpha in the end 😄

src/devices/Ft4222/FtFunction.cs
```diff
@@ -269,7 +269,7 @@ internal class FtFunction
         /// <param name="sizeOfRead">The size of the read buffer</param>
         /// <returns>The status</returns>
         [DllImport("LibFT4222.dll", CallingConvention = CallingConvention.Cdecl)]
-        public static extern FtStatus FT4222_SPISlave_Read(IntPtr ftHandle, byte[] buffer, ushort bufferSize, ref ushort sizeOfRead);
+        public static extern FtStatus FT4222_SPISlave_Read(SafeFtHandle ftHandle, out byte buffer, ushort bufferSize, out ushort sizeOfRead);
```
krwq: I think this should be `in byte`

src/devices/Ft4222/FtFunction.cs
```diff
@@ -367,7 +367,7 @@ internal class FtFunction
         /// <param name="sizeTransferred">The size of the transfer</param>
         /// <returns>The status</returns>
         [DllImport("LibFT4222.dll", CallingConvention = CallingConvention.Cdecl)]
-        public static extern FtStatus FT4222_I2CMaster_ReadEx(IntPtr ftHandle, ushort deviceAddress, byte flag, byte[] buffer, ushort bufferSize, ref ushort sizeTransferred);
+        public static extern FtStatus FT4222_I2CMaster_ReadEx(SafeFtHandle ftHandle, ushort deviceAddress, byte flag, out byte buffer, ushort bufferSize, out ushort sizeTransferred);
```
krwq: `in byte buffer`

krwq: (unless reading a single byte)

src/devices/Ft4222/samples/Ft4222.sample.cs
```diff
@@ -45,13 +47,15 @@ static void Main(string[] args)
...
+                TestGpio();
         }
 
         private static void TestI2c()
         {
-            Ft4222I2c ftI2c = new Ft4222I2c(new I2cConnectionSettings(0, Bno055Sensor.DefaultI2cAddress));
+            var ftI2c = new Ft4222I2c(new I2cConnectionSettings(0, Bno055Sensor.DefaultI2cAddress));
 
-            Bno055Sensor bno055Sensor = new Bno055Sensor(ftI2c);
+            var bno055Sensor = new Bno055Sensor(ftI2c);
```
krwq: nice test!

src/devices/Ft4222/Ft4222.csproj
joperezr: Here we should also use Linux as the RuntimeIdentifier. This is not so the right asset gets picked at runtime, the only reason we pass in linux is in order to use the linux reference assembly when compiling this sensor, and it is important to always use the Linux version of the implementation when compiling as that one is the true reference assembly. For trying this code out locally, you can change it to win if you want or you can add a package reference instead of this which would also work, but for the version we check in please make sure you set the runtime identifier to be Linux

Ellerbach: OK, I'll change back to Linux. I'm currently working on having a proper Linux (and MacOS) version as well. so when I4ll reach this point, what should I pick to check in the code?

src/devices/Ft4222/Ft4222Gpio.cs
Ellerbach: Absolutely, correcting SPI and I2C as well.

src/devices/Ft4222/Ft4222Gpio.cs
Ellerbach: Correct, forgot to remove it from previous tests.

src/devices/Ft4222/Ft4222Gpio.cs
Ellerbach: Just checked the code of the GpioController, correct, I don't need to check the pin validity, it's done in the GpioController. I'll remove all the tests.

src/devices/Ft4222/Ft4222Spi.cs
Ellerbach: Yes, all buffers are initialized before in fact, reason why it's working. out is allowing to pass something that is not initialized. Corrected for in everywhere.

src/devices/Ft4222/FtCommon.cs
Ellerbach: The function I'm passing i is taking unit, that does avoid a cast.

src/devices/Ft4222/FtCommon.cs
Ellerbach: Correct, those are uint. Remove the & 0xFF

src/devices/Ft4222/FtFunction.cs
Ellerbach: Yes, I've corrected all of them.

src/devices/Ft4222/FtFunction.cs
Ellerbach: Corrected everywhere

src/devices/Ft4222/samples/Ft4222.sample.cs
Ellerbach: Thanks :-) This is a complicated sensor, so if working with this one, lots of chances it will work with all the others ;-)

src/devices/Ft4222/Ft4222Gpio.cs
```diff
@@ -0,0 +1,271 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Device.Ft4222;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace System.Device.Gpio
```
joperezr: Let's change the namespace and have it be Iot.Device.Ft4222.Gpio instead. We shouldn't have device bindings that are putting stuff into the System namespace.

src/devices/Ft4222/FtCommon.cs
```diff
@@ -0,0 +1,103 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace System.Device.Ft4222
```
joperezr: Please also change all of these namespaces to instead be under Iot.Device.Ft4222 like the rest of the bindings.

src/devices/Ft4222/DeviceInformation.cs
```diff
@@ -0,0 +1,42 @@
...
+        public string SerialNumber { get; set; }
+        /// <summary>
+        /// The device description
+        /// </summary>
+        public string Description { get; set; }
+        /// <summary>
+        /// The device handle.  This value is not used externally and is provided for information only.
+        /// If the device is not open, this value is 0.
+        /// </summary>
+        public IntPtr FtHandle { get; set; }
```
joperezr: This should probably be a SafeHandle instead.

src/devices/Ft4222/Ft4222.csproj
joperezr: You should always pick Linux. In the end, once we actually publish the Iot.Device.Bindings package, it will use the right one depending on where you publish for, but for development time we need always the Linux one as that one is the correct reference assembly.

src/devices/Ft4222/Ft4222Gpio.cs
joperezr: Right, so GpioController will do some validation upfront for you, that said if there is any validation that is very specific to the driver, then you would have to do that yourself.

src/devices/Ft4222/DeviceInformation.cs
joperezr: Also, who is opening the handle? If this is the class in charge of that, then it should be disposable in order to call Dispose on the SafeHandle.

src/devices/Ft4222/Ft4222Gpio.cs
krwq: @joperezr I'd avoid having device name in the namespace - that is usually causing that it's very annoying to use because you need to write name explicitly. What about `Iot.Device.Gpio` or `Iot.Device.Usb`?

src/devices/Ft4222/FtCommon.cs
krwq: I agree we should unify the namespaces

src/devices/Ft4222/Ft4222Gpio.cs
joperezr: I agree it would be a bit better if we had a different namespace naming scheme, but this is what we already do on all of the existing bindings, so at least for now I would aim for consistency. We can log an issue and discuss about this and see if we want to take in the breaking change and re-organize into different namespaces aside from this PR.

src/devices/Ft4222/DeviceInformation.cs
Ellerbach: This handle is provided for information and is not used. It is part of the information that you get about a specific sub device.
I will remove it from the class as it's not needed

src/devices/Ft4222/Ft4222Gpio.cs
Ellerbach: So which namespace should I use then?

src/devices/Ft4222/FtCommon.cs
Ellerbach: OK, so I changed everything for Iot.Device.Ft4222

src/devices/Ft4222/Ft4222Spi.cs
krwq: `24000000 / 256` shows 93750 for me which seems to be exactly half of the above number

krwq: actually that's correct since it's `<`

src/devices/Ft4222/FtDevice.cs
```diff
@@ -0,0 +1,70 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
```
krwq: nit: extra enter

src/devices/Ft4222/FtFunction.cs
```diff
@@ -0,0 +1,562 @@
...
+        /// <param name="ftHandle">The handle of the open device</param>
+        /// <param name="buffer">The output read buffer</param>
+        /// <param name="bufferSize">The size of the buffer</param>
+        /// <param name="sizeTransferred">The size of the transfer</param>
+        /// <returns>The status</returns>
+        [DllImport("libft4222", CallingConvention = CallingConvention.Cdecl)]
+        public static extern FtStatus FT4222_I2CSlave_Read(SafeFtHandle ftHandle, in byte buffer, ushort bufferSize, out ushort sizeTransferred);
+
+        /// <summary>
+        ///
```
krwq: description missing

krwq:
```suggestion
        /// Reads data from the I2C slave
```

src/devices/Ft4222/FtDevice.cs
krwq:
```suggestion
```

PR #890, opened 2019-12-13
fix scaling of raw values from ADS1115

### LINKED ISSUE #889: Scaling of raw values from ADS1115 is incorrect
**Description**

The Method `Ads1115.RawToVoltage` scales input values incorrectly, when one of the differential input multiplexers (e.g. `AIN0_AIN1`) is selected.
The raw values should always be scaled by 32768, but are scaled by 65535 in these cases instead.

The specification for the device can be found [here](https://www.ti.com/lit/ds/symlink/ads1115.pdf).
The relevant section is 9.5.4:

2019-12-13 12_08_05-Ultra-small, Low-Power, I2C, 16-Bit ADC With Int Ref, PGA, and Prog Comparator d

**Steps to reproduce**

```csharp
I2cDevice i2cDevice = ...;
var device = new Ads1115(i2cDevice, InputMultiplexer.AIN0_AIN1, MeasuringRange.FS4096);
var rawValue = (short) 0x7FFF;
var voltage = device.RawToVoltage(rawValue);
```

**Expected behavior**

`voltage` should be 4.096 (the maximum possible value).

**Actual behavior**

`voltage` is 2.048.

**Versions used**

Add following information:

```
 Version:   3.1.100
 Commit:    cd82f021f4

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.1.100\

Host (useful for support):
  Version: 3.1.0
  Commit:  65f04fb6db

.NET Core SDKs installed:
  3.1.100 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.14 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.14 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.0 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.14 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.0 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

- Version of `System.Device.Gpio` package: 1.0.0
- Version of `Iot.Device.Bindings` package (not needed if bug is in `System.Device.Gpio`): 1.0.0

TobBrandt:
Fixes #889

### DIFF
--- a/src/devices/Ads1115/Ads1115.cs
+++ b/src/devices/Ads1115/Ads1115.cs
@@ -151,14 +151,7 @@ namespace Iot.Device.Ads1115
                     break;
             }
 
-            if ((byte)_inputMultiplexer <= 0x03)
-            {
-                resolution = 65535.0;
-            }
-            else
-            {
-                resolution = 32768.0;
-            }
+            resolution = 32768.0;
 
             return val * (voltage / resolution);
         }


krwq:
cc: @ZhangGaoxing

src/devices/Ads1115/Ads1115.cs
```diff
@@ -151,14 +151,7 @@ public double RawToVoltage(short val)
...
 
-            if ((byte)_inputMultiplexer <= 0x03)
-            {
-                resolution = 65535.0;
-            }
-            else
-            {
-                resolution = 32768.0;
-            }
+            resolution = 32768.0;
```
krwq: @TobBrandt there is also a note under that paragraph in the spec:

```
Single-endedsignalmeasurements,whereVAINN= 0 V and VAINP= 0 V to +FS,only usethe positivecoderangefrom0000hto 7FFFh.However,becauseof deviceoffset,theADS111xcan still outputnegativecodesin caseVAINPis closeto 0 V.
```

Should you be doing full range for single-ended readings? I'm wondering if the `(byte)_inputMultiplexer <= 0x03` wasn't the check for that but I'm not sure

TobBrandt: The way I read this note is as follows: for single ended inputs, the input is expected to be positive relative to ground. Therefore the raw value will only use the positive part of the range (0 to MAX_SHORT), but you can occasionally get small negative values due to measurement tolerances.
That still means you need to divide by MAX_SHORT for single ended inputs.

Btw, the single ended inputs have values  > 0x03 for the input multiplexer. The check is for differential inputs.

I’ll be doing a few measurements when I’m back at the office and post them here.

pgrawehr: Your change affects the differential inputs, which now divide by MAX_SHORT instead of MAX_USHORT. I haven't used the differential mode, but I'm quite sure the single-ended mode works as expected. And yes, it often does give a small negative number if you connect an input to ground, which indicates the sign handling is correct, otherwise a small negative (integer) number would be interpreted as something big.

TobBrandt: The original code worked correctly for the single ended mode. Just the differential mode is wrong by a factor of two.

I've made a few measurements with A0 = 3.3V, A1 = 5V, and gain = +-6.144V. Here are the result:

|InputMultiplexer|Raw value|Calculated Voltage|Actual Voltage|
| ---------------- | ----------- | ------- | ------------- |
|AIN0 = 0x04|0x4461|3.2821875|3.3|
|AIN1 = 0x05|0x6826|4.999125|5|
|AIN0_AIN1 = 0x00|0xDC76|-0.852951|-1.7|

pgrawehr: Seems to be you're right, then. AINO_AIN1 should calculate A0-A1 (If I read the doc correctly), which in your case is -1.7.

krwq: So if I understand correctly you've made measurements as above after the fix and the actual voltage looks correct? If so I'm cool with merging this as is.

TobBrandt: Yes, I did.

krwq:
@ZhangGaoxing merging this since there is no objection, please create an issue if you review this after being merged and notice anything

PR #892, opened 2019-12-13
Enable basic StyleCop rules

pgrawehr:
This is a first instance of adding StyleCop support to the project, enabling most rules I consider useful (and that match the existing project style). See #816.

The changes should not affect the public API, except in very few cases (I think it was one or two places), where protected members where named incorrectly. 

The list of rules can be found int eng/StyleCop.Analyzers.ruleset. All the rules can be found here: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/tree/master/documentation (note that some rules contradict each other and some are IMHO questionable). 

Since this touches about every file in the project, conflicts with other PRs are unavoidable. (I'm just hoping I don't get the blame for all the bugs now, as my name sits on every method in the project...)


pgrawehr:
Builds are already failing due to new files that have been added...

eng/StyleCop.Analyzers.ruleset
```diff
@@ -0,0 +1,182 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="SyleCop Rules" ToolsVersion="15.0">
+  <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
+    <Rule Id="SA0000" Action="Error" />
+	<Rule Id="SA0001" Action="None" /> <!-- XML comment analysis is disabled (default for sample projects) -->
```
krwq: indent off

krwq:
@pgrawehr this is awesome but a bit hard to review. Any chance we can split this PR a bit to make it easier to review? I.e. split your branch into couple of branches:
- changes to build (leave this PR as the last one to merge)
- split diff into couple of files per PR (so that it's max 1-2k lines diff per PR)

presumably we should be able to take style changes before we enforce the style rules globally.

pgrawehr:
@krwq I can absolutelly see your point and was thinking about this as well. If you follow the single commits, you'll see the build changes already separated into their own commits. Looking at it, I should probably make this even more readable.

src/System.Device.Gpio/Interop/Unix/Libc/Interop.access.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+// <autogenerated />
```
krwq: this file is not auto-generated, where is this coming from?

pgrawehr:
So, now this is only for the core part (System.Device.Gpio) and includes the build setting changes as a separate commit. To enable it for the rest, only a few lines have to be changed again.

src/System.Device.Gpio/Interop/Unix/Libc/Interop.epoll_ctl.cs
```diff
@@ -2,7 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+// Files with Interop names are excluded from StyleCop by this line:
```
krwq: nah, this is used for auto-generated files and it's side effect is disabling StyleCop can't you `#pragma disable SAXYZ` instead?

pgrawehr: I didn't know #pragma warning disable works for StyleCop rules. I only knew the SuppressMessage approach so far, but for these files that results in ugly code cluttering, so the autogenerated-approach was a workaround for this (see https://github.com/dotnet/iot/pull/892/commits/f0037bfdbfd243c5e36978e7d6a4244f8ae67193). 

Much better and cleaner now.

krwq:
Thank you @pgrawehr!

joperezr:
> Microsoft Visual Studio Solution File, Format Version 12.00

What was the pourpose to create a solution with everything? I don't think this is a good idea and won't scale very well, so if possible I would like to get it removed.

---
Refers to: AllProjects.sln:2 in 719b0dc. [](commit_id = 719b0dc95d025a166bc252d3fbe56dc4e0a2bf1c, deletion_comment = False)

eng/Analyzers.props
```diff
@@ -0,0 +1,16 @@
...
+<!-- StyleCop.Analyzers configuration-->
+<!-- If the runtime is unknown, GeneratePlatformNotSupportedAssembly is set in 
+System.Device.Gpio.csproj, which creates an auto-generated code 
+file which is not marked as such, making StyleCop fail. -->
+  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.csproj' AND '$(GeneratePlatformNotSupportedAssembly)'!='true' AND '$(MSBuildProjectFile)'!='Iot.Device.Bindings.csproj'">
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)\StyleCop.Analyzers.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
+  <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.csproj' AND '$(GeneratePlatformNotSupportedAssembly)'!='true' AND '$(MSBuildProjectFile)'!='Iot.Device.Bindings.csproj'">
+    <AdditionalFiles Include="$(MSBuildThisFileDirectory)\stylecop.json" Link="stylecop.json" />
+	<PackageReference Include="StyleCop.Analyzers" Version="1.1.118">
```
joperezr: NIT: this line has a tab, please change it to spaces

eng/Analyzers.props
```diff
@@ -0,0 +1,16 @@
+<Project>
+<!-- StyleCop.Analyzers configuration-->
+<!-- If the runtime is unknown, GeneratePlatformNotSupportedAssembly is set in 
+System.Device.Gpio.csproj, which creates an auto-generated code 
+file which is not marked as such, making StyleCop fail. -->
+  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.csproj' AND '$(GeneratePlatformNotSupportedAssembly)'!='true' AND '$(MSBuildProjectFile)'!='Iot.Device.Bindings.csproj'">
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)\StyleCop.Analyzers.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
+  <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.csproj' AND '$(GeneratePlatformNotSupportedAssembly)'!='true' AND '$(MSBuildProjectFile)'!='Iot.Device.Bindings.csproj'">
```
joperezr: If this item is specific to Iot.Device.Bindings, then I would suggest to put it only on that project not on a general props file. Also, if it is applicable to that project, we will want to instead also make it applicable to all projects under src/devices so that developers working on individual projects will also get this feedback during development.

eng/StyleCop.Analyzers.ruleset
```diff
@@ -0,0 +1,182 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="SyleCop Rules" ToolsVersion="15.0">
```
joperezr: Where did we get this list from?

eng/stylecop.json
```diff
@@ -0,0 +1,31 @@
...
+            "indentationSize": 4,
+            "tabSize": 4,
+            "useTabs": false
+        },
+        "orderingRules": {
+            "usingDirectivesPlacement": "outsideNamespace",
+            "blankLinesBetweenUsingGroups": "omit"
+        },
+        "namingRules": {
+            "allowCommonHungarianPrefixes": true,
```
joperezr: Can you explain why is this and the next one needed? Why Hungarian in particular?

src/devices/Directory.Build.props
```diff
@@ -1,6 +1,8 @@
 <Project>
   <Import Project="../../eng/Compilers.props" />
   <Import Project="../../eng/Versions.props" />
+  <!-- Disabled until all projects are merged -->
```
joperezr: NIT: it would be good if we can log an issue abou tthis and have that added to this comment to be able to track progress.

eng/StyleCop.Analyzers.ruleset
krwq: I've pointed @pgrawehr to our corefx StyleCop ruleset so assumed that's where it's from

src/devices/Directory.Build.props
krwq: We have https://github.com/dotnet/iot/issues/816 I'll add a check box there

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -30,8 +27,12 @@ public unsafe partial class RaspberryPi3Driver : GpioDriver
         private const string DeviceTreeRanges = "/proc/device-tree/soc/ranges";
         private const string ModelFilePath = "/proc/device-tree/model";
 
-        private UnixDriver _sysFSDriver = null;
         private readonly IDictionary<int, PinMode> _sysFSModes = new Dictionary<int, PinMode>();
```
joperezr: Just wondering, is it a rule that constants need to be on their own and that they have to be first?

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10DriverPin.Windows.cs
```diff
@@ -42,6 +42,7 @@ public void Dispose()
                 _pin.Dispose();
                 _pin = null;
             }
+
```
joperezr: Looks like we added some tabs in here by mistake. Please remove when you have a chance. Also, is there a stylecop rule that enforces spaces instead of tabs? If so we should enable that.

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
```diff
@@ -71,6 +71,7 @@ public override int Frequency
                 {
                     throw new ArgumentOutOfRangeException(nameof(value), value, "Value must not be negative.");
                 }
+
```
joperezr: Another line with tabs.

src/System.Device.Gpio/System/Device/Pwm/Channels/Windows10PwmChannel.Windows.cs
```diff
@@ -86,6 +87,7 @@ public override double DutyCycle
                 {
                     throw new ArgumentOutOfRangeException(nameof(value), value, "Value must be between 0.0 and 1.0.");
                 }
+
```
joperezr: Here too.

src/System.Device.Gpio/System/Device/Spi/Devices/Windows10SpiDevice.Windows.cs
```diff
@@ -114,6 +116,7 @@ public override void TransferFullDuplex(ReadOnlySpan<byte> writeBuffer, Span<byt
             {
                 throw new ArgumentException($"Parameters '{nameof(writeBuffer)}' and '{nameof(readBuffer)}' must have the same length.");
             }
+
```
joperezr: One more line with tabs

joperezr:
@pgrawehr Left a few comments, hopefully you could address those in your now opened PR. Most are minor things, it is great you added this! Thank you so much for doing it 😄

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: I think it's this order (second rule is whatever is most public goes first):
- constants
- fields
- constructors
- methods
- nested types

src/devices/Directory.Build.props
joperezr: If we could just add the pointer to that to this comment that would be awesome

eng/Analyzers.props
pgrawehr: The `AND '$(MSBuildProjectFile)'!='Iot.Device.Bindings.csproj'` part is temporary only. Without this, the build would fail for now, as that csproj file directly imports (root)/Directory.Build.props, which has the rules already enabled. This will be removed once all parts are merged.

joperezr: Yup I missed that, this sounds good to me.

pgrawehr:
> 
> 
> > Microsoft Visual Studio Solution File, Format Version 12.00
> 
> What was the pourpose to create a solution with everything? I don't think this is a good idea and won't scale very well, so if possible I would like to get it removed.
> 
> Refers to: AllProjects.sln:2 in [719b0dc](https://github.com/dotnet/iot/commit/719b0dc95d025a166bc252d3fbe56dc4e0a2bf1c). [](commit_id = [719b0dc](https://github.com/dotnet/iot/commit/719b0dc95d025a166bc252d3fbe56dc4e0a2bf1c), deletion_comment = False)

For this task here, it is very helpful, as I can edit rules and then immediately build everything and see what breaks. That's the simplest way of building all the projects (including the samples) from within VS. And since the sln file doesn't contain any logic (i.e. no dependency statements), it's not hard to maintain.

krwq:
@pgrawehr @joperezr about sln: I personally don't mind either way. We can also auto-regen sln (dotnet cli has utilties for sln) in the build if there is a need in the future or remove it once the StyleCop prs are done

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10DriverPin.Windows.cs
pgrawehr: YesandNo. There's a StyleCop rule that checks that the file uses only one type of intendation and a setting that says which one. But if you have entire files that are wrong, no error will happen. Not sure why this slipped, though. I'll go over everything at the very end again, ok?

krwq: I'll add checkbox about tabs in the issue and sounds good to me

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: Yes, that's the rule that's enabled now, except that it is 
- constants
- readonly fields
- read-write fields
- constructors
- properties
- nested types (classes)
There are other similar rules in the "ordering" category https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/OrderingRules.md, but I think this one is the most useful rule (and there where actually few violations of it). 

The other rule, SA1202, ElementsMustBeOrderedByAccess is disabled for now (and I personally think it is not helping anything, and requires moving stuff around in files if the visibility changes)

krwq: I thought the properties go before constructors, I got impression everyone does that at least 😄

krwq: SA1202 let's consider afterwards, more important to get bulk ASAP to not affect pending PRs too much

joperezr:
Given that the pourpose of the solution is was just to be able to load all projectsfor this task in mind, let's just remove it since it may just generate confusion on what is the right way of opneing a project. Further more, I think that now if somebody tries to open just a single project, VS will automatically open this solution as it will find it in the path and realize that it contains the project that the user wanted to open. For this and other reasons, let's just get rid of it and only add it in the future if we think it is a scenario that our contributors would benefit of.

joperezr:
For upcomming stylecop PRs, Given that System.Device.Gpio is done, I would recommend just using Iot.Device.Bindings.csproj, or creating a temporary solution with the desired projects without checking this in.

eng/StyleCop.Analyzers.ruleset
pgrawehr: I actually started with the rule set I'm used to and then compared to the corefx set. I added special comments where the corefx has a rule enabled but here it's not (yet).

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10DriverPin.Windows.cs
pgrawehr: Found that I missed SA1027, which is exactly for this.

eng/stylecop.json
pgrawehr: I took this from a template and forgot to check. These two sections are unused here.

eng/StyleCop.Analyzers.ruleset
krwq: fine as a first step, we should eventually enable at minimum the same as corefx and more as a stretch. Let's focus on enforcing this first and once we got some rules running we can iterate from there

krwq:
@pgrawehr I think theoretically you can `dotnet sln add **/*.csproj project.sln` since **/*.csproj will be expanded by shell it would call it (correctly) multiple times and I'm guesing it won't add any duplicates as well

eng/StyleCop.Analyzers.ruleset
pgrawehr: At the end, I'll again go trough everything and document which rules are enabled there and not here. Those won't be many.

PR #894, opened 2019-12-15
Do StyleCopify for Ft4222.

pgrawehr:
Part 2 for #816 (only a small group this time)

### DIFF
--- a/src/devices/Ft4222/Ft4222Gpio.cs
+++ b/src/devices/Ft4222/Ft4222Gpio.cs
@@ -35,13 +35,15 @@ namespace Iot.Device.Ft4222
         /// <summary>
         /// Create a FT4222 GPIO driver
         /// </summary>
-        /// <param name="deviceNumber"></param>
+        /// <param name="deviceNumber">Number of the device in the device list, default 0</param>
         public Ft4222Gpio(int deviceNumber = 0)
         {
             // Check device
             var devInfos = FtCommon.GetDevices();
             if (devInfos.Count == 0)
+            {
                 throw new IOException("No FTDI device available");
+            }
 
             // Select the deviceNumber, only the last one in Mode 0 and Mode 1 can be open.
             // The last one is either B if in Mode 0 or D in mode 1.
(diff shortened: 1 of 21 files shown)


src/devices/Ft4222/Ft4222Gpio.cs
```diff
@@ -35,49 +35,64 @@ public class Ft4222Gpio : GpioDriver
         /// <summary>
         /// Create a FT4222 GPIO driver
         /// </summary>
-        /// <param name="deviceNumber"></param>
+        /// <param name="deviceNumber">Number of the device in the device list, default 0</param>
```
krwq:
```suggestion
        /// <param name="deviceNumber">Number of the devices in the device list, default 0</param>
```

pgrawehr: I don't think so, it's an index, not a count.

src/devices/Ft4222/FtVersion.cs
```diff
@@ -13,9 +13,10 @@ public struct FtVersion
         /// Chip version
         /// </summary>
         public uint ChipVersion;
+
         /// <summary>
         /// Dll version
         /// </summary>
-        public uint dllVersion;
+        public uint DllVersion;
```
joperezr: I agree with this change but do keep in mind that these are breaking changes.

joperezr:
PR looks good to me thanks @pgrawehr. Do you mind sharing what is the workflow you have for fixing the stylecop issues so that other people may help you with the effort? Is it simply to enable the rules and then fix in VS one by one or is there a tool you call to apply codefixes?

krwq:
@joperezr original version of the PR was very large and I think it included all or almost all changes required. I've asked @pgrawehr to split PR a bit so that it's easier to add and address any feedback. @pgrawehr do you have any areas you haven't started yet? Could be useful to create checkboxes in the issues with what's left (and is not pending) if there is anything

pgrawehr:
@krwq: There are some rules which I left open for now (and IMHO needs discussion whether they really should be enabled for this project). The rules that are currently configured are all solved and pass for the whole repository on my branch. Just putting that into reviewable junks now, as you suggested. At the end, we can enable the rules for all projects and write checkboxes for the next steps. What I haven't addressed yet are FxCop rules, but that should be done separately, as those rules are more likely to require functional changes. 

I think it's best if currently nobody helps with this, as it would only create significant merge issues. 

Some of these problems can be auto-solved in VS by hovering over the marked text and then check the solutions:
grafik
(Partly german here, but I think you get the point).

PR #895, opened 2019-12-16
StyleCop fixes, Adx to BoardLed

pgrawehr:
Another bunch of files (Ads1115 skipped, because of open PR, will do at the end)


src/devices/Ags01db/samples/Program.cs
```diff
@@ -4,9 +4,16 @@
 
 namespace Iot.Device.Ags01db.Samples
 {
-    class Program
+    /// <summary>
+    /// Test program main class
```
krwq: I'm wondering if we shouldn't rather make it explicit `internal` (presumably that's what the style rule is complaining about no explicit visibility) - this comment is not exactly useful and with `internal` it should not be required

src/devices/Bmp180/Bmp180.cs
```diff
@@ -69,19 +68,19 @@ public Temperature ReadTemperature()
...
-            int X1 = ((short)_calibrationData.AC3 * B6 ) / 8192;
-            int X2 = ((short)_calibrationData.B1 * B62) / 65536;
-            X3 = ((X1 + X2) + 2) / 4;
-            int B4 = _calibrationData.AC4 * (X3 + 32768) / 32768;
-            uint B7 = (uint)(ReadRawPressure() - B3) * (uint)(50000 >> (short)Sampling.Standard);
-            int p = (B7 < 0x80000000) ? (int)((B7 * 2) / B4) : (int)((B7 / B4) * 2);
-            X1 = (((p * p) / 65536 ) * 3038) / 65536;
-            
-            return Pressure.FromPascal(p + ( ((((p * p) / 65536 ) * 3038) / 65536) + ((-7357 * p) / 65536) + 3791) / 8);
+            int b6 = CalculateTrueTemperature() - 4000;
```
krwq: This will create a merge conflict with https://github.com/dotnet/iot/pull/783 I'm wondering if it won't be better to just temporarily disable all the style checks in this file and create an issue about this to avoid extra work in the other PR. cc: @MarkCiliaVincenti

joperezr:
@pgrawehr sorry I couldn't add my feeback to #892 but I'm just adding it now, can you please include in this PR all the comments I make on that one?

krwq:
@joperezr sorry I merged too quickly :smile:

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -49,15 +49,19 @@ public enum CommunicationProtocol
...
+        {
+            get;
+            set;
+        }
 
         /// <summary>
         /// The temperature calibration factor.
         /// </summary>
-        protected virtual int _tempCalibrationFactor => 1;
+        protected virtual int TempCalibrationFactor => 1;
```
joperezr: This is source breaking change, but I doubt there are a lot of people subclassing outside of the repo so I'm fine with this fix.

krwq: I'd still go with this fix since it's very unlikely to break anyone and if it is well... they should have contributed it back :smiling_imp:

src/devices/Ags01db/samples/Program.cs
pgrawehr: Originally, I was a bit confused on this, because I thought that the main class must be public, but Internal is apparently enough. With that, no summary would be required (yes, the rule is complaining about a public item not being documented)

src/devices/Bmxx80/Bmxx80Base.cs
pgrawehr: I saw that and thought the same. If I remember correctly, there where only like 1 or 2 occassions of fixes like this I had to do.

pgrawehr:
Just checking what's up with these tabs.

src/devices/Bmxx80/Bmxx80Base.cs
joperezr: right, and I believe it is the right thing to fix them (at least the ones that I've spotted) so it's fine to take these breaking changes.

pgrawehr:
@joperezr I missed that rule SA1027 was not enabled, which exactly is for forcing tab/spacing consistency. Just fixing that globaly now, and will commit with the next PRs. I'll continue to create PRs with subsets of files until the remainder is small enough for the rules to be enabled globally.

joperezr:
Awesome! thanks again. If you could also apply the rest of the feedback I gave on the other PR here that would be great as well 😄(like removing the AllProject.sln and such)

eng/StyleCop.Analyzers.ruleset
```diff
@@ -69,6 +69,7 @@
     <Rule Id="SA1136" Action="Error" />
     <Rule Id="SA1137" Action="Error" />
     <Rule Id="SA1139" Action="Error" /> 
+	<Rule Id="SA1167" Action="Error" /> <!-- Elements must have same indentation -->
```
krwq: this element (ironically) has different indentation 😄

pgrawehr: We need stylecop rules for the rule sets...

pgrawehr:
@joperezr I'd like to find an alternate way for the sln problem separately

joperezr:
> I'd like to find an alternate way for the sln problem separately

What about keeping the solution as you have it except just move it one directory up so you have it outside of the iot git repo? That way you will still be able to use it for the StyleCop work but it won't leak into the repository and won't confuse people with a solution that 99.99% of our users won't want/need to open. Would that work for you?

pgrawehr:
Not so good, actually. I need it on different PCs (this StyleCop stuff is something I can do also from underway, when I don't have access to the Pi and the sensors). 
`dotnet new` + `dotnet sln add` can actually build the sln from scratch, so I can create a small batch file to autogenerate the solution when it's needed. But not today any more.

joperezr:
> Not so good, actually. I need it on different PCs

I see, what about working on a special branch that has the solution and creating PRs from there but not merging the solution change? Sorry for being persistent on not wanting the allup solution but I really want to avoid it since I know it will create a bunch of problems and confusion in the future. Also, I'm worried it could have unintended consequences with our build infrastructure, causing it to double restore some projects with wrong configurations. We have a lot of custom build targets on the repo, so passing in the wrong global properties down might actually affect the build.

> so I can create a small batch file to autogenerate the solution when it's needed.

This would be great as long as you keep that batch file code in a gist or something and not checked in for the same reason. If we won't have the solution checked in, then a batch file which creates one might only create confusion. But having that batch might actually solve your problem of working on StyleCop across different machines, so that might be something to explore.

My ultimate point in here is that having an all-inclusive solution will not be used by anybody except of StyleCop editing, and having it will cause confusion, problems when developing, and might even have unintended consequences with our build infrastructure, so that is why I'm pushing on getting it removed.

pgrawehr:
> 
> 
> > Not so good, actually. I need it on different PCs
> 
> Also, I'm worried it could have unintended consequences with our build infrastructure, causing it to double restore some projects with wrong configurations. We have a lot of custom build targets on the repo, so passing in the wrong global properties down might actually affect the build.

I see the point about the confusion, but there shouldn't be any problems with the infrastructure, as the solution itself does not add any properties at all (not even dependencies). That's all part of the projects. If there's problems with properties, we have problems with the different device projects (for instance, I just found that building windows-debug and windows-release in VS fails for quite a bunch of projects due to a wrong reference path, not yet investigated why). 

> 
> > so I can create a small batch file to autogenerate the solution when it's needed.
> 
> This would be great as long as you keep that batch file code in a gist or something and not checked in for the same reason. If we won't have the solution checked in, then a batch file which creates one might only create confusion. But having that batch might actually solve your problem of working on StyleCop across different machines, so that might be something to explore.
Never used that, but I'll try. 
> 
> My ultimate point in here is that having an all-inclusive solution will not be used by anybody except of StyleCop editing, and having it will cause confusion, problems when developing, and might even have unintended consequences with our build infrastructure, so that is why I'm pushing on getting it removed.

I personally think that _not_ having a global solution confuses people, but I might be wrong here. Could also be the fact that we don't have documentation on building at all, it seems...

joperezr:
>  but there shouldn't be any problems with the infrastructure, as the solution itself does not add any properties at all

We don't use the regular infrastructure to build .csprojs, we use the one located in dotnet/arcade, and for a bunch of targets (for example some of our restore targets) we do a global grep for all projects and solutions in the repo, and we restore them. I get your solution itself is not setting any properties, but it is selecting configurations for the projects, and we actually have some custom built in logic for configurations in this repo since we do cross-RID targetting which is not supported by the default msbuild sdk. I helped built both the infrastructure we built with (dotnet/arcade) and the custom stuff in this repo, so that is how I know that there is some potential of adding a global solution that could break or cause double restore or other unwanted consequences. I did some initial validation and it does look like for now we are getting lucky and it has had no effect, but we update the infrastructure we build upon daily so this may very well break any time. Given the only benefit of adding a global solution is the fact of making the progress on StyleCop easier, I don't think it's worth keeping this solution as this is something we can work out in a different way easily (like the suggestions I've made above). 

> I personally think that not having a global solution confuses people

We have analyzed this several times in the past and concluded that this is not true. Having a global solution simply won't scale, and there is no user scenario where you would want to have all of the projects open in VS at the same time. I agree that we could probably provide better docs for building the repo or to make them more discoverable, but if you have specific feedback in mind with [our current instructions](https://github.com/dotnet/iot/blob/master/Documentation/CONTRIBUTING.md) please feel free to log issues about it so we can fix them.

pgrawehr:
@joperezr : Ok, I start to see that the build is way more complex than what I'd expected. I deleted the file now and will check how I can continue. Until the big part is merged, that won't be an issue now anyway - only maybe at the very end when some things don't quite work fully when the rules go live.

joperezr:
Thank you very much, I really appreciate your effort here and I really think that it will definitely help us maintain a much better code base and ensure we follow all the desired guidelines.

src/devices/Bmp180/Bmp180.cs
pgrawehr: When there are conflicts, its probably easiest just to use the "own" part for now, and finally fix it again at the end. There's probably no good way of preventing conflicts with this task. And it seems it's still unclear when that PR can be merged.

krwq:
Thank you @pgrawehr!

krwq:
@joperezr I believe @pgrawehr has already whole change set in his local branch and is slicing it so that we can easily review it, the plan was to enable stylecop in the last PR and we will restart CI on his PR just before merging to make sure we don't accidentally break the build. There is likely not going to be much changes in the last PR - I also try to not merge anything conflicting at the moment

PR #897, opened 2019-12-17
StyleCop rules, Part 4

pgrawehr:
Ak8963 to CharacterLcd, plus a few spacing fixes


src/devices/BrickPi3/samples/SensorTests.cs
```diff
@@ -1,18 +1,18 @@
-﻿using Iot.Device.BrickPi3.Models;
-using Iot.Device.BrickPi3.Sensors;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
krwq: there should be also extra enter after the copyrightheader (not sure if stylecop has any rule for that)

src/devices/Card/CreditCard/CreditCard.cs
```diff
@@ -513,18 +552,17 @@ private bool FillTags()
...
                 }
+
                 return false;
             }
         }
 
         /// <summary>
-        /// 
+        /// Read log records
         /// </summary>
```
krwq: I'm surprised this builds correctly - I recall getting errors when params are not specified

src/devices/BrickPi3/samples/SensorTests.cs
pgrawehr: I wouldn't know. Normally it's even forbidden to have empty lines after comments, but the header seems to be an exception.

src/devices/BrickPi3/Models/SpiMessages.cs
```diff
@@ -11,78 +11,115 @@ public enum SpiMessageType : byte
...
+
         /// <summary>Set proportional factor (KP) of motor position controller</summary>
         SetMotorPositionKP,
+
         /// <summary>Set derivative factor (KD) of motor position controller</summary>
         SetMotorPositionKD,
+
         /// <summary>Set motor angular speed (degrees per second)</summary>
-        setMotorDps,
+        SetMotorDps,
```
joperezr: All of these casing changes are breaking changes. That said I think it's ok to take them as it was a mistake to not catch this in the beginning, but it is something we will probably have to document for vNext so we should just keep that in mind.

pgrawehr: Yea, guess we should just make a note in the release docs. Do we keep them (or some kind of changelog for the end user) somewhere in the repo?

src/devices/BrickPi3/Movement/Motor.cs
```diff
@@ -176,18 +195,22 @@ private void OnPropertyChanged(string name)
         /// </summary>
         public event PropertyChangedEventHandler PropertyChanged;
 
-        private int periodRefresh;
+        private int _periodRefresh;
```
joperezr: Reflection breaking change here as well, but I'm fine to take it too. This one we probably don't even have to document.

src/devices/Card/CreditCard/CreditCard.cs
pgrawehr: Right, that's possible. Maybe this depends on some project-wide warning settings as well? I sometimes get warnings about XML documentation checks being disabled (but didn't investigate on that yet)

src/devices/Buzzer/samples/Buzzer.Samples.csproj
```diff
@@ -6,8 +6,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-  	<ProjectReference Include="$(MSBuildThisFileDirectory)../Buzzer.csproj" />
-  	<ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
+      <ProjectReference Include="$(MSBuildThisFileDirectory)../Buzzer.csproj" />
```
krwq: wow, did style cop check that?

pgrawehr: Nope, I did a search and replace for tabs 😄

src/devices/BrickPi3/Models/SpiMessages.cs
krwq: @pgrawehr, no worries, @joperezr knows how to find all breaking changes between 1.0 and current version with ApiCompat tool. We can use that when writing release notes, no action on your end needed I believe

src/devices/BrickPi3/Movement/Motor.cs
krwq: Yeh, I wouldn't worry especially those are not marked as serializable

src/devices/Buzzer/samples/Buzzer.Samples.csproj
krwq: works too 😄

src/devices/BrickPi3/Models/SpiMessages.cs
joperezr: Yeah unfortunately we don't have a process established yet for breaking changes, but we can use the ApiCompat tool to find out all of these and we can just call them out on the vNext release notes.

PR #900, opened 2019-12-18
StyleCop Rules, Part 6

pgrawehr:
DCMotor to GrovePi

### DIFF
--- a/src/devices/DCMotor/DCMotor.cs
+++ b/src/devices/DCMotor/DCMotor.cs
@@ -17,11 +17,6 @@ namespace Iot.Device.DCMotor
         private const int DefaultPwmFrequency = 50;
 
         /// <summary>
-        /// <see cref="GpioController"/> related with operations on pins
-        /// </summary>
-        protected GpioController Controller;
-
-        /// <summary>
         /// Constructs generic <see cref="DCMotor"/> instance
         /// </summary>
         /// <param name="controller"><see cref="GpioController"/> related with operations on pins</param>
@@ -35,6 +30,15 @@ namespace Iot.Device.DCMotor
         /// 1 means maximum speed, 0 means no movement and -1 means movement in opposite direction.
         /// </summary>
         public abstract double Speed { get; set; }
+
+        /// <summary>
+        /// <see cref="GpioController"/> related with operations on pins
+        /// </summary>
+        protected GpioController Controller
+        {
+            get;
+            set;
+        }
 
         /// <summary>
         /// Disposes the <see cref="DCMotor"/> class
@@ -71,7 +75,9 @@ namespace Iot.Device.DCMotor
         public static DCMotor Create(PwmChannel speedControlChannel)
         {
             if (speedControlChannel == null)
+            {
                 throw new ArgumentNullException(nameof(speedControlChannel));
+            }
 
             return new DCMotor2PinNoEnable(speedControlChannel, -1, null);
         }
@@ -90,7 +96,9 @@ namespace Iot.Device.DCMotor
         public static DCMotor Create(int speedControlPin, GpioController controller = null)
         {
             if (speedControlPin == -1)
+            {
                 throw new ArgumentOutOfRangeException(nameof(speedControlPin));
+            }
 
             controller = controller ?? new GpioController();
             return new DCMotor2PinNoEnable(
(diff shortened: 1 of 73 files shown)


src/devices/DCMotor/DCMotor.cs
```diff
@@ -36,6 +31,15 @@ protected DCMotor(GpioController controller)
         /// </summary>
         public abstract double Speed { get; set; }
 
+        /// <summary>
+        /// <see cref="GpioController"/> related with operations on pins
+        /// </summary>
+        protected GpioController Controller
+        {
+            get;
```
joperezr: Why did we turn this into a property instead of a field as it used to be? This might be an unnecessary breaking change.

src/devices/GoPiGo3/Models/SpiMessages.cs
```diff
@@ -11,86 +11,127 @@ public enum SpiMessageType
...
+
         /// <summary>Set motor position controller's derivative (change) factor</summary>
         SetMotorPositionKd,
+
         /// <summary>Set motor speed (degrees per second)</summary>
         SetMotorDps,
+
         /// <summary>Set motor limits</summary>
-        setMotorLimits,
+        SetMotorLimits,
```
joperezr: One more casing breaking change. I guess since we've taken others like this it would make sense to take this one as well.

src/devices/DCMotor/DCMotor.cs
pgrawehr: I thought that normally members should not be exposed at all, and since the name was like it should be a property, I changed it to one. If you think it should rather be a protected member, I can change that, but then it will change the name.

pgrawehr: Looking at the code, it should probably even be read-only.

src/devices/GrovePi/samples/Program.cs
```diff
@@ -4,45 +4,53 @@
...
 {
-    class Program
+    /// <summary>
+    /// Test class for the GrovePi
+    /// </summary>
+    public class Program
     {
-        static private GrovePi grovePi;
-        static void Main(string[] args)
+        private static GrovePi _grovePi;
```
joperezr: It doesn't matter much here given that this is just a sample, but technically our coding guidelines say that private static fields should be named like: `s_grovePi`. I wonder if we could add that as one of the rules

src/devices/DCMotor/DCMotor.cs
joperezr: Yeah I agree that the original design is probably not the ideal one, I was only wondering if it would be worth it to take the breaking change here or if it was better to just leave as is in this case and make an exception to the rule. I will defer to @krwq as he is the one that originally added the binding.

src/devices/GrovePi/samples/Program.cs
pgrawehr: I think that only works with a custom rule, at least I can't find one for that requirement. SX1309S makes static fields require underscores, but otherwise there doesn't seem to be any rules explicitly for static members (there are some for static readonly members, but these we handle like constants now, so they use PascalCase)

joperezr: I see that is fine, we can always write a new roslyn rule in the future.

src/devices/DCMotor/DCMotor.cs
krwq: I think in general I'd say we should take all of the breaking changes in Iot.Device.Bindings
 which make the style more consistent

src/devices/DCMotor/DCMotor.cs
```diff
@@ -189,7 +211,7 @@ public static DCMotor Create(PwmChannel speedControlChannel, int directionPin, i
         /// </summary>
         /// <param name="speedControlPin">Pin used to control the speed of the motor with software PWM (frequency will default to 50Hz)</param>
         /// <param name="directionPin">Pin used to control the direction of the motor</param>
-        /// <param name="otherDirectionPin">Pin used to control the direction of the motor</param>
+        /// <param name="otherDirectionPin">Pin used to control the reverse direction of the motor</param>
```
krwq: this is actually not correct, original text was. Both pins control the direction, there is not really good description for them because they control the voltage.

src/devices/DCMotor/samples/Program.cs
```diff
@@ -10,11 +10,18 @@
...
 using Iot.Device.DCMotor;
 
-namespace samples
+namespace Iot.Device.DCMotor.Samples
 {
-    class Program
+    /// <summary>
+    /// Test program main class
+    /// </summary>
+    public class Program
```
krwq: internal and remove comments

src/devices/Dhtxx/samples/DhtSensor.sample.cs
```diff
@@ -7,23 +7,34 @@
...
 using Iot.Device.DHTxx;
 
-class Program
+namespace Iot.Device.DHTxx.Samples
 {
-    static void Main(string[] args)
+    /// <summary>
+    /// Test program main class
+    /// </summary>
+    public class Program
```
krwq: internal, remove comments

src/devices/DCMotor/DCMotor.cs
pgrawehr: Ok. The problem here was that StyleCop complains that "Documentation should not be copy-pasted", which happens when two parameters have the same description (which, in most cases, really points to an error). I'll change it to "first pin...", "Second pin..." then.

pgrawehr:
This build ran into some timeout that is probably not related to the changes.

krwq:
ok, passing after retrying

PR #901, opened 2019-12-18
StyleCop Rules, Part 7

pgrawehr:
Hcsr04 to Mcp3xxx

The naming violations in the interop files will be excluded from checks at the end


src/devices/Mcp3xxx/Mcp3xxx.cs
```diff
@@ -25,12 +20,22 @@ public abstract class Mcp3xxx : Mcp3Base
...
+            : base(spiDevice)
         {
             ChannelCount = channelCount;
             _adcResolutionBits = adcResolutionBits;
         }
 
+        /// <summary>
+        /// the number of single ended input channel on the ADC
+        /// </summary>
+        protected byte ChannelCount
```
pgrawehr: Another one of these wrongly named members.

src/devices/Ina219/Ina219.cs
```diff
@@ -241,7 +241,9 @@ public void SetCalibration(ushort calibrationValue, float currentLsb = 1.0F)
             WriteRegister(Ina219Register.Calibration, calibrationValue);
         }
 
-        /// <inheritdoc/>
+        /// <summary>
```
joperezr: Does styleCop not like `<inheritdoc />` ?

src/devices/Mcp25xxx/samples/Mcp25xxx.Sample.cs
```diff
@@ -70,14 +70,13 @@ private static void ReadAllRegistersWithDetails(Mcp25xxx mcp25xxx)
...
-            //ReadAllMessageReceiveRegistersWithDetails(mcp25xxx);
-            //ReadAllAcceptanceFilterRegistersWithDetails(mcp25xxx);
-            //ReadAllBitTimeConfigurationRegistersWithDetails(mcp25xxx);
-            //ReadAllErrorDetectionRegistersWithDetails(mcp25xxx);
-            //ReadAllInterruptRegistersWithDetails(mcp25xxx);
-            //ReadAllCanControlRegistersWithDetails(mcp25xxx);
+            ////ReadAllMessageReceiveRegistersWithDetails(mcp25xxx);
+            ////ReadAllAcceptanceFilterRegistersWithDetails(mcp25xxx);
+            ////ReadAllBitTimeConfigurationRegistersWithDetails(mcp25xxx);
+            ////ReadAllErrorDetectionRegistersWithDetails(mcp25xxx);
```
joperezr: mm not sure I understand why we got the extra double slash here

src/devices/Mcp3xxx/Mcp3xxx.cs
joperezr: Yup, I wouldn't change this for now, and just left it as is with an exception to the rule since that is probably preferable than to taking the breaking change

src/devices/Ina219/Ina219.cs
pgrawehr: The problem is that StyleCop does not like `<inheritdoc/>` when there's nothing to inherit from. This is not overloading anything.

src/devices/Mcp25xxx/samples/Mcp25xxx.Sample.cs
pgrawehr: Four slashes disable any other checks for these lines. Since it is commented out code (in an example), I think that's preferable over fixing the spacing. 
To comment out code (in stylecop-friendly way) you can either add four slashes or use two and a space. Of course, you wouldn't normally commit commented out code, except in samples.

src/devices/Ina219/Ina219.cs
joperezr: you are right! I thought this was the regular Dispose method and not the one that takes in the bool.

src/devices/Mcp25xxx/samples/Mcp25xxx.Sample.cs
joperezr: makes sense

joperezr:
My only remaining comment here would be to revert that change to convert a field into a property. Other than that this is good to go.

src/devices/Ina219/samples/Ina219.Sample.cs
```diff
@@ -12,21 +12,25 @@ namespace Iot.Device.Adc.Samples
     /// </summary>
     public class Program
     {
-        static void Main(string[] args)
+        /// <summary>
+        /// Main entry point
+        /// </summary>
+        /// <param name="args">Unused</param>
+        public static void Main(string[] args)
```
krwq: internal, remove comment

src/devices/Max7219/Fonts.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+
```
krwq: copyright header missing in this file

src/devices/Max7219/samples/Max7219.sample.cs
```diff
@@ -37,29 +36,23 @@ static void Main(string[] args)
...
-                    0b01000010,
-                    0b10100101,
-                    0b10000001,
-                    0b10100101,
-                    0b10011001,
-                    0b01000010,
-                    0b00111100
+                var smiley = new byte[]
+                {
+                    0b00111100, 0b01000010, 0b10100101, 0b10000001, 0b10100101, 0b10011001, 0b01000010, 0b00111100
```
krwq: what's this rule? I preferred the previous style to be honest

krwq: I'd rather disable it here since these form and image

pgrawehr: This is no rule. I suppose this happened due to autoformatting the file. Not sure what exact rules applied there. I'll revert it.

PR #903, opened 2019-12-18
StyleCop Rules, Part 9

pgrawehr:
Pn532 to ServorMotor


src/devices/Pn532/Pn532.cs
```diff
@@ -25,24 +25,37 @@ namespace Iot.Device.Pn532
...
+
         // Operation type for SPI
         private const byte WriteData = 0b0000_0001;
         private const byte ReadStatus = 0b0000_0010;
+
         private const byte ReadData = 0b0000_0011;
+
         // Acknowledge
-        private byte[] AckBuffer = { 0x00, 0x00, 0xFF, 0x00, 0xFF, 0x00 };
+        private byte[] _ackBuffer = { 0x00, 0x00, 0xFF, 0x00, 0xFF, 0x00 };
```
joperezr: What is the logic applied to the above new lines that were added to this file?

src/devices/Pn532/Pn532.cs
```diff
@@ -697,12 +869,8 @@ public byte[] AutoPoll(byte numberPolling, ushort periodMilliSecond, PollingType
         /// <summary>
         /// Set the PN532 as a target, so as a card
         /// </summary>
-        /// <param name="mode"></param>
-        /// <param name="mifare"></param>
-        /// <param name="feliCa"></param>
-        /// <param name="picc"></param>
-        /// <returns></returns>
-        public (TargetModeInitialized modeInialized, byte[] initiator) InitAsTarget(TargetModeInitialization mode, TargetMifareParameters mifare, TargetFeliCaParameters feliCa, TargetPiccParameters picc)
+        public (TargetModeInitialized modeInialized, byte[] initiator) InitAsTarget(TargetModeInitialization mode,
```
joperezr: We probably shouldn't delete the parameter documentation and instead fill it out.

src/devices/Pn532/Pn532.cs
```diff
@@ -1588,43 +1957,68 @@ private int ReadResponseI2C(CommandSet commandSet, Span<byte> readData)
...
                 return -1;
+            }
+
             // Finally, we can read the data
             if (length - 2 > 0)
             {
-                //var buff = new byte[length - 2];
-                //_i2cDevice.Read(buff);
-                //buff.AsSpan().CopyTo(readData);
+                ////var buff = new byte[length - 2];
```
krwq: ?

src/devices/Pn532/Pn532.cs
pgrawehr: A comment line should be preceeded by an empty line, but it seems some extra lines got introduced (maybe due to moving members around). I'll fix it.

src/devices/Pn532/Pn532.cs
pgrawehr: I agree, but I have no idea what this does, so I'd rather not write anything than something wrong.

src/devices/Pn532/Pn532.cs
pgrawehr: Four slashes are used to comment out code (since two slashes alone would trigger the "add a space after //" message). Should we rather remove the commented out code completely? Undocumented inactive code is confusing, especially since the above comment suggests that `Read()` is what we want to do here.

krwq: yes, we should remove it except for cases where it's showcasing something but I don't think that's the case here

pgrawehr:
If this is also merged, I'll prepare the PR to enable StyleCop.

PR #906, opened 2019-12-19
Update pins used by tests to reflect new lab setup

krwq:
Setup of the lab was updated today and now different pins are connected to each other.

This PR only updates pins to unblock CI, I will make proper upgrade by merging https://github.com/dotnet/iot/pull/758


krwq:
@joperezr merging to unblock CI, will leave 3.0 open since there is no pressure there

PR #908, opened 2019-12-20
Test StyleCop is working in PRs

krwq:
double checking the CI works the same as local builds


krwq:
@pgrawehr errors look good to me

PR #911, opened 2019-12-20
Add static modifier to Tm1637.MaxSegments

smdn:
`Tm1637.MaxSegments` should be referable without instantiation.

### DIFF
--- a/src/devices/Tm1637/Tm1637.cs
+++ b/src/devices/Tm1637/Tm1637.cs
@@ -20,7 +20,7 @@ namespace Iot.Device.Tm1637
         /// <summary>
         /// The number of segments that the TM1637 can handle
         /// </summary>
-        public byte MaxSegments => 6;
+        public static byte MaxSegments => 6;
 
         // According to the doc, the clock pulse width minimum is 400 ns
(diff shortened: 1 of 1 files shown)


src/devices/Tm1637/Tm1637.cs
```diff
@@ -20,7 +20,7 @@ public sealed class Tm1637 : IDisposable
         /// <summary>
         /// The number of segments that the TM1637 can handle
         /// </summary>
-        public byte MaxSegments => 6;
+        public static byte MaxSegments => 6;
```
krwq: Actually I got a bit mixed feelings if this should be static or virtual in case we have a similar segment display which could extend this.. @Ellerbach thoughts?

smdn: `MaxSegments = 6` is a constant value determined by the number of anode output pins (GRID1~GRID6) of TM1637.
If we driving an anode common segment display with TM1637, 6(segments) is the maximum.
(It is not related to the number of segments of the display connected to TM1637.)

And class Tm1637 is `sealed`. It looks it is not considered to extend this class.

smdn:
Thank you😄

PR #912, opened 2019-12-21
StyleCop, some remaining issues

pgrawehr:
Fix the remaining open issues as per https://github.com/dotnet/iot/issues/816#issuecomment-567625504

Note: This will show a lot of changes of line 1 of several files, without visible difference. The only change is that the byte-order mark for UTF-8 was added to the files, which prevents usage with a wrong code table.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,6 +16,9 @@ indent_size = 2
 
 # C# files
 [*.cs]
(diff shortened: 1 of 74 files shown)


joperezr:
LGTM thanks @pgrawehr

PR #910, opened 2019-12-20
Fix wrong command in Tm1637.DisplayRaw(byte, byte)

smdn:
`Tm1637.DisplayRaw(byte segmentAddress, byte rawData)` is sending incorrect command.
With current implementation, it always displays `rawData` to the first segment, regardless of the parameter `segmentAddress`. (See repro code and image below)

`DisplayRaw` must send `DataCommand.AddressCommandSetting + segmentAddress` to set the display address, not `DataCommand.FixAddress + segmentAddress`.

Repro code:
```cs
using System;
using System.Device.Gpio;
using Iot.Device.Tm1637;

class Program
{
  static void Main(string[] args)
  {
    var display = new Tm1637(
      pinDio: 16,
      pinClk: 18,
      pinNumberingScheme: PinNumberingScheme.Board
    );

    display.Brightness = 7;
    display.ScreenOn = true;
    display.ClearDisplay();

    display.Display(0, Character.Digit0);
    display.Display(1, Character.Digit1);
    display.Display(2, Character.Digit2);
    display.Display(3, Character.Digit3);
    display.Display(4, Character.Digit4);
    display.Display(5, Character.Digit5);
  }
}
```

This code displays characters like below.
TM1637-broken

Fixed implementation displays characters like below as expected.
TM1637-fixed

PR #913, opened 2019-12-22
Fix TM1637 sample to refer correct enum member

smdn:
The sample for TM1637 throws an exception.

It attempts to parse non existent enum member `Character.Char0`.
It must parse `Character.Digit0`.

```
$ cd src/devices/Tm1637/samples/
$ dotnet build
$ dotnet ./bin/Debug/netcoreapp2.1/Tm1637.sample.dll 
Hello Tm1637!

Unhandled Exception: System.ArgumentException: Requested value 'Char0' was not found.
   at System.Enum.TryParseEnum(Type enumType, String value, Boolean ignoreCase, EnumResult& parseResult)
   at System.Enum.Parse(Type enumType, String value, Boolean ignoreCase)
   at System.Enum.Parse(Type enumType, String value)
   at Tm1637Sample.Program.Main(String[] args) in /home/pi/iot/src/devices/Tm1637/samples/Tm1637.sample.cs:line 55
```

### DIFF
--- a/src/devices/Tm1637/samples/Tm1637.sample.cs
+++ b/src/devices/Tm1637/samples/Tm1637.sample.cs
@@ -52,7 +52,7 @@ namespace Tm1637Sample
             Thread.Sleep(3000);
             for (int i = 0; i < 6; i++)
             {
-                rawData[i] = (Character)Enum.Parse(typeof(Character), $"Char{i}");
+                rawData[i] = (Character)Enum.Parse(typeof(Character), $"Digit{i}");
             }
 
             tm1637.Display(rawData);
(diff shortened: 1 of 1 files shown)


src/devices/Tm1637/samples/Tm1637.sample.cs
```diff
@@ -52,7 +52,7 @@ public static void Main(string[] args)
             Thread.Sleep(3000);
             for (int i = 0; i < 6; i++)
             {
-                rawData[i] = (Character)Enum.Parse(typeof(Character), $"Char{i}");
+                rawData[i] = (Character)Enum.Parse(typeof(Character), $"Digit{i}");
```
krwq: can you also update SegmentOrder so that it works correctly also on 4 segment? i.e.: ` tm1637.SegmentOrder = new byte[] { 2, 1, 0, 3, 5, 4 };`

smdn: Actually, I've faced similar issue before fixing `Display(byte, byte)`.
`Display(ReadOnlySpan<Character>)` seems to be unstable or broken.
It sometime shows like [0 23], or [10 2], not [0123]. Some segment become blank. Or displays character on different segment...
And it is not because of `SegmentOrder` at least.

So I tried to fix `Display(byte, byte)` first to find the reason.
I think `Display(ReadOnlySpan<Character>)` has another problem...

krwq: Ok, will merge this as is since without it I can't even start the sample. Please add anything you're fixing to the sample project for testing purposes if it's not being tested yet.

smdn: Thank you for merging😄

And do you mean I have to add some code which using `Display(byte, byte)` to the sample project?

krwq: yes, anything you find not working :smile:

smdn: Thank you. Sample project updated (#910)

PR #914, opened 2019-12-22
Fast interrupt handling

pgrawehr:
This adds a special overload to the SysFsDriver when it is used as child for the RaspberryPiDriver, where only the interrupt callback is required. 
## Pros
- Fixes that the SysFsDriver is resetting the pull mode of a selected input pin during open.
- Removes extra sleeps for status update, as register-based access is used to get the current pin state
- Fixes validating that the pin is input when setting as an interrupt source
- I have tested interrupt frequencies of about 3khz working quite reliably when generated from an external source, with only loosing like 0.1 % of the pulses (but see below). 

## However:
- Interrupts (input pulses) shorter than about 150us are not recognized reliably, no matter what I do, pulses less than about 10us in length are never recognized (I have not found any information on this limitation on the web). A bit of extra electronics at the respective input pin can work around this limitation - will present a circuit for that. 
- The unit test I provide does not work reliably. It generates a number of interrupts and counts them, but often looses a few. Sometimes it looses a lot (like 50%), and I have not found out what this depends on. Increasing thread priority was a test, but didn't help much either. It could still be that this is dependent on CPU load. 

I'm kind of at a loss here. Has anybody seen any documentation on how fast this should go? Is there another way of reliably counting pulses? 

The PR (after some cleanup) certainly helps in improving the interrupt performance, but it's not where I want it to be - any suggestions are welcome. 

As mentioned above, here's a simple circuit design that helps in detecting pulses that are to short to be otherwise detected. This shows the usage with the ADS1115 ADC, whose "conversion ready" output is to short to be otherwise detected. (I'll present another PR for these improvements). The circuit detects short low-active pulses and makes them long enough to be detected. 

Pulse_lengthener_schem


src/System.Device.Gpio/Interop/Unix/Libc/Interop.threads.cs
```diff
@@ -0,0 +1,63 @@
...
+        IntPtr thread = pthread_self();
+        sched_param sched = new sched_param();
+
+        sched.sched_priority = sched_get_priority_min(SCHED_FIFO);
+        return pthread_setschedparam(thread, SCHED_FIFO, ref sched) == 0;
+    }
+
+    internal static void SetCurrentThreadNormalHighPriority()
+    {
+        nice(19);
```
krwq: Isn't this the lowest priority? Thread is nice when it shares more time :smile:

src/System.Device.Gpio/Interop/Unix/Libc/Interop.threads.cs
```diff
@@ -0,0 +1,63 @@
...
+// Disable these StyleCop rules for this file, as we are using native names here.
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+#pragma warning disable SA1307 // Field should begin with upper-case letter
+
+using System;
+using System.Runtime.InteropServices;
+
+internal partial class Interop
+{
+    internal static bool SetCurrentThreadHighPriority()
```
krwq: I don't think our library should be changing thread priorities, that should be on user or at least be opt-in

krwq:
@pgrawehr have you perhaps looked at https://www.kernel.org/doc/html/latest/admin-guide/gpio/sysfs.html to see if there is any better way to do GPIO events

src/System.Device.Gpio/Interop/Unix/Libc/Interop.threads.cs
pgrawehr: Yes, I know. I just copied this file from devices/Interop and left as-is. The method is currently not used here (and I'm confused about it's name as well, but I didn't check where it's used now)

src/System.Device.Gpio/Interop/Unix/Libc/Interop.threads.cs
pgrawehr: If possible, yes, this should be avoided.

pgrawehr:
@krwq: https://www.kernel.org/doc/Documentation/gpio/sysfs.txt is a bit more extensive than what you just linked, but otherwise the same. Our implementation follows the guidelines there. I just optimized away some unused function calls and explicit delays. 
All the other stuff seems to be for kernel-mode drivers only. The sysfs approach seems to be the only possibility to get interrupts in user mode. The only other option that I currently see (apart from writing a kernel driver) is to poll the gpio interrupt registers, but that would no longer be an "interrupt" driven approach and eat a lot of CPU.

pgrawehr:
I'm just confused: Do the automatic tests involve running the tests on the Pi already? Otherwise, I would be really surprised they all passed.

krwq:
@pgrawehr they've been running on the PIs for several months already

krwq:
@pgrawehr it's my fault if CI is behaving weird today :smile:

pgrawehr:
I cleaned up the branch now, removed the thread priority stuff and tested again - seems to work reliably now. Maybe there was too much other CPU load involved before. 

Something is weird with the automated tests though - they take a lot more time than what they used to.

src/System.Device.Gpio/System/Device/Gpio/WaitForEventResult.cs
```diff
@@ -10,10 +10,15 @@ namespace System.Device.Gpio
...
         /// <summary>
-        /// The event types to wait for.
+        /// The event types that were waited for.
         /// </summary>
         public PinEventTypes EventTypes;
 
+        /// <summary>
+        /// The event type that was actually seen
+        /// </summary>
+        public PinEventTypes DetectedEventTypes;
```
joperezr: Currently this is supposed to be the EventTypes as this is a result structure. If we want to add public surface area like this, we would require to first log an issue with the proposal so that it could be discussed before we add it to our API.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -22,6 +22,23 @@ private void ValidatePinNumber(int pinNumber)
             }
         }
 
+        private void ValidatePinIsInput(int pinNumber)
```
joperezr: If this isn't really used in the Windows implementation and is private, I would keep it in the .Linux specific implementation.

src/System.Device.Gpio/System/Device/Gpio/WaitForEventResult.cs
pgrawehr: @joperezr See Issue #925: The documentation for `EventTypes` said, quite ambiguously "The event types to wait for.", which can be interpreted as the desired event types for the wait. That's what was returned by the linux implementation (but that's IMHO quite useless). The Windows implementation, on the other hand, returned the actually seen event type. 
So we can either 
- Fix the linux implementation and make the documentation more clear
- Add this extra member and clearly distinguish the two cases

Both changes are breaking, the one for windows clients, the other for linux clients.

joperezr: Of those two I'd go with the first option.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -22,6 +22,23 @@ private void ValidatePinNumber(int pinNumber)
             }
         }
 
+        private void ValidatePinIsInput(int pinNumber)
+        {
+            ValidatePinNumber(pinNumber);
+            if (!IsPinOpen(pinNumber))
```
joperezr: Seems like this method will only be called when unregistering a pin callback, which is already ensuring that the pin was opened previously, so this check might just be doing it twice.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -58,6 +58,11 @@ protected override void Dispose(bool disposing)
             }
         }
 
+        private bool IsPinOpen(int pin)
```
joperezr: If this is only used on unregistering callbacks, the controller is already ensuring that the pin is open so this would be doing it twice.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -103,11 +107,17 @@ protected internal override void ClosePin(int pinNumber)
...
             ValidatePinNumber(pinNumber);
 
-            if (_sysFSModes.ContainsKey(pinNumber) && _sysFSModes[pinNumber] == PinMode.Output)
+            lock (s_sysFsInitializationLock)
             {
-                Write(pinNumber, PinValue.Low);
-                SetPinMode(pinNumber, PinMode.Input);
+                if (_sysFSDriver != null)
+                {
+                    _sysFSDriver.ClosePin(pinNumber);
```
joperezr: I believe this is wrong as it could be the case that a pin is being used by the controller but it isn't open in the internal sysfs driver, so this would probably throw saying that we can't close a pin that isn't open.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -166,11 +176,10 @@ protected internal unsafe override PinValue Read(int pinNumber)
         /// <param name="callback">Delegate that defines the structure for callbacks when a pin value changed event occurs.</param>
         protected internal override void RemoveCallbackForPinValueChangedEvent(int pinNumber, PinChangeEventHandler callback)
         {
-            ValidatePinNumber(pinNumber);
+            ValidatePinIsInput(pinNumber);
```
joperezr: I could see some scenarios where also having callbacks for pins set to output could be useful, for example if the setting of the pin value is happening on a different thread and you want to react to it every time so I wouldn't restrict callbacks only to input pins.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -316,11 +325,10 @@ private void SetInputPullModePi4(int pinNumber, PinMode mode)
         /// <returns>A structure that contains the result of the waiting operation.</returns>
         protected internal override WaitForEventResult WaitForEvent(int pinNumber, PinEventTypes eventTypes, CancellationToken cancellationToken)
         {
-            ValidatePinNumber(pinNumber);
+            ValidatePinIsInput(pinNumber);
```
joperezr: Same comment about not restricting the pins to be input for eventing to work.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
pgrawehr: Renaming the method to ValidatePinIsOpen and moving the the .linux part. Then that's what should actually be tested, I believe.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
pgrawehr: See above.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
pgrawehr: Note: The Windows implementation already behaves that way: Any operation on a pin that is not open throws, except for opening it and for closing it (probably because repeatedly closing a pin should just be ignored)

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: No. Note that (despite it's name) the _sysFSModes dictionary contains the modes of _all_ pins, not only the ones in use by SysFs. I'll fix that name.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: We could check our dictionary, but closing a pin that is not open is explicitly not an error.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: See above. Changed to "ValidatePinIsOpen" (ValidatePinNumber just checks that the number is in range)

src/System.Device.Gpio/System/Device/Gpio/WaitForEventResult.cs
pgrawehr: Ok, I'll change it that way.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: @joperezr Ok, now I'm getting what you mean. It's a bit inconsistent tough that the controller and the drivers use different semantics here. All existing driver implementations explicitly allow duplicate opening/closing of pins, the controller however does not. What should be the correct behavior? 
Removing all (or most) validations from the driver is probably also not so nice, since these classes are public, too and could be used by themselves.

joperezr: The idea is that the controller will do all the validation that is in it's power to do, in order to avoid code duplication. This includes checking weather a pin is open or not when trying to peform operations with it, since the controller has a list of all the pins that are opened, so it is able to know if the request is valid or not. The drivers in the other hand are responsible of performing all the validation that is driver-specific, for example, if you try to open pin 100 on a Raspberry pi driver, then the controller won't  know if this is a valid pin number, but the raspberry pi driver will know that this is invalid, so it should throw. That is at least the model we have tried to follow so far, so that is why you still see some validation being made on the driver side, in theory that should only happen when it is a driver-specific check.

pgrawehr:
@joperezr Can you take another look at this, please? I think it should now be stable (I was having some weird test problems, which after a lot of debugging, apparently were caused by events that were pending at the time a callback was registered). 
This should also solve the issues from #925 and (thanks to the ideas that came up in #939) uses libgpiod (if available) for callback handling on the Raspi.

joperezr:
awesome, I'll take a look at this today!

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10DriverPin.Windows.cs
```diff
@@ -141,7 +141,7 @@ void Handler(WinGpio.GpioPin s, WinGpio.GpioPinValueChangedEventArgs a)
 
                 return new WaitForEventResult
                 {
-                    EventTypes = pinEventTypes,
+                    EventTypes = eventOccurred ? pinEventTypes : PinEventTypes.None,
```
joperezr: Is this actually required? I see that we are already setting the PinEventTypes to none on l ine 121, so I believe that this extra check is not required.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -27,7 +27,7 @@ public unsafe partial class RaspberryPi3Driver : GpioDriver
         private const string DeviceTreeRanges = "/proc/device-tree/soc/ranges";
         private const string ModelFilePath = "/proc/device-tree/model";
 
-        private readonly IDictionary<int, PinMode> _sysFSModes = new Dictionary<int, PinMode>();
+        private readonly IDictionary<int, (PinMode CurrentPinMode, bool InUseBySysFs)> _pinModes = new Dictionary<int, (PinMode, bool)>();
```
joperezr: Now that I think of it, should this be a concurrentDictionary instead in case we are using it from multi-threaded scenarios?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -90,7 +72,7 @@ protected internal override void AddCallbackForPinValueChangedEvent(int pinNumbe
             InitializeSysFS();
 
             _sysFSDriver.OpenPin(pinNumber);
-            _sysFSDriver.SetPinMode(pinNumber, GetModeForUnixDriver(_sysFSModes[pinNumber]));
+            _pinModes[pinNumber] = (_pinModes[pinNumber].CurrentPinMode, true);
```
joperezr: Why are we not setting the pinmode any longer? I believe that is still required right?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -103,11 +85,19 @@ protected internal override void ClosePin(int pinNumber)
         {
             ValidatePinNumber(pinNumber);
 
-            if (_sysFSModes.ContainsKey(pinNumber) && _sysFSModes[pinNumber] == PinMode.Output)
+            lock (s_sysFsInitializationLock)
```
joperezr: NIT: if we make this dictionary a concurrentdictionary then I believe we won't need this lock any longer.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -462,7 +452,14 @@ private void InitializeSysFS()
                     return;
                 }
 
-                _sysFSDriver = new SysFsDriver();
+                try
+                {
+                    _sysFSDriver = new LibGpiodDriver(0);
+                }
+                catch (PlatformNotSupportedException)
```
joperezr: Is it PlatformNotSupported what we throw? I thought it was FileNotFound when we try to PInvoke and don't find it.

joperezr: Nvm, I just saw that we catch the dllNotFound on the constructor and then throw a PNSE

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -462,7 +452,14 @@ private void InitializeSysFS()
                     return;
                 }
 
-                _sysFSDriver = new SysFsDriver();
+                try
+                {
+                    _sysFSDriver = new LibGpiodDriver(0);
```
joperezr: >_sysFSDriver [](start = 20, length = 12)

NIT: We should rename _sysFSDriver to instead be something like: _interruptDriver or something since it will be confusing if a libgpioddriver variable is calld sysfs

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -61,25 +64,32 @@ protected internal override void AddCallbackForPinValueChangedEvent(int pinNumbe
 
         private LibGpiodDriverEventHandler PopulateEventHandler(int pinNumber)
         {
-            if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle pinHandle))
+            lock (_pinNumberLock)
```
joperezr: Adding all these locks will hit performance, should we instead just make _pinNumberToSafeLineHandle a ConcurrentDictionary?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: I'll check again, but IIRC this would be conflicting, because we set the Pin mode using the Raspi code (via /dev/gpiomem). Like this, we can also set the Pin Mode to PullUp or PullDown, which is not supported by SysFs.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
pgrawehr: I first considered that, but ConcurrentDictionary has a quite different API than Dictionary, so the two don't just work interchangeably. Also of course, a Concurrent<Something> doesn't help when we do `if (!HasKey(...) { Add(xxx) }` sequences. I'll take another look, though.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: Correct, yes. The pin must have been open at this time (or the controller would have thrown an exception). And when the pin was opened, the PinMode was already set, to any mode possible by _this_ driver, which includes pullup stuff the interrupt driver does not support (at least if it's sysfs)

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: True. Renamed all sysfs variables to interrupt

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: The lock here is wrong altogether, since it's for the driver, not the _pinModes array. I'll just init the driver in the ctor and forget about locking the driver.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
pgrawehr: I replaced the dictionary with a fixed-size array. This is inherently thread-safe (as long as different pins are used, at least) and a lot faster than a dictionary.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10DriverPin.Windows.cs
pgrawehr: No, it was not required, of course.

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -358,6 +403,105 @@ public void WaitForEventFallingEdgeTest()
...
+                controller.UnregisterCallbackForPinValueChangedEvent(InputPin, Callback);
+                // All pulses must be marked as rising edge
+                Assert.Equal(numInterrupts, numRisingEdges);
+                Assert.True(numInterrupts >= numPulses - 1, $"Expected at least {numPulses - 1}, got only {numInterrupts}"); // Allow one missing pulse
+                // That's how long this test should last (at most).
+                // It will usually be around 3s, but on Windows it's around 4s (for reasons unknown)
+                double expectedMillis = numPulses * (pulseLength + waitTime);
+                Assert.True(elapsed < TimeSpan.FromMilliseconds(expectedMillis * acceptableTimeFactor), $"Used {w.Elapsed.TotalMilliseconds}ms, expected max {expectedMillis * acceptableTimeFactor}ms");
+                // Write the state of the test to the log, even if it passed (for detailed analysis)
+                _testOutputHelper.WriteLine($"Got {numInterrupts} out of {numPulses} expected. Expected a duration of {expectedMillis}, actually used {elapsed.TotalMilliseconds}ms");
```
pgrawehr: This should be somewhere in the output, but it looks like the build log is only available if the build failed. Any idea how to find this line (for analysis) in a succeessful build?

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10DriverPin.Windows.cs
```diff
@@ -12,7 +12,7 @@ namespace System.Device.Gpio.Drivers
 {
     internal class Windows10DriverPin : IDisposable
     {
-        private const int ReasonableDebounceTimeoutMillseconds = 50;
+        private const int ReasonableDebounceTimeoutMilliseconds = 0;
```
pgrawehr: 50ms is way to long if you want to count events. This may be feasible for stuff like buttons, but not for digital inputs. This change makes the Windows build pass the fast event test.

pgrawehr: Note: If we want a bounce timeout feature, we should implement it globally, not in one place only (and different from the other drivers)

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
pgrawehr: Ok, I'm now using a combination of ConcurrentDictionary and a lock. IMHO timing in `OpenPin` and `ClosePin` methods is not critical. `Read` and `Write` don't need a lock now.

pgrawehr:
This is hopefully ready for final review (Ok, I've said that a few times already 😄 )

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
joperezr: Hey @pgrawehr sorry I just saw this. You can get this from here https://helix.dot.net/swagger/ui/index#!/WorkItem/WorkItem_ConsoleLog all you will need is the jobId which you should be able to get by looking at your build and search the Send To Helix step, and the logs for that should contain the GUID for job id listed. The id is always "System.Device.Gpio.Tests.dll" so by filling out those two parameters, you should be able to get the console output. Do keep in mind that we have a retention policy on the infrastructure so we won't have the logs for old builds.

pgrawehr: @joperezr Thanks, I'll have a look (I can always force a new build by pushing a dummy commit)

pgrawehr: It's a bit manual work to get the stuff I need, but I found it. The output is in the testresult.xml file, whose path can be found from the https://helix.dot.net/swagger/ui/index#!/WorkItem/WorkItem_Details section. 

It shows that for whatever reason, this test takes about 2100ms on Linux, but around 3800ms on Windows. I have no clue why this is so. Why would Thread.Sleep() be so inaccurate on windows?

pgrawehr: @krwq This is probably something you should once have a look into: It looks like the Windows tests are generally significantly slower than the linux tests. For instance, the relatively simple `PinValueReadAndWrite` test takes between 1ms and 12ms (depending on driver) on linux, but 23ms on Windows. I'll put this up as a separate ticket, as it is not related to this PR.

pgrawehr: Opened #1004

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -24,12 +24,18 @@ public class SysFsDriver : UnixDriver
...
+        private static readonly int s_pinOffset;
+        private int _pollingTimeoutInMilliseconds = Convert.ToInt32(TimeSpan.FromMilliseconds(1).TotalMilliseconds);
+        private int _statusUpdateSleepTime = Convert.ToInt32(TimeSpan.FromMilliseconds(1).TotalMilliseconds);
         private int _pollFileDescriptor = -1;
         private Thread _eventDetectionThread;
         private int _pinsToDetectEventsCount;
 
+        static SysFsDriver()
+        {
+            s_pinOffset = ReadOffset();
```
joperezr: Any reason why we chose to have a static constructor here instead? In general we try to avoid them when possible since they can have unintended effects like by changing completely the way that types are constructed. If there is no specific reason why we moved it to be like this, I would keep it as we had it with just doing an inline definition above.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -66,6 +72,36 @@ public SysFsDriver()
...
         }
 
+        /// <summary>
+        /// The internal timeout for the event wait function.
+        /// </summary>
+        internal TimeSpan PollingTimeout
+        {
+            get
+            {
+                return TimeSpan.FromMilliseconds(_pollingTimeoutInMilliseconds);
```
joperezr: If we intend to do this conversion everytime, why not just sotre it as a TimeSpan instead?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -290,25 +332,49 @@ protected internal override WaitForEventResult WaitForEvent(int pinNumber, PinEv
             AddPinToPoll(pinNumber, ref valueFileDescriptor, ref pollFileDescriptor, out bool closePinValueFileDescriptor);
 
             bool eventDetected = WasEventDetected(pollFileDescriptor, valueFileDescriptor, out _, cancellationToken);
+            if (_statusUpdateSleepTime > 0)
+            {
+                Thread.Sleep(_statusUpdateSleepTime); // Adding some delay to make sure that the value of the File has been updated so that we will get the right event type.
```
joperezr: Is this trying to protect to multi-threaded aps? because if so I think that this won't really work 100% of the time.

joperezr: oh, actually I see what the idea is here. I guess if our event is detected too fast you may have the case that the value file hasn't happened, is that possible with the Sysfs driver though? I thought that SysFS you are basically locking on that file change, so you would only really detect an event if the value of the file actually changes, plus I would imagine that the Read() operation alone would take a few ms which would give some extra time for the value to be right. Since you have done more testing, feel free to correct me if I'm wrong here.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -290,25 +332,49 @@ protected internal override WaitForEventResult WaitForEvent(int pinNumber, PinEv
...
-                EventTypes = eventTypes
+                EventTypes = detectedEventType,
             };
         }
 
         private void SetPinEventsToDetect(int pinNumber, PinEventTypes eventTypes)
         {
             string edgePath = Path.Combine(GpioBasePath, $"gpio{pinNumber + s_pinOffset}", "edge");
+            // Even though the pin is open, we might sometimes need to wait for access
+            SysFsHelpers.EnsureReadWriteAccessToPath(edgePath);
```
joperezr: Did you have to add this because some of our tests were failing because of this?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -402,17 +468,17 @@ private unsafe bool WasEventDetected(int pollFileDescriptor, int valueFileDescri
...
                 {
-                    throw new IOException("Error while trying to initialize pin interrupts.");
+                    throw new IOException("Error while waiting for pin interrupts.");
                 }
 
                 if (waitResult > 0)
                 {
                     pinNumber = events.data.pinNumber;
-
+                    /*
```
joperezr: I don't remember 100% from the top of my head, but I believe this code was required in order to reset the interrupt from epoll and that is why we added it. Is that not the case? If not, we should probably just remove this code instead of having it commented out.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -402,17 +468,17 @@ private unsafe bool WasEventDetected(int pollFileDescriptor, int valueFileDescri
 
             while (!cancellationToken.IsCancellationRequested)
             {
-                // poll every 1 millisecond
+                // Wait until something happens
```
joperezr: I believe we actually can't change it like this, because if we do then we won't be able to detect if a cancellation was requested, that is why we were polling every 1 ms. The idea is that if you are running a multithreaded app, you may have a different thread that will eventually say that it doesn't want to keep blocking this, so it might send a cancellation request, and if you have pollingtimeout set to 0, then this will never be detected by this thread and the execution will continue

src/System.Device.Gpio/System/Device/Gpio/Drivers/UnixDriverDevicePin.Linux.cs
```diff
@@ -42,6 +56,7 @@ public void Dispose()
                 FileDescriptor = -1;
             }
 
+            ActiveEdges = PinEventTypes.None;
```
joperezr: Technically this is not required. This s just a regular enum, so doesn't need resetting to original value on dispose.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -571,10 +641,56 @@ private unsafe void DetectEvents()
...
-                        Thread.Sleep(1); // Adding some delay to make sure that the value of the File has been updated so that we will get the right event type.
-                        PinEventTypes eventTypes = (Read(pinNumber) == PinValue.High) ? PinEventTypes.Rising : PinEventTypes.Falling;
-                        var args = new PinValueChangedEventArgs(eventTypes, pinNumber);
-                        _devicePins[pinNumber]?.OnPinValueChanged(args);
+                        if (_statusUpdateSleepTime > 0)
+                        {
+                            Thread.Sleep(_statusUpdateSleepTime); // Adding some delay to make sure that the value of the File has been updated so that we will get the right event type.
+                        }
+
+                        var currentPin = _devicePins[pinNumber];
```
joperezr: NIT: please try to change all of the instances of `var` into the explicit types intead, it helps a lot making the code more readable

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -571,10 +641,56 @@ private unsafe void DetectEvents()
...
+                                eventType = PinEventTypes.Rising;
+                                secondEvent = PinEventTypes.Falling;
+                            }
+
+                            currentPin.LastValue = newValue;
+                        }
+                        else
+                        {
+                            // Update the value, in case we need it later
+                            currentPin.LastValue = Read(pinNumber);
```
joperezr: I think that this READ should be the very first thing we do after the thread.sleep above, since we want to avoid as much as possible some other events to happen while doing other calculations and this read operation to get wrong value.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -571,10 +641,56 @@ private unsafe void DetectEvents()
...
-                        _devicePins[pinNumber]?.OnPinValueChanged(args);
+                        if (_statusUpdateSleepTime > 0)
+                        {
+                            Thread.Sleep(_statusUpdateSleepTime); // Adding some delay to make sure that the value of the File has been updated so that we will get the right event type.
+                        }
+
+                        var currentPin = _devicePins[pinNumber];
+                        var activeEdges = currentPin.ActiveEdges;
+                        PinEventTypes eventType = activeEdges;
+                        PinEventTypes secondEvent = PinEventTypes.None;
```
joperezr: Do we really get the secondEvent used a lot? If so, I would question whether we should even handle this case instead of just dropping a few events by design, since I guess in this case you won't really get the second event if there where 3 events between event and read operation, or we won't detect 4 etc.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: This may now look weird, yes. In fact, I copied this from line 574 (method DetectEvents()) where a comment says that this is needed. I didn't want to change the SysFs implementation when used standalone, so I first just made sure the two equivalent methods are implemented similarly. 

As a second step then, I reduced the _statusUpdateSleepTime to 0 for the InterruptSysFsDriver, because there the Read() operation uses the registers and is therefore faster and never needs a wait. 

If you want, I can do another test whether that delay is required for SysFs alone. Maybe that comment there on line 574 is just wrong.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: Not 100% sure, but the previous line `private static readonly int s_pinOffset = ReadOffset();` supposedly creates the exact same code. I tend to write explicit static ctors when the operation is lengthy.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: Yes, I saw a few failures here.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
joperezr: You are right that the effect it has is very similar, except that it is much easier to understand and control the execution time of the previous way the code was, than static constructors. Those run at an incontrolable time and may have unintended effects if you had any other static fields that you were auto setting, or if you had subclasses that did. Also, it makes the code a bit less readable, so if there is no specific reason why we really want static constructors, I would suggest to avoid them.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: I removed it because the code doesn't do anything (the values read are all discarded). Everything works fine without this, for all the tests I've done.
I'll remove it.

joperezr: It's true that all values read are discarded, but the important part of this code was that by running the lseek, we were effectively resetting the epoll set, so it was kind of marking the notification as read if you will, so I'm worried now that after removing this, we will continue to get hits in epoll_wait() event when we didn't really get a new event.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: The second event here is guesswork, yes. But it's required to be consistent in the end. 

The logic is as follows: If the previous state is low, we get an event and the current state is still low, we know that there must have been a rise and a fall (otherwise there wouldn't have been an event in the first place). We can't tell whether there have been 4 or 6, but we know we saw an even number of events, which is pretty important for instance if this is connected to a button for instance, where you would only get the raise but not the fall if the contact is short.

Put differently, this should ensure that you don't get two consecutive raises or falls.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: I see your point. But wouldn't we see test failures if that was the case? Especially my FastInterruptTest would certainly detect many more events than expected. I'll make sure this would be detected.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: I don't understand. The cancellationToken is checked in the while loop above, so this will work even if the pollingtimeout is 0. This will only block forever if pollingtimeout is -1, which probably shouldn't be used. Also, the idea is to _increase_ the pollingtimeout to 100ms or so, because this reduces the time during which you are not waiting. And if an event happens while you are not inside the epoll_wait function, it won't be detected. I'll test this again, now that the code is otherwise stable.

joperezr: oh my bad I thought 0 on epoll had the same effect as -1, so that is why I said that if this blocks indefinitely then a cancellation request wouldn't get detected. I actually have never tried setting it to 0, I suppose it would never block so it just checks and then goes on.

>And if an event happens while you are not inside the epoll_wait function, it won't be detected

This I believe is actually not true. That is why we used to do the LSeek before, since even if you get an event in between running the loop but not actually on the epoll_wait function, the epoll_wait will still return true next time you call it. This we have test before to make sure we wouldn't miss events like this, and that is actually why after getting an event we were using the LSeek to make sure that we reset the event so epoll wouldn't treat it as unprocessed again.

pgrawehr: @joperezr It seems you're right. It doesn't actually matter what time I use, all the tests pass anyway (even if I add an extra Sleep(5)). So the timeout should probably just be selected such that it is a compromise between a fast reaction to a Cancel and extra CPU load (due to awaking the CPU and performing useless loops). I'd go for something between 50ms and 100ms. 1ms (the previous value) is to low, IMHO.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: @joperezr I can't confirm your observation, or I wouldn't know how. All the tests pass whether this code is enabled or not. And I'm not getting extra events in any tests, it seems.

src/System.Device.Gpio/System/Device/Gpio/Drivers/UnixDriverDevicePin.Linux.cs
pgrawehr: I know. This is just to make sure the state of the class is consistent. Just in case somebody reuses the disposed instance.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: Set to a constant of 50ms.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: I agree. Changed.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: Oh, and yes: It makes sure that the last reported change corresponds to the current state (that means that the last change was fall when the line stays low for a prolonged time)

pgrawehr:
@joperezr Sorry for bothering, but I'm waiting on your reply to the above points, so this can be finally closed.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
joperezr: I suppose doing this small operation doesn't really affect performance much, so would you mind keeping it for now and logging an issue to do much more deep investigation on whether or not we can remove it so that we do that appart from this PR?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
joperezr: I see, yeah I see your point and I think it makes sense so it's fine to keep it as is

joperezr:
I'm so sorry for taking so long on looking at this @pgrawehr. I think I answered both open questions now, so basically one more thing to do, would be to for now readd back the code that is doing the file read on the sysfs driver since we have to do more investigation in depth before removing it. Adding a tracking issue to remove it would be great but I don't want to hold up this PR more time just while we do that investigation. Again thanks so much for your patience, and once that change is in we can go ahead and merge this.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
pgrawehr: That's acceptable for me, since the preferred solution is now anyway to use libgpiod. I've created #1028 .

pgrawehr: @joperezr Done

TrabacchinLuigi:
Any chance to get this on a preview nuget package ?

joperezr:
Yes, we are overdue now for a preview package so I will work on this today.

joperezr:
@TrabacchinLuigi I have now pushed to NuGet preview 2 of 1.1 release which should include this change: https://github.com/dotnet/iot/releases/tag/1.1-preview2

TrabacchinLuigi:
Super nice! i'll rush to give it a try!
Think i'll run a small benchmark with my raspberry pi 3 and an ne555 astable, wish i had an oscilloscope...

TrabacchinLuigi:
I did not get very nice results, nowhere near the 3Khz now i don't know if it's my test setup, so i'm gonna [share it with you](https://gist.github.com/TrabacchinLuigi/a9907169ca5da545daacf5ac361ca60e) (sorry for the hardcoded pins, the misleading comments, and the poor design, i just needed something fast. so i reused the project i made for IR logging)

To this i connected my NE555 which i expected to span from 7Hz to ~1200Hz but i fear my capacitor it's half the value it has written on because i managed to get only 600Hz, STILL the results where not that good thigs got slightly better with raspberrypi3 driver ... 

[here my results](https://1drv.ms/x/s!AhFYc7b2lqsg2W1XYpX5LKAPeei4?e=YH7Xli)

pgrawehr:
> 
> 
> I did not get very nice results, nowhere near the 3Khz now i don't know if it's my test setup, so i'm gonna [share it with you](https://gist.github.com/TrabacchinLuigi/a9907169ca5da545daacf5ac361ca60e) (sorry for the hardcoded pins, the misleading comments, and the poor design, i just needed something fast. so i reused the project i made for IR logging)
> 
> To this i connected my NE555 which i expected to span from 7Hz to ~1200Hz but i fear my capacitor it's double the value it has written on because i managed to get only 600Hz, STILL the results where not that good thigs got slightly better with raspberrypi3 driver ...
> 
> [here my results](https://1drv.ms/x/s!AhFYc7b2lqsg2W1XYpX5LKAPeei4?e=YH7Xli)

I can't see your results, that page is seemingly password-protected. (you can post files in the comments here). Currently we have tests setups that run at about 1khz on Pi3's, so I expect that to be working. Personally, I tested it with a scope on a Pi4, and it worked quite reliably. (maybe one out of 1000 events lost). You do need to make sure that the pulses are long enough. If the pulses themselves are shorter than about 150ns, they won't be recognized at all. Do use the RaspberryPi3Driver, this is by far the fastest.

Also, you can use `PinEventTypes.Rising | PinEventTypes.Falling` as argument to `RegisterCallbackForPinValueChangedEvent()` to listen for both rising and falling events.

TrabacchinLuigi:
weird, i'll post the file here: 
[raspi.xlsx](https://github.com/dotnet/iot/files/4973884/raspi.xlsx)
pulses are 1.5~2ms so... it should work quite well ... Maybe it's because i registered the event twice instead of using the or operator ...

pgrawehr:
The diff column in your data shows about 1ms between events (0.001 seconds). If the clock is really 600Hz, that's pretty much the value you would expect, wouldn't you? 

Why are you always diffing between 4 events and not with the previous one? You'll probably get more accurate results if you only monitor either rising or falling edges, then you don't need to care about picking the right pair. Note also that the column with the name "Diff ms" is actually the (instantaneous) frequency, which really is near 600 Hz. To get ms from s, just multiply by 1000.

TrabacchinLuigi:
every event print two lanes (so i can draw them on the graph) so i diff from the same event type to measure frequency. 
i catched both of them cause i wanted to draw the full nice graph and see the duty cycle, like a silly oscilloscope 🤓 ...
I just realized the fix is for the RaspberryPi driver not the sysFS... so yes it does work better! and i just got few capacitors to work with so i can test higher speeds now

PR #915, opened 2019-12-23
Ads1115 extensions

pgrawehr:
This adds several extra features to the Ads1115 binding, including:
- `ReadVoltage()` function to directly get the measured value
- `ReadVoltage(InputMultiplexer) overload, which allows selecting the desired channel directly (useful when different channels are required in turns). 
- `EnableComparator` and `EnableConversionReady` methods, to allow using the interrupt pin (see the Readme update for details). 

Note that the last feature requires PR #914, as otherwise the callback is not fast enough to be recognized. Example output of this feature from test program, with sample rate set to 250 Hz and including PR #914 and the extra circuitry described in the readme (without either, the number of received pulses is very very low): 

InterruptResult


pgrawehr:
Do not merge yet - need to investigate a bug that causes the wrong value to be retrieved unter some conditions when switching channels.

pgrawehr:
@joperezr This really took a while, but now this should be ready for review as well.

joperezr:
Awesome @pgrawehr thank you so much for working on this. I apologize for it taking long time for me to look at this, but it is definitely on the things I'll do by end of week.

src/devices/Ads1115/Ads1115.cs
```diff
@@ -18,8 +20,22 @@ public class Ads1115 : IDisposable
 
         private InputMultiplexer _inputMultiplexer;
 
+        private MeasuringRange _measuringRange;
+
+        private DataRate _dataRate;
+        private DeviceMode _deviceMode;
+
+        private GpioController _gpioController;
+        private int _alrtRdyPin;
```
krwq, line 29: nit: please do not use abbreviations in variable names (with some super common exceptions like `i`, `id`, `xml`, etc)

src/devices/Ads1115/Ads1115.cs
```diff
@@ -61,38 +79,136 @@ public DataRate DataRate
             }
         }
 
+        /// <summary>
+        /// ADS1115 operation mode.
+        /// When set to <see cref="DeviceMode.Continuous"/> the chip constantly measures the input and the values can be read directly.
```
krwq, line 84:
```suggestion
        /// When set to <see cref="DeviceMode.Continuous"/> the chip continously measures the input and the values can be read directly.
```

src/devices/Ads1115/Ads1115.cs
```diff
@@ -61,38 +79,136 @@ public DataRate DataRate
...
+        /// If set to <see cref="DeviceMode.PowerDown"/> the chip enters idle mode after each conversion and
+        /// a new value will be requested each time a read request is performed. This is the recommended setting when constantly
```
krwq, line 86:
```suggestion
        /// a new value will be requested each time a read request is performed. This is the recommended setting when frequently
```

src/devices/Ads1115/Ads1115.cs
```diff
@@ -61,38 +79,136 @@ public DataRate DataRate
...
+                _deviceMode = value;
+                SetConfig();
+            }
+        }
+
+        /// <summary>
+        /// Comparator mode.
+        /// Only relevant if the comparator trigger event is set up and is changed by <see cref="EnableComparator(short, short, ComparatorMode, ComparatorQueue, GpioController, int)"/>.
+        /// </summary>
+        public ComparatorMode ComparatorMode
```
krwq, line 106: This could be just `public ComparatorMode ComparatorMode => _comparatorMode` or even better `public ComparatorMode ComparatorMode { get; }` and remove the explicit field

src/devices/Ads1115/Ads1115.cs
```diff
@@ -104,14 +220,255 @@ private void SetConfig()
...
 
         /// <summary>
-        /// Read Raw Data
+        /// Enable conversion ready event.
+        /// The <see cref="AlertReadyAsserted"/> event fires each time a new value is available after this method is called.
+        /// </summary>
+        /// <param name="gpioController">Reference to the GPIO controller</param>
+        /// <param name="pin">Pin to which the ALRT/RDY Pin is connected</param>
+        /// <exception cref="InvalidOperationException">The conversion ready event is already set up</exception>
+        public void EnableConversionReady(GpioController gpioController, int pin)
```
krwq, line 291: how often would the user call this with different gpioControllers? would it make sense to take this in the constructor with default value and change this to paramterless?

src/devices/Ads1115/Ads1115.cs
```diff
@@ -104,14 +220,255 @@ private void SetConfig()
...
+                // Writing a negative value to the max value register and a positive value to the min value register
+                // configures the ALRT/RDY pin to trigger after each conversion (with a transition from high to low, when ComparatorPolarity is Low)
+                WriteComparatorRegisters(short.MaxValue, short.MinValue);
+
+                gpioController.RegisterCallbackForPinValueChangedEvent(pin, ComparatorPolarity == ComparatorPolarity.Low ? PinEventTypes.Falling : PinEventTypes.Rising, ConversionReadyCallback);
+
+            }
+            catch (Exception)
+            {
+                gpioController.ClosePin(pin);
```
krwq, line 321: you can put this in `finally` clause to achieve the same effect

src/devices/Ads1115/Ads1115.cs
```diff
@@ -104,14 +220,255 @@ private void SetConfig()
...
+        /// In window comparator mode, the callback is triggered each time the measured value exceeds the given upper value or gets
+        /// less than the given lower value.
+        /// </summary>
+        /// <param name="lowerValue">Lower value for the comparator</param>
+        /// <param name="upperValue">Upper value for the comparator</param>
+        /// <param name="mode">Traditional or Window comparator mode</param>
+        /// <param name="queueLength">Minimum number of samples that must exceed the threshold to trigger the event</param>
+        /// <param name="gpioController">The GPIO controller, to configure the pin</param>
+        /// <param name="pin">The pin where the ALRT/RDY pin is connected to. </param>
+        public void EnableComparator(short lowerValue, short upperValue, ComparatorMode mode,
```
krwq, line 369: does it make sense to make this private? how would user know what is lower and upper value without reading the spec? I'm fine with exposing register as property on top of that but take uint/int instead of two shorts.

src/devices/Ads1115/Ads1115.cs
```diff
@@ -104,14 +220,255 @@ private void SetConfig()
...
+                _deviceMode = DeviceMode.Continuous;
+                SetConfig();
+                // Writing a negative value to the max value register and a positive value to the min value register
+                // configures the ALRT/RDY pin to trigger after each conversion (with a transition from high to low, when ComparatorPolarity is Low)
+                WriteComparatorRegisters(lowerValue, upperValue);
+
+                gpioController.RegisterCallbackForPinValueChangedEvent(pin, ComparatorPolarity == ComparatorPolarity.Low ? PinEventTypes.Falling : PinEventTypes.Rising, ConversionReadyCallback);
+
+            }
+            catch (Exception)
```
krwq, line 409: finally

src/devices/Ads1115/Ads1115.cs
```diff
@@ -104,14 +220,255 @@ private void SetConfig()
...
+        private ushort ReadConfigRegister()
+        {
+            Span<byte> retBuf = stackalloc byte[2];
+            Span<byte> request = stackalloc byte[1]
+            {
+                (byte)Register.ADC_CONFIG_REG_ADDR
+            };
+
+            _i2cDevice.WriteRead(request, retBuf);
+            return BinaryPrimitives.ReadUInt16BigEndian(retBuf);
```
krwq, line 428: presumably first byte of retBuf will always be zero (most of the devices would do that since you're writing register in the first byte) - if that's the case please just return byte here

src/devices/Ads1115/Ads1115.cs
pgrawehr: This is how the pin is called in the documentation. And also the text on my breakout board reads "ALRT"

src/devices/Ads1115/Ads1115.cs
pgrawehr: Of course, you would always use the same controller (since apparently, you have to wire a pin for that). I guess it would be more consistent with other bindings to make it an optional parameter in the constructor.

src/devices/Ads1115/Ads1115.cs
```diff
@@ -104,14 +220,255 @@ private void SetConfig()
...
+        /// Wait until the current conversion finishes.
+        /// This method must only be called in powerdown mode, otherwise it would timeout, since the busy bit never changes.
+        /// Due to that, we always write the configuration first in power down mode and then enable the continuous bit.
+        /// </summary>
+        /// <exception cref="TimeoutException">A timeout occured waiting for the ADC to finish the conversion</exception>
+        private void WaitWhileBusy()
+        {
+            // In powerdown-mode, wait until the busy bit goes high
+            ushort reg = ReadConfigRegister();
+            int timeout = 5000;
```
krwq, line 441: nit: I'd prefer this piece of code was written in such a way that units of timeout are clear so it's easier to adjust it in the future.

src/devices/Ads1115/Ads1115.cs
```diff
@@ -155,22 +602,51 @@ public double RawToVoltage(short val)
...
         /// <summary>
-        /// Cleanup
+        /// Cleanup.
+        /// Failing to dispose this class, especially when callbacks are active, may lead to undefined behavior.
         /// </summary>
         public void Dispose()
         {
             if (_i2cDevice != null)
             {
+                DisableAlrtReadyPin();
```
krwq, line 649: no abbreviations in function names

src/devices/Ads1115/Ads1115.cs
pgrawehr: No. This catch handler is only executed if an error occurs. If everything succeeds, the pin must not be closed here.

src/devices/Ads1115/Ads1115.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
krwq, line 1: any reason for changing encoding?

src/devices/Ads1115/Ads1115.sln
```diff
@@ -0,0 +1,65 @@
+﻿
```
krwq, line 1: cc: @joperezr

src/devices/Ads1115/Ads1115.cs
pgrawehr: No, see above.

src/devices/Ads1115/DeviceMode.cs
```diff
@@ -12,7 +12,7 @@ public enum DeviceMode
         /// <summary>Continuous mode</summary>
         Continuous = 0x00,
 
-        /// <summary>Power down mode</summary>
+        /// <summary>Power down mode, the chip is shutting down _after_ the next conversion</summary>
```
krwq, line 15: please remove underscores

src/devices/Ads1115/Ads1115.csproj
pgrawehr: All files should be UTF8-BOM for consistency. I didn't do this one on purpose, but VS is apparently updating this automatically since the setting in .editorConfig is set like that.

src/devices/Ads1115/Ads1115.cs
pgrawehr: It would certainly simplify the interface. However, we do offer the ReadRaw() method as well, which returns the same unit (and if you use that, you have to know how to convert the value).

src/devices/Ads1115/Ads1115.cs
pgrawehr: I don't quite understand what you mean, but this reads the 16 bit wide config register (register 0). And all bits are used and returned here.

src/devices/Ads1115/Ads1115.cs
pgrawehr: Yea, this is something I should have seen myself.

src/devices/Ads1115/Ads1115.cs
pgrawehr: Changed, since its the pin number on the gpio side anyway, not the number of the pin on the ADS

src/devices/Ads1115/Ads1115.sln
```diff
@@ -0,0 +1,65 @@
...
+		Linux-Debug|Any CPU = Linux-Debug|Any CPU
+		Linux-Release|Any CPU = Linux-Release|Any CPU
+		Release|Any CPU = Release|Any CPU
+		Windows-Debug|Any CPU = Windows-Debug|Any CPU
+		Windows-Release|Any CPU = Windows-Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{C60BB2A9-E067-4A0D-AF04-5D8B2FF1CE08}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{C60BB2A9-E067-4A0D-AF04-5D8B2FF1CE08}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{C60BB2A9-E067-4A0D-AF04-5D8B2FF1CE08}.Linux-Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
joperezr, line 24: Can we either remove the Solution or update it now that you have removed the configurations on your project?

src/devices/Ads1115/Ads1115.cs
```diff
@@ -14,12 +16,86 @@ namespace Iot.Device.Ads1115
...
+        /// </summary>
+        /// <param name="i2cDevice">The I2C device used for communication.</param>
+        /// <param name="inputMultiplexer">Input Multiplexer</param>
+        /// <param name="measuringRange">Programmable Gain Amplifier</param>
+        /// <param name="dataRate">Data Rate</param>
+        /// <param name="deviceMode">Initial device mode</param>
+        public Ads1115(I2cDevice i2cDevice, InputMultiplexer inputMultiplexer = InputMultiplexer.AIN0, MeasuringRange measuringRange = MeasuringRange.FS4096, DataRate dataRate = DataRate.SPS128,
+            DeviceMode deviceMode = DeviceMode.Continuous)
+        {
+            _i2cDevice = i2cDevice;
```
joperezr, line 52: I know we weren't doing it before either, but should we be checking for null here in case i2cDevice passed it was null?

src/devices/Ads1115/Ads1115.cs
```diff
@@ -14,12 +16,86 @@ namespace Iot.Device.Ads1115
...
+        /// <param name="inputMultiplexer">Input Multiplexer</param>
+        /// <param name="measuringRange">Programmable Gain Amplifier</param>
+        /// <param name="dataRate">Data Rate</param>
+        /// <param name="deviceMode">Initial device mode</param>
+        public Ads1115(I2cDevice i2cDevice,
+            GpioController gpioController, int gpioInterruptPin, bool shouldDispose = true, InputMultiplexer inputMultiplexer = InputMultiplexer.AIN0, MeasuringRange measuringRange = MeasuringRange.FS4096, DataRate dataRate = DataRate.SPS128,
+            DeviceMode deviceMode = DeviceMode.Continuous)
+            : this(i2cDevice, inputMultiplexer, measuringRange, dataRate, deviceMode)
+        {
+            _gpioController = gpioController ?? throw new ArgumentNullException(nameof(gpioController));
```
joperezr, line 85: Should we just initialize a board gpioController in this case and set shouldDispose to true?

src/devices/Ads1115/Ads1115.cs
```diff
@@ -62,37 +140,106 @@ public DataRate DataRate
...
 
-            SetConfig();
+        /// <summary>
+        /// Comparator mode.
+        /// Only relevant if the comparator trigger event is set up and is changed by <see cref="EnableComparator(short, short, ComparatorMode, ComparatorQueue)"/>.
+        /// </summary>
+        public ComparatorMode ComparatorMode
+        {
+            get;
+            private set;
```
joperezr, line 169: No need to call SetConfig() here?

src/devices/Ads1115/Ads1115.cs
```diff
@@ -62,37 +140,106 @@ public DataRate DataRate
...
+            {
+                return _comparatorQueue;
+            }
+        }
+
+        /// <summary>
+        /// This event fires when a new value is available (in conversion ready mode) or the comparator threshold is exceeded.
+        /// Requires setup through <see cref="EnableConversionReady"/> or <see cref="EnableComparator(double, double, ComparatorMode, ComparatorQueue)"/>.
+        /// </summary>
+        public event Action AlertReadyAsserted;
```
joperezr, line 222: NIT: usually .NET events follow the naming pattern of `OnEventRaised` so this case it would probably be something like `OnNewAlertReadyAsserted` or somehting like that: [https://docs.microsoft.com/en-us/dotnet/csharp/event-pattern](https://docs.microsoft.com/en-us/dotnet/csharp/event-pattern)

src/devices/Ads1115/Ads1115.cs
```diff
@@ -104,14 +251,235 @@ private void SetConfig()
...
-            Thread.Sleep(10);
+            WaitWhileBusy();
+
+            if (_deviceMode == DeviceMode.Continuous)
+            {
+                // We need to wait two cycles when changing the configuration in continuous mode,
+                // otherwise we may be getting a value from the wrong input
+                _i2cDevice.Write(writeBuff);
+                WaitWhileBusy();
+                configHi &= 0xFE; // Clear last bit
```
joperezr, line 262: NIT: usually for things like this we just do `&= ~0b01` which makes it easier to read.

joperezr:
Left some minor comments but things look mostly fine

src/devices/Ads1115/Ads1115.sln
pgrawehr: Sure, I'll clean that up or remove it. We should fix that latest after getting rid of those extra configs.

src/devices/Ads1115/Ads1115.cs
pgrawehr: Since the controller argument is not optional here, I thought it would be better to explicitly require the user to pass a valid instance. I'm still not so much a fan of this "lets try to guess what the user wanted" approach. But I can change it if you think it would be useful.

src/devices/Ads1115/Ads1115.cs
pgrawehr: No, the value can only be changed by calling EnableComparator, where the configuration is applied directly.

src/devices/Ads1115/Ads1115.sln
pgrawehr: I've deleted it for now. We should IMHO re-create these in a batch for all bindings when we've cleaned up the platform dependency.

src/devices/Ads1115/Ads1115.cs
pgrawehr: If I read that document correctly, that applies for the event handlers, not the names of the events themselves. 

It has 

```
void OnEventRaised(object sender, EventArgs args);
```
but 
```
public event EventHandler<FileFoundArgs> FileFound;
```

we're talking about the later case here.

src/devices/Ads1115/Ads1115.cs
pgrawehr: The ugly thing is that that requires an extra cast, because `~0b01` is -2 and therefore a signed value which cannot be directly applied to a `byte`.

pgrawehr:
@joperezr I addressed most review comments, please comment on the open questions above, so we can close this.

src/devices/Ads1115/Ads1115.cs
joperezr: Ok, if it is not optional, I'm fine with the way it is.

src/devices/Ads1115/Ads1115.cs
joperezr: You are correct.

src/devices/Ads1115/Ads1115.cs
joperezr: Yeah you are right, I  missed this was a byte and not an int. In that case, I would probably just do `configHi &= 0b_1111_1110;` Since I guess that makes it more readable than having to convert HEX to Binary on my mind 😄

Issue #860, opened 2019-11-21
hcsr04 not working


mringl:
I've a problem running the HCSR04 device on a Pi3 with dotnet core 3.1 preview 2.
I'll never get out of the first while(_controller.Read(_echo) == PinValue.Low)
If I read from _echo after _controller.Write(_trigger, PinValue.High); everything works fine.

I've also confirmed my setup via a python script which works correctly.

krwq:
Thanks for the report. Can you try running the same code as sudo and see if still repros? Also do you have libgpiod installed?

I think except for fixing the issue we should also add a protection against endless looping here.

@joperezr this is another place where it would be useful to know which GpioDriver does the GpioController use since this is really hard to diagnose otherwise

mringl:
Hi, sudo doesn't help, I've also now libgpiod installed but no luck.
I've also tried to use the _controller.WaitForEvent method instead of the loop, no luck.
By the way it is a Rasp 2 B not 3

krwq:
interesting, it could be related to Raspberry Pi 2, we might be using different driver (sysfs).

The eventing is kinda broken on sysfs driver (we currently are not sure if it's possible to fix it). We might be able to fixup the RaspberryPi3 driver though to work with RPi2. I'm surprised installing libgpiod hasn't fixed it, this is something we should look into.

One more thing to try: can you create GpioController using following line:
`new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver())` and pass it to the HCSR04

and see if you have any luck or perhaps some more meaningful error message?

Kash0321:
Same issue here. Is a Raspberry Pi 3B+ in my case... Tested with Python sensor works properly, but not with .NET Core / .NET IoT

I'd tried your last advice (new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver())) with no luck...

Kash0321:
I've found the problem right now, and it isn't a drivers issue... Problem is on Hcsr04 class: https://github.com/dotnet/iot/blob/master/src/devices/Hcsr04/Hcsr04.cs

In line 63, TimeSpan.FromMilliseconds(0.01) returns 0, then, no pulse is sended nor received by the sensor...

            // Trigger input for 10uS to start ranging
            _controller.Write(_trigger, PinValue.High);
            Thread.Sleep(TimeSpan.FromTicks(10)); <---- With this change, now it works
            _controller.Write(_trigger, PinValue.Low);

Kash0321:
[...]

PR #919, opened 2019-12-31
Make HCSR04 more reliable

krwq:
Fixes https://github.com/dotnet/iot/issues/860

Issue is caused by the first call to `_controller.Read` being to slow (due to JITting), that is causing missing the echo pulse and endless looping.

This change is:
- adding protection against indefinite looping
- adding protection against invalid reading (more than 400cm which is max supported distance)
- 10 retries on failed attempt, later we throw InvalidOperationException (unsure about which exception type, please comment if you have better suggestion)
- JITting `Read` in the constructor to avoid always retrying after starting the app
- add constructor taking GpioController

cc: @mringl @Kash0321

### DIFF
--- a/src/devices/Hcsr04/Hcsr04.cs
+++ b/src/devices/Hcsr04/Hcsr04.cs
@@ -42,6 +42,11 @@ namespace Iot.Device.Hcsr04
             _controller.OpenPin(_trigger, PinMode.Output);
 
             _controller.Write(_trigger, PinValue.Low);
+
+            // Call Read once to make sure method is JITted
+            // Too long JITting is causing that initial echo pulse is frequently missed on the first run
+            // which would cause unnecessary retry
+            _controller.Read(_echo);
         }
 
         /// <summary>
@@ -49,6 +54,27 @@ namespace Iot.Device.Hcsr04
         /// </summary>
         private double GetDistance()
         {
+            // Retry at most 3 times.
+            // Try method will fail when context switch occurs in the wrong moment
+            // or something else (i.e. JIT, extra workload) causes extra delay
+            for (int i = 0; i < 3; i++)
+            {
+                if (TryGetDistance(out double result))
+                {
+                    return result;
+                }
+            }
+
+            throw new InvalidOperationException("Could not get reading from the sensor");
+        }
+
+        private bool TryGetDistance(out double result)
+        {
+            // Time when we give up on looping and declare that reading failed
+            // 100ms was chosen because max measurement time for this sensor is around 24ms for 400cm
+            // additionally we need to account 60ms max delay.
+            // Rounding this up to a 100 in case of a context switch.
+            long hangTicks = Environment.TickCount + 100;
             _timer.Reset();
 
             // Measurements should be 60ms apart, in order to prevent trigger signal mixing with echo signal
@@ -66,6 +92,11 @@ namespace Iot.Device.Hcsr04
             // Wait until the echo pin is HIGH (that marks the beginning of the pulse length we want to measure)
             while (_controller.Read(_echo) == PinValue.Low)
             {
+                if (Environment.TickCount > hangTicks)
+                {
+                    result = default;
+                    return false;
+                }
             }
 
             _lastMeasurment = Environment.TickCount;
@@ -75,6 +106,11 @@ namespace Iot.Device.Hcsr04
             // Wait until the pin is LOW again, (that marks the end of the pulse we are measuring)
             while (_controller.Read(_echo) == PinValue.High)
             {
+                if (Environment.TickCount > hangTicks)
+                {
+                    result = default;
+                    return false;
+                }
             }
 
             _timer.Stop();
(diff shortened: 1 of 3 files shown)


src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -42,13 +42,39 @@ public Hcsr04(int triggerPin, int echoPin, PinNumberingScheme pinNumberingScheme
             _controller.OpenPin(_trigger, PinMode.Output);
 
             _controller.Write(_trigger, PinValue.Low);
+
```
pgrawehr: Not directly related to the bug, but mentioned in the issue as well: This ctor should have an overload that takes a GpioController instead of always creating its own.

src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -42,13 +42,39 @@ public Hcsr04(int triggerPin, int echoPin, PinNumberingScheme pinNumberingScheme
...
+            throw new InvalidOperationException("Could not get reading from the sensor");
+        }
+
+        private bool TryGetDistance(out double result)
+        {
+            // Time when we give up on looping and declare that reading failed
+            // 100ms was chosen because max measurement time for this sensor is around 24ms for 400cm
+            // additionally we need to account 60ms max delay.
+            // Rounding this up to a 100 in case of a context switch.
+            long hangTicks = Environment.TickCount + 100;
```
pgrawehr: `Environment.TickCount` is a 32 bit value that overflows after 23 or so days. So this will fail when the value overflows, causing the delay to be 0. Netcore 3.0 or later has a `TickCount64` property as a possible workaround, but I think we can't use that yet.

krwq: Good point, while this would retry it in that case and still work it's not a good solution. `Environment.TickCount - hangTicks > 0` seems to be working fine though since subtraction will "undo" overflowing

krwq:
I've also added protection against invalid readings (more than 400cm) and more retries since I was able to get exceptions by waving hand in front of the sensor (with 10 retries I wasn't able to)

src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -29,14 +29,14 @@ public class Hcsr04 : IDisposable
         /// <summary>
         /// Creates a new instance of the HC-SCR04 sonar.
         /// </summary>
+        /// <param name="gpioController">GPIO controller related with the pins</param>
         /// <param name="triggerPin">Trigger pulse input.</param>
         /// <param name="echoPin">Trigger pulse output.</param>
-        /// <param name="pinNumberingScheme">Pin Numbering Scheme</param>
-        public Hcsr04(int triggerPin, int echoPin, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical)
+        public Hcsr04(GpioController gpioController, int triggerPin, int echoPin)
```
pgrawehr: Now you need to keep track whether we created the controller ourselves or not, because like this, `Dispose()` will clean it up in either case.

krwq: @pgrawehr nah, per our guidelines ownership of GPIO controller belongs to this binding:
https://github.com/dotnet/iot/blob/master/Documentation/Devices-conventions.md#lifetime-management-or-what-should-be-disposed-by-what

if we want to keep track of this we should add extra argument `shouldDispose`

pgrawehr: I know, but then we need this parameter, because otherwise the usage of this is quite problematic still. If we (say) provide a Mcp23xxx as GpioController, we probably don't want it to be disposed when disposing the Hcsr (as there might be other devices still attached to the controller we want to still use or shutdown gracefuly as well using their bindings).

krwq: @pgrawehr there should be one instance of GpioController per device, it should not be shared.

pgrawehr: That only works as long as several GpioController instances can be created for the same physical device and Disposing one of them does not affect the others. This may work for the default Drivers (Raspi, SysFs), but will fail for things like a GpioController created from an MCP23017 (because we generally cannot create two instances of the same I2C device - the bindings would interfere with each other). Note also that with "One GpioController per device" approach, there's no way of blocking the user from doing stupid things like assigning the same pin to several devices at once.

Of course, there could be other ways of working around this issue, like providing Clone() operations that create "Slave" instances which internally reference the master component and do reference counting, but that seems a lot more complex, both in implementation and in usage. 

The `Board.CreateGpioController()`-Approach could be an option as well (then the concrete implementation on an MCP would probably just do nothing on Dispose as that is handled by the master board component). This would probably introduce a breaking change on MCP usage, though (or at least make the current usage variant deprecated).

krwq: @pgrawehr I'll leave this is as is for now and we should start some conversation in separate issue about this. If this is an issue for all devices we should fix it globally and adjust guidelines or fix MCP23xxx to implement GpioDriver correctly.

Do you mind creating an issue about this and creating any sample code which shows the specific issue?

My current understanding is that there should be a way of creating GpioController multiple times but if there is some architectural limitation (i.e. expanders using I2C or whatever) then we should fix the problem not the symptoms

pgrawehr: Ok. We clearly have a conceptual problem there. As stated above, there might be different solutions to it.
I'll collect information on the problem, write it up and then try to actually reproduce it. Might take some time tough, since I'm having quite a bunch of open ends right now (aka too many open branches). Also I'm missing a Socket to get my MCP properly attached to the breadboard and ordering will take some time.

krwq:
Merging this since @Kash0321 has confirmed this is fixing the issue he was seeing. I'll address any feedback separately if needed.

PR #920, opened 2019-12-31
Regenerate device listing

krwq:
I've joined `pwm`/`spi`/`i2c` categories into `protocol` since there were many things which fitted there but on their own they had too little things in there


krwq:
Merging this since this is relatively low risk. Please create an issue or send a PR if you find any issues or have any feedback

PR #926, opened 2020-01-05
New device binding: Pimoroni Explorer Hat Pro (ExplorerHat)

Kash0321:
I'm working with this hat and using dotnet/iot to build [robotics for kids](https://itclub.marcosfrechin.es/) with it. Now I'm trying to write the corresponding device binding library, and these are the first results.

I have limited knowledge of electronics, but I think this code is not bad... What do you think?

The device is [Pimoroni Explorer Hat](https://shop.pimoroni.com/products/explorer-hat)

They have a [python library](https://github.com/pimoroni/explorer-hat) and I intend to translate it into dotnet/iot...

Please feel free to correct, suggest and point out anything that is misimplemented... I would very much like to see it in the dotnet/iot repository, and I would keep working to complete the library


Kash0321:
I forget to mention that I'm not sure how to categorize the device in category.txt, so I left it blank

src/devices/ExplorerHat/ExplorerHat.cs
```diff
@@ -0,0 +1,85 @@
+﻿using System;
```
krwq: copyright header missing

src/devices/ExplorerHat/ExplorerHat.cs
```diff
@@ -0,0 +1,85 @@
...
+        #region IDisposable Support
+
+        private bool _disposedValue = false;
+
+        /// <summary>
+        /// Disposes the <see cref="ExplorerHat"/> instance
+        /// </summary>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
```
krwq: you can set `_controller` to null after disposing and checking for that instead of extra variable

src/devices/ExplorerHat/ExplorerHat.sln
```diff
@@ -0,0 +1,31 @@
+﻿
```
krwq: @joperezr will probably ask you to remove sln :smile:

src/devices/ExplorerHat/Gpio/GpioControllerExtensions.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Device.Gpio;
```
krwq: copyright header missing

src/devices/ExplorerHat/Lighting/Led.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
```
krwq: copyright header

src/devices/ExplorerHat/Lighting/Led.cs
```diff
@@ -0,0 +1,107 @@
...
+        public bool IsOn { get; set; }
+
+        /// <summary>
+        /// Initializes a <see cref="Led"/> instance
+        /// </summary>
+        /// <param name="number">Led number</param>
+        /// <param name="name">Led name</param>
+        /// <param name="pin">Underlying rpi GPIO pin number</param>
+        /// <param name="controller"><see cref="GpioController"/> used by <see cref="Led"/> to manage GPIO resources</param>
+        internal Led(int number, string name, int pin, GpioController controller)
```
krwq: make controller optional (controller = null by default) and then inside of the body of the method `_controller = controller ?? new GpioController()`

src/devices/ExplorerHat/Lighting/Led.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        /// <summary>
+        /// Switch on this led light
+        /// </summary>
+        public void On()
+        {
+            if (!IsOn)
+            {
+                _controller.EnsureOpenPin(Pin, PinMode.Output);
```
krwq: can you open all pins in the constructor, we normally don't late open pins since most of the time people don't replug during runtime anyway

krwq: also we never ensured pins are opened, we simply tried to open them, I'd follow this pattern in here as well

src/devices/ExplorerHat/Lighting/Lights.cs
```diff
@@ -0,0 +1,173 @@
+﻿using System;
```
krwq: copyright header here and everywhere else where it's missing

src/devices/ExplorerHat/Lighting/Lights.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        /// <summary>
+        /// Red led (#3)
+        /// </summary>
+        public Led Three { get => LedArray[2]; }
+
+        /// <summary>
+        /// Green led (#4)
+        /// </summary>
+        public Led Four { get => LedArray[3]; }
```
krwq: should this be `Color LedColor` instead? (haven't seen the device but guessing this is single RGB led)

src/devices/ExplorerHat/Lighting/Lights.cs
```diff
@@ -0,0 +1,173 @@
...
+                return LedArray[key];
+            }
+        }
+
+        /// <summary>
+        /// Gets the <see cref="Led"/> at the specified index
+        /// </summary>
+        /// <param name="key">The color-string-index (blue, yellow, red or green) of the led to get</param>
+        /// <returns>The <see cref="Led"/> at the specified index</returns>
+        public Led this[string key]
```
krwq: you should stick with one way of accessing LEDs, either property per LED or indexer or indexer by string but not all of them since this usually confuses users which one they should use

src/devices/ExplorerHat/Motorization/Motor.cs
```diff
@@ -0,0 +1,135 @@
...
+                else if (value < -1)
+                {
+                    speed = -1;
+                }
+                else
+                {
+                    speed = value;
+                }
+
+                _innerMotor.Speed = speed;
```
krwq: you should just forward this to DCMotor and if you think the default behavior is wrong then file issue on DCMotor, they should work consistently the same.

src/devices/ExplorerHat/Motorization/Motor.cs
```diff
@@ -0,0 +1,135 @@
+﻿using System;
+using System.Device.Gpio;
+
+namespace Iot.Device.ExplorerHat.Motorization
+{
+    /// <summary>
+    /// Represent one of the onboard motors
+    /// </summary>
+    public class Motor : IDisposable
```
krwq: IMO you should use DCMotor directly here since it doesn't seem like there would be any issues with that. If there is anything lacking in the DCMotor please file an issue instead of creating similar class

src/devices/ExplorerHat/Lighting/Led.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
+using System.Device.Gpio;
+using ExplorerHat.Gpio;
+
+namespace Iot.Device.ExplorerHat.Lighting
+{
+    /// <summary>
+    /// Represents a led light
+    /// </summary>
+    public class Led : IDisposable
```
krwq: would it make sense instead of this class have properties in the parent class with something like:

`Color LedColor` or `bool Led1On` etc

src/devices/ExplorerHat/Motorization/Motor.cs
```diff
@@ -0,0 +1,135 @@
...
+        /// Disposes the <see cref="Motor"/> instance
+        /// </summary>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
+            {
+                if (disposing)
+                {
+                    Stop();
+                    _innerMotor.Dispose();
```
krwq: please follow behavior of `_innerMotor` - if behavior is not as expected please file an issue on DCMotor (but even better I'd remove this class and just use it directly)

src/devices/ExplorerHat/README.md
```diff
@@ -0,0 +1,20 @@
+﻿# Explorer HAT Pro (Pimoroni)
+
+## Summary
+
+[Explorer HAT Pro](https://shop.pimoroni.com/products/explorer-hat) is an add-on board for Raspberry Pi.
+
+![Explorer HAT Pro](https://res.cloudinary.com/practicaldev/image/fetch/s--h3gOjXAH--/c_limit%2Cf_auto%2Cfl_progressive%2Cq_auto%2Cw_880/https://user-images.githubusercontent.com/10654401/63101233-e88c4b80-bf78-11e9-87ff-20e7a2809c40.png)
```
krwq: please upload this picture in here instead if possible (if license forbids then I'd recommend just making any picture), otherwise I'd remove the picture since this can be used by 3rd parties to track who opened this readme page which I doubt most of the users would like (I wouldn't)

Kash0321: I take the photo myself. It's my own rpi + explorer hat devices

src/devices/ExplorerHat/Lighting/Lights.cs
Kash0321: The hat has 4 leds, numbered from 1 to 4, and each one of a different color: blue (1), yellow (2), red (3), green (4). 

They are not rgb leds.

src/devices/ExplorerHat/Lighting/Lights.cs
Kash0321: You are right. I'll remove access by indexers.

src/devices/ExplorerHat/ExplorerHat.sln
Kash0321: I forgot to remove that...

src/devices/ExplorerHat/Lighting/Led.cs
Kash0321: This constructor is internal, ExplorerHat class manages GpioController lifecycle. My intend is to share only one instance of GpioController across all objects behind ExplorerHat class

src/devices/ExplorerHat/Lighting/Led.cs
Kash0321: Same as Lights.cs

src/devices/ExplorerHat/DCMotorExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        /// <summary>
+        /// Motor turns backwards at indicated speed
+        /// </summary>
+        /// <param name="motor">DCMotor instance to extend</param>
+        /// <param name="speed">Indicated speed</param>
+        public static void Backwards(this DCMotor.DCMotor motor, double speed = 1)
+        {
+            motor.Speed = Math.Abs(speed) * -1;
```
krwq: I'd prefer we just used `-speed`. I.e. for Forward we will still move backwards if negative speed is passed and that makes sense to me so for Backwards I'd expect it to work same as `Forwards(-speed)`

src/devices/ExplorerHat/ExplorerHat.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </summary>
+        public ExplorerHat(GpioController controller = null, bool shouldDispose = true)
+        {
+            _controller = controller;
+            _shouldDispose = shouldDispose;
+
+            if (_controller is null)
+            {
+                _controller = new GpioController(PinNumberingScheme.Logical);
+            }
```
krwq: these 3 lines and first line in this method can be simplified to just:

`_controller = controller ?? new GpioController();`

src/devices/ExplorerHat/ExplorerHat.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            if (_controller.NumberingScheme != PinNumberingScheme.Logical)
```
krwq: where is this coming from? We don't do similar checks elsewhere, except we provide way to override pins. I'd just remove the check and potentially in the future if needed we can provide ways to override pins

src/devices/ExplorerHat/Lighting/Lights.cs
krwq: fair enough

src/devices/ExplorerHat/README.md
krwq: Can you at least use just `https://user-images.githubusercontent.com/10654401/63101233-e88c4b80-bf78-11e9-87ff-20e7a2809c40.png` instead of the cloudinary proxy server? (I'm not getting reference to cloudinary when I upload the same way)

Kash0321: Yes, of course

src/devices/ExplorerHat/ExplorerHat.cs
Kash0321: I did it because receiving a controller initialized with NumberingScheme.Board, things doesn't works, but I like the "provide ways to override pins" concept.

src/devices/ExplorerHat/DCMotorExtensions.cs
Kash0321: Oh, I like backwards feature, I think that it helps working on two wheels rovers with kids... But I understand your intend to simplifies API. I'll remove it. Anyway, it can be implemented by clients. In that case, would you change the name Forwards by Turn, or Move, or ???...

krwq: I only meant to not do ABS, Backwards is fine

src/devices/ExplorerHat/ExplorerHat.cs
```diff
@@ -0,0 +1,76 @@
...
+                {
+                    Lights.Dispose();
+                    Motors.Dispose();
+
+                    if (_shouldDispose)
+                    {
+                        _controller.Dispose();
+                        _controller = null;
+                    }
+
```
krwq: nit: extra enter after brace

src/devices/ExplorerHat/GpioControllerExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Device.Gpio;
+
+namespace Iot.Device.ExplorerHat
+{
+    /// <summary>
+    /// Extension methods for <see cref="GpioController"/>
+    /// </summary>
+    public static class GpioControllerExtensions
```
krwq: seems this is not doing much, should probably be removed 😄

src/devices/ExplorerHat/Led.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        /// <summary>
+        /// GPIO pin to which led is attached
+        /// </summary>
+        public int Pin { get; private set; }
+
+        /// <summary>
+        /// Led number on Hat
+        /// </summary>
+        public int Number { get; private set; }
```
krwq: Is Number and Name needed for anything specific?

src/devices/ExplorerHat/Led.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <summary>
+        /// Led name on Hat
+        /// </summary>
+        public string Name { get; private set; }
+
+        /// <summary>
+        /// Gets if led is switched on or not
+        /// </summary>
+        /// <value></value>
+        public bool IsOn { get; set; }
```
krwq: This is settable and public which is confusing. I'd do one of the things below (pick whichever feels more intuitive):
- remove On/Off methods and put their body in this property (possibly also rename property), so effectively something along the lines of:

```csharp
private bool _on = false;
public bool On
{
    get
    {
        return _on;
    }
    set
    {
        _on = value;
        _controller.Write(Pin, value);
    }
}
```

- remove property (or make it private)

src/devices/ExplorerHat/category.txt
```diff
@@ -0,0 +1 @@
+
```
krwq: please remove this file or add categories (if you remove, I'll get a warning when regenerating device listing which will remind me to pick categories). If you leave it empty device won't be listed anywhere and I won't get a warning

src/devices/ExplorerHat/GpioControllerExtensions.cs
Kash0321: Sorry

src/devices/ExplorerHat/Led.cs
Kash0321: Not now, that's right. But I'm thinking about make class "Ligths" IEnumerable, so Name (would be renamed to Colour) and Number will help you to identify them when you randomly pick them up. What dou you think?

src/devices/ExplorerHat/Led.cs
Kash0321: I've made set private... See previous comment to consider if it's useful...

src/devices/ExplorerHat/Led.cs
krwq: @Kash0321 I'm a bit skeptical about adding stuff which is not immediately obvious what's for. I.e. imagine adding a "Name" property to Stream so that you know which stream is what.. Normally you'd create some kind of dictionary with (string => Stream) - while you can use properties I'd argue that should belong to the view model of your light and not to the hardware implementation.

src/devices/ExplorerHat/Led.cs
krwq: settable property is useful since you can do: `On = !On` but you can leave as is :)

src/devices/ExplorerHat/samples/ExplorerHat.BasicSample/ExplorerHat.BasicSample.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <ApplicationIcon />
+    <StartupObject />
```
krwq: what are those 2 above? Can you match csproj with other csproj-es (including the name)

krwq: to clarify when there is only one sample you should not create subfolder and name should be `ProjectName.Sample` - please refer to other projects. Same for the content of csproj

Kash0321: Done

src/devices/ExplorerHat/samples/ExplorerHat.Sample.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <ApplicationIcon />
```
krwq: Still seeing `<ApplicationIcon />` and `<StartupObject />`. Please match the content of this and other projects

Kash0321: I'm sorry, I didn't understand. It's done.

krwq:
Thank you @Kash0321!

Issue #927, opened 2020-01-06
Software PWM does not change frequency correctly


krwq:
Changing frequency from 10kHz => 20kHz using the property causes signal to disappear completely. Changing it back to 10kHz it's still off. When setting frequency directly in the constructor and not changing it everything works correctly.

PR #928, opened 2020-01-06
Fix SoftwarePwmChannel to update frequency correctly

krwq:
Fixes: https://github.com/dotnet/iot/issues/927

- rename some variables to make code more legible (i.e. `_period` was previously called `_pulseFreqency` or `_dutyCycle` was called `_percentage`)
- fix issue where changing frequency made PWM to stop working correctly

This has been tested under oscilloscope: 10kHz => 20kHz => 10kHz.

cc: @Ellerbach

### DIFF
--- a/src/devices/SoftPwm/SoftwarePwmChannel.cs
+++ b/src/devices/SoftPwm/SoftwarePwmChannel.cs
@@ -15,22 +15,20 @@ namespace System.Device.Pwm.Drivers
     public class SoftwarePwmChannel : PwmChannel
     {
         private readonly bool _shouldDispose;
-        // use to determine the freqncy of the PWM
-        // PulseFrequency = total frenquency
-        // curent pulse width = when the signal is hi
-        private double _currentPulseWidth;
-        private double _pulseFrequency;
+        // how long the signal is high in its period
+        private double _pulseWidthMs;
+        private double _periodMs;
         private int _frequency;
         // Use to determine the length of the pulse
-        // 100 % = full output. 0%= nothing as output
-        private double _percentage;
+        // 100% = 1.0 = full output. 0% = 0.0 nothing as output
+        private double _dutyCycle;
 
         // Determines if a high precision timer should be used.
-        private bool _usePrecisionTimer = false;
+        private bool _usePrecisionTimer;
 
         private bool _isRunning;
         private bool _isStopped = true;
-        private int _servoPin = -1;
+        private int _pin;
 
         private Stopwatch _stopwatch = Stopwatch.StartNew();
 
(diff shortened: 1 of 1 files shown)


src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -15,22 +15,20 @@ namespace System.Device.Pwm.Drivers
...
-        private double _currentPulseWidth;
-        private double _pulseFrequency;
+        // how long the signal is high in its period
+        private double _pulseWidthMs;
+        private double _periodMs;
         private int _frequency;
         // Use to determine the length of the pulse
-        // 100 % = full output. 0%= nothing as output
-        private double _percentage;
+        // 100% = 1.0 = full output. 0% = 0.0 nothing as output
```
krwq:
```suggestion
        // 100% = 1.0 = full output. 0% = 0.0 = nothing as output
```

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -15,22 +15,20 @@ namespace System.Device.Pwm.Drivers
...
-        private double _pulseFrequency;
+        // how long the signal is high in its period
+        private double _pulseWidthMs;
+        private double _periodMs;
         private int _frequency;
         // Use to determine the length of the pulse
-        // 100 % = full output. 0%= nothing as output
-        private double _percentage;
+        // 100% = 1.0 = full output. 0% = 0.0 = nothing as output
+        private double _dutyCycle;
```
tarekgh: I am not sure if this is a better name than the percentage.

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -15,22 +15,20 @@ namespace System.Device.Pwm.Drivers
...
     {
         private readonly bool _shouldDispose;
-        // use to determine the freqncy of the PWM
-        // PulseFrequency = total frenquency
-        // curent pulse width = when the signal is hi
-        private double _currentPulseWidth;
-        private double _pulseFrequency;
+        // how long the signal is high in its period
+        private double _pulseWidthMs;
+        private double _periodMs;
```
tarekgh: nit: I would name it as _periodInMilliseconds to be explicit.

tarekgh: name _pulseWidthMs same.

src/devices/SoftPwm/SoftwarePwmChannel.cs
krwq: we call argument and property DutyCycle (+/- casing) so this is just consistency

PR #940, opened 2020-01-14
Ft4222: Adding Linux support

Ellerbach:
- This PR add support for the FT4222 chipset for Linux platform
- Added documentation to explain what is required

### DIFF
--- a/src/devices/Ft4222/FtFunction.cs
+++ b/src/devices/Ft4222/FtFunction.cs
@@ -12,15 +12,22 @@ namespace Iot.Device.Ft4222
     /// </summary>
     internal class FtFunction
     {
-        #region ftd2xx.dll
+        #region ftd2xx wrapper
 
         /// <summary>
         /// Create Device Information List
         /// </summary>
         /// <param name="numdevs">number of devices</param>
         /// <returns>The status</returns>
-        [DllImport("ftd2xx")]
-        public static extern FtStatus FT_CreateDeviceInfoList(out uint numdevs);
+        public static FtStatus FT_CreateDeviceInfoList(out uint numdevs)
+        {
+            if (Environment.OSVersion.Platform == PlatformID.Unix)
+            {
+                return FT_CreateDeviceInfoListLinux(out numdevs);
+            }
+
+            return FT_CreateDeviceInfoListWin(out numdevs);
+        }
 
         /// <summary>
         /// Get Device Information Detail
@@ -34,8 +41,15 @@ namespace Iot.Device.Ft4222
         /// <param name="description">Description</param>
         /// <param name="ftHandle">Handle</param>
         /// <returns>The status</returns>
-        [DllImport("ftd2xx")]
-        public static extern FtStatus FT_GetDeviceInfoDetail(uint index, out uint flags, out FtDevice chiptype, out uint id, out uint locid, in byte serialnumber, in byte description, out IntPtr ftHandle);
+        public static FtStatus FT_GetDeviceInfoDetail(uint index, out uint flags, out FtDevice chiptype, out uint id, out uint locid, in byte serialnumber, in byte description, out IntPtr ftHandle)
+        {
+            if (Environment.OSVersion.Platform == PlatformID.Unix)
+            {
+                return FT_GetDeviceInfoDetailLinux(index, out flags, out chiptype, out id, out locid, in serialnumber, in description, out ftHandle);
+            }
+
+            return FT_GetDeviceInfoDetailWin(index, out flags, out chiptype, out id, out locid, in serialnumber, in description, out ftHandle);
+        }
 
         /// <summary>
         /// Open a device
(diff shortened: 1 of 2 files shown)


src/devices/Ft4222/FtFunction.cs
```diff
@@ -12,15 +12,133 @@ namespace Iot.Device.Ft4222
...
         /// <summary>
         /// Create Device Information List
         /// </summary>
         /// <param name="numdevs">number of devices</param>
         /// <returns>The status</returns>
-        [DllImport("ftd2xx")]
-        public static extern FtStatus FT_CreateDeviceInfoList(out uint numdevs);
+        public static FtStatus FT_CreateDeviceInfoList(out uint numdevs)
+        {
+            if (Environment.OSVersion.Platform == PlatformID.Unix)
```
krwq: optional: I'm wondering if you should check for Windows first since Unix implementation is more likely to work on any non-Windows platforms.

src/devices/Ft4222/README.md
```diff
@@ -25,6 +25,12 @@ You will need to unzip the file and go to ```LibFT4222-v1.4.2\imports\LibFT4222\
 
 Alternatively, you can register your dll globally. Run from the directory where your ```LibFT4222.dll``` is located the following command in administrator mode: ```regsvr32.exe LibFT4222.dll```
 
+## Linux Requirements
+
+For Linux, you need to download and install the proper version of ```libft4222.so```. The lib include the driver itself in a static binding. You can find the library [here](https://www.ftdichip.com/Support/SoftwareExamples/libft4222-linux-1.4.2.184.tgz). Once downloaded and extracted, run under administrator priviledges ```install4222.sh```. This will install the required library related to your system.
```
krwq:
```suggestion
For Linux, you need to download and install the proper version of ```libft4222.so```. The lib include the driver itself in a static binding. You can find the library [here](https://www.ftdichip.com/Support/SoftwareExamples/libft4222-linux-1.4.2.184.tgz). Once downloaded and extracted, run provided script ```install4222.sh``` as root. This will install the required library related to your system.
```

src/devices/Ft4222/FtFunction.cs
```diff
@@ -12,15 +12,133 @@ namespace Iot.Device.Ft4222
...
+        /// <param name="chiptype">Device type</param>
+        /// <param name="id">ID</param>
+        /// <param name="locid">Location ID</param>
+        /// <param name="serialnumber">Serial Number</param>
+        /// <param name="description">Description</param>
+        /// <param name="ftHandle">Handle</param>
+        /// <returns>The status</returns>
+        public static FtStatus FT_GetDeviceInfoDetail(uint index, out uint flags, out FtDevice chiptype, out uint id, out uint locid, in byte serialnumber, in byte description, out IntPtr ftHandle)
+        {
+            if (Environment.OSVersion.Platform == PlatformID.Unix)
```
joperezr: Ideally we would make this check only once and cache the result in a static or something.

src/devices/Ft4222/FtFunction.cs
```diff
@@ -12,15 +12,133 @@ namespace Iot.Device.Ft4222
...
+            {
+                return FT_CloseLinux(ftHandle);
+            }
+
+            return FT_CloseWin(ftHandle);
+        }
+
+        #endregion
+
+        #region Linux ftd2xx
```
joperezr: We usually try to keep the Interop methods on their own file

src/devices/Ft4222/FtFunction.cs
krwq: do you mean the delegate? I think enum comparison is more or less as fast as null check. Other option is to use interop interface and implement it per platform

src/devices/Ft4222/FtFunction.cs
Ellerbach: The thing is that this implementation should work for MacOS as well. As on Mac OS as on Windows, there there are 2 different libs.

src/devices/Ft4222/FtFunction.cs
Ellerbach: Those functions are called like 2 or 3 times maximum in all the life cycle of the code. So I won't try to over complicated everything. And they are called during the initialization phase.

src/devices/Ft4222/FtFunction.cs
Ellerbach: You mean doing 2 different files? In this case, it's just because on Linux, the drivers are statically linked to the ft4222 library while on Windows (and Mac) they are in 2 different libraries.

src/devices/Ft4222/FtFunction.cs
joperezr: I'm fine with keeping this as is.

src/devices/Ft4222/FtFunction.cs
joperezr: I meant doing it similar to System.Device.Gpio where we have special interop files for anything that will PInvoke, and the managed code is located on its own class. That said I don't have a big push on this since this is a device binding so we are less strict about this here, so I'm fine with leaving as is.

Issue #941, opened 2020-01-14
Flaky tests in CI (12% (now 3%) chance of failure)


krwq:
I've just run (Linux Release) CI tests 50 times in a row and out of 50 runs 6 failed with following failures:

5x (Fix pending https://github.com/dotnet/iot/pull/945):

```
    System.Device.Gpio.Tests.RaspberryPiDriverTests.InputPullResistorsWork [FAIL]
      Assert.Equal() Failure
      Expected: Low
      Actual:   High
      Stack Trace:
        /_/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.Linux.cs(34,0): at System.Device.Gpio.Tests.RaspberryPiDriverTests.InputPullResistorsWork()
```

1x:

```
    System.Device.Gpio.Tests.SysFsDriverTests.AddCallbackFallingEdgeNotDetectedTest [FAIL]
      Assert.False() Failure
      Expected: False
      Actual:   True
      Stack Trace:
        /_/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs(176,0): at System.Device.Gpio.Tests.GpioControllerTestBase.AddCallbackFallingEdgeNotDetectedTest()
```

cc: @pgrawehr @joperezr

We should try to understand the failures and if this is by design then retry or remove the test

pgrawehr:
Eventing will be way better with #914, but I couldn't get it 100% reliable as well.

krwq:
I'm generally fine with retries on SysFS eventing implementation since we understand there is a limitation but for the PullUp/PullDown I'd expect this to just work and I'm uncertain why the test is failing and I'm a bit concerned with hiding the issue. As a temporary solution retrying seems fine to me but we should create an issue for investigation

krwq:
I'm wondering if perhaps this delay isn't too small:
https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs#L277

And validate if spec
https://www.raspberrypi.org/app/uploads/2012/02/BCM2835-ARM-Peripherals.pdf Section "GPIO Pull-up/down Clock Registers (GPPUDCLKn)" have some extra info which perhaps we missed..

krwq:
Perhaps we should also not store the temporary value in the local since that is opening time window where something else might be fighting with the value of the register

pgrawehr:
[...]

PR #942, opened 2020-01-14
Adding Retry helper to InputPullResistorsWork test since it may fail sometimes

joperezr:
cc: @krwq @pgrawehr 

Contributes to #941 

This will try to mitigate CI failures by adding 5 retries in case this test fails until it ultimately fails the test and throws.

### DIFF
--- a/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.Linux.cs
+++ b/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.Linux.cs
@@ -27,15 +27,18 @@ namespace System.Device.Gpio.Tests
         [Fact]
         public void InputPullResistorsWork()
         {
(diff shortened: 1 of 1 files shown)


joperezr:
@pgrawehr One more strategy we could do here is to add more logging in order to try to understand what went wrong.

pgrawehr:
Yea, I've seen this test fail sometimes on my board, too. Since it _is_ flaky, I haven't figured out what fixes it. I first thought of introducing sleeps after changing the pull mode, but that didn't seem to be enough at times. I will try again.

krwq:
I'm fine with adding retries but do we have any idea what is causing the retries to be needed? Can we open issues about this along with this fix? Until we have full understanding why is this happening I'd prefer we kept the issue open.

I'd expect this scenario to work reliably to be honest and not sure why this is failing.

joperezr:
Logged #943 to track that.

PR #945, opened 2020-01-16
Fix Pull Up/Down intermittently not working

### LINKED ISSUE #943: Test InputPullResistorsWork will fail intermitently
During some sample test runs, @krwq discovered that this test will fail 12% of the time which is why as a mitigation we opted to do #942 which adds a few retries in case the test fails. This workitem is going to track the work of investigating what is wrong with that test and to provide a fix for it in order to make it more stable, or to fix the product if that is appropriate. 

cc @krwq @pgrawehr

krwq:
Fixes #943
Contributes to #941

Did couple of changes before all failures were gone. Out of 50 CI runs 1 failed and issue was unrelated (the other intermittent failure from #941).

Tested only on Release build.

When testing number of iterations in the test was increased to 10k

Changes made:
- added zeroing clock (this fixed most of the problems)
- operating directly on pointer
- despite spec order zeroing clock happens first since that seem to have causing intermittent issues as well
- SpinWait was replaced with regular loop and optimizations in that method were disabled - we can optimize that in the future but I wasn't able to get no intermittent failures without this also was paranoid that pointer operations might get optimized in some way to use local in some hidden away - not sure if that made any difference

cc: @pgrawehr

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -229,6 +229,12 @@ namespace System.Device.Gpio.Drivers
         [MethodImpl(MethodImplOptions.NoOptimization)]
         private void SetInputPullMode(int pinNumber, PinMode mode)
         {
+            /*
+             * NoOptimization is needed to force wait time to be at least minimum required cycles.
+             * Also to ensure that pointer operations optimizations won't be using any locals
+             * which would introduce time period where multiple threads could override value set
+             * to this register.
+             */
             if (IsPi4)
             {
                 SetInputPullModePi4(pinNumber, mode);
@@ -276,6 +282,9 @@ namespace System.Device.Gpio.Drivers
             for (int i = 0; i < 150; i++)
                 ;
 
+            // Spec calls to reset clock after the control signal
+            // Since context switch between those two instructions can potentially
+            // change pull up/down value we reset the clock first.
             *gppudclkPointer &= ~pinBit;
             *gppudPointer &= ~0b11U;
 
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -263,7 +263,6 @@ namespace System.Device.Gpio.Drivers
             *gppudPointer |= modeToPullMode;
 
             // Wait 150 cycles – this provides the required set-up time for the control signal
-            // Thread.SpinWait(150);
             for (int i = 0; i < 150; i++)
                 ;
 
@@ -274,7 +273,6 @@ namespace System.Device.Gpio.Drivers
             *gppudclkPointer |= pinBit;
 
             // Wait 150 cycles – this provides the required hold time for the control signal
-            // Thread.SpinWait(150);
             for (int i = 0; i < 150; i++)
                 ;
 
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -288,6 +288,9 @@ namespace System.Device.Gpio.Drivers
             *gppudclkPointer &= ~pinBit;
             *gppudPointer &= ~0b11U;
 
+            // This timeout is not documented in the spec
+            // but lack of it is causing intermittent failures when
+            // pull up/down is changed frequently.
             for (int i = 0; i < 150; i++)
                 ;
         }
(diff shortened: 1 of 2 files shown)


krwq:
@pgrawehr likely similar change has to be done on PI4 but I don't have a setup at the moment where I could test it

pgrawehr:
Thanks for investigating this. I'll play around with my Pi4 tonight, to see what fixes the issue there. The code does not currently contain any wait magic there, as setting the pulls is now similarly easy as setting the value. However, the official CPU register documentation for the Pi4 has still not been released, so there might be some wait required that's just undocumented.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -258,28 +259,28 @@ private void SetInputPullMode(int pinNumber, PinMode mode)
...
+                ;
 
-            register = *gppudPointer;
-            register &= ~0b11U;
-            *gppudPointer = register;
-            *gppudclkPointer = 0;
+            *gppudclkPointer &= ~pinBit;
+            *gppudPointer &= ~0b11U;
+
+            for (int i = 0; i < 150; i++)
```
joperezr: Can you please update the comment above the process to point out that we had to go against spec because that was the way to make it work? That way if we ever have to change this we will take that into consideration.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -226,6 +226,7 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
         /// </summary>
         /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
         /// <param name="mode">The mode of a pin to set the resistor pull up/down mode.</param>
+        [MethodImpl(MethodImplOptions.NoOptimization)]
```
joperezr: Can you put a comment on top of this option to mention why do we require noOptimization here?

joperezr:
Great find @krwq I'm glad you were able to fix this so fast :)

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: actually I'm not sure this wait is needed, I added this when testing - will double check...

pgrawehr:
Something is weird. I'm (correctly) getting StyleCop errors when building using VS, and also when I'm building `System.Devices.Gpio.sln` using dotnet build directly, but not when executing the batch.

pgrawehr:
There's a problem with the build configuration settings that make Stylecop not run by default. It seems that to fix that, I need to find out where and how `System.Device.Gpio.notsupported.cs` is generated (because that autogenerated file is not StyleCop compliant and fails to declare itself as autogenerated).

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: seems the delay might be needed - I'm getting 2/50 failures without it. Will double check my original results to make sure I wasn't just lucky with no failures

pgrawehr:
@krwq: Addressing the StyleCop stuff in #947, but please build once in VS to fix the issues you are about to introduce.
Note: Empty statements should be written as
```
for (...)
{
}
```

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -276,6 +282,9 @@ private void SetInputPullMode(int pinNumber, PinMode mode)
             for (int i = 0; i < 150; i++)
                 ;
 
+            // Spec calls to reset clock after the control signal
+            // Since context switch between those two instructions can potentially
+            // change pull up/down value we reset the clock first.
```
pgrawehr: Talking about context switches... What happens if two threads try to simultaneously change two (distinct) pins at the same time? Do we need a global lock around this method?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: 100/100 passed with this timeout, 2/50 failed without it

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
krwq: I've put it inside since the comment looked weird under the docs

krwq:
addressed all comments so will merge after CI is green. We should eventually investigate if we can simplify this logic but there is no pressure since this is not a hot path

PR #947, opened 2020-01-16
Remove special treatment of NotSupported stuff

pgrawehr:
It doesn't seem to be used anywhere and breaks Stylecop

### DIFF
--- a/eng/Analyzers.props
+++ b/eng/Analyzers.props
@@ -3,10 +3,10 @@
 <!-- If the runtime is unknown, GeneratePlatformNotSupportedAssembly is set in 
 System.Device.Gpio.csproj, which creates an auto-generated code 
 file which is not marked as such, making StyleCop fail. -->
-  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.csproj' AND '$(GeneratePlatformNotSupportedAssembly)' != 'true'">
+  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.csproj'">
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)\StyleCop.Analyzers.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
-  <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.csproj' AND '$(GeneratePlatformNotSupportedAssembly)' != 'true'">
+  <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.csproj'">
     <AdditionalFiles Include="$(MSBuildThisFileDirectory)\stylecop.json" Link="stylecop.json" />
     <PackageReference Include="StyleCop.Analyzers" Version="1.1.118">
       <PrivateAssets>all</PrivateAssets>
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10DriverPin.Windows.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10DriverPin.Windows.cs
@@ -70,7 +70,7 @@ namespace System.Device.Gpio.Drivers
 
             if (isFirstCallback)
             {
-                _pin.ValueChanged += this.Pin_ValueChanged;
+                _pin.ValueChanged += Pin_ValueChanged;
             }
         }
 


krwq:
just warning that Linux CI might take a while since I'm sending lots of jobs to test the intermittent issue :)

pgrawehr:
Ok, that's why ;-)
This _seems_ to work, however now the generated package contains a default System.Device.Gpio.dll file which is empty, but consists of NotSupported implementations in the official package. Unfortunatelly, I have no clue where to look how this NotSupported stuff is generated. 
@joperezr Can you help, please?
The following problem exists (see the changes in the commit)
- When the build.bat file is run, for some reason the condition in the targets file is not correctly handled, disabling Stylecop always, not only for the NotExisting.dll build
- When the condition is removed, the build fails with StyleCop errors in an auto-generated file named `System.Devices.Gpio.notexisting.cs` 
- I can't find anything about how this file is created,. Apparently, it should have the line `// <autogenerated/>` at the top, to avoid exactly this problem. 
- When building using `dotnet build` or VS, everything works as expected (so errors are reported)

src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -9,7 +9,6 @@
     <TargetRefPath>$(OutputPath)$(TargetFramework)\ref\$(MSBuildProjectName).dll</TargetRefPath>
     <IsPackable>true</IsPackable>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <GeneratePlatformNotSupportedAssembly Condition="'$(RuntimeIdentifier)' == ''">true</GeneratePlatformNotSupportedAssembly>
```
joperezr: We can't remove this. This needs to be in the package in case it is restored in a platform where we haven't provided any implementation for, for example, on OSX. There is currently a plan we have to remove the Windows and Linux specific builds and instead just keep one implementation which works in both places, which would make it so that we can remove this guy, but for now we need to keep it.

joperezr:
>When the build.bat file is run, for some reason the condition in the targets file is not correctly handled, disabling Stylecop always, not only for the NotExisting.dll build
When the condition is removed, the build fails with StyleCop errors in an auto-generated file named System.Devices.Gpio.notexisting.cs
I can't find anything about how this file is created,. Apparently, it should have the line // <autogenerated/> at the top, to avoid exactly this problem.
When building using dotnet build or VS, everything works as expected (so errors are reported)

I'm not sure I understand which build.bat are you refering too. In terms of the condition if you can give me a way to repro I'm happy to take a look and see why a condition is evaluating in some unexpeted way. The notsupported file gets generated as part of a call to a tool named GenAPI, and this file will be located in your intermediate directory.

Here is a brief explanation of System.Device.Gpio is built. The build has two different RuntimeIdentifiers defined (linux and windows). When you call msbuild on the Project without any properties, two inner builds will be dispatched for the same project but this time passing in a RuntimeIdentifier (one for RuntimeIdentifier=win and one for RuntimeIdentifier=linux). Once those two inner builds are done, we will then call this target:

https://github.com/dotnet/iot/blob/c9478593c199456a28cf562b6fb041698d525138/src/System.Device.Gpio/Directory.Build.targets#L20-L23

Which will basically just remove all of the files we need to compile, setup some special paths

https://github.com/dotnet/iot/blob/c9478593c199456a28cf562b6fb041698d525138/src/System.Device.Gpio/Directory.Build.targets#L2-L12

And then finally for this build we do setup that property you found  here

https://github.com/dotnet/iot/blob/c9478593c199456a28cf562b6fb041698d525138/src/System.Device.Gpio/System.Device.Gpio.csproj#L12

Which will only be set on the outer build because RuntimeIdentifier is not set. Setting this property will cause a new build to happen, using this tool https://github.com/dotnet/arcade/tree/master/src/Microsoft.DotNet.GenAPI to generate the not supported source for us, which will be the only thing we will compile given that we removed all of the Compile item previously.

joperezr:
If you are interested in how GenAPI produces the code, you can take a deeper look here https://github.com/dotnet/arcade/blob/bca68c2df2c6230389f10def3880a61db266c6c1/src/Microsoft.DotNet.GenAPI/build/Microsoft.DotNet.GenAPI.targets#L73-L104 for the target that runs and here https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.GenAPI/Program.cs for the code of the task that actually generates the code

pgrawehr:
@joperezr Thanks, that was really helpful. If https://github.com/dotnet/arcade/pull/4649 is accepted, it will be easily possible to fix the problem (by prepending `// <autogenerated />` to the file. Otherwise, more hacks would be required.

joperezr:
Awesome I think the PR you are trying to add over there makes a lot of sense, and keeping it optional is fine for now in order to not break other repos that depend on this already. We can revisit StyleCop for autogenerated files once that file comes through.

pgrawehr:
@joperezr Since the other PR is now merged, this can be merged and closed as well. It will finally correctly activate the StyleCop tests for all builds.

joperezr:
if it's ok by you I'll rebase latest master to trigger a new CI with the new arcade and make sure everything works as expected

joperezr:
oh sorry, looks like you already did that :)

PR #951, opened 2020-01-18
Introduce delay to make switching the pull mode reliable on Pi4

pgrawehr:
This should solve #945 for the Pi4. Before the change, the Pull Up/Down test would (with the retries currently in place) fail 100% of the time, after the change, I couldn't make it fail.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
@@ -304,8 +304,15 @@ namespace System.Device.Gpio.Drivers
         /// </summary>
         /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
         /// <param name="mode">The mode of a pin to set the resistor pull up/down mode.</param>
+        [MethodImpl(MethodImplOptions.NoOptimization)]
         private void SetInputPullModePi4(int pinNumber, PinMode mode)
         {
+            /*
+             * NoOptimization is needed to force wait time to be at least minimum required cycles.
+             * Also to ensure that pointer operations optimizations won't be using any locals
+             * which would introduce time period where multiple threads could override value set
+             * to this register.
+             */
             int shift = (pinNumber & 0xf) << 1;
             uint pull = 0;
             uint bits = 0;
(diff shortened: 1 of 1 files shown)


src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -321,6 +321,9 @@ private void SetInputPullModePi4(int pinNumber, PinMode mode)
             bits &= ~(3u << shift);
             bits |= (pull << shift);
             gpioReg->GPPUPPDN[(pinNumber >> 4)] = bits;
+            for (int i = 0; i < 150; i++)
```
krwq: you will need no optimizations to get that to work, otherwise this might get optimized away

pgrawehr: As you say it... I guess that's true. I will have to retest in release.

krwq: even if it works on release you can't guarantee this will not just stop working in the future or work differently on different runtime

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.Linux.cs
```diff
@@ -321,6 +328,9 @@ private void SetInputPullModePi4(int pinNumber, PinMode mode)
...
```
joperezr: Just curious, is this wait documented on the Pi4 datasheet? (assuming that is already available)

krwq: @joperezr it's not on Pi3, although it does mention to wait 150 cycles when setting so that was a guess it will be needed as well after unsetting and since we got no repro after doing the delay it's presumably correct

pgrawehr: @joperezr : The datasheet for the Pi4 (with the same in-depth-documentation as for the Pi3) does not exist yet. It has been announced since a long time, but it's still missing. [This page](https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2711/README.md) and the short PDF there is all there is for now.

joperezr: I see. I just asked because Ideally I would want to avoid to have workarounds like this as much as possible specially if they aren't documented on the datasheet. Because there is no datasheet for the Pi4 yet, and because we have seen improvements in tests with it, I'm fine with adding this for now and then hopefully once the datasheet is released we can see if this assumption was correct.

PR #958, opened 2020-01-24
Contributing: fix typos

MaherJendoubi:


src/System.Device.Gpio/System/Device/Gpio/WaitEventResult.cs
```diff
@@ -8,6 +8,6 @@ internal enum WaitEventResult
     {
         Error = -1,
         TimedOut = 0,
-        EventOccured = 1
+        EventOccurred = 1
```
joperezr: This is a breaking change, I realize that it was misspelled in the firstplace but this might be one case where we want to keep the original one for compat.

PR #964, opened 2020-01-30
Fix for pre-configured GPIOs

shuebener:
* GPIO "unexport" is not possible when the pin is configured by a device-tree overlay.
* The "edge" file only exists for input pins.
* Read pin value is also possible for output pins.
* Added new method to SysFsDriver.Linux to set the ActiveLow state of a GPIO pin.

-----------------------
By using device-tree overlays, GPIOs are all "exported" by the Linux system itself at startup and therefore cannot be "unexported" by writing to the corresponding files. Because of this, the File.WriteAllText at ClosePin() will throw an IOException.

The ClosePin() method calls the SetPinEventsToDetect() which tries to write to the "edge" file in order to disable any events. But this file only exists for Input pins and not for outputs (at least on my system, a Beaglebone Black). This will then of course lead to an FileNotFoundException.

Also to note, when using a device-tree overlay it is possible to disable the ability to change the direction of a pin, which then will of course also lead to an Exception when trying to use the SetPinMode() method.

-----------------------
I also added a new method "SetActiveLow()". Since I'm new to this project, I'm not sure if this is the correct way to implement a new feature. So feel free to remove this.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
@@ -118,7 +118,15 @@ namespace System.Device.Gpio.Drivers
                 try
                 {
                     SetPinEventsToDetect(pinNumber, PinEventTypes.None);
-                    File.WriteAllText(Path.Combine(GpioBasePath, "unexport"), pinOffset.ToString(CultureInfo.InvariantCulture));
+                    try
+                    {
+                        File.WriteAllText(Path.Combine(GpioBasePath, "unexport"), pinOffset.ToString(CultureInfo.InvariantCulture));
+                    }
+                    catch (IOException)
+                    {
+                        // Unexport failed -> pin is probably configured in device tree
+                    }
+
                     _exportedPins.Remove(pinNumber);
                 }
                 catch (UnauthorizedAccessException e)
@@ -146,6 +154,11 @@ namespace System.Device.Gpio.Drivers
             {
                 try
                 {
+                    if (File.ReadAllText(directionPath).Trim() == sysFsMode) // PIN direction already set
+                    {
+                        return;
+                    }
+
                     File.WriteAllText(directionPath, sysFsMode);
                 }
                 catch (UnauthorizedAccessException e)
@@ -302,15 +315,25 @@ namespace System.Device.Gpio.Drivers
         private void SetPinEventsToDetect(int pinNumber, PinEventTypes eventTypes)
         {
             string edgePath = Path.Combine(GpioBasePath, $"gpio{pinNumber + s_pinOffset}", "edge");
-            string stringValue = PinEventTypeToStringValue(eventTypes);
-            File.WriteAllText(edgePath, stringValue);
+            if (File.Exists(edgePath)) // Only exists for input pins!
+            {
+                string stringValue = PinEventTypeToStringValue(eventTypes);
+                File.WriteAllText(edgePath, stringValue);
+            }
         }
 
         private PinEventTypes GetPinEventsToDetect(int pinNumber)
         {
             string edgePath = Path.Combine(GpioBasePath, $"gpio{pinNumber + s_pinOffset}", "edge");
-            string stringValue = File.ReadAllText(edgePath);
-            return StringValueToPinEventType(stringValue);
+            if (File.Exists(edgePath)) // Only exists for input pins!
+            {
+                string stringValue = File.ReadAllText(edgePath);
+                return StringValueToPinEventType(stringValue);
+            }
+            else
+            {
+                return PinEventTypes.None;
+            }
         }
 
         private PinEventTypes StringValueToPinEventType(string value)
--- a/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
@@ -172,11 +172,6 @@ namespace System.Device.Gpio
                 throw new InvalidOperationException("Can not read from a pin that is not open.");
             }
 
-            if (_driver.GetPinMode(logicalPinNumber) == PinMode.Output)
-            {
-                throw new InvalidOperationException("Can not read from a pin that is set to Output mode.");
-            }
-
             return _driver.Read(logicalPinNumber);
         }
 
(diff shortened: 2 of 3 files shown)


pgrawehr:
By "device tree overlay" you mean something like having the lirc driver active on a certain pin? While I think it's probably ok to ignore the exception there, it is also probably not really a good idea to reuse such a pin, is it?

shuebener:
The device-tree is just a way to tell Linux, what hardware configuration you have. Here is a short example/description about this:
https://learn.adafruit.com/introduction-to-the-beaglebone-black-device-tree/overview

So for the GPIOs you simply define which pin should be an input or output. The result is the same, like when you do it manually by using the export files. So it's perfectly fine to use these pins in software however you like. The benefit of the DT is, that it is loaded very early in the boot process, so there is less chance of problems with the external hardware.

The Beaglebone even detects connected hardware capes by using a EEPROM on the board, to load the hardware configuration automatically.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -118,7 +118,15 @@ protected internal override void ClosePin(int pinNumber)
                 try
                 {
                     SetPinEventsToDetect(pinNumber, PinEventTypes.None);
-                    File.WriteAllText(Path.Combine(GpioBasePath, "unexport"), pinOffset.ToString(CultureInfo.InvariantCulture));
+                    try
```
krwq: can you please split PR into bug fixes from new APIs? otherwise if anything causes regression we will have to undo the whole change

shuebener: I created two new branches on my fork with the changes separated.
Is it possible to update this PR or do I have to create new PRs?

pgrawehr: You can update a PR by committing to the same branch again (you can force-push, if required). Create a new PR for the second change.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -172,11 +172,6 @@ public PinValue Read(int pinNumber)
                 throw new InvalidOperationException("Can not read from a pin that is not open.");
             }
 
-            if (_driver.GetPinMode(logicalPinNumber) == PinMode.Output)
```
joperezr: Will this work for all drivers? I believe the reason why we h ad this in the first place is because even though SysFS case would work, other drivers might not. We should test all drivers before going ahead and doing this change.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -146,6 +154,11 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
             {
                 try
                 {
+                    if (File.ReadAllText(directionPath).Trim() == sysFsMode) // PIN direction already set
```
joperezr: I'm not sure this extra check is really saving us anything, since you are still creating the stream and checking it's contents, which probably takes equivalent time than re-writing the value. On the other hand, in the case this value is different we would be doubling the time for that, so I don't think that this extra check is really worth it.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
```diff
@@ -302,15 +315,25 @@ protected internal override WaitForEventResult WaitForEvent(int pinNumber, PinEv
         private void SetPinEventsToDetect(int pinNumber, PinEventTypes eventTypes)
         {
             string edgePath = Path.Combine(GpioBasePath, $"gpio{pinNumber + s_pinOffset}", "edge");
-            string stringValue = PinEventTypeToStringValue(eventTypes);
-            File.WriteAllText(edgePath, stringValue);
+            if (File.Exists(edgePath)) // Only exists for input pins!
```
joperezr: Does it even make sense to allow detecting events for output pins? If not, then we should probably add this validation directly on GpioController so that no driver will allow this check.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
joperezr: Also, we have a test that is guarding the behavior that we will throw if attempting to read from an output pin, so if we decide to take this, then we would have to go and modify/remove that test.

pgrawehr: I also think we should leverage that. I haven't seen a case where reading from an output pin would be a problem.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
shuebener: For my case, writing to the direction fill will throw an exception, because the direction cannot be changed. I just added this check because the exception message "requires root" was a bit confusing for me when running the app as root. So it should also be fine to not add this check since it is not necessary to use this method/constructor.

```
# echo "in" > /sys/class/gpio/gpio48/direction
-bash: /sys/class/gpio/gpio48/direction: Permission denied
```

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.Linux.cs
shuebener: I don't know if there are gpio drivers around which will support this functionality. But at least for the Beaglebone linux distributions with SysFs I haven't seen it jet.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
shuebener: If there may be a driver not supporting this, I feel that it would make sense to throw this exception only from there instead of denying it for everyone.

krwq:
Hey @shuebener do you think you'll have time to finish up this PR? Please do split fixes in this PR from new APIs

krwq:
@shuebener ping

joperezr:
[Triage] PR seems abandoned. We will close in our next triage meeting if we don't see any activity.

PR #967, opened 2020-02-03
LibGpiodDriver: Use process name as consumer name

microhobby:
The pin number that are in use we already have by the gpiolib interface.
For purposes of description, it is more valuable to have the process
name, that are using the resource, as consumer name.

This way we have the info about who are using the pin.

Signed-off-by: Matheus Castello <matheus@castello.eng.br>

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
@@ -19,6 +19,8 @@ namespace System.Device.Gpio.Drivers
 
         protected internal override int PinCount => Interop.libgpiod.gpiod_chip_num_lines(_chip);
 
+        private string consumerName = System.Diagnostics.Process.GetCurrentProcess().ProcessName;
+
         public LibGpiodDriver(int gpioChip = 0)
         {
             try
@@ -157,14 +159,13 @@ namespace System.Device.Gpio.Drivers
             int requestResult = -1;
             if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle pinHandle))
             {
-                string consumer = pinNumber.ToString();
                 if (mode == PinMode.Input)
                 {
-                    requestResult = Interop.libgpiod.gpiod_line_request_input(pinHandle, consumer);
+                    requestResult = Interop.libgpiod.gpiod_line_request_input(pinHandle, consumerName);
                 }
                 else
                 {
-                    requestResult = Interop.libgpiod.gpiod_line_request_output(pinHandle, consumer);
+                    requestResult = Interop.libgpiod.gpiod_line_request_output(pinHandle, consumerName);
                 }
 
                 pinHandle.PinMode = mode;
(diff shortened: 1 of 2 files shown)


microhobby:
Detailed:

The pin number that are in use we already have by the gpiolib interface.
dotnetiotConsumerPinNumber

For purposes of description, it is more valuable to have the process
name, that are using the resource, as consumer name.
dotnetIoTLibpgiodUsed

This way we have the info about who are using the pin.
dotnetIotDemo4RunningHtop

For example, developers of the libgpiod use the program name as consumer name in their command line utilities:
https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/tree/tools/gpioget.c#n101

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -19,6 +19,8 @@ public class LibGpiodDriver : UnixDriver
 
         protected internal override int PinCount => Interop.libgpiod.gpiod_chip_num_lines(_chip);
 
+        private string consumerName = System.Diagnostics.Process.GetCurrentProcess().ProcessName;
```
krwq: - this can be static
- naming convention for static is: `s_consumerName` for instance fields: `_consumerName`

src/System.Device.Gpio/System/Device/Gpio/LibgpiodDriverEventHandler.Linux.cs
```diff
@@ -32,7 +34,7 @@ public LibGpiodDriverEventHandler(int pinNumber, SafeLineHandle safeLineHandle)
 
         private void SubscribeForEvent(SafeLineHandle pinHandle)
         {
-            int eventSuccess = Interop.libgpiod.gpiod_line_request_both_edges_events(pinHandle, $"Listen {_pinNumber} for both edge event");
+            int eventSuccess = Interop.libgpiod.gpiod_line_request_both_edges_events(pinHandle, consumerName);
```
krwq: I'm wondering if this should be some combination of old and new version. cc: @joperezr

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Thanks for the review @krwq

src/System.Device.Gpio/System/Device/Gpio/LibgpiodDriverEventHandler.Linux.cs
microhobby: In my opinion this "listening ..." information is useless and somewhat replicated. First, we are not using the line info internally for any purpose. For the description information of the line info, it is already described if the pin is as input or output, there is no need to say it again in the consumer tag.

joperezr: I think I agree with @microhobby, I don't think that the info is all that useful and simply having the process name might be enough, plus as @microhobby pointed out, it does follow what libgpiod guys use for their utilities. I'm fine with keeping this just with the consumer name for now, and then if somebody asks for the more descriptive message then we can have a discussion on whether or not to add it back.

microhobby: @krwq if you have concerns about this, let me know what you think.

buyaa-n: Agree with @joperezr, plus looks more clean :), thank you for demonstration @microhobby

joperezr:
Thanks for helping with this @microhobby!

Issue #410, opened 2019-05-08
LibGpiod: Input Pull Down / Pull Up Support


BWalti:
The LibGpiodDriver currently only supports "PinMode.Input" as input variant. InputPullDown and InputPullUp are not supported: https://github.com/dotnet/iot/blob/733296b63311104122b49bf1e2c247c7bd690391/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs#L100-L104

Wouldn't it be possible to use the corresponding flags and pass them through?

Offending code:
https://github.com/dotnet/iot/blob/733296b63311104122b49bf1e2c247c7bd690391/src/Native/System.Device.Gpio.Native/libgpiod.c#L53-L56

A test in another library showing, that it could be possible:
https://github.com/brgl/libgpiod/blob/2e2ca25d49c2e1b263009ae8f970871dd45af369/tests/tests-line.c#L472-L473
?

EDIT: as @buyaa-n pointed out, the above referenced code was not about pulling up / down. But there is a corresponding ticket indeed: [brgl/libgpiod#31](https://github.com/brgl/libgpiod/issues/31). END EDIT

I had to switch back to Windows 10 IoT as there PullDown is supported (active high). Otherwise I would have to do some soldering :)

PS: I guess the original library now lies in git.kernel.org:
https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/tree/include/gpiod.h?h=v1.2.x#n870

shaggygi:
I think there are other checks that would also need to be changed if/when #135 is added, as well.

BWalti:
I'm not an expert on that field, but in my understanding #135 is about output lines?
Maybe I'm wrong, but I guessed that passing the flag would cause the internal resistors to be correctly applied. I would not expect other side effects. But I'm probably naive :)

shaggygi:
No, you are correct.  My point is more about we will need to check for the different types of inputs and outputs.  This includes for both dev boards and device bindings that implement GpioController (e.g. Mcp23xxxx, etc.).

BWalti:
I did not think about other boards. My bad! I guess not all boards/chips with libgpiod support have PullUp / PullDown internal resistors?
I see, maybe its bigger to cleanly solve than I first thought..

shaggygi:
[...]

PR #968, opened 2020-02-03
Add support for libgpiod bias pin configuration

microhobby:
libgpiod v1.5 now has abstractions to request pins with bias flags: pull-up and pull-down.
For use the bias configurations with kernel gpiolib interfaces we need Kernel v5.5 or later.

News about the libgpiod v1.5:
https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/tree/NEWS#n13

Fixes https://github.com/dotnet/iot/issues/410


src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -103,8 +113,30 @@ protected internal override PinMode GetPinMode(int pinNumber)
...
         {
-            // Libgpiod Api do not support pull up or pull down resistors for now.
-            return mode != PinMode.InputPullDown && mode != PinMode.InputPullUp;
+            switch (mode)
+            {
+                case PinMode.Input:
+                    return true;
+                case PinMode.InputPullDown:
+                case PinMode.InputPullUp:
+                    // for use the bias flags we need Kernel version 5.5 or later
```
joperezr: Correct me if I'm wrong but I believe we would also need libgpiod version to be 1.5 correct? That one will be a bit harder to check, but we could try calling the new PInvoke inside a try catch to see if the method is present on the current installation of libgpiod.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -19,6 +20,15 @@ public class LibGpiodDriver : UnixDriver
 
         protected internal override int PinCount => Interop.libgpiod.gpiod_chip_num_lines(_chip);
 
+        public enum RequestFlag : ulong {
```
joperezr: This should probably be internal or private.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -103,8 +113,30 @@ protected internal override PinMode GetPinMode(int pinNumber)
...
-            // Libgpiod Api do not support pull up or pull down resistors for now.
-            return mode != PinMode.InputPullDown && mode != PinMode.InputPullUp;
+            switch (mode)
+            {
+                case PinMode.Input:
+                    return true;
+                case PinMode.InputPullDown:
+                case PinMode.InputPullUp:
+                    // for use the bias flags we need Kernel version 5.5 or later
+                    var linuxVersionMatch = Regex.Split(RuntimeInformation.OSDescription, @"Linux (.*?)\.(.*?)\.");
```
joperezr: Would it be possible to try to avoid using Regex in this case and instead use string manipulation? Most of the time using Regex leads to performance bottlenecks and potential security issues, so if we can try using string manipulation instead I think it would be better.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -157,14 +189,25 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
...
+
+                switch (mode)
                 {
-                    requestResult = Interop.libgpiod.gpiod_line_request_output(pinHandle, consumer);
+                    case PinMode.Input:
+                        requestResult = Interop.libgpiod.gpiod_line_request_input(pinHandle, consumer);
+                        break;
+                    case PinMode.InputPullDown:
+                        flags = (int) RequestFlag.GPIOD_LINE_REQUEST_FLAG_BIAS_PULL_DOWN;
+                        requestResult = Interop.libgpiod.gpiod_line_request_input_flags(pinHandle, consumer, flags);
```
joperezr: Same comment as above. I assume this requires 1.5 for libgpiod, so in that case you probably want to add this in a try catch block as it is likely that the method won't bind and throw an exception here.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Yes, correct. What do you think, keep the check for the kernel, removing the Regex, or just the check for bind libgpiod?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Yes ok, my doubt is if we check this early on the IsPinModeSupported

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
joperezr: Well, the idea of this method is to return whether or not a pin mode is supported on the current hardware, so in order for us to tell the truth here, we would have to make both checks, that is to ensure Kernel is 5.5 plus also ensure libgpiod version is recent enough.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
joperezr: If we add this check in IsPinModeSupported then we theoretically won't need to add it again here (although wrapping the native call might still be helpful for other reasons) since when you call on GpioController.SetPinMode, it will first callinto IsPinModeSupported, and will call this driver method only if that previous call returned true.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Thanks for the review. I fix it on 5344c80

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Thanks for the review. I fix it on d423507

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: The libpgiod has a function that return the version, so I thought it better to use it in checks. I add the binding on 47ff809

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Thanks for review. I'm making all the versions checks on the IsPinModeSupported.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Thanks for review @joperezr . I also added the check for the libgpiod in IsPinModeSupported d28e719. Let me know what you think.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -103,8 +112,48 @@ protected internal override PinMode GetPinMode(int pinNumber)
...
-            return mode != PinMode.InputPullDown && mode != PinMode.InputPullUp;
+            switch (mode)
+            {
+                case PinMode.Input:
+                    return true;
+                case PinMode.InputPullDown:
+                case PinMode.InputPullUp:
+                    // for use the bias flags we need libgpiod version 1.5 or later
+                    IntPtr libgpiodVersionPtr = Interop.libgpiod.gpiod_version_string();
+                    var libgpiodVersionMatch = Marshal.PtrToStringAnsi(libgpiodVersionPtr).Trim().Split('.');
```
krwq: Is Trim actually needed? Can https://docs.microsoft.com/en-us/dotnet/api/system.version.-ctor?view=netframework-4.8#System_Version__ctor_System_String_ be used instead?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -103,8 +112,48 @@ protected internal override PinMode GetPinMode(int pinNumber)
...
+                    bool isLibgpiod1dot5 = false;
+
+                    if (libgpiodVersionMatch.Length > 1)
+                    {
+                        libgpiodVersion = float.Parse($"{libgpiodVersionMatch[0]}.{libgpiodVersionMatch[1]}");
+                        isLibgpiod1dot5 = (libgpiodVersion >= 1.5);
+                    }
+
+                    // for use the bias flags we need Kernel version 5.5 or later
+                    var linuxVersionMatch = RuntimeInformation.OSDescription.Replace("Linux ", "").Trim().Split('.');
```
krwq: Can we re-use or copy following file to get the version instead? https://github.com/dotnet/runtime/blob/master/src/libraries/Common/tests/CoreFx.Private.TestUtilities/System/PlatformDetection.Unix.cs

microhobby: Hi @krwq, thanks for the review. I don't think so, this appears to check the version of the Distro and not the Kernel version. Here we have to check the Kernel version. If I'm wrong or if there is another way to get this kernel version more easily let me know.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Thank you for the review @krwq. Thank you very much for the tip, this simplified the code. I am sending a new commit using `Version`, let me know what you think.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -105,8 +115,46 @@ protected internal override PinMode GetPinMode(int pinNumber)
...
+                case PinMode.InputPullUp:
+                    // for use the bias flags we need libgpiod version 1.5 or later
+                    IntPtr libgpiodVersionPtr = Interop.libgpiod.gpiod_version_string();
+                    string libgpiodVersionMatch = Marshal.PtrToStringAnsi(libgpiodVersionPtr);
+                    Version libgpiodVersion = new Version(libgpiodVersionMatch);
+                    bool isLibgpiod1dot5 = (libgpiodVersion.Major >= 1 && libgpiodVersion.Minor >= 5);
+
+                    // for use the bias flags we need Kernel version 5.5 or later
+                    string[] linuxVersionMatch = RuntimeInformation.OSDescription.Replace("Linux ", string.Empty).Split('.');
+                    Version linuxVersion = new Version();
```
krwq: would it make sense to do something like: `Version linuxVersion = (linuxVersionMatch.Length > 1) ? ... : null;`
and then assign `bool isKernelLinux5dot5 = linuxVersion != null && ...;` to avoid extra allocations?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -105,8 +115,46 @@ protected internal override PinMode GetPinMode(int pinNumber)
...
+                    }
+
+                    // check if we have the correct versions
+                    if (isKernelLinux5dot5 && isLibgpiod1dot5)
+                    {
+                        return true;
+                    }
+                    else
+                    {
+                        Console.Error.WriteLine($"Using Kernel v{linuxVersion} but v5.5 or later is required.\n" +
```
krwq: I don't believe `IsPinModeSupported` should be printing anything since it's used for probing. You can move this logic to separate method and then whenever this mode is trying to be used only then improve current error message.

Also the linux kernel version can be checked only once:
create a static method for probing: `IsLinuxKernelVersion1_5OrHigher` and then assign this to a static field. Note this will be lazy initialized on the first usage of this class.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -105,8 +115,46 @@ protected internal override PinMode GetPinMode(int pinNumber)
...
+                case PinMode.InputPullDown:
+                case PinMode.InputPullUp:
+                    // for use the bias flags we need libgpiod version 1.5 or later
+                    IntPtr libgpiodVersionPtr = Interop.libgpiod.gpiod_version_string();
+                    string libgpiodVersionMatch = Marshal.PtrToStringAnsi(libgpiodVersionPtr);
+                    Version libgpiodVersion = new Version(libgpiodVersionMatch);
+                    bool isLibgpiod1dot5 = (libgpiodVersion.Major >= 1 && libgpiodVersion.Minor >= 5);
+
+                    // for use the bias flags we need Kernel version 5.5 or later
+                    string[] linuxVersionMatch = RuntimeInformation.OSDescription.Replace("Linux ", string.Empty).Split('.');
```
krwq: Note that you should also check this string actually contains `Linux` and not something completely different to avoid any weird behaviors in case something changes in the future.

microhobby: Hi @krwq , thanks for the review and tips. I am sorry for the delay in working on it, the quarantine for the COVID-19 started to take effect this week here in the state where I live in Brazil. It was a chaotic week. I hope to be able to work on it this weekend. I hope you are safe. Thanks.

buyaa-n: Thank you @microhobby for your contribution! 

Question: do we really need to check both libgpiod version and Linux kernel? Could someone install libgpiod  1.5 on linux with kernel lower than 5.5? Isn't enough to check only libgpiod version installed?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -159,13 +207,24 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
...
+                int flags;
+
+                switch (mode)
                 {
-                    requestResult = Interop.libgpiod.gpiod_line_request_output(pinHandle, s_consumerName);
+                    case PinMode.Input:
+                        requestResult = Interop.libgpiod.gpiod_line_request_input(pinHandle, s_consumerName);
+                        break;
+                    case PinMode.InputPullDown:
+                        flags = (int)RequestFlag.GPIOD_LINE_REQUEST_FLAG_BIAS_PULL_DOWN;
```
buyaa-n: Doesn't seem to need this `flag` variable, for me better to just use the enum below inline

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
buyaa-n: I have tried install Libgpoid 1.5 on Ubuntu with kernel version 4.4 and got this error: 
`configure: error: "libgpiod needs linux headers version >= v5.5.0"`. Which means we don't need to check kernel version in addition to Libgpoid version, please remove kernel version check

microhobby: Hi @buyaa-n, thanks for the review and checks.

The problem I saw, when I first wrote the patches, was in the case of a cross-compile, I can just use headers from Linux repo and cross compile 1.5 on my machine with no problems. For example in my tests I can use libgpiod 1.5 with an old Kernel version which returns the following error: 

```
Unhandled exception. System.IO.IOException: Error setting pin mode, for pin: 23, error: 22
   at System.Device.Gpio.Drivers.LibGpiodDriver.SetPinMode(Int32 pinNumber, PinMode mode)
``` 
The kernel does not understand the argument that libpgiod sends and then returns an error 22 (Invalid argument). 

The initial idea was to make it more verbose in error handling that kernel 5.5 or later should be used together with libgpiod 1.5. But with your reviews I start to think that it can be adding complexity without much value, this is a tool for developers so, I think that just checking for libgpiod 1.5 or later is enough.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: Thanks @buyaa-n fixed

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
microhobby: I sent two new commits removing the kernel check and improving the libgpiod version check with @krwq suggestions. I wish you are safe and healthy, thanks.

buyaa-n: Great, thank you for doing this @microhobby, hope you are safe too

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -133,6 +143,13 @@ internal partial class libgpiod
         /// <returns>GPIO chip pointer handle or NULL if an error occurred.</returns>
         [DllImport(LibgpiodLibrary, SetLastError = true)]
         internal static extern SafeChipHandle gpiod_chip_open_by_number(int number);
+
+        /// <summary>
+        /// Get the API version of the library as a human-readable string.
+        /// </summary>
+        /// <returns>Human-readable string containing the library version.</returns>
+        [DllImport(LibgpiodLibrary, SetLastError = true)]
+        internal static extern IntPtr gpiod_version_string();
```
joperezr: I assume this one is available in older versions of libgpiod as well?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.Linux.cs
```diff
@@ -21,6 +21,28 @@ public class LibGpiodDriver : UnixDriver
 
         private static string s_consumerName = System.Diagnostics.Process.GetCurrentProcess().ProcessName;
 
+        // for use the bias flags we need libgpiod version 1.5 or later
+        private static bool IsLibgpiodVersion1_5orHigher()
+        {
+            IntPtr libgpiodVersionPtr = Interop.libgpiod.gpiod_version_string();
```
joperezr: If we can validate that this method is also available in older versions of libgpiod that is perfect. If not, we could just add a try catch and in case the PInvoke is unsuccesful or doesn't bind we can just return false and assume it is not 1.5 or later.

joperezr:
Sorry for taking so long to review this @microhobby, I just asked a small question around whether or not the version PInvoke is available in older versions, but other than that this LGTM. Do you mind also rebasing as there seems to be a conflict on your PR?

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
buyaa-n: I see that method in version 0.2.x 😄 :https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/tree/include/gpiod.h?h=v0.2.x#n1094

joperezr: awesome, thanks for validating @buyaa-n. In that case let's go ahead and merge this in 😄

microhobby: Thanks for the check @buyaa-n! Thanks for the review @joperezr.
I did the merge with master and resolved the conflict, forgive me for the delay.
Stay safe!

buyaa-n: Thank you @microhobby, merging the PR, stay safe!

PR #986, opened 2020-02-20
Fixes #183: Allow reading output pins

### LINKED ISSUE #183: Cannot read from output pin?

stevedesmond-ca:
Fixes #183
Per #183, this PR removes the arbitrary restriction on `GpioController`'s ability to `Read()` from a pin currently set to `PinMode.Output`.

I took the path that @buyaa-n's [comment](https://github.com/dotnet/iot/issues/183#issuecomment-467940757) suggested, of moving the appropriate place for handling this to the driver level.

That said, this PR does not add this restriction to any particular drivers, as the only devices I have to test with (Raspberry Pi v2 and v3) both work as expected with the ability to read pin values when their `direction` is set to `out`.

EDIT (@krwq): added "Fixes #183" so that github automatically closes the issue after merging

### DIFF
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -116,6 +116,16 @@ namespace System.Device.Gpio.Tests
         }
 
         [Fact]
+        public void CanReadFromOutputPin()
+        {
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+            {
+                controller.OpenPin(OutputPin, PinMode.Output);
+                Assert.Equal(PinValue.Low, controller.Read(OutputPin));
+            }
+        }
+
+        [Fact]
         [Trait("SkipOnTestRun", "Windows_NT")] // Currently, the Windows Driver is defaulting to InputPullDown instead of Input when Closed/Opened.
         public void OpenPinDefaultsModeToInput()
         {
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -121,7 +121,12 @@ namespace System.Device.Gpio.Tests
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 controller.OpenPin(OutputPin, PinMode.Output);
+
+                controller.Write(OutputPin, PinValue.Low);
                 Assert.Equal(PinValue.Low, controller.Read(OutputPin));
+
+                controller.Write(OutputPin, PinValue.High);
+                Assert.Equal(PinValue.High, controller.Read(OutputPin));
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -115,16 +115,6 @@ public void ThrowsIfWritingOnInputPin()
             }
         }
 
-        [Fact]
```
krwq: can you please replace this with test testing the scenario you're fixing?

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -116,12 +116,12 @@ public void ThrowsIfWritingOnInputPin()
...
 
         [Fact]
-        public void ThrowsIfReadingFromOutputPin()
+        public void CanReadFromOutputPin()
         {
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 controller.OpenPin(OutputPin, PinMode.Output);
-                Assert.Throws<InvalidOperationException>(() => controller.Read(OutputPin));
+                Assert.Equal(PinValue.Low, controller.Read(OutputPin));
```
krwq: please also `gpioController.Write` and assert that value has changed afterwards and we can merge like that

krwq: seems this is incidental that this value is "Low". On Windows it seems to be "High", please pass value to OpenPin to get stable behavior. cc: @joperezr (double check this is not a bug)

stevedesmond-ca: :sweat_smile: Yep, forgot about Windows and Linux not doing the same thing here. I believe it's expected behavior, and there's already a comment about a similar side-effect [on line 21](https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs#L21).

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -121,7 +121,12 @@ public void CanReadFromOutputPin()
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 controller.OpenPin(OutputPin, PinMode.Output);
```
krwq: optional: there is an overload to open pin and immediately write to it

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
krwq: Perhaps something we can iterate on separately if anyone complains

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
krwq: actually I don't think there is one

joperezr:
Thanks a lot @stevedesmond-ca!

PR #990, opened 2020-02-25
Fix bug #884 in TimeEnvelope of samples/led-more-blinking-lights

### LINKED ISSUE #884: TimeEnvelope.IsLastMultiple in example is incorrect.
**Describe the bug**

**IsLastMultiple** should return true when on last increment of the passed value.  However, it returns false.  The current conditions checks if **_time - value == 0**, this will actually return true on the First Multiple. The condition should be **_count - value == _time**.  

For example, if _count is set to 1000 and value is 200 then IsLastMultiple should return true when _time is 800.   The current logic returns false because  800 - 200 != 0, however, 1000 - 200 == 800.

**Steps to reproduce**

```csharp
var te = new TimeEnvelope(1000);
var multiple = 0;
te.AddTime(100);
while(te.Time != 0)  {
  if (te.IsMultiple(200)) {
     multiple++;
    Console.Writeline($"Multiple Hit {multiple}");
  }   
  if (te.IsFirstMultiple(200)) {
     Console.Writeline("First Multiple Hit");
  }
  if (te.IsLastMultiple(200)) {
    Console.Writeline("Last Multiple Hit");
  }
  te.AddTime(100);
}
```

**Expected behavior**

```
Multiple Hit (1)             // at 200
First Multiple Hit          // at 200
Multiple Hit (2)             // at 400
Multiple Hit (3)             // at 600
Multiple Hit (4)             // at 800 
Last Multiple Hit           // at 800
```

**Actual behavior**

```
Multiple Hit (1)             // at 200
First Multiple Hit          // at 200
Last Multiple Hit           // at 200
Multiple Hit (2)             // at 400
Multiple Hit (3)             // at 600
Multiple Hit (4)             // at 800
```

mtripsky:
Fixes #884

- the method IsLastMultiple(int value) did not use _count to correctly decide when the value is last multiple. 
-  simplified all methods returning bool, instead of returning true/false, it returns directly result of if check.
- it was tested against code snippet included in the bug description to test expected behavior.
```csharp
var te = new TimeEnvelope(1000);
var multiple = 0;
te.AddTime(100);
while(te.Time != 0)  {
  if (te.IsMultiple(200)) {
     multiple++;
    Console.Writeline($"Multiple Hit {multiple}");
  }   
  if (te.IsFirstMultiple(200)) {
     Console.Writeline("First Multiple Hit");
  }
  if (te.IsLastMultiple(200)) {
    Console.Writeline("Last Multiple Hit");
  }
  te.AddTime(100);
}
```
ScreenShot of expected behaviour after my changes:
Screenshot 2020-02-25 at 12 29 46

ScreenShot of the original behaviour before my changes:
Screenshot 2020-02-25 at 12 58 57

### DIFF
--- a/samples/led-more-blinking-lights/TimeEnvelope.cs
+++ b/samples/led-more-blinking-lights/TimeEnvelope.cs
@@ -58,31 +58,16 @@ internal class TimeEnvelope
 
     public bool IsFirstMultiple(int value)
     {
-        if (_time == value)
-        {
-            return true;
-        }
-
-        return false;
+        return _time == value;
     }
 
     public bool IsLastMultiple(int value)
     {
-        if (_time - value == 0)
-        {
-            return true;
-        }
-
-        return false;
+        return _count - value == _time;
     }
 
     public bool IsMultiple(int value)
     {
-        if (_time % value == 0)
-        {
-            return true;
-        }
-
-        return false;
+        return _time % value == 0;
     }
 }


mtripsky:
Could somebody explain my why some of the checks are failing when I changed only one file with no impact on the library?

joperezr:
> Could somebody explain my why some of the checks are failing when I changed only one file with no impact on the library?

This one seemed to be caused by some issue with our testing infrastructure, as it looks like the job failed to be triggered on our testing queue. I'll requeue it now.

Issue #984, opened 2020-02-19
DHT22 Wrong temperature readings.


BramEsendam:
The temperature output dropped to to 1°C when the real temperature got above 26°C.

**Steps to reproduce**

```csharp
using var dht = new Dht22(26);
while (true)
 {
         // Try to read the temperature.
         var temp = dht.Temperature;
         if (!dht.IsLastReadSuccessful) continue;

          // Try to read the humidity.
          var humidity = dht.Humidity;
          if (!dht.IsLastReadSuccessful) continue;

          Console.WriteLine($"Temperature: {temp.Celsius:0.0} °C, 'Humidity: {humidity:0.0} %");

          break;
}
```

**Expected behavior**

Output: Temperature: 26.0 °C, 'Humidity: 36.1 %

**Actual behavior**

Output: Temperature: 1.0 °C, 'Humidity: 36.1 %

**Versions used**

Add following information:

- Dotnet on the build machine: `Version: 3.1.1 Commit:  a1388f194c`
- Dotnet on the raspberry pi 3B: `Version: 3.1.1 Commit:  a1388f194c`
- Version of `System.Device.Gpio` package: `1.0.0`
- Version of `Iot.Device.Bindings` package: `1.0.0`

BramEsendam:
![8a60d15c0ab975022315428a4098dcf6](https://user-images.githubusercontent.com/35614609/74871854-76982300-535c-11ea-917e-c6a7a007e503.png)

I ran it with a small python script as well.
And it resulted in the correct values.

BramEsendam:
I was running a Dht11 sensor before this, and i never had any problems with it.
Except of the 2c inaccuracy :P

joperezr:
> I was running a Dht11 sensor before this, and i never had any problems with it.

So when you say ya were running it before, do you mean with python or with .net core? If you were doing it with .net core and it was working correctly, did you recompile or updated the nuget package to get a newer version?

cc: @Ellerbach @ZhangGaoxing as FYI.

BramEsendam:
I was running the same code. Except for using the dht22 class i was using the dht11 one.

fomenkodmitry:
Sustained, similar bug.
You can see the code here: https://github.com/fomenkodmitry/AspNetCore-raspberrypi4-dht22
Platform: Rasberry pi 4 model b, net core 3.1

```bash
[...]

PR #1008, opened 2020-03-12
Fixing DHT22 temperature measurement not correct

Ellerbach:
- Fixing issue #984 by correcting the calculation for DHT22 temperature

### DIFF
--- a/src/devices/Dhtxx/Devices/Dht22.cs
+++ b/src/devices/Dhtxx/Devices/Dht22.cs
@@ -29,7 +29,7 @@ namespace Iot.Device.DHTxx
 
         internal override Temperature GetTemperature(byte[] readBuff)
         {
-            var temp = (readBuff[2] & 0x7F) + readBuff[3] * 0.1;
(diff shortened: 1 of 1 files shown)


BramEsendam:
@joperezr When will this be released on NuGet?

joperezr:
We will be shipping a preview package to NuGet in around 3 weeks, but in the meantime you could always just reference our daily build packages. The latest (as of now) package version of Iot.Device.Bindings (version [1.1.0-prerelease.20162.2](https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=dotnet5&package=Iot.Device.Bindings&protocolType=NuGet&version=1.1.0-prerelease.20162.2)) already has this fix available. In order to consume that version, simply add the following source to your NuGet.config (or add a new NuGet.config right next to your .csproj if you don't have one already)

```xml
<configuration>
  <packageSources>
    <!-- Your other NuGet sources like nuget.org goes here -->
    <add key="dotnet5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
  </packageSources>
</configuration>
```

PR #1011, opened 2020-03-16
Fixing exception in decoding credit card information

Ellerbach:
- Fixing an exception trapping to be a bit more broader. 
- Adding OverFlowException which happens time to time when the element is mal formed in a specific way

### DIFF
--- a/src/devices/Card/CreditCard/BerSplitter.cs
+++ b/src/devices/Card/CreditCard/BerSplitter.cs
@@ -43,7 +43,7 @@ namespace Iot.Device.Card.CreditCardProcessing
                     index += resSize.Item1;
 
                 }
-                catch (ArgumentOutOfRangeException)
+                catch (Exception ex) when (ex is ArgumentOutOfRangeException || ex is OverflowException)
                 {
                     // We may have a non supported Tag
                     break;


src/devices/Card/CreditCard/BerSplitter.cs
```diff
@@ -43,7 +43,7 @@ public BerSplitter(ReadOnlySpan<byte> toSplit)
                     index += resSize.Item1;
 
                 }
-                catch (ArgumentOutOfRangeException)
+                catch (Exception ex) when (ex is ArgumentOutOfRangeException || ex is OverflowException)
```
joperezr: Should we have different catch blocks instead and have the OverflowException instead write an error saying that the element was "malformed in some specific way"?

Ellerbach: Honestly, I was even hesitating to just catch all exceptions. For this part, we're trying to analyze if a message is a ber-tlv encoded one. If not, then this is most likely another type of data. Most of the time, the one coming if the buffer out of range. But seems like time to time the overflow one is coming. It's basically due to the fact that that either the index returned will be out of the array or the size of the buffer to create will be too big/small.

joperezr: Ok, I’m fine with this change as long as getting into this failed state won’t mean that we can give better feedback or better messages to consumers as to why this happened.

PR #1016, opened 2020-03-19
Fixing missing GpioController and ShouldDispose

### LINKED ISSUE #1013: Add a default constructor on bindings to take GpioController
When you want to use other ```GpioController``` than the default driver one, you can't in any of the binding devices.

With introduction of [FT4222](https://github.com/dotnet/iot/tree/master/src/devices/Ft4222) we need to be able to pass a specific ```GpioController``` to any class using a GPIO Controller. 

Ideal solution would be to add or modify all devices constructor to add the possibility to add your own ```GpioController``` if you want. 
We will need also to add a flag to dispose automatically or not the ```GpioController``` at the end.

See usage of the [FT4222 GpioController](https://github.com/dotnet/iot/tree/master/src/devices/Ft4222#gpio)

So for example taking the DHT sensor, we would need to change the constructor and adjust the code from:
```csharp
        /// <summary>
        /// Create a DHT sensor
        /// </summary>
        /// <param name="pin">The pin number (GPIO number)</param>
        /// <param name="pinNumberingScheme">The GPIO pin numbering scheme</param>
        public DhtBase(int pin, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical)
        {
            _protocol = CommunicationProtocol.OneWire;
            _controller = new GpioController(pinNumberingScheme);
            _pin = pin;

            _controller.OpenPin(_pin);
            // delay 1s to make sure DHT stable
            Thread.Sleep(1000);
        }
```

to:
```csharp
private bool _autoDispose;

        /// <summary>
        /// Create a DHT sensor
        /// </summary>
        /// <param name="pin">The pin number (GPIO number)</param>
        /// <param name="pinNumberingScheme">The GPIO pin numbering scheme</param>
        /// <param name="gpioController">A GPIO Controller if not using the default driver one</param>
        /// <param name="autoDispose">True to auto dispose the GPIO Controller when dispose</param>
        public DhtBase(int pin, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical, GpioController gpioController = null, bool autoDispose = true)
        {
            _protocol = CommunicationProtocol.OneWire;
            _autoDispose = autoDispose;
            if (gpioController == null)
            { _controller = new GpioController(pinNumberingScheme); }
            else
            {
                _controller = gpioController;
            }
            _pin = pin;

            _controller.OpenPin(_pin);
            // delay 1s to make sure DHT stable
            Thread.Sleep(1000);
        }

        /// <inheritdoc/>
        public void Dispose()
        {
            if (_autoDispose)
            {
                _controller?.Dispose();
            }
            _i2cDevice?.Dispose();
        }
```
And also adjust the constructor for the DHTxx:
```csharp
        public Dht22(int pin, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical, GpioController gpioController = null, bool autoDispose = true)
            : base(pin, pinNumberingScheme, gpioController, autoDispose)
        {
        }
```

That would be needed for all the devices using a ```GpioController```.

And DHT may not be a device for which the FT4222 would work. But it's to give an example.

Other alternative option: add the FT4222 into ```System.Device.Gpio```. And on the driver side, check if it does exist and pick it if no other driver.

Ellerbach:
- Fixes #1013 and adding GpioController and ShouldDispose everywhere it was needed
- Adding ShouldDispose as true in all the constructors
- Adjusting all Dispose function to take into consideration the new behavior

### DIFF
--- a/src/devices/CharacterLcd/LcdInterface.Gpio.cs
+++ b/src/devices/CharacterLcd/LcdInterface.Gpio.cs
@@ -45,9 +45,10 @@ namespace Iot.Device.CharacterLcd
             private bool _useLastByte;
 
             private GpioController _controller;
+            private bool _shouldDispose;
             private PinValuePair[] _pinBuffer = new PinValuePair[8];
 
-            public Gpio(int registerSelectPin, int enablePin, int[] dataPins, int backlightPin = -1, float backlightBrightness = 1.0f, int readWritePin = -1, GpioController controller = null)
+            public Gpio(int registerSelectPin, int enablePin, int[] dataPins, int backlightPin = -1, float backlightBrightness = 1.0f, int readWritePin = -1, GpioController controller = null, bool shouldDispose = true)
             {
                 _rwPin = readWritePin;
                 _rsPin = registerSelectPin;
@@ -66,6 +67,7 @@ namespace Iot.Device.CharacterLcd
                 }
 
                 _controller = controller ?? new GpioController(PinNumberingScheme.Logical);
+                _shouldDispose = shouldDispose;
 
                 Initialize();
             }
@@ -242,7 +244,12 @@ namespace Iot.Device.CharacterLcd
 
             protected override void Dispose(bool disposing)
             {
-                _controller?.Dispose();
+                if (_shouldDispose)
+                {
+                    _controller?.Dispose();
+                    _controller = null;
+                }
+
                 base.Dispose(disposing);
             }
         }
--- a/src/devices/CharacterLcd/LcdInterface.Gpio.cs
+++ b/src/devices/CharacterLcd/LcdInterface.Gpio.cs
@@ -66,8 +66,8 @@ namespace Iot.Device.CharacterLcd
                     throw new ArgumentException($"The length of the array given to parameter {nameof(dataPins)} must be 4 or 8");
                 }
 
+                _shouldDispose = controller == null ? true : shouldDispose;
                 _controller = controller ?? new GpioController(PinNumberingScheme.Logical);
-                _shouldDispose = shouldDispose;
 
                 Initialize();
             }
--- a/src/devices/DCMotor/DCMotor.cs
+++ b/src/devices/DCMotor/DCMotor.cs
@@ -15,13 +15,16 @@ namespace Iot.Device.DCMotor
     public abstract class DCMotor : IDisposable
     {
         private const int DefaultPwmFrequency = 50;
+        private bool _shouldDispose;
 
         /// <summary>
         /// Constructs generic <see cref="DCMotor"/> instance
         /// </summary>
         /// <param name="controller"><see cref="GpioController"/> related with operations on pins</param>
-        protected DCMotor(GpioController controller)
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
+        protected DCMotor(GpioController controller, bool shouldDispose)
         {
+            _shouldDispose = shouldDispose;
             Controller = controller;
         }
 
@@ -57,8 +60,11 @@ namespace Iot.Device.DCMotor
         {
             if (disposing)
             {
-                Controller?.Dispose();
-                Controller = null;
+                if (_shouldDispose)
+                {
+                    Controller?.Dispose();
+                    Controller = null;
+                }
             }
         }
 
@@ -79,7 +85,7 @@ namespace Iot.Device.DCMotor
                 throw new ArgumentNullException(nameof(speedControlChannel));
             }
 
-            return new DCMotor2PinNoEnable(speedControlChannel, -1, null);
+            return new DCMotor2PinNoEnable(speedControlChannel, -1, null, true);
         }
 
         /// <summary>
@@ -105,7 +111,8 @@ namespace Iot.Device.DCMotor
             return new DCMotor2PinNoEnable(
                 new SoftwarePwmChannel(speedControlPin, DefaultPwmFrequency, 0.0, controller: controller),
                 -1,
-                controller);
+                controller,
+                shouldDispose);
         }
 
         /// <summary>
@@ -134,7 +141,7 @@ namespace Iot.Device.DCMotor
                 throw new ArgumentOutOfRangeException(nameof(directionPin));
             }
 
-            return new DCMotor2PinNoEnable(speedControlChannel, directionPin, controller);
+            return new DCMotor2PinNoEnable(speedControlChannel, directionPin, controller, shouldDispose);
         }
 
         /// <summary>
@@ -167,7 +174,8 @@ namespace Iot.Device.DCMotor
             return new DCMotor2PinNoEnable(
                 new SoftwarePwmChannel(speedControlPin, DefaultPwmFrequency, 0.0, controller: controller),
                 directionPin,
-                controller);
+                controller,
+                shouldDispose);
         }
 
         /// <summary>
@@ -207,7 +215,8 @@ namespace Iot.Device.DCMotor
                 speedControlChannel,
                 directionPin,
                 otherDirectionPin,
-                controller);
+                controller,
+                shouldDispose);
         }
 
         /// <summary>
@@ -248,7 +257,8 @@ namespace Iot.Device.DCMotor
                 new SoftwarePwmChannel(speedControlPin, DefaultPwmFrequency, 0.0, controller: controller),
                 directionPin,
                 otherDirectionPin,
-                controller);
+                controller,
+                shouldDispose);
         }
     }
 }
--- a/src/devices/DCMotor/DCMotor.cs
+++ b/src/devices/DCMotor/DCMotor.cs
@@ -20,11 +20,9 @@ namespace Iot.Device.DCMotor
         /// Constructs generic <see cref="DCMotor"/> instance
         /// </summary>
         /// <param name="controller"><see cref="GpioController"/> related with operations on pins</param>
-        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
-        protected DCMotor(GpioController controller, bool shouldDispose = true)
+        protected DCMotor(GpioController controller)
         {
             Controller = controller;
-            ShouldDispose = shouldDispose;
         }
 
         /// <summary>
@@ -43,11 +41,6 @@ namespace Iot.Device.DCMotor
         }
 
         /// <summary>
-        /// True to dispose the Gpio Controller
-        /// </summary>
-        protected bool ShouldDispose { get; set; }
-
-        /// <summary>
         /// Disposes the <see cref="DCMotor"/> class
         /// </summary>
         public void Dispose()
@@ -64,11 +57,8 @@ namespace Iot.Device.DCMotor
         {
             if (disposing)
             {
-                if (ShouldDispose)
-                {
-                    Controller?.Dispose();
-                    Controller = null;
-                }
+                Controller?.Dispose();
+                Controller = null;
             }
         }
 
--- a/src/devices/DCMotor/DCMotor.cs
+++ b/src/devices/DCMotor/DCMotor.cs
@@ -20,9 +20,11 @@ namespace Iot.Device.DCMotor
         /// Constructs generic <see cref="DCMotor"/> instance
         /// </summary>
         /// <param name="controller"><see cref="GpioController"/> related with operations on pins</param>
-        protected DCMotor(GpioController controller)
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
+        protected DCMotor(GpioController controller, bool shouldDispose = true)
         {
             Controller = controller;
+            ShouldDispose = shouldDispose;
         }
 
         /// <summary>
@@ -41,6 +43,11 @@ namespace Iot.Device.DCMotor
         }
 
         /// <summary>
+        /// True to dispose the Gpio Controller
+        /// </summary>
+        protected bool ShouldDispose { get; set; }
+
+        /// <summary>
         /// Disposes the <see cref="DCMotor"/> class
         /// </summary>
         public void Dispose()
@@ -57,8 +64,11 @@ namespace Iot.Device.DCMotor
         {
             if (disposing)
             {
-                Controller?.Dispose();
-                Controller = null;
+                if (ShouldDispose)
+                {
+                    Controller?.Dispose();
+                    Controller = null;
+                }
             }
         }
 
@@ -87,13 +97,14 @@ namespace Iot.Device.DCMotor
         /// </summary>
         /// <param name="speedControlPin">Pin used to control the speed of the motor with software PWM (frequency will default to 50Hz)</param>
         /// <param name="controller"><see cref="GpioController"/> related to the <paramref name="speedControlPin"/></param>
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
         /// <returns><see cref="DCMotor"/> instance</returns>
         /// <remarks>
         /// <paramref name="speedControlPin"/> can be connected to either enable pin of the H-bridge.
         /// or directly to the input related with the motor (if H-bridge allows inputs to change frequently).
         /// Connecting motor directly to GPIO pin is not recommended and may damage your board.
         /// </remarks>
-        public static DCMotor Create(int speedControlPin, GpioController controller = null)
+        public static DCMotor Create(int speedControlPin, GpioController controller = null, bool shouldDispose = true)
         {
             if (speedControlPin == -1)
             {
@@ -113,6 +124,7 @@ namespace Iot.Device.DCMotor
         /// <param name="speedControlChannel"><see cref="PwmChannel"/> used to control the speed of the motor</param>
         /// <param name="directionPin">Pin used to control the direction of the motor</param>
         /// <param name="controller"><see cref="GpioController"/> related to the <paramref name="directionPin"/></param>
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
         /// <returns><see cref="DCMotor"/> instance</returns>
         /// <remarks>
         /// <paramref name="speedControlChannel"/> can be connected to either enable pin of the H-bridge.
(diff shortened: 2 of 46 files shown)


src/devices/CharacterLcd/LcdInterface.Gpio.cs
```diff
@@ -66,6 +67,7 @@ public Gpio(int registerSelectPin, int enablePin, int[] dataPins, int backlightP
                 }
 
                 _controller = controller ?? new GpioController(PinNumberingScheme.Logical);
+                _shouldDispose = shouldDispose;
```
joperezr: I would recommend to by default set this to false whenever the controller parameter wasn't null. So basically, if somebody passed in a controller, let's default shouldDispose to false.

src/devices/DCMotor/DCMotor.cs
```diff
@@ -20,9 +20,11 @@ public abstract class DCMotor : IDisposable
         /// Constructs generic <see cref="DCMotor"/> instance
         /// </summary>
         /// <param name="controller"><see cref="GpioController"/> related with operations on pins</param>
-        protected DCMotor(GpioController controller)
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
+        protected DCMotor(GpioController controller, bool shouldDispose = true)
```
joperezr: You don't really need to default shouldDispose here since this is a protected constructor and we are only calling it from our subclasses.

src/devices/DCMotor/DCMotor.cs
```diff
@@ -141,14 +153,15 @@ public static DCMotor Create(PwmChannel speedControlChannel, int directionPin, G
...
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
         /// <returns><see cref="DCMotor"/> instance</returns>
         /// <remarks>
         /// PWM pin <paramref name="speedControlPin"/> can be connected to either enable pin of the H-bridge.
         /// or directly to the input related with the motor (if H-bridge allows inputs to change frequently).
         /// <paramref name="directionPin"/> should be connected to H-bridge input corresponding to one of the motor inputs.
         /// Connecting motor directly to GPIO pin is not recommended and may damage your board.
         /// </remarks>
-        public static DCMotor Create(int speedControlPin, int directionPin, GpioController controller = null)
+        public static DCMotor Create(int speedControlPin, int directionPin, GpioController controller = null, bool shouldDispose = true)
```
joperezr: Same comment as the CharacterLCd here, we should probably have a Create Overload that doesn't take a shouldDispose, and set it to false in case controller was passed in. Another way to do this is to default it to false instead, and only set it to true if the controller that was passed in was null.

src/devices/CharacterLcd/LcdInterface.Gpio.cs
Ellerbach: The current behavior was that everything was disposed if you passed one GpioController. So the shouldDispose is then by default true. 
Also this create consistency across all the bindings.

src/devices/DCMotor/DCMotor.cs
Ellerbach: Right, I'll adjust this one.

src/devices/DCMotor/DCMotor.cs
Ellerbach: Same comment then for the previous one, it's more about consistency.

src/devices/CharacterLcd/LcdInterface.Gpio.cs
pgrawehr: I would prefer false as well, but it would be a breaking change for existing bindings. 
@krwq Your oppinion might be useful here.

src/devices/DCMotor/DCMotor.cs
joperezr: I see, I'm fine with keeping it as is for consistency

src/devices/DCMotor/DCMotor.cs
```diff
@@ -87,13 +87,14 @@ public static DCMotor Create(PwmChannel speedControlChannel)
...
         /// <param name="controller"><see cref="GpioController"/> related to the <paramref name="speedControlPin"/></param>
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
         /// <returns><see cref="DCMotor"/> instance</returns>
         /// <remarks>
         /// <paramref name="speedControlPin"/> can be connected to either enable pin of the H-bridge.
         /// or directly to the input related with the motor (if H-bridge allows inputs to change frequently).
         /// Connecting motor directly to GPIO pin is not recommended and may damage your board.
         /// </remarks>
-        public static DCMotor Create(int speedControlPin, GpioController controller = null)
+        public static DCMotor Create(int speedControlPin, GpioController controller = null, bool shouldDispose = true)
```
joperezr: Wait with your latest changes you are not really disposing anything right? It looks like you are only taking in the extra parameter but not really assigning to anything, and also not checking it before disposing.

src/devices/Dhtxx/DhtBase.cs
```diff
@@ -84,11 +89,14 @@ public virtual double Humidity
         /// </summary>
         /// <param name="pin">The pin number (GPIO number)</param>
         /// <param name="pinNumberingScheme">The GPIO pin numbering scheme</param>
-        public DhtBase(int pin, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical)
+        /// <param name="gpioController"><see cref="GpioController"/> related with operations on pins</param>
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
+        public DhtBase(int pin, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical, GpioController gpioController = null, bool shouldDispose = true)
```
joperezr: Same here, it doesn't look like you are actually checking this value before disposing.

src/devices/ExplorerHat/Led.cs
```diff
@@ -65,21 +67,19 @@ public void Off()
...
         /// <summary>
         /// Disposes the <see cref="Led"/> instance
         /// </summary>
         protected virtual void Dispose(bool disposing)
         {
-            if (!_disposedValue)
+            if (disposing)
             {
-                if (disposing)
+                if (_shouldDispose)
```
joperezr: This is not really disposing, this is only really turning the LED off. I realize you didn't introduce this, but now that you are touching this do you mind fixing that? Just dispose the controller after calling Off. Also, I believe Off should be called regardless in this case.

src/devices/ExplorerHat/Lights.cs
```diff
@@ -104,24 +106,22 @@ public void Off()
 
         #region IDisposable Support
 
-        private bool _disposedValue = false; // Para detectar llamadas redundantes
+        private bool _shouldDispose;
```
joperezr: The pourpuse of the previous line was different though, it was meant to make sure that the double call to Dispose() wouldn't crash the device

src/devices/ExplorerHat/Lights.cs
```diff
@@ -104,24 +106,22 @@ public void Off()
...
 
         /// <summary>
         /// Disposes the <see cref="Lights"/> instance
         /// </summary>
         protected virtual void Dispose(bool disposing)
         {
-            if (!_disposedValue)
+            if (disposing)
```
joperezr: I think you shouldn't change this condition from my comment above. This was really just meant to catch the case of double call to Dispose()

src/devices/ExplorerHat/Motors.cs
```diff
@@ -80,22 +82,20 @@ internal Motors(GpioController controller)
...
-        private bool _disposedValue = false;
+        private bool _shouldDispose;
 
         /// <summary>
         /// Disposes the <see cref="Motors"/> instance
         /// </summary>
         protected virtual void Dispose(bool disposing)
         {
-            if (!_disposedValue)
+            if (disposing)
```
joperezr: Same comment here.

src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -147,10 +150,13 @@ private bool TryGetDistance(out double result)
         /// <inheritdoc/>
         public void Dispose()
         {
-            if (_controller != null)
+            if (_shouldDispose)
             {
```
joperezr: NIT: you can also change the bellow code to just be:

```c#
_controller?.Dispose();
_controller = null;
```
in order to remove the if-check

src/devices/Nrf24l01/Nrf24l01.cs
```diff
@@ -121,15 +122,18 @@ public byte PacketSize
...
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
         public Nrf24l01(SpiDevice sensor, int ce, int irq, byte packetSize, byte channel = 2,
-            OutputPower outputPower = OutputPower.N00dBm, DataRate dataRate = DataRate.Rate2Mbps, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical)
+            OutputPower outputPower = OutputPower.N00dBm, DataRate dataRate = DataRate.Rate2Mbps, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical, GpioController gpioController = null, bool shouldDispose = true)
         {
             _sensor = sensor;
             _ce = ce;
             _irq = irq;
             PacketSize = packetSize;
+            _shouldDispose = shouldDispose;
```
joperezr: You are not checking this value on the Dispose() call and are always disposing.

src/devices/Pcx857x/Pcx857x.cs
```diff
@@ -39,10 +40,12 @@ public abstract class Pcx857x : GpioDriver
         /// The GPIO controller for the <paramref name="interrupt"/>.
         /// If not specified, the default controller will be used.
         /// </param>
-        public Pcx857x(I2cDevice device, int interrupt = -1, GpioController gpioController = null)
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
+        public Pcx857x(I2cDevice device, int interrupt = -1, GpioController gpioController = null, bool shouldDispose = true)
```
joperezr: NIT: Try not to default the values on abstract classes so that we make sure that we are flowing this value from all of the subclasses. That's the same feedback that I was suggesting on DCMotor

joperezr:
Left a few minor comments where  you missed Disposing or other minor things, but looks good in general. Thanks a lot for fixing @Ellerbach

src/devices/CharacterLcd/LcdInterface.Gpio.cs
krwq: One thing we should definitely do is to always dispose when controller is null (or throw an exception when user passes null and asks for not disposing).

Considering our current guidelines is to always dispose and to not re-use GpioController IMO we should always dispose but I'm not strongly opinionated. I think it's much easier to explain what this method does if we consistently dispose or not dispose - if we choose to not dispose then we should change guidelines to match.

Ellerbach: OK, so let me adjust the code to:
- dispose whatever the shouldDispose parameter is if gpioController is null
- In case gpioController is not null, then follow the shouldDispose wish

Recommendatin is not to reuse GpioController but in some cases like when using FT4222 or a controller that can only be created once, you don't really have other options.

src/devices/Pcx857x/Pcx857x.cs
Ellerbach: Yes, that said, I just added on top, the initial class had anyway some default values.

Ellerbach:
> Left a few minor comments where you missed Disposing or other minor things, but looks good in general. Thanks a lot for fixing @Ellerbach

OK, so I have adjusted the misses and also made consistency to always dispose when the controller is null. Please check again all is ok! Thanks again

src/devices/Dhtxx/DhtBase.cs
```diff
@@ -269,7 +277,11 @@ internal virtual void ReadThroughI2c()
         /// <inheritdoc/>
         public void Dispose()
         {
-            _controller?.Dispose();
+            if (_shouldDispose)
+            {
```
joperezr: NIT: looks like here we are missing to also reassign _controller to Null, I know it wasn't introduced by your change but can we do that in case double Dispose is called?

src/devices/ExplorerHat/Led.cs
```diff
@@ -76,7 +80,10 @@ protected virtual void Dispose(bool disposing)
             {
                 if (disposing)
                 {
-                    Off();
+                    if (_shouldDispose)
```
joperezr: Looks like my last comment on here hasn't been addressed yet. For convenience, this was it:

```
This is not really disposing, this is only really turning the LED off. I realize you didn't introduce this, but now that you are touching this do you mind fixing that? Just dispose the controller after calling Off. Also, I believe Off should be called regardless in this case.
```

src/devices/ExplorerHat/Lights.cs
```diff
@@ -115,10 +119,13 @@ protected virtual void Dispose(bool disposing)
             {
                 if (disposing)
                 {
-                    LedArray[0].Dispose();
-                    LedArray[1].Dispose();
-                    LedArray[2].Dispose();
-                    LedArray[3].Dispose();
+                    if (_shouldDispose)
```
joperezr: NIT, looks like we forgot to also dispose the controller here.

src/devices/ExplorerHat/Motors.cs
```diff
@@ -91,8 +95,11 @@ protected virtual void Dispose(bool disposing)
             {
                 if (disposing)
                 {
-                    _motorArray[0].Dispose();
-                    _motorArray[1].Dispose();
+                    if (_shouldDispose)
```
joperezr: not disposing the controller here either.

src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -32,11 +33,13 @@ public class Hcsr04 : IDisposable
...
         /// <param name="triggerPin">Trigger pulse input.</param>
         /// <param name="echoPin">Trigger pulse output.</param>
-        public Hcsr04(GpioController gpioController, int triggerPin, int echoPin)
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
+        public Hcsr04(GpioController gpioController, int triggerPin, int echoPin, bool shouldDispose = true)
         {
             _echo = echoPin;
             _trigger = triggerPin;
             _controller = gpioController;
+            _shouldDispose = shouldDispose;
```
joperezr: you don't necessarily have to fix this, but I do wanted to point out that in a few bindings like this one,

joperezr: Disregard this comment.

src/devices/Dhtxx/DhtBase.cs
Ellerbach: It's a readonly value, so I don't think it can be assigned to null?

src/devices/ExplorerHat/Lights.cs
Ellerbach: the controller should not be disposed in this dispose as every individual Led is disposing it. I moved out of the if the individual calls to each Led.

joperezr: would that be a problem though? it will mean that perhaps the first LED will dispose it, and then the second one will try to call Off() before disposing it again and that will fail. Wouldn't it be better to just have the LEDs dispose and have the Ligths Dispose dispose the actual controller?

Ellerbach: Yes, great catch. Adding Off(); before disposing. Then the controller will be disposed at the Lights level, not at every individual level.

pgrawehr:
While working on #1044: It might really be worth considering to make the GpioController a mandatory parameter, even if that's breaking. Otherwise, the interface is error-prone, as people leave out the paramter, even if they need to provide one (i.e. for Arduino, etc). This will either cause an exception, because the standard GpioController is likely not available, or worse still, use the wrong controller if there are multiple (and make people wonder why it doesn't work). 
Also, as I designed it now, using a GpioController without a Board reference should be deprecated.

Ellerbach:
> While working on #1044: It might really be worth considering to make the GpioController a mandatory parameter, even if that's breaking

I did change all the constructors to add the GpioController as a paramter. If null, then the default one will be used. If not null, then the shouldDispose parameter will decide if to be disposed or not at the end.

pgrawehr:
> 
> I did change all the constructors to add the GpioController as a paramter. If null, then the default one will be used. If not null, then the shouldDispose parameter will decide if to be disposed or not at the end.

Yes, I know. But I think this might not be optimal. I'm concerned about the fact that the default one will be wrong or just not work at all. I.e. if one is working with an Arduino or an FT4222, attempting to use the default GpioController will just throw an exception, because the default controller for such boards cannot be auto-detected. 

So it might be better to require the GpioController to be not null upon constructing the bindings, to avoid this kind of possible error. This would prevent the above misstake as well as later prevent "dangling" controllers that are not associated to a board. (Note: the exception to this rule would of course be bindings where the controller truly is optional, such as the MCP chips, where it's optionally used for interrupt handling only).

Ellerbach:
> I'm concerned about the fact that the default one will be wrong or just not work at all. I.e. if one is working with an Arduino or an FT4222, attempting to use the default GpioController will just throw an exception, because the default controller for such boards cannot be auto-detected

Yes and no. I get your point. Now, if you are using an Arduino or a FT4222, you know it and it's on purpose. So you will create the GpioController before and pass it to the constructor. So in short, you'll do the work of passing it. Now, for the very fast majority, the build in driver is just what they need. So they'll just pass it. I guess the important is to make sure we can, if needed pass an Arduino or a FT4222 or any other GpioController we'll build over time. But keep it simple for most people.

pgrawehr:
> Now, if you are using an Arduino or a FT4222, you know it and it's on purpose. So you will create the GpioController before and pass it to the constructor. So in short, you'll do the work of passing it. Now, for the very fast majority, the build in driver is just what they need. So they'll just pass it. I guess the important is to make sure we can, if needed pass an Arduino or a FT4222 or any other GpioController we'll build over time. But keep it simple for most people.

I'm not sure that "most people" will be working on (typically) a Raspi. Right now, that's certainly the case, but with Arduino and FT4222 supoort, this library could become feasible also as platform for school projects and rapid prototyping. On the other hand, keeping it simple is certainly a good concept. 

Maybe some more complex solutions are necessary, like keeping a static reference somewhere, so that the GpioController can be "auto-detected". Has other drawbacks, I know, but just my 2ct.

PR #1021, opened 2020-03-23
Bug fix on SPI for FT4222

Ellerbach:
- Fixing bug on SPI ReadByte()
- Adjusting documentation to properly shows Linux support

### DIFF
--- a/src/devices/Ft4222/Ft4222Spi.cs
+++ b/src/devices/Ft4222/Ft4222Spi.cs
@@ -248,7 +248,7 @@ namespace Iot.Device.Ft4222
         {
             Span<byte> toRead = stackalloc byte[1];
             Read(toRead);
-            return toRead[1];
+            return toRead[0];
         }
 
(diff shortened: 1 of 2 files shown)


src/devices/Ft4222/Ft4222Spi.cs
```diff
@@ -248,7 +248,7 @@ public override byte ReadByte()
         {
             Span<byte> toRead = stackalloc byte[1];
             Read(toRead);
-            return toRead[1];
+            return toRead[0];
```
joperezr: Great catch!

Ellerbach: Well, I don't really know how it manged to be here! And I'm using the FT4222 on a new project which uses SPI. And I needed to read a byte. so it was quite straight forward to find :-) Anyway, the good news is that the FT4222 works really perfectly for SPI and GPIO at the same time.

PR #1024, opened 2020-03-25
Adding driver for AD7193 (Analog Devices 24-bit ADC)

andrasfuchs:
This is the addition of the driver for AD7193 analog-to-digital converter IC with a sample demonstrating both single and continuous conversion modes.


joperezr:
@andrasfuchs thanks a lot for your contribution! Before we go ahead and review this PR, did you check our existing device binding for ADS1115 ADC? seems like a very similar device and instead of adding a new project, I wonder if we could refactor that in some way so that we can have shared common logic with two derived more specific devices for this one and the Ads1115. Do you mind checking the implementation to see if there is any logic in there that you could reuse and could be refactored?

andrasfuchs:
@joperezr Yes, I started with the Ads1115 project as an example when I started to work on the AD7193's driver.

I had the impression that all ADCs could have a common base class, because they share many functions, but I'm not experienced enough yet with other ADCs from other manufacturers to make the proper suggestion.

One thing I would definitely miss from the Ads1115 implementation if I had to use it, is the continuous mode support (that I implemented with AD7193), it's usually a much more performant way to measure the values at a high sample rate.

My plan is to implement one more driver for the more advanced [Analog Devices AD7606](https://www.analog.com/media/en/technical-documentation/data-sheets/AD7606B.pdf) in the next few weeks, and I'm sure I will have a better understanding of the commonalities and differences among ADC later on.

Ideally we would have an common interface for them with the basic functions (like single and continuous conversion) to make the life of the end user easier. 

I would also include some metadata for the devices in general, like manufacturer, model name, datasheet URI etc., because that would make a very good base for a catalog later one, where you will have hundreds of devices supported.

pgrawehr:
Note that I already prepared an extension for the ADS1115 driver, that (among other stuff) adds continuous support. See PR #915. There's only a small bug in that driver for now, which should be quick to fix (I actually know what needs to be done).

joperezr:
> One thing I would definitely miss from the Ads1115 implementation if I had to use it, is the continuous mode support (that I implemented with AD7193), it's usually a much more performant way to measure the values at a high sample rate.

Also one more thing to note (apart from @pgrawehr saying that he already has a PR adding this support) is that it is totally ok to have a base class with common functionality and then having subclasses extending that so you could still add this support only on your device. All that said, if you can't find a good way to have this code refactored and plug it in to our current ADCs, it is fine for now to keep it as is and then refactor it on a future PR.

src/devices/Ad7193/Ad7193.sln
```diff
@@ -0,0 +1,79 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
```
joperezr: We Usually don't add sln files to the repo. This might actually change in the future but at least for consistency lets keep this file out from your PR for now.

src/devices/Ad7193/Ad7193.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <Configurations>Windows-Debug;Linux-Debug;Linux-Release;Windows-Release;Debug;Release</Configurations>
```
joperezr: Project shouldn't require configurations once you have built using build.cmd once, so please go ahead and remove this line.

src/devices/Ad7193/Ad7193.csproj
```diff
@@ -0,0 +1,37 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!--Cases used when running in VS, with the solution config either Windows-Debug or Linux-Debug (or -Release) -->
```
joperezr: Please also remove all these next lines and just keep the ProjectReference to System.Device.Gpio which is passing in the RuntimeIdentifier as Linux. You can take a look at our other projects to see what do we usually do.

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+using global::Iot.Units;
+
+namespace Iot.Device.Ad7193
+{
+    /// <summary>
+    /// Represents the Analog Devices AD7193, the 4-channel, 4.8 kHz, ultralow noise, 24-bit sigma-delta ADC with PGA
+    /// </summary>
+    public class Ad7193 : IDisposable
+    {
+        #region Metadata for IDevice
```
joperezr: We usually don't include the below members for metadata and instead just add them to the Readme file. There is no real need to have them as public API when they are really just info about the device, so they should really instead just be info in a Readme file.

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+        /// </summary>
+        public const string DataSheetURI = "https://www.analog.com/media/en/technical-documentation/data-sheets/AD7193.pdf";
+        #endregion
+
+        #region Metadata for ISpiDevice
+
+        /// <summary>
+        /// The list of SPI modes that are valid for this device
+        /// </summary>
+        public const SpiMode ValidSpiModes = SpiMode.Mode3;
```
joperezr: NIT: these two SPI related properties shouldn't be part of the public API, if you need them to do validation during execution just keep them as private or internal

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+        /// </summary>
+        public const int MaximumSpiFrequency = 10000000;
+        #endregion
+
+        #region metadata for IAdcDevice
+
+        /// <summary>
+        /// Number of ADCs on the device
+        /// </summary>
+        public const int ADCCount = 1;
```
joperezr: Same goes for all these, if they are not configurable (properties) which would affect the interaction with the device, lets keep them off from the surface area.

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+        private readonly object _spiTransferLock = new object();
+        private readonly Stopwatch _stopWatch = new Stopwatch();
+
+        private SpiDevice _spiDevice = null;
+
+        /// <summary>
+        /// The list of received ADC values
+        /// </summary>
+        // protected ConcurrentQueue<AdcValue> adcValues = new ConcurrentQueue<AdcValue>();
+        protected BlockingCollection<AdcValue> adcValues = new BlockingCollection<AdcValue>();
```
joperezr: Why protected if there are no subclasses to this type? It is better to have it as private for now, and then if we ever extend this to have subclasses we can change the visibility of the member

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+
+        /// <summary>
+        /// The event that is fired every time a new value is received from the ADC
+        /// </summary>
+        public event EventHandler<AdcValueReceivedEventArgs> AdcValueReceived;
```
joperezr: NIT: should the name of this event be more something like `OnValueChange` or something like that?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+
+        private readonly StringBuilder _sb = new StringBuilder();
+
+        private readonly uint[] _registerCache = { 0x00, 0x080060, 0x000117, 0x000000, 0xa2, 0x00, 0x000000, 0x000000 };
+        private readonly byte[] _registerSize = { 1, 3, 3, 3, 1, 1, 3, 3 };
+
+        /// <summary>
+        /// The external reference voltage value. The default is 2.5V on REFIN1+ and REFIN1- (on the Digilent Pmod AD5 board)
+        /// </summary>
+        public double VReference { get; set; } = 2.50f;
```
joperezr: What is the use case of exposing this?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+
+        /// <summary>
+        /// Gain level on the ADC
+        /// </summary>
+        public Gain PGAGain
+        {
+            set
+            {
+                _registerCache[(byte)Register.Configuration] &= 0xFF_FFF8;          // keep all bit values except Gain bits
```
joperezr: NIT: small thing but we usually do something like `.. &= ~0b111;` for cases like this, which makes it more obvious that what we are doing is to flip those last three bits.

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+                _registerCache[(byte)Register.Configuration] |= (uint)value;
```
joperezr: It won't matter much but perhaps you want to check that the value is not greater than 0b111 or you could also just AND it with 0b111 as well.

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+        protected BlockingCollection<AdcValue> adcValues = new BlockingCollection<AdcValue>();
+
+        /// <summary>
+        /// The event that is fired every time a new value is received from the ADC
+        /// </summary>
+        public event EventHandler<AdcValueReceivedEventArgs> AdcValueReceived;
+
+        private readonly StringBuilder _sb = new StringBuilder();
+
+        private readonly uint[] _registerCache = { 0x00, 0x080060, 0x000117, 0x000000, 0xa2, 0x00, 0x000000, 0x000000 };
```
joperezr: What is the point of keeping a register Cache? I'm not saying that is wrong, but usually you really want to go and check it everytime from the device itself instead of using something from the cache.

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+
+        /// <summary>
+        /// The external reference voltage value. The default is 2.5V on REFIN1+ and REFIN1- (on the Digilent Pmod AD5 board)
+        /// </summary>
+        public double VReference { get; set; } = 2.50f;
+
+        /// <summary>
+        /// Gain level on the ADC
+        /// </summary>
+        public Gain PGAGain
```
joperezr: No Getter?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,838 @@
...
+                _registerCache[(byte)Register.Configuration] |= (uint)value;
+
+                SetRegisterValue(Register.Configuration, _registerCache[(byte)Register.Configuration]);
+            }
+        }
+
+        /// <summary>
+        /// Human-readable representation of the current status register
+        /// </summary>
+        public string Status
```
joperezr: Seems like this getter is actually performing a lot of operations so I guess that we should probably move this to be a GetStatus method instead.

joperezr:
Hey @andrasfuchs I left some initial feedback that should help get your code to a state where it follows the same patterns as the rest of the repo. Take a look at it and let me know if you have any questions.

src/devices/Ad7193/Ad7193.sln
pgrawehr: Hmmm... I would definitelly keep them for the individual bindings. You'll need them anyway, and it helps keeping things together a lot. Most recent changes kept it IIRC.

src/devices/Ad7193/Ad7193.csproj
pgrawehr: @joperezr (for this and the next few comments) That was my change. And it's the only way I found that made the build pass in VS for all platforms. Without this, if you build on Windows, you get an error saying that System.Device.Gpio.dll was not found (it's using a wrong path). 
This is relevant both for running samples on Windows (for bindings where this is possible) as well as to make sure that any unit tests run on all platforms. We have many unit tests that are not executed on windows due to this (although they are technically platform-independent). I'll create separate PR/issues after further investigation, but I've just tried out this concept here as Andras has run into much the same issues as I have.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: Is there any chance that you will include metadata in the classes themselves? I think they are very useful from the user's perspective, just like in the case of any Windows driver.
This would make it possible to compare different devices, ADCs programmatically, or list all supported devices in the framework, programmatically.
I also use KiCad EDA (a PCB designer program), and they store these pieces of information in their files as well, which is very useful.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: The SPI properties are related to the technical restriction we must apply if we use this particular device, so I think it is very good if we have this information on the class.
On the other hand, I agree that some check should be done relating to these settings in the constructor to check if the SpiDevice in the right mode/at the right frequency.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: Would you be more open to keep metadata in separate class, or do you insist on removing them completely no matter how they are presented?

src/devices/Ad7193/Ad7193.cs
andrasfuchs: This is protected, because I read in one of your contribution documents that you prefer to have a base class with the basic functionality of the driver, and let others extend on that class for the more device specific functions by inheriting from the driver. 
In this case it's important for some advanced functions to keep the access to the ADC values, I think, if someone would, let's say, implement the channel-separation/ordering logic for these values, to have separate buffers for every channel while using the multi-channel continuous conversion.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: I don't think so, but I'm not a native speaker, you are probably a better judge of this :)

When the event fires, the old value doesn't change, but a new value got converted and was added to the list of values in the buffer. 
It's more like when in network communications we receive a new packet that we potentially need to process it.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: The reference voltage can be set up in a way on the PCB, that is comes from an external source, connected to a particular pin on the IC. 
If the user decides to use an external voltage, they need to set this value manually to reflect their design.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: I didn't know this syntax, thank you!

src/devices/Ad7193/Ad7193.cs
andrasfuchs: There are some rare cases when the cache is very useful: when we are in continuous mode, the device send the ADC values continuously on the bus. If we don't ask the ADC to include the current status register in the results, we can win a little bandwidth, but we don't know on which channel that particular ADC value was converted on.
If we previously selected only one channel, that's now a problem, we know that all values are from channel x. 
In this case we should not break the flow of the data, just because we want to know the channel number. It's alright, if we get it from cache.
The situation is similar to values like the gain and filter, these values are typically set before the start of the continuous conversion and they remain the same for sure until we stop that conversion.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: It should have one, I agree.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: Status, Mode and Config are all like this, and I was hesitant about their naming as well.
They basically show the register values in a human readable form, they should be named differently.
What do you think? GetStatus or GetStatusText would be more appropriate? 
Or I could even more these into a RegisterToString(Register register) method, and change the Status, Mode and Config properties to uints with their getters and setters.

andrasfuchs:
Thank you @joperezr for the feedback, I appreciate it.

src/devices/Ad7193/Ad7193.sln
joperezr: I don't believe new bindings have added solutions, at least not that I'm aware of, and if there are any is only because we missed them on the Pull Request. You may be right that perhaps adding them is the right thing to do, but I would rather do this as a bulk for all devices in order to always have consistency on the repo rather than needing to special case.

src/devices/Ad7193/Ad7193.csproj
joperezr: The code for this binding is AnyCPU and is all IL that applies for all platforms, so there is no need to build it for more than one configuration/platform. If you keep it the way we have it for other bindings and make sure you first build using build.cmd/build.sh from the root, then building individual projects from VS should definitely work for all platforms, if that is not the case it is definitely a bug.

src/devices/Ad7193/Ad7193.cs
joperezr: It would be good to have an actual scenario or need for the APIs first before we expose all these methods that we will then have to support in future versions. That said, if you think it is valuable to keep them, I would much rather have just one method called somehting like GetDeviceMetadata() which would in turn return an object that contained all of these properties. That way you don't polute your surface area, and you keep it simple so that it is a class that is easy to  use and interact with without the need to look into samples and such.

src/devices/Ad7193/Ad7193.csproj
pgrawehr: I do think this is a bug/inconsistency in the build environment. I'll come up with a separate ticket on this, including a more extensive explanation of the problem.

src/devices/Ad7193/Ad7193.cs
joperezr: I understand that, but seems to be that there are no subclasses today of this base class. My suggestion was that once somebody does indeed add a subclass from this base, then we can change it back to be protected, but there is no need to have it protected today if nobody is using it.

src/devices/Ad7193/Ad7193.cs
joperezr: I see, my comment wasn't really about using proper english (I'm also not a native speaker 😄), but it was instead more about how EventHandlers traditionally look in C# APIs. They usually have the pattern of "OnSomethingHappened" so that it is clear that this is an eventhandler and that people that want to use it should just append delegates to it. Given your explanation, probably the right name for it should be `OnValueReceived`?

src/devices/Ad7193/Ad7193.csproj
pgrawehr: See #1037

src/devices/Ad7193/Ad7193.cs
andrasfuchs: Alright, I changed the scope of it to private.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: Oh, I see, sure! :) I renamed it.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: Created two new methods for these kind of operations: GetRegisterBits and SetRegisterBits

andrasfuchs:
@joperezr Could you look at this PR again?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
+﻿using System;
```
krwq, line 1: copyright header missing

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Device.Spi;
+using System.Diagnostics;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Ad7193.Metadata;
+using global::Iot.Units;
```
krwq, line 10: is global:: necessary?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+
+        /// <summary>
+        /// Metadata of AD7193
+        /// </summary>
+        protected static Ad7193Metadata metadata = new Ad7193Metadata();
+
+        /// <summary>
+        /// The list of received ADC values
+        /// </summary>
+        // private ConcurrentQueue<AdcValue> _adcValues = new ConcurrentQueue<AdcValue>();
```
krwq, line 32: remove commented out code

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+        /// <summary>
+        /// Polarity select bit: true is unipolar, false is bipolar operation.
+        /// </summary>
+        public bool Unipolar
+        {
+            get
+            {
+                return (GetRegisterBits(Register.Configuration, BitMask.ConfigurationU) == 1);
+            }
+
```
krwq, line 59: remove extra space

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+        private readonly StringBuilder _sb = new StringBuilder();
+
+        private readonly uint[] _registerCache = { 0x00, 0x080060, 0x000117, 0x000000, 0xa2, 0x00, 0x000000, 0x000000 };
+        private readonly byte[] _registerSize = { 1, 3, 3, 3, 1, 1, 3, 3 };
+
+        /// <summary>
+        /// The external reference voltage value. The default is 2.5V on REFIN1+ and REFIN1- (on the Digilent Pmod AD5 board)
+        /// </summary>
+        public double VReference { get; set; } = 2.50f;
```
krwq, line 48: can we use full name (VoltageReference)?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+
+        /// <summary>
+        /// Polarity select bit: true is unipolar, false is bipolar operation.
+        /// </summary>
+        public bool Unipolar
+        {
+            get
+            {
+                return (GetRegisterBits(Register.Configuration, BitMask.ConfigurationU) == 1);
```
krwq, line 57: remove redundant brackets around

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+        /// <summary>
+        /// Gain level on the ADC
+        /// </summary>
+        public Gain PGAGain
+        {
+            get
+            {
+                return (Gain)(GetRegisterBits(Register.Configuration, BitMask.ConfigurationG));
+            }
+
```
krwq, line 76: remove extra enter

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+        }
+
+        /// <summary>
+        /// Power-down mode. In power-down mode, all AD7193 circuitry, except the bridge power-down switch, is powered down. The bridge power-down switch remains active because the user may need to power up the sensor prior to powering up the AD7193 for settling reasons. The external crystal, if selected, remains active.
+        /// </summary>
+        public bool IsPowerDown
+        {
+            get
+            {
+                GetRegisterValue(Register.Mode);
```
krwq, line 127: is the return value ignored intentionally?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+        }
+
+        /// <summary>
+        /// Ready bit for the ADC. This bit is set when data is written to the ADC data register.
+        /// </summary>
+        public bool IsReady
+        {
+            get
+            {
+                GetRegisterValue(Register.Status);
```
krwq, line 139: ditto

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+                return (GetRegisterBits(Register.Mode, BitMask.StatusRDY) != 0b1);
```
krwq, line 140: remove extra brackets around expression

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+        public bool ContinuousRead
+        {
+            get
+            {
+                return _continuousRead;
+            }
+
+            set
+            {
+                if (value)
```
krwq, line 171: This could be collapsed to:

```csharp
SetRegisterValue(Register.Communications, value ? 0b0101_1100 : 0b0101_1000);
```

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+        }
+
+        /// <summary>
+        /// Enables or disables DAT_STA Bit (appends status register to data register when reading)
+        /// </summary>
+        public bool AppendStatusRegisterToData
+        {
+            get
+            {
+                return (GetRegisterBits(Register.Mode, BitMask.ModeDAT_STA) == 1);
```
krwq, line 191: remove extra brackets

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+            }
+
+            set
+            {
+                _registerCache[(byte)Register.Mode] = SetRegisterBits(Register.Mode, BitMask.ModeDAT_STA, value ? 1U : 0);
```
krwq, line 196: can the caching happen directly in SetRegisterBits? how do you know cache is uninitialized vs initialized with the right value?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+                while (_stopWatch.IsRunning)
+                {
+                    elapsedTicks = _stopWatch.ElapsedTicks;
+                    if (elapsedTicks >= nextSampleAt)
+                    {
+                        nextSampleAt = elapsedTicks + samplePerTicks;
+                        ReadADCValue();
+                        samples++;
+                        jitter = (elapsedTicks - (samples * samplePerTicks));
+                        if (JitterCorrection)
```
krwq, line 444: i have a mixed feelings on adding utilities like that to devices. One one hand they're useful but on the other hand I suspect these might be useful elsewhere as well. Can you please write in the docs of this device more about this so that if I find similar code to be useful in the future for different device we could unify those?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+                    }
+                }
+            });
+        }
+
+        /// <summary>
+        /// Reads the current ADC result value on the selected channel
+        /// </summary>
+        /// <returns>24-bit raw value of the last ADC result (+ status byte if enabled)</returns>
+        public uint? ReadADCValue()
```
krwq, line 464: We usually use TryReadValue/TryReadADCValue instead of nullable. Also if there can be multiple types of error and not just "error" you might want to return status or just throw an exception if this is super unlikely

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+            _adcValues.Add(adcValue);
+
+            // call the event handler
+            OnValueReceived?.Invoke(this, new AdcValueReceivedEventArgs(adcValue));
+
+            return raw;
+        }
+
+        /// <summary>
+        /// Reads a single value on the selected channel
```
krwq, line 495: can you document how is this different from ReadADCValue

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+            if (Unipolar)
+            {
+                voltage = (double)adcValue / 16777216.0;
+            }
+            else if (!Unipolar)
+            {
+                voltage = ((double)adcValue / 8388608.0) - 1.0;
+            }
+
+            voltage *= (VReference / pgaGain);
```
krwq, line 552: remove extra brackets here and everywhere when you spot (i.e. below)

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+        {
+            byte registerAddress = (byte)register;
+            byte byteNumber = _registerSize[registerAddress];
+            byte[] writeBuffer = new byte[byteNumber + 1];
+
+            byte commandByte = (byte)((byte)CommunicationsRegisterBits.ReadOperation | GetCommAddress(registerAddress));
+            writeBuffer[0] = commandByte;
+
+            byte[] readBuffer = new byte[writeBuffer.Length];
+            lock (_spiTransferLock)
```
krwq, line 581: usually we make classes not thread safe and the caller can lock them if they want. I haven't seen any thread being spawned in the ctor - might make sense to remove the locks overall since that will match what we do elsewhere

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+                _spiDevice.TransferFullDuplex(writeBuffer, readBuffer);
+            }
+
+            byte[] trimmedReadBuffer = new byte[readBuffer.Length - 1];
+            Array.Copy(readBuffer, 1, trimmedReadBuffer, 0, trimmedReadBuffer.Length);
+            readBuffer = trimmedReadBuffer;
+
+            _registerCache[registerAddress] = ByteArrayToUInt32(readBuffer);
+
+            // Debug.WriteLine($"Read Register - address: {registerAddress.ToString("X2")}, command: {commandByte.ToString("X2")}, received: {String.Join(' ', readBuffer.Select(x => x.ToString("X2")))}");
```
krwq, line 592: remove commented out code

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+        {
+            return (((x) & 0x07) << 3);
+        }
+
+        /// <summary>
+        /// Returns the human-readable representation of a register
+        /// </summary>
+        /// <param name="register">The register you need the text-representation of</param>
+        /// <returns></returns>
+        public string RegisterToString(Register register)
```
krwq, line 654: most of the devices made Register enum internal, while those are useful when debugging it's more confusing for users who didn't read datasheet

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+
+            return result;
+        }
+
+        /// <summary>
+        /// Disposes the AD7193 object and closes the SPI device
+        /// </summary>
+        public void Dispose()
+        {
+            if (_spiDevice != null)
```
krwq, line 772: null check is not needed (since you use `?.` operator below)

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+            {
+                _spiDevice?.Dispose();
+                _spiDevice = null;
+            }
+        }
+
+        private uint GetRegisterBits(Register reg, BitMask bitmask)
+        {
+            uint result = _registerCache[(byte)reg] & (uint)bitmask;     // clear all bit(s) except the bit(s) we would like to read
```
krwq, line 781: comment above

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,855 @@
...
+
+            if (bitmask > 0)
```
krwq, line 783: I'm having hard time following what this piece of code is doing. Seems to me it will remove the last n bits of the result (with n being number of 0s in the binary representation of the bitmask) but what's not clear is why do we do that. Could you write a comment explaining what's happening here?

src/devices/Ad7193/Ad7193.cs
andrasfuchs: yes

src/devices/Ad7193/Ad7193.cs
andrasfuchs: yes

src/devices/Ad7193/Ad7193.cs
andrasfuchs: This is just the question of preference, I think, but true. I changed it.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: The Register enum can't be internal because of the RegisterToString() method that can be used outside of he library. As you said it's useful for debugging, or in cases when the user needs more detailed information about the state of the device.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: I can't see extra space here, but I removed the brackets as you asked at other places.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: What is a copyright header?

src/devices/Ad7193/Ad7193.cs
andrasfuchs: Strange, when I removed it in the past it caused a compilation error, but now it looks fine. I removed it.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: The cache is initialized with the initial values according to the specification of AD7193.
You're right, the caching should happen in the SetRegisterBits().

andrasfuchs: I moved the cache updater logic into the SetRegisterValue() method, which is also called by the SetRegisterBits() method.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: I think I had some trouble with threads while reading the value when I used the event handler in the test application, and that's why I decided to use the lock in the driver. 
It would work if I used it in the sample application as well, but then everyone who uses the event-based value acquisition should know that they need to use locks.
This is a safer solution, although I understand it is not critical, since everyone could do the lock manually when they use this driver.
Would you still prefer to remove it from here and let the users do the locking?

src/devices/Ad7193/Ad7193.cs
andrasfuchs: Done

src/devices/Ad7193/Ad7193.cs
andrasfuchs: Many device has registers and the registers are compartmentalized into smaller values.
For example in the case of [AD7193](https://www.analog.com/media/en/technical-documentation/data-sheets/AD7193.pdf), the 8-bit Status register stores variables like the the 1-bit ready, 1-bit error and 4-bit channel index variables.
To extract these values separately I created the GetRegisterBits method, so I can pass a mask indicating the location of the - let's say - ready bit in the register, and the method returns me only 1-bit of information: 1 if the ready bit is set, 0 is it's not set.
I've just made a description for this method, to make this more understandable.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: Yes, I had mixed feelings as well. It would be great to have a base class for all the ADCs, so the helper methods like these could go there, unifying the way they do the processing.
At this point we don't have a device class hierarchy, so I think it is a good place to have it, for now.

src/devices/Ad7193/Ad7193.cs
andrasfuchs: I checked and it can be only one general error, if the Status register indicates it.
I need to test if it is common to have a few errors during conversions, I'll do it today.

andrasfuchs: Alright, so in order to get an error, you must misconfigure something, otherwise the conversion goes on without any errors even at the maximum sampling rate.
Because of this, I changed the return value to a non-nullable uint and added a logic to throw an Exception if something goes wrong. It should happen fairly rarely.

andrasfuchs: I did some further testing, and I was wrong: the error bit gets set even if the signal on the ADC channel is out of range. Because of this, I changed the method to TryReadADCValue(out uint value) returning a bool.

krwq:
@andrasfuchs I'm expecting to take a look at this by the end of the week (hopefully tomorrow) - it's already on my TODO list but it's relatively large. Unfortunatelly I'm working on IoT repo only part time 😞

andrasfuchs:
@krwq Alright, no worries, I appreciate your help.

src/devices/Ad7193/Ad7193.cs
krwq: See i.e. https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Pwm/PwmChannel.Linux.cs#L1-L3

We got guidelines from higher up that every `.cs` file should have it

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+
+        /// <summary>
+        /// The external reference voltage value. The default is 2.5V on REFIN1+ and REFIN1- (on the Digilent Pmod AD5 board)
+        /// </summary>
+        public double VoltageReference { get; set; } = 2.50f;
+
+        /// <summary>
+        /// Polarity select bit: true is unipolar, false is bipolar operation.
+        /// </summary>
+        public bool Unipolar
```
krwq, line 52: it would be preferred to create an enum `Polarity` with both values although bool is alright too

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            set
+            {
+                SetRegisterBits(Register.Configuration, BitMask.ConfigurationG, (uint)value);
+            }
+        }
+
+        /// <summary>
+        /// The value of the current Status register
+        /// </summary>
+        public uint Status
```
krwq, line 84: This should be an enum or API should be internal if it's used anywhere. uint requires users to look in the spec for status codes

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            {
+                _registerCache[(byte)Register.Status] = ReadRegisterValue(Register.Status);
+                return _registerCache[(byte)Register.Status];
+            }
+        }
+
+        /// <summary>
+        /// The value of the current Mode register
+        /// </summary>
+        public uint Mode
```
krwq, line 96: enum please

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            {
+                _registerCache[(byte)Register.Mode] = ReadRegisterValue(Register.Mode);
+                return _registerCache[(byte)Register.Mode];
+            }
+        }
+
+        /// <summary>
+        /// The value of the current Configuration register
+        /// </summary>
+        public uint Config
```
krwq, line 108: enum or internal

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+                return _registerCache[(byte)Register.Configuration];
+            }
+        }
+
+        /// <summary>
+        /// Power-down mode. In power-down mode, all AD7193 circuitry, except the bridge power-down switch, is powered down. The bridge power-down switch remains active because the user may need to power up the sensor prior to powering up the AD7193 for settling reasons. The external crystal, if selected, remains active.
+        /// </summary>
+        public bool IsPowerDown
+        {
+            get
```
krwq, line 122: no setter?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+        /// </summary>
+        public bool HasErrors
+        {
+            get
+            {
+                return GetRegisterBits(Register.Status, BitMask.StatusERR) == 0b1;
+            }
+        }
+
+        private bool _continuousRead = false;
```
krwq, line 151: all fields should be in the beginning of the class

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            set
+            {
+                SetRegisterBits(Register.Configuration, BitMask.ConfigurationPseudo, (uint)value);
+            }
+        }
+
+        /// <summary>
+        /// Changes the currently selected channel
+        /// </summary>
+        public Channel ActiveChannels
```
krwq, line 222: Should this be called `SelectedChannel` instead?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            set
+            {
+                SetRegisterBits(Register.Mode, BitMask.ModeAVG, (uint)value);
+            }
+        }
+
+        /// <summary>
+        /// Set the filter output data rate select bits. The 10 bits of data programmed into these bits determine the filter cutoff frequency, the position of the first notch of the filter, and the output data rate for the part.
+        /// </summary>
+        public ushort Filter
```
krwq, line 254: Please rename it to Low/HighPassFilterHz and make units clear. If unit is not Hz or multiple it would be better if you did translation here

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            }
+
+            set
+            {
+                SetRegisterBits(Register.Mode, BitMask.ModeFS, value);
+            }
+        }
+
+        /// <summary>
+        /// The value of the offset calibration register
```
krwq, line 268: Is there a section in Readme explaining how to do it and what this actually means? Please make the units clear and ideally rename property so it has `Calibration` somewhere in there

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            set
+            {
+                SetRegisterBits(Register.Offset, BitMask.Offset, value);
+            }
+        }
+
+        /// <summary>
+        /// The value of the full-scale calibration register
+        /// </summary>
+        public uint FullScale
```
krwq, line 286: same comment as for Offset

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Gets the metadata class for the device
+        /// </summary>
+        /// <returns></returns>
+        public static IDeviceMetadata GetDeviceMetadata()
+        {
+            return Ad7193.metadata;
```
krwq, line 305: `metadata` should be Capitalized if it's public/protected or `_underscored` if it's private.

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+        }
+
+        /// <summary>
+        /// Resets the ADC
+        /// </summary>
+        public void Reset()
+        {
+            for (int i = 0; i < 6; i++)
+            {
+                _spiDevice.Write(new byte[] { 0xFF });
```
krwq, line 336: 2 requests:
- can this be done by writing 6 bytes in 1 write instead
- please change to use Span. I.e.:

```csharp
public void Reset()
{
    Span<byte> bytes = stackalloc byte[6];
    bytes.Fill(0xFF); // or just use it as an array, i.e. bytes[0] = 0xFF;
    // ...
}
```

this way you don't allocate 6 arrays on this call

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Initiates internal calibration, including zero-scale and full-scale calibrations
+        /// </summary>
+        public void Calibrate()
```
krwq, line 343: would it make sense to make calibration take offset and full scale instead of having two properties? This way it's clearer what those are for. Most of the time you set both fields anyway when you need to calibrate

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+
+        /// <summary>
+        /// Waits after a single conversion until the DOUT/!RDY goes low to indicate the completion of a conversion
+        /// </summary>
+        public void WaitForADC()
+        {
+            while (!IsReady)
+            {
+                ReadRegisterValue(Register.Status);
+                Thread.Sleep(5);
```
krwq, line 362: where is the sleep value coming from? Is the number documented in the spec? If so please mention the page

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Initiates Single Conversion (device will go into low power mode when conversion complete, and DOUT/!RDY goes low to indicate the completion of a conversion)
+        /// </summary>
+        public void StartSingleConversion()
+        {
+            // set the single conversion mode bits
+            SetRegisterBits(Register.Mode, BitMask.ModeMD, 0b001);
```
krwq, line 372: what if you're in the middle of conversion already and call this?

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+
+            _stopWatch.Restart();
+            Task samplingTask = Task.Run(() =>
+            {
+                long samples = 0;
+                long nextSampleAt = _stopWatch.ElapsedTicks;
+                long elapsedTicks = 0;
+                long jitter = 0;
+                long maxJitterCorrectionPerSample = Math.Max(Stopwatch.Frequency / 100000, 1);
+                while (_stopWatch.IsRunning)
```
krwq, line 415: How does user get value here? How does it know it got updated? How does user stop this loop? Should you create an event which you call in here? I'd prefer we didn't have any threads in the binding - it's usually better to add helper methods after the binding is complete

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+            }
+
+            double voltage = 0;
+
+            // 0 - bipolar (ranges from ±19.53 mV to ±2.5 V) ; 1 - unipolar (ranges from 0 mV to 19.53 mV to 0 V to 2.5 V)
+            if (Unipolar)
+            {
+                voltage = (double)adcValue / 16777216.0;
+            }
+            else if (!Unipolar)
```
krwq, line 530:
```suggestion
            else
```

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+
+            StartSingleConversion();
+
+            WaitForADC();
+
+            return TryReadADCValue(out uint rawValue) ? rawValue : (uint?)null;
+        }
+
+        /// <summary>
+        /// Converts the raw ADC result to Volts
```
krwq, line 494: Please mention the range depending on Uni/BiPolar setting

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+
+            return voltage;
+        }
+
+        /// <summary>
+        /// Converts the ADC result to Celsius
+        /// </summary>
+        /// <param name="adcValue">The raw ADC result</param>
+        /// <returns></returns>
+        protected Temperature ADCValueToTemperature(ulong adcValue)
```
krwq, line 545: if this is for analog thermometers then please remove the helper. We should decide on the helpers once we got abstraction in place

src/devices/Ad7193/Ad7193.cs
krwq: I'd prefer we removed all the threads and helpers which are not essential. This logic should be made on top of ADC. Even though we currently don't have abstraction we could plug that into it's still better to design APIs so that they don't have anything which adds too much complexity

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+                ReadRegisterValue(Register.Offset),
+                ReadRegisterValue(Register.FullScale)
+            };
+
+            return result;
+        }
+
+        private static int GetCommAddress(int x)
+        {
+            return ((x) & 0x07) << 3;
```
krwq, line 625: brackets around x are not necessary

src/devices/Ad7193/Ad7193.cs
krwq: I'd prefer we kept the debugging stuff for library implementers not for everyone

src/devices/Ad7193/Ad7193.cs
```diff
@@ -0,0 +1,842 @@
...
+                value &= (uint)bitmask;
+                result |= value;
+            }
+
+            WriteRegisterValue(reg, result);
+
+            return result;
+        }
+
+        private uint ByteArrayToUInt32(byte[] buffer)
```
krwq, line 805: please remove those helpers and use BinaryPrimitives instead which have utilities for all these

src/devices/Ad7193/AdcValue.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace Iot.Device.Ad7193
```
krwq, line 1: copyright header

src/devices/Ad7193/AdcValueReceivedEventArgs.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
```
krwq, line 1: copyright header

src/devices/Ad7193/Channel.cs
```diff
@@ -0,0 +1,79 @@
+﻿using System;
```
krwq, line 1: copyright header

src/devices/Ad7193/Channel.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// </summary>
+    [Flags]
+    public enum Channel
+    {
+        /// <summary>
+        /// Channel CH0
+        /// (Pseudo Bit = 0) -> +:AIN1 and -:AIN2
+        /// (Pseudo Bit = 1) -> +:AIN1 and -:AINCOM
+        /// </summary>
+        CH00 = 0b00_0000_0001,
```
krwq, line 16: Channel0, Channel1 etc.. also for bits I'd be easier to read if you used `1 << 0`, `1 << 1`, `1 << 2`, ...

src/devices/Ad7193/Channel.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// Temperature sensor
+        /// </summary>
+        TEMP = 0b01_0000_0000,
+
+        /// <summary>
+        /// Short
+        /// (Pseudo Bit = 0) -> +:AIN2 and -:AIN2
+        /// (Pseudo Bit = 1) -> +:AINCOM and -:AINCOM
+        /// </summary>
+        Shrt = 0b10_0000_0000
```
krwq, line 77: what is short channel?

src/devices/Ad7193/Channel.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// Channel CH7
+        /// (Pseudo Bit = 0) -> +:AIN7 and -:AIN8
+        /// (Pseudo Bit = 1) -> +:AIN8 and -:AINCOM
+        /// </summary>
+        CH07 = 0b00_1000_0000,
+
+        /// <summary>
+        /// Temperature sensor
+        /// </summary>
+        TEMP = 0b01_0000_0000,
```
krwq, line 70: so it is builtin 😄

src/devices/Ad7193/Gain.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace Iot.Device.Ad7193
+{
+    /// <summary>
+    /// PGA Gain settings
+    /// </summary>
+    public enum Gain
+    {
+        /// <summary>
+        /// PGA gain off
```
krwq, line 9: I'd prefer you labeled that as 1x. `Off` might suggest 0

src/devices/Ad7193/Metadata/Ad7193Metadata.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Device.Spi;
+using System.Text;
+
+namespace Ad7193.Metadata
```
krwq, line 6: does this need separate namespace?

src/devices/Ad7193/Metadata/Ad7193Metadata.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    /// <summary>
+    /// Metadata class for AD7193
+    /// </summary>
+    public class Ad7193Metadata : IDeviceMetadata, ISpiDeviceMetadata, IAdcDeviceMetadata
+    {
+        /// <inheritdoc/>
+        public string Manufacturer => "Analog Devices";
```
krwq, line 14: Looking at the content, I'd prefer you put most of those in the readme.

src/devices/Ad7193/Metadata/Ad7193Metadata.cs
```diff
@@ -0,0 +1,46 @@
...
+        public string ProductCategory => "ADC";
+
+        /// <inheritdoc/>
+        public string ProductDescription => "4-Channel, 4.8 kHz, Ultralow Noise, 24-Bit Sigma-Delta ADC with PGA";
+
+        /// <inheritdoc/>
+        public string DataSheetURI => "https://www.analog.com/media/en/technical-documentation/data-sheets/AD7193.pdf";
+
+        /// <inheritdoc/>
+        public SpiMode ValidSpiModes => SpiMode.Mode3;
```
krwq, line 29: You should put all SPI settings directly on the ADC class, i.e.:
```
const int DefaultSpiFrequency =  ...;
const SpiMode DefaultSpiMode = ...;
```

I'm not sure about adding validation for Mode3. We can leave this logic for now but some devices are sometimes embedded in others and some settings occasionally change while device is still the same

src/devices/Ad7193/Metadata/ISpiDeviceMetadata.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Device.Spi;
+using System.Text;
+
+namespace Ad7193.Metadata
```
krwq, line 6: please remove all interfaces with just one implementation. If you'd like the interface to get added for all ADCs please create separate issue.

src/devices/Ad7193/Ad7193.cs
joepr25: I agree with @krwq here, basically we try device bindings to abstract as much as possible everything related to the actual communication between the device as well as the data-sheet contents, because the whole point of them is to have easy methods to use and only exposing what adds functionality of the sensor/device. Having public API like this will just make it harder to use.

src/devices/Ad7193/Register.cs
```diff
@@ -0,0 +1,53 @@
+﻿namespace Iot.Device.Ad7193
```
krwq, line 1: copyright header - here and everywhere where it's missing

src/devices/Ad7193/Register.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// AD7193 Register Map
+    /// </summary>
+    public enum Register : byte
```
krwq, line 6: this should be internal

src/devices/Ad7193/category.txt
```diff
@@ -0,0 +1,3 @@
+adc
+spi
+sigma-delta
```
krwq, line 3: sigma-delta? please remove spi, spi is currently for things exposing SPI, i.e.: SoftwareSpi or FT4222

src/devices/Ad7193/samples/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+		Console.WriteLine($"Channel {adcValue.Channel.ToString().PadLeft(2)}: {adcValue.Voltage.ToString("0.0000").PadLeft(11)} V | {adcValue.Raw.ToString("N0").PadLeft(13)} | {sps.ToString("N1").PadLeft(9)} SPS");
+	}
+}
+```
+
+## Result
+You can see that the values are oscillating between 0.0 V and 1.0 V at the rate of 4800 SPS, as expected.
+
+![Results in a PowerShell window](PowerShell_RaspberryPi_Output.png)
```
krwq, line 140: cannot put comment there:
- please remove the checked in DLL
- please make sure all pictures are less than 100kB (i.e., convert to JPEG, lower quality of JPEG)

andrasfuchs:
@krwq Thank you for the feedback!

@krwq, @joperezr Thank you guys for spending time on this PR. At this point I feel like that I can't justify more working hours on this, because it functions well for all my use cases, and I'm already behind on my project, because of the many requested changes on this PR.

If you feel like joining in, and modify the PR according to your suggestions, you are more than welcome, otherwise please leave open the PR for a few more weeks, and if I can spare some time, I will work on it a little more.

krwq:
Thanks for letting us know @andrasfuchs. Take your time on that, if anyone needs this binding shortly they can help 😄

src/devices/Ad7193/Ad7193.cs
andrasfuchs: The thermometer is part of the AD7193 IC.

src/devices/Ad7193/samples/README.md
andrasfuchs: I reduced the JPG file sizes, but the DLL is crucial for the deployment script to work, so I left it there.

src/devices/Ad7193/category.txt
andrasfuchs: Sigma-delta is a one of the two common types of ADCs.

src/devices/Ad7193/samples/README.md
krwq: @andrasfuchs:
- we cannot have dlls in the repo, this is because no one can easily verify fidelity and also adds lots of risk (i.e. what if your machine is infected and the dll is also inadvertently infected - everyone using the script will be infected)
- all scripts should not be in the device samples - you should put them in the documentation with instructions and you can provide a link of where to get the dll not the dll itself, we should not do that per binding and rather globally since those won't differ per binding

krwq:
@andrasfuchs there still seem to be some unaddressed comments here

andrasfuchs:
> @andrasfuchs there still seem to be some unaddressed comments here

Sorry, I don't think I'll be able to work on this in the near future. 
Feel free to make the suggested modifications or discard the pull request.

joperezr:
[Triage] PR seems abandoned and author won't have time to finish it. We will create an issue for this binding and attach the progress made on this PR so if anybody is interested in the future they can build upon the progress that was made here. We will close this PR next week if there is no activity by then.

PR #1026, opened 2020-03-25
Added support for MCP9808

HakanL:
Added support for temperature sensor MCP9808. Used LM75 as template.

### DIFF
--- a/src/devices/Device-Index.md
+++ b/src/devices/Device-Index.md
@@ -36,6 +36,7 @@
 * [Mcp25xxx device family - CAN bus](Mcp25xxx/README.md)
 * [Mcp3428 - Analog to Digital Converter (I2C)](Mcp3428/README.md)
 * [MCP3xxx family of Analog to Digital Converters](Mcp3xxx/README.md)
+* [Mcp9808 - Microchip's MCP9808 I2C Temp sensor](Mcp9808/README.md)
 * [MLX90614 - Infra Red Thermometer](Mlx90614/README.md)
 * [MotorHat](MotorHat/README.md)
 * [MPR121 - Proximity Capacitive Touch Sensor Controller](Mpr121/README.md)
--- a/src/devices/Device-Index.md
+++ b/src/devices/Device-Index.md
@@ -36,7 +36,6 @@
 * [Mcp25xxx device family - CAN bus](Mcp25xxx/README.md)
 * [Mcp3428 - Analog to Digital Converter (I2C)](Mcp3428/README.md)
 * [MCP3xxx family of Analog to Digital Converters](Mcp3xxx/README.md)
-* [Mcp9808 - Microchip's MCP9808 I2C Temp sensor](Mcp9808/README.md)
 * [MLX90614 - Infra Red Thermometer](Mlx90614/README.md)
 * [MotorHat](MotorHat/README.md)
 * [MPR121 - Proximity Capacitive Touch Sensor Controller](Mpr121/README.md)
--- a/src/devices/Mcp9808/Mcp9808.cs
+++ b/src/devices/Mcp9808/Mcp9808.cs
@@ -57,13 +57,18 @@ namespace Iot.Device.Mcp9808
             _i2cDevice = i2cDevice;
 
             Disabled = false;
+
+            if (!Init())
+            {
+                throw new Exception("Unable to identify manufacturer/device id");
+            }
         }
 
         /// <summary>
         /// Checks if the device is a MCP9808
         /// </summary>
         /// <returns>True if device has been correctly detected</returns>
-        public bool Init()
+        private bool Init()
         {
             if (Read16(Register8.MCP_MANUF_ID) != 0x0054)
             {
@@ -93,6 +98,8 @@ namespace Iot.Device.Mcp9808
         public void Wake()
         {
             SetShutdown(false);
+
+            // Sleep 250 ms, which is the typical temperature conversion time for the highest resolution (page 3 in datasheet, tCONV)
             Thread.Sleep(250);
         }
 
--- a/src/devices/Mcp9808/Mcp9808.cs
+++ b/src/devices/Mcp9808/Mcp9808.cs
@@ -42,7 +42,6 @@ namespace Iot.Device.Mcp9808
             set
             {
                 SetShutdown(value);
-                _disable = value;
             }
         }
 
@@ -152,6 +151,8 @@ namespace Iot.Device.Mcp9808
             }
 
             Write16(Register8.MCP_CONFIG, curVal);
+
+            _disable = isShutdown;
         }
 
         /// <summary>
@@ -171,7 +172,7 @@ namespace Iot.Device.Mcp9808
         internal ushort Read16(Register8 reg)
         {
             _i2cDevice.WriteByte((byte)reg);
-            var buf = new byte[2];
+            Span<byte> buf = stackalloc byte[2];
             _i2cDevice.Read(buf);
 
             return (ushort)(buf[0] << 8 | buf[1]);
--- a/src/devices/Mcp9808/Mcp9808.cs
+++ b/src/devices/Mcp9808/Mcp9808.cs
@@ -1,0 +1,170 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.I2c;
+using Iot.Units;
+using System.Threading;
+
+namespace Iot.Device.Mcp9808
+{
+    /// <summary>
+    /// Microchip's MCP9808 I2C Temp sensor
+    /// </summary>
+    public class Mcp9808 : IDisposable
+    {
+        private I2cDevice _i2cDevice;
+
+        /// <summary>
+        /// MCP9808 I2C Address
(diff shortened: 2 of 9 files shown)


src/devices/Device-Index.md
```diff
@@ -36,6 +36,7 @@
 * [Mcp25xxx device family - CAN bus](Mcp25xxx/README.md)
 * [Mcp3428 - Analog to Digital Converter (I2C)](Mcp3428/README.md)
 * [MCP3xxx family of Analog to Digital Converters](Mcp3xxx/README.md)
+* [Mcp9808 - Microchip's MCP9808 I2C Temp sensor](Mcp9808/README.md)
```
krwq: you should not touch this file, it's auto-generated

src/devices/Mcp9808/Mcp9808.cs
```diff
@@ -0,0 +1,199 @@
...
+            _i2cDevice = i2cDevice;
+
+            Disabled = false;
+        }
+
+        /// <summary>
+        /// Checks if the device is a MCP9808
+        /// </summary>
+        /// <returns>True if device has been correctly detected</returns>
+        public bool Init()
```
krwq: I think this should be private, I'd call it in the constructor and throw if doesn't match

src/devices/Mcp9808/Mcp9808.cs
```diff
@@ -0,0 +1,199 @@
...
+            return Read8(Register8.MCP_RESOLUTION);
+        }
+
+        /// <summary>
+        /// Wakes-up the device
+        /// </summary>
+        public void Wake()
+        {
+            SetShutdown(false);
+            Thread.Sleep(250);
```
krwq: Please point to the spec page for where this delay is coming from

HakanL:
All good comments, thank you, I will work on those.

HakanL:
Please re-check.

src/devices/Mcp9808/Mcp9808.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// </summary>
+        public bool Disabled
+        {
+            get
+            {
+                return _disable;
+            }
+            set
+            {
+                SetShutdown(value);
```
joperezr: NIT: should we check here `if (value != _disable)` so that we only do it when we are changing the value?

src/devices/Mcp9808/Mcp9808.cs
```diff
@@ -0,0 +1,206 @@
...
+        {
+            return Read8(Register8.MCP_RESOLUTION);
+        }
+
+        /// <summary>
+        /// Wakes-up the device
+        /// </summary>
+        public void Wake()
+        {
+            SetShutdown(false);
```
joperezr: Should Wake() method be changing the value of _disabled?

src/devices/Mcp9808/Mcp9808.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// Read MCP9808 Temperature (℃)
+        /// </summary>
+        /// <returns>Temperature</returns>
+        private double GetTemperature()
+        {
+            ushort value = Read16(Register8.MCP_AMBIENT_TEMP);
+
+            if (value == 0xFFFF)
+            {
+                return double.NaN;
```
joperezr: Is this an error state? should we throw an exception in this case?

src/devices/Mcp9808/Mcp9808.cs
```diff
@@ -0,0 +1,206 @@
...
+            }
+
+            double temp = value & 0x0FFF;
+            temp /= 16.0;
+            if ((value & 0x1000) != 0)
```
joperezr: Do you mind also adding a comment into what does this check mean? Is that to catch an overflow?

src/devices/Mcp9808/Mcp9808.cs
HakanL: I don't think there's any harm in setting the registry twice, maybe for some reason you'd want to re-send the reg-write?

src/devices/Mcp9808/Mcp9808.cs
```diff
@@ -0,0 +1,206 @@
...
+            return (Register16)Read16(reg);
+        }
+
+        internal ushort Read16(Register8 reg)
+        {
+            _i2cDevice.WriteByte((byte)reg);
+            var buf = new byte[2];
+            _i2cDevice.Read(buf);
+
+            return (ushort)(buf[0] << 8 | buf[1]);
```
joperezr: Is this endianess going to be correct on all machines? Usually we suggest to instead of manually doing this, to call BinaryPrimitives.ReadInt16BigEndian or BinaryPrimitives.ReadInt16LittleEndian which will handle that on all precessors for you.

src/devices/Mcp9808/Mcp9808.cs
HakanL: Good point, fixed.

src/devices/Mcp9808/Mcp9808.cs
HakanL: I just copied it from an Arduino library for the MCP9808. I assume there's some reason to it in the datasheet, but didn't look into that.

src/devices/Mcp9808/Mcp9808.cs
HakanL: It's not related to the machine's endianess, it's how the MCP9808 is sending it, high byte first, then low byte.

src/devices/Mcp9808/Mcp9808.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        internal Register16 ReadRegister16(Register8 reg)
+        {
+            return (Register16)Read16(reg);
+        }
+
+        internal ushort Read16(Register8 reg)
+        {
+            _i2cDevice.WriteByte((byte)reg);
+            var buf = new byte[2];
```
joperezr: NIT: you don't have to change it but if you want to reduce allocations on the binding, you could also just use the stack instead by doing something like:

```c#
Span<byte> buf = stackalloc byte[2];
```

src/devices/Mcp9808/Mcp9808.cs
joperezr: I see, now that you mention that, can you check if that Arduino library has permisive license? If we are porting code from an existing open source library, we are required to make sure that the ported code is compatible with our license. If you point me to the source of that library I can take a look if it does.

src/devices/Mcp9808/Mcp9808.cs
HakanL: It means it doesn't have a current temperature ready yet. The Arduino library (Adafruit) I copied the logic from returned NaN, seemed like a good idea, but I'm open to suggestions.

joperezr:
Left a few comments but this looks great otherwise. Thanks a bunch for your contribution @HakanL

src/devices/Mcp9808/Mcp9808.cs
HakanL: Good idea, I tried that first but got a build error, maybe it just wasn't set up correctly, it worked now, I will change it.

HakanL:
More fixes pushed, please re-check.

src/devices/Mcp9808/Mcp9808.cs
joperezr: I see, that is fine then, the problem here is that I'm not 100% sure that our Temperature Units class will handle that, and the property will try to convert this value into a Temperature assuming this was Celcius.

joperezr: Assuming our Temperature class does handle it, then perhaps all we need to do here is document the property to say that if you get a nan value back (or whatever FromCelcius(double.NAN) will return) then it means that the current temperature is not ready yet.

src/devices/Mcp9808/Mcp9808.cs
HakanL: I didn't actually port their library, I just took these 3 lines and re-wrote them to C# (plus the part with NaN). But here it is: https://github.com/adafruit/Adafruit_MCP9808_Library

src/devices/Mcp9808/Mcp9808.cs
HakanL: I can do either way, document or throw exception, whatever you think is best :)

src/devices/Mcp9808/Mcp9808.cs
joperezr: I see, that's ok though since it looks like they are under BSD license which means that is ok 😄

src/devices/Mcp9808/Mcp9808.cs
joperezr: I think documenting is fine, let's just make sure that Temperature.FromCelcius will work when we send it NaN

HakanL: I have added a comment to this, plus I tested the logic in Temperature with NaN, it works fine, it will return NaN even after * and +.

src/devices/Mcp9808/Mcp9808.cs
HakanL: IMHO these classes are fairly low-level and shouldn't have too much logic/"optimization", as long as it doesn't cause a fault-state within the device if the registry is written to multiple times.

PR #1033, opened 2020-04-02
Adding PN5180 RFID/NFC reader

Ellerbach:
Adding PN5180 RFID/NFC reader:
- Full communication thru SPI and GPIO Controller
- Full documentation
- Full example including leveraging existing implementation of Mifare and Credit Card to cover and show how to use the public functions and properties
- Full support with non native SPI and GPIO Controllers like FT4222
- ISO 14443 Type A and Type B full support
- EEPROM, Radio Frenquence access and manafgement
- Status of current implementation in the main read me


src/devices/Pn5180/Command.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
```
krwq, line 1: license header

Ellerbach: Yes, missing everywhere. Will add everywhere

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+    /// </summary>
+    public class Pn5180 : CardTransceiver, IDisposable
+    {
+        private const int TimeoutWaiting = 10_000;
+
+        private readonly SpiDevice _spiDevice;
+        private readonly GpioController _gpioController;
+        private bool _shouldDispose;
+        private int _pinBusy;
+        private int _pinNss;
```
krwq, line 31: is this enable/disable pin? if so please make sure everything works if it's hard-wired (or -1 passed and pin connected to whatever makes it enabled)

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+
+        private static List<SelectedPiccInformation> _activeSelected = new List<SelectedPiccInformation>();
+
+        /// <summary>
+        /// A radio Frequence configuration element size is 5 bytes
+        /// Byte 1 = Register Address
+        /// next 4 bytes = data of the register
+        /// </summary>
+        public const int RadioFrequenceConfiguraitonSize = 5;
```
krwq, line 40: I think this should be

```suggestion
        public const int RadioFrequencyConfiguraitonSize = 5;
```
same for comment above

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+
+        /// <summary>
+        /// PN532 SPI Clock Frequency
+        /// </summary>
+        public const int SpiClockFrequency = 7_000_000;
```
krwq, line 45: should this be `DefaultSpiClockFrequency`?

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+
+        /// <summary>
+        /// Only SPI Mode supported is Mode0
+        /// </summary>
+        public const SpiMode SpiMode = System.Device.Spi.SpiMode.Mode0;
```
krwq, line 50: DefaultSpiMode?

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            _gpioController = gpioController ?? new GpioController(PinNumberingScheme.Logical);
+            _shouldDispose = shouldDispose;
+            _pinBusy = pinBusy;
+            _pinNss = pinNss;
+            _gpioController.OpenPin(_pinBusy, PinMode.Input);
+            _gpioController.OpenPin(_pinNss, PinMode.Output);
+
+            // Check the version
+            var (product, firmware, eeprom) = GetVersion();
+            if ((product.Major == 0) || (firmware.Major == 0) || (eeprom.Major == 0))
```
krwq, line 94: should this be `&&` or `>= someVersion`? I suspect 1.0.0 should not create an exception, right?

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+        public void Dispose()
+        {
+            // If the RF field is on, try to switch it off
+            try
+            {
+                SetRadioFrequence(false);
+            }
+            catch (Exception)
+            {
+                // We don't do anything on purpose. Trying to switch off
```
krwq, line 112: when would above throw? can we make it not throw and i.e. return some kind of status or bool flag? (i.e. TrySetRadioFrequency)

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            }
+
+            if (_shouldDispose)
+            {
+                _spiDevice?.Dispose();
```
krwq, line 117: I'm slightly not sure about it. I think there might be a common case where you might want to dispose SPI but not GPIO, I guess all or nothing makes more sense as the message is clearer and we recommend to use 1 GpioController per device

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            }
+        }
+
+        #region EEPROM
+
+        /// <summary>
+        /// Get the Product, Firmware and EEPROM versions of the PN8150
+        /// </summary>
+        /// <returns>A tuple with the Product, Firmware and EEPROM versions</returns>
+        public (Version product, Version firmware, Version eeprom) GetVersion()
```
krwq, line 128: `GetVersions` or 3 separate APIs

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            var eeprom = new Version(versionAnswer[5], versionAnswer[4]);
+            return (product, firmware, eeprom);
+        }
+
+        /// <summary>
+        /// Get the PN5180 identifier, this is a 16 byte long
+        /// </summary>
+        /// <param name="identifier">A 16 byte buffer</param>
+        /// <returns>True if success</returns>
+        public bool GetIdentifier(Span<byte> identifier)
```
krwq, line 149: might be worth putting word `output` somewhere in the param name

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+        }
+
+        /// <summary>
+        /// Read the full EEPROM
+        /// </summary>
+        /// <param name="eeprom">At least 255 bytes buffer</param>
+        /// <returns>True if success</returns>
+        public bool ReadAllEeprom(Span<byte> eeprom)
+        {
+            if (eeprom.Length > 255)
```
krwq, line 166: the description says `at least 255`, this says `less than 255`

krwq: Also since you used the word ReadAll I'm fine with `eeprom.Length != 255`

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            if ((byte)address + eeprom.Length > 255)
+            {
+                throw new ArgumentException($"Size of EEPROM is 255 bytes, {nameof(address)} + {nameof(eeprom)} must be less than 255");
+            }
+
+            Span<byte> dumpEeprom = stackalloc byte[3];
+            dumpEeprom[0] = (byte)Command.READ_EEPROM;
+            dumpEeprom[1] = (byte)address;
+            dumpEeprom[2] = (byte)eeprom.Length;
+            LogInfo.Log($"{nameof(ReadEeprom)}, {nameof(dumpEeprom)}: {BitConverter.ToString(dumpEeprom.ToArray())}", LogLevel.Debug);
```
krwq, line 206: slightly bothersome that people opting out of logging pay the ToArray and ToString cost as well. Fine in this PR. If we ever get complain about this or memory related with usage of this library then we can fix.

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+        /// <summary>
+        /// Read data from a card.
+        /// </summary>
+        /// <param name="toRead">>The span of byte to read</param>
+        /// <param name="expectedToRead">The expected number of bytes to read</param>
+        /// <returns>True if success. Will return false if the number of bytes to read is not the same as the expected number to read</returns>
+        /// <remarks>Using this function you'll have to manage yourself the possible low level communication protocol.
+        /// This function write directly to the card all the bytes. Please make sure you'll first load specific radio frequence settings,
+        /// detect a card, select it and then send data</remarks>
+        public bool ReadDataFromCard(Span<byte> toRead, int expectedToRead)
```
krwq, line 344: I'd remove expectedToRead and force users to use correct span length since it's very cheap to slice. Also expectedToRead might suggest that you might read less and still success

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+                LogInfo.Log($"{nameof(ReadDataFromCard)}: wrong number of expected bytes, clearing the cache", LogLevel.Debug);
+                // Clear all
+                ReadDataFromCard(new byte[numBytes]);
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// REad all the data from the card
```
krwq, line 363:
```suggestion
        /// Read all the data from the card
```

src/devices/Pn5180/Pn5180.cs
krwq: Btw. how would user know how much to read? Usually these APIs are shapped as folowing:

`bool TryReadDataFromCard(Span<byte> toRead, out int bytesWritten)` so that you can pass too long buffer.

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+        /// </summary>
+        /// <returns>A byte array with all the read elements, null if nothing can be read</returns>
+        /// <remarks>Using this function you'll have to manage yourself the possible low level communication protocol.
+        /// This function write directly to the card all the bytes. Please make sure you'll first load specific radio frequence settings,
+        /// detect a card, select it and then send data</remarks>
+        public byte[] ReadAllDatFromCard()
```
krwq, line 369:
```suggestion
        public byte[] ReadAllDataFromCard()
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            {
+                return toRead.ToArray();
+            }
+
+            return null;
+        }
+
+        /// <summary>
+        /// Get the number of bytes to read and the valid number of bits in the last byte
+        /// If the full byte is valid then the value of the valid bit is 0
```
krwq, line 390: you might want to mention if validBits==3 means 3 most significant or least significant bits

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+                        toReceive.Slice(2, numBytes - 2).CopyTo(dataFromCard);
+                        // Send Ack with the right mark
+                        card.LastBlockMark = !card.LastBlockMark;
+                        Span<byte> rBlockChain = new byte[2] { (byte)(0b1010_1010 | (card.LastBlockMark ? 1 : 0)), targetNumber };
+                        var numBytes2 = TranceiveTypeB(card, rBlockChain, toReceive);
+
+                        if (numBytes2 >= 2)
+                        {
+                            // If not the right target, then not for us
+                            if (!(toReceive[1] == targetNumber))
```
krwq, line 512:
```suggestion
                            if (toReceive[1] != targetNumber)
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+                            {
+                                return -1;
+                            }
+
+                            if (!(toReceive[0] == (0b0000_1010 | (card.LastBlockMark ? 1 : 0))))
```
krwq, line 517: !=

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+                        }
+
+                        numBytes = ReadWithTimeout(dataFromCard, (int)(toReceive[2] * card.Card.FrameWaitingTime / 1000));
+                        goto IBlock;
+                    }
+                }
+
+                return numBytes;
+            }
+            else
```
krwq, line 547: I'm not sure where this `if .. else` start but you might want to swap the condition there so else is first (easier to read)

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            {
+                // Case of a type A card
+                return TranceiveBuffer(dataToSend, dataFromCard);
+            }
+
```
krwq, line 552: extra enter 😄

src/devices/Pn5180/Pn5180.cs
pgrawehr: This is still missspelled ;-)

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+        /// <param name="configuration">A span of bytes for the configuration. Should be a multiple of 5 with the size of <see ref="GetRadioFrequenceConfigSize"/></param>
+        /// <returns>True if success</returns>
+        public bool RetrieveRadioFrequenceConfiguration(ReceiverRadioFrequenceConfiguration receiver, Span<byte> configuration) => RetrieveRadioFrequenceConfiguration((byte)receiver, configuration);
+
+        private bool RetrieveRadioFrequenceConfiguration(byte config, Span<byte> configuration)
+        {
+            // Page 41 documentation PN5180A0XX-C3.pdf
+            if ((configuration.Length > 195) || (configuration.Length % 5 != 0) || (configuration.Length == 0))
+            {
+                throw new ArgumentException($"Configuration buffer size must be a multiple of 5 and no larger than 195 bytes");
```
krwq, line 776:
```suggestion
                throw new ArgumentException($"Configuration buffer size must be a positive multiple of 5 and no larger than 195 bytes");
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+        /// <param name="configuration">A span of bytes for the configuration. Should be a multiple of 5 with the size of <see ref="GetRadioFrequenceConfigSize"/></param>
+        /// <returns>True if success</returns>
+        public bool UpdateRadioFrequenceConfiguration(ReceiverRadioFrequenceConfiguration receiver, Span<byte> configuration) => UpdateRadioFrequenceConfiguration((byte)receiver, configuration);
+
+        private bool UpdateRadioFrequenceConfiguration(byte config, Span<byte> configuration)
+        {
+            // Page 41 documentation PN5180A0XX-C3.pdf
+            if ((configuration.Length > 252) || (configuration.Length % 6 != 0) || (configuration.Length == 0))
+            {
+                throw new ArgumentException($"Configuration buffer size must be a multiple of 5 and no larger than 195 bytes");
```
krwq, line 816:
```suggestion
                throw new ArgumentException($"Configuration buffer size must be a positive multiple of 6 and no larger than 195 bytes");
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            try
+            {
+                SpiReadRegister(Register.RF_STATUS, status);
+            }
+            catch (TimeoutException)
+            {
+                return false;
+            }
+
+            return (status[2] & 0b0000_0100) == 0b0000_0100;
```
krwq, line 901: any chance you could cast this to enum and use HasFlag or at minimum comment what this bit is? disregard if this was ported

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+        }
+
+        private bool SetRadioFrequence(bool field)
```
krwq, line 904: can you make param name a verb?

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+
+            Span<byte> atqa = stackalloc byte[2];
+            Span<byte> irqStatus = stackalloc byte[4];
+            Span<byte> uidSak = stackalloc byte[7];
+            Span<byte> uid = stackalloc byte[10];
+            Span<byte> sak = stackalloc byte[1];
+            Span<byte> sakInterm = stackalloc byte[5];
+            int numBytes;
+
+            DateTime dtTimeout = DateTime.Now.AddMilliseconds(timeoutPollingMilliseconds);
```
krwq, line 953: this code here (this method) looks super similar to PN532, is it possible to unify them both?

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+                    // Read 5 bytes sak. Byte 1 will tell us if we have the full UID or if we need to read more
+                    ReadDataFromCard(sakInterm.Slice(0, 5));
+                    // Switches back on the CRC off in RX and TX direction
+                    CrcReceptionTransfer = true;
+                    // Now finish the anticollision with the new NVB
+                    uidSak[1] = 0x70;
+                    // Add the previous elements to this last anticollision to send
+                    sakInterm.Slice(0, 5).CopyTo(uidSak.Slice(2));
+                    ret &= SendDataToCard(uidSak);
+
```
krwq, line 1006: you can probably `if (!ret) return false` here

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+                rNak[1] = potentioalActive.Card.TargetNumber;
+                ret = SendDataToCard(rNak);
+                var numByts = ReadWithTimeout(rNak, (int)(potentioalActive.Card.FrameWaitingTime / 1000));
+                if ((!ret) || (numByts != 2))
+                {
+                    _activeSelected.Remove(potentioalActive);
+                    continue;
+                }
+
+                if (rNak[1] != potentioalActive.Card.TargetNumber)
```
krwq, line 1091:
```suggestion
                if (rNak[1] != potentialActive.Card.TargetNumber)
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+                }
+
+                // The card is still active, we keep it
+            }
+
+            // Find the first slot available
+            var target = _activeSelected.Select(m => m.Card.TargetNumber);
+            for (int i = 1; i < 14; i++)
+            {
+                if (!target.Where(m => m == i).Any())
```
krwq, line 1110: I think `target.Any(m => m == i)` might be faster

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+                if (!ret)
+                {
+                    // If we can't select the card, we still output the data but we let
+                    // the application know that the card can't be selected
+                    return false;
+                }
+
+                // Add the card to the list
+                var selected = new SelectedPiccInformation() { Card = card, LastBlockMark = false };
+                _activeSelected.Add(selected);
```
krwq, line 1186: do we ever remove it from the list?

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+
+            return true;
+        }
+
+        /// <summary>
+        /// Deselect a 14443 Type B card
+        /// </summary>
+        /// <param name="card">The card to deselect</param>
+        /// <returns>True if success</returns>
+        public bool DeselecCardTypeB(Data106kbpsTypeB card)
```
krwq, line 1235:
```suggestion
        public bool DeselectCardTypeB(Data106kbpsTypeB card)
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            crc[0] = (byte)(crcRet & 0xFF);
+            crc[1] = (byte)(crcRet >> 8);
+        }
+
+        /// <summary>
+        /// Provide a calculation of CRC for Type A cards
+        /// </summary>
+        /// <param name="buffer">The buffer to process</param>
+        /// <param name="crc">The CRC, Must be a 2 bytes buffer</param>
+        public void CalculateCrcA(Span<byte> buffer, Span<byte> crc)
```
krwq, line 1306:
```suggestion
        public void CalculateCrcA(ReadOnlySpan<byte> buffer, Span<byte> crc)
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            {
+                throw new ArgumentException($"CRC A can only be 2 bytes long");
+            }
+
+            var crcRet = CalculateCrc(buffer, 0x6363);
+            crc[0] = (byte)(crcRet & 0xFF);
+            crc[1] = (byte)(crcRet >> 8);
+        }
+
+        private ushort CalculateCrc(Span<byte> buffer, ushort crcB)
```
krwq, line 1318:
```suggestion
        private ushort CalculateCrc(ReadOnlySpan<byte> buffer, ushort crcB)
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+        {
+            // Page 42 of ISO14443-3.pdf
+            for (int i = 0; i < buffer.Length; i++)
+            {
+                var crcInterim = buffer[i];
```
krwq, line 1323: use explicit type

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            // the slave device to latch the data.
+            // The BUSY line is used to indicate that the system is BUSY and cannot receive any data
+            // from a host.Recommendation for the BUSY line handling by the host:
+            // 1.Assert NSS to Low
+            // 2.Perform Data Exchange
+            // 3.Wait until BUSY is high
+            // 4.Deassert NSS
+            // 5.Wait until BUSY is low
+            // Wait for the PN8150 to be ready
+            int timeout = TimeoutWaiting;
```
krwq, line 1424: I'd prefer we measured time instead of cycles. This will vary too greatly depending on where it runs

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1496 @@
...
+            }
+
+            _gpioController.Write(_pinNss, PinValue.Low);
+
+            // In order to write data to or read data from the PN5180, "dummy reads" shall be
+            // performed.The Figure 8 and Figure 9 are illustrating the usage of this "dummy reads" on
+            // the SPI interface.
+            for (int i = 0; i < toRead.Length; i++)
+            {
+                _spiDevice.TransferFullDuplex(new byte[] { 0xFF }, toRead.Slice(i, 1));
```
krwq, line 1479: please preallocate this since this will allocate in a loop right now and it doesn't even change

src/devices/Pn5180/README.md
```diff
@@ -0,0 +1,307 @@
...
+}
+```
+
+The [example](./samples/Pn5180sample.cs) contains as well an implementation to fully dump the content of a credit card.
+
+## Current implementation
+
+Communication support:
+- [X] SPI: fully supported
+- [X] GPIO Controller: fully supported
```
krwq, line 274: do you mean SPI over GPIO (SoftPwm)?

src/devices/Pn5180/SelectedPiccInformation.cs
```diff
@@ -0,0 +1,14 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Iot.Device.Rfid;
+
+namespace Iot.Device.Pn5180
+{
+    internal class SelectedPiccInformation
+    {
+        public Data106kbpsTypeB Card { get; set; }
+        public bool LastBlockMark { get; set; }
```
krwq, line 12: I'm having hard time figuring out what does this mean and how it is PiccInformation, should this actually belong here?

src/devices/Pn5180/samples/Pn5180sample.cs
```diff
@@ -0,0 +1,423 @@
...
+            {
+                Console.WriteLine($"Not a valid choice, please choose the test you want to run");
+            }
+        }
+
+        private static Pn5180 Rpi()
+        {
+            var spi = SpiDevice.Create(new SpiConnectionSettings(0, 1) { ClockFrequency = Pn5180.SpiClockFrequency, Mode = Pn5180.SpiMode, DataFlow = DataFlow.MsbFirst });
+
+            // REset the device
```
krwq, line 94: Reset

src/devices/Pn5180/samples/Pn5180sample.cs
```diff
@@ -0,0 +1,423 @@
...
+            {
+                PullTypeBCards();
+            }
+            else
+            {
+                Console.WriteLine($"Not a valid choice, please choose the test you want to run");
+            }
+        }
+
+        private static Pn5180 Rpi()
```
krwq, line 90: Can you change name to something like HardwareSpi? I suspect there will be new board replacing RPi in 10y and no new dev will even know what Rpi means at that time :)

src/devices/Pn5180/samples/Pn5180sample.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj">
+      <AdditionalProperties>$(AdditionalProperties);RuntimeIdentifier=linux</AdditionalProperties>
+    </ProjectReference>
+  </ItemGroup>
+
+  <ItemGroup>
```
krwq, line 14: Cannot put comment there but can you please replace PNG with JPEG so it's not 1.5MB 😄

src/devices/Pn5180/Pn5180.cs
krwq: :smile: good call

src/devices/Pn5180/Pn5180.cs
Ellerbach: Both pins are needed. Without those pins, the device can't work properly, there won't be any communication at all. I will add a test on negative numbers because for sure, the pin number can't be negative.

src/devices/Pn5180/Pn5180.cs
Ellerbach: It's actually the maximum speed. Will rename accordingly

src/devices/Pn5180/Pn5180.cs
Ellerbach: it should actually. Versions numbers are not 0. they have lots of variety but not 0. Those are 3 different versions, so if any is 0, then for sure it's not a good version. It's not that easy neither to test something >= for any of them. All up, it's a very basic test just to make sure that there is a basic one.

src/devices/Pn5180/Pn5180.cs
Ellerbach: it's actually obsolete, I can remove fully the try/catch. I changed it for the SetRadioFrequency to return a boolean which in this case we won't have to test.

src/devices/Pn5180/Pn5180.cs
Ellerbach: I've mentionned in the /// comments that should dispose will dispose both if true.

src/devices/Pn5180/Pn5180.cs
Ellerbach: I will go for a GetVersions

src/devices/Pn5180/Pn5180.cs
Ellerbach: yes, good catch, will adjust both the description and the condition

src/devices/Pn5180/Pn5180.cs
Ellerbach: I'll check the log level and call the Log function if log level is passed. a comparison will be much less than a conversion. Will fix everywhere

Ellerbach: So something like this will make it:
```
if (LogLevel >= LogLevel.Debug)
            {
                LogInfo.Log($"{nameof(ReadEeprom)}, {nameof(dumpEeprom)}: {BitConverter.ToString(dumpEeprom.ToArray())}", LogLevel.Debug);
            }
```

src/devices/Pn5180/Pn5180.cs
Ellerbach: I will change the ```byte[] ReadAllDataFromCard``` to ```bool ReadDataFromCard(Span<byte> toRead, out int bytesRead)``` which makes more sense.
And I changed to private the ReadData with the expected length. It's used mainly internally.

src/devices/Pn5180/Pn5180.cs
Ellerbach: ho yes, done!

src/devices/Pn5180/Pn5180.cs
Ellerbach: the status for this specific field is just 1 bit out of a 4 bytes configuration. So I'm making the test only once like this. It's clearly documented in the documentation of the device

src/devices/Pn5180/Pn5180.cs
Ellerbach: I made this one as similar as possible as the PN532, keeping the philosophy. Issue with the PN532 is that it's working in a different way as the PN532 returns a byte array that then you have to try to understand what you get as a result. 
But, whenever I'll have a fix to do to the PN532, I can add *similar* functions to the PN532 to listen to specific types of cards. That said, parameters will be a bit different as the PN532 dos not have this concept of receiver and transmitter frequencies settings, it has only 1 for both and the choices as more limited on the PN532. So It won't be the same settings.
But clearly, trying to have a similar approach and similar naming to make it easier to use.

src/devices/Pn5180/Pn5180.cs
Ellerbach: Yes and no, I've seen weird behaviors sometimes which I think I fixed by adjusting correctly the various timings. You may miss writing but still receive something from the previous time. So I4ll just keep it like this and I changed &= for a simple = which makes more sense.

src/devices/Pn5180/Pn5180.cs
Ellerbach: yes, in the test right before if the card is not seen or not responding.

src/devices/Pn5180/README.md
Ellerbach: Will rephrase

src/devices/Pn5180/SelectedPiccInformation.cs
Ellerbach: It's related to the 14443-4 communication protocol. You have a bit changing which is called the Block Mark. And you have to adjust when you write or get a value to read.

src/devices/Pn5180/samples/Pn5180sample.csproj
Ellerbach: It's because there a transparent background. Transparent background are a recommendation now as more and more websites uses dark modes (docs.microsoft.com is now) and pictures are rendering correctly only without the background. That said, I can for sure adjust the size of the picture, no need to have a native 4K size.

src/devices/Pn5180/SelectedPiccInformation.cs
krwq: fair enough, thanks for clarifying!

src/devices/Pn5180/Pn5180.cs
krwq: fair enough

src/devices/Pn5180/Pn5180.cs
krwq: we can fix it in the future if we got it slightly wrong 😄

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1551 @@
...
+            {
+                LogInfo.Log($"{nameof(SendDataToCard)}: {nameof(TimeoutException)} in {nameof(SpiReadRegister)}", LogLevel.Debug);
+                return (-1, -1);
+            }
+        }
+
+        /// <inheritdoc/>
+        public override int Transceive(byte targetNumber, ReadOnlySpan<byte> dataToSend, Span<byte> dataFromCard)
+        {
+            // Check if we have a Mirafare Card authentication request
```
krwq, line 454:
```suggestion
            // Check if we have a Mifare Card authentication request
```

(I think)

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1551 @@
...
+                        numBytes = ReadWithTimeout(dataFromCard, (int)(toReceive[2] * card.Card.FrameWaitingTime / 1000));
+                        goto IBlock;
+                    }
+                }
+
+                return numBytes;
+            }
+        }
+
+        private int TranceiveTypeB(SelectedPiccInformation card, ReadOnlySpan<byte> dataToSend, Span<byte> dataFromCard)
```
krwq, line 594:
```suggestion
        private int TransceiveTypeB(SelectedPiccInformation card, ReadOnlySpan<byte> dataToSend, Span<byte> dataFromCard)
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1551 @@
...
+                if (!ret)
+                {
+                    return -1;
+                }
+            }
+
+            return numBytes;
+        }
+
+        private int TranceiveBuffer(ReadOnlySpan<byte> dataToSend, Span<byte> dataFromCard)
```
krwq, line 647:
```suggestion
        private int TransceiveBuffer(ReadOnlySpan<byte> dataToSend, Span<byte> dataFromCard)
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1551 @@
...
+            }
+
+            return true;
+        }
+
+        #endregion
+
+        #region SPI primitives
+
+        private void SpiWriteRegister(Command command, Register register, Span<byte> data)
```
krwq, line 1432:
```suggestion
        private void SpiWriteRegister(Command command, Register register, ReadOnlySpan<byte> data)
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1551 @@
...
+            Span<byte> toSend = stackalloc byte[2]
+            {
+                (byte)Command.READ_REGISTER,
+                (byte)register
+            };
+            SpiWrite(toSend);
+            SpiRead(readBuffer);
+        }
+
+        private void SpiWriteRead(Span<byte> toSend, Span<byte> toRead)
```
krwq, line 1457:
```suggestion
        private void SpiWriteRead(ReadOnlySpan<byte> toSend, Span<byte> toRead)
```

src/devices/Pn5180/Pn5180.cs
```diff
@@ -0,0 +1,1551 @@
...
+        {
+            SpiWrite(toSend);
+            SpiRead(toRead);
+        }
+
+        private void SpiWrite(Span<byte> toSend)
```
krwq, line 1463:
```suggestion
        private void SpiWrite(ReadOnlySpan<byte> toSend)
```

Issue #1037, opened 2020-04-04
Sample projects fail to properly build on Windows


pgrawehr:
**Describe the bug**
Attempting to execute a sample on Windows (within Visual Studio) results in build errors. 

This is irespective of the fact that many bindings don't work on Windows, just because the hardware is not there. This is about an ordinary Windows 10 x64 installation on a desktop computer. 

**Steps to reproduce**

* Perform a clean checkout and do a "build" from the console. 
* Open src/devices/Dhtxx/DhtSensor.csproj
* Add src/devices/Dhtxx/samples/DhtSensor.sample.csproj 
* Building now basically works, but execution fails with an error from UnixI2cDevice, which shouldn't even be there and definitelly shouldn't be used. So aparently, the linux library is being used. Also, Intellisense / Resharper is not happy and marking all references to System.Device types in red.
* So add a reference to System.Device.Gpio.csproj
* This now adds new solution configurations. Choosing "Debug" behaves as before. So we choose "Windows-Debug"
The build now fails: 
```
1>------ Erstellen gestartet: Projekt: System.Device.Gpio, Konfiguration: Windows-Debug Any CPU ------
1>System.Device.Gpio -> C:\projects\iot4\artifacts\bin\System.Device.Gpio\Windows-Debug\netstandard2.0\win\System.Device.Gpio.dll
2>------ Erstellen gestartet: Projekt: Dhtxx, Konfiguration: Debug Any CPU ------
2>CSC : error CS0006: Metadatendatei "C:\projects\iot4\artifacts\bin\System.Device.Gpio\Windows-Debug\netstandard2.0\ref\System.Device.Gpio.dll" wurde nicht gefunden.
2>Die Erstellung des Projekts "Dhtxx.csproj" ist abgeschlossen -- FEHLER.
3>------ Erstellen gestartet: Projekt: DhtSensor.sample, Konfiguration: Debug Any CPU ------
3>CSC : error CS0006: Metadatendatei "C:\projects\iot4\artifacts\bin\System.Device.Gpio\Windows-Debug\netstandard2.0\ref\System.Device.Gpio.dll" wurde nicht gefunden.
3>Die Erstellung des Projekts "DhtSensor.sample.csproj" ist abgeschlossen -- FEHLER.
========== Erstellen: 1 erfolgreich, 2 fehlerhaft, 0 aktuell, 0 übersprungen ==========
```
[...]

PR #1038, opened 2020-04-05
Remove platform dependency

pgrawehr:
This makes the System.Device.Gpio library build platform-independent. The drivers to use for a specific platform are now determined at runtime. 

The target is to avoid breaking backwards compatibility for existing clients. Whether that is 100% possible will yet need to be determined.

Work in progress. Tasks to be completed:
* [x] Check for side effects (i.e. attempts to load inexistent libraries)
* [x] Check that the public APIs haven't changed (edited: @joperezr has done this)
* [x] Clean up build scripts / Nuget package contents. (edited: @joperezr has done this)
* [X] Nuget Symbols packages should include the PDB files (Looks like that hasn't been the case even before) - Debug information type is "Embedded", so this is unneccessary


src/System.Device.Gpio/Directory.Build.targets
```diff
@@ -1,60 +1,18 @@
 <Project>
-  <!-- Targets for building the notsupported assembly -->
-  <Target Name="SetupEnvironmentForNotSupportedAssembly">
```
joperezr, line 3: There is one more problem with doing this which we haven't discussed, and that is what will happen when somebody tries to run this on platforms like OSX. I think that we should perhaps still throw not supported there, but I'm not sure what is the best way to do that.

src/System.Device.Gpio/Directory.Build.targets
```diff
@@ -1,60 +1,18 @@
...
 
   <Target Name="GetSymbolsForSnupkg" Condition="'$(DebugType)' == 'Portable'">
     <!-- Add the linux and windows specific pdbs -->
-    <ItemGroup>
-      <!-- DebugSymbolsProjectOutputGroupOutput doesn't support PackagePath and instead the output path is set to lib/$(TargetFramework)/$(TargetPath) 
-           Setting TargetPath to ./../../ will get us back to the root of the package, so now we are able to place the pdbs inside
-           the runtimes folder.
-      -->
-      <DebugSymbolsProjectOutputGroupOutput Include="$(OutputPath)linux/$(AssemblyName).pdb">
-        <TargetPath>./../../runtimes/linux/lib/$(TargetFramework)/$(AssemblyName).pdb</TargetPath>
```
joperezr, line 50: Let's just get RID of the target completely as well as the PropertyGroup above given that this target is now empty.

src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -25,17 +21,8 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
   </ItemGroup>
-  <ItemGroup Condition="'$(TargetsLinux)' == 'true'">
```
joperezr, line 28: I can't remember from the top of my head where I was defining TargetsLinux and TargetsWindows (I believe it was on root's Directory.Build.Props) but we should remove that as well.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
```diff
@@ -24,11 +24,23 @@ public GpioController(PinNumberingScheme numberingScheme)
         {
         }
 
+        private static GpioDriver GetBestDriverForBoard()
```
joperezr, line 27: We should just move the constructor as well as these methods back to GpioController.cs. No need to have them split by filename now that we will only have one build.

src/System.Device.Gpio/System/Device/Gpio/Drivers/HummingBoardDriver.Linux.cs
```diff
@@ -1,126 +0,0 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-using System.Threading;
-using System.Threading.Tasks;
-
-namespace System.Device.Gpio.Drivers
-{
-    public partial class HummingBoardDriver : GpioDriver
```
joperezr, line 10: For all of these, we should probably add a check in their constructor to make sure that they are being initialized in a supported platform, so that if somebody tries to initialize  a libgpioddriver in windows, we will still throw platformNotSupported.

src/System.Device.Gpio/Directory.Build.targets
pgrawehr: It will be hard to test things on OSX, since I don't have a Mac. But I believe if we really do all the testing for a specific hardware at runtime, this should work on OSX the same. And things like Arduino or FT4222 should work just fine. It would just throw an Exception if trying to instantiate i.e. the Windows10Driver, but that's very much the same as if you try this on linux.

src/System.Device.Gpio/System/Device/Gpio/GpioController.Linux.cs
pgrawehr: True. I'll clean that up.

src/System.Device.Gpio/System/Device/Gpio/Drivers/HummingBoardDriver.Linux.cs
pgrawehr: I agree. That part I wasn't so sure about in detail. There are several possibilities here:
1. Add check in .ctor (not very nice, since ctors throwing for reasons other than invalid arguments is not good style)
2. Provide a public `Initialize` method for this check (cleaner when it comes to error handling, but may lead to usage errors)
3. Deprecate the usage of the existing .ctors of these classes and put the test logic to the corresponding `Board` implementation. Then it needs to be done (and implemented) only once. 

I'd prefer to do 3, since the problem not only happens for all the GpioDrivers but also for all the I2C drivers, the PWM drivers, the SPI drivers, etc. I'd then do 1 just for backwards compatibility (and provide a new ctor when created from the Board).

src/System.Device.Gpio/Directory.Build.targets
pgrawehr: There's a difference between "the library is not supported on Platform X" vs "there's no hardware available on Platform X for which we have a driver".

pgrawehr:
Fixes #1037 as well

src/System.Device.Gpio/System/Device/Gpio/Drivers/HummingBoardDriver.Linux.cs
pgrawehr: 1) already works, at least if you're not so very peculiar about the exception type being thrown. (DllNotFoundException, NotSupportedException, PlatformNotSupportedException, throw the dice). 
I'll start with 3) but on a separate branch, otherwise there's probably to much stuff on the same PR in the end.

src/System.Device.Gpio/System.Device.Gpio.csproj
pgrawehr: Yes, removed this one.

joperezr:
There is a bunch of stuff on Directory.Build.Targets that can also get removed. Same goes for eng\CrossTargetingRIDS.targets. If you are ok with it, I could also commit into this PR and remove a bunch of infra stuff that wouldn't be required after this. Since I was the one that set most of this up, then it'll probably be easier for me to find all these things no longer needed.

src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -1,16 +1,12 @@
...
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <LangVersion>preview</LangVersion>
-    <!-- Only produce reference assembly in one of our inner builds -->
-    <ProduceReferenceAssembly Condition="'$(TargetsLinux)' == 'true'">true</ProduceReferenceAssembly>
     <TargetRefPath>$(OutputPath)$(TargetFramework)\ref\$(MSBuildProjectName).dll</TargetRefPath>
     <IsPackable>true</IsPackable>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <GeneratePlatformNotSupportedAssembly Condition="'$(RuntimeIdentifier)' == ''">true</GeneratePlatformNotSupportedAssembly>
-    <Configurations>Windows-Debug;Linux-Debug;Linux-Release;Windows-Release;Debug;Release</Configurations>
+    <Configurations>Debug;Release</Configurations>
```
joperezr, line 9: I believe these are already the defaults in an SDK project so we can remove them as well.

src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -1,16 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <RuntimeIdentifiers>win;linux</RuntimeIdentifiers>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <LangVersion>preview</LangVersion>
-    <!-- Only produce reference assembly in one of our inner builds -->
-    <ProduceReferenceAssembly Condition="'$(TargetsLinux)' == 'true'">true</ProduceReferenceAssembly>
```
joperezr, line 8: I agree that we won't need a ref assembly any longer, but please also remove the TargetRefPath property after this line since that won't be used either any longer

src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -25,17 +21,8 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
   </ItemGroup>
-  <ItemGroup Condition="'$(TargetsLinux)' == 'true'">
-    <!--Excluding Windows implementations-->
-    <Compile Remove="**\*.Windows.cs" />
-    <Compile Remove="Interop\Windows\**\*.cs" />
```
joperezr, line 31: This does mean that now the Linux implementation will depend on some winmd files, obviously not at runtime but just wanted to note that. I don't expect this to be a problem, but there are some runtimes that will fail to load an assembly when one of the dependencies is not present, that said I think that only applies to AOT so we are most likely safe here.

src/System.Device.Gpio/System/Device/Gpio/Drivers/HummingBoardDriver.cs
```diff
@@ -7,7 +7,7 @@ namespace System.Device.Gpio.Drivers
     /// <summary>
     /// A GPIO driver for the HummingBoard.
     /// </summary>
-    public partial class HummingBoardDriver // Different base classes declared in HummingboardDriver.Linux.cs and HummingboardDriver.Windows.cs
+    public class HummingBoardDriver : Windows10Driver
```
joperezr, line 10: This change I don't fully like here, it is definitely less than ideal to have HummingBoardDriver which could be running in Linux to have a base class called Windows10Driver. We should either think on how to either make the HummingBoardDriver extend GpioDriver in Windows as well, or to rename Windows10Driver so that we can make something less platform-named. Both changes would be breaking.

joperezr:
>     public class LibGpiodDriver : UnixDriver

All These drivers that are Linux specific (LibGpiod, RaspberryPi3, UnixDriver, etc) should now have a constructor which first thing it does is to check OS, and if platform is not Unix then throw PNSE. We defenitely don't want to fail late for something we know won't work.

---
Refers to: src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs:12 in 1dff47a. [](commit_id = 1dff47a2a3e1c99bf4d7ffd5a03cd6a23f8bba74, deletion_comment = False)

joperezr:
Left some comments, mostly what I want to see is if you could add logic on most GPIO driver constructors to make sure we throw Platform not supported when running on a OS that the driver won't support. Other than that most things look great, as I said if you are ok I'm happy to contribute to your PR by removing all logic  related to packaging and cross rid targeting cleanup, and also to verify that the surface area is still the same. Last thing to discuss here, would be the differences between base clases in Windows vs Linux, since Windows adds a base class to some of their Drivers so that is something we will have to design on how to fix as change will most likely be breaking.

pgrawehr:
> 
> 
> There is a bunch of stuff on Directory.Build.Targets that can also get removed. Same goes for eng\CrossTargetingRIDS.targets. If you are ok with it, I could also commit into this PR and remove a bunch of infra stuff that wouldn't be required after this. Since I was the one that set most of this up, then it'll probably be easier for me to find all these things no longer needed.

Sure, go ahead. I was hoping to get some support on this from you, just because you created this setup.

src/System.Device.Gpio/System.Device.Gpio.csproj
pgrawehr: What is AOT? 

Apparently, it runs on all existing platforms (at least all builds are green). For Windows and Linux I tested that the exceptions only pop up when the corresponding (in this case: the wrong) classes are instantiated. But that's the behavior we expect.

src/System.Device.Gpio/System/Device/Gpio/Drivers/HummingBoardDriver.cs
pgrawehr: The reason for changing it that way was that the HummingBoardDriver.Linux.cs implementation file previously consisted of only "NotImplementedException" methods, so it newer worked. Therefore I thought it was anyway only supported on Windows. 

If this was ever supported on linux, we could just provide a HummingBoardLinuxDriver (and move the magic of choosing the right one to the HummingBoard class or at least the GetBestDriverForBoard() method).

pgrawehr:
> 
> 
>  Last thing to discuss here, would be the differences between base clases in Windows vs Linux, since Windows adds a base class to some of their Drivers so that is something we will have to design on how to fix as change will most likely be breaking.

Please verify this, but I think it's not, because so far we did not have a driver that actually worked on both platforms. I.e. there was a RaspberryPi3Driver that inherited from WindowsDriver on Windows and UnixDriver on Unix, but this class didn't work on windows (the implementation consisted of exceptions only). So while technically, it might be a breaking change, nobody used these classes on the wrong platform.

src/System.Device.Gpio/System/Device/Gpio/Drivers/HummingBoardDriver.Linux.cs
pgrawehr: Should be working as expected now

src/System.Device.Gpio/System.Device.Gpio.csproj
joperezr: AOT is ahead-of-time so basically it is when IL is compiled down to native from compile time. There are several frameworks that work like that, like UWP or Xamarin. Again, .NET Core is not like that so I think we are safe so we probably don't need to do anything special here.

src/System.Device.Gpio/System/Device/Gpio/Drivers/HummingBoardDriver.cs
joperezr: having it never worked in the past is different as have it never work at all in the future. Imagine we eventually decide to add a HummingBoard driver in the future for Linux, and if so and we keep this as is, it means we would have to have an implementation of Windows10Driver on Linux as well so that we don't introduce a breaking change. This is obviosly less than ideal. In hindsight, we probably should have never exposed drivers with platforms name on them, it would be better to have a "GenericDriver" or "PlatformDriver" and have different implementations for it (what today is Wiindows10Driver and UnixDriver) but I guess that boat has sailed and will brings us some trouble now that we are unifying things.

joperezr:
> Sure, go ahead. I was hoping to get some support on this from you, just because you created this setup.

Sounds good, I'll push some changes today to remove stuff as well as address some nits so that we can get your PR ready to go

joperezr:
> Please verify this, but I think it's not, because so far we did not have a driver that actually worked on both platforms. I.e. there was a RaspberryPi3Driver that inherited from WindowsDriver on Windows and UnixDriver on Unix, but this class didn't work on windows (the implementation consisted of exceptions only). So while technically, it might be a breaking change, nobody used these classes on the wrong platform.

Actually I believe that RaspberryPi3Driver did work on Windows, because it was just basically inheriting from WindowsDriver, so it would really just forward the calls to WindowsDriver. HummingBoardDriver should be doing the same, I'll address this in a commit, push it to your branch and see what you think. With my change I'll try to make sure we don't break scenarios that worked in 1.0 and also to not introduce a breaking change, and finally to ensure that we won't be adding base classes to some drivers that are platform specific in case we someday want to add implementation to them.

src/System.Device.Gpio/System/Device/Gpio/Drivers/HummingBoardDriver.cs
pgrawehr: We could make the HummingBoardDriver aggregate its underlying driver (UnixDriver or Windows10Driver) instead of inheriting it. That would make the actual implementation flexible, and, despite it technically being a breaking change, probably nobody would really notice.

joperezr:
@pgrawehr @krwq  please take a look at ce4568a to see how I'm dealing with making sure we don't introduce breaking changes but at the same time allowing some drivers to use the implementation from other drivers. @tarekgh I'm particularly interested in your take on the changes that went in to the SetRegister and ClearRegister on the RaspberryDriver as I want to make sure that this extra layer of indirection won't hit perf on the LCD Display

krwq:
@joperezr I don't think this will regress it but if it does I think we should fix it separately. Overall this change over-benefits temporarily breaking a single device bindings. We should definitely test this after this is merged

joperezr:
> Overall this change over-benefits temporarily breaking a single device bindings

Even though the LCD one will be the most affected (if at all) all bindings are potentially touched/affected by this, as this will add indirection to pretty much all calls in GPIO. I immediately thought about the LCD one because I know that is the most perf-sensitive one but I still plan on running some perf measurements of before and after the change with just reading/writing registers in general.

krwq:
I think SoftwareSpi and SoftwarePwm are at the highest risk of breaking something although in case of say SysFs it probably will not matter

tarekgh:
@krwq could you try this changed on the RGB Matrix and look if you see any difference? we'll not be able to tell without trying. But I agree we shouldn't block this PR for that.

krwq:
@tarekgh I'm currently in advanced stage of moving out so will be hard as all my stuff is almost completely packed. LED matrices are still not packed but most of my Raspberry PIs are packed and the remaining ones are configured for the router I already packed. I'll see if I can test it soon but there is a high chance I won't make it in time. If not then I'll fix the binding if it's broken after we're done moving (I'm expecting at least 4 months though) or someone can pick this up if they hit any issue sooner. I'll see if I have space to perhaps leave 1 segment somewhere for playing around but that might not cover threshold scenarios we want to try.

tarekgh:
@krwq no worries at all. Nothing really urgent.

krwq:
@tarekgh I'll make sure that someone has a spare matrix somewhere 😄

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPiRegisterDriver.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            get { throw new PlatformNotSupportedException(); }
+            set { throw new PlatformNotSupportedException(); }
+        }
+
+        protected internal virtual ulong SetRegister
+        {
+            get { throw new PlatformNotSupportedException(); }
+            set { throw new PlatformNotSupportedException(); }
+        }
```
pgrawehr, line 33: Do we need these to be declared here? They're only used in the derived class.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPiRegisterDriver.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace System.Device.Gpio
+{
+    internal class RaspberryPiRegisterDriver : GpioDriver
+    {
+        private GpioDriver _internalDriver;
+
+        public RaspberryPiRegisterDriver()
+        {
+            _internalDriver = new Windows10Driver();
```
pgrawehr: Since RaspberryPi3LinuxDriver derives from this (is that really intended?), this will throw on Linux.

pgrawehr: Never mind, this is already fixed in the next commit

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPiRegisterDriver.cs
joperezr: They are also used in Windows when using the raspberrypi3driver. We could remove them if we decided to add an ifcheck on RaspberryPi3Driver everytime these methods get called but we don’t want to add another perf hit on this critical scenario. One thing I didn’t explore was to just have these two be delegate functions that get set on the constructor of raspberrypi3driver, that way we only have to do the check once. I’ll try exploring that option as well.

pgrawehr: Actually, I wonder what these are intended for. Our code nowhere uses these properties and they would only allow a derived class to access a very specific register.

joperezr: These two properties are basically a shortcut we had to create in order to support the RGBMatrix device binding. We only use them there: https://github.com/dotnet/iot/blob/master/src/devices/RGBLedMatrix/Gpio.cs

Basically this device requires you to be able to write to GPIO at a super high rate, and by adding these properties we basically allow you to write a) more than one pin at a time and b) skip validation overhead we do when you try to write to a pin using the controller. You can only call these two if you manually passed in the driver when creating the GpioController, and you keep a handle to the driver since you call them via the driver. They are also not super easy to use, since they are modifying the registers inside the Pi directly in order to set the value of the pins.

pgrawehr: I see. Maybe providing the pointer to /dev/gpiomem would be more flexible and even easier to use (for bindings that know what they're doing). And redirecting would not have a performance impact.

joperezr: That is more or less what this is doing, except that instead of giving back the whole gpiomem, it only gives back the specific registers for changing the gpio values.

joperezr:
@pgrawehr from my side I think things are ready to go, is there anything you want to add to this or are there any questions/feedback with the changes I pushed? If not, we can go ahead and merge.

pgrawehr:
It's fine from my end. Build passes, the samples I've tested worked, too. 

Next step will probably be to have a look trough the samples and clean up those projects. 
What about now creating solutions for all the bindings (together with their samples and the tests, if any) in the same go?

joperezr:
>Next step will probably be to have a look trough the samples and clean up those projects.

Oh you are right, I forgot about those, I'll include one more commit to fix that. As for the solutions for all bindings, let's do that in a separate PR since it is really a different concept. It is also easy to do, it is just a matter of creating a script to go into each folder and take care of it for us.

joperezr:
I'll merge after CI passes

joperezr:
Looks like the FastEvent test is very flaky in Windows so I'll go ahead and disable that for now

krwq:
@joperezr is there an issue about this flaky test? if not can you create it and if there is one can you add a comment in the code to point to that issue?

joperezr:
done with https://github.com/dotnet/iot/issues/1054

PR #1039, opened 2020-04-06
Arduino/Firmata device support

pgrawehr:
Fixes #262 

This adds support for controlling an Arduino board directly from a PC. The arduino board will run a special firmware called "Firmata", that interprets the commands from the PC and set it to hardware. See Readme for details.

This also includeds an `ArduinoBoard` class, which is a prototype for #878. I'll later push more commits that will show how that would look like with the existing Windows/Linux drivers (but that's somehow also dependent on #1038, because things get complicated otherwise, see #1037 for one of the issues) 

@Ellerbach Can you check whether that interface would fit your FT4222 as well?

The first implementation for the firmata driver was from https://github.com/ms-iot/remote-wiring. MIT license.

Issue #974, opened 2020-02-06
PinValueChangedEventArgs uses wrong PinNumberingScheme


dontspeak:
Im initiating the gpiocontroller with **board** numbering scheme, but when I receive change events, the `PinValueChangedEventArgs  `does provides the **logical** pin:

```
GpioController controller = new GpioController(PinNumberingScheme.Board);
controller.OpenPin(16, PinMode.InputPullUp); // Set "16"
controller.RegisterCallbackForPinValueChangedEvent(16, PinEventTypes.Falling, MyEvent);

public void MyEvent(object sender, PinValueChangedEventArgs args)
{
Log(args.PinNumber); // Gives me "23"
}
```

The `PinNumber `should depend on the `PinNumberingScheme`? I.E. in that case, it should return the according board pin number (which is 16) and not the logical one (which is 23).

RPI 4b
.NET CORE 3.1.1

pgrawehr:
I guess you are right. From what I've seen so far, the board numbering scheme is only marginally supported.

joperezr:
Thanks for logging this @dontspeak, that was a great find! I believe you are right, we usually just handle the numbering scheme on the controller, but all the actual drivers use logical numbering so I don't doubt that this is the case today.

Ellerbach:
[Triage] Closing this one as addressed on the various PR

Issue #874, opened 2019-12-01
Running SysFsDriverTests or RaspberryPiDriverTests causes PWM to be unusable


krwq:
After running SysFsDriverTests or RaspberryPiDriverTests PWM no longer works.

I have 100% repro: If I run PWM tests in a loop or run LibGpiodDriverTests everything works without any issues.
If I run SysFS or RPI3 GPIO tests once after that PWM becomes completely unusable and stops reacting to any changes (regardless if using PWM manually through `/sys/class/pwm/` or our APIs) - PI has to be rebooted.

Found during testing https://github.com/dotnet/iot/pull/758

I'm currently on latest from master and my PI3 is running Raspbian after `sudo apt-get update` and `sudo apt-get upgrade` (was reproing before updates as well).

I'll try to further isolate the problem if I have time.

pgrawehr:
The reason for your problem is #857. You need to reset the pin to PWM mode after some other test has set it to in or out.

krwq:
@pgrawehr possible but I don't believe any of these drivers should be even touching PWM pin...

krwq:
actually seems they do touch it...

krwq:
Hmm.. I'm kinda leaning toward not allowing to open any pins which have mode set to something which we do not understand to avoid similar problems in the future... @pgrawehr - that should at least prevent the problems like this since they are really weird to debug

pgrawehr:
@krwq I wouldn't completely disallow it, but make sure they are not opened in different modes at the same time. Otherwise, you would still not be able to open a pin in PWM mode if another application has set it to Output. 
In either case, I think that comes down to the "Board" approach you mentioned in https://github.com/dotnet/iot/issues/809#issuecomment-558862627, since only then we have a common board-wide controller that can know which pins are open for what and which pins have which special function(s).

krwq:
Yes, other thing I'm considering is perhaps let user decide what to do on dispose, i.e.:

`gpioController.OpenPin(18, PinMode.Output, XYZ.RestorePreviousMode);` // XYZ.DoNotClose, XYZ.DontCare

pgrawehr:
I didn't encounter this scenario, but I think you're right. Not sure whether the XYZ parameter should be given on `Open()` or `Close()` though.

krwq:
@pgrawehr IMO Open because:
- Dispose auto-calls Close which would make it easier to use without having to close pins manually
[...]

PR #1044, opened 2020-04-15
[WIP] Board abstraction

### LINKED ISSUE #878: Hardware abstraction - design conversation
Moving side conversation from #857
https://github.com/dotnet/iot/issues/857#issuecomment-557268367
https://github.com/dotnet/iot/issues/857#issuecomment-561810527
Related: #874

This issue is created to talk about abstraction of hardware in some reasonable way. See above issue (at minimum comments above) to get some background on the problems and possible solution.

This is a very early and initial problem description so it's not complete yet.

cc: @pgrawehr 

Initially we have though of isolating GPIO/PWM/I2C/SPI as separate entities and we have shipped that design but more we are exploring the hardware the more we find that GPIO is not exactly separate. I expect this separation to create more and more issues which will be hard to debug or fix in the future if we don't start to think about them right now.

Here is some broad thought dump into what I think should be covered here
- GPIO can be used to emulate I2C/SPI/PWM (see SoftwareSpi, SoftwarePwm)
- sometimes emulation is supported or abstracted away by OS driver and using both at the same time is causing weird issues (#874 and more generic issue about the same problem: #857)
- in some cases one device can emulate another one (i.e. MPU9250 has embedded I2C bus) but in some cases our abstraction cannot handle that very well (I.e. as part of #523 @Ellerbach and self have tried to come up how to create implementation of I2cDevice so that it can be used by embedded device without code changes but have found tiny differences which were big enough that it didn't feel feasible)
- some devices work with multiple protocols (i.e. I2C/SPI/Serial port) and underlying protocol is often the same but we are lacking abstraction to be able to handle them as one - perhaps some register based abstraction?
- some devices/protocols can be completely abstract (#875, #262, #78)
- some devices are a collections of devices (see: SenseHat)

pgrawehr:
Fixes #878 and #974 

This is a suggestion/preview how everything should come together for #878. 

This PR includes #1039 and #1038, because these three (big) changes somewhat depend on each other. Please leave specific comments on Arduino and Platform-Independent build on those PRs. The relevant changes for this PR are all in the System.Device.Gpio assembly and the base classes defined there:
- New `Board` class abstracting a piece of hardware with different interfaces (for now: GPIO, SPI, I2C, PWM, Analog In).
- Implementations for Raspberry, generic Unix, Windows and Arduino/Firmata prepared (more TBD, also things like MCP23xxx or ADS1115 possible, and of course FT4222)
- The `Board` should now be used for creating all interface elements (GpioController, I2cDevice, etc.) This ensures consistent access and allows verification of pin usage - both for error checking (attempting to use a pin for I2C and GPIO at the same time) and setting the proper pin mode, if required (i.e. on the Raspberry).
- This `Board` class is provided to all interface elements as back-reference, to be able to manage Pins or other common behavior like logical-to-physical pin mappings. 
- Therefore added new Ctor to the abstract I2cDevice, SpiDevice, PwmChannel classes
- Marked existing Ctors/Static Create methods obsolete, but should still work.

Some of the advantages:
- Only one place is needed to check whether a board class is compatible with the actual hardware.
- No need to repeat device scans all over the place (especially since all types of drivers are now theoretically accessible regardless of the library target OS).
- Some boards (i.e. Arduino) require one common master component anyway.
- Chaining components (Reading from a DHT11 connection to an MCP23017 to an arduino to a Raspi) is much cleaner to do for a client. (Requires #1016 as well). 
- and more...

### DIFF
--- a/src/System.Device.Gpio/System/Device/Board.cs
+++ b/src/System.Device.Gpio/System/Device/Board.cs
@@ -1,0 +1,246 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Device.Analog;
+using System.Device.Boards;
+using System.Device.Gpio;
+using System.Device.I2c;
+using System.Device.Pwm;
+using System.Device.Spi;
+using System.IO;
+using System.Text;
+
+namespace System.Device
+{
+    public abstract class Board : MarshalByRefObject, IDisposable
+    {
+        private readonly PinNumberingScheme _defaultNumberingScheme;
+        private readonly object _pinReservationsLock;
+        private readonly Dictionary<int, PinReservation> _pinReservations;
+        private bool _initialized;
+        private bool _disposed;
+
+        protected Board(PinNumberingScheme defaultNumberingScheme)
+        {
+            _defaultNumberingScheme = defaultNumberingScheme;
+            _pinReservations = new Dictionary<int, PinReservation>();
+            _pinReservationsLock = new object();
+            _initialized = false;
+            _disposed = false;
+        }
+
+        ~Board()
+        {
+            Dispose(false);
+        }
+
+        public event Action<string, Exception> LogMessages;
+
+        protected bool Initialized
+        {
+            get
+            {
+                return _initialized;
+            }
+        }
+
+        protected bool Disposed
+        {
+            get
+            {
+                return _disposed;
+            }
+        }
+
+        public PinNumberingScheme DefaultPinNumberingScheme
+        {
+            get
+            {
+                return _defaultNumberingScheme;
+            }
+        }
+
+        protected void Log(string message, Exception exception = null)
+        {
+            LogMessages?.Invoke(message, null);
+        }
+
+        /// <summary>
+        /// Converts pin numbers in the active <see cref="PinNumberingScheme"/> to logical pin numbers.
+        /// Does nothing if <see cref="PinNumberingScheme"/> is logical
+        /// </summary>
+        /// <param name="pinNumber">Pin numbers</param>
+        /// <returns>The logical pin number</returns>
+        public abstract int ConvertPinNumberToLogicalNumberingScheme(int pinNumber);
+
+        /// <summary>
+        /// Converts logical pin numbers to the active numbering scheme.
+        /// This is the opposite of <see cref="ConvertPinNumberToLogicalNumberingScheme"/>.
+        /// </summary>
+        /// <param name="pinNumber">Logical pin number</param>
+        /// <returns>The pin number in the given pin numbering scheme</returns>
+        public abstract int ConvertLogicalNumberingSchemeToPinNumber(int pinNumber);
+
+        /// <summary>
+        /// Reserves a pin for a specific usage. This is done automatically if a known interface (i.e. GpioController) is
+        /// used to open the pin, but may be used to block a pin explicitly, i.e. for UART.
+        /// </summary>
+        /// <param name="pinNumber">The pin number, in the boards default numbering scheme</param>
+        /// <param name="usage">Intended usage of the pin</param>
+        /// <param name="owner">Class that owns the pin (use "this")</param>
+        public virtual void ReservePin(int pinNumber, PinUsage usage, object owner)
+        {
+            if (!_initialized)
+            {
+                Initialize();
+            }
+
+            int logicalPin = ConvertPinNumberToLogicalNumberingScheme(pinNumber);
+            lock (_pinReservationsLock)
+            {
+                if (_pinReservations.TryGetValue(logicalPin, out var reservation))
+                {
+                    throw new InvalidOperationException($"Pin {pinNumber} has already been reserved for {reservation.Usage} by class {reservation.Owner}.");
+                }
+
+                PinReservation rsv = new PinReservation(logicalPin, usage, owner);
+                _pinReservations.Add(logicalPin, rsv);
+            }
+
+            ActivatePinMode(logicalPin, usage);
+        }
+
+        public virtual void ReleasePin(int pinNumber, PinUsage usage, object owner)
+        {
+            if (!_initialized)
+            {
+                throw new InvalidOperationException("Cannot release a pin if board is not initialized.");
+            }
+
+            int logicalPin = ConvertPinNumberToLogicalNumberingScheme(pinNumber);
+            lock (_pinReservationsLock)
+            {
+                if (_pinReservations.TryGetValue(logicalPin, out var reservation))
+                {
+                    if (reservation.Owner != owner || reservation.Usage != usage)
+                    {
+                        throw new InvalidOperationException($"Cannot release Pin {pinNumber}, because you are not the owner or the usage is wrong. Class {reservation.Owner} has reserved the Pin for {reservation.Usage}");
+                    }
+
+                    _pinReservations.Remove(logicalPin);
+                }
+                else
+                {
+                    throw new InvalidOperationException($"Cannot release Pin {pinNumber}, because it is not reserved.");
+                }
+            }
+        }
+
+        /// <summary>
+        /// Override this method if something special needs to be done to use the pin for the given device.
+        /// Many devices support multiple functions per Pin, but not at the same time, so that some kind of
+        /// multiplexer needs to be set accordingly.
+        /// </summary>
+        /// <param name="pinNumber">The logical pin number to use.</param>
+        /// <param name="usage">The intended usage</param>
+        protected virtual void ActivatePinMode(int pinNumber, PinUsage usage)
+        {
+        }
+
+        public virtual void Initialize()
+        {
+            if (_disposed)
+            {
+                throw new ObjectDisposedException(ToString());
+            }
+
+            _initialized = true;
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            _disposed = true;
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        private sealed class PinReservation
+        {
+            public PinReservation(int pin, PinUsage usage, object owner)
+            {
+                Pin = pin;
+                Usage = usage;
+                Owner = owner;
+            }
+
+            public int Pin { get; }
+            public PinUsage Usage { get; }
+
+            /// <summary>
+            /// Component that owns the pin (used mainly for debugging)
+            /// </summary>
+            public object Owner { get; }
+        }
+
+        public abstract GpioController CreateGpioController(PinNumberingScheme pinNumberingScheme);
+        public abstract I2cDevice CreateI2cDevice(I2cConnectionSettings connectionSettings);
+
+        public abstract SpiDevice CreateSpiDevice(SpiConnectionSettings settings);
+
+        public abstract PwmChannel CreatePwmChannel(
+            int chip,
+            int channel,
+            int frequency = 400,
+            double dutyCyclePercentage = 0.5);
+
+        public abstract AnalogController CreateAnalogController(int chip);
+
+        public static Board DetermineOptimalBoardForHardware(PinNumberingScheme defaultNumberingScheme = PinNumberingScheme.Logical)
+        {
+            if (Environment.OSVersion.Platform == PlatformID.Unix)
+            {
+                Board board = null;
+                try
+                {
+                    board = new RaspberryPiBoard(defaultNumberingScheme, true);
+                    board.Initialize();
+                }
+                catch (Exception x) when ((x is NotSupportedException) || (x is IOException))
+                {
+                    board?.Dispose();
+                    board = null;
+                }
+
+                if (board != null)
+                {
+                    return board;
+                }
+
+                try
+                {
+                    board = new UnixBoard(defaultNumberingScheme, true);
+                    board.Initialize();
+                }
+                catch (Exception x) when ((x is NotSupportedException) || (x is IOException))
+                {
+                    board?.Dispose();
+                    board = null;
+                }
+
+                if (board != null)
+                {
+                    return board;
+                }
+            }
+            else
+            {
+                // TODO: Create WindowsBoard()
+            }
+
+            throw new PlatformNotSupportedException("Could not find a matching board driver for this hardware");
+        }
+    }
+}
--- a/src/System.Device.Gpio/System/Device/Boards/WindowsBoard.cs
+++ b/src/System.Device.Gpio/System/Device/Boards/WindowsBoard.cs
@@ -1,0 +1,59 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Device.Analog;
+using System.Device.Gpio;
+using System.Device.Gpio.Drivers;
+using System.Device.I2c;
+using System.Device.Pwm;
+using System.Device.Spi;
+using System.Text;
+
+namespace System.Device.Boards
+{
+    public class WindowsBoard : Board
+    {
+        public WindowsBoard(PinNumberingScheme defaultNumberingScheme)
+            : base(defaultNumberingScheme)
+        {
+            if (defaultNumberingScheme != PinNumberingScheme.Board)
+            {
+                throw new NotSupportedException("This board only supports logical pin numbering");
+            }
+        }
+
+        public override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber)
+        {
+            return pinNumber;
+        }
+
+        public override int ConvertLogicalNumberingSchemeToPinNumber(int pinNumber)
+        {
+            return pinNumber;
+        }
+
+        public override GpioController CreateGpioController(PinNumberingScheme pinNumberingScheme)
+        {
+            return new GpioController(DefaultPinNumberingScheme, new Windows10Driver(this), this);
+        }
+
+        public override I2cDevice CreateI2cDevice(I2cConnectionSettings connectionSettings)
+        {
+            return new Windows10I2cDevice(connectionSettings, this);
+        }
+
+        public override SpiDevice CreateSpiDevice(SpiConnectionSettings settings)
+        {
+            return new Windows10SpiDevice(settings, this);
+        }
+
+        public override PwmChannel CreatePwmChannel(int chip, int channel, int frequency = 400, double dutyCyclePercentage = 0.5)
+        {
+            throw new NotImplementedException();
+        }
+
+        public override AnalogController CreateAnalogController(int chip)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
@@ -42,7 +42,8 @@ namespace System.Device.Gpio.Drivers
             GPIOD_LINE_REQUEST_FLAG_BIAS_PULL_UP = (1UL << 5)
         }
 
-        public LibGpiodDriver(int gpioChip = 0)
+        public LibGpiodDriver(Board board, int gpioChip = 0)
+        : base(board)
         {
             if (Environment.OSVersion.Platform != PlatformID.Unix)
             {
@@ -66,6 +67,12 @@ namespace System.Device.Gpio.Drivers
             {
                 throw ExceptionHelper.GetPlatformNotSupportedException(ExceptionResource.LibGpiodNotInstalled);
             }
+        }
+
+        [Obsolete("Use Board.CreateGpioController instead")]
+        public LibGpiodDriver(int gpioChip = 0)
+            : this(null, gpioChip)
+        {
         }
 
         protected internal override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEventTypes eventTypes, PinChangeEventHandler callback)
--- a/src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
@@ -9,17 +9,64 @@ namespace System.Device.Gpio
 {
     public abstract class GpioDriver : IDisposable
     {
+        [Obsolete("Available for backwards compatibility only")]
+        protected GpioDriver()
+        {
+            Board = null;
+        }
+
+        protected GpioDriver(Board board)
+        {
+            Board = board;
+        }
+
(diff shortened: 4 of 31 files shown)


shaggygi:
I briefly looked over this and is very interesting.  I question if `Board` should really be named `Device` instead.  Some devices can offer features (GpioController, SPI, etc.) at the component level, as well.  Take the [MCP2210](#1059) as @pgrawehr mentioned for example.  In addition, we usually think of hardware, like RPi, BeagleBone, etc. as "boards" since we are usually seeing/touching them as their circuit board is exposed (for hobby/cost reasonings).  However, they are "devices".  If you bought a plastic case and put an RPi in it... it usually exposes only the pins... just like integrated components (e.g. DIP16, SOIC8, etc.) = bindings (e.g. Mcp2308).

Scenario:  Let's say I designed some hardware with a combination of 3 RPi, a few Adafruit modules and other random supporting components.  The final module was packaged up in an enclosure and provided some specific features like blink some lights, update an LCD, activate a horn, etc..  But it also exposed some connectors that offer I2C bus and PWM pins.  Most likely there would be a "device"-binding API created for this module as a whole.  To me, this scenario would be a "device" and not a "board".  My point is this module, with a combination of components is no different than a combination of components that make up a RPi.  Both are "devices".

Lastly, this core API is called System."Device".Gpio 😄.  But as I write this, my brain starts to hurt as you have to start thinking about other possible future System.Device APIs (e.g. System.Device.Bluetooth, System.Device.Hid, etc.).  Wouldn't they need to have a respective "Device" (e.g. BleDevice, etc.)?  Therefore, this System.Device.Gpio `Device` might need to be `GpioDevice`.  But then... would you want to create a `RaspberryPi` that implements `GpioDevice` as it could implement some of those other APIs.  This would not be able to implement multiple abstract classes. 🤕 🍺 🤕 

Just my 2 pennies to think about.

pgrawehr:
@shaggygi Thanks for the feedback. "Board" was the name proposed [here](https://github.com/dotnet/iot/issues/878#issuecomment-563363677), and I think it is a bit less ambiguous than the very generic term "device" and is not used anywhere right now (we already have PwmDevice, SpiDevice, etc), even though it may also not be adequate all the time. 

In your scenario, you'd have your "3RpiDevice", consisting of 3 "RpiBoards", with each one of these boards having Gpio Pins, and I2C etc. You could also connect something like an MCP2210 to one of the Rpis, then you have another board, which provides more GPIO pins. So it would be something like:
```
var uart = firstRpiBoard.CreateUart("/dev/ttyUSB1");
var mcp2210 = new Mcp2210(uart); // Assuming the mcp communicates using uart, but could also use I2C or whatever
mcpGpioPins = mcp2210.CreateGpioController();
mcpGpioPins.Open(1).Write(PinValue.High);
```

The "board"'s implementation is pretty small, it is mostly a factory instance for other objects, like GpioControllers, I2CControllers etc. So any binding would not implement both I2cDevice and GpioDevice but instead the Board and offer CreateXyz-Methods (which could return internal instances that wrap one's own functionality if needed, of course). This is especially useful for hardware like the MCP (or the FT4222, or the Arduino/Firmata) where there's one common instance that actually talks to the hardware but several logical controllers. And on the Raspi it would be used to verify that only one controller is using a pin at a time.

shaggygi:
Good points and maybe I'm missing something.  You can have the following:

```csharp
var uart = firstRpiBoard.CreateUart("/dev/ttyUSB1");
var mcp2210 = new Mcp2210(uart); // Assuming the mcp communicates using uart, but could also use I2C or whatever
mcpGpioPins = mcp2210.CreateGpioController();
// Along with:
var rpiPins = firstRpiBoard.CreateGpioController();
```
I'm thinking we are on the same page for above.  I just can't get past implementing `mcp2210` as a "board".

pgrawehr:
> I'm thinking we are on the same page for above. I just can't get past implementing `mcp2210` as a "board".

I don't know how that piece of hardware actually looks like, but maybe you should just put the drawing of a PCB on top of it ;-) I'm open for better names, but "Device" is just to broad and ambiguous IMHO. And "Controller" is also used all over the place already. Maybe something like "PinManager"?

shaggygi:
Yup, I believe Device is probably ambiguous, just not sure a better name.  The MCP2201 is nothing more than a IC that has USB, GPIO and SPI pins.  There are MCP2210 breakout boards (just like Adafruit provides for many ICs).  I can see the MCP2210 device binding API implementing this proposal to include methods like `CreateGpioController` just as I could see the MCP23xxx implementing in the future.  But still don't think either would be consider a "board" (understanding that is just a name that is being discussed).

The IoT Plug and Play group is using the word "capability".  Not sure if that is any better.
Take a look and see if anything might be better: https://github.com/Azure/IoTPlugandPlay/blob/master/samples/SampleDevice.capabilitymodel.json

pgrawehr:
Capability isn't the right word we're searching, either, I believe. That would be (like what is described over there) a set of properties describing what a system/piece of hardware can do, not the implementation thereof.

pgrawehr:
Idea: Maybe we name the base class something like "PinManager", but leave it open for the concrete classes to use matching names. I.e. "ArduinoBoard" would probably be suitable, but for MCP2201 we could still go for "Mcp2201Chip" or so.

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
```diff
@@ -7,8 +7,15 @@ namespace System.Device.Pwm
     /// <summary>
     /// Represents a single PWM channel.
     /// </summary>
-    public abstract partial class PwmChannel : IDisposable
+    public abstract class PwmChannel : IDisposable
     {
+        public Board Board { get; }
+
+        protected PwmChannel(Board board)
```
joperezr, line 14: This would be removing the parameterless constructor on PwmChannel which would be a breaking change. If we go with these changes we would have to add that constructor back.

src/System.Device.Gpio/System/Device/Spi/SpiDevice.cs
```diff
@@ -7,13 +7,23 @@ namespace System.Device.Spi
     /// <summary>
     /// The communications channel to a device on a SPI bus.
     /// </summary>
-    public abstract partial class SpiDevice : IDisposable
+    public abstract class SpiDevice : IDisposable
     {
+        public SpiDevice(SpiConnectionSettings settings, Board board)
```
joperezr, line 12: Same here, this is removing the paramterless constructor which we can't take as it is a breaking change.

src/System.Device.Gpio/System/Device/I2c/I2cDevice.cs
```diff
@@ -9,11 +9,42 @@ namespace System.Device.I2c
...
+            get;
+            private set;
+        }
+
         /// <summary>
         /// The connection settings of a device on an I2C bus. The connection settings are immutable after the device is created
         /// so the object returned will be a clone of the settings object.
         /// </summary>
-        public abstract I2cConnectionSettings ConnectionSettings { get; }
+        public I2cConnectionSettings ConnectionSettings
```
joperezr, line 41: Removing abstract from a member is a breaking change.

src/System.Device.Gpio/System/Device/Spi/SpiDevice.cs
```diff
@@ -56,6 +66,7 @@ public abstract partial class SpiDevice : IDisposable
         /// </summary>
         /// <param name="settings">The connection settings of a device on a SPI bus.</param>
         /// <returns>A communications channel to a device on a SPI bus running on Windows 10 IoT.</returns>
+        [Obsolete]
```
joperezr, line 69: I honestly don't agree with this change. I think that most people using this will still probably not care about the Board abstraction, and what this is doing is probably forcing consumers to use it. Given this Create method is the main entrypoint, I would keep it without Obsolete and simply just add the other option of doing it via a Board.

src/System.Device.Gpio/System/Device/Spi/SpiDevice.cs
```diff
@@ -7,13 +7,23 @@ namespace System.Device.Spi
     /// <summary>
     /// The communications channel to a device on a SPI bus.
     /// </summary>
-    public abstract partial class SpiDevice : IDisposable
+    public abstract class SpiDevice : IDisposable
     {
+        public SpiDevice(SpiConnectionSettings settings, Board board)
+        {
+            ConnectionSettings = settings;
```
joperezr, line 14: One thing that has worked well for Spi and I2c is the factory pattern for creating instances, I would encourage to keep using it specially since we have differnet platform implementations underneath.

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
```diff
@@ -48,6 +55,7 @@ protected virtual void Dispose(bool disposing)
         /// <param name="frequency">The frequency in hertz.</param>
         /// <param name="dutyCyclePercentage">The duty cycle percentage represented as a value between 0.0 and 1.0.</param>
         /// <returns>A PWM channel running on Windows 10 IoT.</returns>
+        [Obsolete("Use Board.CreatePwmChannel instead")]
```
joperezr, line 58: Same comment than with SpiDevice. I think with an introduction of a feature like this, we shouldn't force people down that path every time and instead keep allowing both ways for creating objects so people that don't want to go through the Board abstraction are not force d to.

src/System.Device.Gpio/System/Device/I2c/I2cConnectionSettings.cs
```diff
@@ -39,5 +62,15 @@ internal I2cConnectionSettings(I2cConnectionSettings other)
         /// The bus address of the I2C device.
         /// </summary>
         public int DeviceAddress { get; }
+
+        public int SdaPin
```
joperezr, line 66: I also don't agree too much with this. One problem with this type that we learned in hindsight was that it was leaking implementation details for a very specific implementation (hardware I2c). Adding these would now also leak software implementation. We should talk about  this more on how to better approach the connection settings.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
```diff
@@ -68,6 +69,12 @@ public LibGpiodDriver(int gpioChip = 0)
             }
         }
 
+        [Obsolete("Use Board.CreateGpioController instead")]
```
joperezr, line 72: Comment in general for all the Gpio Driver Obsoletes: I don't think that we should go through this path into forcing people to use Board abstraction. One of the features we have in our design is to support people to create new GpioDrivers that internally use 2 or more drivers for cases where you are using pins from different providers. Doing that via the Board now would add an extra layer of complexity that wouldn't be desired.

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
```diff
@@ -9,17 +9,64 @@ namespace System.Device.Gpio
...
+
         /// <summary>
         /// The number of pins provided by the driver.
         /// </summary>
         protected internal abstract int PinCount { get; }
 
+        /// <summary>
+        /// Explicitly initializes the driver (instead of waiting for the first pin to be opened)
+        /// </summary>
+        public virtual void Initialize()
```
joperezr, line 36: Should this be abstract if it is empty?

src/System.Device.Gpio/System/Device/Boards/WindowsBoard.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Device.Gpio;
+using System.Device.Gpio.Drivers;
+using System.Device.I2c;
+using System.Device.Pwm;
+using System.Device.Spi;
+using System.Text;
+
+namespace System.Device.Boards
+{
+    public class WindowsBoard : Board
```
joperezr, line 13: One learning we had from V1 was that it was a wrong design to have the platform on the name of the type as this just increased surface area and complexity. I imagine instead following the factory pattern we have in SpiDevice, where we only have one public class Board, and then have a Create factory method which will in turn decide which one of these implementations to use. Unfortunately we can't fix this aspect on the original API due to compat reasons, but I would vote against making this mistake again.

src/System.Device.Gpio/System/Device/Board.cs
```diff
@@ -0,0 +1,246 @@
...
+using System.Device.Analog;
+using System.Device.Boards;
+using System.Device.Gpio;
+using System.Device.I2c;
+using System.Device.Pwm;
+using System.Device.Spi;
+using System.IO;
+using System.Text;
+
+namespace System.Device
```
joperezr, line 12: This should probably be on the Boards namespace

src/System.Device.Gpio/System/Device/Board.cs
```diff
@@ -0,0 +1,246 @@
...
+            _defaultNumberingScheme = defaultNumberingScheme;
+            _pinReservations = new Dictionary<int, PinReservation>();
+            _pinReservationsLock = new object();
+            _initialized = false;
+            _disposed = false;
+        }
+
+        ~Board()
+        {
+            Dispose(false);
```
joperezr, line 33: Why do we need a finalizer?

src/System.Device.Gpio/System/Device/Board.cs
```diff
@@ -0,0 +1,246 @@
...
+
+        public abstract PwmChannel CreatePwmChannel(
+            int chip,
+            int channel,
+            int frequency = 400,
+            double dutyCyclePercentage = 0.5);
+
+        public abstract AnalogController CreateAnalogController(int chip);
+
+        public static Board DetermineOptimalBoardForHardware(PinNumberingScheme defaultNumberingScheme = PinNumberingScheme.Logical)
```
joperezr, line 201: This should only be used for implementation detail, shouldn't be exposed

joperezr:
Sorry it took so long to review the PR. I left some comments specifically on the API, I haven't looked at the implementation yet since I think we should first finalize on the shape before we look into the implementation. I do want to add some overall comments to the feature:
- To be honest, I haven't yet seen much value that the Board abstraction is adding other than pin management, and I think that the fact of now having two ways of creating controllers (or devices) is just going to cause more confusion. I also took a peek on the issue that this is fixing and couldn't find enough examples of why this is needed, so it would be good if either you or @krwq could elaborate more on what are the real advantages and scenarios that this would fix.
- If we did decide to go with it, in my mind this should be less of a disruptive change where you only get the Board abstraction if you want it, but it shouldn't pollute all of the surface area in the rest of S.D.G and it should certainly not obsolete the previous ways of doing things.
- I also think we should probably separate the Analog part from this PR, as that one is a completely new feature that we want to work on regardless, so it would be good to for now just focus on adding the Board onto what we already have.
- Regarding Device vs Board: I don't have a good suggestion here. The only thing to keep in mind is that once we are in agreement in general for what we want to go with here, we will need to take the proposal into API Review (as changes to System.Device.Gpio have to be reviewed by the review board) so they will probably have additional comments here.

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
pgrawehr: Basically yes, but then it would be breaking.

src/System.Device.Gpio/System/Device/Board.cs
pgrawehr: I added it for completeness, and I think it makes sense here: All of the derived (and aggregated) classes implement IDisposable and potentially allocate unmanaged resources such as device driver handles. These should be cleaned up even if the user forgets to call Dispose.

src/System.Device.Gpio/System/Device/Boards/WindowsBoard.cs
pgrawehr: So you would suggest to make this internal instead? Or change the name?

src/System.Device.Gpio/System/Device/Board.cs
pgrawehr: This is the factory method you suggested to use, so making it internal is not really helpful, I suppose.

pgrawehr:
> 
> 
> Sorry it took so long to review the PR. I left some comments specifically on the API, I haven't looked at the implementation yet since I think we should first finalize on the shape before we look into the implementation. I do want to add some overall comments to the feature:
> 
>     * To be honest, I haven't yet seen much value that the Board abstraction is adding other than pin management, and I think that the fact of now having two ways of creating controllers (or devices) is just going to cause more confusion. I also took a peek on the issue that this is fixing and couldn't find enough examples of why this is needed, so it would be good if either you or @krwq could elaborate more on what are the real advantages and scenarios that this would fix.

There are several reasons for this approach:
- Pin management: There's currently no way of reserving a pin for a certain function or to make sure it is only used for one function at a time. While that part may be a convenience function, the problem denoted in #874 is not: There's currently no way of setting a pin to a mode other than input or output. Many boards (i.e. Raspi, but also Arduino/Firmata) require a special method call to set a pin to i2c, pwm or spi mode (if it isn't set to that by the bootup configuration). 

- Boards requiring a common master interface: Some devices (let's keep calling them "boards" for now) require one common master interface to communicate with them, and the Gpio Controller is only a sub-feature-set of that master interface. These boards are instantiated with a common connection instance (i.e. a serial port number) and all further communication technically goes trough that master interface. Examples for this includes boards such as Arduino, FT4222, MCP2210, but also devices such as I2C port expanders, which are devices that have one primary instance but provide pin controllers. The standard devices (i.e. a Raspi) can benefit from this as well, for instance if it is controlled over a network. 

- More obvious control of what happens where: `myi2cPortExpander.CreateGpioController()` or `myArduinoBoardOnCom1.CreateAnalogController()` helps the user to understand what device is providing the pins he's going to access. 

> 
>     * If we did decide to go with it, in my mind this should be less of a disruptive change where you only get the Board abstraction if you want it, but it shouldn't pollute all of the surface area in the rest of S.D.G and it should certainly not obsolete the previous ways of doing things.

Meanwhile, I mostly agree to that. I think this could simply be done by keeping a global static instance of the main board class (the one the default factory methods would be using). 

> 
>     * I also think we should probably separate the Analog part from this PR, as that one is a completely new feature that we want to work on regardless, so it would be good to for now just focus on adding the Board onto what we already have.

I agree, that is basically a separate topic. But it shows that there could be more such extensions coming up (i.e. proper support for one-wire communication pins, UART pins, etc.). I will create a separate PR to discuss that interface. 

> 
>     * Regarding Device vs Board: I don't have a good suggestion here. The only thing to keep in mind is that once we are in agreement in general for what we want to go with here, we will need to take the proposal into API Review (as changes to System.Device.Gpio have to be reviewed by the review board) so they will probably have additional comments here.

As long as it's mostly about naming, I don't mind much either.

src/System.Device.Gpio/System/Device/Boards/WindowsBoard.cs
joperezr: Have it internal. So you basically end up having only one public (abstract) Board class without a constructor and with a factory Create method. This Create method, will give you a WindowsBoard when running in Windows and a LinuxBoard when running in Linux. That way Windows* and Linux* is not present anywhere on the public facing API, but we still have the separation logic on our code so stuff still makes sense.

src/System.Device.Gpio/System/Device/Board.cs
joperezr: In that case the derived and aggregated classes are the ones that should have the finalizer (if applicable), not this one. In general in .NET we actually try to not use finalizers at all when possible, and instead use the Dispose pattern and always using SafeHandles whenever we have a type that holds native resources.

src/System.Device.Gpio/System/Device/Board.cs
joperezr: I see, I didn't realize it was returning a Board. In that case the naming is confusing, factory methods are usually called something like Create()

src/System.Device.Gpio/System/Device/Board.cs
joperezr: you may find this page helpful with guidance about Dispose patterns and finalizers: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

joperezr:
While I agree that having better pin management would be beneficial in some scenarios, I find that usually the people that consumes our APIs are also the ones that wire the PCB and sensors/devices, so most of the time they have had to already kind of figure out the pin layout and management themselves which is why at least initially we didn't provide a Board "warpper" class and instead had developers pick and choose the controllers they wanted to use. Also, we expect that most of the final applications that will be using S.D.G won't be doing it directly, but instead would use one (or many) of the Iot.Device.Bindings classes which are in turn using S.D.Gpio. Given that most of those bindings are developed in this repo, we so far haven't found many cases where a Board abstraction was necessary or would have helped a lot. That said, this by no means say that we shouldn't prototype this, but I really want to avoid disrupting the whole S.D.Gpio API surface because of a prototype. 

My suggestion for driving this forward is the following: Let's try to come up with a design for that Board class that helps you accomplish pin management and the other issues that it was intended to address, but considering that we want to place that Board class on Iot.Device.Bindings for now. So that means that we want to try not to touch System.Device.Gpio at all for now, and only prototype more or less how this would look like by adding a Board class. The benefits of that is that we don't have to have the right design from the beggining and we can iterate as Board class would be experimental, and we also don't have to go to API Review process because we wouldn't be touching S.D.Gpio. Does that sound like a good plan to you?

pgrawehr:
That suggestion sounds reasonable, and it would allow the use of the board class for the devices that would clearly benefit from it (and generally from a common interface). FT4222, Arduino, MCP, etc. But it would make it hard (or impossible) to use it on the Raspi, because the GpioController(s) need to have callbacks into the board for the pin management to work properly. That at least requires some kind of (probably small) interface.

mi-hol:
>  But it would make it hard (or impossible) to use it on the Raspi, because the GpioController(s) need to have callbacks into the board for the pin management to work properly. That at least requires some kind of (probably small) interface.

Is this great change still trying to overcome the encountered hurdle?

pgrawehr:
Closing this for now - superseeded by #1128 (@joperezr which could take a review, but I know its a big one)

joperezr:
> (@joperezr which could take a review, but I know its a big one)

Thanks a lot for your patience @pgrawehr we are finally wrapping up dotnet 5 so now we have time to start burning the debt that built up in iot repo, and your board abstraction is definitely something we want to make sure we get reviewed and merged.

PR #1046, opened 2020-04-19
Adding formatting to Pressure similar to Temperature

MarkCiliaVincenti:
Pressure now looks more like the Temperature struct.

### DIFF
--- a/src/devices/Units/Pressure.cs
+++ b/src/devices/Units/Pressure.cs
@@ -189,17 +189,17 @@ namespace Iot.Units
             {
                 return string.Format(culture, $"{{0:F{numDigits}}} kPa", Kilopascal);
             }
-            
+
             if (formatArgs.StartsWith("HPA"))
             {
                 return string.Format(culture, $"{{0:F{numDigits}}} hPa", Hectopascal);
             }
-            
+
             if (formatArgs.StartsWith("INHG"))
             {
                 return string.Format(culture, $"{{0:F{numDigits}}} inHg", InchOfMercury);
             }
-            
+
             if (formatArgs.StartsWith("MMHG"))
             {
                 return string.Format(culture, $"{{0:F{numDigits}}} mmHg", MillimeterOfMercury);
--- a/src/devices/Units/Pressure.cs
+++ b/src/devices/Units/Pressure.cs
@@ -10,7 +10,7 @@ namespace Iot.Units
     /// <summary>
     /// Structure representing pressure
     /// </summary>
-    public struct Pressure : IEquatable<Pressure>, IComparable<Pressure>, IComparable
+    public struct Pressure : IFormattable, IComparable<Pressure>, IEquatable<Pressure>
     {
         private const double MillibarRatio = 0.01;
         private const double KilopascalRatio = 0.001;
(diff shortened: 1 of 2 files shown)


MarkCiliaVincenti:
I'm not sure why it's failing.

pgrawehr:
The AddCallbackFallingEdgeNotDetectedTest is known to be not 100% reliable, so just try again (by doing a dummy commit). If it keeps failing, something is fishy...

src/devices/Units/Pressure.cs
```diff
@@ -120,51 +120,110 @@ public static Pressure FromMillimeterOfMercury(double value)
             return new Pressure(value / MillimeterOfMercuryRatio);
         }
 
+        /// <inheritdoc cref="IComparable{T}.CompareTo" />
```
krwq, line 123: is cref required? usually just `<inheritdoc/>` is enough

src/devices/Units/Pressure.cs
```diff
@@ -120,51 +120,110 @@ public static Pressure FromMillimeterOfMercury(double value)
...
+        /// var s = p.ToString("PA2"); -> "101325.01 Pa"
+        /// var s = p.ToString("HPA2"); --> "1013.25 hPa"
+        /// </code>
+        /// </example>
         /// </summary>
-        public override bool Equals(object obj)
+        /// <param name="formatArgs">Format string</param>
+        /// <param name="culture">Culture to use. Affects the format of the number.</param>
+        /// <returns>String representation</returns>
+        public string ToString(string formatArgs, IFormatProvider culture)
```
krwq, line 169: I'm assuming those are made up formats and not officially documented anywhere, if so please create an issue about them (if there are similar for temperature also mention them) since we need to make sure everyone agrees on that before we ship that - it will be super breaking to change that in the future so let's make sure we get those right.

I think it's fine not to block PR on that (just please make sure you check case insensitive (pass comparer)) and that you create an issue

krwq: note: there is a chance we will remove those formatters if we don't agree on them in the issue by the time we start shipping preview.

src/devices/Units/tests/PressureTests.cs
```diff
@@ -47,7 +47,7 @@ public void ToStringDefault()
             using (new SetCultureForTest("de-CH"))
             {
                 Pressure p = Pressure.FromHectopascal(1024.24);
-                Assert.Equal("1024.24hPa", p.ToString());
+                Assert.Equal("1024.2 hPa", p.ToString());
```
krwq, line 50: please add at least 3 tests:
- ToString with just units no number of decimal places
- ToString with single digit number of decimal places
- ToString with double digit number of decimal places (if that makes sense)

src/devices/Units/Pressure.cs
krwq, line 169: I don't feel strongly about it but I think until we fully design how this formatter should work we should remove it. I'm suspecting this currently won't cover all cases. Some people will want to put space between number and unit, some will prefer to display it slightly differently, some will have translated unit name or use different alphabel (not sure about it, just guessing). I'd rather throw on ToString or return just .Pascal.ToString

pgrawehr: I've created #1047 on this, since it was originally my idea.

krwq:
@MarkCiliaVincenti I'm closing this PR since we're using UnitsNet now. Probably makes sense to re-create PR if there is anything not covered by UnitsNet left here. Thanks for this contribution as it was one of the PRs/issues which convinced me to moving to UnitsNet.

PR #1049, opened 2020-04-21
ValidatePinCount should use the overridable member

### LINKED ISSUE #1048: GPIO Pin extension
We are using a Raspberry PI 3 with a extended GPIO up to GPIO 44. unfortunately I would like to use the RaspberryPi3Driver but I cannot override the Pin Mapping Method or the checks. because they are private/internal or protected.

`
        protected internal override int PinCount => 28;
        private void ValidatePinNumber(int pinNumber)
        protected internal override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber)
`

It should be possible to override these. Maybe I am missing out something here.

pgrawehr:
Fixes #1048

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
@@ -16,7 +16,7 @@ namespace System.Device.Gpio.Drivers
 
         private void ValidatePinNumber(int pinNumber)
         {
-            if (pinNumber < 0 || pinNumber > 27)
+            if (pinNumber < 0 || pinNumber >= PinCount)
             {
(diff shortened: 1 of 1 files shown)


joperezr:
LGTM, thanks @pgrawehr

JTrotta:
@pgrawehr  Are you sure >= is correct?
Usualy GPIO numbering is zero based. I suggest:
` if (pinNumber < 0 || pinNumber > PinCount)`

krwq:
@JTrotta if you got 3 pins then 0, 1, 2 would be valid so `>=` looks good to me

JTrotta:
@krwq form your example PinCount should be 3. Pin 3 does not exists bu it would be validated, I suppose.

krwq:
@JTrotta isn't that check for throwing an exception though?

PR #1051, opened 2020-04-25
Replace Units with UnitsNet

krwq:
**Note: This PR is against feature-units branch**

This is a prototype/experiment work which is part of https://github.com/dotnet/iot/issues/869

- Removes Units
- Adds package reference to UnitsNet
- Adjust code to make it build

I'm still voting for yes on this library

I've run our Units tests before removing them against this library. There are some differences with formatting and treating NaN and Infinities.

cc: @joperezr @pgrawehr @MarkCiliaVincenti @JTrotta @Ellerbach 

<details>

<summary>Test run output</summary>

```
A total of 1 test files matched the specified pattern.
[xUnit.net 00:00:00.82]     Units.Tests.PressureTests.ConversionFactorTests [FAIL]
[xUnit.net 00:00:00.82]     Iot.Units.Tests.TemperatureTests.FromCelsiusPositiveInfinity [FAIL]
[xUnit.net 00:00:00.82]     Iot.Units.Tests.TemperatureTests.FromKelvinNaN [FAIL]
  X Units.Tests.PressureTests.ConversionFactorTests [6ms]
  Error Message:
   Assert.Equal() Failure
Expected: 102423.94009 (rounded from 102423.940094101)
Actual:   102424 (rounded from 102424)
  Stack Trace:
     at Units.Tests.PressureTests.ConversionFactorTests() in C:\src\iot\src\devices\Units\tests\PressureTests.cs:line 21
  X Iot.Units.Tests.TemperatureTests.FromCelsiusPositiveInfinity [1ms]
  Error Message:
   System.ArgumentException : PositiveInfinity or NegativeInfinity is not a valid number.
Parameter name: val
  Stack Trace:
     at UnitsNet.InternalHelpers.Guard.EnsureValidNumber(Double value, String paramName) in C:\projects\unitsnet\UnitsNet\InternalHelpers\Guard.cs:line 25
   at UnitsNet.QuantityValue..ctor(Double val) in C:\projects\unitsnet\UnitsNet\QuantityValue.cs:line 39
   at UnitsNet.QuantityValue.op_Implicit(Double val) in C:\projects\unitsnet\UnitsNet\QuantityValue.cs:line 66
   at Iot.Units.Tests.TemperatureTests.FromCelsiusPositiveInfinity() in C:\src\iot\src\devices\Units\tests\TemperatureTests.cs:line 78
  X Iot.Units.Tests.TemperatureTests.FromKelvinNaN [1ms]
  Error Message:
[...]

PR #1052, opened 2020-04-25
[prototype work] Replace Units with UnitsNet

krwq:
This is WIP and just used so that we can see if CI passes, we still didn't decide if we want this work.

This consists of:
- https://github.com/dotnet/iot/pull/1051
  - above PR is initial implementation (removing our Units and using UnitsNet)
  - see PR description for our test runs after porting to UnitsNet - this should give us basic overview on differences

Currently known differences:
- names are only slightly different (i.e. `Celsius` => `DegreesCelsius`, `Kelvin` => `Kelvins`, `Hectopascal` => `Hectopascals` - it's mostly just plural or adding word `Degrees`)
- UnitsNet does not support taking NaN as an input
- default Formatting is slightly different (doesn't matter)
- `MeanSeaLevel` or any constants are not in UnitsNet as far as I've seen - I've moved it to WeatherHelpers


joperezr:
Closing this as CI has been added to the other branch.

PR #1056, opened 2020-04-28
Treat warnings as errors

pgrawehr:
- Ensure warnings cannot be ignored (such as missing documentation in the core assembly...)
- Add documentation where missing

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -15,8 +15,6 @@
     <LangVersion>Latest</LangVersion>
     <!-- Generate snupkg package -->
     <IncludeSymbols>true</IncludeSymbols>
-    <NoWarn>1701;1702</NoWarn>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -15,6 +15,8 @@
     <LangVersion>Latest</LangVersion>
     <!-- Generate snupkg package -->
     <IncludeSymbols>true</IncludeSymbols>
+    <NoWarn>1701;1702</NoWarn>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
   <ItemGroup>
--- a/eng/Compilers.props
+++ b/eng/Compilers.props
@@ -6,6 +6,16 @@
     <CompilerVersion>3.6.0-2.20166.2</CompilerVersion>
   </PropertyGroup>
 
+  <PropertyGroup Condition="!($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn></NoWarn> <!-- Make sure the list is empty - by default it is not -->
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
+  <PropertyGroup Condition="($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <NoWarn>CS1591</NoWarn> <!-- Documentation on public members missing -->
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
   <ItemGroup>
     <PackageReference Condition="'$(UseSdkCompilers)' != 'true'" Include="Microsoft.Net.Compilers.Toolset" Version="$(CompilerVersion)" IsImplicitlyDefined="true" PrivateAssets="all" />
   </ItemGroup>
--- a/eng/Compilers.props
+++ b/eng/Compilers.props
@@ -8,12 +8,13 @@
 
   <PropertyGroup Condition="!($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <NoWarn></NoWarn> <!-- Make sure the list is empty - by default it is not -->
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <NoWarn>$(NoWarn.Replace("1591", ""))</NoWarn> <!-- Make sure CS1591 is not in the list (by default it is) -->
   </PropertyGroup>
   <PropertyGroup Condition="($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
-    <NoWarn>CS1591</NoWarn> <!-- Documentation on public members missing -->
+    <NoWarn>$(NoWarn)</NoWarn> <!-- No changes required here (For test projects, missing documentation is ok) -->
+  </PropertyGroup>
+  <PropertyGroup>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <ItemGroup>
--- a/eng/Compilers.props
+++ b/eng/Compilers.props
@@ -6,15 +6,10 @@
     <CompilerVersion>3.6.0-2.20166.2</CompilerVersion>
   </PropertyGroup>
 
-  <PropertyGroup Condition="!($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <NoWarn>$(NoWarn.Replace("1591", ""))</NoWarn> <!-- Make sure CS1591 is not in the list (by default it is) -->
-  </PropertyGroup>
-  <PropertyGroup Condition="($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
-    <GenerateDocumentationFile>false</GenerateDocumentationFile>
-    <NoWarn>$(NoWarn)</NoWarn> <!-- No changes required here (For test projects, missing documentation is ok) -->
-  </PropertyGroup>
   <PropertyGroup>
+    <_IsTestProject Condition="$(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase))">true</_IsTestProject>
+    <NoWarn Condition="'$(_IsTestProject)' != 'true'">$(NoWarn.Replace("1591", ""))</NoWarn> <!-- Removing CS1591 from warning exclusion list as we want to make sure documentation is enforced -->
+    <GenerateDocumentationFile Condition="'$(_IsTestProject)' != 'true'">true</GenerateDocumentationFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <ItemGroup>
--- a/samples/Directory.Build.props
+++ b/samples/Directory.Build.props
@@ -3,6 +3,10 @@
   <Import Project="../eng/Compilers.props" />
   <Import Project="../eng/Versions.props" />
   <PropertyGroup>
+    <!--
+         CS8321: The local function is not declared
+         CS1591: Missing XML comment for publicly visible type or member
+    -->
     <NoWarn>$(NoWarn);CS8321;CS1591</NoWarn>
     <DeterministicSourcePaths>false</DeterministicSourcePaths>
     <IsPackable>false</IsPackable>
--- a/samples/Directory.Build.props
+++ b/samples/Directory.Build.props
@@ -3,7 +3,7 @@
   <Import Project="../eng/Compilers.props" />
   <Import Project="../eng/Versions.props" />
   <PropertyGroup>
-    <NoWarn>$(NoWarn);CS8321</NoWarn>
+    <NoWarn>$(NoWarn);CS8321;CS1591</NoWarn>
     <DeterministicSourcePaths>false</DeterministicSourcePaths>
     <IsPackable>false</IsPackable>
     <MainLibraryPath>$(MSBuildThisFileDirectory)../src/System.Device.Gpio/</MainLibraryPath>
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
@@ -22,6 +22,9 @@ namespace System.Device.Gpio.Drivers
         private readonly int _pinCount;
         private SafeChipHandle _chip;
 
+        /// <summary>
+        /// The number of pins this driver controls
+        /// </summary>
         protected internal override int PinCount => _pinCount;
 
         // for use the bias flags we need libgpiod version 1.5 or later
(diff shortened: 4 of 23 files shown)


Directory.Build.props
```diff
@@ -15,6 +15,8 @@
     <LangVersion>Latest</LangVersion>
     <!-- Generate snupkg package -->
     <IncludeSymbols>true</IncludeSymbols>
+    <NoWarn>1701;1702</NoWarn>
```
joperezr, line 18: Can you add a comment to explain what these two are for?

Directory.Build.props
```diff
@@ -15,6 +15,8 @@
...
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
joperezr, line 19: Note that we are also setting this to true in src\devices\Directory.Build.props which I know doesn't import this or doesn't get imported by S.D.G or by I.D.B but we could probably find a central place where it will apply for both, for example in Compilers.props

src/System.Device.Gpio/System/Device/Gpio/PinValue.cs
```diff
@@ -26,13 +26,42 @@ namespace System.Device.Gpio
         /// </summary>
         public static PinValue Low => new PinValue(0);
 
+        /// <summary>
+        /// Implicit conversion from int. 0 means low, everything else means high.
```
joperezr, line 30: nit: instead of saying "low" or "high" should we use `<see cref="PinValue.Low"` and similar for High?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
```diff
@@ -9,6 +9,10 @@
 
 namespace System.Device.Gpio.Drivers
 {
+    /// <summary>
+    /// This driver uses the Libgpiod library to get user-level access to the gpio ports.
+    /// It superseeds the SysFsDriver
```
joperezr, line 14: Should we mention that it requires target machine to install libgpiod library?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
```diff
@@ -18,6 +22,9 @@ public class LibGpiodDriver : UnixDriver
         private readonly int _pinCount;
         private SafeChipHandle _chip;
 
+        /// <summary>
+        /// The number of pins this driver controls
+        /// </summary>
```
joperezr, line 27: NIT: inherit doc as well here?

src/System.Device.Gpio/System/Device/Gpio/Drivers/UnixDriver.cs
```diff
@@ -4,8 +4,15 @@
...
 namespace System.Device.Gpio.Drivers
 {
+    /// <summary>
+    /// The base class for the standard unix drivers
+    /// </summary>
     public abstract class UnixDriver : GpioDriver
     {
+        /// <summary>
+        /// Construct an instance of an unix driver.
+        /// Just checks that the platform is Unix.
```
joperezr, line 14: NIT: I would probably ommit the fact that all we do here is check platform as we may change this in the future and will likely miss an update to the summary

src/System.Device.Gpio/System/Device/Gpio/Drivers/UnixDriver.cs
```diff
@@ -14,6 +21,10 @@ protected UnixDriver()
             }
         }
 
+        /// <summary>
+        /// Static factory method
+        /// </summary>
+        /// <returns>An instance of GpioDriver, depending on which one fits</returns>
```
joperezr, line 27: Now that I think of it, given that we have a factory pattern we missed adding a platform check to it to throw when called from Windows. You don't have to fix it in this PR if you don't want to but just wanted to mention it

src/System.Device.Gpio/System/Device/I2c/I2cDevice.cs
```diff
@@ -75,12 +75,17 @@ public static I2cDevice Create(I2cConnectionSettings settings)
...
         public void Dispose()
         {
             Dispose(true);
             GC.SuppressFinalize(this);
         }
 
+        /// <summary>
+        /// Disposes this instance
+        /// </summary>
+        /// <param name="disposing">True if explicitly disposing, false if in finalizer</param>
```
joperezr, line 88: NIT: use `<see langword="true"` instead of true and false

src/System.Device.Gpio/System/Device/Spi/SpiDevice.cs
```diff
@@ -68,12 +68,17 @@ public static SpiDevice Create(SpiConnectionSettings settings)
...
         public void Dispose()
         {
             Dispose(true);
             GC.SuppressFinalize(this);
         }
 
+        /// <summary>
+        /// Disposes this instance
+        /// </summary>
+        /// <param name="disposing">True if explicitly disposing, false if in finalizer</param>
```
joperezr, line 81: Same comment for true and false

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
```diff
@@ -29,12 +29,17 @@ public abstract partial class PwmChannel : IDisposable
...
         public void Dispose()
         {
             Dispose(true);
             GC.SuppressFinalize(this);
         }
 
+        /// <summary>
+        /// Disposes this instance
+        /// </summary>
+        /// <param name="disposing">True if explicitly disposing, false if in finalizer</param>
```
joperezr, line 42: Same comment for using langword for true and false

joperezr:
Left a few comments but most are just small NITs, this looks good otherwise. Thanks for adding it @pgrawehr !

Directory.Build.props
pgrawehr: I'll have to search again - but it was something very obscure that affected old compiler versions only or smth.

src/System.Device.Gpio/System/Device/Gpio/Drivers/UnixDriver.cs
pgrawehr: If you try this on Windows, both LibGpiodDriver and SysFsDriver will fail, so you already end up with an exception.

Directory.Build.props
joperezr: I will honestly prefer not to add a NoWarn (especially globally) if we don't fully understand why are we getting it. If we do end up needing it, I would much rather have a pragma disable in the code that is emiting the warning so that we only scope the no warn to the case we understand on why is it happening.

src/System.Device.Gpio/System/Device/Gpio/Drivers/UnixDriver.cs
joperezr: you are correct, I guess it is just really a matter of making the code more readable as this way it would be straight obvious just by looking at the UnixDriver.cs that this driver is not supported on WIndows (although I guess you could infer it by looking at the file name 😆 )

Directory.Build.props
pgrawehr: Moved to Compilers.props and added a condition to disable CS1591 (documentation) for test assemblies.

eng/Compilers.props
```diff
@@ -6,7 +6,17 @@
     <CompilerVersion>3.6.0-2.20166.2</CompilerVersion>
   </PropertyGroup>
 
+  <PropertyGroup Condition="!($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn></NoWarn> <!-- Make sure the list is empty - by default it is not -->
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
joperezr, line 12: NIT: this is always true so let's just define it in an unconditioned propertyGroup instead.

eng/Compilers.props
```diff
@@ -6,7 +6,17 @@
     <CompilerVersion>3.6.0-2.20166.2</CompilerVersion>
   </PropertyGroup>
 
+  <PropertyGroup Condition="!($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn></NoWarn> <!-- Make sure the list is empty - by default it is not -->
```
joperezr, line 11: I would just remove this as I don't think it is a good idea to hard reset it to empty here. There are some warnings that we may get coming from Arcade(the central infrastructure used in dotnet repos) which may be expected so we don't want to shoot ourselves in the foot here and remove those.

eng/Compilers.props
```diff
@@ -6,7 +6,17 @@
...
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
+  <PropertyGroup Condition="($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <NoWarn>CS1591</NoWarn> <!-- Documentation on public members missing -->
```
joperezr, line 16: Same comment here, instead of hard setting it to just this warning, let's instead do `<NoWarn>$(NoWarn);CS1591</NoWarn>` so as to not remove the warnings comming from our central infrastructure.

samples/Directory.Build.props
```diff
@@ -3,7 +3,7 @@
   <Import Project="../eng/Compilers.props" />
   <Import Project="../eng/Versions.props" />
   <PropertyGroup>
-    <NoWarn>$(NoWarn);CS8321</NoWarn>
+    <NoWarn>$(NoWarn);CS8321;CS1591</NoWarn>
```
joperezr, line 6: This shouldn't be required as this also imports Compilers.props right? Perhaps in Compilers.props we have to just adjust the condition of nowarn for this warning to exclude it for Tests and samples

tools/DevicesApiTester/DeviceApiTester.csproj
```diff
@@ -9,6 +9,11 @@
     <Company>The .NET Foundation</Company>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!-- Ignore missing documentation tags (mostly already documented parameter members, anyway)-->
+    <NoWarn>1591</NoWarn>
```
joperezr, line 14: NIT: do not reset NoWarn as well in here, so just do `<NoWarn>$(NoWarn);1591</NoWarn>`

krwq:
Honestly:
- I'm fine and agree with TreatWarningAsErrors when doing full build
- I find this switch super annoying when writing code (incremental builds) - I frequently put some code for debugging and having to re-iterate because whitespace was off by 1 on code I was gonna remove anyway is super annoying to me.

tools/templates/DeviceBindingTemplate/dotnet_new_device-binding_csharp/_DeviceBinding.cs
```diff
@@ -9,12 +9,19 @@
 
 namespace Iot.Device._DeviceBinding
 {
+    /// <summary>
+    /// Add documentation here
```
joperezr, line 13: I'm fine with keeping this, but just wondering if it would actually be better to remove them so that people get compiler errors when they don't have documentation. If we add these they could potentially just leave this and we may not catch it in review. That said, I would also like our template to generate code that can compile, so perhaps keeping it is right :). I'll let you be the one to decide here.

eng/Compilers.props
pgrawehr: The problem is that the default list we inherit contains CS1591, which we explicitly _do not_ want to disable. So without this hard reset, the change basically does not work.

samples/Directory.Build.props
pgrawehr: Question is: Do we want to exclude it for all samples or just those in the samples subfolder?

tools/templates/DeviceBindingTemplate/dotnet_new_device-binding_csharp/_DeviceBinding.cs
pgrawehr: I was exactly thinking the same, but the compiler beat me to it: The build fails without these changes (this somehow also gets compiled)

pgrawehr:
> 
> 
> Honestly:
> 
>     * I'm fine and agree with TreatWarningAsErrors when doing full build
> 
>     * I find this switch super annoying when writing code (incremental builds) - I frequently put some code for debugging and having to re-iterate because whitespace was off by 1 on code I was gonna remove anyway is super annoying to me (same for documentation errors during debuggin).

I'm used to strict conventions, so this doesn't anoy me any more - but I can understand you, I was feeling the same when we first added StyleCop to my company's build system. Since VS immediately points out the errors, its no longer as bad as it used to be, as you see the problems pop up immediatelly (with red lines), so normally no re-iteration is required. Disabling the rules for "normal" development would likely remove this and you'd only get the errors when you thought you were all done - even more annoying IMHO. 

What I regularly do for testing or debugging purposes is to put "// <autogenerated />" or "#pragma warning disable ..." at the top of the file I'm working on, to disable all these rules temporarily.

krwq:
I'm ok with this change as is as long as I'm democratically outvoted 😄

eng/Compilers.props
pgrawehr: `<NoWarn>$(NoWarn.Replace("1591", ""))</NoWarn>` does the trick, it seems.

joperezr:
I totally see your point @krwq about strict rules sometimes being annoying, but I think that the closest we can make the actual full build to the build people get in their VS will reduce the number of iterations people have to do on their PRs, and will also help driving CI successful runs up, so I'm fine with the change as well.

eng/Compilers.props
joperezr: 😆 feels hacky but I guess it's ok if it works. The only last thing that makes me uncomfortable is that now we depend on the import order, as we are depending on Compilers.props being imported after the NoWarn gets set, but I guess this is fine for now as whenever we break this it will be easy to diagnose as it will break the build.

eng/Compilers.props
```diff
@@ -6,7 +6,18 @@
...
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn>$(NoWarn.Replace("1591", ""))</NoWarn> <!-- Make sure CS1591 is not in the list (by default it is) -->
+  </PropertyGroup>
+  <PropertyGroup Condition="($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <NoWarn>$(NoWarn)</NoWarn> <!-- No changes required here (For test projects, missing documentation is ok) -->
+  </PropertyGroup>
+  <PropertyGroup>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
```
joperezr, line 19: How about changing this to one single property group to make it more readable like:

```suggestion
  <PropertyGroup>
    <_IsTestProject Condition="$(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase))">true</_IsTestProject>
    <NoWarn Condition="'$(_IsTestProject)' != 'true'">$(NoWarn.Replace("1591", ""))</NoWarn> <!-- Removing CS1591 from warning exclusion list as we want to make sure documentation is enforced -->
    <GenerateDocumentationFile Condition="'$(_IsTestProject)' != 'true'">true</GenerateDocumentationFile>
    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
  </PropertyGroup>
```

samples/Directory.Build.props
joperezr, line 6: It would be nice if in here we add the meaning of the warnings for future reference.

```suggestion
    <!--
         CS8321: The local function is not declared
         CS1591: Missing XML comment for publicly visible type or member
    -->
    <NoWarn>$(NoWarn);CS8321;CS1591</NoWarn>
```

eng/Compilers.props
```diff
@@ -8,12 +8,13 @@
 
   <PropertyGroup Condition="!($(MSBuildProjectName.EndsWith('Tests', StringComparison.OrdinalIgnoreCase)))">
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <NoWarn></NoWarn> <!-- Make sure the list is empty - by default it is not -->
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <NoWarn>$(NoWarn.Replace("1591", ""))</NoWarn> <!-- Make sure CS1591 is not in the list (by default it is) -->
```
krwq, line 11: can you write what 1591 is?

pgrawehr: Already fixed with joperezr's commit. It's all about this: "Missing documentation for publicly visible type"

eng/Compilers.props
pgrawehr: @joperezr Thanks for sorting this out - I'm still learning the quirks of MsBuild projects and the property handling.

PR #1060, opened 2020-04-30
Added sln file to doc that will be added with device-binding template

shaggygi:
Added sln file to example in device-binding README.


shaggygi:
@joperezr 👍 have another PR coming soon on the template code updates, as well.

PR #1065, opened 2020-05-03
Improved tests

StevenBoelens:
Test using temperature conversion failed on a system with a decimal comma. This has been fixed.
Also added a test that explicitly tests formatting using both a culture with decimal point as well as a culture with decimal comma that working with an explicit culture produces consistent test results.

Symptom before fixing (on a system using culture "nl-NL"):
Test Name:	Iot.Units.Tests.TemperatureTests.ToStringWithFormatArgsAndCulture
Test FullName:	Units.Tests.Iot.Units.Tests.TemperatureTests.Iot.Units.Tests.TemperatureTests.ToStringWithFormatArgsAndCulture
Test Source:	C:\Users\sboelens\source\repos\iot\src\devices\Units\tests\TemperatureTests.cs : line 165
Test Outcome:	Failed
Test Duration:	0:00:00

Test Name:	Iot.Units.Tests.TemperatureTests.ToStringWithFormatArgsAndCulture
Test Outcome:	Failed
Result StackTrace:	at Iot.Units.Tests.TemperatureTests.ToStringWithFormatArgsAndCulture() in C:\Users\sboelens\source\repos\iot\src\devices\Units\tests\TemperatureTests.cs:line 173
Result Message:	
Assert.Equal() Failure
                  ↓ (pos 8)
Expected: It is 68.4°F outside, this equals 20.21°C
Actual:   It is 68,4°F outside, this equals 20,21°C
                  ↑ (pos 8)

### DIFF
--- a/src/devices/Units/tests/TemperatureTests.cs
+++ b/src/devices/Units/tests/TemperatureTests.cs
@@ -169,8 +169,54 @@ namespace Iot.Units.Tests
             var s = t.ToString("K2", cf);
             Assert.Equal("293.36°K", s);
 
-            s = $"It is {t:F1} outside, this equals {t:C2}".ToString(cf);
+            s = ((FormattableString)$"It is {t:F1} outside, this equals {t:C2}").ToString(cf);
             Assert.Equal("It is 68.4°F outside, this equals 20.21°C", s);
+        }
+
+        /// <summary>
+        /// Test to see if using an explicit format provider works with decimal points and decimal commas.
+        /// In order to use an explicit format provider it is necessary to specify the string as a FormattableString instead of a normal string.
+        /// The test uses a global setting of en-US and then overrules the format to use both the en-US culture (with decimal point) and de-DE culture which uses decimal comma.
+        /// The test then checks to see if the results remain the same when using the de-DE culture as globally set culture.
+        /// </summary>
+        [Fact]
+        public void ToStringWithExplicitCultureAndFormatArgsAndCulture()
+        {
+            using (new SetCultureForTest("en-US"))
+            {
+                CultureInfo cf = new CultureInfo("en-US", false);
+                Temperature t = Temperature.FromCelsius(20.2123);
+                var s = t.ToString("K2", cf);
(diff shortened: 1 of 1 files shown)


pgrawehr:
So only the test was wrong, not the implementation?

StevenBoelens:
> So only the test was wrong, not the implementation?

Yes, only the test was wrong.

src/devices/Units/tests/TemperatureTests.cs
```diff
@@ -169,10 +169,56 @@ public void ToStringWithFormatArgsAndCulture()
...
             Assert.Equal("It is 68.4°F outside, this equals 20.21°C", s);
         }
 
+        /// <summary>
+        /// Test to see if using an explicit format provider works with decimal points and decimal commas.
+        /// In order to use an explicit format provider it is necessary to specify the string as a FormattableString instead of a normal string.
+        /// The test uses a global setting of en-US and then overrules the format to use both the en-US culture (with decimal point) and de-DE culture which uses decimal comma.
+        /// The test then checks to see if the results remain the same when using the de-DE culture as globally set culture.
+        /// </summary>
+        [Fact]
```
joperezr, line 182: NIT: looks like this could be converted to a Theory and just pass in culture and expected values in order to avoid most of the duplication here.

krwq:
@joperezr I feel like we should merge the feature-units branch which will remove all of these test cases and also fix the problem.

krwq:
Here is the changeset I'm talking about: https://github.com/dotnet/iot/pull/1072

StevenBoelens:
Switching to the UnitsNet library sounds like a good plan. This probably also fixes the error that temperature in Kelvin is not specified as 273°K but 273K without the ° symbol.

krwq:
@StevenBoelens we've merged the UnitsNet PR, can you try running tests now to see if your run is green and the issue with Kelvin is resolved?

StevenBoelens:
A rebuild of the entire library on my system (with decimal comma instead of decimal point) no longer fails. This seems resolved.

PR #1066, opened 2020-05-03
Fix Dot enum for Large4Digit7SegmentDisplay

gLes:
The values for LeftUpper and LeftLower dots were interchanged

I discovered this while testing my project that uses this library.

I did not notice this error during my original contribution because I did not use these dots separately until now in my project and even though they are all separately enumerated in the sample I just failed to notice that they lit up in the wrong order (manual testing ¯\_(ツ)_/¯).

### DIFF
--- a/src/devices/Display/Dot.cs
+++ b/src/devices/Display/Dot.cs
@@ -26,17 +26,17 @@ namespace Iot.Device.Display
         /// <summary>
         /// Left colon
         /// </summary>
-        LeftColon = LeftLower | LeftUpper,
-
-        /// <summary>
-        /// Left lower dot
-        /// </summary>
-        LeftLower = 0b0000_0100,
+        LeftColon = LeftUpper | LeftLower,
 
         /// <summary>
         /// Left upper dot
         /// </summary>
-        LeftUpper = 0b0000_1000,
+        LeftUpper = 0b0000_0100,
+
+        /// <summary>
+        /// Left lower dot
+        /// </summary>
+        LeftLower = 0b0000_1000,
 
         /// <summary>
         /// Decimal point (between third and fourth digits)


gLes:
I even created a workaround extension method for anyone interested in using this hardware (but that's probably only me 😃):

```csharp
using Iot.Device.Display;

namespace Extensions
{
    public static class DotExtensions
    {
        /// <summary>
        /// Swap bits for <see cref="Dot.LeftLower"/> and <see cref="Dot.LeftUpper"/>
        /// as a temporary workaround
        /// </summary>
        /// <param name="dots">dots value to modify</param>
        /// <returns>dots value with the <see cref="Dot.LeftLower"/> and <see cref="Dot.LeftUpper"/> swapped</returns>
        /// <remarks>TODO remove when https://github.com/dotnet/iot/pull/1066 is merged and published</remarks>
        public static Dot LeftDotsWorkaround(this Dot dots) =>
            dots.HasFlag(Dot.LeftLower) == dots.HasFlag(Dot.LeftUpper)
                ? dots // If they are both set or unset then return value unchanged
                : (Dot)((byte)dots ^ (byte)Dot.LeftColon); // Otherwise swap the two bits by XORing the value with LeftColon (LeftLower | LeftUpper)
    }
}
```

joperezr:
Thanks for logging the issue! Also thank you for providing a workaround as an extension method for users that hit this, would you like to take it a step further and submit a PR to fix the issue? Let me know if you are interested and I will gladly help you if needed on setting up and contributing.

gLes:
This conversation is already attached to my PR to fix this issue 🙂

https://github.com/dotnet/iot/pull/1066/files

gLes:
Hi @joperezr,

What's up with this issue? Can we put it up for review?

PR #1067, opened 2020-05-04
Added SHTC3 binding

famblard86:
Add Shtc3 device
base on Sht3x device with datasheet:

https://www.sensirion.com/fileadmin/user_upload/customers/sensirion/Dokumente/2_Humidity_Sensors/Datasheets/Sensirion_Humidity_Sensors_SHTC3_Datasheet.pdf


src/devices/README.md
```diff
@@ -63,6 +63,7 @@ Our vision: the majority of .NET bindings are written completely in .NET languag
 * [MPU6500/MPU9250 - Gyroscope, Accelerometer, Temperature and Magnetometer (MPU9250 only)](Mpu9250/README.md)
 * [Sense HAT](SenseHat/README.md)
 * [SHT3x - Temperature & Humidity Sensor](Sht3x/README.md)
+* [SHTC3 - Temperature & Humidity Sensor](Shtc3/README.md)
```
krwq, line 66: please do not modify this file, it's auto-generated

src/devices/Shtc3/Measure.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// Temperature
+        /// </summary>
+        public Temperature Temperature { get; }
+
+        /// <summary>
+        /// Humidity
+        /// </summary>
+        public double Humidity { get;  }
+    }
+
```
krwq, line 35: nit: extra enter

src/devices/Shtc3/Measure.cs
```diff
@@ -0,0 +1,36 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Iot.Units;
+
+namespace Iot.Device.Shtc3
+{
+    /// <summary>
+    /// Shtc3 measure
+    /// </summary>
+    public readonly struct Measure
```
krwq, line 12: - humidty arg should contain units in name and description (i.e. relativeHumidity), note that name relativeHumidity implies return value from 0..1.
- Please call this Shtc3Measurement or have an overload like `bool TryRead(out Temperature temperature, out double relativeHumidity)` (we currently use that instead of separate struct in many places)

src/devices/Shtc3/README.md
```diff
@@ -0,0 +1,26 @@
+# SHTC3 - Temperature & Humidity Sensor
+SHTC3 is a digital humidity and temperature sensor designed especially for battery-driven high-volume consumer electronics application.
```
krwq, line 2: empty line after title (otherwise in some special cases formatting might get broken when editing)

src/devices/Shtc3/README.md
```diff
@@ -0,0 +1,26 @@
...
+To reduce power cosumption this project use capability of sensor to allow measurement in low power mode and active sleep mode.
+
+## Usage
+```C#
```
krwq, line 6: this actually sometimes misrenders if there is no extra enter between title and code block

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,290 @@
...
+                if (value == _status)
+                {
+                    return;
+                }
+
+                if (value == Status.Idle)
+                {
+                    Wakeup();
+                }
+                else
```
krwq, line 65: this should be more explicit and if value is out of range you should throw ArgumentOutOfRangeException

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+        /// <summary>
+        /// Try read Temperature and Humidity
+        /// </summary>
+        /// <param name="sensorMeasure">measure return by sensor</param>
+        /// <param name="LowPower">"true" measure in low power mode, "false"(default) measure in normal power mode</param>
+        /// <param name="ClockStretching">"true" allow clock stretching, "false"(default) without clock stretching</param>
+        /// <returns></returns>
+        public bool TryGetTempAndHumi(out Measure sensorMeasure, bool LowPower = false, bool ClockStretching = false)
```
krwq, line 81: - no abbreviations in names (Humi => Humidity)
- the name suggests you will return two values and not structure consider just having two out args and remove the structure
- argument names should start with lower case (LowPower, ClockStretching => lowPower, clockStretching)

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,290 @@
...
+        private bool TryReadSensorData(Register cmd, out double temperature, out double humidity)
+        {
+            Write(cmd);
+
+            Span<byte> readBuff = stackalloc byte[6];
+
+            _i2cDevice.Read(readBuff);
+
+            // Details in the Datasheet P7
+            int st = (readBuff[0] << 8) | readBuff[1];      // Temp
```
krwq, line 106: consider using `BinaryPrimitives.ReadInt16LittleEndian`/`BinaryPrimitives.ReadInt16BigEndian`

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,290 @@
...
+            int srh = (readBuff[3] << 8) | readBuff[4];     // Humi
+
+            // check 8-bit crc
+            bool tCrc = CheckCrc8(readBuff.Slice(0, 2), readBuff[2]);
+            bool rhCrc = CheckCrc8(readBuff.Slice(3, 2), readBuff[5]);
+            if (tCrc == false || rhCrc == false)
```
krwq, line 112:
```suggestion
            if (!tCrc || !rhCrc)
```

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,290 @@
...
+                humidity = double.NaN;
+                return false;
+            }
+
+            temperature = st;
+            humidity = srh;
+            return true;
+        }
+
+        private static Register GetMeasurementCmd(bool LowPower, bool ClockStretching)
```
krwq, line 124: arguments should be lowerCase

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        /// <summary>
+        /// SHTC3 Soft Reset
+        /// </summary>
+        public void Reset()
+        {
+            Write(Register.SHTC3_RESET);
+        }
+
+        private int _id = 0;
```
krwq, line 173: fields should be first in the class

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        /// <summary>
+        /// Sensor Id
+        /// </summary>
+        public int Id
+        {
+            get
+            {
+                if (_id == 0)
```
krwq, line 182: is 0 invalid number? if not then use different value (perhaps -1, if all values are fine as ID then use `int?`)

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,290 @@
...
+                {
+                    Id = ReadId();
+                }
+
+                return _id;
+            }
+            set
+            {
+                _id = value;
```
krwq, line 191: I don't think this should be settable

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,290 @@
...
+
+            return crc == crc8;
+        }
+
+        private void Write(Register register)
+        {
+            byte msb = (byte)((short)register >> 8);
+            byte lsb = (byte)((short)register & 0xFF);
+
+            Span<byte> writeBuff = stackalloc byte[]
```
krwq, line 269: you can use BinaryPrimitives to Write Int16 directly to span

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,290 @@
...
+            {
+                msb,
+                lsb
+            };
+
+            _i2cDevice.Write(writeBuff);
+
+            // wait SCL free
+            DelayHelper.DelayMilliseconds(20, false);
```
krwq, line 278: please point to doc page where is 20 coming from

src/devices/Shtc3/Shtc3.csproj
```diff
@@ -0,0 +1,34 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
krwq, line 1: please base your project on others

src/devices/Shtc3/Shtc3.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <Compile Include="..\Common\System\Device\DelayHelper.cs" Link="common\DelayHelper.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="common\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
+    <ProjectReference Include="..\Units\Units.csproj" />
```
krwq, line 31: this needs an update, we've recently started using UnitsNet package instead (which is automatically dependency on every device, just simply remove this reference)

famblard86:
thank you, I will work on those.

src/devices/Shtc3/Shtc3.cs
famblard86: just based on SHT3x, maybe the table 5 page 4  suggest to reduce to 15 ms ?

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,287 @@
...
+        public int? Id
+        {
+            get
+            {
+                if (_id is null)
+                {
+                    _id = ReadId();
+                }
+
+                return _id;
```
krwq, line 194: consider: `return _id = _id ?? ReadId();`

src/devices/Shtc3/Shtc3.csproj
```diff
@@ -0,0 +1,20 @@
...
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <Compile Include="..\Common\System\Device\DelayHelper.cs" Link="DelayHelper.cs" />
+    <None Include="README.md" />
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="UnitsNet" Version="4.58.0" />
```
krwq, line 17: we define this globally, please remove

famblard86: sorry @krwq i don't understand ? without UnitsNet nugget the build fail

pgrawehr:
There were build errors. You'll likely have to rebase your branch, as some method have changed their return types.

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,282 @@
...
+                }
+
+                _status = value;
+            }
+        }
+
+        /// <summary>
+        /// Try read Temperature and Humidity
+        /// </summary>
+        /// <param name="temperature">Temperature return by sensor</param>
```
krwq, line 84:
```suggestion
        /// <param name="temperature">Temperature returned by sensor</param>
```

same below

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,282 @@
...
+        /// <param name="relativeHumidity">Humidity return by sensor</param>
+        /// <param name="lowPower">"true" measure in low power mode, "false"(default) measure in normal power mode</param>
```
krwq, line 86:
```suggestion
        /// <param name="lowPower">"true" measured in low power mode, "false"(default) measured in normal power mode</param>
```

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,282 @@
...
+        /// <param name="clockStretching">"true" allow clock stretching, "false"(default) without clock stretching</param>
```
krwq, line 87:
```suggestion
        /// <param name="clockStretching">"true" allow clock stretching, "false" (default) without clock stretching</param>
```

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,282 @@
...
+        /// <returns></returns>
```
krwq, line 88:
```suggestion
        /// <returns>True if operation was successful</returns>
```

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,282 @@
...
+
+            _i2cDevice.Read(readBuff);
+
+            // check 8-bit crc
+            if (!CheckCrc8(readBuff.Slice(0, 2), readBuff[2]))
+            {
+                return null;
+            }
+
+            var id = BinaryPrimitives.ReadInt16BigEndian(readBuff.Slice(0, 2));
```
krwq, line 210:
```suggestion
            int id = BinaryPrimitives.ReadInt16BigEndian(readBuff.Slice(0, 2));
```

src/devices/Shtc3/Shtc3.cs
krwq: fine as is

src/devices/Shtc3/samples/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+            I2cConnectionSettings settings = new I2cConnectionSettings(1, Iot.Device.Shtc3.Shtc3.I2cAddress);
+            I2cDevice device = I2cDevice.Create(settings);
+
+            using (Shtc3 sensor = new Shtc3(device))
+            {
+                Console.WriteLine($"Sensor Id: {sensor.Id}");
+                while (true)
+                {
+                    // Set sensor ready to use because while apply sleep mode bellow
+                    sensor.Status = Status.Idle;
```
krwq, line 26: I'd prefer the status was only exposed by method/properties rather than direct access, i.e.:

```csharp
sensor.Sleep(); // sets status to sleep
sensor.WakeUp(); // and/or call it automatically on use
// not sure what Status = Idle is actually doing...
// etc
```

src/devices/Shtc3/samples/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        private static void ConsoleWriteInfo(Temperature temperature, Ratio relativeHumidity)
+        {
+            Console.WriteLine($"Temperature: {temperature.DegreesCelsius:0.#}\u00B0C");
+            Console.WriteLine($"Humidity: {relativeHumidity.Percent:0.#}%");
+            // WeatherHelper supports more calculations, such as saturated vapor pressure, actual vapor pressure and absolute humidity.
+            Console.WriteLine($"Heat index: {WeatherHelper.CalculateHeatIndex(temperature, relativeHumidity).DegreesCelsius:0.#}\u00B0C");
+            Console.WriteLine($"Dew point: {WeatherHelper.CalculateDewPoint(temperature, relativeHumidity).DegreesCelsius:0.#}\u00B0C");
+        }
+
```
krwq, line 59: nit: extra enter here and below

krwq:
@famblard86 sorry for the delay, we're busy with shipping .NET 5... I'll try to review quicker from now on. Overall PR looks good to me. Main feedback is to not expose Status directly and instead use methods/properties

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,282 @@
...
+    public class Shtc3 : IDisposable
+    {
+        // CRC const
+        private const byte CRC_POLYNOMIAL = 0x31;
+        private const byte CRC_INIT = 0xFF;
+
+        /// <summary>
+        /// Default I2C address
+        /// </summary>
+        public const int I2cAddress = 0x70;
```
Ellerbach, line 26: Maybe `DefaultI2cAddress`? Will be more consistent with other bindings.

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,282 @@
...
+        private I2cDevice _i2cDevice;
+        private int? _id = null;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Shtc3"/> class.
+        /// </summary>
+        /// <param name="i2cDevice">The I2C device used for communication.</param>
+        public Shtc3(I2cDevice i2cDevice)
+        {
+            _i2cDevice = i2cDevice;
```
Ellerbach, line 37: Maybe check that `i2cDevice` is not null?

src/devices/Shtc3/Shtc3.cs
```diff
@@ -0,0 +1,282 @@
...
+            }
+
+            temperature = st;
+            humidity = srh;
+            return true;
+        }
+
+        private static Register GetMeasurementCmd(bool lowPower, bool clockStretching)
+        {
+            Register mea = Register.SHTC3_MEAS_T_RH_POLLING_NPM;
```
Ellerbach, line 135: rather than allocating this Register, just return for each case the result.
like:
```csharp
if (clockStretching)
{
    return Register.SHTC3_MEAS_T_RH_CLOCKSTR_LPM;
}
```

src/devices/Shtc3/Shtc3.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
Ellerbach, line 4: Should be 3.1 now

Ellerbach (approved):
Thanks for this binding! Couple of comments on my side on top of @krwq ones.

Ellerbach (approved):
Thanks! Looks good to me now

src/devices/Shtc3/Status.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Shtc3
+{
+    /// <summary>
+    /// Shtc3 Status
+    /// </summary>
+    public enum Status
```
krwq, line 10: should this be private?

krwq:
LGTM, @famblard86 should Status be internal now?

famblard86:
yes thank you has fix it

krwq (approved):

krwq:
Thank you @famblard86!

joperezr:
Seems like CI didn't run for that last change for some reason and it broke the build with:

`error CS0053: Inconsistent accessibility: property type 'Status' is less accessible than property 'Shtc3.Status'`

I'll make a change now to fix that

Issue #1068, opened 2020-05-05
Switch samples to PackageRef


richlander:
The samples in this repo use project references, instead of PackageReference to the NuGet packages (that the repo produces) on NuGet.org. This has two (related) problems:

- The projects can only be built as-is within the repo.
- The projects do not demonstrate the best-practice for customer usage. For example, the packages they would need are not even listed in a comment in the projects.

Examples:

- https://github.com/dotnet/iot/blob/master/samples/led-blink/led-blink.csproj
- https://github.com/dotnet/iot/blob/master/samples/force-sensitive-resistor/force-sensitive-resistor.csproj

richlander:
I added a proposal for how to handle this in #1069.

joperezr:
We can totally do this but we have to be aware of the trade-offs. This would mean that if we make a breaking change on either the main library(not very likely) or a binding the sample uses (somewhat likely) then we would have issues like:
- Passing CI when making the breaking change but then getting a broken CI on the next build when we consume the new packages with the changes. This in case we exclusively use package reference on samples.
- Failing CI when making the breaking changes so we adjust the samples on the same PR, but if people grab the project and paste it outside the repo it would just be broken as the package ref wouldn't satisfy the code. This in case we use your proposal.

Also one question, would we do this to only depend on packages from NuGet.org? If so, those get updated even less often than daily ones so problems might accumulate and samples might fall out of date. I'm not saying we shouldn't do it, I just want to mention the trade-offs we would see.

richlander:
Resolved in https://github.com/dotnet/iot/pull/1069

PR #1069, opened 2020-05-05
Add PreferPackageReference

richlander:
This is a proposal for #1068


samples/led-blink/led-blink.csproj
```diff
@@ -2,12 +2,26 @@
...
+   -->
+
+  <ItemGroup Condition="!$(PreferPackageReference)">
     <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="$(PreferPackageReference)">
+    <PackageReference Include="System.Device.Gpio" Version="1.1.0-prerelease.20153.1" />
+  </ItemGroup>
+
```
rainersigwald, line 26: Maybe something like this, to simplify the "if you're copy-pasting this for your own example" steps?

```suggestion
  <ItemGroup>
    <PackageReference Include="System.Device.Gpio" Version="1.1.0-prerelease.20153.1" />
  </ItemGroup>

  <!--  This sample uses conditional references, but you only need one of them.
        The ProjectReference is used by the repo build, to ensure that the samples work
        with new builds of the System.Device.Gpio project.
        The PackageReference is intended for most user use-cases.
        If you want to test the package, type `dotnet build /p:PreferPackageReference=true`
        If you want to copy this project and use outside of the repo, then
        delete the second ItemGroup.
   -->
  <ItemGroup Condition="'$(PreferPackageReference)' != 'true'">
    <PackageReference Remove="System.Device.Gpio" />
    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
  </ItemGroup>


```

richlander:
Related: https://github.com/NuGet/Home/issues/8213#issuecomment-502787749

richlander:
@dsplaisted -- good idea. This repo produces two projects/packages. Not all the samples use both of them. That said, I assume it likely isn't a big deal to add both ProjectRefs??

Here is a sample that uses both: https://github.com/dotnet/iot/blob/master/samples/force-sensitive-resistor/force-sensitive-resistor.csproj

The sample in the PR uses just one of them.

richlander:
@joperezr -- i can update all the samples to this scheme if you are good with it.

samples/led-blink/led-blink.csproj
```diff
@@ -2,11 +2,26 @@
...
+  Details:
+  This sample includes references to one or more NuGet packages as both ProjectReference and PackageReference.
+  The PackageReferences are intended for most use cases.
+  The project (as-is) will prefer the ProjectReference. If you want to test the project with the package, type: 
+  dotnet build /p:PreferPackageReference=true
+  The ProjectReference is used by the repo build to test project updates with the samples before they are published 
+  as packages to NuGet.org. This also ensures that the samples don't break.
+  -->
+
+  <ItemGroup Condition="'$(PreferPackageReference)' != 'true'">
```
joperezr, line 24: NIT: you could make this condition exclusive by default so people could still copy paste and have it working without the need to delete an ItemGroup. For example:

```xml
<ItemGroup Condition="'$(ForceProjectReference)' != 'true'">
  <PackageReference Include="...
</ItemGroup>

<ItemGroup Condition="'$(ForceProjectReference)' == 'true'">
  <ProjectReference Include="...
</ItemGroup>
```

This way just setting that property to true in the samples/Directory.Build.props will cause the ProjectReference to be used, but if the project gets copied outside of this repo, it just flips to use PackageReference instead without people having to edit it.

dsplaisted:
@richlander I was thinking of an MSBuild item transform in the Directory.Build.targets.  Something like this (though my syntax may be wrong):

```xml
<ItemGroup Condition="'$(ForceProjectReference)' == 'true'">
  <ProjectReference Include="%(PackageReference->'$(MainLibraryPath)System.Device.Gpio.csproj')"
                    Condition="'%(PackageReference.Identity)' == 'System.Device.Gpio'" />
  <ProjectReference Include="%(PackageReference->'$(IotBindingsLibraryPath)Iot.Device.Bindings.csproj')"
                    Condition="'%(PackageReference.Identity)' == 'Iot.Device.Bindings'" />
</ItemGroup>
```

joperezr:
>I was thinking of an MSBuild item transform in the Directory.Build.targets. Something like this (though my syntax may be wrong):

That would be even cleaner because then the sample .csproj would just look like a regular project. We can also then add a Target into the .targets file that runs as InitialTarget (so it runs before restore) and removes them from PackageReference in case ForceProjectReference is set. Does that sound right @dsplaisted ?

richlander:
@joperezr  The current state of the PR has that characteristic (regular project). I think we're now down to how to the best syntax for D.B.p.

joperezr:
What happens if you have both projectReference and PackageReference to the same project?

richlander:
@joperezr  -- see this: https://github.com/NuGet/Home/issues/8213#issuecomment-502787749

samples/Directory.Build.props
```diff
@@ -12,5 +12,26 @@
     <IsPackable>false</IsPackable>
     <MainLibraryPath>$(MSBuildThisFileDirectory)../src/System.Device.Gpio/</MainLibraryPath>
     <IotBindingsLibraryPath>$(MSBuildThisFileDirectory)../src/Iot.Device.Bindings/</IotBindingsLibraryPath>
+    <PreferPackageReference>false</PreferPackageReference>
```
joperezr, line 15: NIT: you don't need this. You are not checking anywhere if it is set to false, you are just checking if it is not true, whcih by default that's the case (it is empty by default).

samples/Directory.Build.props
```diff
@@ -12,5 +12,26 @@
     <IsPackable>false</IsPackable>
     <MainLibraryPath>$(MSBuildThisFileDirectory)../src/System.Device.Gpio/</MainLibraryPath>
     <IotBindingsLibraryPath>$(MSBuildThisFileDirectory)../src/Iot.Device.Bindings/</IotBindingsLibraryPath>
+    <PreferPackageReference></PreferPackageReference>
```
joperezr, line 15:
```suggestion
```

I meant you can just remove it entirely. When you don't have it defined it evaluates to empty, for example if you were to use `'$(SomeRandomPropertyThatDoesNotGetDefinedAnywhereAtAll)'` will evaluate to `''`

joperezr:
Changes look good to me. Two more thoughts before we merge:
- If we are only referencing versions that go into NuGet, that means that these guys will get versioned every month or so, which means it is possible that if people use the package reference they may have a broken project if the samples already are using newer/modified API only available from the projectReference. It's ok if we are ok with it, just wanted to mention it.
- Can we apply these changes to the rest of the samples on the same PR?

richlander:
I thought about your first point. There is no solution to that, other than publishing preview builds relatively often (which I think would be good).

Ya, I'll update the rest of the projects.

samples/Directory.Build.props
```diff
@@ -13,4 +13,24 @@
...
+    This enables building the samples with the latest version of those two dependent projects, by default.
+    If changes to these projects break the samples, then the build will break.
+    
+    If you want to use the PackageReferences instead, set the PreferPackageReference property to true, either
+    in the project or on the commandline, like:
+
+    dotnet build /p:PreferPackageReference=true
+
+    The samples can be copied outside the repo, and will use the PackageReferences.
+    There is a comment each of the samples. Once the samples are copied outside the repo, that comment is no longer meaningful.
```
rainersigwald, line 28: What comment?

samples/Directory.Build.props
```diff
@@ -13,4 +13,24 @@
...
+    This file adds two ProjectReferences to all sample projects (even if they don't depend on one or both projects).
+    This enables building the samples with the latest version of those two dependent projects, by default.
+    If changes to these projects break the samples, then the build will break.
+    
+    If you want to use the PackageReferences instead, set the PreferPackageReference property to true, either
+    in the project or on the commandline, like:
+
+    dotnet build /p:PreferPackageReference=true
+
+    The samples can be copied outside the repo, and will use the PackageReferences.
```
rainersigwald, line 27:
```suggestion
    The samples can be copied outside the repo, and will use the PackageReferences as long as this file isn't included.
```

?

samples/serialport-arduino/arduino-demo.csproj
```diff
@@ -4,12 +4,12 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <RootNamespace>arduino_demo</RootNamespace>
-    <!-- this can also be passed from the command line (i.e. dotnet publish -r linux-arm) -->
     <RuntimeIdentifier>linux-arm</RuntimeIdentifier>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.IO.Ports" Version="$(SystemIOPortsPackageVersion)" />
+    <PackageReference Include="System.Device.Gpio" Version="1.1.0-prerelease.20153.1" />
```
joperezr, line 11: Just as a comment we will have to come up with a way of updating all of these versions on all of our packages every time we ship a new NuGet.

samples/Directory.Build.props
richlander: Funny. I removed that. Will change the wording in this file.

samples/serialport-arduino/arduino-demo.csproj
richlander: Yes. We should be doing that anyway.

PR #1072, opened 2020-05-09
Feature units

krwq:
I think we should merge this branch and work directly on master. I have pretty much no objection to using UnitsNet


pgrawehr:
I'm in favor of merging this. I've already updated a few bindings and will present a PR, but I need to do a bit of testing first. 

I've got a question, though: How would we go about adding special units, structures or extensions? I.e. for NMEA or GPS support, we'd need a "GeographicPosition" entity, and for some purposes we might want to have extension methods to units (Anlge normalization, or more physical formulas). 

Where would we put that, now that the entire "Units" namespace is gone?

JTrotta:
For NMEA take a look a this well working [parser](https://github.com/deriven/hologps/blob/master/src/NMEA/NMEAParser.cs)

krwq:
@pgrawehr I'd recommend starting with an issue directly on UnitsNet github, if they don't want to solve the issue then we'll think about it.

As a temporary solution we should put a structure wherever we need it.

@JTrotta we cannot base a code on anything GPL licensed.

krwq:
cc: @joperezr

src/devices/Directory.Build.props
```diff
@@ -10,4 +10,8 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="UnitsNet" Version="4.58.0" />
```
joperezr, line 15: NIT: should we instead just add this to the projects that actually require it? I think that is cleaner. Also the version we depend on should be moved to Versions.props.

krwq: Iot.Device.Bindings depends on it anyway, for all tiny project this won't change anything anywaym this way we don't need to pollute csprojes... Once (if) we split projects into packages then I'd change that

pgrawehr: I'm with krwq here. Once this is properly used, there will be few projects _not_ requiring it (maybe some output devices).

joperezr: That's fair, I'm fine with keeping it here then. That said, I do think we still want to move the version to Versions.props on this PR since it is actually used in a few places so we want to centralize the version.

src/devices/OneWire/FamilyBindings/OneWireThermometerDevice.cs
```diff
@@ -4,7 +4,7 @@
 
 using System.Collections.Generic;
 using System.Threading.Tasks;
-using Iot.Units;
+using UnitsNet;
```
joperezr, line 7: NIT: looks like this isn't used

krwq:
Seems UnitsNet using was needed in the OneWireThermometerDevice (it uses Temperature type but not any of the methods since it calls into something else to get it)

krwq:
Merging since only done very minimal changes since approve

PR #1073, opened 2020-05-09
Bmp280 test

### LINKED ISSUE #1027: Bmp280 Sensor gives wrong readings
While testing the Bmp280 sensor, I found that mine was giving very wrong readings (~940hPa when it should be in the order of ~1010hPa). There's definitelly no storm outside now, and it's cold but sunny. 

Investigating further, I found a small bug in the calibration calculation for temperatures: 

Line 181 of Bmxx80Base.cs should be changed to: 
```
double var1 = ((((double)adcTemperature / 16384.0) - ((double)_calibrationData.DigT1) / 1024.0)) * (double)_calibrationData.DigT2;
```            
There are missing parentheses in that formula. 

I'll come up with a patch for this (including an unit test), however, since my sensor still gives the wrong values (also with the python reference implementation), it's hard for me to know that's really all. 

See also #1022

CC @djaus2

pgrawehr:
Fixes #1027

Reading is apparently correct now. Added self test for math formula and extended example a bit to show some weatherhelper usage.

### DIFF
--- a/src/devices/Bmxx80/Bme680.cs
+++ b/src/devices/Bmxx80/Bme680.cs
@@ -19,7 +19,8 @@ namespace Iot.Device.Bmxx80
     /// </summary>
     public class Bme680 : Bmxx80Base
     {
-        private readonly Temperature _ambientTemperatureDefault;
+        private static readonly Temperature DefaultAmbientTemperature = Temperature.FromDegreesCelsius(20);
+        private readonly Temperature _ambientTemperatureUserDefault;
 
         /// <summary>
         /// Default I2C bus address.
@@ -66,7 +67,7 @@ namespace Iot.Device.Bmxx80
         public Bme680(I2cDevice i2cDevice, Temperature ambientTemperatureDefault)
             : base(DeviceId, i2cDevice)
         {
-            _ambientTemperatureDefault = ambientTemperatureDefault;
+            _ambientTemperatureUserDefault = ambientTemperatureDefault;
             _communicationProtocol = CommunicationProtocol.I2c;
         }
 
@@ -75,7 +76,7 @@ namespace Iot.Device.Bmxx80
         /// </summary>
         /// <param name="i2cDevice">The <see cref="I2cDevice"/> to create with.</param>
         public Bme680(I2cDevice i2cDevice)
-            : this(i2cDevice, Temperature.FromDegreesCelsius(20))
+            : this(i2cDevice, DefaultAmbientTemperature)
         {
         }
 
@@ -466,7 +467,7 @@ namespace Iot.Device.Bmxx80
             _bme680Calibration = (Bme680CalibrationData)_calibrationData;
             if (!TryReadTemperature(out var temp))
             {
-                temp = _ambientTemperatureDefault;
+                temp = _ambientTemperatureUserDefault;
             }
 
             ConfigureHeatingProfile(Bme680HeaterProfile.Profile1, Temperature.FromDegreesCelsius(320), Duration.FromMilliseconds(150), temp);
--- a/src/devices/Bmxx80/Bme680.cs
+++ b/src/devices/Bmxx80/Bme680.cs
@@ -19,6 +19,8 @@ namespace Iot.Device.Bmxx80
     /// </summary>
     public class Bme680 : Bmxx80Base
     {
+        private readonly Temperature _ambientTemperatureDefault;
+
         /// <summary>
         /// Default I2C bus address.
         /// </summary>
@@ -59,10 +61,22 @@ namespace Iot.Device.Bmxx80
         /// Initialize a new instance of the <see cref="Bme680"/> class.
         /// </summary>
         /// <param name="i2cDevice">The <see cref="I2cDevice"/> to create with.</param>
-        public Bme680(I2cDevice i2cDevice)
+        /// <param name="ambientTemperatureDefault">Assumed ambient temperature for startup. Used for initialization of the gas measurement
+        /// if the temperature cannot be read during a reset.</param>
+        public Bme680(I2cDevice i2cDevice, Temperature ambientTemperatureDefault)
             : base(DeviceId, i2cDevice)
         {
+            _ambientTemperatureDefault = ambientTemperatureDefault;
             _communicationProtocol = CommunicationProtocol.I2c;
+        }
+
+        /// <summary>
+        /// Initialize a new instance of the <see cref="Bme680"/> class.
+        /// </summary>
+        /// <param name="i2cDevice">The <see cref="I2cDevice"/> to create with.</param>
+        public Bme680(I2cDevice i2cDevice)
+            : this(i2cDevice, Temperature.FromDegreesCelsius(20))
+        {
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


pgrawehr:
@krwq: I think this is ready for review, should be quick. It does not much more than adding a test to the BMP280 implementation.

src/devices/Bmxx80/Bme680.cs
```diff
@@ -450,7 +450,11 @@ protected override void SetDefaultConfiguration()
             FilterMode = Bme680FilteringMode.C0;
 
             _bme680Calibration = (Bme680CalibrationData)_calibrationData;
-            TryReadTemperature(out var temp);
+            if (!TryReadTemperature(out var temp))
+            {
+                temp = Temperature.FromDegreesCelsius(22.0);
```
krwq, line 455: I'm confused - 22C if we can't read? would it make sense to throw or something?

src/devices/Bmxx80/Bmx280Base.cs
```diff
@@ -272,34 +270,31 @@ protected override void SetDefaultConfiguration()
...
-            if (var1 == 0)
+            // This uses the recommended approach with floating point math
+            double var1, var2, p;
+            var1 = (TemperatureFine / 2.0) - 64000.0;
+            var2 = var1 * var1 * ((double)_calibrationData.DigP6) / 32768.0;
+            var2 = var2 + var1 * ((double)_calibrationData.DigP5) * 2.0;
+            var2 = (var2 / 4.0) + (((double)_calibrationData.DigP4) * 65536.0);
+            var1 = (((double)_calibrationData.DigP3) * var1 * var1 / 524288.0 + ((double)_calibrationData.DigP2) * var1) / 524288.0;
+            var1 = (1.0 + var1 / 32768.0) * ((double)_calibrationData.DigP1);
+            if (var1 == 0.0)
```
krwq, line 288: can we still hit the case where this is 0?

src/devices/Bmxx80/Bmxx80Base.cs
```diff
@@ -182,7 +182,7 @@ protected Temperature CompensateTemperature(int adcTemperature)
             double var2 = (adcTemperature / 131072.0) - (_calibrationData.DigT1 / 8192.0);
             var2 *= var2 * _calibrationData.DigT3 * TempCalibrationFactor;
 
-            TemperatureFine = (int)(var1 + var2);
+            TemperatureFine = (var1 + var2);
```
krwq, line 185: brackets are not necessary

src/devices/Bmxx80/Bme680.cs
pgrawehr: I just selected a reasonable default, previously the return value was ignored, which made the variable undefined (note that with the usage of UnitsNet, the value is not necessarily Zero, but really unusable). 

I'm not sure I want to throw here, as it would eventually cause the ctor to throw, just because an initial temperature couldn't be read.

src/devices/Bmxx80/Bmx280Base.cs
pgrawehr: Unlikely, but the function is taken directly from the spec sheet, and there this condition is used.

src/devices/Bmxx80/Bme680.cs
krwq: I'd rather have this something more obvious it's invalid, i.e. FromKelvin(0) or something

pgrawehr: It should be a reasonable ambient temperature. It is used to pre-configure the heater of the gas sensor. A small error will not matter much, as that heater is heats up to 250-350 °C, but 0°K is very far off (and unlikely to be accurate, unless the world has ended). The spec says to use the temperature measurement or a hardcoded value. 

Maybe I should make it an (optional) parameter?

pgrawehr: @krwq: Ping. I'm not sure how to best proceed here.

pgrawehr:
@krwq: Any idea what's wrong with the Windows builds? They seem to be failing quite often.

src/devices/Bmxx80/Bme680.cs
pgrawehr: Updated with an argument

pgrawehr:
@krwq : Ping. Can we finally close this, please?

src/devices/Bmxx80/Bme680.cs
```diff
@@ -450,7 +464,11 @@ protected override void SetDefaultConfiguration()
             FilterMode = Bme680FilteringMode.C0;
 
             _bme680Calibration = (Bme680CalibrationData)_calibrationData;
-            TryReadTemperature(out var temp);
+            if (!TryReadTemperature(out var temp))
+            {
+                temp = _ambientTemperatureDefault;
```
joperezr, line 469: Should we do a Debug.WriteLine here somewhere to log that there was an issue reading the temperature?

src/devices/Bmxx80/Bme680.cs
```diff
@@ -59,12 +61,24 @@ public class Bme680 : Bmxx80Base
...
+            _ambientTemperatureDefault = ambientTemperatureDefault;
             _communicationProtocol = CommunicationProtocol.I2c;
         }
 
+        /// <summary>
+        /// Initialize a new instance of the <see cref="Bme680"/> class.
+        /// </summary>
+        /// <param name="i2cDevice">The <see cref="I2cDevice"/> to create with.</param>
+        public Bme680(I2cDevice i2cDevice)
+            : this(i2cDevice, Temperature.FromDegreesCelsius(20))
```
joperezr, line 78: NIT: instead of passing in 20, declare a static constant for this.

src/devices/Bmxx80/tests/Bmp280Tests.cs
```diff
@@ -0,0 +1,92 @@
...
+        DIG_P9 = 0x9E,
+
+        STATUS = 0xF3,
+        CONFIG = 0xF5,
+
+        PRESSUREDATA = 0xF7,
+        TEMPDATA_MSB = 0xFA
+    }
+
+    public sealed class Bmp280SensorTests : IDisposable
```
joperezr, line 43: Why do you need the tests to be disposable? Is it for test cleanup? if so we should probably call dispose on the _i2cDevice

joperezr:
Left a few comments but this looks good in general. Also I'm re-running CI now that Windows queues should be fixed

src/devices/Bmxx80/Bme680.cs
pgrawehr: Of course, we could do that (even inside the TryReadTemperature method), but we generally don't do this. In several other PRs/tickets we saw that a generic logging framework for the whole project would probably be helpful (i.e. an interface where one could attach a console, or an NLog writer or something). I'd rather go that way (in a separate PR).

src/devices/Bmxx80/tests/Bmp280Tests.cs
pgrawehr: Yes, forgot about that. Normally, you'd verify mocks in `Dispose()`, but here I finally had to write a "manual" mock, so forgot to clean that up.

src/devices/Bmxx80/tests/Bmxx80.Tests.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <LangVersion>latest</LangVersion>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <RootNamespace>Iot.Device.Imu.Tests</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.0.1" />
+    <PackageReference Include="xunit" Version="2.4.1" />
```
joperezr, line 13: NIT: we don't have to fix it on this PR, but these references to xunit on all of our UnitTests projects should be against the same version instead of what we have now where different projects depend on different versions. Perhaps we should even add some infrastructure so that if this is a unit test project then these references get automatically added for you so that we control and make sure all versions match.

PR #1075, opened 2020-05-14
Fixing the high resolution version of SoftwarePwmChannel

tannergooding:
The `usePrecisionTimer` variant of SoftwarePwmChannel didn't work on my PI 4 with Ubuntu 20.04, so this PR fixes it and does some other cleanup to the implementation.

### DIFF
--- a/src/devices/Common/System/Device/DelayHelper.cs
+++ b/src/devices/Common/System/Device/DelayHelper.cs
@@ -12,10 +12,51 @@ namespace System.Device
     /// </summary>
     internal static class DelayHelper
     {
-        // GetTimestamp() currently can take ~300ns. We hope to improve this to get better
-        // fidelity for very tight spins.
-        //
-        // SpinWait currently spins to approximately 1μs before it will yield the thread.
+        /* GetTimestamp() currently can take ~300ns. We hope to improve this to get better
+         * fidelity for very tight spins.
+         *
+         * SpinWait currently spins to approximately 1μs before it will yield the thread.
+         */
+
+        private const long TicksPerSecond = TimeSpan.TicksPerSecond;
+        private const long TicksPerMillisecond = TimeSpan.TicksPerMillisecond;
+        private const long TicksPerMicrosecond = TimeSpan.TicksPerMillisecond / 1000;
+
+        /// <summary>A scale that normalizes the hardware ticks to <see cref="TimeSpan" /> ticks which are 100ns in length.</summary>
+        private static readonly double s_tickFrequency = (double)TicksPerSecond / Stopwatch.Frequency;
+
+        /// <summary>
+        /// Delay for at least the specified <paramref name="time" />.
+        /// </summary>
+        /// <param name="time">The amount of time to delay.</param>
+        /// <param name="allowThreadYield">
+        /// True to allow yielding the thread. If this is set to false, on single-proc systems
+        /// this will prevent all other code from running.
+        /// </param>
+        public static void Delay(TimeSpan time, bool allowThreadYield)
+        {
+            long start = Stopwatch.GetTimestamp();
+            long delta = (long)(time.Ticks / s_tickFrequency);
+            long target = start + delta;
+
+            if (!allowThreadYield)
+            {
+                do
+                {
+                    Thread.SpinWait(1);
+                }
+                while (Stopwatch.GetTimestamp() < target);
+            }
+            else
+            {
+                SpinWait spinWait = new SpinWait();
+                do
+                {
+                    spinWait.SpinOnce();
+                }
+                while (Stopwatch.GetTimestamp() < target);
+            }
+        }
 
         /// <summary>
         /// Delay for at least the specified <paramref name="microseconds"/>.
@@ -27,26 +68,8 @@ namespace System.Device
         /// </param>
         public static void DelayMicroseconds(int microseconds, bool allowThreadYield)
         {
-            long start = Stopwatch.GetTimestamp();
-            ulong minimumTicks = (ulong)(microseconds * Stopwatch.Frequency / 1_000_000);
-
-            if (!allowThreadYield)
-            {
-                do
-                {
-                    Thread.SpinWait(1);
-                }
-                while ((ulong)(Stopwatch.GetTimestamp() - start) < minimumTicks);
-            }
-            else
-            {
-                SpinWait spinWait = new SpinWait();
-                do
-                {
-                    spinWait.SpinOnce();
-                }
-                while ((ulong)(Stopwatch.GetTimestamp() - start) < minimumTicks);
-            }
+            var time = TimeSpan.FromTicks(microseconds * TicksPerMicrosecond);
+            Delay(time, allowThreadYield);
         }
 
         /// <summary>
@@ -59,11 +82,14 @@ namespace System.Device
         /// </param>
         public static void DelayMilliseconds(int milliseconds, bool allowThreadYield)
         {
-            // We have this as a separate method for now to make calling code clearer
-            // and to allow us to add additional logic to the millisecond wait in the
-            // future. If waiting only 1 millisecond we still have ample room for more
-            // complicated logic. For 1 microsecond that isn't the case.
-            DelayMicroseconds(milliseconds * 1000, allowThreadYield);
+            /* We have this as a separate method for now to make calling code clearer
+             * and to allow us to add additional logic to the millisecond wait in the
+             * future. If waiting only 1 millisecond we still have ample room for more
+             * complicated logic. For 1 microsecond that isn't the case.
+             */
+
+            var time = TimeSpan.FromTicks(milliseconds * TicksPerMillisecond);
+            Delay(time, allowThreadYield);
         }
     }
 }
--- a/src/devices/SoftPwm/SoftwarePwmChannel.cs
+++ b/src/devices/SoftPwm/SoftwarePwmChannel.cs
@@ -2,51 +2,50 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Device.Gpio;
+using System.Device.Pwm;
 using System.Diagnostics;
 using System.Threading;
-using System.Threading.Tasks;
 
 namespace System.Device.Pwm.Drivers
 {
-    /// <summary>
-    /// Software PWM channel implementation
-    /// </summary>
+    /// <summary>Software PWM channel implementation</summary>
     public class SoftwarePwmChannel : PwmChannel
     {
-        private readonly bool _shouldDispose;
-        // how long the signal is high in its period
-        private double _pulseWidthInMilliseconds;
-        private double _periodInMilliseconds;
-        private int _frequency;
-        // Use to determine the length of the pulse
-        // 100% = 1.0 = full output. 0% = 0.0 = nothing as output
-        private double _dutyCycle;
+        // Frequency represents the number of times the pin should "pulse" (go from low to high and back) per second
+        // DutyCycle represents the percentage of time the pin should be in the high state
 
-        // Determines if a high precision timer should be used.
-        private bool _usePrecisionTimer;
+        // So, if the Frequency is 1 and the Duty Cycle is 0.5, the pin will go High once per second and stay on for 0.5 seconds
+        // While if the Frequency is 400 and the Duty Cycle is 0.5, the pin will go High 400 times per second staying on for 0.00125 seconds each time, for a total of 0.5 seconds
 
-        private bool _isRunning;
-        private bool _isStopped = true;
         private int _pin;
 
-        private Stopwatch _stopwatch = Stopwatch.StartNew();
+        private int _frequency;
+        private double _dutyCycle;
 
-        private Thread _runningThread;
         private GpioController _controller;
-        private bool _runThread = true;
+        private readonly bool _shouldDispose;
 
-        /// <summary>
-        /// The frequency in hertz.
-        /// </summary>
+        private TimeSpan _pinHighTime;
+        private TimeSpan _pinLowTime;
+
+        private Thread _thread;
+        private Stopwatch _stopwatch;
+
+        private bool _isRunning;
+        private bool _isTerminating;
+
+        /// <summary>The frequency in hertz.</summary>
         public override int Frequency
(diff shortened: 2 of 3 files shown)


src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -91,10 +86,20 @@ public SoftwarePwmChannel(int pinNumber, int frequency = 400, double dutyCycle =
 
             if (frequency <= 0)
             {
-                throw new ArgumentOutOfRangeException(nameof(frequency), "Frequency must be a positive value.");
+                throw new ArgumentOutOfRangeException(nameof(frequency), frequency, "Frequency must be a positive non-zero value.");
             }
 
-            if (controller == null)
+            if (dutyCycle < 0.0 || dutyCycle > 1.0)
```
tannergooding, line 92: `dutyCycle` wasn't being checked at all, this fixes that.

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -105,106 +110,89 @@ public SoftwarePwmChannel(int pinNumber, int frequency = 400, double dutyCycle =
...
-            _runningThread = new Thread(RunSoftPWM);
-            _runningThread.Start();
+            _thread.Start();
         }
 
         private void UpdatePulseWidthParameters()
         {
-            _periodInMilliseconds = 1000.0 / _frequency;
-            _pulseWidthInMilliseconds = _dutyCycle * _periodInMilliseconds;
+            double cycleTicks = TimeSpan.TicksPerSecond / (double)_frequency;
```
tannergooding, line 131: I'm not quite sure why the old logic used `1000` but the new logic just works based on `TimeSpan Ticks` which are well-defined to be `100ns`

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -105,106 +110,89 @@ public SoftwarePwmChannel(int pinNumber, int frequency = 400, double dutyCycle =
...
-                        Wait(_pulseWidthInMilliseconds);
-                    }
-                    else
-                    {
-                        Task.Delay(TimeSpan.FromMilliseconds(_pulseWidthInMilliseconds)).Wait();
-                    }
-
-                    // The pulse if over and so set the pin to low and then wait until it's time for the next pulse
-                    _controller.Write(_pin, PinValue.Low);
+                    Thread.Yield();
```
tannergooding, line 146: The old logic just wasted cycles in the background thread when the PWM wasn't running. This makes it a good citizen and yields until the user asks to start running again.

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -105,106 +110,89 @@ public SoftwarePwmChannel(int pinNumber, int frequency = 400, double dutyCycle =
...
-                        Task.Delay(TimeSpan.FromMilliseconds(_periodInMilliseconds - _pulseWidthInMilliseconds)).Wait();
-                    }
+                if (_pinHighTime != TimeSpan.Zero)
+                {
+                    _controller.Write(_pin, PinValue.High);
+                    Wait(_pinHighTime);
                 }
-                else
+
+                if (_pinLowTime != TimeSpan.Zero)
```
tannergooding, line 156: The old logic was checking if the high time was zero, but not if the low time was. We do both.

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -105,106 +110,89 @@ public SoftwarePwmChannel(int pinNumber, int frequency = 400, double dutyCycle =
...
 
         /// <summary>
         /// A synchronous wait is used to avoid yielding the thread
         /// This method calculates the number of CPU ticks will elapse in the specified time and spins
         /// in a loop until that threshold is hit. This allows for very precise timing.
         /// </summary>
-        /// <param name="milliseconds">The milliseconds to wait for</param>
-        private void Wait(double milliseconds)
+        /// <param name="ticks">The time to wait.</param>
+        private void Wait(TimeSpan timeout)
```
tannergooding, line 172: The wait logic was just centralized here. `_stopwatch` being initialized determines if we are high precision or not and we `Sleep` if we are not high precision. Using `Task.Delay` isn't really worthwhile if we aren't actually async.

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -105,106 +110,89 @@ public SoftwarePwmChannel(int pinNumber, int frequency = 400, double dutyCycle =
...
-            double desiredTicks = milliseconds / 1000.0 * Stopwatch.Frequency;
-            double finalTick = initialTick + desiredTicks;
-            while (_stopwatch.ElapsedTicks < finalTick)
+            if (_stopwatch is null)
+            {
+                Thread.Sleep(timeout);
+                return;
+            }
+
+            TimeSpan elapsed = _stopwatch.Elapsed;
```
tannergooding, line 180: The old logic was trying to use the raw ticks, but wasn't  necessarily account for large or small frequencies, which is where the logic was broken.
Some devices, for example, report a 1ns frequency while others report 100ns and others may report even lower resolutions (such as 1 micro or 1 millisecond).
The new logic just relies on the existing `Stopwatch.Elapsed` logic which normalizes the "raw ticks" into "timespan ticks" so everything can be cleanly and correctly handled.

src/devices/SoftPwm/SoftwarePwmChannel.cs
tannergooding: The old logic was also constantly turning the pin off, which isn't necessarily desirable if the pin is also being controlled by other logic.

krwq:
cc: @Ellerbach 

@tannergooding do you have any servo motors to test this with? This was the main scenario this was used with

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -105,106 +110,89 @@ public SoftwarePwmChannel(int pinNumber, int frequency = 400, double dutyCycle =
...
+        private void Wait(TimeSpan timeout)
         {
-            long initialTick = _stopwatch.ElapsedTicks;
-            long initialElapsed = _stopwatch.ElapsedMilliseconds;
-            double desiredTicks = milliseconds / 1000.0 * Stopwatch.Frequency;
-            double finalTick = initialTick + desiredTicks;
-            while (_stopwatch.ElapsedTicks < finalTick)
+            if (_stopwatch is null)
+            {
+                Thread.Sleep(timeout);
```
krwq, line 176: doesn't this have 10ms resolution?

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -105,106 +110,89 @@ public SoftwarePwmChannel(int pinNumber, int frequency = 400, double dutyCycle =
...
+                return;
+            }
+
+            TimeSpan elapsed = _stopwatch.Elapsed;
+            TimeSpan target = elapsed + timeout;
+
+            while (target > _stopwatch.Elapsed)
```
krwq, line 183: consider using DelayHelpers for that (we didn't have that when this was written I believe)

Ellerbach: No we did not :-) That was one of the very first PR done :-) Rest came later on

tannergooding:
> do you have oscilloscope or any servo motors to test this with? This was the main scenario for SoftPwm I believe

I don't have the former, but I do have the latter. I tested this with various LEDs and with a couple different motors and examined the numbers reported between the two.

If you print the various internal numbers at different points you can see several issues that would crop up:
1. For sufficiently high frequencies (say `65536`) with a low enough duty cycle (say `1/255`), you would end up with sub 100ns times (`LowTime: 5.9838E-05 milliseconds; HighTime: 0.0151989 milliseconds`) that would result in timings getting off and the odd flicker from the device. https://docs.microsoft.com/en-us/windows/win32/sysinfo/acquiring-high-resolution-time-stamps goes into more details about some of the issues you can hit with high precision timers like this, but it ultimately stems from even if the frequency is reported as say 1ns (like on Unix) it might take say 300ns to actually query the hardware and get a result back.
2. Because the values were being tracked as double, they were too overly precise and resulted in rounding errors where `LowTime + HighTime != Period`
3. Because we didn't conditionalize the low signal, you could end up with RGB LEDs glowing (because you connect + to the 3.3V and each color to a capacitor which goes to the GPIO)

src/devices/SoftPwm/SoftwarePwmChannel.cs
tannergooding: Looks like it doesn't currently support taking a `TimeSpan`. That overload could be added and then everything could be done in terms of 100ns ticks (which is likely the maxmimum resolution any hardware can realistically support due to latency, etc).

krwq:
@tannergooding I'm good with this as long as ServoMotor works reasonably with SoftPwm (I don't expect it to be as good as hardware)

src/devices/SoftPwm/SoftwarePwmChannel.cs
krwq: it would be useful to add such support, feel free to add it

tannergooding: I'll have an update for this sometime later this afternoon

tannergooding: Should be resolved. There is a new `Delay(TimeSpan time, bool allowThreadYield)` helper that allows delays as precise as 100ns (this will realistically be lower due to hardware latency in querying the timer).

tannergooding:
> I'm good with this as long as ServoMotor works reasonably with SoftPwm (I don't expect it to be as good as hardware)

At least on my own devices it appears to work correctly. It would likely be good if someone else tested as well (just for sanity) to validate there isn't anything expected that I wasn't aware of.

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -150,42 +153,19 @@ private void Run()
                 if (_pinHighTime != TimeSpan.Zero)
```
krwq, line 153: Can you add an assert so that `_pingHighTime` and `_pinLowTime` won't become zero while `_isRunning`

krwq:
@JeremyKuhne any chance you could take a look at the DelayHelpers changes?

src/devices/SoftPwm/SoftwarePwmChannel.cs
tannergooding: But they can become zero while running.

It is perfectly valid to change the Duty Cycle to `1.0` or `0.0` in which case the pin will always be high or always be low.

krwq: What I'm concerned about is that when both are zero/one this will use 100% of the CPU for no reason, for those two cases might make sense to special case to optimize CPU usage

tannergooding: Why would it use 100% CPU? You can't have both be zero and so you will always hit one of the two cases and then delay for `period`.

So if your Frequency is 60, then the period is `16.67ms`, if you have a duty cycle of `0.5` you will:
* Go High
* Delay for `8.33ms`
* Go Low
* Delay for `8.33ms`
* Repeat

Likewise, if the Duty Cycle is `1.0` you would:
* Go High
* Delay for `16.66ms`
* See LowTime is 0, so do nothing
* Repeat

tannergooding: For a Duty Cle of `0.0`, it would naturally be inversed:
* See HighTime is 0, so do nothing
* Go Low
* Delay for 16.66ms
* Repeat

krwq: I meant just sanity check that they both don't become zeros, either way I'm fine with this as is

krwq:
Thanks!

src/devices/Common/System/Device/DelayHelper.cs
```diff
@@ -45,10 +54,24 @@ public static void DelayMicroseconds(int microseconds, bool allowThreadYield)
                 {
                     spinWait.SpinOnce();
                 }
-                while ((ulong)(Stopwatch.GetTimestamp() - start) < minimumTicks);
+                while (Stopwatch.GetTimestamp() < target);
```
krwq, line 57: I think the previous logic here was preventing overflows (you'd get double overflow so would just work)

src/devices/SoftPwm/SoftwarePwmChannel.cs
tannergooding: Actually, it's possible for both to be zero.

Given there are two threads involved, the other thread can update the Duty Cycle between when `_pinHighTime` and `_pinLowTime` is read, and for that cycle it would end up with both being zero. It would then fix itself in the next cycle.

I don't think its worth trying to account for that given things will always normalize themselves in the cycle following the update (which is no more than 1s).

---

We could assert in `UpdatePulseWidthParameters` that `pinHighTicks` and `pinLowTicks` aren't both zero, but I think that's the closest to what you were asking that we could do?

src/devices/Common/System/Device/DelayHelper.cs
tannergooding: I don't think there is a real need to account for overflow here, it just makes the code harder to read at little benefit (we don't account for this in `System.Diagnostics.Stopwatch` either, for similar reasons).
Even if `Frequency` is 1ns ticks, it would take 292.3 years to overflow, in which case the fix is for the user to install updates and reboot their computer 😄

Windows currently returns 100ns ticks (but supports more precise), Unix/POSIX returns 1ns ticks (and is hardcoded to do so). Hardware has a latency of ~300ns to even query the timer.

krwq: fair enough 😄

src/devices/SoftPwm/SoftwarePwmChannel.cs
krwq: Seems like not worth to add a check

JeremyKuhne:
Awesome :shipit:

Issue #1074, opened 2020-05-13
Output-only SpiDevice?


HakanL:
Would it make sense to have an Output-only SpiDevice class? For example `SpiOutputDevice`, which the SpiDevice can inherit from. The use-case is for hardware devices that don't have any read-capabilities, like LED/Pixel/PWM controllers like the TLC59711. It would be nice to create an implementation for a device that could take a SpiOutputDevice in the constructor which could be implemented either in hardware or software (SoftwareSpi).

krwq:
I don't feel strongly but I'd rather go with parity with Stream so rather go in direction of having CanRead/CanWrite assuming we have a use case where it would mitigate accidental use. If your device is output only (i.e. LED strips are as well and we already use them) just don't call Read...

HakanL:
It would just allow for a cleaner class structure. Take for example the `SoftwareSpi` class, it requires chip select, miso input pins, but for some output-only devices those aren't available/used. If we had an OutputSpiDevice then we could just take that as a constructor parameter into a device-implementation and it would be up to the user to choose if they want to use a hardware SPI (SpiDevice), or a software-implementation. We would then have separation of concerns, each device wouldn't have to implement their own bit-banging (when hardware isn't available).
Further, the SpiDevice class is written in a way where you have to implement the TransferFullDuplex, which does both a read and a write. For an output-only device that doesn't make sense. Yes, it could be done without the proposed changes, but it would be a hack, isn't it better to create a nice abstraction in the base classes? SpiDevice would inherit from SpiOutputDevice for example.

krwq:
on the other hand you can't base your class on more than one abstract class - for SoftwareSpi I guess we could make it work if `-1` is used instead of legit pin number. I have mixed feelings on this - too much abstraction usually does more harm long term so better be sure about the abstractions you're adding

HakanL:
Correct, but the SpiDevice would inherit from SpiOutputDevice, not multiple base-classes (which obviously isn't possible), it would just be optional to create a SoftwareSpiOutput device, or just a SoftwareSpi (which would handle the outputs as well) depending on what your hardware can support.
I'm personally not a fan of hard-coded special numbers (like -1), but yes you could do some if/elses in there, but IMHO it would be cleaner to have the abstraction, especially since the main Write method in SoftwareSpi will do a read (via TransferFullDuplex), so you'd have to add a big chunk of if-cases.
[...]

PR #1076, opened 2020-05-15
Improve Software Spi

HakanL:
#1074 
Allows the chip select and miso to be optional
Re-built all samples, to my knowledge these changes should not have broken anything

### DIFF
--- a/src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
+++ b/src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
@@ -19,8 +19,8 @@ namespace System.Device.Spi
         /// Initializes a new instance of the <see cref="SpiConnectionSettings"/> class.
         /// </summary>
         /// <param name="busId">The bus ID the device is connected to.</param>
-        /// <param name="chipSelectLine">The chip select line used on the bus.</param>
-        public SpiConnectionSettings(int busId, int chipSelectLine)
+        /// <param name="chipSelectLine">The chip select line used on the bus. Optional, -1 if not used</param>
+        public SpiConnectionSettings(int busId, int chipSelectLine = -1)
         {
             BusId = busId;
             ChipSelectLine = chipSelectLine;
--- a/src/devices/SoftwareSpi/SoftwareSpi.cs
+++ b/src/devices/SoftwareSpi/SoftwareSpi.cs
@@ -25,10 +25,11 @@ namespace Iot.Device.Spi
         /// <summary>
         /// Software implementation of the SPI.
         /// </summary>
+        /// <remarks>Note that there is a ChipSelectLine in the SPIConnectionSettings as well, either that or the cs property will be used.</remarks>
         /// <param name="clk">Clock pin.</param>
         /// <param name="miso">Master Input Slave Output pin. Optional, set to -1 to ignore</param>
         /// <param name="mosi">Master Output Slave Input pin.</param>
-        /// <param name="cs">Chip select pin (or negated chip select). Optional, set to -1 to ignore. Note that there is a ChipSelectLine in the SPIConnectionSettings as well, either will be used.</param>
+        /// <param name="cs">Chip select pin (or negated chip select). Optional, set to -1 to ignore.</param>
         /// <param name="settings">Settings of the SPI connection.</param>
         /// <param name="controller">GPIO controller used for pins.</param>
         /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
@@ -43,12 +44,12 @@ namespace Iot.Device.Spi
             _miso = miso;
             _mosi = mosi;
 
-            if (_cs > -1 && _settings.ChipSelectLine > -1 && _cs != _settings.ChipSelectLine)
+            if (_cs != -1 && _settings.ChipSelectLine != -1 && _cs != _settings.ChipSelectLine)
             {
                 throw new ArgumentException("Both cs and settings.ChipSelectLine can't both be set");
             }
 
-            if (cs > -1)
+            if (cs != -1)
             {
                 _cs = cs;
             }
@@ -58,13 +59,13 @@ namespace Iot.Device.Spi
             }
 
             _controller.OpenPin(_clk, PinMode.Output);
-            if (_miso > -1)
+            if (_miso != -1)
             {
                 _controller.OpenPin(_miso, PinMode.Input);
             }
 
             _controller.OpenPin(_mosi, PinMode.Output);
-            if (_cs > -1)
+            if (_cs != -1)
             {
                 _controller.OpenPin(_cs, PinMode.Output);
             }
@@ -75,7 +76,7 @@ namespace Iot.Device.Spi
             // aka. CPHA - tells us when read/write is 'captured'
             bool onPulseEnd = ((int)_settings.Mode & 1) == 1;
 
-            if (_cs > -1)
+            if (_cs != -1)
             {
                 _controller.Write(_cs, !(bool)_settings.ChipSelectLineActiveState);
             }
@@ -113,6 +114,7 @@ namespace Iot.Device.Spi
             else
             {
                 _bitTransfer = new ScopeData(
+                    enter: () => { },
                     exit: () =>
                     {
                         _controller.Write(_clk, !idle);
@@ -120,7 +122,7 @@ namespace Iot.Device.Spi
                     });
             }
 
-            if (_cs > -1)
+            if (_cs != -1)
             {
                 _chipSelect = new ScopeData(
                     enter: () =>
@@ -131,6 +133,10 @@ namespace Iot.Device.Spi
                     {
                         _controller.Write(_cs, !(bool)_settings.ChipSelectLineActiveState);
                     });
+            }
+            else
+            {
+                _chipSelect = new ScopeData(() => { }, () => { });
             }
         }
 
@@ -195,11 +201,7 @@ namespace Iot.Device.Spi
 
         private bool ReadWriteBit(bool bitToWrite)
         {
-            if (_miso == -1)
-            {
-                throw new ArgumentException("Cannot read without a miso pin specified");
-            }
-
+            // _mosi is checked higher up the call path
             _controller.Write(_mosi, bitToWrite);
             return (bool)_controller.Read(_miso);
         }
@@ -212,6 +214,11 @@ namespace Iot.Device.Spi
         /// <inheritdoc />
         public override void Read(Span<byte> data)
         {
+            if (_miso == -1)
+            {
+                throw new ArgumentException("Cannot read without a miso pin specified");
+            }
+
             Span<byte> dataToWrite = stackalloc byte[data.Length];
             TransferFullDuplex(dataToWrite, data);
         }
@@ -219,19 +226,8 @@ namespace Iot.Device.Spi
         /// <inheritdoc />
         public override void Write(ReadOnlySpan<byte> data)
         {
-            if (_miso == -1)
-            {
-                // Output/Write only
-                TransferWriteOnly(data);
-
-                return;
-            }
-            else
-            {
-                // Write and read
-                Span<byte> dataToRead = stackalloc byte[data.Length];
-                TransferFullDuplex(data, dataToRead);
-            }
+            // Output/Write only
+            TransferWriteOnly(data);
         }
 
         /// <inheritdoc />
@@ -272,20 +268,20 @@ namespace Iot.Device.Spi
             internal Action _enter;
             internal Action _exit;
 
-            public ScopeData(Action enter = null, Action exit = null)
+            public ScopeData(Action enter, Action exit)
             {
-                _enter = enter;
-                _exit = exit;
+                _enter = enter ?? throw new ArgumentNullException(nameof(enter));
+                _exit = exit ?? throw new ArgumentNullException(nameof(exit));
             }
 
             public void Enter()
             {
-                _enter?.Invoke();
+                _enter.Invoke();
             }
 
             public void Exit()
             {
-                _exit?.Invoke();
+                _exit.Invoke();
             }
         }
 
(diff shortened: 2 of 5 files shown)


HakanL:
I saw the failed tests, but I don't know what they failed on, doesn't seem to be related (PinValueReadAndWrite).

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -19,20 +19,20 @@ public class SoftwareSpi : SpiDevice
...
 
         /// <summary>
         /// Software implementation of the SPI.
         /// </summary>
         /// <param name="clk">Clock pin.</param>
-        /// <param name="miso">Master Input Slave Output pin.</param>
+        /// <param name="miso">Master Input Slave Output pin. Optional, set to -1 to ignore</param>
         /// <param name="mosi">Master Output Slave Input pin.</param>
-        /// <param name="cs">Chip select pin (or negated chip select).</param>
+        /// <param name="cs">Chip select pin (or negated chip select). Optional, set to -1 to ignore. Note that there is a ChipSelectLine in the SPIConnectionSettings as well, either will be used.</param>
```
krwq, line 31: last sentence should go into remarks section - it's too long for parameter name

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -19,20 +19,20 @@ public class SoftwareSpi : SpiDevice
...
         /// <param name="settings">Settings of the SPI connection.</param>
         /// <param name="controller">GPIO controller used for pins.</param>
         /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
-        public SoftwareSpi(int clk, int miso, int mosi, int cs, SpiConnectionSettings settings = null, GpioController controller = null, bool shouldDispose = true)
+        public SoftwareSpi(int clk, int miso, int mosi, int cs = -1, SpiConnectionSettings settings = null, GpioController controller = null, bool shouldDispose = true)
```
krwq, line 35: bummer that miso is before mosi, we cannot change this right now especially they have the same type

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -42,20 +42,44 @@ public SoftwareSpi(int clk, int miso, int mosi, int cs, SpiConnectionSettings se
             _clk = clk;
             _miso = miso;
             _mosi = mosi;
-            _cs = cs;
+
+            if (_cs > -1 && _settings.ChipSelectLine > -1 && _cs != _settings.ChipSelectLine)
```
krwq, line 46: use `!=` or `==` instead of `>` here and everywhere

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -96,15 +120,18 @@ public SoftwareSpi(int clk, int miso, int mosi, int cs, SpiConnectionSettings se
...
-                    _controller.Write(_cs, _settings.ChipSelectLineActiveState);
-                },
-                exit: () =>
-                {
-                    _controller.Write(_cs, !(bool)_settings.ChipSelectLineActiveState);
-                });
+            if (_cs > -1)
+            {
+                _chipSelect = new ScopeData(
+                    enter: () =>
```
krwq, line 126: can we pass in no-op instead, otherwise we have to check for null on every data transfer

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -113,7 +140,7 @@ public SoftwareSpi(int clk, int miso, int mosi, int cs, SpiConnectionSettings se
         /// <inheritdoc />
         public override void TransferFullDuplex(ReadOnlySpan<byte> dataToWrite, Span<byte> dataToRead)
         {
-            if (dataToRead.Length != dataToRead.Length)
+            if (dataToWrite.Length != dataToRead.Length)
```
krwq, line 143: good catch

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -160,8 +219,19 @@ public override void Read(Span<byte> data)
         /// <inheritdoc />
         public override void Write(ReadOnlySpan<byte> data)
         {
-            Span<byte> dataToRead = stackalloc byte[data.Length];
-            TransferFullDuplex(data, dataToRead);
+            if (_miso == -1)
```
krwq, line 222: Since you already have TransferWriteOnly I'd just always use it

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -144,12 +171,44 @@ public override void TransferFullDuplex(ReadOnlySpan<byte> dataToWrite, Span<byt
...
         }
 
+        private void WriteBit(bool bitToWrite)
+        {
+            _controller.Write(_mosi, bitToWrite);
+        }
+
         /// <inheritdoc />
         public override void Read(Span<byte> data)
         {
```
krwq, line 214: Read should always throw when MISO is -1

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -210,18 +280,12 @@ public ScopeData(Action enter = null, Action exit = null)
 
             public void Enter()
             {
-                if (_enter != null)
-                {
-                    _enter();
-                }
+                _enter?.Invoke();
```
krwq, line 283: please remove the null checks in here - we will be paying relatively high cost for that where this is only needed for chip select and we can pass no op for that case

src/devices/SoftwareSpi/SoftwareSpi.cs
HakanL: Yea I know, I was thinking of adding an overloaded constructor, but figured it would be more confusing as the order would change with the same parameter types.

src/devices/SoftwareSpi/SoftwareSpi.cs
HakanL: I didn't want to change too much, but I guess the devices won't know if the data is read or not. Let me change it.

src/devices/SoftwareSpi/SoftwareSpi.cs
HakanL: I'm moving it up the call path instead so it doesn't have to checked for each bit.

src/devices/SoftwareSpi/SoftwareSpi.cs
HakanL: The code was like this, I just changed it to ? instead of `if`. But I agree, I will remove the check completely.

HakanL:
Fixes from code review pushed.

src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
```diff
@@ -19,8 +19,8 @@ private SpiConnectionSettings()
         /// Initializes a new instance of the <see cref="SpiConnectionSettings"/> class.
         /// </summary>
         /// <param name="busId">The bus ID the device is connected to.</param>
-        /// <param name="chipSelectLine">The chip select line used on the bus.</param>
-        public SpiConnectionSettings(int busId, int chipSelectLine)
+        /// <param name="chipSelectLine">The chip select line used on the bus. Optional, -1 if not used</param>
+        public SpiConnectionSettings(int busId, int chipSelectLine = -1)
```
krwq, line 23: if you make it optional here you should at least check if there is a way to support that on Linux (I think there might be), otherwise please leave as is and file a separate issue (might be a good idea anyway since that PR might need more work)

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -89,22 +114,30 @@ public SoftwareSpi(int clk, int miso, int mosi, int cs, SpiConnectionSettings se
             else
             {
                 _bitTransfer = new ScopeData(
+                    enter: () => { },
```
krwq, line 117: why didn't this need null check before (if it did have it please remove the check)

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -89,22 +114,30 @@ public SoftwareSpi(int clk, int miso, int mosi, int cs, SpiConnectionSettings se
...
+                        _controller.Write(_cs, _settings.ChipSelectLineActiveState);
+                    },
+                    exit: () =>
+                    {
+                        _controller.Write(_cs, !(bool)_settings.ChipSelectLineActiveState);
+                    });
+            }
+            else
+            {
+                _chipSelect = new ScopeData(() => { }, () => { });
```
krwq, line 139: might be worth to actually make constructor default to null and replace null with nop internally, that won't pollute the code and still get rid of null checks

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -144,24 +177,57 @@ public override void TransferFullDuplex(ReadOnlySpan<byte> dataToWrite, Span<byt
...
             Span<byte> dataToWrite = stackalloc byte[data.Length];
             TransferFullDuplex(dataToWrite, data);
         }
 
         /// <inheritdoc />
         public override void Write(ReadOnlySpan<byte> data)
         {
-            Span<byte> dataToRead = stackalloc byte[data.Length];
-            TransferFullDuplex(data, dataToRead);
+            // Output/Write only
```
krwq, line 229: Looking at the function name I don't think it requires any comment (it's probably more confusing than lack of it)

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -202,26 +268,20 @@ private class ScopeData
             internal Action _enter;
             internal Action _exit;
 
-            public ScopeData(Action enter = null, Action exit = null)
+            public ScopeData(Action enter, Action exit)
             {
-                _enter = enter;
-                _exit = exit;
+                _enter = enter ?? throw new ArgumentNullException(nameof(enter));
```
krwq, line 273: consider `?? () => {}` instead - since lack of enter/exit action is equivalent to NOP

krwq: I'm still curious why we didn't get NullRef before since it was lacking one of the actions in one place

src/devices/SoftwareSpi/SoftwareSpi.cs
krwq: nvm, there was a null check

src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
HakanL: From what I can find, passing in 0 seems to mean ignore ChipSelect on both Linux and Windows. Do you recommend we use 0 as the ignore value in the SoftwareSpi class as well, and change the default to that?

src/devices/SoftwareSpi/SoftwareSpi.cs
HakanL: Sure I can change that, I was just thinking it would indicate an incorrect use, but it's fine, I've changed it.

src/devices/SoftwareSpi/SoftwareSpi.cs
HakanL: Yea I agree, changed.

src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
krwq: no, 0 is a valid value for PhysicalNumbering numbering. As an implementation detail we can pass 0 to Linux struct when user passes -1, For 0 I'd pass through and I'm not opinionated if we should throw or just let it work so IMO just let it work unless we got some other opinions (@joperezr?).

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -113,7 +145,7 @@ public SoftwareSpi(int clk, int miso, int mosi, int cs, SpiConnectionSettings se
         /// <inheritdoc />
         public override void TransferFullDuplex(ReadOnlySpan<byte> dataToWrite, Span<byte> dataToRead)
         {
-            if (dataToRead.Length != dataToRead.Length)
+            if (dataToWrite.Length != dataToRead.Length)
```
krwq, line 148: TransferFullDuplex should also check if we are able to write and throw otherwise

HakanL: Yes, you're right, it's a public method. I've moved the check to `TransferFullDuplex` instead.

src/System.Device.Gpio/System/Device/Spi/SpiConnectionSettings.cs
HakanL: ok, I've added an implementation detail to both UnixSpiDevice and Win10SpiDevice to check for -1 and pass 0 to the OS.

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -160,8 +225,7 @@ public override void Read(Span<byte> data)
         /// <inheritdoc />
         public override void Write(ReadOnlySpan<byte> data)
         {
-            Span<byte> dataToRead = stackalloc byte[data.Length];
-            TransferFullDuplex(data, dataToRead);
+            TransferWriteOnly(data);
```
krwq, line 228: missing miso check here (similarly as you added for TransferFullDuplex)

HakanL: Miso isn't used for write-only though.

krwq: Sorry, still thinking of LED strip, add a check in Read then 😄

HakanL: :) Already in TransferFullDuplex that is called by Read, do you want a check in Read as well?

krwq:
rerunning CI to double check units branch didn't break anything here

krwq:
Thanks @HakanL!

PR #1077, opened 2020-05-15
Added support for the TLC-59711 LED PWM driver

HakanL:
Inspired by this work: https://github.com/adafruit/Adafruit_TLC59711/ (but not directly ported)


src/devices/Tlc59711/Tlc59711.cs
```diff
@@ -0,0 +1,155 @@
...
+        private byte _brightnessGreen;
+        private byte _brightnessBlue;
+        private byte[] _dataToWrite;
+
+        /// <summary>
+        /// Creates a new instance of the Tlc59711.
+        /// </summary>
+        public Tlc59711(byte numberOfDrivers, SpiDevice spiDevice)
+        {
+            _numberOfDrivers = numberOfDrivers;
```
krwq, line 35: - can this be somehow read by SPI? if so might be worth doing it and making it optional.
- numberOfDrivers should be an `int`
- optional: for some reason I feel like arguments should be swapped but I'm not certain if we have any official guidelines on that

src/devices/Tlc59711/Tlc59711.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// <param name="pwm">PWM level (0 = minimum, 65535 = maximum)</param>
+        public void SetPWM(int channel, ushort pwm)
+        {
+            if (channel > 12 * _numberOfDrivers)
+            {
+                throw new ArgumentOutOfRangeException("channel");
+            }
+
+            _pwmBuffer[channel] = pwm;
+
```
krwq, line 70: nit: extra enter

src/devices/Tlc59711/Tlc59711.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            _numberOfDrivers = numberOfDrivers;
+
+            _spiDevice = spiDevice ?? throw new ArgumentNullException(nameof(spiDevice));
+
+            // default 100% brigthness
+            _brightnessRed = _brightnessGreen = _brightnessBlue = 0x7f;
+
+            // 12 channels per driver
+            _pwmBuffer = new ushort[numberOfDrivers * 12];
```
krwq, line 43: I feel like you should change multiplication of common expressions into some kind of static function (i.e. NumberOfDrivesToNumberOfChannels) - `12` seems somewhat magic number - in here you got comment but elsewhere it's not there - it will be just easier to follow the code this way even if it's not shorter.

If it's used only in one place feel free to leave the magic constant with a comment since there is not much benefit of the function

src/devices/Tlc59711/Tlc59711.cs
HakanL: No it can't be read, you can daisy-chain as many devices as you like, but there's no reading anything from the device, the device is input-only (why I started down the path of output-only SPI device).
I can change the parameter to be int, just figured it could never be that large.
I can swap parameters and default drivers to 1 (most common)

src/devices/Tlc59711/Tlc59711.cs
HakanL: The number of channels (per driver) is constant, so I have moved it to a private const now.

HakanL:
Fixes pushed, please re-check.

src/devices/Tlc59711/Tlc59711.cs
```diff
@@ -0,0 +1,155 @@
...
+            _spiDevice?.Dispose();
+            _spiDevice = null;
+        }
+
+        /// <summary>
+        /// Call this to set the PWM level for a channel.
+        /// </summary>
+        /// <param name="channel">Channel, 0-based</param>
+        /// <param name="pwm">PWM level (0 = minimum, 65535 = maximum)</param>
+        public void SetPWM(int channel, ushort pwm)
```
krwq, line 65: I'd expect this work similarly to Pca9685, i.e.:
- this should allow creating PwmChannel
- global changing values is optional
- utilities for LED and similar I don't think they necessarily belong here (I'd rather have a separate PR with extension methods for all PwmChannels)

note the benefits of being able to construct PwmChannel:
- consistent APIs
- being able to pass it to other bindings which accept PwmChannel (i.e. ServoMotor)
- more in line with our on-going conversation of abstracting capabilities of the current board (i.e. board with TLC59711 with 1 driver would be equivalent of having 12 PwmChannels on board where having the hardware would be transparent for users)

krwq:
I'll check with our legal team the code can be based on BSD (I think yes but will double check).

src/devices/Tlc59711/Tlc59711.cs
HakanL: Well, to some degree, but the TLC59711 does not allow you to change the frequency, it will only allow you to control the DutyCycle. The TLC is a dedicated LED driver (which internally is using PWM).

HakanL:
> I'll check with our legal team the code can be based on BSD (I think yes but will double check).

BSD?

src/devices/Tlc59711/Tlc59711.cs
krwq: Would it make sense to just report the correct frequency but make setter a no-op? Most of the PWMs can be used to control LEDs. You can mention the limitation in the README.md

HakanL:
> I'll check with our legal team the code can be based on BSD (I think yes but will double check).

Ah, from the Adafruit repo that I used for inspiration on the methods, yes. BTW I used a similar library (same company, Adafruit) for the Mcp9808 device that I contributed, and that was approved (also BSD).

krwq:
https://github.com/adafruit/Adafruit_TLC59711/blob/master/license.txt - unoficially seem like dotnet/runtime already is using BSD code so we presumably can as well (we will need to update https://github.com/dotnet/iot/blob/master/THIRD-PARTY-NOTICES.TXT file, you can use similar file in dotnet/runtime as an example how to do that)

HakanL:
Ok, we can wait for that, but I just wanted to let you know that I just used that library to get an idea of what the API should look like, and some implementation details, it's not a port. It may not make a difference, but just FYI.

krwq:
@HakanL as long as it's inspired it's legally considered "derivative work" - as much as I hate legal issues we need to clear these first 😄

src/devices/Tlc59711/Tlc59711.cs
HakanL: I actually can't find the frequency this device is using, they have something called Enhanced Spectrum PWM, but they don't specify the frequency (maybe it's dynamic, I don't know). I don't disagree with you on using a standard API for (LED) control in the library, but I'm not sure it's a great fit with the PwmChannel IMO.

HakanL:
Got it, wasn't sure what the restrictions were, no problem. But please be aware that it's the same for the Mcp9808 device that I contributed, I derived that work from the Adafruit library as well.

krwq:
Thanks, we'll probably just need to list what's based on adafruit code to the best of our knowledge

richlander:
Ya, we can use BSD, as long as we add an entry to the TPN.

It occurs to me. We need more than one TPN. The root one should be dedicated to the GPIO library that's effectively the official deliverable of this repo.

Ideally, there would be a TPN in each binding directory, and we'd produce a merged one. If that's too much, then a single TPN at the root of the bindings directory is fine, with a link to the root of the directory to which it applies.

Works?

src/devices/Tlc59711/Tlc59711.cs
krwq: I think the frequency is f_osc is between 7-12MHz (per spec it's typically 10Mhz) divided by PWM steps which is 65536 which gives around 150Hz. Probably easiest to check with the oscilloscope. If you think this is different enough it's ok not to implement PwmChannel but I suspect someone will do that in the future anyway given that ServoMotor won't work otherwise 😄 Please at least make the API as similar as possible to Pca9685

krwq:
@HakanL do you think you'll have time to address remaining feedback in this PR?

src/devices/Tlc59711/Tlc59711.cs
HakanL: Most LED drivers use a much higher PWM frequency than 150Hz, so I don't think that number is correct. If it's not known (or it may be variable and controlled by the driver, ie their "Enhanced Spectrum PWM" I don't think it makes sense to add something in the class here that may not be correct. Unfortunately I have moved and don't have my electronics workbench set up currently, so I can't verify with a scope. There's also a risk that the frequency changes with the PWM settings. I don't think the PwmChannel is a good fit.

HakanL:
> @HakanL do you think you'll have time to address remaining feedback in this PR?

I added a comment, I don't know if there's anything more I can do about this PR?

krwq:
@HakanL what I meant is something in here implementing PwmChannel so that you can compose it with other things (i.e. ServoMotor)

HakanL:
> @HakanL what I meant is something in here implementing PwmChannel so that you can compose it with other things (i.e. ServoMotor)

This chip is strictly a LED driver, it's not designed to drive a servo motor (that requires specific timing in the PWM).

krwq:
@HakanL in that case I'd suggest to change the APIs so that they take `Color` and be more of shape like `SetRgbLedColor(int led, Color color)` but still, considering this is 16-bit I'd still be tempted to try it with servo motor

src/devices/Tlc59711/Tlc59711.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        /// <summary>
+        /// Call this to set the RGB value for a group of 3 channels
+        /// </summary>
+        /// <param name="ledNumber">LED number (channel number of the "red" pin divided by 3)</param>
+        /// <param name="red">Red level (0 = minimum, 65535 = maximum)</param>
+        /// <param name="green">Green level (0 = minimum, 65535 = maximum)</param>
+        /// <param name="blue">Blue level (0 = minimum, 65535 = maximum)</param>
+        public void SetLED(int ledNumber, ushort red, ushort green, ushort blue)
```
krwq, line 82: I'd recommend creating also API which takes Color.

src/devices/Tlc59711/Tlc59711.cs
```diff
@@ -0,0 +1,157 @@
...
+            _spiDevice?.Dispose();
+            _spiDevice = null;
+        }
+
+        /// <summary>
+        /// Call this to set the PWM level for a channel.
+        /// </summary>
+        /// <param name="channel">Channel, 0-based</param>
+        /// <param name="pwm">PWM level (0 = minimum, 65535 = maximum)</param>
+        public void SetPWM(int channel, ushort pwm)
```
krwq, line 65: Please remove this API if this is not suitable for regular PWM usage, alternative please create a way to create PwmChannel

src/devices/Tlc59711/Tlc59711.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        /// <summary>
+        /// Set the brightness of LED channels to specific value
+        /// </summary>
+        /// <param name="brightnessRed">Brightness between 0 and 127</param>
+        /// <param name="brightnessGreen">Brightness between 0 and 127</param>
+        /// <param name="brightnessBlue">Brightness between 0 and 127</param>
+        public void SetBrightness(byte brightnessRed, byte brightnessGreen, byte brightnessBlue)
+        {
+            _brightnessRed = Math.Min(brightnessRed, (byte)127);
```
krwq, line 152: this is confusing, in other place RGB channels are 16-bit but in here you limit them to less than 8

src/devices/Tlc59711/Tlc59711.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        /// <summary>
+        /// Set the brightness of all LED channels to same value
+        /// </summary>
+        /// <param name="brightness">Brightness between 0 and 127</param>
+        public void SetBrightness(byte brightness)
+        {
+            brightness = Math.Min(brightness, (byte)127);
+
+            _brightnessRed = _brightnessGreen = _brightnessBlue = brightness;
```
krwq, line 141: how is this working? This won't change anything on the device, you're just updating fields here...

src/devices/Tlc59711/Tlc59711.csproj
```diff
@@ -0,0 +1,16 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <LangVersion>preview</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="Tlc59711.cs" />
+    <None Include="README.md" />
+    <ProjectReference Include="..\SoftwareSpi\SoftwareSpi.csproj" />
```
krwq, line 12: why does this depend on SoftwareSpi?

src/devices/Tlc59711/Tlc59711.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// Call this after every change to write the new PWM levels to the device.
+        /// </summary>
+        public void Write()
+        {
+            int pos = 0;
+
+            uint command;
+
+            // Magic word for write
+            command = 0x25;
```
krwq, line 99: you will need to write down the comment with complete bit layout of this

src/devices/Tlc59711/Tlc59711.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            SetPWM((byte)(ledNumber * 3), red);
+            SetPWM((byte)(ledNumber * 3 + 1), green);
+            SetPWM((byte)(ledNumber * 3 + 2), blue);
+        }
+
+        /// <summary>
+        /// Call this after every change to write the new PWM levels to the device.
+        /// </summary>
+        public void Write()
```
krwq, line 92: this should be called automatically, user should not have to do this, you should rather provide API for changing all colors at the same time (i.e. with Span or image buffer - i.e. similarly as Wsb2812b code does it)

joperezr:
[Triage] @HakanL would you have some time to address the feedback we gave here? This seems abandoned, so we'll go ahead and close this one if we don't see activity by our next triage meeting.

HakanL:
> [Triage] @HakanL would you have some time to address the feedback we gave here? This seems abandoned, so we'll go ahead and close this one if we don't see activity by our next triage meeting.

There were a lot of comments added. I think it may be best if someone else creates a (sample) class for LED drivers, apparently I was way off in this PR.

joperezr:
[Triage] Sounds good, we'll go ahead and close it. Let us know if you are interested in contributing bindings 😄

PR #1084, opened 2020-05-22
Support PWM on Beagle Bone

### LINKED ISSUE #1083: Beagle Bone support for PWM
## Summary

The `System.Device.Gpio` library expects the pwm sysfs naming to conform to the mainline kernel's documentation.

https://github.com/dotnet/iot/blob/00f9da195e6c3bb296450ab88303e031ad55aff8/src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.cs#L41-L42

> If CONFIG_SYSFS is enabled in your kernel configuration a simple sysfs
interface is provided to use the PWMs from userspace. It is exposed at
/sys/class/pwm/. Each probed PWM controller/chip will be exported as
pwmchipN, where N is the base of the PWM chip.
> ...
> The PWM channels are numbered using a per-chip index from 0 to npwm-1.
>
> When a PWM channel is exported a pwmX directory will be created in the
pwmchipN directory it is associated with, where X is the number of the
channel that was exported.

https://www.kernel.org/doc/Documentation/pwm.txt

However, the Beagle Bone's fork of the kernel does not adhere to this convention.

https://github.com/beagleboard/linux/commit/0e09cd3599153a865e87e212ffed6d485488dd4f

Instead of `/sys/class/pwm/pwmchipY/pwmX`, it uses `/sys/class/pwm/pwmchipY/pwm-Y:X`, breaking the library.

## Proposed Solution

Expose `private readonly string _channelPath` as `protected virtual string ChannelName { get; }` and derive a `BeagleBonePwmChannel` that inherits from `UnixPwmChannel`, overriding `ChannelName` to provide a working pwm solution for the BeagleBone's version of the Linux kernel.

Something like this:

```cs
internal class BeagleBonePwmChannel : UnixPwmChannel
{
    protected override string ChannelName { get => $"{ChipPath}/pwm-{Chip}:{Channel}"; }
}
``` 

The version of the kernel can be detected by reading the `/proc/version` file which returns the following information.

```txt
Linux version 4.14-100-ti-r113 (root@x1-am57xx-beagle-x15-2gb) (gcc version 6.3.0.20170516 (Debian 6.3.0-18+deb9u1)) #1 SMP PREEMPT Wed July 31 00:01:10 UTC 2019
```

There a couple of things that might be sufficient to tell us we're running on the Beagle Bone kernel here, but I'm afraid they might be fragile.  

* We could look for `ti-`, but just because it's a Texas Instruments chip doesn't mean we're running beagle bone's forked version of the kernel. 
* We could also look for `beagle`, but we have no guarantee that the Beagle Bone organization won't rename it's build server

This would at least allow the `PwmChannel.Create()` method to continue it's automatic detection of an appropriate `PwmChannel` implementation and is similar, in spirit, to how the `GpioController` finds the "best driver" to use by reading `/proc/cpuinfo`.

https://github.com/dotnet/iot/blob/00f9da195e6c3bb296450ab88303e031ad55aff8/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs#L382-L384

rubberduck203:
Fixes #1083


src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
```diff
@@ -67,14 +67,27 @@ protected virtual void Dispose(bool disposing)
...
+            return new Channels.UnixPwmChannel(
+                chip,
+                channel,
+                frequency,
+                dutyCyclePercentage);
+        }
+
+        private static bool IsBeagleBoneKernel()
+        {
+            var kernelVersionInfo = System.IO.File.ReadAllText(@"/proc/version");
```
joperezr, line 89: NIT: can we add a try catch here in case /proc/version file can't be read for a permission issue and if so just return false?

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
```diff
@@ -67,14 +67,27 @@ protected virtual void Dispose(bool disposing)
                     frequency,
                     dutyCyclePercentage);
             }
-            else
+
+            if (IsBeagleBoneKernel())
```
joperezr, line 71: NIT: instead of two if blocks, can we do if()... else if ()... else?

It is just the convention we usually follow.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.cs
```diff
@@ -53,6 +51,16 @@ internal class UnixPwmChannel : PwmChannel
             SetFrequency(frequency, dutyCycle, currentDutyCycleNs);
         }
 
+        /// <summary>The PWM chip number.</summary>
+        protected int Chip { get; }
```
joperezr, line 55: Why do we need them to be properties? We could keep them regular members and just change their visibility from private to protected right?

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.cs
```diff
@@ -36,10 +34,10 @@ internal class UnixPwmChannel : PwmChannel
             int frequency = 400,
             double dutyCycle = 0.5)
         {
-            _chip = chip;
-            _channel = channel;
-            _chipPath = $"/sys/class/pwm/pwmchip{_chip}";
-            _channelPath = $"{_chipPath}/pwm{_channel}";
```
joperezr, line 42: Another way to do this is just to also take an extra parameter which takes the channel Path format right? that way you don't need the extra properties or change visibility of other members.

joperezr:
Left some minor comments but the changes look good overall, again I'm sorry for the delay on reviewing and thanks a lot for the contribution.

src/System.Device.Gpio/System/Device/Pwm/Channels/BeagleBonePwmChannel.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace System.Device.Pwm.Channels
```
krwq, line 1: copyright header missing

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
```diff
@@ -67,14 +67,27 @@ protected virtual void Dispose(bool disposing)
...
             {
-                return new Channels.UnixPwmChannel(
-                    chip,
-                    channel,
-                    frequency,
-                    dutyCyclePercentage);
+                return new Channels.BeagleBonePwmChannel(
+                chip,
```
krwq, line 74: please add +1 indent to args

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.cs
rubberduck203: Sure. I think that would work.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.cs
rubberduck203: Yeah. It is a way to do it. I thought about doing it that way actually.
Correct me if I'm wrong, but wouldn't that require a format string like `"pwm-{0}:{1}"`?
That felt less readable than `"pwm-{Chip}:{Channel}"` to me.

I can play with it.

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
rubberduck203: I wasn't sure if it would be appropriate to return false if it failed, but I'm happy to make this change if you feel like that's good behavior.

joperezr: what we want is to not regress existing scenarios and have this be kind of like "A nice addition when it's needed" kind of thing, so by surrounding this with a try catch it means that if for whatever reason it fails we always silently fall back to old behavior, but in normal cases it will just work and provide better experience.

joperezr: One other thing I forgot to mention, can you please switch the `var`'s for explicit types? as well as make the Contains to be InvariantCulture so that the comparison happens fast

rubberduck203: Sure. That makes sense.
I'll push an update up tomorrow morning.

All the `var`s? 
Is there a published style guide somewhere?
It would be nice to have some of these things in the [CONTRIBUTING.md](https://github.com/dotnet/iot/blob/master/Documentation/CONTRIBUTING.md) file.

joperezr: We have a link saying that we follow the same code styling than our other dotnet repos here: https://github.com/dotnet/iot/blob/master/Documentation/Coding-guidelines.md
If you go there and specifically check number 10, you'll see that:
> We only use var when it's obvious what the variable type is (e.g. var stream = new FileStream(...) not var stream = OpenStandardInput()).

rubberduck203: This is done.
I tried to document the trade offs being made with a comment.
Please let me know if it captured the intent well enough.

src/System.Device.Gpio/System/Device/Pwm/Channels/BeagleBonePwmChannel.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>The sysfs name of the PWM channel</summary>
+        /// <returns>
+        /// A string like "pwm-X:Y" where X is the Channel number and Y is the Chip number.
+        /// </returns>
+        /// <remarks>
+        /// The BeagleBone kernel uses a non-standard naming convention for PWM Channels.
+        /// https://github.com/beagleboard/linux/commit/0e09cd3599153a865e87e212ffed6d485488dd4f
+        /// </remarks>
+        protected override string ChannelName => $"pwm-{_chip}:{_channel}";
```
rubberduck203, line 40: Wait. I think I got this the wrong way around when I updated it.
The doc isn't matching the code.

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.cs
rubberduck203: I implemented this on another branch.
Personally, I'm not sure how I feel about it.
Your call. I'll bring it in if you think it's preferable.

https://github.com/dotnet/iot/compare/master...rubberduck203:beagle-bone-pwm-alt?expand=1

rubberduck203: @joperezr any feedback on this?

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
```diff
@@ -79,6 +87,23 @@ protected virtual void Dispose(bool disposing)
...
 
+        private static bool IsBeagleBoneKernel()
+        {
+            try
+            {
+                string kernelVersionInfo = IO.File.ReadAllText(@"/proc/version");
+                // The String.Contains(string, StringComparison) overload wasn't introduced until .Net Standard 2.1
+                return kernelVersionInfo.IndexOf("beagle", StringComparison.InvariantCulture) > 0;
+            }
+            catch (SystemException ex) when (ex is IO.IOException || ex is Security.SecurityException || ex is UnauthorizedAccessException)
```
joperezr, line 98: One more comment here, can we make t his catch all exceptions so that we alwasy just fall back to old behavior in case of anything going wrong with the new codepath?

src/System.Device.Gpio/System/Device/Pwm/Channels/UnixPwmChannel.cs
joperezr: I like the way you changed this now.

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
rubberduck203: Done @joperezr. I think that's addresses all of your feedback.
Thanks for your patience and working through this with me.

joperezr: Thank you for working through and for the great contribution!

PR #1088, opened 2020-05-26
Adding Instructions to donwload .NET 5 preview SDK

joperezr:
FYI: @StevenBoelens @krwq @pgrawehr

### DIFF
--- a/Documentation/CONTRIBUTING.md
+++ b/Documentation/CONTRIBUTING.md
@@ -5,7 +5,10 @@ How to Contribute
 
 ## Recommended Software
 1. (If building on Windows) - **[Visual Studio 2019](https://visualstudio.microsoft.com/vs/preview/)** (Community, Professional, Enterprise) with the latest updates. This is not required for building the repo, but it will help a lot if the intent is to change the code or add new bindings.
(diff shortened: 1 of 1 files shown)


Documentation/CONTRIBUTING.md
```diff
@@ -5,7 +5,10 @@ How to Contribute
 
 ## Recommended Software
 1. (If building on Windows) - **[Visual Studio 2019](https://visualstudio.microsoft.com/vs/preview/)** (Community, Professional, Enterprise) with the latest updates. This is not required for building the repo, but it will help a lot if the intent is to change the code or add new bindings.
-1. **[.NET Core SDK](https://dotnet.microsoft.com/download/dotnet-core/3.0)** To ensure the install worked, make sure you can call `dotnet` tool from a command prompt
+1. **.NET Core SDK** To ensure the install worked, make sure you can call `dotnet` tool from a command prompt. We now require .NET 5 preview versions of the SDKs because of our central infrastructure, you can find the installers per platform here:
```
krwq, line 8: are we planning to update this version on every release now? can we instead create some generic aka.ms link which will always point to the right version?

joperezr: not every release, every time that Arcade pushes a new version into consumers which happens more or less often. I wouldn't create a link for this as it would be one more thing to have to mantain, instead everytime that we get an update via Arcade I'll go ahead and update this doc on the same PR

PR #1090, opened 2020-05-28
Added LLC200D3SH binding

DazFahy:
Add LLC200D3SH Optomax digital liquid level switch device

device with datasheet:
https://cdn-shop.adafruit.com/product-files/3397/3397_datasheet_actual.pdf


src/devices/Llc200d3sh/DetectionType.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Iot.Device.Llc200d3sh
```
krwq, line 1: - copyright header missing
- namespace should be type of device and not the device name

src/devices/Llc200d3sh/Llc200d3sh.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
+        public Llc200d3sh(int pin, GpioController gpioController = null,
+                          PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical, bool shouldDispose = true)
+        {
+            _controller = gpioController != null ? gpioController : new GpioController(pinNumberingScheme);
+            _pin = pin;
+
+            _controller.OpenPin(_pin, PinMode.Input);
+
+            _shouldDispose = gpioController == null ? true : shouldDispose;
```
krwq, line 29: `shouldDispose || gpioController == null`

src/devices/Llc200d3sh/DetectionType.cs
krwq: considering this device works same as any other liquid switch I suggest to make a generic type like WaterLevelSwitch or similar with options to flip how `1` should be interpreted as

src/devices/Llc200d3sh/DetectionType.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Optomax digital liquid level switch detection types
+    /// </summary>
+    public enum DetectionType
```
krwq, line 6: this name suggests the device has more than one mode of detection and not that this represents detection result. I suggest to either rename reading method to something like `IsMaxWaterLevel` or change this to something like this to also accommodate for other similar sensors (perhaps we could have a way of creating specific sensor which would give you pre-configured class)

```csharp
enum WaterLevel
{
    WaterLevelMax,
    WaterLevelMin,
    WaterNotDetected,
}
```

src/devices/Llc200d3sh/Llc200d3sh.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
```
krwq, line 1: copyright header missing

src/devices/Llc200d3sh/Llc200d3sh.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
krwq, line 1: please copy other csproj and follow the pattern

src/devices/Llc200d3sh/Llc200d3sh.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <Compile Remove="samples\**" />
+    <None Include="README.md" />
+	  <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
+	  <!--<ProjectReference Include="..\Common\Iot\Device\Common\CommonHelpers.csproj" />-->
```
krwq, line 13: there shouldn't be any commented out lines (unless there is a comment explaining why it's there but most likely it should be removed anyway)

src/devices/Llc200d3sh/README.md
```diff
@@ -0,0 +1,11 @@
+# Optomax digital liquid level switch
+
+Optomax digital liquid level switch is a device that can detect the presence of liquid/water. GPIO can be used to communicate with the device.
```
krwq, line 3: I got other water switch sensors which I guess work identical - I discourage using specific device name here except for the case where you'd preconfigure the class, so i.e. you could have base class like `WaterLevelSensor` and derived class `Llc200d3sh` - I think it should still live somewhere in this folder considering how simple the implementation is

krwq: for the sample you can show it on any sensor so basically the diagram can stay as is but if you go with two classes then please show how to configure the class yourself and how to use the specialized version

src/devices/Llc200d3sh/samples/Program.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
```
krwq, line 1: copyright header

src/devices/Llc200d3sh/DetectionType.cs
krwq: also the other example is that I could technically pour other fluid in there and it will show it as Water which further confirms me we should perhaps not even use word "Liquid" instead of water

src/devices/Llc200d3sh/README.md
DazFahy: Can you please supply the names and data sheets of some of the other liquid switch sensors you know about so I can have a look at their details.

src/devices/Llc200d3sh/DetectionType.cs
DazFahy: I don't like using the name water in the sensors code as you point out it can detect liquid as well. I am also not sure about the enum containing min and max as it is a simple switch, therefore when water is detected do you use the min or max enum value.  It could also confuse it with sensors which can detect the amount of liquid present I.e. etape

I agree detectiontype is the wrong name

DazFahy: As you point out we could rename the reading method and remove the enum altogether.  Although I think the reading method name would be better as something like `IsLiquidPresent`

src/devices/Llc200d3sh/DetectionType.cs
DazFahy: Would you have the namespace as LiquidLevelSwitch or as LiquidLevel. If you used LiquidLevel then in theory you could also include sensors that measure the volume of liquid in mm such as etape. Or does this make the namespace too generic?

krwq: Both seem fine to me

src/devices/Llc200d3sh/DetectionType.cs
krwq: sounds good to me, I like the proposed name

src/devices/Llc200d3sh/README.md
krwq: @DazFahy - I haven't got any spec for them, they were super cheap since they were basically breaking the circuit when water wasn't there (or opposite) so I only connected them between `High` and input and pulled down the input so when circuit was broken it was producing `Low`.

I think it was this one: https://www.aliexpress.com/item/4000265498553.html?spm=a2g0o.productlist.0.0.7d8c22abhyyPaK&algo_pvid=414159f0-5dd1-4c99-ac0a-245d4683c933&algo_expid=414159f0-5dd1-4c99-ac0a-245d4683c933-0&btsid=0be3769015911337997453345e8e02&ws_ab_test=searchweb0_0,searchweb201602_,searchweb201603_ or equivalent

src/devices/LiquidLevelSwitch/LiquidLevelSwitchBase.cs
```diff
@@ -0,0 +1,59 @@
...
+        private readonly int _pin;
+        private readonly PinMode _switchMode;
+        private readonly bool _shouldDispose;
+
+        private GpioController _controller;
+
+        /// <summary>Initializes a new instance of the <see cref="LiquidLevelSwitchBase" /> class.</summary>
+        /// <param name="pin">The pin.</param>
+        /// <param name="gpioController">The gpio controller.</param>
+        /// <param name="switchMode">The switch mode.</param>
```
krwq, line 24: this is not obvious how it works, perhaps something like `liquidPresentPinState`

src/devices/LiquidLevelSwitch/LiquidLevelSwitchBase.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <param name="pinNumberingScheme">The pin numbering scheme.</param>
+        /// <param name="shouldDispose">if set to <c>true</c> [should dispose].</param>
```
krwq, line 26: Seems like we use "True to dispose the Gpio Controller" everywhere else, I'd just follow the pattern

krwq:
```suggestion
        /// <param name="shouldDispose">True to dispose the Gpio Controller.</param>
```

src/devices/LiquidLevelSwitch/LiquidLevelSwitchBase.cs
```diff
@@ -0,0 +1,59 @@
...
+        public LiquidLevelSwitchBase(int pin, PinMode switchMode, GpioController gpioController = null,
+                                     PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical, bool shouldDispose = true)
+        {
+            _controller = gpioController != null ? gpioController : new GpioController(pinNumberingScheme);
```
krwq, line 30:
```suggestion
            _controller = gpioController ?? new GpioController(pinNumberingScheme);
```

src/devices/LiquidLevelSwitch/LiquidLevelSwitchBase.cs
```diff
@@ -0,0 +1,59 @@
...
+            _pin = pin;
+            _switchMode = switchMode;
+
+            _controller.OpenPin(_pin, System.Device.Gpio.PinMode.Input);
```
krwq, line 34: `using System.Device.Gpio` and shorten fully qualified name

src/devices/LiquidLevelSwitch/LiquidLevelSwitchBase.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            _shouldDispose = shouldDispose || gpioController == null;
+        }
+
+        /// <summary>Determines whether [is liquid present].</summary>
```
krwq, line 39: I don't think we use square brackets in the xml docs anywhere, please rephrase

src/devices/LiquidLevelSwitch/LiquidLevelSwitchBase.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <returns>
+        ///   <c>true</c> if [is liquid present]; otherwise, <c>false</c>.</returns>
+        public bool IsLiquidPresent()
+        {
+            var value = _controller.Read(_pin);
+
+            return value == (_switchMode == PinMode.Low ? PinValue.Low : PinValue.High);
```
krwq, line 46: if you rename the arg name then this could be just:
`return _controller.Read(_pin) == _liquidPresentPinState;`

src/devices/LiquidLevelSwitch/PinMode.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.LiquidLevelSwitch
+{
+    /// <summary>
+    /// Liquid level liquid detection pin mode
+    /// </summary>
+    public enum PinMode
```
krwq, line 10: I thought you were using System.Device.Gpio.PinMode... I'd remove this enum and just replace it with `[System.Device.Gpio.]PinMode liquidPresentPinState` or similar name

src/devices/LiquidLevelSwitch/LiquidLevelSwitchBase.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System;
+using System.Device.Gpio;
+
+namespace Iot.Device.LiquidLevelSwitch
+{
+    /// <summary>
+    /// Represents base class for Liquid Level Switches
+    /// </summary>
+    public abstract class LiquidLevelSwitchBase : IDisposable
```
krwq, line 13: I think `Base` suffix is not needed here since name is already generic enough without it

src/devices/LiquidLevelSwitch/README.md
```diff
@@ -0,0 +1,17 @@
+# Digital liquid level switch
+
+Digital liquid level switches are devices that can detect the presence of liquid/water. GPIO can be used to communicate with the devices.
+
+## Devices
+
+The implementation supports the following devices:
+
+- LLC200D3SH sensor ([Datasheet](https://cdn-shop.adafruit.com/product-files/3397/3397_datasheet_actual.pdf))
```
krwq, line 9: if you remove `abstract` then you can mention that any switch with single pin output will do as long as it's configured

src/devices/LiquidLevelSwitch/LiquidLevelSwitchBase.cs
DazFahy: Should I remove the specific implementation Llc200d3sh and change the sample to use the LiquidLevelSwitch?

DazFahy: Should I remove the specific implementation Llc200d3sh and change the sample to use the LiquidLevelSwitch?

krwq: No preference on removing but regardless I think LiquidLevelSwitch should be showcased in the sample (so both or just generic implementation)

src/devices/LiquidLevel/LiquidLevelSwitch.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        private GpioController _controller;
+
+        /// <summary>Creates a new instance of the LiquidLevelSwitch.</summary>
+        /// <param name="dataPin">The data pin</param>
+        /// <param name="liquidPresentPinState">Whether the pin value used to detect liquid is low or high</param>
+        /// <param name="pinNumberingScheme">Use the logical or physical pin layout</param>
+        /// <param name="gpioController">A Gpio Controller if you want to use a specific one</param>
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
+        public LiquidLevelSwitch(int dataPin, PinValue liquidPresentPinState, GpioController gpioController = null,
```
krwq, line 27: generally for formatting args you either move all of them on new lines or none of them, i.e. here I'd do (aligned to current indent level):

```csharp
public LiquidLevelSwitch(
    int dataPin,
    PinValue liquidPresentPinState,
    GpioController gpioController = null,
    PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical,
    bool shouldDispose = true)
{
...
```

src/devices/LiquidLevel/LiquidLevelSwitch.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        private readonly int _dataPin;
+        private readonly PinValue _liquidPresentPinState;
+        private readonly bool _shouldDispose;
+
+        private GpioController _controller;
+
+        /// <summary>Creates a new instance of the LiquidLevelSwitch.</summary>
+        /// <param name="dataPin">The data pin</param>
+        /// <param name="liquidPresentPinState">Whether the pin value used to detect liquid is low or high</param>
```
krwq, line 23:
```suggestion
        /// <param name="liquidPresentPinState">Data pin state representing liquid being present</param>
```

DazFahy: Changes made, however 2 of the checks failed, looks like an issue with the builds pulling some .net 5 preview nuget packages.

krwq:
LGTM, I suspect retrying should fix the issue (happening right now)

krwq:
Thanks @DazFahy!

PR #1098, opened 2020-06-05
Added support for APA102

HumJ0218:
I'm not sure if it can be named as `AddressableLed.Apa102` or just `Apa102`

There are many models compatible with APA102. 

And it's seems like APA102 is no longer in production?

### DIFF
--- a/src/devices/Apa102/Apa102.cs
+++ b/src/devices/Apa102/Apa102.cs
@@ -1,0 +1,80 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Device.Spi;
+using System.Drawing;
+
+namespace Iot.Device.Apa102
+{
+    /// <summary>
+    /// Driver for APA102. A double line transmission integrated control LED
+    /// </summary>
+    public class Apa102 : IDisposable, IEnumerable<Color>
+    {
+        /// <summary>
+        /// Colors of LEDs
+        /// </summary>
+        public Span<Color> Pixels => _pixels;
+
+        private readonly SpiDevice _spi;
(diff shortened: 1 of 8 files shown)


krwq:
I recently got APA102 couple months ago so I guess they still make them, do you know how fast you can change colors per second? They apparently offer much higher refresh rate than other LED strips

src/devices/Apa102/Apa102.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// Update color data to LEDs
+        /// </summary>
+        public void Flush()
+        {
+            for (var i = 0; i < _pixels.Length; i++)
+            {
+                var pixel = _buffer.AsSpan((i + 1) * 4);
+                pixel[0] = (byte)((_pixels[i].A >> 3) | 0b11100000); // global brighrness (alpha)
```
krwq, line 51:
```suggestion
                pixel[0] = (byte)((_pixels[i].A >> 3) | 0b11100000); // global brightness (alpha)
```

src/devices/Apa102/Apa102.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Collections.Generic;
+using System.Device.Spi;
+using System.Drawing;
+
+namespace Iot.Device.Apa102
+{
+    /// <summary>
+    /// Driver for APA102. A double line transmission integrated control LED
+    /// </summary>
+    public class Apa102 : IDisposable, IEnumerable<Color>
```
krwq, line 16: I'd remove the IEnumerable, you already expose Span

src/devices/Apa102/Apa102.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
krwq, line 1: please follow other projects, i.e.: https://github.com/dotnet/iot/blob/master/src/devices/Ak8963/Ak8963.csproj

PR #1100, opened 2020-06-09
Use of units, Part 1

pgrawehr:
Change of types to units for several bindings.

Note: String formatting (in samples) needs to be revisited later, but there are some corresponding PRs open in UnitsNet that will result in breaking changes.


src/devices/Bh1750fvi/Bh1750fvi.cs
```diff
@@ -89,7 +90,7 @@ private void SetLightTransmittance(double transmittance)
         /// Get BH1750FVI Illuminance
         /// </summary>
         /// <returns>Illuminance (Lux)</returns>
```
krwq, line 92: consider changing description since it's generic illuminance now

src/devices/Bmxx80/Bme280.cs
```diff
@@ -121,7 +122,7 @@ protected override void SetDefaultConfiguration()
         /// </summary>
         /// <param name="adcHumidity">The humidity value read from the device.</param>
         /// <returns>The percentage relative humidity.</returns>
```
krwq, line 124: description

src/devices/Bmxx80/Bme680.cs
```diff
@@ -421,11 +421,11 @@ public override bool TryReadTemperature(out Temperature temperature)
         /// the measurement was valid. Contains <see cref="double.NaN"/> otherwise.
         /// </param>
         /// <returns><code>true</code> if measurement was not skipped, otherwise <code>false</code>.</returns>
```
krwq, line 423: description

src/devices/Bmxx80/Bme680HeaterProfileConfig.cs
```diff
@@ -33,7 +34,7 @@ public class Bme680HeaterProfileConfig
         /// <param name="heaterResistance">The heater resistance in Ohm.</param>
         /// <param name="heaterDuration">The heating duration in ms.</param>
```
krwq, line 35: description

krwq: I've likely missed some cases, please re-view all descriptions in this PR to double check they don't contain explicit units

src/devices/Bh1750fvi/Bh1750fvi.cs
```diff
@@ -89,7 +89,7 @@ private void SetLightTransmittance(double transmittance)
         /// <summary>
         /// Get BH1750FVI Illuminance
         /// </summary>
-        /// <returns>Illuminance (Lux)</returns>
+        /// <returns>Illuminance (Default unit: Lux)</returns>
```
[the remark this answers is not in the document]
pgrawehr: Can do, but maybe the unit is better known than the quantity, so I thought to leave the name of the unit as reference.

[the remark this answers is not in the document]
krwq: IMO this suggests that unit is Lux but it is a generic representation of illuminance - I personally find the part in the brackets confusing even though I know what you mean..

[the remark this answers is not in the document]
krwq: either way we can fix this later if people complain, let's just keep going with the conversion, in the end it's just a comment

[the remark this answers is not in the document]
pgrawehr: I'll continue on this later, but I need to be sorting all my open branches first...

Issue #1085, opened 2020-05-25
Is there a need to implement 74595 drivers?


HumJ0218:
In some devices there are not enough GPIO lines

74595s can be a output-only GPIO driver:

SPI > 595 - 595 - ...
GPIO > 0 1 2 3 4 5 6 7 8 9 10 11 ...

So we can use SPI to drive devices with parallel data lines (ex. HD44780 the LCD screen)

pgrawehr:
Please specify the exact type of chip you mean and ideally provide a link to the data sheet. 

We do have support for the MCP23xxx GPIO expanders, they should do what you want. They have 8 or 16 configurable input/output pins. The MCP23S17 is for SPI, the MCP32017 for I2C.

PR #1104, opened 2020-06-16
Add binding for SN74hc595 shift register

richlander:
* Adds binding and test app/sample (tested on Debian ARM64 with .NET 5.0 Preview 5).
* Sample tests using one or two shift registers (both were tested).
* Uses scenario namespace (`Iot.Device.ShiftRegister`) instead of product namespace -- raising that as a conversation. I think it may help discoverability.
* Models `PinMapping` struct as a nested class, in large part due to the namespace choice.
* Includes `GpioControllerExtensions` extension methods that aided writing this binding. Currently not namespaced.

Some of these choices differ from what I've seen for other bindings. I am happy to make changes. I wanted to try some different ideas that felt natural to me as I was writing this binding.


src/devices/Sn74hc595/GpioControllerExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+    public static void OpenPinAndWrite(this GpioController controller, int pin, PinValue value)
+    {
+        controller.OpenPin(pin, PinMode.Output);
+        controller.Write(pin, value);
+    }
+
+    /// <summary>
+    /// Writes a value to multiple pins.
+    /// </summary>
+    public static void WriteValueToPins(this GpioController controller, int value, params int[] pins)
```
HumJ0218, line 34: duplicate  `GpioController.Write(ReadOnlySpan<PinValuePair> pinValuePairs)` ?

src/devices/Sn74hc595/GpioControllerExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        for (int i = 0; i < pins.Length; i++)
+        {
+            controller.Write(pins[i], value);
+        }
+    }
+
+    /// <summary>
+    /// Writers multiple values to the same pin.
+    /// </summary>
+    public static void WriteValuesToPin(this GpioController controller, int pin, params int[] values)
```
HumJ0218, line 45: What is the purpose of this?

HumJ0218:
The other day I submited a similar Issue #1085 , so I'm particularly concerned about this PR. You control 74595 via GPIO pins and I wanted a SPI interface. 

By using GPIO pins, users can shift bits one by one to show some amazing effects, but it might be much slower than SPI when user want to use it as a serial-parallel converter(it's what I use the most)

IMO you can make them both at the same time, just for more useful

richlander:
@HumJ0218 -- Thanks for the feedback. The [data sheet](https://www.ti.com/lit/ds/symlink/sn74hc595.pdf) I was using makes no mention of SPI. I did find this [post](https://playingwithatmega8.blogspot.com/2016/06/spi-port-extension.html), which seems to talk to your use case. Is that a good representation? If not, can you point me to a better one?

My scenario is the GPIO one (lighting LEDs). Here's an earlier version of a demo: https://www.youtube.com/watch?v=PvEqyu42ZDA&t=6s. I've since massively cleaned up the wiring.

It seems like there is room for multiple bindings of the same device if they enable different scenarios. I see you closed your issue. Why was that?

Curious to hear what you think.

src/devices/Sn74hc595/GpioControllerExtensions.cs
richlander: I assume you mean this: https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs#L335

High-level, yes, it is a duplicate. Practically, it isn't. The referenced API has a lot of ceremony and duplication for writing a single value to multiple pins. If you write code with both APIs, you will immediately see how different they are.

Anyway, my top-level question is what to do with these extension methods, in general.

src/devices/Sn74hc595/GpioControllerExtensions.cs
richlander: This API is the opposite use case of the other extension you called out. The first one writes the same value to multiple pins. This one writes multiples values to a single pin. In this case, this how I manage the leading and falling edge of the pin (which is required by the protocol).

Use: https://github.com/richlander/iot/blob/Sn74hc595/src/devices/Sn74hc595/Sn74hc595.cs#L84

joperezr:
> using System;

NIT: missing license headers. Same applies to the rest of the *.cs files.

---
Refers to: src/devices/Sn74hc595/GpioControllerExtensions.cs:1 in 4c8f139. [](commit_id = 4c8f139a174cc7da2163537c0551f3061b847dc4, deletion_comment = False)

src/devices/Sn74hc595/GpioControllerExtensions.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Device.Gpio;
+using System.Linq;
+
+/// <summary>
+/// Extension methods for GpioController
+/// </summary>
+public static class GpioControllerExtensions
```
joperezr, line 9: No namespace here? Also not sure if GpioController extensions belong inside a binding, in my mind this file should instead live under src\devices\Common

src/devices/Sn74hc595/GpioControllerExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    /// <summary>
+    /// Opens a set of pins.
+    /// </summary>
+    public static void OpenPins(this GpioController controller, PinMode mode, params int[] pins)
```
joperezr, line 14: I would be interested in your opinion of why do you think these extensions were required, to see if we have a gap in our API. Specifically abtout your OpenPinAndWrite, I wonder if our Open methods should have an optional arg Initial value

src/devices/Sn74hc595/GpioControllerExtensions.cs
joperezr: > Put another way, the product already has higher-level helpers, and I find them too cumbersome to actually use in practice.

That is good feedback and we should probably take a deeper look there then.

It's fine if you want to have these helpers for your binding, but I would suggest a few things: a) have the helpers be internal as we don't want your binding to expose these if they are really just meant to be used internally, if you think they might be useful outside this binding, then let's move them to common and add a namespace. b) for this specific one, I would definitely call the Write that takes in the Span of PinValuePairs, the reason is that this specific method is optimized in some drivers, so if you instead use regular Write's in a loop, you are effectively loosing those perf gains that might be important for your binding.

src/devices/Sn74hc595/GpioControllerExtensions.cs
joperezr: I had the same question when I saw it at first but so with your answer I take that this is so that you get an event of some sorts by doing a 0 to 1 or 1 to 0?

src/devices/Sn74hc595/GpioControllerExtensions.cs
richlander: Ya, I knew that wasn't right. I just decided to wait to hear feedback on the file all-up.

src/devices/Sn74hc595/GpioControllerExtensions.cs
richlander: Excellent q. I agree that that API option/question you suggest/put on the table is a better idea. I think we should do that or nothing at all. The only challenge is that the initial value is only relevant for one PinMode type.

HumJ0218:
@richlander Although SPI-related content is not mentioned in the data sheet, the timing of the 74595 can be considered identical to SPI.

I made some modules, including 7-segment LED, 16-segment LED, 8-relays, etc. They are all by SPI interface to drive the 74595. Especially the segment LED, by using 74595 it can be static driven instead of scanning, and PWM signal connect to OE to adjusting the brightness. If you use GPIO, just like software SPI, speed and efficiency are comparable to hardware SPI.

src/devices/Sn74hc595/GpioControllerExtensions.cs
richlander: Yes,

richlander:
@HumJ0218 -- What do you suggest to do? Have both bindings? Just one?

richlander:
We should also decide the namespace I chose. See if that works.

richlander:
Got an SPI variant working, and integrated into the main API (it's offered as another constructor). I wrote the code elsewhere, so need to integrate it into this PR now. There were a fair number of changes to make it work well.

src/devices/Sn74hc595/Sn74hc595.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
joperezr, line 4: Do you actually require netcoreapp3.1? In the end this will ship in Iot.Device.bindings which currently targets netcoreapp2.1, so  I'm assuming you don't depend on 3.1 or build would be failing right now so could you move back to 2.1 here?

richlander: Good call. will fix.

src/devices/Sn74hc595/Sn74hc595.cs
```diff
@@ -0,0 +1,361 @@
...
+    public class Sn74hc595
+    {
+        // Spec: https://www.ti.com/lit/ds/symlink/sn74hc595.pdf
+        // Tutorial: https://www.youtube.com/watch?v=6fVbJbNPrEU
+        // Using with SPI:
+        // https://forum.arduino.cc/index.php?topic=571144.0
+        // http://www.cupidcontrols.com/2013/12/turn-on-the-spi-lights-spi-output-shift-registers-and-leds/
+        private GpioController _controller;
+        private SpiDevice _spiDevice;
+        private bool _shouldDispose;
```
joperezr, line 23: NIT: readonly? same for the pins data, srclk and rclk bellow.

src/devices/Sn74hc595/Sn74hc595.csproj
richlander: I left the sample app as 3.1. I think that will help usability. Thoughts?

joperezr: I'm fine with that, we only really cared about the other one because the project in that case isn't what is actually building the binding that we are shipping. Samples can target anything as long as they support netcoreapp2.1

joperezr:
NIT: add a sln file for the binding, we know have that as one of the parts from the device binding template. FWIW yours should look like:
```xml

Microsoft Visual Studio Solution File, Format Version 12.00
# Visual Studio 15
VisualStudioVersion = 15.0.26124.0
MinimumVisualStudioVersion = 15.0.26124.0
Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sn74hc595", "Sn74hc595.csproj", "{6577AF1A-A054-4C89-A667-390EB3243006}"
EndProject
Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "samples", "samples", "{B07FF1B4-9F77-4FB3-8257-2D93714B8362}"
EndProject
Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sn74hc595-test", "samples\Sn74hc595-test.csproj", "{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}"
EndProject
Global
	GlobalSection(SolutionConfigurationPlatforms) = preSolution
		Debug|Any CPU = Debug|Any CPU
		Debug|x64 = Debug|x64
		Debug|x86 = Debug|x86
		Release|Any CPU = Release|Any CPU
		Release|x64 = Release|x64
		Release|x86 = Release|x86
	EndGlobalSection
	GlobalSection(SolutionProperties) = preSolution
		HideSolutionNode = FALSE
	EndGlobalSection
	GlobalSection(ProjectConfigurationPlatforms) = postSolution
		{6577AF1A-A054-4C89-A667-390EB3243006}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Debug|Any CPU.Build.0 = Debug|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Debug|x64.ActiveCfg = Debug|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Debug|x64.Build.0 = Debug|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Debug|x86.ActiveCfg = Debug|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Debug|x86.Build.0 = Debug|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Release|Any CPU.ActiveCfg = Release|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Release|Any CPU.Build.0 = Release|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Release|x64.ActiveCfg = Release|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Release|x64.Build.0 = Release|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Release|x86.ActiveCfg = Release|Any CPU
		{6577AF1A-A054-4C89-A667-390EB3243006}.Release|x86.Build.0 = Release|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Debug|Any CPU.Build.0 = Debug|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Debug|x64.ActiveCfg = Debug|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Debug|x64.Build.0 = Debug|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Debug|x86.ActiveCfg = Debug|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Debug|x86.Build.0 = Debug|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Release|Any CPU.ActiveCfg = Release|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Release|Any CPU.Build.0 = Release|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Release|x64.ActiveCfg = Release|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Release|x64.Build.0 = Release|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Release|x86.ActiveCfg = Release|Any CPU
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC}.Release|x86.Build.0 = Release|Any CPU
	EndGlobalSection
	GlobalSection(NestedProjects) = preSolution
		{0A8644DD-A8D1-4DD3-92E1-6B0C327713CC} = {B07FF1B4-9F77-4FB3-8257-2D93714B8362}
	EndGlobalSection
EndGlobal

```

src/devices/Sn74hc595/Sn74hc595.cs
```diff
@@ -0,0 +1,361 @@
...
+        {
+            _spiDevice = spiDevice;
+        }
+
+        /// <summary>
+        /// Count of shift units.
+        /// The number of (daisy-chained) shift registers. Minimum is 1.
+        /// Not the count of registers on a single unit.
+        /// </summary>
+        public int Count => _count;
```
joperezr, line 86: Nit: Might have to do with the fact that I haven't worked with one of these before, but Count feels a bit vague to refer to number of devices daisy chained. Not sure of what could be a better name though, `DaisyChainLength` or something more descriptive?

src/devices/Sn74hc595/Sn74hc595.cs
```diff
@@ -0,0 +1,361 @@
...
+
+        /// <summary>
+        /// Reports if Sn74hc595 is controlled with SPI.
+        /// </summary>
+        public bool UsesSpi => _spiDevice is object;
+
+        /// <summary>
+        /// Reports if Sn74hc595 is controlled with GPIO.
+        /// </summary>
+        public bool UsesGpio => _controller is object;
```
joperezr, line 102: Nit: I'm not sure of what a good user scenario would be for these two properties, although I'm not against keeping them.

src/devices/Sn74hc595/Sn74hc595.cs
```diff
@@ -0,0 +1,361 @@
...
+            {
+                ShiftByte(0);
+            }
+        }
+
+        /// <summary>
+        /// Shifts single value to next register
+        /// Does not perform latch.
+        /// </summary>
+        public void Shift(PinValue value)
```
joperezr, line 134: do we expect folks to call this directoy or to always call ShiftByte instead?

src/devices/Sn74hc595/Sn74hc595.cs
```diff
@@ -0,0 +1,361 @@
...
+        {
+            if (_spiDevice is object)
+            {
+                _spiDevice.WriteByte(value);
+                return;
+            }
+
+            for (int i = 0; i < 8; i++)
+            {
+                var data = (128 >> i) & value;
```
joperezr, line 162: NIT: use explicit type instead of var

src/devices/Sn74hc595/Sn74hc595.cs
richlander: Exactly. I wasn't sure what to call that either. We should come up with a general term since it isn't specific to this chip. Maybe @terrajobst has an idea on that.

src/devices/Sn74hc595/Sn74hc595.cs
richlander: I added those because some of the methods throw in the SPI case.

See:
- https://github.com/dotnet/iot/pull/1104/files#diff-255452453cfa55f55ad064a57e6b07ccR36
- https://github.com/dotnet/iot/pull/1104/files#diff-9de118a60b0d902626974c85247fd348R136

This is the downside of integrating SPI into this API. I think its worth it, but w/o these properties, it is hard to know which APIs are safe to call.

src/devices/Sn74hc595/Sn74hc595.cs
richlander: "directory"? Both Shift and ShiftByte are separate use cases. This comment pretty much clarifies that: https://github.com/dotnet/iot/pull/1104#issuecomment-644595161

src/devices/Sn74hc595/Sn74hc595.cs
terrajobst: `DeviceChainCount`?

richlander: That’s headed in a good direction. Here are some options:

- ‘DeviceChainCount’
- ‘DeviceCount’
- ‘ChainedDeviceCount’

The only confusing thing is what the count is when not chained. It would be weird if the count jumped from 0 to 2. I like ‘DeviceCount’ best for that reason. Chaining seems to be more the mechanism that is used to increase the device count. Thoughts?

src/devices/Sn74hc595/Sn74hc595.cs
richlander: Will make a ShiftBit method. Add in the comment that it is only supported via GPIO.

src/devices/Sn74hc595/Sn74hc595.cs
richlander: Going to go with `DeviceCount`.

richlander:
I think is now ready to go. Thanks!

src/devices/Sn74hc595/Sn74hc595.cs
```diff
@@ -0,0 +1,354 @@
...
+        /// Cleanup.
+        /// Failing to dispose this class, especially when callbacks are active, may lead to undefined behavior.
+        /// </summary>
+        public void Dispose()
+        {
+            if (_shouldDispose)
+            {
+                _controller?.Dispose();
+                _controller = null;
+                _spiDevice?.Dispose();
```
joperezr, line 232: We usually always dispose spidevices given that in most cases they will only serve one single device whereas a gpiocontroller can be used outside. All that said, given we have the shouldDispose what you are doing is probably the right thing to do.

src/devices/Sn74hc595/Sn74hc595.cs
```diff
@@ -0,0 +1,354 @@
...
+        {
+            if (_spiDevice is object)
+            {
+                _spiDevice.WriteByte(value);
+                return;
+            }
+
+            for (int i = 0; i < 8; i++)
+            {
+                int data = (128 >> i) & value;
```
joperezr, line 166: Nit: instead of using 128, can you use `0b_1000_0000` in order to make it clear what is happening here?

richlander: Fixed

src/devices/Sn74hc595/Sn74hc595.cs
richlander: I could go either way. I took your feedback and moved the SPI logic out of the condition.

PR #1107, opened 2020-06-18
Add IWritablePinSegment

richlander:
I've been working on enabling [LED animations](https://github.com/richlander/iot/blob/samples-update/samples/led-bar-graph/AnimateLeds.cs) to work over a variety of pin multiplexing schemes. That's what this interface is for.

I'm not enamored by this particular interface name. It would be good is someone could come up with something better.

Reading and writing is very similar when using GPIO controller pins directly. As you move to more exotic multiplexing options, reading and writing require different approaches. Reading may be done via polling. Also, many scenarios don't require reading. So, I'm proposing to make this interface for writing only. If people want reading, I'd suggest we make another interface for reading. Implementations can implement one or both.

I also have an interface implementation for a shift register @ https://github.com/richlander/iot/blob/samples-update/samples/led-bar-graph/ShiftRegisterSegment.cs. It is waiting on #1104 to be merged. Otherwise, it would be part of this PR.

I updated my test program to use the shift register implementation. Creating the implementation and switching to it took about 10mins. That was a good proof case for the idea. The program, at least visually, does the same thing as the prior GPIO-based implementation. There are other implementations we could make, like charlieplexing.

I don't know if this repo/namespace location is the correct one. I just put it there to start. Looking for input on that. It's also an open question of whether the interface and its implementations should be in the same place. I don't think they should.

### DIFF
--- a/src/devices/Common/Iot/Device/Common/IWritablePinSegment.cs
+++ b/src/devices/Common/Iot/Device/Common/IWritablePinSegment.cs
@@ -1,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.Common
+{
+    /// <summary>
+    /// Interface that abstracts pin multiplexing.
+    /// </summary>
+    public interface IWritablePinSegment
+    {
+        /// <summary>
+        /// Length of segment (number of pins)
+        /// </summary>
+        int Length { get; }
+
+        /// <summary>
+        /// Writes a high or low value to pin
+        /// </summary>
+        void Write(int pin, int value);
+    }
+}
--- a/src/devices/Common/Iot/Device/Common/IWritablePinSegment.cs
+++ b/src/devices/Common/Iot/Device/Common/IWritablePinSegment.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Device.Gpio;
 
 namespace Iot.Device.Common
 {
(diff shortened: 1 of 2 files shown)


src/devices/Common/Iot/Device/Common/IWritablePinSegment.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// Length of segment (number of pins)
+        /// </summary>
+        int Length { get; }
+
+        /// <summary>
+        /// Writes a high or low value to pin
+        /// </summary>
+        void Write(int pin, int value);
```
krwq, line 22: I suggest to use PinValue instead of `int`. Also I think this abstraction is very similar to already existing GpioController - I suspect if you added something like MappingGpioController/MappingGpioDriver you could achieve similar effect

richlander: I thought of that, and it may be the right option. My reason for not choosing it is that `PinValue` seems very GPIO-oriented, which this is not.

richlander: I'm still thinking about this. Maybe you are right. Maybe `PinValue` can be considered generic. I can take a run at making that change.

richlander: This is the problem ... I have to cast `PinValue` to `int` here to enable that: https://github.com/richlander/iot/blob/Sn74hc595/src/devices/Sn74hc595/Sn74hc595.cs#L115

Thoughts?

richlander: I could switch the `Shift` method to `PinValue` and then leave [ShiftByte](https://github.com/richlander/iot/blob/Sn74hc595/src/devices/Sn74hc595/Sn74hc595.cs#L138) as `byte`. Thoughts?

richlander: Done.

richlander:
I am going to close this and instead extend #1148.

PR #1111, opened 2020-06-20
Added Sharp Memory LCD LSxxxB7DHxx

HumJ0218:
added 3-wire interface memory LCDs - LS013B7DH03, LS013B7DH05, LS027B7DH01


src/devices/MemoryLCD/Utility.cs
```diff
@@ -0,0 +1,276 @@
...
+
+using System.Collections.Generic;
+
+namespace Iot.Device.MemoryLcd
+{
+    internal static class Utility
+    {
+        public static byte GetAgByte(int b)
+        {
+            return s_byteTable[b];
```
HumJ0218, line 13: The AG byte here is reverse of row number byte, I'm not sure if it's better to use a byte dictionary or directly calculate it

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,208 @@
...
+using System.Device.Gpio;
+using System.Device.Spi;
+using System.Threading;
+
+namespace Iot.Device.MemoryLcd
+{
+    /// <summary>
+    /// Memory LCD
+    /// </summary>
+    public abstract class LSxxxB7DHxx
```
pgrawehr, line 15: This should implement IDisposable to close the open pins and the Spi device connection.

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,208 @@
...
+                    Thread.Sleep(s_powerup_extcomin_delay);
+                }
+            }
+        }
+
+        private void WriteSpi(ReadOnlySpan<byte> bytes)
+        {
+            if (_gpio != null && _scs > -1)
+            {
+                _gpio.Write(_scs, PinValue.High);
```
pgrawehr, line 196: IIRC Chip select lines are usually low active. Is it really correct to set it high here?

src/devices/MemoryLCD/samples/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            mlcd.AllClear();
+
+            Console.WriteLine("DataUpdate1Line. Press any key to continue...");
+            Console.ReadKey();
+
+            byte[] lineBuffer = new byte[mlcd.BytesPerLine];
+            for (int i = 0; i < mlcd.PixelHeight; i++)
+            {
+                random.NextBytes(lineBuffer);
```
pgrawehr, line 49: Would it be possible to provide (as a separate class or maybe as extension methods) a high-level interface to this, which allows writing text or bitmap instances directly? This low-level interface that doesn't even define the pixel format to use is not really nice to use.

src/devices/MemoryLCD/LSxxxB7DHxx.cs
HumJ0218: yes for sure, see datasheet section 6. all 3 model had been tested on my device

IMO such the timing may facilitate compatibility with other SPI devices during transmission

src/devices/MemoryLCD/samples/Program.cs
HumJ0218: yes, and low-level interfaces also need to be retained for more efficient control

src/devices/MemoryLCD/LS013B7DH03.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Create a memory LCD device
+        /// </summary>
+        /// <param name="spi">SPI controller</param>
+        /// <param name="gpio">GPIO controller</param>
+        /// <param name="scs">Chip select signal</param>
+        /// <param name="disp">Display ON/OFF signal</param>
+        /// <param name="extcomin">External COM inversion signal input</param>
+        public LS013B7DH03(SpiDevice spi, GpioController gpio = null, int scs = -1, int disp = -1, int extcomin = -1)
```
krwq, line 23: I'd suggest expanding names of pins (`chipSelect`, `display`, `externalCom`)

src/devices/MemoryLCD/LS027B7DH01.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Create a memory LCD device
+        /// </summary>
+        /// <param name="spi">SPI controller</param>
+        /// <param name="gpio">GPIO controller</param>
+        /// <param name="scs">Chip select signal</param>
+        /// <param name="disp">Display ON/OFF signal</param>
+        /// <param name="extcomin">External COM inversion signal input</param>
+        public LS027B7DH01(SpiDevice spi, GpioController gpio = null, int scs = -1, int disp = -1, int extcomin = -1)
```
krwq, line 23: same

HumJ0218:
@krwq Is there a standardized rule in our project as to whether the parameter name should be used using the pin name in datasheet(eg. `scs`) or the full name of the pin(eg. `chipSelectSignal`)?

krwq:
@HumJ0218 I do not think we've set the guidelines on that. I usually prefer long names always and never oppose that but if name is common abbreviation it is fine with me. `scs` specifically sounds weird to me as usually it's called just `cs` and I'm not sure what the front `s` stands for... (I'm fine with `cs` but `scs` is confusing to me). I'd say it's gut feeling guideline...

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// </summary>
+        public void DataUpdate1Line(byte lineIndex, Span<byte> data, bool frameInversion = false)
+        {
+            if (data.Length != BytesPerLine)
+            {
+                throw new Exception($"Data writing period must be {PixelWidth}ck ({BytesPerLine} bytes)");
+            }
+
+            // m(1), ag(1), d(18), dummy(2)
+            byte[] buffer = new byte[BytesPerLine + 4];
```
Ellerbach, line 106: Why not directly create a Span<byte> for buffer? I have hard time to understand what's the purpose of the `line` buffer. That should simplify a bit the lisibility.

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// </summary>
+        public void DataUpdateMultipleLines(Span<byte> lineIndex, Span<byte> data, bool frameInversion = false)
+        {
+            if (data.Length != lineIndex.Length * BytesPerLine)
+            {
+                throw new Exception($"Data writing period must be {lineIndex.Length * PixelWidth}ck ({lineIndex.Length * BytesPerLine} bytes)");
+            }
+
+            // m(1), ag1(1), d1(18), dummy2(1), ag2(1), d2(18), ... , dummyn(1), agn(1), dn(18), dummy(2)
+            byte[] buffer = new byte[2 + (2 + BytesPerLine) * lineIndex.Length];
```
Ellerbach, line 128: Same here, the `fromLine` and `line` are not really necessary

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,300 @@
...
+            WriteSpi(buffer);
+        }
+
+        /// <summary>
+        /// Maintains memory internal data (maintains current display).
+        /// </summary>
+        public void Display(bool frameInversion = false)
+        {
+            // m(1), dummy(1)
+            byte[] buffer = new byte[2];
```
Ellerbach, line 154: Span<byte> rather than buffer
```csharp
Span<byte> bufffer = stackalloc byte[2];
```

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,300 @@
...
+            WriteSpi(buffer);
+        }
+
+        /// <summary>
+        /// Clears memory internal data and writes white.
+        /// </summary>
+        public void AllClear(bool frameInversion = false)
+        {
+            // m(1), dummy(1)
+            byte[] buffer = new byte[2];
```
Ellerbach, line 167: same as already mentionned

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,300 @@
...
+
+        /// <summary>
+        /// External COM inversion signal input (H: enable)<br/>When EXTMODE is "Lo", connect the EXTCOMIN to VSS.
+        /// </summary>
+        private readonly int _extcomin;
+        #endregion
+
+        #region  Delay constants for LCD timing
+
+        private static readonly int s_powerup_disp_delay = 1; // (>30us)
```
Ellerbach, line 57: why not const if it's a readonly?
Also the naming convention should be sleepPowerupDisplayDelay or something equivalent on a readonly int. Or SleepPowerUpDisplayDelay if const is used. 
If const is used, then all constant should be placed at beginning of the class
This apply to all the next lines

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,300 @@
...
+        private static readonly int s_powerup_extcomin_delay = 1; // (>30us)
+        // private static readonly int s_scs_high_delay = 0; // (>3us)
```
Ellerbach, line 59: No dead code please, if it's not used, please remove!

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,300 @@
...
+        private readonly int _disp;
+
+        /// <summary>
+        /// External COM inversion signal input (H: enable)<br/>When EXTMODE is "Lo", connect the EXTCOMIN to VSS.
+        /// </summary>
+        private readonly int _extcomin;
+        #endregion
+
+        #region  Delay constants for LCD timing
+
```
Ellerbach, line 56: Please add a link on the documentation where you picked the timing.
Also, if you need µ seconds, you can if you're using the Delay Helper. Would be much better than the traditional Thread.Sleep

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,300 @@
...
+                if (_scs > -1)
+                {
+                    _gpio.OpenPin(_scs, PinMode.Output);
+                }
+
+                if (_disp > -1)
+                {
+                    _gpio.OpenPin(_disp, PinMode.Output);
+                    _gpio.Write(_disp, PinValue.High);
+                    Thread.Sleep(s_powerup_disp_delay);
```
Ellerbach, line 209: See comment at the beginning of this file about using DelayHelper. See here: https://github.com/dotnet/iot/blob/master/src/devices/Common/System/Device/DelayHelper.cs

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,300 @@
...
+                }
+
+                if (_extcomin > -1)
+                {
+                    _gpio.OpenPin(_extcomin, PinMode.Output);
+                    _gpio.Write(_extcomin, PinValue.Low);
+                    Thread.Sleep(s_powerup_extcomin_delay);
```
Ellerbach, line 216: +1, same as above. If any other place, please apply the right pattern for those delays

src/devices/MemoryLCD/Utility.cs
Ellerbach: I would not use a dictionary for a good reason: your binding can perfectly work on nanoFrmaework which doesn't support dictionaries :-) So just maybe do the math or just have a static table, that would do the trick as well. In terms of size, the math will be better, in terms of performance, the static table would be better. You don't even need a dictionary for that

Ellerbach, line 13: And please use the proper naming convention for the variable as well `sByteTable` or equivalent.

src/devices/MemoryLCD/LS013B7DH03.cs
Ellerbach, line 23: And it seems like what you are doing when you are passing the gpio and the pins is to emulate SPI, correct?
If so, then it's not needed to implement the gpio part, just use SoftwareSpi, that should do the same.

Ellerbach, line 23: And if you really need to use Gpiocontroller, please use the disposed pattern mentioned here: https://github.com/dotnet/iot/tree/master/src/devices#gpiocontroller

Ellerbach (commented):
Thanks for adding those LCD! Couple of comments and suggestions.

HumJ0218:
> @HumJ0218 I do not think we've set the guidelines on that. I usually prefer long names always and never oppose that but if name is common abbreviation it is fine with me. `scs` specifically sounds weird to me as usually it's called just `cs` and I'm not sure what the front `s` stands for... (I'm fine with `cs` but `scs` is confusing to me). I'd say it's gut feeling guideline...

I prefer to use the pin name defined in the datasheet as the parameter name `scs`

I guess the pin name `SCS` are different from most other devices, is because of these module are using different pin value(high level) when selected. For this reason(and also the datasheet), I think it might be more appropriate to use `scs` than `cs`.

joperezr:
[Triage] We would prefer to use `cs` so that it matches the rest of our API and in order to abstract the concept of the datasheet to consumers. We are completely fine if you want to add scs on the tripple slash comments so that it is also easy to find to people. On a side note, do you think you will have some time in the near future to address the pending comments here?

HumJ0218:
@joperezr I think it is necessary to have a guideline for parameter names so as not to have similar disputes in the future.

And, I'd love to be involved in the discussion

joperezr:
[Triage] We added the following issue to track the addition of the documentation: https://github.com/dotnet/iot/issues/1256

Do you still plan on addressing the rest of the feedback that was given here? Let us know if you are not interested so that we can close the PR in our next triage meeting.

HumJ0218:
@joperezr working on it, and new posts this weekend

joperezr:
awesome, keep in mind you will most likely have to rebase on top of master as we have recently added nullable annotations to all bindings, so this will likely generate warnings in your code that you'll have to fix.

src/devices/MemoryLCD/AgBytes.cs
```diff
@@ -0,0 +1,272 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
Ellerbach, line 3: It's only 2 lines now :-)
Please adjust everywhere

src/devices/MemoryLCD/Extends/LSxxxB7DHxxExtends.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
```
Ellerbach, line 1: You're missing the header here (only 2 lines now)

src/devices/MemoryLCD/LS013B7DH03.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Create a <see cref="LS013B7DH03"/> device
+        /// </summary>
+        /// <param name="spi">SPI controller</param>
+        /// <param name="gpio">GPIO controller</param>
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
+        /// <param name="scs">Chip select signal</param>
+        /// <param name="disp">Display ON/OFF signal</param>
+        /// <param name="extcomin">External COM inversion signal input</param>
+        public LS013B7DH03(SpiDevice spi, GpioController? gpio = null, bool shouldDispose = true, int scs = -1, int disp = -1, int extcomin = -1)
```
Ellerbach, line 24: I would prefer long and explicit names. I think @krwq already commented on this.

src/devices/MemoryLCD/samples/Program.cs
```diff
@@ -0,0 +1,131 @@
+﻿using System;
```
Ellerbach, line 1: You're missing the header here (only 2 lines now)

src/devices/MemoryLCD/MemoryLcd.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
joperezr, line 4: I'm curious, any reason why you went with 3.1 as opposed to 2.1? I ask because iot.device.bindings ships for 2.1 and 5.0, so that means that this binding would only be available for the 5.0 version so ideally we would want to be in both if possible.

src/devices/MemoryLCD/AgBytes.cs
```diff
@@ -0,0 +1,271 @@
...
+namespace Iot.Device.MemoryLcd
+{
+    /// <summary>
+    /// Gate Line Address Setup
+    /// </summary>
+    internal static class AgBytes
+    {
+        public static byte[] Table { get; } = new byte[]
+        {
+            0b00000000,
```
joperezr, line 13: NIT: You don't have to fix this if you don't want to, but it is usually easier to read these guys when using an underscore as separators, for example: `0b_0000_0100`

src/devices/MemoryLCD/Extends/LSxxxB7DHxxExtends.cs
```diff
@@ -0,0 +1,70 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Drawing;
+using Iot.Device.MemoryLcd;
+
+namespace MemoryLcd.Extends
```
joperezr, line 8: Can you change the namespace to put this under Iot.Device.MemoryLcd as well please?

src/devices/MemoryLCD/Extends/LSxxxB7DHxxExtends.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    /// <summary>
+    /// Extend methods for LSxxxB7DHxx
+    /// </summary>
+    public static class LSxxxB7DHxxExtends
```
joperezr, line 13: We usually name these types of classes Extensions instead of Extends, so in this case this should be named `LSxxxB7DHxxExtensions`

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,256 @@
...
+
+            outputBuffer[0] = (byte)(frameInversion ? ModeSelectionPeriodByte.FrameInversion : ModeSelectionPeriodByte.Dummy);
+
+            WriteSpi(outputBuffer);
+        }
+
+        /// <summary>
+        /// Clears memory internal data and writes white.
+        /// </summary>
+        public void AllClear(bool frameInversion = false)
```
joperezr, line 172: Can you also add in the summary what does frameInversion mean or when to use it? Also it seems like this method could probably just be called Clear or ClearAll instead of AllClear?

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,256 @@
...
+            BytesPerLine = (PixelWidth + 7) / 8;
+
+            _lineNumberBuffer = Enumerable.Range(0, PixelHeight).Select(m => (byte)m).ToArray();
+            _frameBuffer = new byte[BytesPerLine * PixelHeight];
+
+            Init();
+        }
+
+        /// <summary>
+        /// Updates data of only one specified line.
```
joperezr, line 109: Same comment here, looks like you are not having the parameters in the summary.

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,256 @@
...
+        /// </summary>
+        public void DataUpdate1Line(byte lineIndex, Span<byte> data, bool frameInversion = false)
```
joperezr, line 111:
```suggestion
        public void UpdateRow(int row, Span<byte> data, bool frameInversion = false)
```
Consider renaming method and parameters.

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,256 @@
...
+            outputBuffer[1] = AgBytes.Table[lineIndex];
+            data.CopyTo(outputBuffer.Slice(2, BytesPerLine));
+
+            WriteSpi(outputBuffer);
+        }
+
+        /// <summary>
+        /// Updates arbitrary multiple lines data.
+        /// </summary>
+        public void DataUpdateMultipleLines(Span<byte> lineIndex, Span<byte> data, bool frameInversion = false)
```
joperezr, line 131:
```suggestion
        public void UpdateMultipleRows(Span<int> rowIndex, Span<byte> data, bool frameInversion = false)
```
Similar to the other comment about updating the name.

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,256 @@
...
+            outputBuffer[0] = (byte)(ModeSelectionPeriodByte.Mode | (frameInversion ? ModeSelectionPeriodByte.FrameInversion : ModeSelectionPeriodByte.Dummy));
+
+            for (int i = 0; i < lineIndex.Length; i++)
+            {
+                int from = BytesPerLine * i;
+                int to = 1 + (2 + BytesPerLine) * i;
+
+                outputBuffer[to] = AgBytes.Table[lineIndex[i] + 1];
+
+                data.Slice(from, BytesPerLine).CopyTo(outputBuffer.Slice(to + 1, BytesPerLine));
```
joperezr, line 150: Any reason why this method is not simply calling the other method updating of the single row?

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,256 @@
...
+            }
+
+            WriteSpi(outputBuffer);
+        }
+
+        /// <summary>
+        /// Maintains memory internal data (maintains current display).
```
joperezr, line 157: Does this mean that the other UpdateRow methods won't really change the display until you call this method? If not I'm not sure I understand what this method is doing.

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,256 @@
...
+        /// </summary>
+        public void DataUpdateMultipleLines(Span<byte> lineIndex, Span<byte> data, bool frameInversion = false)
+        {
+            if (data.Length != lineIndex.Length * BytesPerLine)
+            {
+                throw new Exception($"Data writing period must be {lineIndex.Length * PixelWidth}ck ({lineIndex.Length * BytesPerLine} bytes)");
+            }
+
+            // m(1), ag1(1), d1(18), dummy2(1), ag2(1), d2(18), ... , dummyn(1), agn(1), dn(18), dummy(2)
+            Span<byte> outputBuffer = stackalloc byte[2 + (2 + BytesPerLine) * lineIndex.Length];
```
joperezr, line 139: It is dangerous to blindly do stackalloc from size that is provided as a parameter as you could easily hit the max size which would crash the app. Before doing this we should some validation that the data passed in will fit the LCD and fail early if not.

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,256 @@
...
+        /// <param name="display">Display ON/OFF signal</param>
+        /// <param name="externalCom">External COM inversion signal input</param>
+        internal LSxxxB7DHxx(SpiDevice spi, GpioController? gpio = null, bool shouldDispose = true, int chipSelect = -1, int display = -1, int externalCom = -1)
+        {
+            _spi = spi ?? throw new ArgumentNullException(nameof(spi));
+
+            if (chipSelect != -1 || display != -1 || externalCom != -1)
+            {
+                _shouldDispose = gpio == null || shouldDispose;
+                _gpio = gpio;
```
joperezr, line 88: Should we error out in case pins were passed in but gpio is null? or should we in this case just initialize a default gpiocontroller?

src/devices/MemoryLCD/LSxxxB7DHxx.cs
```diff
@@ -0,0 +1,256 @@
...
+            Span<byte> outputBuffer = stackalloc byte[2];
+
+            outputBuffer[0] = (byte)(ModeSelectionPeriodByte.AllClear | (frameInversion ? ModeSelectionPeriodByte.FrameInversion : ModeSelectionPeriodByte.Dummy));
+
+            WriteSpi(outputBuffer);
+        }
+
+        private void Init()
+        {
+            if (_gpio != null)
```
joperezr, line 184: What happens if gpio is not defined?

krwq:
Triage: @HumJ0218 do you think you will have time to address @joperezr's comments?

joperezr:
[Triage] @HumJ0218 do you still plan on working on this one? We added some feedback and it hasn't been addressed, let us know if you are interested on still working on this so that we keep the PR open.

joperezr (changes requested):
Please take a look at the unresolved comments and the build errors when you are able to.

PR #1126, opened 2020-07-03
Add Charlieplex binding

richlander:
Add binding for [charlieplexing](https://en.wikipedia.org/wiki/Charlieplexing).

Open questions:

- `CharlieLoad` isn't a great name. The load will most often be an LED, but not always. I'll need some help there.
- Note sure if `Iot.Device.Multiplex` is a great namespace name.

Update:

- `Charlieplex` -> `CharlieplexSegment`
- `CharlieLoad` -> `CharlieplexSegmentNode`
- `Iot.Device.Multiplex` -> `Iot.Device.Multiplexing`


pgrawehr:
>     * Note sure if `Iot.Device.Multiplex` is a great namespace name.

Maybe better `Iot.Device.Multiplexer`. There could be other multiplexer devices as well. I also think that the folder name and the namespace name should be equal.

src/devices/Charlieplex/Charlieplex.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// <summary>
+        /// The number of load devices (like LEDs) that can be addressed.
+        /// </summary>
+        public int LoadCount => _nodeCount;
+
+        /// <summary>
+        /// Write a PinValue to a load.
+        /// Address scheme is 0-based. Given 8 loads, addresses would be 0-7.
+        /// </summary>
+        public void Write(int node, PinValue value, int delay = 10)
```
pgrawehr, line 70: Document arguments. And prefer TimeSpan for delays.

src/devices/Charlieplex/Charlieplex.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            for (int i = 0; i < delay; i++)
+            {
+                WriteSegment();
+            }
+        }
+
+        private void WriteSegment()
+        {
+            // int delay = 1;
```
pgrawehr, line 82: Remove commented code

src/devices/Charlieplex/Charlieplex.cs
```diff
@@ -0,0 +1,175 @@
...
+                _controller.SetPinMode(node.Cathode, PinMode.Input);
+            }
+
+        }
+
+        /// <summary>
+        /// Provides the set of Charlie loads given the set of pins and the count provided.
+        /// If count = 0, then the Charlieplex maximum is used for the pins provided (n^2-n).
+        /// </summary>
+        public static CharlieplexNode[] GetCharlieLoads(int[] pins, int loadCount = 0)
```
pgrawehr, line 115: Does this really need to be public? Inputs should be validated in the Ctor. Maybe instead offer a property to retrieve the _carlieNodes member?

richlander: This makes it so that someone can call this API on any machine to get the correct wiring layout. That's useful. It's also what the tests uses for the wiring layout, again, runnable on any machine. The issue is that the constructor needs a GpioController, which only works on a Pi and similar devices.

src/devices/Charlieplex/Charlieplex.cs
richlander: I tried taking a TimeSpan. It made calling the API so much more verbose. Currently, the API can be called with three ints. that's really nice.

richlander: I adopted TimeSpan for DisplaySegment. That creates asymmetry, but better accounts for their different use cases.

src/devices/Charlieplex/Charlieplex.cs
richlander: As I read docs / blog posts for other solutions, I found their description of how to wire the circuit confusing at times, or not super well connected to the code. This API is intended to solve that. It also covers the case where you want to enable fewer nodes than the max for the number of pins.

richlander:
Sorry for making this PR before it was ready. I realized that a few hours after creating the PR initially. It's working pretty well now. There are pictures and videos in the README.

As a second phase, I think I'll add cancellation token support to the DisplaySegment API.

src/devices/Charlieplex/Charlieplex.csproj
```diff
@@ -0,0 +1,17 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Device.Gpio" Version="1.1.0-prerelease.20153.1" />
+  </ItemGroup>
+
+<ItemGroup>
```
joperezr, line 12: NIT: Identation

src/devices/Charlieplex/Charlieplex.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Device.Gpio" Version="1.1.0-prerelease.20153.1" />
```
joperezr, line 9: Remove this ItemGroup, we should only have the ProjectReference on device bindings

src/devices/Charlieplex/tests/tests.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
joperezr, line 1: NIT: Usually we call the UnitTests projects like:
<BindingName>.Tests.csproj so that they run as expected and test logs are identifyiable to a binding so please rename this to be Charlieplex.Tests.csproj

src/devices/Charlieplex/tests/tests.csproj
```diff
@@ -0,0 +1,22 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="../Charlieplex.csproj"/>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.6.1" />
+    <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.2">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="coverlet.collector" Version="1.3.0">
```
joperezr, line 16: NIT: I don't think you need coverlet reference right?

src/devices/Charlieplex/tests/tests.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../Charlieplex.csproj"/>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.6.1" />
+    <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.2">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
```
joperezr, line 13: you shouldn't need any of these two metadata on the PackageReference as nobody will reference this to package it.

src/devices/Charlieplex/CharlieplexSegment.cs
```diff
@@ -0,0 +1,244 @@
...
+                _controller?.Dispose();
+                _controller = null;
+            }
+        }
+
+        /// <summary>
+        /// Represents a node in a Charlieplexed circuit.
+        /// https://wikipedia.org/wiki/Charlieplexing
+        /// </summary>
+        public struct Node
```
joperezr, line 226: NIT: Should it make sense to extract this struct out of the class instead? I ask specifically since you have a method that is basically returning you an array of these so it is a type that is exposed by your API. No strong feelings about this so I'm fine with what you choose

src/devices/Charlieplex/CharlieplexSegment.cs
```diff
@@ -0,0 +1,244 @@
...
+{
+    /// <summary>
+    /// Provides support for Charlieplex multiplexing.
+    /// https://wikipedia.org/wiki/Charlieplexing
+    /// </summary>
+    public class CharlieplexSegment : IDisposable
+    {
+        private readonly bool _shouldDispose;
+        private GpioController _controller;
+        private int[] _pins;
```
joperezr, line 16: Should all of these 4 members be readonly?

src/devices/Charlieplex/CharlieplexSegment.cs
```diff
@@ -0,0 +1,244 @@
...
+        private int _nodeCount;
+        private Node[] _nodes;
+        private Node _lastNode;
+
+        /// <summary>
+        /// Initializes a new Charlieplex type that can be use for multiplex over a relatively small number of GPIO pins.
+        /// </summary>
+        /// <param name="pins">The set of pins to use.</param>
+        /// <param name="nodeCount">The count of nodes (like LEDs) that will be addressable. If 0, then the Charlieplex maximum is used for the pins provided (n^2-n).</param>
```
joperezr, line 25: Seems like most of the time you will want this to be 0 right? is there any scenario where it makes sense to manually specify the nodeCount?

joperezr: nvm I think this was a dumb question I guess.

src/devices/Charlieplex/CharlieplexSegment.cs
```diff
@@ -0,0 +1,244 @@
...
+        public int NodeCount => _nodeCount;
+
+        /// <summary>
+        /// Write a PinValue to a node, to update Charlieplex segment.
+        /// Address scheme is 0-based. Given 8 nodes, addresses would be 0-7.
+        /// Displays nodes in their updated configuration for the specified duration.
+        /// </summary>
+        /// <param name="node">Node to update.</param>
+        /// <param name="value">Value to write.</param>
+        /// <param name="duration">Time to display segment, in milliseconds.</param>
```
joperezr, line 87: Why do we default to 10 ms? would it make sense to instead just have a Write which writes to the node the value but doesn't do it for a specific time? Basically the equivalent of what a gpioController.Write() would be.

src/devices/Charlieplex/CharlieplexSegment.cs
```diff
@@ -0,0 +1,244 @@
...
+        /// <param name="gpioController">The GPIO Controller used for interrupt handling.</param>
+        /// <param name="shouldDispose">True (the default) if the GPIO controller shall be disposed when disposing this instance.</param>
+        public CharlieplexSegment(int[] pins, int nodeCount = 0,  GpioController gpioController = null, bool shouldDispose = true)
+        {
+            if (pins.Length < 2)
+            {
+                throw new ArgumentException($"{nameof(CharlieplexSegment)}: 2 or more pins must be provided.");
+            }
+
+            var charlieCount = (int)Math.Pow(pins.Length, 2) - pins.Length;
```
joperezr, line 35: NIT: can we change var -> explicit types?

src/devices/Charlieplex/CharlieplexSegment.cs
```diff
@@ -0,0 +1,244 @@
...
+                nodeCount = charlieCount;
+            }
+
+            if (gpioController is null)
+            {
+                gpioController = new GpioController();
+            }
+
+            // first two pins will be needed as Output.
+            gpioController.OpenPin(pins[0]);
```
joperezr, line 52: NIT: we have an OpenPin that also sets the pinmode.

joperezr:
Overall looks good but left a few comments. Also we seem to be missing a fritz diagram 😄

src/devices/Charlieplex/tests/tests.csproj
richlander: I don't even know where that came from. Will remove.

src/devices/Charlieplex/CharlieplexSegment.cs
richlander: Yes, it will be frequently specified. Once you get past a small number of pins, the numbers get big quickly and won't necessarily match the number of LEDs you are using.

src/devices/Charlieplex/CharlieplexSegment.cs
richlander: Good question. The underlying question is whether the LED should be lit as as a result of calling `.Write`.

richlander:
Thanks for the feedback. I was thinking about the fritizing diagram. I think its too complicated to be useful. The diagrams in the README are basically it. Same ones as at https://en.wikipedia.org/wiki/Charlieplexing. This is a very different type of binding.

src/devices/Charlieplex/CharlieplexSegment.cs
richlander: I think one of two options make sense:

- Keep as-is
- Pull out and then rename to `CharlieplexSegmentNode`

This seems very similar to what I did with the PinMapping type for the shift register: https://github.com/dotnet/iot/blob/master/src/devices/Sn74hc595/Sn74hc595.cs#L272

It's the pattern where there is a type that is both tightly bound to another type and that has a very narrow role to play. That seems like it is well suited for a nest type. Is there good API design guidance for that case?

richlander: I talked to @terrajobst. He said go with option 2 above. I did.

src/devices/Charlieplex/CharlieplexSegment.cs
richlander: I thought more about this. Switched to default of 0 by default.

richlander:
Applied all feedback. Thanks!

richlander:
I talked to @terrajobst. He recommend `.Multiplexing` as the better namespace name. I prefer that. It's updated.

richlander:
Separate from this PR, but I'm wondering if we should move the shift register binding I made to this same `.Multiplexing` namespace. Thoughts? 

https://github.com/dotnet/iot/blob/master/src/devices/Sn74hc595/Sn74hc595.cs#L9

src/devices/Charlieplex/Charlieplex.cs
pgrawehr: Sounds reasonable then. Just maybe prefer List<CharliplexNode> as return type instead of an array?

PR #1128, opened 2020-07-03
Board approach, iteration 2

pgrawehr:
Adds a Board base class to simplify handling common features of a piece of hardware. 

Also adds support for proper pin mode handling on the Raspberry Pi. I.e. it is now possible to use pins interchangeably as PWM and GPIO, or switch between software and hardware SPI on the same set of pins.


src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -15,7 +15,7 @@ namespace System.Device.Gpio
     /// <summary>
     /// Represents a general-purpose I/O (GPIO) controller.
     /// </summary>
-    public sealed class GpioController : IDisposable
+    public class GpioController : IDisposable
```
colombod, line 18: Brilliant! This simplifies soo many things

colombod:
This set of abstraction and changes are key to simplify and consolidate other boards design like the piTop 4 module. hope this is merged soon. Very excellent work.

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,407 @@
...
+
+        /// <summary>
+        /// This method shall be overriden by clients, providing just the basic interface to an I2cDevice.
+        /// </summary>
+        /// <param name="connectionSettings">I2C Connection settings</param>
+        /// <param name="pinAssignment">Pins assigned to the device</param>
+        /// <returns>An I2cDevice connection</returns>
+        protected abstract I2cDevice CreateSimpleI2cDevice(I2cConnectionSettings connectionSettings, int[] pinAssignment);
+
+        public I2cDevice CreateI2cDevice(I2cConnectionSettings connectionSettings, int[] pinAssignment, PinNumberingScheme pinNumberingScheme)
```
colombod, line 245: this interface will always create a device manager even if the same connection settings are used. it would be more interesting if there was the option here to track the connections setting used so far, that would make the board something that can be inspected like the pitop boards and plates in this [repo](https://github.com/colombod/pi-top) .

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,407 @@
...
+        {
+            if (pinAssignment == null || pinAssignment.Length != 2)
+            {
+                throw new ArgumentException($"Invalid argument. Must provide exactly two pins for I2C", nameof(pinAssignment));
+            }
+
+            return new I2cDeviceManager(this, connectionSettings, RemapPins(pinAssignment, pinNumberingScheme), CreateSimpleI2cDevice);
```
colombod, line 252: tracking connection here with getOrCreate instead of going off with a new operation every time could be interesting

src/devices/Board/Board.cs
pgrawehr: So you would want to make sure that each device (here: every instance that uses the same address) can be created only once? Could indeed further reduce usage errors, since most devices won't like it if they're being used i.e. from multiple bindings at the same time.

colombod: Yes you are spot on. In my simple implementation reusing a connection settings yield to the already built device. the interesting part is that the device managers can then auto remove themselves if disposed. Do you think this could imply the need for some sort of reference counting? there are some useful disposables classes in rx.Net that could be very useful like refcount disposables, serial disposables and disposable factories.

src/devices/Board/I2cDeviceManager.cs
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        public override void WriteRead(ReadOnlySpan<byte> writeBuffer, Span<byte> readBuffer)
+        {
+            _i2cDeviceImplementation.WriteRead(writeBuffer, readBuffer);
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
```
colombod, line 77: if a device is recreated with same connection do we need to track some sort of reference counting?

pgrawehr: Normally, you would now create an I2cDevice for a particular address and then fed this into a binding (ie. a BME680, an MCP28xxx or whatever) and not directly use the device any further. If you create another instance of I2cDevice for the same physical device and use it differntly, or create another instance of a binding with it, you'll very likely run into weird problems, because the binding expects that any state change of the device only happens trough that binding. 

Reference counting would be needed if we want multiple instances for the same device, but I think we shouldn't allow that. So if we extend these checks, we should just prevent creation of the same I2cDevice a second time (until the first one is disposed). For that, remembering whether a particular device was already created is sufficient (can use a Dictionary or a Hashtable for that).

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -132,6 +133,36 @@ protected internal override int ConvertPinNumberToLogicalNumberingScheme(int pin
         /// <inheritdoc/>
         protected internal override void Write(int pinNumber, PinValue value) => _internalDriver.Write(pinNumber, value);
 
+        /// <summary>
+        /// Returns the currently set pin mode by directly reading the hardware
+        /// </summary>
+        /// <param name="pinNumber">Pin number</param>
+        /// <returns>(Alternate) Pin mode. 0 = Alt0, 1= Alt1... -1 Gpio Input, -2 Gpio Output</returns>
+        protected internal int GetAlternatePinMode(int pinNumber)
```
krwq, line 141: should this be an enum?

krwq: Ah, this is about current pin mode, not all possible options...

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -367,6 +367,102 @@ private void SetInputPullModePi4(int pinNumber, PinMode mode)
...
+            uint* registerPointer = &_registerViewPointer->GPFSEL[pinNumber / 10];
+            uint register = *registerPointer;
+            // get the three bits of the register
+            register = (register >> shift) & 0b111;
+
+            switch (register)
+            {
+                case 0b000:
+                    // Input
+                    return -1;
```
krwq, line 446: what are negative values, you didn't allow them for setter

krwq: also, assuming we go with IGpioPin design this could be an implementation detail of RaspberryPi3Pin.PinMode

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -196,9 +218,9 @@ public PinValue Read(int pinNumber)
         /// </summary>
         /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
         /// <param name="value">The value to be written to the pin.</param>
-        public void Write(int pinNumber, PinValue value)
+        public virtual void Write(int pinNumber, PinValue value)
```
krwq, line 221: I'm curious, why all these need to be virtual instead of replacing the driver? the initial idea was that you should be changing the driver, not the controller (although maybe would be simpler this way if we did that from the start)

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -212,14 +234,50 @@ public void Write(int pinNumber, PinValue value)
...
+        /// <summary>
+        /// Returns the currently set pin mode by directly reading the hardware
+        /// </summary>
+        /// <param name="pinNumber">Pin number</param>
+        /// <returns>(Alternate) Pin mode. 0 = Alt0, 1= Alt1... -1 Gpio Input, -2 Gpio Output</returns>
+        protected internal int GetAlternatePinMode(int pinNumber)
+        {
+            int logicalPinNumber = GetLogicalPinNumber(pinNumber, NumberingScheme);
+            // TODO: Maybe use an interface for this feature query instead?
+            RaspberryPi3Driver driver = _driver as RaspberryPi3Driver;
```
krwq, line 246: no, it should be generic or implementation specific, this will be very annoying to support long term

src/System.Device.Gpio/System/Device/Gpio/PinValue.cs
```diff
@@ -82,6 +82,21 @@ public override bool Equals(object obj)
         /// </summary>
         public static bool operator !=(PinValue a, PinValue b) => !a.Equals(b);
 
+        /// <summary>
+        /// Returns the inverse of the value
+        /// </summary>
+        /// <param name="v">Input value</param>
+        /// <returns>High, when the input is low, low when the input is high</returns>
+        public static PinValue operator !(PinValue v)
```
krwq, line 90: 👍

src/devices/Board/AlternatePinMode.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Iot.Device.Board
+{
+    /// <summary>
+    /// Alternate pin modes, for pins that support them
+    /// </summary>
+    public enum AlternatePinMode
+    {
+        // Note: Do not change the enum values without reviewing the impact
+        Unknown = -2,
```
krwq, line 13: I saw you were using int directly instead of this enum, now the negative values are clearer, you should use this enum directly

src/devices/Board/AlternatePinMode.cs
```diff
@@ -0,0 +1,29 @@
...
+        NotSupported = -1,
+        Gpio = 0,
+
+        // the names are Raspi specific, but this is considered a generic representation of alternate modes (if any)
+        Alt0 = 1,
```
krwq, line 18: I'm wondering if this should rather be: Pwm0, Pwm1, Spi0Mosi, etc (hope you got the idea), the Alt0, Alt1... seem very generic, might be fine as an intermediate abstraction layer but it's not very helpful unless you got a spec in front of you

src/devices/Board/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Board.Tests")]
```
krwq, line 3: this is actually doable entirely from the csproj, see: https://github.com/dotnet/iot/blob/e3449040842ca153151cf6148aec133993b8d904/src/devices/Amg88xx/Amg88xx.csproj#L15-L20 (in pending PR)

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,407 @@
+﻿using System;
```
krwq, line 1: you're missing copyright header in couple of files

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,407 @@
...
+
+        public event Action<string, Exception> LogMessages;
+
+        protected bool Initialized
+        {
+            get
+            {
+                return _initialized;
+            }
+        }
```
krwq, line 43:
```suggestion
        protected bool Initialized => _initialized;
```

here and below

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,407 @@
...
+        {
+            get
+            {
+                return _defaultNumberingScheme;
+            }
+        }
+
+        protected void Log(string message, Exception exception = null)
+        {
+            LogMessages?.Invoke(message, null);
```
krwq, line 63: I'd prefer logging wasn't part of Board abstraction, it can use it but it should not implement or abstract it

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,407 @@
...
+        public abstract int ConvertPinNumber(int pinNumber, PinNumberingScheme inputScheme, PinNumberingScheme outputScheme);
+
+        /// <summary>
+        /// Reserves a pin for a specific usage. This is done automatically if a known interface (i.e. GpioController) is
+        /// used to open the pin, but may be used to block a pin explicitly, i.e. for UART.
+        /// </summary>
+        /// <param name="pinNumber">The pin number, in the boards default numbering scheme</param>
+        /// <param name="usage">Intended usage of the pin</param>
+        /// <param name="owner">Class that owns the pin (use "this")</param>
+        public virtual void ReservePin(int pinNumber, PinUsage usage, object owner)
```
krwq, line 82: I'd prefer this was not part of the initial version of abstraction, this needs more thought. Perhaps this is needed long term and this is how the API would look like but I'd prefer we did it separately as this PR is already large enough for this rather niche concept

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,407 @@
...
+        public abstract PinUsage DetermineCurrentPinUsage(int pinNumber);
+
+        /// <summary>
+        /// Override this method if something special needs to be done to use the pin for the given device.
+        /// Many devices support multiple functions per Pin, but not at the same time, so that some kind of
+        /// multiplexer needs to be set accordingly.
+        /// </summary>
+        /// <param name="pinNumber">The logical pin number to use.</param>
+        /// <param name="usage">The intended usage</param>
+        protected virtual void ActivatePinMode(int pinNumber, PinUsage usage)
```
krwq, line 187: if anything this should be throwing by default but I'd rather keep this concept out of the initial abstraction

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,407 @@
...
+            public int Pin { get; }
+            public PinUsage Usage { get; }
+
+            /// <summary>
+            /// Component that owns the pin (used mainly for debugging)
+            /// </summary>
+            public object Owner { get; }
+        }
+
+        public virtual GpioController CreateGpioController(int[] pinAssignment = null)
```
krwq, line 230: I'm not sure what pinAssignment would do here

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
pgrawehr: I would really like to use an enum here. Unfortunately, the set of alternate pin modes is hardware dependent. So if I go with the names that are most common here, they would be Raspberry Pi-specific and be confusing to use on some other hardware (the Arduino binding for instance uses a similar semantics, but with different values)

Maybe return a class instead? (That would be handled much like a handle for most of what the client does)?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
pgrawehr: Similar problem as with the enum above: I was intending to use 0-based addressing, so that Alt0 is 0, Alt1 is 1, etc. Then Input and Output fall out of that scheme, while technically they belong to the same set (There are 7 possible modes (3 bit): Input, Output, Alt0, Alt1, Alt2, Alt3, Alt4, Alt5 for each pin). 

A class type could work around that as well.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: @joperezr and me came up with the suggestion to unseal the GpioController because of several issues:

- If you want to add features to the GpioController (such as pin management, for the board approach), there's currently no way to do that (because GpioController is sealed). 
- Since all methods (except the .ctor) on the Driver class are protected, it is not even possible to create a replacement for the GpioController and call into existing driver implementations.
- A sealed class cannot be mocked for unit testing

Therefore, the controller is now unsealed and all methods that you might want to override or mock are virtual. The change should not be breaking for anyone.

src/devices/Board/AlternatePinMode.cs
pgrawehr: Oh, apparently I got confused myself. 

I think I'll try a class instead.

src/devices/Board/Board.cs
pgrawehr: Yes, and documentation as well, I suppose

src/devices/Board/Board.cs
pgrawehr: I haven't gotten friend with this new syntax yet...

src/devices/Board/Board.cs
pgrawehr: I think I'll put up a small generic logging API, to be usable for all bindings.

src/devices/Board/Board.cs
pgrawehr: That is to allow creating a GpioController that handles only a set of pins. I added it to be consistent with the other CreateXYZ()-Methods, where you can also specify the pins to use. 

When we do the IGpioPin approach discussed elsewhere, this would be a duplicate.

src/devices/Board/Board.cs
pgrawehr: I have to check whether it can be removed easily. I think the method is internally used already.

src/devices/Board/Board.cs
pgrawehr: I'll consider this for later, I think. Adding a test that no two I2cDevice instances with the same ConnectionSettings can be done without changing the API later.

src/devices/Board/Board.cs
pgrawehr: That's probably to much hasle. Since you'll not normally use the device directly (but some binding instead), caching seems unnecessary.

src/devices/Board/Board.cs
pgrawehr: ... but on a separate PR.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: I was trying to keep the impact on the System.Device assembly as low as possible (by not adding new methods in the base classes), but I'm happy to do it right.

src/devices/Board/AlternatePinMode.cs
pgrawehr: The user on the top-level API will not have to worry what these mean. Even though `GetHardwareModeForPinUsage` is public, the user rarely needs to use this directly.

src/devices/Board/Board.cs
Ellerbach: Can we reuse the logger I've been using for the nfc readers? 
https://github.com/dotnet/iot/blob/master/src/devices/Card/LogInfo.cs 

Should be generic enough. We can move to the common folder maybe?

pgrawehr: I'll check what we already have, yes. Just this shouldn't be part of this PR (even if it means that these methods will change again)

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
krwq: if you have some decent design with struct (class is too heavy to be used instead of enum) then fine, otherwise I'm fine with leaving as is

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
krwq: fine with me

krwq: cc: @joperezr

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
```diff
@@ -0,0 +1,128 @@
+﻿using System;
```
krwq, line 1: copyright header missing

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// Platform-dependent description of an alternate pin mode (i.e. a pin that is used for I2C or SPI transfer).
+    /// Depending on the hardware in use, the use will reflect the effective mode ("PWM", "SPI") or the state of the
+    /// mode multiplexer of the CPU ("ALT0", "ALT1").
+    /// </summary>
+    public class AlternatePinMode : IEquatable<AlternatePinMode>
+    {
+        /// <summary>
+        /// Default Gpio mode
+        /// </summary>
+        public static AlternatePinMode Gpio = new AlternatePinMode("GPIO", 0);
```
krwq, line 17: since you're using statics this is fine as you don't need to allocate it all the time. I'll just double check this is immutable

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
```diff
@@ -0,0 +1,128 @@
...
+                return false;
+            }
+
+            return other.ModeValue == ModeValue && other.ModeName == ModeName;
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            return ModeValue.GetHashCode() ^ ModeName.GetHashCode();
```
krwq, line 109: nope, use this instead https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=dotnet-plat-ext-3.1

As is it is really easy to create collisions (even though it is right now meant to be used with user coming data, we have historically seen this happening eventually)

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -146,6 +146,18 @@ protected internal override int ConvertPinNumberToLogicalNumberingScheme(int pin
         /// <inheritdoc/>
         protected internal override void Write(int pinNumber, PinValue value) => _internalDriver.Write(pinNumber, value);
 
+        /// <inheritdoc />
+        protected internal override AlternatePinMode GetAlternatePinMode(int pinNumber)
```
krwq, line 150: @joperezr if we go ahead with GpioPin then these methods should be moved there as a property

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -367,6 +367,83 @@ private void SetInputPullModePi4(int pinNumber, PinMode mode)
...
+                case 0b111:
+                    return AlternatePinMode.Alt3;
+                case 0b011:
+                    return AlternatePinMode.Alt4;
+                case 0b010:
+                    return AlternatePinMode.Alt5;
+            }
+
+            // This cannot happen.
+            throw new InvalidOperationException("Invalid register value");
```
krwq, line 444: I wonder if we should just assume that if we can't read it then it's GPIO

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -60,25 +70,26 @@ public GpioController(PinNumberingScheme numberingScheme, GpioDriver driver)
...
-            return (NumberingScheme == PinNumberingScheme.Logical) ? pinNumber : _driver.ConvertPinNumberToLogicalNumberingScheme(pinNumber);
+            return _driver.ConvertPinNumberToLogicalNumberingScheme(pinNumber);
         }
 
         /// <summary>
         /// Opens a pin in order for it to be ready to use.
         /// </summary>
         /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
-        public void OpenPin(int pinNumber)
+        public virtual void OpenPin(int pinNumber)
```
krwq, line 90: @Ellerbach @Ellerbach if we ship this in this release then likely we won't be able to relatively safely change return type as previously source compatible change can now be overriden by someone and now it won't be source compatible

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -105,10 +116,22 @@ public void OpenPin(int pinNumber, PinMode mode)
...
+            ClosePin(pinNumber, NumberingScheme);
+        }
+
+        /// <summary>
+        /// Closes an open pin.
+        /// </summary>
+        /// <param name="pinNumber">The pin number.</param>
+        /// <param name="numberingScheme">Numbering scheme for the given pin</param>
+        /// <remarks>Internal use, to make Dispose work consistently</remarks>
+        protected virtual void ClosePin(int pinNumber, PinNumberingScheme numberingScheme)
```
krwq, line 128: isn't numbering scheme global per GpioController? (maybe I missed this change somewhere in here)

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -120,9 +143,9 @@ public void ClosePin(int pinNumber)
         /// </summary>
         /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
         /// <param name="mode">The mode to be set.</param>
-        public void SetPinMode(int pinNumber, PinMode mode)
+        public virtual void SetPinMode(int pinNumber, PinMode mode)
```
krwq, line 146: if ClosePin takes numbering scheme why doesn't this? (similar comment elsewhere)

krwq:
@pgrawehr I haven't finished reviewing this PR because I noticed there is multiple things changed here at once and I'll likely need to re-review later. There are couple of things being changed here at the same time:
- support for alternative pin modes
- unsealing GpioController
- some bug fixes
- tests
- perhaps more

I think to move this forward overall quicker I think you should split this and do it step-by-step. I recommend with support for alternative pin modes as it seems short term most useful. It will be much easier to understand what problems you're solving and focus on getting this right. You can leave this PR open meantime though. Please let me know if you disagree.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: Yes, it currently is. The reason for this extra overload is the Dispose method, because we only know the logical numbers of the open pins. Also, this would probably be needed if adding the IGpioPin interface, as callback on dispose. 

Maybe it should even be public, then.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: We could add these overloads, yes. This would allow specifying, for a particular call, that you use another than the default numbering scheme. But I'd rather do that on a separate PR. For the close, it is required for different reasons (see above).

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
pgrawehr: This here cannot realistically happen. The value is masked above to 0b111 and all cases are handled.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
krwq: @pgrawehr I think if GpioController was created with specific numberingScheme then you should use that all the way, you should not close pins which are not managed by the gpio controller

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
krwq: technically you can add extra fields in the AlternatePinMode but assuming this actually shouldn't ever happen we shouldn't try to fix this anyway (I missed the comment previously)

pgrawehr: Hmm... True. But that applies to the inverse only, not this method. I'll add a check there.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
pgrawehr: Used a class with static members now. This syntactically looks identical to using an enum, but is extensible.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: This has been done now

src/devices/Board/Board.cs
pgrawehr: I left this in for now, but plan to replace it once we agree on a logging system/library.

src/devices/Board/Board.cs
pgrawehr: This is one of the main features of the new approach, actually. It manages the pin usages and keeps track of who uses which pin for what purpose. This is public to also allow reserving pins directly from a client (i.e. for UART, which is not currently managed here)

src/devices/Board/Board.cs
pgrawehr: This also belongs to the main feature: It allows (see the RaspberryPiBoard specific implementation) to set the alternate pin mode for a pin. This internally calls SetAlternatePinMode on the Raspi

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
pgrawehr: All of these done (i think)

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
pgrawehr: Yes, an instance of AlternatePinMode is immutable.

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
pgrawehr: Not sure how that would be any better. Collisions are to be expected for this method, that should not cause any harm (except a performance degradation). 
Besides that this would require an extra dependency. System.HashCode is not generally available.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: Already discussed and solved separtely

pgrawehr:
@krwq: Did another round of updates. Please review again or suggest more splitting. I could do the System part separatelly, but like this, the use case(s) are better visible, I think.

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
krwq: fine

src/devices/Interop/Windows/WinUser.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
```
krwq, line 1: another one with missing header. as a side not, haven't read the reminder of code yet but is this actually needed?

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
pgrawehr: Ups, forgot to check the files in the System assembly.

src/devices/Interop/Windows/WinUser.cs
pgrawehr: Yes, this is required for the "KeyboardGpioDriver", which is a nice driver for tests on standard PC hardware (the functions are used to blink the keyboard leds)

src/devices/Board/I2cDeviceManager.cs
```diff
@@ -0,0 +1,131 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Device.I2c;
+using System.Text;
+
+namespace Iot.Device.Board
+{
+    internal class I2cDeviceManager : I2cDevice
```
Ellerbach, line 11: So this is maybe where we should as well think of #1335?

src/devices/Board/PinUsage.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        /// <summary>
+        /// Pin used for RS-232
+        /// </summary>
+        Uart = 5,
+
+        /// <summary>
+        /// Pin used for analog input
+        /// </summary>
+        AnalogIn = 6,
```
Ellerbach, line 44: Shall we add an AnalogOut as well?

Ellerbach (commented):
I **love** the Windows Keyboard Gpio Driver :-D

src/devices/Board/I2cDeviceManager.cs
pgrawehr: We could add it here, but I think it would better be one level up, since this is just an extended version of the I2cDevice class.

src/devices/Board/PinUsage.cs
pgrawehr: Is an option, but I'm not sure... Many devices I know use the PWM interface to also control those.

Ellerbach: GrovePi offers real Analog Output for example. It will then perfectly fit here.

src/devices/Board/I2cDeviceManager.cs
Ellerbach: If we have the level up, then not sure we need this one. At the same time, that can do the work the time to have the higher level which can anyway be added.

src/devices/Board/PinUsage.cs
pgrawehr: I know, some arduinos also have real DACs. What I meant is that some boards use the same interface methods to set a value to a DAC like to set a PWM pin (using just the duty cycle and ignoring the frequency). But that should probably stay implementation-dependent. I'll add the value.

colombod: Pi-top also have ping used for analog

joperezr:
[Triage] We are waiting on https://github.com/dotnet/iot/issues/1335 to go in so that there is an I2cBus in System.Device.Gpio library, and then we can adapt this PR to use that instead of having it's own implementation of I2cBus.

krwq:
@pgrawehr I'll review this once you rebase and pull in I2cBus changes

pgrawehr:
@krwq: This is ready for another round of review. I'll have to run a few more tests on actual hardware, but the API isn't likely to change now.

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// Platform-dependent description of an alternate pin mode (i.e. a pin that is used for I2C or SPI transfer).
+    /// Depending on the hardware in use, the use will reflect the effective mode ("PWM", "SPI") or the state of the
+    /// mode multiplexer of the CPU ("ALT0", "ALT1").
+    /// </summary>
+    public class AlternatePinMode : IEquatable<AlternatePinMode>
+    {
+        /// <summary>
+        /// Default Gpio mode
+        /// </summary>
+        public static AlternatePinMode Gpio = new AlternatePinMode("GPIO", 0);
```
krwq, line 20: nit:

```suggestion
        public static readonly AlternatePinMode Gpio = new AlternatePinMode("GPIO", 0);
```

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <summary>
+        /// The mode is not known
+        /// </summary>
+        public static AlternatePinMode Unknown = new AlternatePinMode("Unknown", 0xffffffff);
```
krwq, line 25: ditto here and below

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -350,6 +350,87 @@ private void SetInputPullModePi4(int pinNumber, PinMode mode)
...
+            /*
+             * There are 6 registers(4-byte ints) that control the mode for all pins. Each
+             * register controls the mode for 10 pins. Each pin uses 3 bits in the register
+             * containing the mode.
+             */
+
+            // Define the shift to get the right 3 bits in the register
+            int shift = (pinNumber % 10) * 3;
+            // Gets a pointer to the register that holds the mode for the pin
+            uint* registerPointer = &_registerViewPointer->GPFSEL[pinNumber / 10];
```
krwq, line 404: make sure to throw here when pinNumber is gonna cause OutOfRange here. This is writing to memory directly so you should protected against that here

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
```diff
@@ -125,5 +125,33 @@ protected virtual void Dispose(bool disposing)
         {
             // Nothing to do in base class.
         }
+
```
krwq, line 128: I suggest to split alternative pin mode to a separate PR than Board

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,708 @@
...
+            _i2cBuses = new Dictionary<int, I2cBusManager>();
+            _pinReservationsLock = new object();
+            _initialized = false;
+            _disposed = false;
+        }
+
+        /// <summary>
+        /// (Temporary) Register to this event to receive log messages.
+        /// </summary>
+        public event Action<string, Exception?>? LogMessages;
```
krwq, line 61: this should not be public

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,708 @@
...
+        /// <param name="defaultNumberingScheme">Default numbering scheme for the board.</param>
+        /// <remarks>
+        /// The constructor will never throw an exception. Call <see cref="Initialize"/> to initialize the hardware and check
+        /// whether an instance can actually run on the current hardware.
+        /// </remarks>
+        protected Board(PinNumberingScheme defaultNumberingScheme)
+        {
+            _defaultNumberingScheme = defaultNumberingScheme;
+            _pinReservations = new Dictionary<int, List<PinReservation>>();
+            _i2cBuses = new Dictionary<int, I2cBusManager>();
```
krwq, line 52: should this just be `Dictionary<int, I2cBus>`

krwq: optional: also you could probably make it work with HashSet as the I2cBus contains busId

src/devices/Board/Board.cs
krwq: (at least not in this PR)

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,708 @@
...
+
+        /// <summary>
+        /// (Temporary) Register to this event to receive log messages.
+        /// </summary>
+        public event Action<string, Exception?>? LogMessages;
+
+        /// <summary>
+        /// True if the board instance is initialized
+        /// </summary>
+        protected bool Initialized => _initialized;
```
krwq, line 66: this feels like it shouldn't be public

src/devices/Board/Board.cs
krwq: let's do it separately regardless of decision

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,708 @@
...
+        }
+
+        /// <summary>
+        /// Converts pin numbers from one numbering scheme to another.
+        /// </summary>
+        /// <param name="pinNumber">Pin number to convert</param>
+        /// <param name="inputScheme">The numbering scheme of the input pin</param>
+        /// <param name="outputScheme">The desired numbering scheme</param>
+        /// <returns>The converted pin number</returns>
+        public abstract int ConvertPinNumber(int pinNumber, PinNumberingScheme inputScheme, PinNumberingScheme outputScheme);
```
krwq, line 106: you can go with two designs here:
- take PinNumberingScheme as input and assume inputScheme is whatever was passed in everywhere
- PinNumberingScheme becomes part of every API taking pinNumber

currently you mixed both. My preference is first option but we can discuss it during the triage meeting

src/devices/Board/PinUsage.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Board
+{
+    /// <summary>
+    /// Designated (or active) usage of a pin
+    /// </summary>
+    public enum PinUsage
```
krwq, line 9: I'm confused. I though these will be covered by AlternatePinMode. Should those two abstractions be joined?

krwq:
I think AlternativePinMode/PinUsage should be split out of this PR. I feel this specific concept needs to be cleared out and it should work independently of Board abstraction

src/devices/Board/Board.cs
pgrawehr: Removed, it's not used in this class so far, anyway.

src/devices/Board/Board.cs
pgrawehr: Removed

src/devices/Board/Board.cs
pgrawehr: This is protected, because derived classes should also be able to throw if this is false.

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// Alternate mode 5 for Raspberry Pi
+        /// </summary>
+        public static AlternatePinMode Alt5 = new AlternatePinMode("Alt5", 0b010);
+
+        /// <summary>
+        /// Creates a new mode.
+        /// </summary>
+        /// <param name="modeName">Common name of the mode</param>
+        /// <param name="modeValue">Internal value to identify this mode</param>
+        public AlternatePinMode(string modeName, uint modeValue)
```
Ellerbach, line 62: seems like AlternatePinMode is immutable, why not use the new record from c#9? With init in the 2 properties?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -350,6 +350,87 @@ private void SetInputPullModePi4(int pinNumber, PinMode mode)
             }
         }
 
+        /// <inheritdoc />
+        protected internal override void SetAlternatePinMode(int pinNumber, AlternatePinMode altMode)
```
Ellerbach, line 354: I know it's internal but I haven't seen anywhere explicit where the pinNumber is checked to be valid. I most likely have missed it. So if you're sure it's checked, ignore this comment :-)

src/devices/Board/CustomBoard.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    /// <summary>
+    /// A board that can be customized with user-specified drivers.
+    /// This should only be used if the drivers can't be auto-detected properly.
+    /// </summary>
+    public class CustomBoard : GenericBoard
+    {
+        private readonly GpioDriver _gpioDriver;
+        private readonly Func<int, I2cBus> _i2CBusCreator;
+        private readonly Func<SpiConnectionSettings, SpiDevice> _spiDeviceCreator;
```
Ellerbach, line 19: Can a board have multiple SPI devices? In this configuration, I'm not sure. Same question for PWM.
Question is then: shall it be a collection for both?
And tricky part: a SPI device, can be the same as another one with a little difference on the chip select. Which can make all this a bit more complicated

src/devices/Board/CustomBoard.cs
```diff
@@ -0,0 +1,83 @@
...
+            _spiDeviceCreator = spiDeviceCreator;
+            _pwmChannelCreator = pwmChannelCreator;
+        }
+
+        /// <inheritdoc />
+        public override int ConvertPinNumber(int pinNumber, PinNumberingScheme inputScheme, PinNumberingScheme outputScheme)
+        {
+            if (inputScheme != outputScheme)
+            {
+                throw new NotSupportedException("Only logical pin numbering supported");
```
Ellerbach, line 44: this message is not aligned with the check. If only logical is supported, that's what should be tested, not the fact that they are both different.

src/devices/Board/SpiDeviceManager.cs
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            _board = board;
+            ConnectionSettings = connectionSettings;
+            try
+            {
+                _board.ReservePin(pins[0], PinUsage.Spi, this);
+                _board.ReservePin(pins[1], PinUsage.Spi, this);
+                _board.ReservePin(pins[2], PinUsage.Spi, this);
+                if (pins.Length == 4)
```
Ellerbach, line 29: as before you checked for pin lenght <3 and message about this, you should most likelt check for >=4. I agree, it doesn't change much anyway

src/devices/Board/SpiDeviceManager.cs
```diff
@@ -0,0 +1,139 @@
...
+                    // numbering must be used here I think. So the value would be 0 or 1 for CE0 and CE1 in ALT 0 mode, which are
+                    // the logical pins 7 and 8.
+                    _board.ReservePin(pins[3], PinUsage.Spi, this);
+                }
+
+                _device = createOperation(connectionSettings, pins);
+            }
+            catch (Exception x)
+            {
+                Console.WriteLine($"Exception: {x}");
```
Ellerbach, line 41: You may remove the Console.WriteLine

src/devices/Board/SpiDeviceManager.cs
```diff
@@ -0,0 +1,139 @@
...
+                _board.ReleasePin(pins[0], PinUsage.Spi, this);
+                _board.ReleasePin(pins[1], PinUsage.Spi, this);
+                _board.ReleasePin(pins[2], PinUsage.Spi, this);
+                if (pins.Length == 4)
```
Ellerbach, line 45: same as before

Ellerbach (commented):
Few comments, most interesting ones are around the Spi Device Manager

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
pgrawehr: I'll check again, but generally the pin number is checked by the GpioController.

src/devices/Board/CustomBoard.cs
pgrawehr: Yes, a board can have multiple SPI devices. There's a special feature called "Pin sharing" that allows the board to specify that for some features (like SPI, or previously I2C), the same set of pins can be opened multiple times. In other words: It is possible to open the same pins multiple times in SPI mode, and reference counting is performed to ensure all devices are closed before a pin is used for something else.

src/devices/Board/SpiDeviceManager.cs
pgrawehr: Sure... That's were logging should be used instead.

krwq: You can create an internal method Log/Info/Error which doesn't do anything and change it once we figure out the convention we want to introduce

krwq: (or Debug.WriteLine for now)

src/devices/Board/SpiDeviceManager.cs
pgrawehr: Now testing for exactly 3 or 4 pins.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
pgrawehr: You were right, I had missed the test.

src/System.Device.Gpio/System/Device/Gpio/AlternatePinMode.cs
pgrawehr: I guess (from the specification proposal) that this would be a good idea, however I'm a bit hesitant in using it since it has not been fully documented yet, it seems. The [language reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/) doesn't mention the `record` keyword yet, and there are few usable examples about it.

arkadiuszwojcik: Init is also working with classes. So: public string ModeName { get; init; } should also work.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -23,6 +23,41 @@ public class RaspberryPi3Driver : GpioDriver
...
+
+        /// <summary>
+        /// Alternate mode 4 for Raspberry Pi
+        /// </summary>
+        public static readonly ExtendedPinMode Alt4Mode = new RaspberryPiExtendedPinMode("Alt4", 0b011);
+
+        /// <summary>
+        /// Alternate mode 5 for Raspberry Pi
+        /// </summary>
+        public static readonly ExtendedPinMode Alt5Mode = new RaspberryPiExtendedPinMode("Alt5", 0b010);
```
microhobby, line 59: Forgive me if I got here just now. Just for the knowledge this should not be done or decided in user space, it is not a good practice (at least in Linux). This should be decided in the description of the hardware in the Device Tree Source or Device tree overlay.

Don't get me wrong, nothing against its implementation.

microhobby: Sorry for being annoying with everything that touches what the Linux Kernel already does. I just have a concern that dotnet/iot becomes a library that is not a first class citizen when using Linux resources. We also have to be careful about how to recommend this use. For use in development environment it may be ok ...

pgrawehr: @microhobby I'm sorry, but I don't understand. Where does the linux kernel already provide this information or these functions? The latest libgpiod version for the raspi is still 1.2. Of course if that driver supports these features, that can be extended as well.

krwq: @microhobby the only reason we consider APIs like this are issues like: https://github.com/dotnet/iot/issues/874 - kernel currently doesn't seem to be solving this problem correctly, how would you recommend to do it otherwise?

microhobby: @pgrawehr the alternative modes are set by the Kernel on the boot, the pinctrl sub system handle this. The pinctrl subsystem load these information from the hardware description, the device tree source. All boards have it own device tree source: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm/boot/dts

gpiolib and libgpiod do not change alternative functions, were not designed for that. They were designed only to change pin states that are in the alternative GPIO function. Even because changing the alternate function of a pin is not a user space taks, as I said it happens in kernel space at boot by the pinctrl driver.

microhobby: @krwq in fact I think it’s not that the kernel isn’t solving the problem correctly, actually we’re overwriting what the kernel knows about the hardware. 

Since for RPI it seems we are using `/dev/mem` anyway I don't see why not follow with this solution, we should only be careful when recommending this. For development and testing environments I'm not seeing problems.

This is my humble opinion, let me know what you guys think.

src/devices/Board/ManagedGpioController.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Reflection;
+using System.Text;
+using System.Threading;
+
+namespace Iot.Device.Board
+{
+    /// <summary>
+    /// A GPIO Controller instance that manages pin usage
+    /// </summary>
+    internal class ManagedGpioController : GpioController
```
krwq, line 19: would it make sense to instead have GpioDriver tied to a board?

src/devices/Board/PwmChannelManager.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Device.Pwm;
+
+namespace Iot.Device.Board
+{
+    internal class PwmChannelManager : PwmChannel
```
krwq, line 9: I'm confused about this class. The name says manager but yet it implements PwmChannel

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,692 @@
...
+                board?.Dispose();
+                board = null;
+            }
+
+            if (board != null)
+            {
+                return board;
+            }
+
+            throw new PlatformNotSupportedException("Could not find a matching board driver for this hardware");
```
joperezr, line 689: Is there a reason why we don't try as well with CustomBoard?

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,692 @@
...
+            {
+                return board;
+            }
+
+            try
+            {
+                board = new GenericBoard(defaultNumberingScheme);
+                board.Initialize();
+            }
+            catch (Exception x) when ((x is NotSupportedException) || (x is IOException))
```
joperezr, line 678: Do you mind adding a comment of when do you expect IOException? I understand the NotSupported but not the IOException.

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,692 @@
...
+            try
+            {
+                board = new RaspberryPiBoard(defaultNumberingScheme);
+                board.Initialize();
+            }
+            catch (Exception x) when ((x is NotSupportedException) || (x is IOException))
+            {
+                board?.Dispose();
+                board = null;
+            }
```
joperezr, line 666: NIT: All this try and catching seems repetitive, we could extract it to a helper that takes in a Type, or perhaps just a board directly

src/devices/Board/GenericBoard.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Device.Pwm;
+using System.Device.Spi;
+using System.Text;
+
+namespace Iot.Device.Board
+{
+    /// <summary>
+    /// A generic board class. Uses generic implementations for GPIO, I2C etc
+    /// </summary>
+    public class GenericBoard : Board
```
joperezr, line 18: Is there a specific reason why this is public? I would rather have this (and the RaspberryPiBoard and CustomBoard) internal, and just have the Create method on the base class to create the right one for you. We could also even have an overload of Create that takes in an Enum so that we allow callers to specify which board they want to create. In the past on dotnet/iot we have found that using factory pattern is  better (like in SPI and I2C) for future refactorings since it was way easier to adapt and develop that when compared to GPIO (which doesn't have the factory pattern)

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,692 @@
...
+        /// <summary>
+        /// True if the board instance is initialized
+        /// </summary>
+        protected bool Initialized => _initialized;
+
+        /// <summary>
+        /// True if this instance is disposed.
+        /// Any attempt to use it after this becomes true results in undefined behavior.
+        /// </summary>
+        protected bool Disposed
```
joperezr, line 67: NIT:
`protected bool Disposed => _disposed;`

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,692 @@
...
+            get
+            {
+                return _disposed;
+            }
+        }
+
+        /// <summary>
+        /// The default pin numbering scheme
+        /// </summary>
+        public PinNumberingScheme DefaultPinNumberingScheme
```
joperezr, line 78: NIT: Same here, you can use the compact return;

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,692 @@
...
+        }
+
+        /// <summary>
+        /// Converts pin numbers from one numbering scheme to another.
+        /// </summary>
+        /// <param name="pinNumber">Pin number to convert</param>
+        /// <param name="inputScheme">The numbering scheme of the input pin</param>
+        /// <param name="outputScheme">The desired numbering scheme</param>
+        /// <returns>The converted pin number</returns>
+        public abstract int ConvertPinNumber(int pinNumber, PinNumberingScheme inputScheme, PinNumberingScheme outputScheme);
```
joperezr, line 93: I think I agree with Krzysztof here, we've learned that using PinNumberingScheme is more confusing that not, so I would just take an optional parameter in the Create factory method, and just assume that all ints are in that numbering scheme going forward.

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,692 @@
...
+        /// <param name="owner">Class that owns the pin (use "this")</param>
+        /// <exception cref="InvalidOperationException">The pin is already reserved</exception>
+        public virtual void ReservePin(int pinNumber, PinUsage usage, object owner)
+        {
+            if (!_initialized)
+            {
+                Initialize();
+            }
+
+            int logicalPin = ConvertPinNumber(pinNumber, DefaultPinNumberingScheme, PinNumberingScheme.Logical);
```
joperezr, line 110: I would just assume that this is in _defaultNumberingScheme.

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,692 @@
...
+
+            lock (_pinReservationsLock)
```
joperezr, line 112: Is it possible to avoid this lock and instead just use ConcurrentDictionary for the pinReservations?

src/devices/Board/Board.cs
```diff
@@ -0,0 +1,692 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Returns the current usage of a pin
+        /// </summary>
+        /// <param name="pinNumber">Pin number in the current numbering scheme</param>
+        /// <returns>The current usage of a pin</returns>
+        public abstract PinUsage DetermineCurrentPinUsage(int pinNumber);
```
joperezr, line 212: Why is this abstract? can't you just check the pinReservations list?

src/devices/Board/Board.cs
pgrawehr: Because the CustomBoard _requires_ custom drivers. It's "default" implementation would be equivalent to the GenericBoard.

src/devices/Board/Board.cs
pgrawehr: Cleaned it up. This is actually never throwing an exception in the current implementation

src/devices/Board/GenericBoard.cs
pgrawehr: This is the one you would use if you can't use or don't want to use the RaspberryPiBoard. It uses all the default implementations.

src/devices/Board/Board.cs
pgrawehr: With the removal of the numbering scheme support, this went away, too.

src/devices/Board/Board.cs
pgrawehr: The logic is not straight-forward, and it's only used when opening/closing pins, therefore performance shouldn't be an issue.

src/devices/Board/Board.cs
pgrawehr: This uses a separate list (moved it here now). The reason this is abstract (or virtual) is that an implementation is suggested to query the hardware, if possible. This allows determining the pin usage without changing it first.

pgrawehr:
@joperezr I think I have addressed the open issues. Should be ready to go now.

joperezr (approved):
Thank you so much for driving this through @pgrawehr I know how hard it has been and it is a long journey, but I couldn't be more happy to merge this in 😄 Thanks for the great contribution!

PR #1129, opened 2020-07-04
New character map for SPLC780 chip

pgrawehr:
By chance stepped over a display that had this new character map. Other than the different character map, the display is compatible to the HD4470 chips.

### DIFF
--- a/src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
+++ b/src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
@@ -13,8 +13,11 @@ namespace Iot.Device.CharacterLcd
     /// </summary>
     public class LcdCharacterEncodingFactory
     {
+        // Default maps for the HD44780 controller
         private static readonly Dictionary<char, byte> DefaultA00Map;
         private static readonly Dictionary<char, byte> DefaultA02Map;
+        // This map is used on SPLC780 controllers, it seems. They're otherwise compatible to the HD44780.
+        private static readonly Dictionary<char, byte> DefaultSplC780Map;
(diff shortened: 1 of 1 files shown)


src/devices/CharacterLcd/LcdCharacterEncodingFactory.cs
```diff
@@ -31,12 +34,16 @@ static LcdCharacterEncodingFactory()
...
+            // This map for instance can be found here: https://www.microchip.com/forums/m977852.aspx
+            DefaultSplC780Map = new Dictionary<char, byte>();
+
             // Inserts ASCII characters ' ' to 'z', which are common to most known character sets
             for (char c = ' '; c <= 'z'; c++)
             {
                 DefaultCustomMap.Add(c, (byte)c);
                 DefaultA00Map.Add(c, (byte)c);
                 DefaultA02Map.Add(c, (byte)c);
+                DefaultSplC780Map.Add(c, (byte)c);
```
lindexi, line 46: Why not deine the `c` as byte?

pgrawehr: Well, `byte c= ' '` doesn't work either, so this would just move the cast to the for loop. Not much better IMHO.

Ellerbach:
Cool :-) One more display :-D Never found this one! Smart integration!

pgrawehr:
Not sure how I would order a specific one (or make sure I get several of the same kind for a particular product)... This one was part of an Arduino "advent calendar" box.

pgrawehr:
@krwq : this should be very quick to review, since it only extends existing functionalities - no new API at all

PR #1130, opened 2020-07-06
Add a generic GPIO driver for Allwinner SoCs and Orange Pi Zero/Lite/Lite2 pin map

ZhangGaoxing:


ZhangGaoxing:
I moved the code to `Iot.Device.Gpio`, but how to reference https://github.com/dotnet/iot/tree/master/src/System.Device.Gpio/Interop/Unix and System.Device.Gpio 😅
@krwq @joperezr

joperezr:
Not sure what you mean on how to reference System.Device.Gpio or why you would need that, but referencing native PInvokes should be easy by just making sure that your project also compiles those interop files. They don't expose any surface area and all the purpose they have is to be able to PInvoke to native libraries that S.D.Gpio already depends on so I don't see harm on doing that. Of course, it will be important for you to think of what to do in the Windows case.

ZhangGaoxing:
@krwq @joperezr

ZhangGaoxing:
Is anybody there? @krwq @joperezr

joperezr:
Sorry @ZhangGaoxing for some reason this PR fell through the cracks and we missed it. Will review now.

joperezr:
Can we change the project name? I would make this be two bindings, one would be the OrangePi, and the other one would be Sunxi one. Each one with its own namespace and samples.

src/devices/Gpio/Iot.Device.Gpio.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="../../System.Device.Gpio/**/*.cs" />
```
joperezr, line 8: Why do you need to compile the source code of the library as opposed to just reference the library? This is not right since it will expose the surface area of System.Device.Gpio on Iot.Device.Bindings which we definitely don't want as consumers of this library will have conflicts as types will be exposed in two places.

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,469 @@
...
+        public SunxiDriver(int cpuxPortBaseAddess, int cpusPortBaseAddess)
+        {
+            CpuxPortBaseAddess = cpuxPortBaseAddess;
+            CpusPortBaseAddess = cpusPortBaseAddess;
+        }
+
+        /// <summary>
+        /// The number of pins provided by the driver.
+        /// </summary>
+        protected internal override int PinCount => throw new PlatformNotSupportedException("This driver is generic so it can not enumerate how many pins are available.");
```
joperezr, line 62: Once you change to start referencing System.Device.Gpio as opposed to compilingt he source as you are doing now, you will have to change all of these abstract method signatures to instead be `protected override ...`. So dropping the internal given that via a reference you won't be seeing that any longer.

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,469 @@
...
+        /// <summary>
+        /// CPUS-PORT base address.
+        /// </summary>
+        protected virtual int CpusPortBaseAddess { get; }
+        // final_address = mapped_address + (target_address & map_mask) https://stackoverflow.com/a/37922968
+        private readonly int _mapMask = Environment.SystemPageSize - 1;
+
+        private static readonly object s_initializationLock = new object();
+        private static readonly object s_sysFsInitializationLock = new object();
+        private const string GpioMemoryFilePath = "/dev/mem";
```
joperezr, line 42: Constants should be declared at the beginning of the class. Your build is getting a bunch of errors because of code style issues like this, I won't comment on all but all should be addressed before we can merge this.

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,469 @@
...
+        private UnixDriver _interruptDriver;
+        private readonly IDictionary<int, PinState> _pinModes = new Dictionary<int, PinState>();
```
joperezr, line 44: Same here, readonly members need to appear before non-readonly

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,469 @@
...
+        /// Closes an open pin.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        protected internal override void ClosePin(int pinNumber)
+        {
+            if (_pinModes.ContainsKey(pinNumber))
+            {
+                if (_pinModes[pinNumber].InUseByInterruptDriver)
+                {
+                    _interruptDriver.ClosePin(pinNumber);
```
joperezr, line 91: I wonder if we should make the SunxiDriver derive from UnixDriver instead that way it will be able to call these methods.

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,469 @@
...
+            }
+
+            return _pinModes[pinNumber].CurrentPinMode;
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (_gpioPointer0 != default)
+            {
+                Interop.munmap(_gpioPointer0, 0);
```
joperezr, line 341: In order to call these methods you will have to compile the Interop files that declares them.

joperezr:
Left some initial feedback. Once that is addressed I'm happy to take another look and do let me know if you need help developing this. BTW, if we find that it is really hard to develop a Gpio driver outside of System.Device.Gpio, we can discuss about moving this OrangePi drivers into System.Device.Gpio which might make things easier.

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
ZhangGaoxing: `UnixDriver` is an abstract class🤔

ZhangGaoxing: If I use `ProjectReference`, how to solve `CS1540` error😅 @joperezr

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+        protected SunxiDriver()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SunxiDriver"/>.
+        /// </summary>
+        /// <param name="cpuxPortBaseAddess">CPUX-PORT base address (This can be find in the corresponding SoC datasheet).</param>
+        /// <param name="cpusPortBaseAddess">CPUS-PORT base address (This can be find in the corresponding SoC datasheet).</param>
+        public SunxiDriver(int cpuxPortBaseAddess, int cpusPortBaseAddess)
```
krwq, line 59: should the addresses be unsigned or just IntPtr?

krwq, line 59:
```suggestion
        public SunxiDriver(int cpuxPortBaseAddress, int cpusPortBaseAddress)
```

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+        {
+            CpuxPortBaseAddess = cpuxPortBaseAddess;
```
krwq, line 61: property name also has typo

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+            SetPinMode(pinNumber, PinMode.Input);
+        }
+
+        /// <summary>
+        /// Closes an open pin.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        protected override void ClosePin(int pinNumber)
+        {
+            if (_pinModes.ContainsKey(pinNumber))
```
krwq, line 93: should you throw when it's already in use by interrupts? is it safe to use it like that?

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+            int cfgNum = unmapped.Port / 8;
+            int cfgShift = unmapped.Port % 8;
+            int pulNum = unmapped.Port / 16;
+            int pulShift = unmapped.Port % 16;
+
+            // Get register address, register pointer
+            int cfgAddress, pulAddress;
+            uint* cfgPointer, pulPointer;
+
+            cfgAddress = (CpuxPortBaseAddess + unmapped.PortController * 0x24 + cfgNum * 0x04) & _mapMask;
```
krwq, line 133: can you please point to spec

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+        }
+
+        /// <summary>
+        /// Removes a handler for a pin value changed event.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        /// <param name="callback">Delegate that defines the structure for callbacks when a pin value changed event occurs.</param>
+        protected override void RemoveCallbackForPinValueChangedEvent(int pinNumber, PinChangeEventHandler callback)
+        {
+            _interruptDriver.OpenPin(pinNumber);
```
krwq, line 272: why do we need to open pin in order to remove the callback?

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+        /// <summary>
+        /// Blocks execution until an event of type eventType is received or a cancellation is requested.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        /// <param name="eventTypes">The event types to wait for.</param>
+        /// <param name="cancellationToken">The cancellation token of when the operation should stop waiting for an event.</param>
+        /// <returns>A structure that contains the result of the waiting operation.</returns>
+        protected override WaitForEventResult WaitForEvent(int pinNumber, PinEventTypes eventTypes, CancellationToken cancellationToken)
+        {
+            _interruptDriver.OpenPin(pinNumber);
```
krwq, line 287: should we throw if pin is already in use? (or will OpenPin?)

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+            }
+            catch (PlatformNotSupportedException)
+            {
+                _interruptDriver = new InterruptSysFsDriver(this);
+            }
+        }
+
+        private void Initialize()
+        {
+            if (_gpioPointer0 != default)
```
krwq, line 379: can you do IntPtr.Zero/0 instead of `default`?

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+
+                int fileDescriptor = Interop.open(GpioMemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
+                if (fileDescriptor == -1)
+                {
+                    throw new IOException($"Error {Marshal.GetLastWin32Error()} initializing the Gpio driver.");
+                }
+
+                IntPtr mapPointer0 = Interop.mmap(IntPtr.Zero, Environment.SystemPageSize - 1, (MemoryMappedProtections.PROT_READ | MemoryMappedProtections.PROT_WRITE), MemoryMappedFlags.MAP_SHARED, fileDescriptor, CpuxPortBaseAddess & ~_mapMask);
+                IntPtr mapPointer1 = Interop.mmap(IntPtr.Zero, Environment.SystemPageSize - 1, (MemoryMappedProtections.PROT_READ | MemoryMappedProtections.PROT_WRITE), MemoryMappedFlags.MAP_SHARED, fileDescriptor, CpusPortBaseAddess & ~_mapMask);
+                if (mapPointer0.ToInt64() == -1 || mapPointer1.ToInt64() == -1)
```
krwq, line 399: should you unmap in here if one of them is -1 and the other one isn't?

krwq: (might be a good idea to create SafeMappedMemoryHandle or something to take care of this automatically)

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+            int portController = (pinNumber - port) / 32;
+
+            return (portController, port);
+        }
+
+        private static int MapPortController(char portController)
+        {
+            return portController switch
+            {
+                'A' => 0,
```
krwq, line 438: you can simplify this:

```csharp
return (portController >= 'A' && portController <= 'M') ? (int)(portController - 'A') : throw new Exception();
```

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+
+        private class PinState
+        {
+            public PinState(PinMode currentMode)
+            {
+                CurrentPinMode = currentMode;
+                InUseByInterruptDriver = false;
+            }
+
+            public PinMode CurrentPinMode
```
krwq, line 463: please put it in one line (suggestion won't remove old lines)
```suggestion
            public PinMode CurrentPinMode { get; set; }
```

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,476 @@
...
+                _interruptDriver.Dispose();
+                _interruptDriver = default;
+            }
+        }
+
+        private void InitializeInterruptDriver()
+        {
+            try
+            {
+                _interruptDriver = new LibGpiodDriver(0);
```
krwq, line 369: I wish we had some better way of doing this.. might be worth filing issue.. we should instead have something like `GetBestDriverForBoard()` or something along the lines

src/devices/Gpio/Drivers/OrangePiZeroDriver.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// <summary>
+        /// Converts a board pin number to the driver's logical numbering scheme.
+        /// </summary>
+        /// <param name="pinNumber">The board pin number to convert.</param>
+        /// <returns>The pin number in the driver's logical numbering scheme.</returns>
+        protected override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber)
+        {
+            return pinNumber switch
+            {
+                3 => MapPinNumber('A', 12),
```
krwq, line 31: (optional) you might instead want to create array with 27 elements and map it to number or -1 if not present... switch will check it one by one and be relatively slow.

src/devices/Gpio/samples/Program.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
```
krwq, line 1: copyright header missing

src/devices/Gpio/samples/Program.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Threading;
+
+namespace AllwinnerGpioDriver.Samples
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            int pinNum = 7;
+            using GpioController gpio = new GpioController(PinNumberingScheme.Board, new OrangePiZeroDriver());
```
krwq, line 13: Board? wasn't this Physical/Logical before?

src/devices/Gpio/Drivers/OrangePiLite2Driver.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Iot.Device.Gpio.Drivers
+{
+    /// <summary>
+    /// A GPIO driver for the Orange Pi Lite 2.
+    /// </summary>
+    /// <remarks>
+    /// SoC: Allwinner H6 (sun50iw6p1)
+    /// </remarks>
+    public class OrangePiLite2Driver : Sun50iw6p1Driver
```
krwq, line 15: IMO those drivers should be internal and we should change detection logic, on the other hand we did other drivers public. I feel like this library will grow in size indefinitely if we keep going like that. Perhaps we should split the drivers into separate package or something to only pull what's needed on the given setup? cc: @Ellerbach @joperezr

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
ZhangGaoxing: I referred to `RaspberryPi3LinuxDriver.cs` before coding😅https://github.com/dotnet/iot/blob/ddeecef97ebf6eb2901b169c46a02af47bd234eb/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs#L200

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
ZhangGaoxing: This also needs to be changed...... https://github.com/dotnet/iot/blob/ddeecef97ebf6eb2901b169c46a02af47bd234eb/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs#L641

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
ZhangGaoxing: Actually `LinuxDriver.Create()` can solve this problem. But before solving this problem, I think we should solve `error CS1540` first. I can't access `LinuxDriver` methods.

ZhangGaoxing:
Anyone to help me with this?😅 @joperezr @krwq

joperezr:
Hey @ZhangGaoxing I'll take a look and see why this is not building and help out early next week.

joperezr:
I took a look today and see what the problem is with the SunxiDriver and the way you try to call protected internal methods via the _interruptdriver. While I understand that the extra level of inderection is not ideal, would it be possible to instead switch all the places where you use the interruptDriver so that it instead goes through a controller? With this I mean changing:

```diff
        private void InitializeInterruptDriver()
        {
-            try
-            {
-                _interruptDriver = new LibGpiodDriver(0);
-            }
-            catch (PlatformNotSupportedException)
-            {
-                _interruptDriver = new InterruptSysFsDriver(this);
-            }
+          _interruptController = new GpioController();
        }
```

That way you will automatically get the right driver picked for you depending on whether or not libgpiod is installed, and you will be able to call all Pin and interrupt operations via the controller as all of that surface is public. I realize that this will introduce an extra layer of indirection, but it seems like the least complex way of working around the current design. Does that sound like a good plan to you?

joperezr:
> <Project Sdk="Microsoft.NET.Sdk">

Also, this project should just be removed. If you want to add individual projects for developing the Gpio drivers that is ok, but it shouldn't be compiling System.Device.Gpio code directly, and instead should just reference the project like the rest of bindings do. That way, we will ensure that this  individual projects work the same as the Iot.Device.Bindings package.

---
Refers to: src/devices/Gpio/Iot.Device.Gpio.csproj:1 in 61152e1. [](commit_id = 61152e1e34c2ae08b53f3b874df1e6f40c237957, deletion_comment = False)

src/devices/Gpio/samples/Program.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System;
+using Iot.Device.Gpio.Drivers;
+using System.Device.Gpio;
+using System.Threading;
+
+namespace AllwinnerGpioDriver.Samples
```
joperezr, line 9: Please create a separate project for the samples.

ZhangGaoxing:
@joperezr @krwq If I use `GpioController` to instead `UnixDriver`, some methods like `AddCallbackForPinValueChangedEvent()` become inaccessible. I don't think the existing design can add additional GPIO drivers externally. 🙁

joperezr:
> ome methods like AddCallbackForPinValueChangedEvent() become inaccessible.

[Triage] This is accessible just with a different name. Use RegisterCallbackForPinValueChangedEvent instead.

> I don't think the existing design can add additional GPIO drivers externally

[Triage] We are aware that there are some difficulties with this and we are interested in seeing if using GpioController would make it work in your case. We could also consider making GpioController un-sealed. Is there any other API that you don't have access to via GpioController in order to finish this?

joperezr:
[Triage] We have a PR up that is unsealing GpioController which might impact this work in the sense that you could just subclass it for Orange Pi, so that could be another option too. Also keep in mind that we recently enabled nullability on the whole repo, so this might require some following up with code in your PR. Are you still interested in working on this? If not, let us know so that we can close the PR in our next triage meeting.

ZhangGaoxing:
I have other things to do, but I will continue with this work. I'll take a closer look at it at the weekend.

ZhangGaoxing:
I removed the use of `GpioController` to implement pin value interrupts, because the internal `GpioController` will cause repeated opening of pins. I can't call methods in `UnixDriver`, so I inherit `SysFsDriver` directly. So far, I have tested on Orange Pi Zero using Armbian and found no problems. I'm going to add README this weekend.

ZhangGaoxing:
@joperezr

src/devices/Gpio/Drivers/OrangePiLite2Driver.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        private readonly int[] _pinNumberConverter = new int[27]
```
krwq, line 17: should this be static? I assume this won't change

ZhangGaoxing:
@joperezr @krwq

joperezr:
[Triage] this is on our radar of PRs to take a look by next week.

microhobby:
@ZhangGaoxing Sorry if I don't understand your goal here (I didn't go into your commits), but isn't the libpgiod API generic enough? Please if possible let me know what we have here that libgpiod can't do.

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,468 @@
...
+        /// <returns>The pin number in the driver's logical numbering scheme.</returns>
+        protected override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber) => throw new PlatformNotSupportedException("This driver is generic so it can not perform conversions between pin numbering schemes.");
+
+        /// <summary>
+        /// Opens a pin in order for it to be ready to use.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        protected override void OpenPin(int pinNumber)
+        {
+            Initialize();
```
joperezr, line 83: I think initialize should be called from the constructor instead because you can't prevent from folks calling ClosePin before calling OpenPin. Also seems like any operation will require Initialize to have happened, so no need to delay it to the first OpenPin call.

ZhangGaoxing:
@joperezr @krwq Does anyone come to review this PR? After this PR is merged, I can continue to add the GPIO driver of Rockchip. 😄 https://github.com/ZhangGaoxing/rockchip-gpio-driver

src/devices/Gpio/Drivers/OrangePiLite2Driver.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
krwq, line 3: You will need to do 2 line copyright headers like i.e.:

https://github.com/dotnet/iot/blob/master/src/devices/Ak8963/Ak8963I2c.cs#L1-L2

src/devices/Gpio/Drivers/OrangePiLite2Driver.cs
```diff
@@ -0,0 +1,37 @@
...
+        private static readonly int[] _pinNumberConverter = new int[]
+        {
+            -1, -1, -1, MapPinNumber('H', 6), -1, MapPinNumber('H', 5), -1, MapPinNumber('H', 4), MapPinNumber('D', 21), -1,
+            MapPinNumber('D', 22), MapPinNumber('D', 24), MapPinNumber('C', 9), MapPinNumber('D', 23), -1, MapPinNumber('D', 26),
+            MapPinNumber('C', 8), -1, MapPinNumber('C', 7), MapPinNumber('C', 2), -1, MapPinNumber('C', 3), MapPinNumber('D', 25),
+            MapPinNumber('C', 0), MapPinNumber('C', 5), -1, MapPinNumber('H', 3)
+        };
+
+        /// <inheritdoc/>
+        protected override int PinCount => _pinNumberConverter.Count(n => n != -1);
```
krwq, line 27: consider caching this value

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
```diff
@@ -0,0 +1,387 @@
...
+                    pulValue |= 0b01U << (pulShift * 2);
+                    break;
+                default:
+                    throw new ArgumentException("Unsupported pin mode.");
+            }
+
+            *cfgPointer = cfgValue;
+            *pulPointer = pulValue;
+
+            if (_pinModes.ContainsKey(pinNumber))
```
krwq, line 169: consider using TryGetValue

krwq (approved):
LGTM after fixing copyright headers, remainder is optional feedback

src/devices/Gpio/Drivers/OrangePiLite2Driver.cs
ZhangGaoxing: _pinNumberConverter will never be null here, I think there should be no exceptions thrown here.

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
ZhangGaoxing: To eliminate this warning, I need extra steps to deal with the situation where pinState is null (although it cannot be null). 😅
```C#
//  VS warns me that pinState may be null
if (_pinModes.TryGetValue(pinNumber, out PinState pinState))
{
}
```

src/devices/Gpio/Drivers/OrangePiLite2Driver.cs
krwq: that's not what I meant, I meant that you need to iterate through the list every time you want to check the PinCount and this value is actually constant (`=>` means this will compute every time you access the property, `{ get; } = ...;` would compute it only once)

src/devices/Gpio/README.md
```diff
@@ -0,0 +1,79 @@
...
+**sunxi** represents the family of ARM SoCs from Allwinner Technology. This project contains a **full function(PULL-UP, PULL-DOWN)** generic GPIO driver `SunxiDriver` for Allwinner SoCs and some special GPIO drivers like `OrangePiZeroDriver`, `OrangePiLiteDriver`, `OrangePiLite2Driver`.
+
+## Getting started
+
+### Generic GPIO driver: `SunxiDriver`
+
+```C#
+// Beacuse this is a generic driver, the pin scheme can only be Logical.
+// The base addresses can be found in the corresponding SoC datasheet.
+using GpioController gpio = new GpioController(PinNumberingScheme.Logical, new SunxiDriver(cpuxPortBaseAddress: 0x01C20800, cpusPortBaseAddress: 0x01F02C00));
```
krwq, line 12: I think the sample shouldn't start with the most complex option, start with options where passing in the address is not necessary

src/devices/Gpio/Drivers/Sunxi/SunxiDriver.cs
krwq: I'm not sure what you mean but fine as is

krwq (approved):
2 recommendations left, let me know if you prefer to merge as is.

src/devices/Gpio/Drivers/OrangePiLite2Driver.cs
ZhangGaoxing: Haha, OK, I see what you mean. Actually, I am too lazy to count the number of pins. I'll change it.

microhobby:
Just commenting that I don't agree with this. It is not the correct way to go on adding all possible GPIO implementations for each existing SoC vendor messing up with `/dev/mem`, since we already have an implementation that covers all SoC generically using the Linux Kernel interfaces, which use the hardware resources and exceptions in the best possible way.

@ZhangGaoxing don't get me wrong, nothing personal.

ZhangGaoxing:
@microhobby  Yes, what you said makes sense. But I want to ask a question directly. How do I operate a pin that is set to non-GPIO mode? How do I use what you call Linux Kernel interfaces to pull up a GPIO pin? In fact, most of the GPIO operations can be done using SysFsDriver, but this PR is just to implement some special operations of GPIO. According to your point of view, actually there is no need for WiringPi. 😄

microhobby:
@ZhangGaoxing yes, in my view WiringPi is not needed at all. We already have pull-up, pull-down configurations in GPIOLIB. Pin multiplexing must be done in Kernel space by device trees or overlays. I already mentioned this here: https://github.com/dotnet/iot/pull/1397

And by the way, if you miss something that should be done in Kernel space at GPIOLIB I'm sure your technical knowledge will be welcome in the Linux Kernel community. When this is added to the Kernel pinctrl subsystem there is a whole mechanism that deals generically, and handles exceptions to a kernel level taking into account the entire system, with all implementations of SoC vendors that kernel has support. IMHO this is the correct way to follow.

ZhangGaoxing:
@microhobby Actually, before you come to this conclusion, you have made the assumption that libgpiod is available on any device. ARM is different from x86, after a product (board) is launched, the Linux kernel version will probably never change again. We can't guarantee that these new technologies can be used in any ARM board. This is why I choose to inherit SysFsDriver to implement the SunxiDriver, because libgpiod is not available on the sunxi boards I tested (for now). Although sysfs is no longer recommended, the popularity of libgpiod may take much longer than we think. Maybe you use more Raspberry Pi, and the maintenance of Raspberry Pi is also more timely, but Raspberry Pi is not the best choice, because there are cheaper and better performance boards, and the number of people using these boards is far more than those using Raspberry Pi. This driver just gives them a choice.

microhobby:
@ZhangGaoxing If you are going to use this because the vendor does not have updated software or does not provide means for you to compile updated software, especially the kernel, it is a mistake to choose this hardware from that vendor. But is not the case of Alwinner SoCs, they are very well maintained and have DTS updated by the community in mainline. And it is not difficult to compile a Kernel and upgrade to a ready to use updated distro that has packages with the latest version of libgpiod, and others by far more critical updated packages. I assure you that it is much less work and time spend than all this work that you make here, it's just a matter of configuration, everything is there. 

But, I really understand your point of view. Users are so used to just using the software that the manufacturer of the development board provides that they forget that first it is a development board for prototyping and second according to the software that is provided generally, most of the time, it was not meant to last a 10-year life cycle product but only as a basis, you should not use it as something reliable for your project, there are a lot of security issues and other details. 

My point is, we should use the interfaces here in the right way and recommend the most reliable way to use the system resources since the beginning, I have seen many "temporary solutions" made during the development go to the field with the promise "we will fix this later", and only be fixed when a issue breaks out. My opinion here is an alert, what I want and I think that all contributors of the project want, it is whoever uses the library to succeed in their projects, apply it in production in final products and earn money with it (why not 🤑). But many projects start out as something more "maker" DIY and it's develops to a final product, so the importance of starting in the "pro" way from the beginning.

Another point, using the interfaces that the updated versions of the Kernel provide also helps us in maintaining the project here. Having different drivers for different platforms for different SoC vendors only adds to the complexity of project maintenance.

To be clear, I really appreciate your work here, my intention is not to discourage you, just giving my humble opinion based on my experience.

Obs.: and I use more boards from the NXP iMX family 😁

joperezr:
@microhobby I understand your concern, and it is very valid. One of the reasons why we are not currently opting for adding all of these other drivers into the main System.Device.Gpio library is because we currently think that the ideal driver for non-Raspberry Pi's should be libgpiod driver. That said, we shouldn't block the addition of faster (even if they are more dangerous) drivers to the bindings library, as the user has to do a conscious decision in order to use them and instantiate this driver and pass it to the controller. We can add a note on the driver's notes and mention that the driver is dealing with internal memory registers, so it is dangerous. That said, we don't currently have that note on our RaspberryPiDriver which is also using registers directly (it will only use `/dev/gpiomem` for most cases, but it could also go into `/dev/mem` as well in some other cases.

joperezr (approved):
Sorry it took so long to get this in. Thank you for all the hard work @ZhangGaoxing!

PR #1131, opened 2020-07-06
Add device binding for STUSB4500

chrfin:
This adds a device binding for the [STUSB4500](https://github.com/usb-c/STUSB4500), an autonomous USB-C PD controller for Power Sinks / UFP.


src/devices/StUsb4500/samples/StUsb4500.Sample.cs
```diff
@@ -0,0 +1,180 @@
...
+using System.Device.Spi;
+using System.Linq;
+using System.Threading;
+using Iot.Device.StUsb4500.Enumerations;
+using Iot.Device.StUsb4500.Objects;
+using UnitsNet;
+
+namespace Iot.Device.StUsb4500.Samples
+{
+    /// <summary>Samples for StUsb4500.</summary>
```
Ellerbach, line 18: Thank you for the very complete example!

src/devices/README.md
```diff
@@ -205,6 +205,10 @@ Our vision: the majority of .NET bindings are written completely in .NET languag
 
 * [SPI, GPIO and I2C drivers for FT4222](Ft4222/README.md)
 
+### USB PD Power Sinks
```
Ellerbach, line 208: this is not necessary. Add a category.txt file to the root of the director with this category and it will be build automatically.

src/devices/Device-Index.md
```diff
@@ -63,6 +63,7 @@
 * [Solomon Systech Ssd1351 - CMOS OLED](Ssd1351/README.md)
 * [SPI, GPIO and I2C drivers for FT4222](Ft4222/README.md)
 * [Still image recording library](Media/README.md)
+* [STUSB4500 - Autonomous USB-C PD controller for Power Sinks / UFP](StUsb4500/README.md)
```
Ellerbach, line 66: This is automatic as well

Ellerbach (changes requested):
Thanks a lot, it looks very good, few adjustments needed. Also please add a proper readme.

src/devices/StUsb4500/Objects/RequestDataObject.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        /// <summary>Gets or sets a value indicating whether the give back flag is set.</summary>
+        public bool GiveBackFlag { get => GetBit(27); set => UpdateBit(27, value); }
+
+        /// <summary>Gets or sets the object position.</summary>
+        /// <remarks>This is stored as a 3-bit value (range 0 - 7). 000b is Reserved and shall not be used.</remarks>
+        public byte ObjectPosition
+        {
+            get => (byte)(Value >> 28 & ObjectPositionMask);
+            set => Value = (Value & ObjectPositionMask) | (uint)((value & ObjectPositionMask) << 28);
```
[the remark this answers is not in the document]
chrfin: As this is a number (it says e.g. "This is PDO #x") I do not think an enum would be better. And the range is limited by the bit-operations in the setter to be within the valid range. Or would you like an exception in case the value is >7?

src/devices/StUsb4500/StUsb4500.cs
```diff
@@ -0,0 +1,559 @@
...
+
+            Span<byte> prtStatusBuffer = stackalloc byte[]
+                {
+                    (byte)StUsb4500Register.PRT_STATUS
+                };
+            Span<byte> rxHeaderBuffer = stackalloc byte[]
+                {
+                    (byte)StUsb4500Register.RX_HEADER_LOW
+                };
+            ////Span<byte> rxByteCntBuffer = stackalloc byte[]
```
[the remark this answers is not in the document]
chrfin: This is left in there on purpose - please see the comment a few lines below where _rxByteCntBuffer_ whould have been used.

src/devices/StUsb4500/StUsb4500.cs
```diff
@@ -0,0 +1,559 @@
...
+            }
+            catch (ArgumentException exception) when (exception.Message.Contains("readBuffer")) // sometimes we get a "readBuffer cannot be empty." exception
+            {
+                if (retryCount <= 0) // avoid endless tries in case of other problems
+                {
+                    throw;
+                }
+
+                Console.WriteLine(exception.Message);
+                return ReadSourcePdo(retryCount - 1); // try again, second try almost always works
```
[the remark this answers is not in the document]
chrfin: You can try again immediately as the function's first command is a reset anyhow.

src/devices/StUsb4500/StUsb4500.cs
```diff
@@ -0,0 +1,559 @@
...
+                _i2cDevice.WriteRead(rxHeaderBuffer, headerBuffer);
+                var messageHeader = new UsbPdMessageHeader((ushort)(headerBuffer[0] + (headerBuffer[1] << 8)));
+                messageType = messageHeader.DataMessageType;
+                objectCount = messageHeader.NumberOfDataObjects;
+            }
+            while (objectCount <= 0 && messageType != UsbPdDataMessageType.USBPD_DATAMSG_Source_Capabilities && !cts.IsCancellationRequested);
+
+            if (cts.IsCancellationRequested)
+            {
+                return new PowerDeliveryObject[0];
```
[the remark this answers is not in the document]
chrfin: Thinking about it again I also think null is better - as null is more like "nothing received at all"...

src/devices/StUsb4500/StUsb4500.cs
```diff
@@ -0,0 +1,559 @@
...
+            }
+
+            // The following check is disabled as it can not really prevent the error mentioned below, as the buffer-change can happen after the check and before the read.
```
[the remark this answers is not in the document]
chrfin: ...as this is based on the official C-code from ST I wanted to keep it in there so other see my rational behind not executing this part from the original code.

src/devices/StUsb4500/Objects/RequestDataObject.cs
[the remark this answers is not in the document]
Ellerbach: ok, then number will be fine. Just make sure you check the range

src/devices/StUsb4500/StUsb4500.cs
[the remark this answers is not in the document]
Ellerbach: I don't think it adds value at all if it should not be called, it should just be removed :-)
If you want to add a comment, then refer to the original source code and says that here, some additional check can be done but they are generating problem and they have been removed.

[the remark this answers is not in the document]
chrfin: ok, thanks - will change it

chrfin:
@Ellerbach Thanks for the review - I updated the code according to your feedback

src/devices/StUsb4500/StUsb4500.cs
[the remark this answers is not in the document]
Ellerbach: Same comment, I really don't see any value to leave something that is not used and won't be used.
Exception would be in a sample file.

src/devices/StUsb4500/Objects/BatteryObject.cs
```diff
@@ -0,0 +1,70 @@
...
+        private const uint MaxVoltageMask = 0b0011_1111_1111_0000_0000_0000_0000_0000;
+
+        /// <summary>Gets or sets the operating power.</summary>
+        /// <remarks>This is stored with the factor 4 as a 10-bit value (range 0 - 1023) => 0...255.75W.</remarks>
+        public Power OperatingPower
+        {
+            get => Power.FromWatts((ushort)(Value & OperatingPowerMask) / 4.0);
+            set
+            {
+                CheckArgumentInRange(value.Value, 255.75M);
```
Ellerbach, line 27: Thanks for adding the test for the arguments. You should use, and that's for everywhere value.TheUnitToTest (replace TheUnitToTest by the real one). value.Value will return the value in which it has been created. So imagine you create in millivolts, you'll et values which are 1000 times higher and it will be rejected.
So same for the next lines, you'll need to get them with the target unit first.
This apply everywhere unfortunately.
See here for Power for example: https://github.com/angularsen/UnitsNet/blob/a7e55169673f68ef25f15e3f0305f23dc894a6f4/UnitsNet/GeneratedCode/Quantities/Power.g.cs#L162

Ellerbach (commented):
Thanks for all the changes. The way to manipulate the UnitsNet has to be adjusted. Otherwise, looks good!

src/devices/StUsb4500/Enumerations/CableConnection.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.StUsb4500.Enumerations
```
krwq, line 5: This should be directly under `Iot.Device.StUsb4500` although consider renaming StUsb4500 to just `Usb` and renaming struct to contain `StUsb4500` in the name.

src/devices/StUsb4500/Objects/ObjectBase.cs
```diff
@@ -0,0 +1,64 @@
...
+
+using System;
+using System.Runtime.CompilerServices;
+
+namespace Iot.Device.StUsb4500.Objects
+{
+    /// <summary>
+    /// Base class for shared object logic.
+    /// </summary>
+    public abstract class ObjectBase
```
krwq, line 13: these should be internal static helper class and not public abstract base class

src/devices/StUsb4500/Objects/ObjectBase.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        /// <summary>Gets or sets the value.</summary>
+        public uint Value { get; protected set; }
```
krwq, line 16: what's the purpose of this value?

src/devices/StUsb4500/Objects/ObjectBase.cs
chrfin: @krwq what would be the benefit of using a helper instead of a base class?

src/devices/StUsb4500/Objects/ObjectBase.cs
chrfin: @krwq  "objects" transfered to/from the StUsb4500 are based on one "value" which is stored in this base-property and can be bitwise-edited via the methods below.

src/devices/StUsb4500/Enumerations/CableConnection.cs
chrfin: I did rename the namespace and the enumeration itself, but I would like to keep the folder, as it improves the overall structure by clearly separating the enums from the "logic" IMO

src/devices/StUsb4500/Objects/ObjectBase.cs
Ellerbach: I think the open question is more: do you need this property to be publicly exposed? Or should it be an internal property only?

Ellerbach (approved):
Thanks, looks good to me now. Thanks for fixing all the units!

src/devices/StUsb4500/Enumerations/UsbCCableConnection.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Usb.Enumerations
```
krwq, line 5: seems the namespace is still Enumerations

src/devices/StUsb4500/Enumerations/UsbPdControlMessageType.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Usb.Enumerations
```
krwq, line 5: ditto

krwq: same for all enums

src/devices/StUsb4500/Enumerations/UsbCCableConnection.cs
chrfin: @krwq I would like to ask why you request this change, as I would like to keep the folder/namespace, as it improves the overall structure by clearly separating the enums from the "logic" IMO?

joperezr:
@chrfin do you plan on addressing @krwq's comments above? We think this PR is close to merging but we would need that feedback addressed. We'll also trigger a new round of the build because we recently enabled nullability on the whole repo which might trigger some warnings that would need resolving in the PR.

chrfin:
@joperezr @krwq I'm actually waiting on his feedback on my question - or at least a "non-negotiable" and than I will fix it immediately...

joperezr:
oh great, sorry we missed that. @krwq do you mind responding?

It does look like the new changes that went in with enabling nullable broke the state of your PR. @chrfin do you need help rebasing your branch against latest master in order to be able to fix the new warnings that are breaking the build? If so let us know and we are happy to help.

chrfin:
@joperezr I did rebase the branch/PR and added some fixes for NRTs - please verify, thanks!

src/devices/StUsb4500/Enumerations/UsbCCableConnection.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
krwq, line 3: we have recently converted to 2 line copyright header (remove third line without modification to the first two)

src/devices/StUsb4500/Enumerations/UsbCCableConnection.cs
krwq: Consistency. No (or at least most) other bindings does that - if we decide on this structure we should do it for all bindings.

src/devices/StUsb4500/Helper/UsbPdObjectHelper.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
```
krwq, line 1: copyright header missing

src/devices/StUsb4500/Helper/UsbPdObjectHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace Iot.Device.Usb.Helper
```
krwq, line 6: same, we only use namespaces when otherwise it would cause collisions or mess (i.e. Spi/I2c code touches similar concepts but differ slightly) - in here it's all about USB and therefore it should be in the same namespace (if we decide to have these helpers also for other bindings then you could consider changing it)

src/devices/StUsb4500/Objects/BatteryObject.cs
```diff
@@ -0,0 +1,70 @@
...
+                Value = (Value & ~MaxVoltageMask) | (Convert.ToUInt32(value.VoltsDc * 20) << 20 & MaxVoltageMask);
+            }
+        }
+
+        /// <summary>Gets the power of this PDO.</summary>
+        public override Power Power => OperatingPower;
+
+        /// <summary>Initializes a new instance of the <see cref="BatteryObject"/> class.</summary>
+        /// <param name="value">The value.</param>
+        public BatteryObject(uint value)
```
krwq, line 61: can you either rename the value into something meaningful or explain what value is in the xml docs?

krwq: Also since this is representation of some specific encoding could you put the name of that encoding in the name of the class (BatteryXyz, "Object" is rather meaningless - it could be suffixed on any class)

src/devices/StUsb4500/Objects/FixedSupplyObject.cs
```diff
@@ -0,0 +1,92 @@
...
+            get => Value.GetBit(29);
+            set => Value = Value.UpdateBit(29, value);
+        }
+
+        /// <summary>Gets the power of this PDO.</summary>
+        public override Power Power => Power.FromWatts(Voltage.VoltsDc * OperationalCurrent.Amperes);
+
+        /// <summary>Initializes a new instance of the <see cref="FixedSupplyObject"/> class.</summary>
+        /// <param name="value">The value.</param>
+        public FixedSupplyObject(uint value)
```
krwq, line 83: similar comment as on BatteryObject: rename, document/rename value

src/devices/StUsb4500/Objects/PowerDeliveryObject.cs
```diff
@@ -0,0 +1,44 @@
...
+        public abstract Power Power { get; }
+
+        /// <summary>Initializes a new instance of the <see cref="PowerDeliveryObject"/> class.</summary>
+        /// <param name="value">The value.</param>
+        protected PowerDeliveryObject(uint value) => Value = value;
+
+        /// <summary>Creates a new PDO from the given value.</summary>
+        /// <param name="value">The value.</param>
+        /// <returns>A PDO of the type defined by the value.</returns>
+        public static PowerDeliveryObject CreateFromValue(uint value)
```
krwq, line 28: 1. same comment about name and value documentation
2. you should be consistent: either go with constructor everywhere or Create static method everywhere, if this is an implementation detail it should be protected or internal

krwq (commented):
General comment:
- please avoid "Object" suffix in the name of the class, please put something related to the class i.e. what encoding it's using
- for constructors taking "value" please describe what that value is and where to take it from
- think about how users should use the classes - should they check type themselves?
- please update copyright headers, each file should have exactly two lines (we have switched from 3 to 2 lines recently)
- please rebase - we have enabled nullable reference types across the repo, you'll need to adjust the code (although I expect it will mostly stay the same)

src/devices/StUsb4500/samples/StUsb4500.Sample.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// <summary>Updates the local PDOs.</summary>
+        /// <param name="stUsb">The device.</param>
+        private static void UpdateLocalPdo(StUsb4500 stUsb)
+        {
+            Console.WriteLine($"Update Local sink PDOs...{Environment.NewLine}");
+            PowerDeliveryObject[] sinkPdos = stUsb.SinkPowerDeliveryObjects;
+            if (sinkPdos.Length > 1 && sinkPdos[1] is FixedSupplyObject fixedSupplyObject1)
+            {
+                fixedSupplyObject1.Voltage = ElectricPotentialDc.FromVoltsDc(12);
+                fixedSupplyObject1.OperationalCurrent = ElectricCurrent.FromAmperes(2.0);
```
krwq, line 148: if OperationalCurrent is available on all or most of the implementations it should be on the base class to avoid having to repeat the code like here. If it's not available on some of the implementations you should just add property: SupportsOperationallCurrent and check for it in here. Only when it's available on a single implementation is when it should not be on the base class

krwq: (similar comment for other properties)

src/devices/StUsb4500/samples/StUsb4500.Sample.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+
+            Console.WriteLine($"P(max) = {maxPower:0.##}{Environment.NewLine}");
+        }
+
+        /// <summary>Prints the RDO.</summary>
+        /// <param name="stUsb">The device.</param>
+        private static void PrintRdo(StUsb4500 stUsb)
+        {
+            Console.WriteLine("RDO (negotiated power):");
```
krwq, line 127: please avoid using ToString in the sample, most likely users will want to know how to access the properties and printing them while informational is not useful in the real app. If the code seems repetitive when accessing properties you should consider changing APIs so it's not.

src/devices/StUsb4500/Objects/BatteryObject.cs
chrfin: @krwq I did update the documentation to make it as clear as I can think of including a note to check the specification where this value is described in detail.

src/devices/StUsb4500/Objects/PowerDeliveryObject.cs
chrfin: @krwq @2) The `CreateFromValue` is a helper method if the user does not know the type and then yes, they should check the returned type. If the user knows the type he can directly create the correct type. If this can be made anyway better I'm open for suggestions. Sadly the three possible return types do not share any their properties in all three types...

src/devices/StUsb4500/samples/StUsb4500.Sample.cs
chrfin: @krwq This property is available in 2 of 3 possibilities. And 2 other properties are available in other 2 of the three. I'm not sure what way around is better, but to avoid mistakingly choosing the wrong one I tend to the current solution. Or should I change it to the suggested `Supports*` approach and throw `NotSupportedException` if the wrong one is used?

src/devices/StUsb4500/samples/StUsb4500.Sample.cs
chrfin: @krwq I need to disagree as these values are useful to be printed in a real app - this is exactly what we do and is the reason I created this device binding. We show some of these values in our web-interface to inform the user of our device if a valid USB-PD source is connected (as USB-Pd is only one of several ways to power our device). It is also a copy of the original example from the official STUSB4500 repo (https://github.com/usb-c/STUSB4500/tree/master/Firmware).

chrfin:
> General comment:
> 
> *   please avoid "Object" suffix in the name of the class, please put something related to the class i.e. what encoding it's using
> *   for constructors taking "value" please describe what that value is and where to take it from
> *   think about how users should use the classes - should they check type themselves?
> *   please update copyright headers, each file should have exactly two lines (we have switched from 3 to 2 lines recently)
> *   please rebase - we have enabled nullable reference types across the repo, you'll need to adjust the code (although I expect it will mostly stay the same)

1.  The "Object" is not in the sense of a ".NET object" - it is the actual name of these given by the USB-PD specification.
2.  Documentation and name of "value" should be better now - please let me know if I can improve more
3.  Please see my comment above - maybe consolidate everything in the base-class with a "Type" property and `Supports*` flags and throwing `NotSupportedExceptions` in case of invalid usage?
4.  Headers are updated :-)
5.  Rebase is also already done and nullability should be working now

src/devices/StUsb4500/samples/StUsb4500.Sample.cs
```diff
@@ -0,0 +1,177 @@
...
+using System;
+using System.Collections.Generic;
+using System.Device.Gpio;
+using System.Device.I2c;
+using System.Device.Spi;
+using System.Linq;
+using System.Threading;
+using UnitsNet;
+
+namespace Iot.Device.Usb.Samples
```
Ellerbach, line 13: Now we've moved to C# 9, maybe you can do a more modern example by removing the namespace, the class and the main function?

Ellerbach (approved):
Looks good to me. I let the last comments to be resolved by @krwq and you

src/devices/StUsb4500/samples/StUsb4500.Sample.cs
krwq: fine as is

krwq (approved):
LGTM, Thanks!

chrfin:
Great thanks - I will update the sample (possibly on monday - in a training this week) and then resolve the last comment above :-)

chrfin:
@Ellerbach Sample is updated now :-)

PR #1134, opened 2020-07-12
Add device binding for uFire_ISE

R10Rasmus:
This adds a device binding for the [u-fire Isolated ISE Probe Interface](https://github.com/u-fire/Isolated_ISE/) , an I2C slave device that can interface with any voltage based sensor such as a pH or ORP probe.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1134) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1134) before we can accept your contribution.<br/><br/>:x: R10Rasmus [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1134)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1134) it.</sub>

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+        public UFireIse(I2cDevice i2cDevice)
+        {
+            _device = i2cDevice;
+        }
+
+        /// <summary>
+        /// Measure mV
+        /// </summary>
+        /// <returns>mV read from ISE Probe Interface</returns>
+        public ElectricPotential MeasuremV()
```
Ellerbach, line 66: Now you have a UnitsNet, the name should be changed to remove the unit. So ```Measure``` or ```ProbePotential``` or anything like this
Also, this is valid for all variables and properties like this in all the PR.
I understand that because all probes are different, the range will be different. That said, for the comment, would be good to explain if there is any typical range. And adjust the comment to say that the best way to use the measurement is in millivolt. So adjust comment with something like:
/// Measure the ISE Probe Interface, typical measure are in the millivolt range

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+using UnitsNet;
+
+namespace Iot.Device.UFire
+{
+    /// <summary>
+    /// μFire ISE (Ion Specific Electrode) Probe Interface controller
+    /// </summary>
+    public class UFireIse : IDisposable
+    {
+        private const int ISE_TEMP_MEASURE_TIME = 750;
```
Ellerbach, line 17: If those constants are part of the documentation, please mention it and the reference page.
For the naming convention, it is not supposed to be in capital letters except for registers and constants which are part of the documentation.
So if they are not part of the documentation, please consider renaming them to be aligned with the main naming convention so like:
```csharp
private const int IseTemperatureMeasureTime = 750;
```

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+            _mV = Read_register((byte)Register.ISE_MV_REGISTER);
+            _mV = Convert.ToSingle(Math.Round(_mV, 2));
+
+            return new ElectricPotential(_mV, UnitsNet.Units.ElectricPotentialUnit.Millivolt);
+        }
+
+        /// <summary>
+        /// Measure temperature
+        /// </summary>
+        /// <returns>Temperature. Temperature range: -2 to 35 C. Temperature Precision: 0.05 C</returns>
```
Ellerbach, line 81: This is great way for describing what is returned!

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+            _temp = new Temperature(Read_register((byte)Register.ISE_TEMP_REGISTER), UnitsNet.Units.TemperatureUnit.DegreeCelsius);
+
+            return _temp;
+        }
+
+        /// <summary>
+        /// Set temperature
+        /// </summary>
+        /// <param name="temp">Temperature</param>
+        public void SetTemp(Temperature temp)
```
Ellerbach, line 96: If this is to compensate the temperature or to correct the temperature, then consider using a function with a more explicit name: SetTemperatureCompensation

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+            _mV = Convert.ToSingle(Math.Round(_mV, 2));
+
+            return new ElectricPotential(_mV, UnitsNet.Units.ElectricPotentialUnit.Millivolt);
+        }
+
+        /// <summary>
+        /// Measure temperature
+        /// </summary>
+        /// <returns>Temperature. Temperature range: -2 to 35 C. Temperature Precision: 0.05 C</returns>
+        public Temperature MeasureTemp()
```
Ellerbach, line 82: Consider using the full name for temperature: MeasureTemperature

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Changes the default I2C address
+        /// </summary>
+        /// <param name="i2cAddress">The new I2C address</param>
+        public void SetI2CAddress(byte i2cAddress)
+        {
+            if (i2cAddress >= 1 && i2cAddress <= 127)
```
Ellerbach, line 236: You may consider throwing an Argument Exception if the value is out of range

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+                Send_command((byte)Command.ISE_I2C);
+            }
+        }
+
+        /// <summary>
+        /// Low level reads of EEPROM
+        /// </summary>
+        /// <param name="address">Address to read</param>
+        /// <returns>Low level data from EEPROM</returns>
+        public float ReadEEPROM(float address)
```
Ellerbach, line 248: So why do you keep a float if the value is a byte address? 
if you use ```public float ReadEeprom(byte address)``` it makes it a bit easier.
Also consider throwing an Argument Exception then is address == 0xFF
And please consider using the correct naming convention so Eeprom rather than EEPROM when using for function, properties, variables

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+
+            return 0;
+        }
+
+        /// <summary>
+        /// Low level write of EEPROM
+        /// </summary>
+        /// <param name="address">Address</param>
+        /// <param name="data">Data to write to the address</param>
+        public void WriteEEPROM(float address, float data)
```
Ellerbach, line 265: Same comment. Also consider adding the range for the variable in the comments

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+
+        }
+
+        private void Send_command(byte data)
+        {
+            Span<byte> bytes = stackalloc byte[2];
+            bytes[0] = (byte)Register.ISE_TASK_REGISTER;
+            bytes[1] = data;
+            _device.Write(bytes);
+            DelayHelper.DelayMilliseconds(ISE_COMMUNICATION_DELAY, allowThreadYield: true);
```
Ellerbach, line 291: Please refer to somewhere in the documentation for the delays

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+            data[1] = _device.ReadByte();
+            data[2] = _device.ReadByte();
+            data[3] = _device.ReadByte();
+
+            return Convert.ToSingle(Round_total_digits(BitConverter.ToSingle(data, 0), 7));
+        }
+
+        private void Write_register(byte register, float? data)
+        {
+            byte[] dataArray = { 0, 0, 0, 0 };
```
Ellerbach, line 310: Would make sense to have either the Span<> either the byte array but not both. One is redondant.

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+
+            Span<byte> bytes = stackalloc byte[4];
+            bytes[0] = dataArray[0];
+            bytes[1] = dataArray[1];
+            bytes[2] = dataArray[2];
+            bytes[3] = dataArray[3];
+
+            Change_register(register);
+            _device.Write(bytes);
+            DelayHelper.DelayMilliseconds(ISE_COMMUNICATION_DELAY, allowThreadYield: true);
```
Ellerbach, line 325: For all wait delay, please refer to the page of the documentation. This comment is valid for all the wait you do in the code. A simple comment like ```// Page 42 of documentation says delay of xx ms```
You can either refer the document like ```// See documentation https://www.url.com/folder/doc.pdf``` with the line if it happens only once. Either put this reference to the documentation at the top of the file for example if it's the same everywhere. If someone has to maintain the code, it makes things much easier.

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -0,0 +1,388 @@
...
+        }
+
+        private void Change_register(byte register)
```
Ellerbach, line 328: Please refer to the naming convention, it should be ```ChangeRegister```
Please adjust as well all other functions and names like this

src/devices/uFire_ISE/UFireOrp.cs
```diff
@@ -0,0 +1,75 @@
...
+using UnitsNet;
+
+namespace Iot.Device.UFire
+{
+    /// <summary>
+    /// Get ORP measuremens from μFire ISE Probe Interface
+    /// </summary>
+    public class UFireOrp : UFireIse
+    {
+        private const float POTENTIAL_REGISTER_ADDRESS = 0x64;
```
Ellerbach, line 16: Please see the comment on naming. Also why a float if it's a byte?

src/devices/uFire_ISE/UFireOrp.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// ORP measuremens
+        /// </summary>
+        public ElectricPotential ORP = new ElectricPotential();
```
Ellerbach, line 21: Please refer as well to naming convention, should be ```Orp``` and as ORP is not a well know abbreviation, should be the full name. Same comment for ISE

src/devices/uFire_ISE/UFireOrp.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Eh measuremens
+        /// </summary>
+        public ElectricPotential Eh = new ElectricPotential();
```
Ellerbach, line 26: Same comment for Eh, consider using the full name and not the abbreviation

src/devices/uFire_ISE/UFireOrp.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <param name="i2cDevice">The I2C device to be used</param>
+        public UFireOrp(I2cDevice i2cDevice)
+            : base(i2cDevice)
+        {
+        }
+
+        /// <summary>
+        /// The probe potential
+        /// </summary>
+        public ElectricPotential ProbePotential
```
Ellerbach, line 40: ```public ElectricPotential ProbePotential => ElectricPotential.FromMillivolts(ReadEEPROM(POTENTIAL_REGISTER_ADDRESS));```

src/devices/uFire_ISE/UFireOrp.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+        }
+
+        /// <summary>
+        /// Tries to measurer ORP (Oxidation-Reduction Potential).
+        /// </summary>
+        /// <param name="orp">ORP (Oxidation-Reduction Potential) measurement</param>
+        /// <returns>True if it could measure ORP (Oxidation-Reduction Potential) else false</returns>
+        public bool TryMeasureORP(out ElectricPotential orp)
```
Ellerbach, line 54: So full naming should be then ```TryMeasureOxidationReducationPotential```

src/devices/uFire_ISE/UFirePh .cs
```diff
@@ -4,95 +4,101 @@
...
         /// <summary>
-        /// pH measuremen
+        /// pH (Power of Hydrogen) units measurement
         /// </summary>
-        public float PH = 0;
+        public float Ph = 0;
 
         /// <summary>
-        /// poH measuremen
+        /// pOH units measurement
```
Ellerbach, line 25: what is pOH vs pH, please add in the comments. Also, like for the rest of similar comments, consider using the full name rather than the abbreviation

src/devices/uFire_ISE/UFirePh .cs
```diff
@@ -4,95 +4,101 @@
...
         /// <param name="i2cDevice">The I2C device to be used</param>
-        public UFire_pH(I2cDevice i2cDevice)
+        public UFirePh(I2cDevice i2cDevice)
             : base(i2cDevice)
         {
         }
 
         /// <summary>
-        /// Starts a pH measurement.
+        /// Tries to measurer pH (Power of Hydrogen) .
```
Ellerbach, line 39: nit measure

src/devices/uFire_ISE/samples/uFire_ISE.Sample.cs
```diff
@@ -12,8 +12,9 @@ namespace Iot.Device.UFire.Sample
     /// </summary>
     public static class Program
     {
-        private const int BUSID = 1;
-        private const int I2CADDRESS = 0x3F;
+        private const int BusId = 1;
+        // 0x3F is the default address of all sensors
+        private const int I2cAddress = 0x3F;
```
Ellerbach, line 17: So if this is the default address of all sensors, then it would make sense to expose this const in the main base UFireIse class as public, and also consider changing for byte, so something like this:
```csharp
public class UFireIse : IDisposable
{
    /// <summary>
    /// 0x3F is the default address of all sensors
    /// </summary>
    public const byte I2cAddress = 0x3F;
}
```

Ellerbach (commented):
Thanks for the changes, couple of more remarks. Also, you should consider running ```build.cmd``` or ```build.sh``` in the root of the cloned repo. This will also find issues in the code for naming, spaces, extra lines, missing docs, etc.

src/devices/uFire_ISE/UFireIse.cs
R10Rasmus: I have asked the manufacturer to document this delay. I will get back to you when I hear back from them.

src/devices/uFire_ISE/UFirePh .cs
R10Rasmus: pOH is a unit and I can not write it in any other way then pOH. I have added a comment explaining the relationship between pH and pOH

src/devices/uFire_ISE/UFireIse.cs
```diff
@@ -14,9 +14,17 @@ namespace Iot.Device.UFire
     /// </summary>
     public class UFireIse : IDisposable
     {
+        /// <summary>
+        /// 0x3F is the default address of all sensors
+        /// </summary>
+        public static byte I2cAddress = 0x3F;
+
+        // A temperature measurement takes 750ms, see https://www.ufire.co/docs/uFire_ISE/#use
         private const int ISE_TEMP_MEASURE_TIME = 750;
```
Ellerbach, line 23: Thanks for adding the links. So would make sense then to adjust all the naming for the constants in this file and in the classes with the standard notation: IseTemperatureMeasureTime, IseMvMeasureTime, etc

src/devices/uFire_ISE/UFirePh .cs
```diff
@@ -22,7 +22,7 @@ public class UFirePh : UFireIse
         public float Ph = 0;
```
Ellerbach, line 22: As I see you are using Ph = -1 for wrong values, then maybe initalize to -1?
Also consider using explicit property:
```csharp
public float Ph { get; set; } = -1;
```

src/devices/uFire_ISE/UFirePh .cs
```diff
@@ -22,7 +22,7 @@ public class UFirePh : UFireIse
...
 
         /// <summary>
-        /// pOH units measurement
+        /// pOH units measurement, for the relationship between pH and pOH see https://www.chem.purdue.edu/gchelp/howtosolveit/Equilibrium/Calculating_pHandpOH.htm#pOH
         /// </summary>
         public float Poh = 0;
```
Ellerbach, line 27: So if Ph + Poh = 14, then probably a better way to have this done is like this:
```csharp
public float Poh => Ph >=0 ? Math.Abcs(14 - Ph) : -1;
```

src/devices/uFire_ISE/UFirePh .cs
```diff
@@ -36,14 +36,14 @@ public UFirePh(I2cDevice i2cDevice)
...
         /// <param name="pH">The measure pH value</param>
         /// <param name="temp">Temperature compensation is available by passing the temperature.</param>
         /// <returns>True if it could measure pH (Power of Hydrogen) else false</returns>
         public bool TryMeasurepH(out float pH, Temperature? temp = null)
         {
-            ElectricPotential mV = MeasuremV();
+            ElectricPotential mV = Measure();
             if (mV.Value == -1)
             {
                 pH = -1;
```
Ellerbach, line 49: next line not needed, and remove all the double pH = -1. Or maybe you want to use the property, so then you should consider:
```csharp
Ph = pH;
```

src/devices/uFire_ISE/samples/uFire_ISE.Sample.cs
```diff
@@ -67,26 +65,34 @@ private static void Basic(I2cDevice device)
         {
             using (UFireIse uFire_ISE = new UFireIse(device))
```
Ellerbach, line 66: uFireIse rather than uFire_ISE

src/devices/uFire_ISE/samples/uFire_ISE.Sample.cs
```diff
@@ -67,26 +65,34 @@ private static void Basic(I2cDevice device)
...
             {
-                Console.WriteLine("mV:" + uFire_ISE.MeasuremV());
+                Console.WriteLine("mV:" + uFire_ISE.Measure().Millivolts);
             }
         }
 
         private static void Orp(I2cDevice device)
         {
             using (UFireOrp uFire_orp = new UFireOrp(device))
```
Ellerbach, line 74: uFireOrp rather than uFire_orp

Ellerbach (commented):
Please run the build locally on your machine and correct the errors you will see from the local build. Correcting those errors will most likely clear all the errors that are present in the CI/CD.
Please add as well README.md at least in the root of the main device. 
And you should per naming conventions rename the root directory to: src/devices/UFireIse

src/devices/uFire_ISE/UFirePh .cs
Ellerbach: And also, rather than -1 for non valid, maybe better to use float.NaN?

src/devices/uFire_ISE/UFirePh .cs
Ellerbach: Same as above. Maybe better to use float.NaN.

src/devices/uFire_ISE/UFirePh .cs
Ellerbach: Also, are you sure that -1 is not a correct value returned by Measure?

R10Rasmus: yes I am sure it return -1 as a value if it get an error.

R10Rasmus:
The main root of the main device (uFireIse) do contain a README.md. Do I need to put it in some other location?

Also the failure of the CI/CD looks like a error in the System.Device.Gpio.Tests.dll.WorkItemExecution (the error is: The Work Item Failed). Do I need to change anything, to make this test pass?

src/devices/uFireIse/UFireIse.cs
```diff
@@ -0,0 +1,393 @@
...
+            DelayHelper.DelayMilliseconds(IseCommunicationDelay, allowThreadYield: true);
+        }
+
+        private float ReadRegister(byte register)
+        {
+            byte[] data = { 0, 0, 0, 0 };
+
+            ChangeRegister(register);
+
+            data[0] = _device.ReadByte();
```
Ellerbach, line 305: Why not something like this?:
```csharp
Span<byte> data = stackalloc new byte[4];
_device.Read(data);
```
This will read the 4 bytes directly.

src/devices/uFireIse/UFirePh .cs
```diff
@@ -0,0 +1,169 @@
...
+                    temp_multiplier *= -1;
+                }
+
+                Ph += Convert.ToSingle(temp_multiplier);
+            }
+
+            if (Ph <= 0.0 || Ph >= 14.0)
+            {
+                pH = float.NaN;
+                pH = float.NaN;
```
Ellerbach, line 79: for the second like I guess it's ```Ph = float.NaN```

src/devices/uFireIse/UFirePh .cs
```diff
@@ -0,0 +1,169 @@
...
+
+                return false;
+            }
+
+            if (float.IsNaN(Ph) || float.IsInfinity(Ph))
+            {
+                pH = float.NaN;
+                pH = float.NaN;
```
Ellerbach, line 87: for the second like I guess it's Ph = float.NaN

src/devices/uFireIse/UFirePh .cs
```diff
@@ -0,0 +1,169 @@
...
+
+                return false;
+            }
+
```
Ellerbach, line 91: You should here add: ```Ph = pH``` or your property will never be updated?

src/devices/uFireIse/UFirePh .cs
```diff
@@ -0,0 +1,169 @@
...
+using UnitsNet;
+
+namespace Iot.Device.UFire
+{
+    /// <summary>
+    /// Get pH measuremens from μFire ISE (ion-selective electrode) Probe Interface
+    /// </summary>
+    public class UFirePh : UFireIse
+    {
+        private const float PROBE_MV_TO_PH = 59.2F;
```
Ellerbach, line 16: As for all the other constants which are not really defined like this in the doc, please use the proper notation. So would be ProbeMvToPh
TempCorrectionFactor

src/devices/uFireIse/README.md
```diff
@@ -0,0 +1,15 @@
+﻿# μFire ISE Probe Interface - I2C Driver
+
+## Summary
+
+The μFire ISE Probe Interface is an I²C sensor that can read a pH probe. Attach a waterproof temperature sensor for temperature compensation with the attached connector.
+
```
Ellerbach, line 6: Usually we ask to add the usage of the device in the bindings. How to create the device, basic usage for reading, things like this.

Ellerbach (commented):
Thanks for all the modifications. It looks very good new. very few things still pending and add some more details in the README. Then would be great if @krwq or @joperezr can have a look.

src/devices/Device-Index.md
```diff
@@ -65,6 +65,7 @@
 * [Still image recording library](Media/README.md)
 * [TCS3472x Sensors](Tcs3472x/README.md)
 * [TM1637 - Segment Display](Tm1637/README.md)
+* [uFire_ISE - I2C Driver](uFire_ISE/README.md)
```
Ellerbach, line 68: You should not change this file, it is generated automatically. Add a txt file in the root of the directory with the device description

src/devices/README.md
```diff
@@ -211,6 +211,7 @@ Our vision: the majority of .NET bindings are written completely in .NET languag
 * [Adafruit Seesaw - extension board (ADC, PWM, GPIO expander)](Seesaw/README.md)
 * [MotorHat](MotorHat/README.md)
 * [Pca9685 - I2C PWM Driver](Pca9685/README.md)
+* [uFire_ISE - I2C Driver](uFire_ISE/README.md)
```
Ellerbach, line 214: Same, please don't change this file, it is changed automatically. Please add a txt file at root of the directory, you'll find examples in all bindings.

Ellerbach:
> The main root of the main device (uFireIse) do contain a README.md. Do I need to put it in some other location?

It's ok, I totally missed the file. I added a comment in this review, please add some examples how to use the sensor, you can copy/paste from the example, add some comments. It's always very helpful to have this.

Ellerbach:
> Also the failure of the CI/CD looks like a error in the System.Device.Gpio.Tests.dll.WorkItemExecution (the error is: The Work Item Failed). Do I need to change anything, to make this test pass

It's ok, previously for some of the previous commits, you had issues with your code, now issue seems to be internal to the CI/CD.

src/devices/uFireIse/UFireIse.cs
```diff
@@ -298,33 +298,35 @@ private void SendCommand(byte data)
...
             ChangeRegister(register);
 
-            data[0] = _device.ReadByte();
-            data[1] = _device.ReadByte();
-            data[2] = _device.ReadByte();
-            data[3] = _device.ReadByte();
-
-            return Convert.ToSingle(RoundTotalDigits(BitConverter.ToSingle(data, 0), 7));
+            Span<byte> data = stackalloc byte[4]
+            {
```
Ellerbach, line 304: You don't need to initialize as you will read the data right after. So you can get rid of this part from line 304 to line 309.

Ellerbach:
Thanks again for all the changed, just realize that the folder name is still not the perfect one. It should start with a uppercase. I gave the right way of writing it: https://github.com/dotnet/iot/pull/1134#pullrequestreview-474490265
Otherwise, looks really good now with all the adjustments!

R10Rasmus:
Sorry I made a copay/past mistake :(Also sorry for the many checkin's. The folder name was not straightforward to change

R10Rasmus:
thanks for all the feedback :) I hope we together got the code ready to be merged in :)

Ellerbach (approved):
@krwq and @joperezr, looks good to me now. Please have a look for potential additional feedbacks.

src/devices/UFireIse/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+The μFire ISE Probe Interface is an I²C sensor that can read a pH probe. Attach a waterproof temperature sensor for temperature compensation with the attached connector.
+
+## Usage
+
+You can find an example in the [sample](./samples/uFireIse.Sample.cs) directory. 
+
+
+### Basic
+   It is possible to read the basic value (Electric Potential) from the probe.
```
krwq, line 13:
```suggestion
It is possible to read the basic value (Electric Potential) from the probe.
```

src/devices/UFireIse/README.md
```diff
@@ -0,0 +1,66 @@
...
+   
+```csharp
+using (UFireIse uFireIse = new UFireIse(device))
+	{
+		Console.WriteLine("mV:" + uFireIse.Measure().Millivolts);
+	}
```
krwq, line 19:
```suggestion
using (UFireIse uFireIse = new UFireIse(device))
{
    Console.WriteLine("mV:" + uFireIse.Measure().Millivolts);
}
```

src/devices/UFireIse/README.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+### Orp 
+   To read the ORP (OxidationReducationPotential) value use this example
```
krwq, line 23:
```suggestion
To read the ORP (OxidationReducationPotential) value use this example
```

src/devices/UFireIse/README.md
```diff
@@ -0,0 +1,66 @@
...
+   
+```csharp
+using (UFireOrp uFireOrp = new UFireOrp(device))
+	{
```
krwq, line 27: please fix indentation as above

src/devices/UFireIse/Register.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <summary>
+        /// config register
+        /// </summary>
+        ISE_CONFIG_REGISTER = 0x26,
+
+        /// <summary>
+        /// firmware version
+        /// </summary>
+        ISE_TASK_REGISTER = 0x27
+
```
krwq, line 73: remove extra enter

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+        public UFireIse(I2cDevice i2cDevice)
+        {
+            _device = i2cDevice;
+        }
+
+        /// <summary>
+        /// Measure a value from the ISE Probe Interface, typical measure are in the millivolt range.
+        /// </summary>
+        /// <returns>value from ISE Probe Interface, typical measure are in the millivolt range. On error it return -1 as value</returns>
+        public ElectricPotential Measure()
```
krwq, line 75: ReadVoltage/ReadElectricPotential

If this sensor was measuring only one thing then just "Read" would be ok

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+            _measurement = Convert.ToSingle(Math.Round(_measurement, 2));
+
+            return new ElectricPotential(_measurement, UnitsNet.Units.ElectricPotentialUnit.Millivolt);
+        }
+
+        /// <summary>
+        /// Measure temperature
+        /// </summary>
+        /// <returns>Temperature. Temperature range: -2 to 35 C. Temperature Precision: 0.05 C</returns>
+        public Temperature MeasureTemperature()
```
krwq, line 91: ReadTemperature

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+            _temperature = new Temperature(ReadRegister((byte)Register.ISE_TEMP_REGISTER), UnitsNet.Units.TemperatureUnit.DegreeCelsius);
+
+            return _temperature;
+        }
+
+        /// <summary>
+        /// Set temperature compensation
+        /// </summary>
+        /// <param name="temp">Compensation temperature</param>
+        public void SetTemperatureCompensation(Temperature temp)
```
krwq, line 105: expand parameter and please put clearer description on what this means

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+            _temperature = temp;
+        }
+
+        /// <summary>
+        /// Calibrates the probe using a single point using a mV value.
+        /// </summary>
+        /// <param name="solution">Value. Range: -1024 mV to 1024 mV</param>
+        public void CalibrateSingle(ElectricPotential solution)
+        {
+            if (solution.Millivolts >= -1024 && solution.Millivolts <= 1024)
```
krwq, line 117: you should throw otherwise and not silently quit

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+        {
+            WriteRegister((byte)Register.ISE_TEMP_REGISTER, Convert.ToSingle(temp.DegreesCelsius));
+            _temperature = temp;
+        }
+
+        /// <summary>
+        /// Calibrates the probe using a single point using a mV value.
+        /// </summary>
+        /// <param name="solution">Value. Range: -1024 mV to 1024 mV</param>
+        public void CalibrateSingle(ElectricPotential solution)
```
krwq, line 115: this needs more explanation. Calibrate to what? are you marking where 0 is? If this needs longer explanation please make sure at least sample explains how to use it

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+
+                DelayHelper.DelayMilliseconds(IseMvMeasureTime, allowThreadYield: true);
+            }
+        }
+
+        /// <summary>
+        /// Calibrates the dual-point values for the high reading, in mV, and saves them in the devices's EEPROM.
+        /// </summary>
+        /// <param name="solution">Value. Range: -1024 mV to 1024 mV</param>
+        public void CalibrateProbeHigh(ElectricPotential solution)
```
krwq, line 145: can we merge three methods into one with longer explanation? I assume you need to pass all 3 points in order to make this usable

krwq: (internally you can have 3 methods but public API should be more obvious how to use)

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Returns the firmware version of the device.
+        /// </summary>
+        /// <returns>Firmware version</returns>
+        public Version GetVersion()
+        {
+            return new Version(ReadRegister((byte)Register.ISE_VERSION_REGISTER).ToString());
```
krwq, line 162: does it make sense to use constructor which doesn't take string? Is this major version? Minor? if so perhaps:
`return new Version(ReadRegister(....), 0)` or something along the lines

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+        /// When there are high and low calibration points, the device will automatically use them to adjust readings.To disable dual-point adjustment, call reset to remove all calibration data.
+        /// </summary>
+        /// <returns>The known value (reference value) for calibrate the high value</returns>
+        public ElectricPotential GetCalibrateHighReference()
+        {
+            return new ElectricPotential(ReadRegister((byte)Register.ISE_CALIBRATE_REFHIGH_REGISTER), UnitsNet.Units.ElectricPotentialUnit.Millivolt);
+        }
+
+        /// <summary>
+        /// Dual point uses two measures for low and high points. It needs the measured value (reading value) and the known value (reference value). Calling calibrateProbeLow saves both the reading and reference value.
```
krwq, line 176:
```suggestion
        /// Dual point uses two measures for low and high points. It needs the measured value (reading value) and the known value (reference value). Calling CalibrateProbeLow saves both the reading and reference value.
```

please ensure this name is correct

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+        /// <returns>The measured value (reading value) for calibrate the low value</returns>
+        public ElectricPotential GetCalibrateLowReading()
+        {
+            return new ElectricPotential(ReadRegister((byte)Register.ISE_CALIBRATE_READLOW_REGISTER), UnitsNet.Units.ElectricPotentialUnit.Millivolt);
+        }
+
+        /// <summary>
+        ///  Resets all the stored calibration information
+        /// </summary>
+        public void Reset()
```
krwq, line 209: `ResetCalibration`

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+            if (refLow.Millivolts >= -1024 && refLow.Millivolts <= 1024 &&
+                refHigh.Millivolts >= -1024 && refHigh.Millivolts <= 1024 &&
+                readLow.Millivolts >= -1024 && readLow.Millivolts <= 1024 &&
+                readHigh.Millivolts >= -1024 && readHigh.Millivolts <= 1024)
+            {
+                WriteRegister((byte)Register.ISE_CALIBRATE_REFLOW_REGISTER, Convert.ToSingle(refLow.Millivolts));
+                WriteRegister((byte)Register.ISE_CALIBRATE_REFHIGH_REGISTER, Convert.ToSingle(refHigh));
+                WriteRegister((byte)Register.ISE_CALIBRATE_READLOW_REGISTER, Convert.ToSingle(readLow));
+                WriteRegister((byte)Register.ISE_CALIBRATE_READHIGH_REGISTER, Convert.ToSingle(readHigh));
+            }
```
krwq, line 236: please throw if out of range

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+        /// <param name="readLow">the measured low point. Range: -1024 mV to 1024 mV</param>
+        /// <param name="readHigh">the measured high point. Range: -1024 mV to 1024 mV</param>
+        public void SetDualPointCalibration(ElectricPotential refLow, ElectricPotential refHigh, ElectricPotential readLow, ElectricPotential readHigh)
+        {
+            if (refLow.Millivolts >= -1024 && refLow.Millivolts <= 1024 &&
+                refHigh.Millivolts >= -1024 && refHigh.Millivolts <= 1024 &&
+                readLow.Millivolts >= -1024 && readLow.Millivolts <= 1024 &&
+                readHigh.Millivolts >= -1024 && readHigh.Millivolts <= 1024)
+            {
+                WriteRegister((byte)Register.ISE_CALIBRATE_REFLOW_REGISTER, Convert.ToSingle(refLow.Millivolts));
```
krwq, line 232: consider making these properties since you got both getter and setter

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+                WriteRegister((byte)Register.ISE_CALIBRATE_READLOW_REGISTER, Convert.ToSingle(readLow));
+                WriteRegister((byte)Register.ISE_CALIBRATE_READHIGH_REGISTER, Convert.ToSingle(readHigh));
+            }
+        }
+
+        /// <summary>
+        /// Changes the default I2C address
+        /// </summary>
+        /// <param name="i2cAddress">The new I2C address</param>
+        public void SetI2CAddress(byte i2cAddress)
```
krwq, line 243: I assume after you do that you won't be able to communicate to the device, right? If so then rename this to SetI2cAddressAndDispose (and call Dispose)

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+        {
+            if (i2cAddress >= 1 && i2cAddress <= 127)
+            {
+                WriteRegister((byte)Register.ISE_SOLUTION_REGISTER, Convert.ToByte(i2cAddress));
+                SendCommand((byte)Command.ISE_I2C);
+            }
+            else
+            {
+                throw new ArgumentOutOfRangeException();
```
krwq, line 252: provide  argument to the exception - might be useful to provide what range is valid in the message

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Low level reads of EEPROM
+        /// </summary>
+        /// <param name="address">Address to read</param>
+        /// <returns>Low level data from EEPROM</returns>
+        public float ReadEeprom(byte address)
```
krwq, line 261: does it ever make sense to read the eeprom as byte or other type than float?

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+        private void SendCommand(byte data)
+        {
+            Span<byte> bytes = stackalloc byte[2];
+            bytes[0] = (byte)Register.ISE_TASK_REGISTER;
+            bytes[1] = data;
+            _device.Write(bytes);
+            DelayHelper.DelayMilliseconds(IseCommunicationDelay, allowThreadYield: true);
+        }
+
+        private float ReadRegister(byte register)
```
krwq, line 300: ReadFloat, also it should take typed Register and cast in here

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+            Span<byte> data = stackalloc byte[4];
+
+            _device.Read(data);
+
+            return Convert.ToSingle(RoundTotalDigits(BitConverter.ToSingle(data.ToArray(), 0), 7));
+        }
+
+        private void WriteRegister(byte register, float? data)
+        {
+            Span<byte> bytes = stackalloc byte[4]
```
krwq, line 313: what does null mean here (or passing all zeros)? Default?

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+             {
+                0,
+                0,
+                0,
+                0
+             };
+
+            if (data != null)
+            {
+                bytes = BitConverter.GetBytes(RoundTotalDigits(data.Value));
```
krwq, line 323: BitConverter.GetBytes will allocate array, can you use overload which takes Span and instead write to it?

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+            }
+
+            ChangeRegister(register);
+            _device.Write(bytes);
+            DelayHelper.DelayMilliseconds(IseCommunicationDelay, allowThreadYield: true);
+        }
+
+        private void ChangeRegister(byte register)
```
krwq, line 331: what does this actually do? what does it change it to?

krwq, line 331: you should take typed register here and everywhere where applies

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+            _device.WriteByte(value);
+
+            DelayHelper.DelayMilliseconds(IseCommunicationDelay, allowThreadYield: true);
+
+        }
+
+        private int Bit_set(int v, int index, bool x)
+        {
+            int mask = 1 << index;
+            v &= ~mask;
```
krwq, line 358: might be better to:
```
if (x)
{
  v |= mask;
}
else
{
  v &= ~=mask;
}
```

rather than always clearing first

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+        {
+            return float.IsNaN(x) || x == 0 ? 0 : (int)Math.Floor(Math.Log10(Math.Abs(x))) + 1;
+        }
+
+        private float RoundTotalDigits(float x, int? digits = 7)
+        {
+            int numberOfDigits = digits.Value - Magnitude(x);
+            if (numberOfDigits > 15)
+            {
+                numberOfDigits = 15;
```
krwq, line 377: should this throw instead?

src/devices/UFireIse/UFireOrp.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Get ORP (oxidation-reduction potential) measuremens from μFire ISE (ion-selective electrode) Probe Interface
+    /// </summary>
+    public class UFireOrp : UFireIse
+    {
+        private const byte POTENTIAL_REGISTER_ADDRESS = 0x64;
+
+        /// <summary>
+        /// Oxidation-reduction potential (ORP) measuremens
+        /// </summary>
+        public ElectricPotential OxidationReducationPotential = new ElectricPotential();
```
krwq, line 20: fields should not be public (if I missed something before also fix)

src/devices/UFireIse/UFireOrp.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// The probe potential
+        /// </summary>
+        public ElectricPotential ProbePotential => ElectricPotential.FromMillivolts(ReadEeprom(POTENTIAL_REGISTER_ADDRESS));
+
+        /// <summary>
+        /// Tries to measure ORP (Oxidation-Reduction Potential).
+        /// </summary>
+        /// <param name="orp">ORP (Oxidation-Reduction Potential) measurement</param>
+        /// <returns>True if it could measure ORP (Oxidation-Reduction Potential) else false</returns>
+        public bool TryMeasureOxidationReducationPotential(out ElectricPotential orp)
```
krwq, line 46: Reduction or Reducation? (here and anywhere applies)

src/devices/UFireIse/UFirePh .cs
```diff
@@ -0,0 +1,172 @@
...
+    /// </summary>
+    public class UFirePh : UFireIse
+    {
+        private const float ProbeMvToPh = 59.2F;
+        private const float TemperatureCorrectionFactor = 0.03F;
+
+        /// <summary>
+        /// pH (Power of Hydrogen) units measurement
+        /// </summary>
+        public float Ph = 0;
```
krwq, line 22: I assume UnitsNet doens't have unit for that? Acidity or something?

src/devices/UFireIse/UFirePh .cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <param name="pH">The measure pH value</param>
+        /// <param name="temp">Temperature compensation is available by passing the temperature.</param>
+        /// <returns>True if it could measure pH (Power of Hydrogen) else false</returns>
+        public bool TryMeasurepH(out float pH, Temperature? temp = null)
+        {
+            // It return -1 on error
+            ElectricPotential mV = Measure();
+            if (mV.Value == -1)
+            {
+                pH = float.NaN;
```
krwq, line 50: Ph = pH = float.NaN;

src/devices/UFireIse/UFirePh .cs
```diff
@@ -0,0 +1,172 @@
...
+            Ph = pH;
+
+            return true;
+        }
+
+        /// <summary>
+        /// Calibrates the probe using a single point using a pH value.
+        /// </summary>
+        /// <param name="solutionpH">pH value</param>
+        public void CalibrateSingle(float solutionpH)
```
krwq, line 102: please explain how this works

src/devices/UFireIse/UFirePh .cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <summary>
+        /// Returns the dual-point calibration low-reading value.
+        /// </summary>
+        /// <returns></returns>
+        public new float GetCalibrateLowReading()
+        {
+            return MVtopH(GetCalibrateLowReading());
+        }
+
+        private float PHtomV(float pH)
```
krwq, line 161: PhToMillivolts - this name took me couple of seconds to figure out what it means

src/devices/UFireIse/UFirePh .cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            return (7 - pH) * ProbeMvToPh;
+        }
+
+        private float MVtopH(float mV)
+        {
+            return Convert.ToSingle(Math.Abs(7.0 - (mV / ProbeMvToPh)));
+        }
+
```
krwq, line 170: extra enter

src/devices/UFireIse/samples/UFireIse.Sample.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// </summary>
+    public static class Program
+    {
+        private const int BusId = 1;
+
+        private static void PrintHelp()
+        {
+            Console.WriteLine("Command:");
+            Console.WriteLine("    B           Basic");
+            Console.WriteLine("    O           Orp");
```
krwq, line 22: what does Orp mean?

src/devices/UFireIse/samples/UFireIse.Sample.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        /// <summary>
+        /// Example program main entry point
+        /// </summary>
+        /// <param name="args">Command line arguments see <see cref="PrintHelp"/></param>
+        public static void Main(string[] args)
+        {
+            I2cConnectionSettings settings = new I2cConnectionSettings(BusId, UFireIse.I2cAddress);
+            I2cDevice device = I2cDevice.Create(settings);
```
krwq, line 34: 1. you should always wrap this in using statement
2. only open when you actually use it (i.e. not only printing help)

src/devices/UFireIse/samples/UFireIse.Sample.cs
```diff
@@ -0,0 +1,107 @@
...
+                    Console.WriteLine("pOH:" + uFire_pH.Poh);
+                }
+                else
+                {
+                    Console.WriteLine("Not possible to measure pH");
+                }
+            }
+        }
+    }
+
```
krwq, line 106: extra enter

src/devices/UFireIse/README.md
```diff
@@ -0,0 +1,66 @@
...
+		}
+		else
+		{
+			Console.WriteLine("Not possible to measure pH");
+		}
+	}
+```
+
+### Ph 
+   To read the Ph (Power of Hydrogen) value use this example
```
krwq, line 40: please cover calibration in the sample or make summary on how to use it clear

krwq:
Sorry for late review... I've added bunch of comments but overall looks good (especially for first time contribution)

krwq:
(wrong button)

src/devices/UFireIse/UFireIse.cs
R10Rasmus: It should not be properties  because the  refLow, refHigh, readLow and readHigh values are only relevant when running the function.

src/devices/UFireIse/UFireIse.cs
R10Rasmus: it do only make sense to the the eeprom as float

src/devices/UFireIse/UFireIse.cs
R10Rasmus: it use 0 as default

src/devices/UFireIse/UFireIse.cs
R10Rasmus: GetBytes do not have a Span overload

src/devices/UFireIse/UFireOrp.cs
R10Rasmus: It makes it possible to the Oxidation-reduction potential (ORP) measurement without doing a measuremen (the retunres the old measurement)

src/devices/UFireIse/UFirePh .cs
R10Rasmus: yes UnitsNet do not have a unit for pH (Power of Hydrogen)

R10Rasmus:
I hope I got everything, also see comments :)

src/devices/UFireIse/UFireIse.cs
krwq: In such case I'd recommend instead of making it nullable to change signature to: `private void WriteRegister(byte register, float data = 0)`

src/devices/UFireIse/UFireIse.cs
krwq: @tannergooding do you perhaps know how to do this in non-allocating way?

src/devices/UFireIse/UFireOrp.cs
krwq: I meant that you should either make them properties or private

krwq:
@R10Rasmus thanks! I'll be taking a look right now (I've responded to comments already). As a side note to make our life easier in the future it's usually a good idea to not Resolve conversations where you've responded to and allow author of the question/comment to close (unless it's a non-actionable comment like "I like what you did here")

src/devices/UFireIse/README.md
```diff
@@ -0,0 +1,77 @@
...
+}
+```
+
+### Orp 
+To read the ORP (OxidationReducationPotential) value use this example
+   
+```csharp
+using (UFireOrp uFireOrp = new UFireOrp(device))
+{
+	if (uFireOrp.TryMeasureOxidationReducationPotential(out ElectricPotential orp))
```
krwq, line 28: should this be `ReductionPotential`? I assume you changed the code but sample is still left (you might want to search for Reducation and fix all)

src/devices/UFireIse/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+### Calibration 
+Calibration of the probe using a single solution. 
+Put the probe in a solution where the pH (Power of Hydrogen) value is known (in this example we assume it is 7).
+   
+```csharp
+ using (UFirePh uFire_pH = new UFirePh(device))
+{
+    uFire_pH.CalibrateSingle(7);
```
krwq, line 46: really nice, might also be worth adding if this calibration stays if you reboot device/app or you need to re-calibrate every time

src/devices/UFireIse/Register.cs
```diff
@@ -0,0 +1,82 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.UFire
+{
+    /// <summary>
+    /// Register with all adress values for μFire ISE (Ion Specific Electrode) Probe Interface controller
+    /// </summary>
+    public enum Register : byte
```
krwq, line 10: this should not be public

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+
+        /// <summary>
+        /// Use temperature compensation
+        /// </summary>
+        public bool TemperatureCompensation
+        {
+            get
+            {
+                return _temperatureCompensation;
+            }
```
krwq, line 48: nit: this can be written as:

```suggestion
            get => _temperatureCompensation;
```

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UFireIse"/> class.
+        /// </summary>
+        /// <param name="i2cDevice">The I2C device to be used</param>
+        public UFireIse(I2cDevice i2cDevice)
+        {
+            _device = i2cDevice;
```
krwq, line 68: consider throwing ArgumentNullException if null was provided

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+        }
+
+        /// <summary>
+        /// Read a value from the ISE Probe Interface, typical measure are in the millivolt range.
+        /// </summary>
+        /// <returns>value from ISE Probe Interface, typical measure are in the millivolt range. On error it return -1 as value</returns>
+        public ElectricPotential Read()
```
krwq, line 75: ReadElectricPotential

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+        /// </summary>
+        /// <param name="temp">The temperature used for compensation</param>
+        public void SetTemperatureCompensation(Temperature temp)
+        {
+            WriteRegister(Register.ISE_TEMP_REGISTER, Convert.ToSingle(temp.DegreesCelsius));
+            _temperature = temp;
+        }
+
+        /// <summary>
+        /// Calibrates the probe using a single solution. Put the probe in a solution where the pH (Power of Hydrogen) is know.
```
krwq, line 112:
```suggestion
        /// Calibrates the probe using a single solution. Put the probe in a solution where the pH (Power of Hydrogen) is known.
```

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+        {
+            if (solution.Millivolts >= -1024 && solution.Millivolts <= 1024)
+            {
+                WriteRegister(Register.ISE_SOLUTION_REGISTER, Convert.ToSingle(solution.Millivolts));
+                SendCommand((byte)Command.ISE_CALIBRATE_SINGLE);
+
+                DelayHelper.DelayMilliseconds(IseMvMeasureTime, allowThreadYield: true);
+            }
+
+            throw new ArgumentOutOfRangeException("The value range is -1024 mV to 1024 mV");
```
krwq, line 125: this will now always throw (missing `else` - make sure to use braces around throw)

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+        {
+            if (solution.Millivolts >= -1024 && solution.Millivolts <= 1024)
+            {
+                WriteRegister(Register.ISE_SOLUTION_REGISTER, Convert.ToSingle(solution.Millivolts));
+                SendCommand((byte)Command.ISE_CALIBRATE_LOW);
+
+                DelayHelper.DelayMilliseconds(IseMvMeasureTime, allowThreadYield: true);
+            }
+
+            throw new ArgumentOutOfRangeException("The value range is -1024 mV to 1024 mV");
```
krwq, line 142: ditto

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+        {
+            if (solution.Millivolts >= -1024 && solution.Millivolts <= 1024)
+            {
+                WriteRegister(Register.ISE_SOLUTION_REGISTER, Convert.ToSingle(solution.Millivolts));
+                SendCommand((byte)Command.ISE_CALIBRATE_HIGH);
+
+                DelayHelper.DelayMilliseconds(IseMvMeasureTime, allowThreadYield: true);
+            }
+
+            throw new ArgumentOutOfRangeException("The value range is -1024 mV to 1024 mV");
```
krwq, line 159: ditto

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+        }
+
+        /// <summary>
+        /// Returns the firmware version of the device.
+        /// </summary>
+        /// <returns>Firmware version</returns>
+        public Version GetVersion()
+        {
+            return new Version((int)ReadFloat((byte)Register.ISE_VERSION_REGISTER), 0);
```
krwq, line 168: I'm surprised the version is float here... should this be integer?

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+        /// When there are high and low calibration points, the device will automatically use them to adjust readings.To disable dual-point adjustment, call ResetCalibration to remove all calibration data.
+        /// </summary>
+        /// <returns>The measured value (reading value) for calibrate the low value</returns>
+        public ElectricPotential GetCalibrateLowReading()
+        {
+            return new ElectricPotential(ReadFloat(Register.ISE_CALIBRATE_READLOW_REGISTER), UnitsNet.Units.ElectricPotentialUnit.Millivolt);
+        }
+
+        /// <summary>
+        ///  Resets all the stored calibration information
```
krwq, line 213: will it use factory default after or will it be unusable until calibrated? (you mind want to clarify)

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+                readLow.Millivolts >= -1024 && readLow.Millivolts <= 1024 &&
+                readHigh.Millivolts >= -1024 && readHigh.Millivolts <= 1024)
+            {
+                WriteRegister(Register.ISE_CALIBRATE_REFLOW_REGISTER, Convert.ToSingle(refLow.Millivolts));
+                WriteRegister(Register.ISE_CALIBRATE_REFHIGH_REGISTER, Convert.ToSingle(refHigh));
+                WriteRegister(Register.ISE_CALIBRATE_READLOW_REGISTER, Convert.ToSingle(readLow));
+                WriteRegister(Register.ISE_CALIBRATE_READHIGH_REGISTER, Convert.ToSingle(readHigh));
+            }
+
+            throw new ArgumentOutOfRangeException("The value range is -1024 mV to 1024 mV for refLow, refHigh, readLow and readLow");
```
krwq, line 244: ditto

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+            SendCommand((byte)Command.ISE_MEMORY_WRITE);
+        }
+
+        /// <summary>
+        /// Get Firmware version
+        /// </summary>
+        /// <returns>The version of the firmware</returns>
+        public Version GetFirmwareVersion()
+        {
+            return new Version(0, (int)ReadByte(Register.ISE_FW_VERSION_REGISTER));
```
krwq, line 297: should you swap arguments here? in the other Version you've read float, now you read byte (this one seems more correct). Likely you don't need to cast byte to int as it should upgrade the type automatically

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,415 @@
...
+
```
krwq, line 298: nit: extra enter

src/devices/UFireIse/UFireIse.cs
```diff
@@ -0,0 +1,391 @@
...
+            return v;
+        }
+
+        private int Magnitude(float x)
+        {
+            return float.IsNaN(x) || x == 0 ? 0 : (int)Math.Floor(Math.Log10(Math.Abs(x))) + 1;
+        }
+
+        private float RoundTotalDigits(float x, int? digits = 7)
```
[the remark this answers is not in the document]
krwq: didn't see the answer so unresolving

src/devices/UFireIse/UFireOrp.cs
krwq: likely it also should be something like:
`public ElectricPotential LastOxidationReducationPotential { get; private set; } = new ElectricPotential();`

note `Last` prefix to make sure users won't assume it will read

krwq: similar below for other public fields

src/devices/UFireIse/UFirePh.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        /// <summary>
+        /// pH (Power of Hydrogen) units measurement
+        /// </summary>
+        public float Ph = 0;
+
+        /// <summary>
+        /// pOH units measurement, for the relationship between pH and pOH see https://www.chem.purdue.edu/gchelp/howtosolveit/Equilibrium/Calculating_pHandpOH.htm#pOH
+        /// </summary>
+        public float Poh => Ph >= 0 ? Math.Abs(14 - Ph) : float.NaN;
```
krwq, line 27: you will need to remove this formula unless you find a link with MIT or other permissive license

krwq: I think removing a link to non-permissive page is ok and remove absolute value (since Ph is positive anyway) since this formula is trivial to figure out from just reading the definition

krwq: on the other hand from some sources seems relationship doesn't seem to always be true. it won't be for measuring i,e, acidity of soil. IMO we should remove this property at all

krwq, line 27: ditto

src/devices/UFireIse/UFireIse.cs
R10Rasmus: I have changed it back to a string, because the manufacturer do not provide any information about the format of the version number. This insures it supports any format it sends back.

src/devices/UFireIse/UFireIse.cs
R10Rasmus: I have changed it back to a string, because the manufacturer do not provide any information about the format of the version number. This insures it supports any format it sends back.

src/devices/UFireIse/UFireIse.cs
krwq: They seem to leave it as byte: https://github.com/u-fire/Isolated_ISE/blob/b2d58b40a7873cda98796f714272695bec365cea/python/RaspberryPi/uFire_ISE.py#L97

You can either assume it's major 0 and minor = whatever_you_read or swap minor and major, up to you but be consistent

krwq: @justind000 do you perhaps know if the version in ISE_VERSION_REGISTER is major or minor? How should we interpret it? (or doesn't really matter?)

src/devices/UFireIse/UFireIse.cs
krwq: uFire treats it as byte: https://github.com/u-fire/Isolated_ISE/blob/b2d58b40a7873cda98796f714272695bec365cea/python/RaspberryPi/uFire_ISE.py#L151 - change it to major or minor. regardless be consistent with other version API - I don't think float is correct

src/devices/UFireIse/UFireIse.cs
R10Rasmus: if we look at https://github.com/u-fire/ISE_Probe/blob/master/python/RaspberryPi/shell.py#L91 the GetVersion is major  and GetFirmwareVersion is minor. That is why I first made GetVersion return major  and GetFirmwareVersion  return minor.

krwq:
@R10Rasmus I think you've missed couple of comments (see not resolved issues - you might need to expand in some places) - other than that LGTM

src/devices/UFireIse/UFireIse.cs
R10Rasmus: @krwq I have pushed a new version where GetVersion return major and GetFirmwareVersion return  minor. Is this ok, or do I need to change it, sow both return major or minor?

src/devices/uFire_ISE/UFireIse.cs
R10Rasmus: manufacturer has updated it document with this delay, see https://www.ufire.co/docs/uFire_ISE/#use

src/devices/UFireIse/UFireIse.cs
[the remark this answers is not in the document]
R10Rasmus: RoundTotalDigits has been removed

R10Rasmus:
now the last not resolved issues has a question to @tannergooding and a question to @krwq

src/devices/UFireIse/UFireIse.cs
krwq: be consistent, either both minor of both major

krwq: not sure in that case... perhaps we should join both versions together into a single `GetVersion`? Honestly I'd be fine if you returned byte from both methods at this point 😄

krwq (approved):
LGTM, for Versions please either join them together into a single version or return byte

src/devices/UFireIse/UFireIse.cs
R10Rasmus: it now return byte

src/devices/UFireIse/UFireIse.cs
tannergooding: You want `TryWriteBytes`

R10Rasmus: I now use TryWriteBytes

krwq (approved):

krwq:
Thank you @R10Rasmus!

Ellerbach:
@R10Rasmus , thank you for the PR and all the work done to bring µFire to .NET IoT! Super happy to this one merged!

R10Rasmus:
@Ellerbach no problem, I hope in the future I can bring more of µFire products to .NET IoT.

PR #1137, opened 2020-07-19
Updating the sample into a full CLI application

Fortinbra:
Updating the sample application with more complicated examples, ported from the python WS2812b example library. These were ported from strandtest.py in https://github.com/jgarff/rpi_ws281x.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1137) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1137) before we can accept your contribution.<br/><br/>:x: Fortinbra [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1137)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1137) it.</sub>

src/devices/Ws28xx/Ws28xx.sln
```diff
@@ -1,13 +1,13 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26124.0
+# Visual Studio Version 16
```
joperezr, line 3: NIT: Can we please undo the changes into this file?

joperezr:
THanks a lot for your contribution @Fortinbra! We usually try to keep the samples as minimal as possible so that they are only showcasing the usage of the sensor, so I question a bit if this PR (although it makes the sample much more usable and cool) will add extra code that might overshadow the actual usage of the actual binding. I think some of the examples you are providing are really cool and useful but I wonder if it would make sense to either separate this into different samples or simplify it so that it is not necessarily a full command line app and instead just a sample program mostly just showcasing the binding usage, what do you think?

pgrawehr:
@joperezr We have a samples directory as well. Maybe that's the better place for extensive "showcases"?

src/devices/Ws28xx/Ws28xx.sln
Fortinbra: I'll undo those changes.

Fortinbra:
@joperezr and @pgrawehr  Thank you for your feedback, I'll simplify this into a simple app, removing the command line aspect. One thing I've always wanted to see in the samples are some that are a little more complicated, so show some of the more advanced capabilities of the bindings.

src/devices/Ws28xx/samples/Ws28xx.Sample.cs
```diff
@@ -13,47 +14,157 @@ internal class Program
...
+            }
+
+        }
+
+        private static void CancelHandler(object sender, ConsoleCancelEventArgs e)
+        {
+            BitmapImage img = s_neo.Image;
             img.Clear();
-            while (true)
+            s_neo.Update();
```
joperezr, line 64: NIT: should we also add Dispose code here in the Cancellation handler?

joperezr: Oh I just noticed that apparently this device binding isn't disposable which is wrong as the SPIDevice it holds should be disposed from here. Do you want to address that on this PR as well as it should be a straight forward change? If not I can add a follow up to do that.

Fortinbra: > Oh I just noticed that apparently this device binding isn't disposable which is wrong as the SPIDevice it holds should be disposed from here. Do you want to address that on this PR as well as it should be a straight forward change? If not I can add a follow up to do that.

Let's add a follow up for it. Thanks!

PR #1138, opened 2020-07-22
Update Uln2003.cs

### LINKED ISSUE #1133: Uln2003 - wrong time of steps
**Describe the bug**
The time of step (_stepMicrosecondsDelay) has one little mistake. The variable _steps is used to calculate the delay. Therefore, the delay is calculated well only for _steps equal to full revolutions.

**Steps to reproduce**
Try to set RPM to 1, and do only one step.
```csharp
Uln2003 motor = new Uln2003(bluePin, pinkPin, yellowPin, orangePin)
motor.RPM = 1;
motor.Step(1);
```
No motor or driver is needed to see this error, just check the execution time of the function.

**Expected behavior**
One step should take about 15ms (half step = 60000/4096) or 30ms (full step = 60000/2048).

**Actual behavior**
Regardless of the value of the steps passed to the step function, the execution time for RPM = 1 will be one minute

**Versions used**
Cloned repository

**Possible solution**
Add a new variable with number of steps required to full rotation?
steps_fix

I'm new to C # and github so forgive me if the description doesn't match the standards.

wsad4ryba:
Fixes #1133 
- Correct calculation of time to wait between steps

### DIFF
--- a/src/devices/Uln2003/Uln2003.cs
+++ b/src/devices/Uln2003/Uln2003.cs
@@ -22,6 +22,7 @@ namespace Iot.Device.Uln2003
         private int _steps = 0;
         private int _engineStep = 0;
         private int _currentStep = 0;
+        private int _stepsToRotate = 4096;
         private StepperMode _mode = StepperMode.HalfStep;
         private bool[,] _currentSwitchingSequence = _halfStepSequence;
         private bool _isClockwise = true;
@@ -102,12 +103,15 @@ namespace Iot.Device.Uln2003
                 {
                     case StepperMode.HalfStep:
                         _currentSwitchingSequence = _halfStepSequence;
+                        _stepsToRotate = 4096;
                         break;
                     case StepperMode.FullStepSinglePhase:
                         _currentSwitchingSequence = _fullStepSinglePhaseSequence;
+                        _stepsToRotate = 2048;
                         break;
                     case StepperMode.FullStepDualPhase:
                         _currentSwitchingSequence = _fullStepDualPhaseSequence;
+                        _stepsToRotate = 2048;
                         break;
                 }
             }
@@ -136,7 +140,7 @@ namespace Iot.Device.Uln2003
             _stopwatch.Restart();
             _isClockwise = steps >= 0;
             _steps = Math.Abs(steps);
-            _stepMicrosecondsDelay = RPM > 0 ? 60 * 1000 * 1000 / _steps / RPM : StepperMotorDefaultDelay;
+            _stepMicrosecondsDelay = RPM > 0 ? 60 * 1000 * 1000 / _stepsToRotate / RPM : StepperMotorDefaultDelay;
             _currentStep = 0;
 
             while (_currentStep < _steps)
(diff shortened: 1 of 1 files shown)


src/devices/Uln2003/Uln2003.cs
```diff
@@ -136,7 +140,7 @@ public void Step(int steps)
             _stopwatch.Restart();
             _isClockwise = steps >= 0;
             _steps = Math.Abs(steps);
-            _stepMicrosecondsDelay = RPM > 0 ? 60 * 1000 * 1000 / _steps / RPM : StepperMotorDefaultDelay;
+            _stepMicrosecondsDelay = RPM > 0 ? 60 * 1000 * 1000 / _stepsToRotate / RPM : StepperMotorDefaultDelay;
```
joperezr, line 143: >_steps [](start = 66, length = 6)

Do we also have to update three lines bellow the while loop to use _stepsToRotate instead of _steps? Do we still need _steps at all?

wsad4ryba: No for first question.
Yes for second. We need _steps to be able to move the motor for fraction of rotation.
If we use _stepsToRotate in while loop inside Step, we would end up with method to always rotate once.
Maybe we can add a new method Rotate:
```csharp
public void Rotate(int rotations)
        {
            Step(rotations * _stepsToRotate); // If the rotations is negative, the motor moves in the reverse direction
        }
```
Thanks to that, people who want to use step motor for multiple rotations will have an easier task.
I and probably some other people want to use step motor to precise alignment, so we need Step with _steps in while loop.

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1138) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1138) before we can accept your contribution.<br/><br/>:x: wsad4ryba [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1138)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1138) it.</sub>

src/devices/Uln2003/Uln2003.cs
```diff
@@ -63,7 +65,8 @@ public class Uln2003 : IDisposable
         /// <param name="pin4">The GPIO pin number which corresponds pin D on ULN2003 driver board.</param>
         /// <param name="controller">The controller.</param>
         /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
-        public Uln2003(int pin1, int pin2, int pin3, int pin4, GpioController controller = null, bool shouldDispose = true)
+        /// <param name="stepsToRotate">Amount of steps needed to rotate motor once in HalfStepMode.</param>
+        public Uln2003(int pin1, int pin2, int pin3, int pin4, GpioController controller = null, bool shouldDispose = true, int stepsToRotate = 4096)
```
wsad4ryba, line 69: Thanks to new parameter we could use other step motors with different amount of steps needed to rotate.

wsad4ryba:
I know it's not pretty, but it's probably the easiest way to achieve new functionality.
Are you going to merge this?

joperezr:
Sorry forgot to hit merge 😂

wsad4ryba:
😂

Issue #1139, opened 2020-07-23
PN5180 ListenToCardIso14443TypeA hangup


arnoud-koolecontrols:
Don't know if this is the place to post problems with device bindings.
Please let me know if it should be posted somewhere else.

**PN5180 ListenToCardIso14443TypeA hangup**
I am looking into the device binding with the PN5180 to read and write some Mifare 1k classic cards.
I noticed that when a Mifare classic card is put in and out of the field the PN5180 stops working and the Busy line of the PN5180 stays high. 

**Steps to reproduce**
Call the ListenToCardIso14443TypeA  in a loop and hold a Mifare classic 1k card with a UID4 in front of the reader and away again. Repeat this until the hang-up occurs. 

**Possible fix**
at line 1036 in Pn5180.cs the anticollision is done. 
```csharp
for (int i = 0; i < 3; i++)
{
    // Select SL1
    uidSak[0] = (byte)(0x93 + i * 2);
    // NVB = Number of valid bits
    uidSak[1] = 0x20;
    SendDataToCard(uidSak.Slice(0, 2));
    // Read 5 bytes sak. Byte 1 will tell us if we have the full UID or if we need to read more
    ReadDataFromCard(sakInterm.Slice(0, 5));
```
To prevent the hang-up a check if 5 bytes are received would be enough.
```csharp
for (int i = 0; i < 3; i++)
{
    // Select SL1
    uidSak[0] = (byte)(0x93 + i * 2);
    // NVB = Number of valid bits
    uidSak[1] = 0x20;
    SendDataToCard(uidSak.Slice(0, 2));
    (numBytes, _) = GetNumberOfBytesReceivedAndValidBits();
    if (numBytes != 5)
    {
        return false;   //this will prevent a hangup
    }
    // Read 5 bytes sak. Byte 1 will tell us if we have the full UID or if we need to read more
    ReadDataFromCard(sakInterm.Slice(0, 5));
```

krwq:
cc: @Ellerbach

arnoud-koolecontrols:
After fixing this hang-up I also notice the ListenToCardIso14443TypeA stops working after reading some blockdata of a mifare classic 1k card.

After authenticating for a block the MFC_CRYPTO_ON bit in the SYSTEM_CONFIG is set high. 

**Possible fix**
at line 1022 in Pn5180.cs the flags are configured

```csharp
// Clears all interrupt
[...]

PR #1142, opened 2020-07-27
PN5180 ListenToCardIso14443TypeA hangup fix

arnoud-koolecontrols:
Line 1010: If MifareAuthenticate is done the MFC_CRYPTO_ON flag in the SYSTEM_CONFIG register is turned on. This should be turned of again before sending the REQALL

Line 1043: If a card is pulled out of the field while sending a part of it's UID it will result in a sak ending with values 0xFF. These are send back resulting in a hangup of the PN5180. Added a check to stop when no 5 bytes are received.

Fixes mentioned in #1139


Ellerbach:
@krwq, I am good to go with those modifications. This correct properly the bug described and make sure that if the card is pulled out during the peering, the process will be stopped.
Thanks @arnoud-koolecontrols for the PR.

krwq:
Awesome, thanks for reviewing @Ellerbach!

PR #1147, opened 2020-08-03
delete error code from readme

HumJ0218:


krwq:
Seems one of the devices on Linux queue (ddvsorpil022) might have some issue with I2C, I'll see if we can get it rebooted. Windows failure seems a flaky test. cc: @joperezr

PR #1148, opened 2020-08-04
Bindings for MBI5027 and SN74HC595

richlander:
- Adds generic shift register binding -- tested with MBI5027, MBI5168, and SN74HC595.
- Adds [MBI5027](http://archive.fairchip.com/pdf/MACROBLOCK/MBI5027.pdf) binding, based on the generic binding
- Re-implements the SN74HC595 binding in terms (base class) of the generic binding
- Moves SN74HC595 to `Iot.Device.Multiplexing` namespace (breaking change)
- Changes the location of the pinmapping type for SN74HC595 (breaking change)
- Changes the order of pins in the pinmapping type for SN74HC595 (breaking change)

I learned a bunch by adding a second shift register implementation, which caused me to re-thinking the choices I made with the SN74HC595. It has not gone out in a stable NuGet package, so I think we can/should make these changes.

Existing SN74HC595 type: https://github.com/dotnet/iot/tree/master/src/devices/Sn74hc595


HumJ0218:
MBI5027 is sink driver, so ```sr.ShiftBit(1)``` means "pull down the output pin". This makes the electrical characteristics(current direction) different from the 74595. Should it be changed to make the two classes compatible with electrical characteristics?

HumJ0218:
or I suggest adding an enumeration flag to base class to identify the output characteristics of the component (source current, sink current, pull-up, pull-down, etc.)

or put these definitions in `Gpio.PinValue` and use then here?

richlander:
> MBI5027 is sink driver, sr.ShiftBit(1) means "pull down the output pin". This makes the electrical characteristics(current direction) different from 74595.

You are totally right. The MBI5027 and SN74HC595 are identical as it relates to shift register operation but are exactly opposite as it relates to electrical signal / circuit design. From a code standpoint, I don't think it matters. I'm far from an expert on electronics, however, the code I wrote works and was the most intuitive outcome from reading the datasheet. It would make sense to do something different if writing code could make the MBI5027 work like the SN74HC595, or vice versa. Different code will not have that result. IMO, the circuit you create needs to match the electrical model of the specific shift register, and the code does not.

The MBI5027 datasheet doesn't make reference to pull up/down pin modes in reference to the latch pin. I guess that's another way of thinking about this. The data and storage register clock pins are abstracted from the electrical model of the chip, with every shift register I've seen. It makes no sense for them to be pull up/down, as they are not directly connected to the outputs. It's the latch and the output enable pins that are more directly attached to the outputs.

I guess the simple questions are: 

- Is the code in the PR wrong?
- Would exposing a different API provide consuming code with a better result?

Similarly, do you think it is wrong to provide the `ShiftRegister` class as a universal solution for shift registers, including for use with MBI5027?

HumJ0218:
Leave the circuit connection to the user, the code only completes the logical function, I think there is no problem. I didn't understand the problem before, but now I do. Thank you for your explanation.

src/devices/ShiftRegister/ShiftRegister.cs
```diff
@@ -0,0 +1,260 @@
...
+            // value reset to low in preparation for next use.
+            _controller.Write(_latch, 0);
+        }
+
+        /// <summary>
+        /// Switch output register to high-impedance state.
+        /// Temporarily disables register outputs, but does not delete values.
+        /// Requires use of GPIO controller.
+        /// </summary>
+        public void OutputDisable()
```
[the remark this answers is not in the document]
richlander: This use case seems out of scope of the design guidelines for a property.

src/devices/Mbi5027/Mbi5027PinMapping.cs
```diff
@@ -0,0 +1,67 @@
...
+            Clk   = 20    // SR pin 3   -- storage register clock
+            LE    = 21    // SR pin 4   -- latch / publish storage register
+            OE    = 12    // SR pin 21  -- blank
+            Sdo   = 25    // SR pin 22  -- clear
+        */
+
+        /// <summary>
+        /// Serial data in pin.
+        /// </summary>
+        public int Sdi { get; set; }
```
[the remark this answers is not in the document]
richlander: Yes. I have been going back and forth on that. Where I landed is using the datasheet names for the concrete shift registers and longer names the more abstract `ShiftRegister` class. I can adopt some variant of these names for `ShiftRegister`. I prefer just `Latch` but the rest look good to me. Fair?

src/devices/ShiftRegister/ShiftRegister.cs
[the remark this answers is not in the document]
Ellerbach: Why? If the state doesn't change once set, then it's a property. You can set to enable or disable. Only 2 states and they can't randomly change. You can even keep track of it and return the set value.

src/devices/Mbi5027/Mbi5027PinMapping.cs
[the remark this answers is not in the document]
Ellerbach: I always prefer long names as it's then easy to find the abbreviation. I also realized not all datasheets uses the same short names. So I ended up having long names and speaking names rather then abbreviation.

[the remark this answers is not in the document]
richlander: Sure, but this binding is for a specific device with a specific datasheet. This is for the pinmapping. IMO, matching the pinmapping to the datasheet is the most intuitive approach. I cannot think of anything else being easier.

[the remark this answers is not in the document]
Ellerbach: OK, fair enough

src/devices/ShiftRegister/ShiftRegister.cs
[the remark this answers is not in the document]
richlander: But not changing it isn't the model. You can change change Output Enable/disable as many times you want. They cannot randomly change, but you can change them from code.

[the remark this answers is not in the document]
richlander: I talked to @terrajobst. He said that this one is OK. I'll implement your suggested pattern.

src/devices/Device-Index.md
```diff
@@ -35,6 +35,7 @@
 * [Mcp23xxx - I/O Expander device family](Mcp23xxx/README.md)
 * [Mcp25xxx device family - CAN bus](Mcp25xxx/README.md)
 * [Mcp3428 - Analog to Digital Converter (I2C)](Mcp3428/README.md)
+* [Mbi5027 -- 16-bit shift register](Mbi5027/README.md)
```
Ellerbach, line 38: Device-Index .md is generated automatically. What you need is the category.txt file in the same directory as your csproj

src/devices/Mbi5027/Mbi5027.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Enable open/short eror detection mode.
+        /// Requires use of GPIO controller.
+        /// </summary>
+        public void EnableDetectionMode()
+        {
+            if (GpioController is null || _pinMapping.Clk == 0 || _pinMapping.OE == 0 || _pinMapping.LE == 0)
```
Ellerbach, line 51: You may want to check anything <= 0?

src/devices/ShiftRegister/ShiftRegister.cs
```diff
@@ -0,0 +1,260 @@
...
+        /// <summary>
+        /// Initialize a new shift register connected through GPIO.
+        /// </summary>
+        /// <param name="pinMapping">The pin mapping to use by the binding.</param>
+        /// <param name="bitLength">Bit length of register, including chained registers.</param>
+        /// <param name="gpioController">The GPIO Controller used for interrupt handling.</param>
+        /// <param name="shouldDispose">True (the default) if the GPIO controller shall be disposed when disposing this instance.</param>
+        public ShiftRegister(ShiftRegisterPinMapping pinMapping, int bitLength, GpioController gpioController = null,  bool shouldDispose = true)
+        {
+            if (gpioController == null)
```
Ellerbach, line 41: To simply the next lines:
```csharp
_shouldDispose = shouldDispose || ( GpioController == null);
_controller = gpioController ?? new GpioController();
```

src/devices/ShiftRegister/ShiftRegister.cs
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        /// <summary>
+        /// Switch output register low-impedance state.
+        /// Enables register outputs with existing values.
+        /// Requires use of GPIO controller.
+        /// </summary>
+        public void OutputEnable()
+        {
+            if (_controller is null || _pinMapping.OE == 0)
```
Ellerbach, line 204: Same comment as somewhere else, maybe check <=0?
And thinking of it, 0 is a valid GPIO. So for all this, should be <0

richlander:
I believe that I have resolved all the feedback now. We can merge now if folks are good with the changes.

Ellerbach (approved):

Ellerbach:
> I believe that I have resolved all the feedback now. We can merge now if folks are good with the changes.

Seems like yes :-) Looks good to me now! thanks! Giving a chance for @krwq to have a look.

richlander:
FYI: I went with `OutputEnable` as the property name instead of just `Output`. The former matches the datasheet and made it more obvious, at least to me, what the bool related to. In any case, it is now a property, based on your feedback (@Ellerbach). Thanks for that.

src/devices/Mbi5027/Mbi5027.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Enable open/short eror detection mode.
+        /// Requires use of GPIO controller.
+        /// </summary>
+        public void EnableDetectionMode()
+        {
+            if (GpioController is null || _pinMapping.Clk < 0 || _pinMapping.OE < 0 || _pinMapping.LE < 0 || _pinMapping.Sdo < 0)
```
krwq, line 51: I think only -1 should be considered invalid although I'm fine with `<0`

src/devices/Mbi5027/Mbi5027.cs
```diff
@@ -0,0 +1,186 @@
...
+            ChangeModeSignal(0, 0);
+            ChangeModeSignal(1, 0);
+            ChangeModeSignal(1, 0);
+            ChangeModeSignal(1, 0);
+            GpioController.Write(_pinMapping.OE, 0);
+        }
+
+        private void ChangeModeSignal(PinValue oe, PinValue le)
+        {
+            GpioController.Write(_pinMapping.OE, oe);
```
krwq, line 172: is there any time constraint we should work with? i.e. should there be some delay here?

src/devices/Mbi5027/Mbi5027PinMapping.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace Iot.Device.Multiplexing
+{
+    /// <summary>
+    /// Represents pin bindings for the MBI5027.
+    /// </summary>
+    public struct Mbi5027PinMapping
+    {
+        /// <param name="sdi">Serial data in pin</param>
+        /// <param name="clk">Shift register pin (shift to data register)</param>
+        /// <param name="le">Register clock pin (latch)</param>
```
krwq, line 17: should we call it just `latch` or `latchEnabled` (same for oe -> `outputEnabled`)

src/devices/Mbi5027/Mbi5027PinMapping.cs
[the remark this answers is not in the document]
krwq: I see you've talked about this already. I also prefer expanded names, comments can have datasheet names

src/devices/Mbi5027/Mbi5027.cs
richlander: This is a great question. It works as-is. The problem is that we don't have any delays that are at all meaningful. For example, I could add a delay for 1ms, but that would be overkill, by more than 3 orders of magnitude. The chip runs at 25Mhz clock frequency. What's the guidance in that situation?

src/devices/Mbi5027/Mbi5027PinMapping.cs
richlander: I stuck with the names specified in the datasheet. That seemed the most intuitive/sensible to me. I picked better names in the `ShiftRegister` samples, since there is no datasheet for it. I'll update that one to use `LatchEnable`, which I agree is better.

src/devices/Mbi5027/Mbi5027.cs
richlander: Checking for `-1` in particular seems not all that useful. It's not a special number. Why is it different or more likely than `-2`?

src/devices/Mbi5027/Mbi5027PinMapping.cs
richlander: Here are the two data sheets I was targeting. Looking at them makes it more obvious why I targeted the specific pin names.

MBI5027
image

SN74HC595
image

src/devices/Mbi5027/Mbi5027.cs
krwq: we picked `-1` by convention but `< 0` is fine as well

src/devices/Mbi5027/Mbi5027.cs
krwq: let's leave as is for now, if we find this problematic then we'll figure it out

src/devices/Mbi5027/Mbi5027PinMapping.cs
```diff
@@ -0,0 +1,77 @@
...
+            Clk = clk;
+            LE = le;
+            OE = oe;
+            Sdo = sdo;
+        }
+
+        /// <summary>
+        /// Minimal pin bindings for the MBI5027.
+        /// </summary>
+        public static Mbi5027PinMapping Minimal => new Mbi5027PinMapping(16, 20, 21);
```
krwq, line 32: this should be a method or return a cached instance (same below)

krwq: actually if this is a struct then I'm ok with this

src/devices/ShiftRegister/ShiftRegister.cs
```diff
@@ -0,0 +1,243 @@
...
+        /// </summary>
+        public void ShiftByte(byte value, bool latch = true)
+        {
+            if (_spiDevice is object)
+            {
+                _spiDevice.WriteByte(value);
+                return;
+            }
+
+            for (int i = 0; i < 8; i++)
```
krwq, line 143: this part looks like software SPI. Is it possible to use SoftwareSpi in a specific mode (CPHA=false, CPOL=false; or mode=0)  for this?

If so we could remove part of the abstraction like (UsesSpi, UsesGpio). BitLength can also be set on both Software and HardwareSpi (through settings)

src/devices/Mbi5027/Mbi5027.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        /// <summary>
+        /// Initialize a new shift register connected through GPIO.
+        /// </summary>
+        /// <param name="pinMapping">The pin mapping to use by the binding.</param>
+        /// <param name="bitLength">Bit length of register, including chained registers. Default is 16 bits.</param>
+        /// <param name="gpioController">The GPIO Controller used for interrupt handling.</param>
+        /// <param name="shouldDispose">Option (true by default) to dispose the GPIO controller when disposing this instance.</param>
+        public Mbi5027(Mbi5027PinMapping pinMapping, int bitLength = 16, GpioController gpioController = null,  bool shouldDispose = true)
+        : base(new ShiftRegisterPinMapping(pinMapping.Sdi, pinMapping.Clk, pinMapping.LE, pinMapping.OE), bitLength, gpioController, shouldDispose)
```
krwq, line 28: you could likely still do that with SoftwareSpi, you'd just need to manually pass in GpioController

src/devices/Mbi5027/Mbi5027PinMapping.cs
richlander: Ya. I don't think there is much value of returning a cached instance (even if it was a class). Isn't there a minimum number of uses to expect to merit creating a cached instance? I suspect this property would get accessed <=1 times per app lifetime.

src/devices/ShiftRegister/ShiftRegister.cs
richlander: I'm not seeing a way to do what you are suggesting.

GPIO access is about changing one bit at a time. SPI is about updating on byte at a time. They are different patterns. This code allows you, as a secondary pattern, to update a byte at a time from GPIO.

richlander: For example, the `Mbi5027` needs to interact with `OE`, `Sdo`, and `Le` for error detection. `OE` and `Sdo` would not be used by  `SoftwareSpi`, but `Le` would be. It would seem odd to share the same pin across these two different use cases.

krwq (approved):

PR #1151, opened 2020-08-10
Adding CCS811 gas sensor

Ellerbach:
Add CCS811 Gas Sensor
- Full sensor implementation for measurement, temperature and humidity adjustment, baseline
- Interruption with event support, threshold
- Full documentation including specific I2C settings for Raspberry Pi and other I2C devices with lack of I2C stretching support
- Full example to test all the features thru simple menus
- Example using as well external GPIO and I2C chipset like FT4222

PR #1154, opened 2020-08-13
Ccs811 orthograph adjustment

Ellerbach:
- Fix orthograph issue in PR #1151 Volatil -> Volatile


danmosemsft:
Thanks. @richlander and I were chatting, he mentioned that it might be good soon to do an all-up pass through the API for consistent naming etc, and a review of the public abstractions.

krwq:
@danmosemsft makes sense, for now we have been doing sweep passes whenever we have identified something specific. For System.Device.Gpio we are much more careful with naming to make sure we don't introduce any breaking changes. For Iot.Device.Bindings (like i.e. this binding) we have been allowing some breaking changes if they were justified but we've tried to avoid them as much as possible (also depending on how commonly given API is used).

krwq:
unrelated failures again

Ellerbach:
> Thanks. @richlander and I were chatting, he mentioned that it might be good soon to do an all-up pass through the API for consistent naming etc, and a review of the public abstractions.

And maybe good as well to give clear naming recommendations for the function variables and naming as well. You get the experience thru PR. Maybe good to have couple of example like do/don't somewhere. I4ll think of it and will propose something.

PR #1159, opened 2020-08-20
Documentation adjustment with do and don't

Ellerbach:
Adjusting documentation for binding:
- adding do and don't as examples
- adding recommendation
- adjusting UnitsNet usage in the device convention


Documentation/Devices-conventions.md
```diff
@@ -32,9 +32,11 @@ There is no specific guidelines what simple and main scenario is but you should
 
 ## Units
 
-In the most common case units should match [International System of Units (aka. SI)](https://en.wikipedia.org/wiki/International_System_of_Units) with some exceptions:
+Use [UnitsNet](https://github.com/angularsen/UnitsNet) whenever it is possible on any public functions, event or properties. This supports a lot of different type of units.
```
krwq, line 35:
```suggestion
Use [UnitsNet](https://github.com/angularsen/UnitsNet) whenever it is possible on any public functions, event or properties. This supports a lot of different types of units. 
```

src/devices/README.md
```diff
@@ -229,7 +229,11 @@ Bindings must:
 
 * include a .NET Core project file for the main library.
 * include a descriptive README, with a fritzing diagram.
+  * Best if you include the coverage of the binding. If you did not implement all features, mention what has been implemented and what is not. It is ok not to implement every features but it's important to know what is missing
+  * Refer to multiple manufacturer and add links where the binding can be found. Sometimes, they have different names depending on where you can find them while it's the same chipset used or compatible ones 
 * include a buildable sample (layout will be described below).
+  * Best if you include a sample that covers usage of all available fonctions and properties
```
krwq, line 235:
```suggestion
  * Best if you include a sample that covers usage of all available functions and properties
```

src/devices/README.md
```diff
@@ -251,6 +255,138 @@ iot/
           Foo.Tests.cs
 ```
 
+When creating a new binding, please follow the recommendations [from corefx coding guideliness](https://github.com/dotnet/corefx/tree/master/Documentation#coding-guidelines). On top, we recommend to use very descriptive namings, self explanatory names. Please make sure you don't use abbreviation. Describe any unit a sensor may use as well as the usual range of values. Check as well [the devices convention](../../Documentation/Devices-conventions.md). You will find below couple of examples of do and don't.
```
krwq, line 258: you might want to update that with runtime coding guideliness equivalent somewhere from here:
https://github.com/dotnet/runtime/tree/master/docs/coding-guidelines

src/devices/README.md
```diff
@@ -251,6 +255,138 @@ iot/
...
+
+### Using units
+
+| Do | Don't |
+|---|---|
+| Always use [UnitsNet](https://github.com/angularsen/UnitsNet) when possible: | Don't implement you own: |
```
krwq, line 264:
```suggestion
| Always use [UnitsNet](https://github.com/angularsen/UnitsNet) when possible | Don't implement your own |
```

src/devices/README.md
```diff
@@ -251,6 +255,138 @@ iot/
...
+|---|---|
+| Always add range of the units in the XML description|Never leave a unit without a full range description|
+| ```/// <param name="equivalentCO2">The equivalent CO2 (eCO2) output range for CCS811 is from 400ppm up to 29206ppm.</param>``` | ```/// <param name="equivalentCO2">The equivalent CO2 (eCO2)</param>``` |
+| Even when using the UnitsNet, make sure you give the usual range of a sensor value. | This makes it difficult to understand  which range the sensor is using. |
+
+### Naming of functions
+
+| Do |
+|---|
+| When reading a sensor and outputting data, use ```Try``` for example: |
```
krwq, line 284:
```suggestion
| When reading a sensor can fail prefix a method with ```Try``` for example: |
```

src/devices/README.md
```diff
@@ -251,6 +255,138 @@ iot/
...
+| ```public bool TryGetTemperature(out Temperature temperature)``` |
+| This makes it easy to understand that if you read and the retruned value is ```false``` the reading is not correct. And this means as well this function should not raise an exception when trying to read the sensor data. |
```
krwq, line 286: You might want to add that exception in such cases are still ok in some cases, i.e.: IO errors or user error (basically if you have to try .. catch the whole method then it's an anti-pattern and likely you should just throw)

src/devices/README.md
```diff
@@ -251,6 +255,138 @@ iot/
...
+        var mode = ReadRegister(Register.MEAS_MODE);
+        // Clear previous mode
+        mode = (byte)(mode & 0b1000_1111);
+        mode = (byte)(mode | (((byte)(value)) << 4));
+        WriteRegister(Register.MEAS_MODE, mode);
+    }
+}
+```
+
+| Do | Don't |
```
krwq, line 321: I'd put this above the `Try...` rule and also add that it is acceptable to use abbreviation for commonly known abbreviations (i.e. Html/Xml)

src/devices/README.md
```diff
@@ -251,6 +255,138 @@ iot/
...
+/// register and the Algorithms will use the new value in its
+/// calculations(until it adjusts it as part of its internal Automatic
+/// Baseline Correction). Please refer to documentation to understand when to restore a
+/// previous baseline: https://www.sciosense.com/wp-content/uploads/2020/01/Application-Note-Baseline-Save-and-Restore-on-CCS811.pdf</remarks>
+
+```
+
+| Do |
+|---|
+| Always reference specific elements with a poiunter documentation/reference to a specific part of the docuementation. This does including specific timing wait, intiailization phase and anything that would trigger a question like ```why this wait?``` or ```Why this function has to be called 2 times in a row```. If the documentation is the same for all the sensor implementation, refer the documentation once at the beginning and then mention the page. |
```
krwq, line 343:
```suggestion
| Always reference specific elements with a pointer documentation/reference to a specific part of the documentation. This does including specific timing wait, initialization phase and anything that would trigger a question like ```why this wait?``` or ```Why this function has to be called 2 times in a row```. If the documentation is the same for all the sensor implementation, refer the documentation once at the beginning and then mention the page. |
```

src/devices/README.md
```diff
@@ -251,6 +255,138 @@ iot/
...
+_controller.Write(_pinWake, PinValue.Low);
+// Doc says wait 50 micro seconds, page 58 https://urltothedoc
+DelayHelper.DelayMicroseconds(50, true);
+```
+
+### GpioController
+
+| Do|
+|---|
+| In a constructor if you are using a ```GpioController``` make sure you are using as well a bollean ```_shouldDispose```. Some platforms or some specific doungles like FT4222 does allow the creation of one controller so if used with multiple bindings, it can't be disposed just like that. In the ```Dispose``` function, make sure you close the possible open pins if ```_shouldDispose``` is false. |
```
krwq, line 356:
```suggestion
| In a constructor if you are using a ```GpioController``` make sure you are using as well a boolean ```_shouldDispose```. Some platforms or some specific doungles like FT4222 does allow the creation of one controller so if used with multiple bindings, it can't be disposed just like that. In the ```Dispose``` function, make sure you close the possible open pins if ```_shouldDispose``` is false. |
```

src/devices/README.md
```diff
@@ -251,6 +255,138 @@ iot/
...
+| Example of an correct implementation: |
+
+```csharp
+private GpioController _controller;
+private bool _shouldDispose;
+private int _thePinToUse;
```
krwq, line 362: `_pinToUse`

src/devices/README.md
```diff
@@ -251,6 +255,138 @@ iot/
...
+
+public MyBinding(int thePinToUse, GpioController controller = null, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical, bool shouldDispose = true)
+{
+  _shouldDispose = gpioController == null ? true : shouldDispose;
```
krwq, line 366: `_shouldDispose = gpioController == null || shouldDispose;`

src/devices/README.md
```diff
@@ -251,6 +255,138 @@ iot/
...
+  }
+}
+```
+
+### Licences
+
 We are currently not accepting samples that rely on native libraries for hardware interaction. This is for two reasons: we want feedback on the System.Device API and we want to encourage the use of 100% portable .NET solutions. If a native library is used to enable precise timing, please file an issue so that we can discuss your proposed contribution further.
 
 We will only accept samples that use the MIT or compatible licenses (BSD, Apache 2, ...). We will not accept samples that use GPL code or were based on an existing GPL implementation. It is critical that these samples can be used for commercial applications without any concern for licensing.
+
```
krwq, line 391: extra enter

src/devices/README.md
```diff
@@ -251,6 +255,137 @@ iot/
...
+| ```/// <param name="equivalentCO2">The equivalent CO2 (eCO2) output range for CCS811 is from 400ppm up to 29206ppm.</param>``` | ```/// <param name="equivalentCO2">The equivalent CO2 (eCO2)</param>``` |
+| Even when using the UnitsNet, make sure you give the usual range of a sensor value. | This makes it difficult to understand  which range the sensor is using. |
+
+### Naming of functions
+
+| Do |
+|---|
+| When reading a sensor can fail prefix a method name with ```Try``` for example: |
+| ```public bool TryGetTemperature(out Temperature temperature)``` |
+| This makes it easy to understand that if you read and the retruned value is ```false``` the reading is not correct. And this means as well this function should not raise an exception when trying to read the sensor data if the data is not valid for example. You still can get exceptions like IO Errors or user errors. For example if the function takes parameters with a specific range, and the user is out of this range, then an Argument Exception can be raised. |
```
krwq, line 286:
```suggestion
| This makes it easy to understand that if you read and the returned value is ```false``` the reading is not correct. And this means as well this function should not raise an exception when trying to read the sensor data if the data is not valid for example. You still can get exceptions like IO Errors or user errors. For example if the function takes parameters with a specific range, and the user is out of this range, then an Argument Exception can be raised. |
```

krwq (approved):

joperezr:
This is all goodness, thanks @Ellerbach

Issue #1093, opened 2020-05-29
Request a way to keep the camera open in VideoDevice


HumJ0218:
There are many cameras that grab pictures immediately after initialization, and there will be an unstable picture situation, ex. a full-green or black frame.

How about adding a method to keep the camera open. So users can make a delay before capture a image to avoid the green frame?

@ZhangGaoxing

krwq:
I wouldn't mind if camera stayed open for whole lifetime of the class

HumJ0218:
@krwq Thank you for your explanation. My understanding is that the ABCDE five lines are inputs from a 5-32 decoder to select the current pixel line. I have this doubt because I bought a 128x64 RGB LED screen with a label `HUB75` besides interface and only have ABC three lines as line selection lines. After some research I found that these three lines are inputs of a group of shift registers. The outputs lead the pixel lines.

My question is whether there are multiple HUB75 implementations, and they are not compatible with each other.

krwq:
I'm not sure - I suspect people implement everything specially for their devices and once they get it to work they just don't care about making it compatible with other versions (unless they have both)

Ellerbach:
@HumJ0218 , did you have a chance to look at the PR #1163? 
This should solve your issue. Please let us know so we can work on the PR and merge it. Let me know if you need any help.

HumJ0218:
@Ellerbach OK. I'll try it on weekend

Ellerbach:
[Triage] Regarding the request and the PR #1163, this seems to be addressed.

PR #1163, opened 2020-08-22
Adding continuous recording for SoundDevice and VideoDevice

Ellerbach:
- Add continuous recording for SoundDevice by chunk of 1 second to a file
- Add continuous capture for VideoDevice
    - This should address issue #1093
    - Added a simple MJPEG ASP.NET Core sample to illustrate advanced usage
- Adjusted documentation for SoudDevice
- Adjusted documentation for VideoDevice


pgrawehr:
What hardware/OS did you test this with? I guess for once it wasn't a Pi, because that one doesn't have a built-in recording device.

Ellerbach:
> What hardware/OS did you test this with? I guess for once it wasn't a Pi, because that one doesn't have a built-in recording device.

For the sound part a Pi Zero with mono :-) See: https://github.com/Ellerbach/Nabaztag.Net/tree/master/Nabaztag.Server and more generally the full project. The hardware is using a custom made shield with a custom build sound card having 2 mics and which does not support unmute and other commands like this.
For the video part, a Pi 3 with a very old webcam. I deploy a more complex version of the MJPEG server in a container using Azure IoT Edge and a connection to Azure IoT Hub with the twins (for the timezone property). So output is like this. The webcam is in my garden in a green house:
serre

HumJ0218:
@Ellerbach I've tried the project `MjpegStream`, but it didn't work out. Here are my steps:

1. cloned your code with VS. `Ellerbach/iot/audiovideo`
2. found the project `MjpgStream` and tried to publish but was prompted for missing the required development environment. (I don't have .NET 5.)
3. deleted `global.json`. (use .NET Core 3.1)
4. `.UseUrls("http://*:5000")`
5. publish as linux_arm. Sent to raspberrypi 4 (with a HDMI to CSI module instead of camera. They should be the same)
6. access "http://ip:5000/image" with browser

then the request was jammed. I found the program swinging around in the `while` loop in `Get`. Looks like `_camera.NewImageReady` event was never triggered.

Which step am I missing?

HumJ0218:
If I run MJPG-streamer directly, I can get the data stream normally.

HumJ0218:
If I run MJPG-streamer directly, I can get image data normally.

( It looks like I can start an MJPG separately and grab the image or stream to do what I want.

Ellerbach:
> Which step am I missing?

You may want to check first what your camera support. Maybe the basic JPEG mode may not work, maybe the resolution doesn't work. 
When running the basic sample, you should be able to see the list of compatible modes.

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -5,9 +5,11 @@
 using System;
 using System.Buffers.Binary;
 using System.IO;
+using System.Runtime.CompilerServices;
```
krwq, line 8: are you using the new namespaces?

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -147,6 +159,24 @@ public override void Record(uint recordTimeSeconds, Stream outputStream)
...
+            try
+            {
+                OpenRecordingPcm();
+                PcmInitialize(_recordingPcm, header, ref @params, ref dir);
+                ReadStream(outputStream, header, ref @params, ref dir);
+                CloseRecordingPcm();
+            }
+            finally
+            {
+                Dispose();
```
krwq, line 174: why do you dispose from Record method?

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -177,114 +207,156 @@ public override void Record(uint recordTimeSeconds, Stream outputStream)
...
-            try
+        public override void StartRecording(string outputFilePath)
+        {
+            Record(outputFilePath);
+        }
+
+        public override void StopRecording()
+        {
+            _record = false;
+            while (_recordingThread.IsAlive)
```
krwq, line 221: should you just `_recordingThread.Join()`

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -177,114 +207,156 @@ public override void Record(uint recordTimeSeconds, Stream outputStream)
...
 
-            try
+        public override void StartRecording(string outputFilePath)
+        {
+            Record(outputFilePath);
+        }
+
+        public override void StopRecording()
+        {
+            _record = false;
```
krwq, line 220: this would probably be better with Task and CancellationToken

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
```diff
@@ -177,114 +207,156 @@ public override void Record(uint recordTimeSeconds, Stream outputStream)
...
-        private void WriteWavHeader(Stream wavStream, WavHeader header)
+        public override void WriteWavHeader(Stream wavStream, WavHeader header)
         {
             Span<byte> writeBuffer2 = stackalloc byte[2];
             Span<byte> writeBuffer4 = stackalloc byte[4];
 
-            Encoding.ASCII.GetBytes(header.Chunk.ChunkId, writeBuffer4);
-            wavStream.Write(writeBuffer4);
+            writeBuffer4 = Encoding.ASCII.GetBytes(header.Chunk.ChunkId);
+            wavStream.Write(writeBuffer4.ToArray(), 0, writeBuffer4.Length);
```
krwq, line 273: this is allocating quite a bit... You probably don't benefit much fron using Span here. Actually might be harmful since you need to convert Array -> Span -> Array. In .net5 you got Span overloads for Write

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
Ellerbach: yes, that should work as well.

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
Ellerbach: It was the original logic and didn't changed it. It's not really necessary. So will remove.

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
Ellerbach: it's not needed, will remove. I did few tests and should be removed.

src/devices/Media/SoundDevice/Devices/UnixSoundDevice.cs
Ellerbach: What would be the benefit here? All sound is anyway single thread, you can't open 2 streams with Alsa.

krwq (approved):

PR #1164, opened 2020-08-25
Fixing DHT issues, adding and adjusting documentation

### LINKED ISSUE #1112: dht -11 can't work
**Describe the bug**

OS Name:     raspbian
 OS Version:  10
 OS Platform: Linux
 RID:         linux-arm
 Base Path:   /opt/dotnet/sdk/3.1.300/

i use 1-Wire Protocol

https://github.com/dotnet/iot/blob/master/src/devices/Dhtxx/README.md

it show 
NaN ,
NaN

### LINKED ISSUE #1153: Dht22 failed to read. I can read it using Python to make sure that the connection is OK.
Dht22 failed to read. I can read it using Python to make sure that the connection is OK.

 
using (Iot.Device.DHTxx.Dht22 dht = new Iot.Device.DHTxx.Dht22(26, PinNumberingScheme.Logical))
                {
                    while (true)
                    {
                        Console.WriteLine($"Temperature: {dht.Temperature.Celsius.ToString("0.0")} ℃, Humidity: { dht.Humidity} %");
                        Thread.Sleep(1000);
                    }
                }

### LINKED ISSUE #1161: Wrong temperature calculation for DHT21 and derived devices
The calculation of the temperature seems to be wrong. It is encoded in two bytes, according to the datasheet.
The current calculation is:
https://github.com/dotnet/iot/blob/21cf491456542716ff69192cb9f50da7e0dc2f58/src/devices/Dhtxx/Devices/Dht21.cs#L34

I think it should be corrected like this: `(((readBuff[2] & 0x7F) << 8) | readBuff[3]) * 0.1;`

Ellerbach:
Fixing DHT sensor issues:
- Fixes #1153, fixes #1161, should fix #1112 
- Making consistent calculation for temperatures for all DHT sensors
- Fixing DHT21 temperature calculation
Adjusting the documentation:
- Adding more sample for GPIO 1 wire usage
- Adding documentation for propre usage of DHT sensor
- Adding FAQ
Adding more sample:
- Sample implementing both I2C and 1 wire
- Ability to choose the sensor and the pin

### DIFF
--- a/src/devices/Dhtxx/DhtBase.cs
+++ b/src/devices/Dhtxx/DhtBase.cs
@@ -283,7 +283,7 @@ namespace Iot.Device.DHTxx
             }
             else
             {
-                _controller.ClosePin(_pin);
+                _controller?.ClosePin(_pin);
             }
 
             _i2cDevice?.Dispose();
--- a/src/devices/Dhtxx/DhtBase.cs
+++ b/src/devices/Dhtxx/DhtBase.cs
@@ -58,7 +58,7 @@ namespace Iot.Device.DHTxx
         /// Get the last read temperature
         /// </summary>
         /// <remarks>
-        /// If last read was not successfull, it returns <code>default(Temperature)</code>
+        /// If last read was not successful, it returns <code>default(Temperature)</code>
         /// </remarks>
         public virtual Temperature Temperature
         {
@@ -73,7 +73,7 @@ namespace Iot.Device.DHTxx
         /// Get the last read of relative humidity in percentage
         /// </summary>
         /// <remarks>
-        /// If last read was not successfull, it returns <code>default(Ratio)</code>
+        /// If last read was not successful, it returns <code>default(Ratio)</code>
         /// </remarks>
         public virtual Ratio Humidity
         {
@@ -280,6 +280,10 @@ namespace Iot.Device.DHTxx
             if (_shouldDispose)
             {
                 _controller?.Dispose();
+            }
+            else
+            {
+                _controller.ClosePin(_pin);
             }
 
             _i2cDevice?.Dispose();
--- a/src/devices/Dhtxx/DhtBase.cs
+++ b/src/devices/Dhtxx/DhtBase.cs
@@ -30,14 +30,14 @@ namespace Iot.Device.DHTxx
         protected readonly int _pin;
 
         /// <summary>
-        /// I2C device used to communicate with the device
-        /// </summary>
-        protected readonly I2cDevice _i2cDevice;
-
-        /// <summary>
         /// True to dispose the Gpio Controller
         /// </summary>
         protected readonly bool _shouldDispose;
+
+        /// <summary>
+        /// I2C device used to communicate with the device
+        /// </summary>
+        protected I2cDevice _i2cDevice;
 
         /// <summary>
         /// <see cref="GpioController"/> related with the <see cref="_pin"/>.
@@ -284,10 +284,17 @@ namespace Iot.Device.DHTxx
             }
             else
             {
-                _controller?.ClosePin(_pin);
+                if (_controller != null)
+                {
+                    if (_controller.IsPinOpen(_pin))
+                    {
+                        _controller.ClosePin(_pin);
+                    }
+                }
             }
 
             _i2cDevice?.Dispose();
+            _i2cDevice = null;
         }
     }
 }
--- a/src/devices/Dhtxx/DhtBase.cs
+++ b/src/devices/Dhtxx/DhtBase.cs
@@ -35,14 +35,14 @@ namespace Iot.Device.DHTxx
         protected readonly I2cDevice _i2cDevice;
 
         /// <summary>
-        /// <see cref="GpioController"/> related with the <see cref="_pin"/>.
-        /// </summary>
-        protected readonly GpioController _controller;
-
-        /// <summary>
         /// True to dispose the Gpio Controller
         /// </summary>
         protected readonly bool _shouldDispose;
+
+        /// <summary>
+        /// <see cref="GpioController"/> related with the <see cref="_pin"/>.
+        /// </summary>
+        protected GpioController _controller;
 
         // wait about 1 ms
         private readonly uint _loopCount = 10000;
@@ -280,6 +280,7 @@ namespace Iot.Device.DHTxx
             if (_shouldDispose)
             {
                 _controller?.Dispose();
+                _controller = null;
             }
             else
             {
--- a/src/devices/Dhtxx/README.md
+++ b/src/devices/Dhtxx/README.md
@@ -67,7 +67,7 @@ using (Dht12 dht = new Dht12(device))
             $"Heat index: {WeatherHelper.CalculateHeatIndex(tempValue, humValue).Celsius:0.#}\u00B0C");
         Console.WriteLine(
             $"Dew point: {WeatherHelper.CalculateDewPoint(tempValue, humValue).Celsius:0.#}\u00B0C");
-}   }
+    }
     else
     {
         Console.WriteLine("Error reading DHT sensor");
--- a/src/devices/Dhtxx/README.md
+++ b/src/devices/Dhtxx/README.md
@@ -41,7 +41,7 @@ using (Dht11 dht = new Dht11(26))
     }
 }
 ```
-**Note:** _Specifically on the RPi with the DHT22, 1-Wire works using Raspian but not with Windows 10 IoT Core. The device has to switch the 1-wire pin between input and output and vice versa. It seems that Windows IoT Core OS can't switch the pin direction quick enough. There have been suggestions for using two pins; one for input and one for output. This solution has not been implemented here, but these are some handy links that may help setting that up:_
+**Note:** _On the RPi with any of the DHT sensor, 1-Wire works using Raspian but not with Windows 10 IoT Core. The device has to switch the 1-wire pin between input and output and vice versa. It seems that Windows IoT Core OS can't switch the pin direction quick enough. There have been suggestions for using two pins; one for input and one for output. This solution has not been implemented here, but these are some handy links that may help setting that up:_
 - https://github.com/ms-iot/samples/tree/develop/GpioOneWire
 - And on Hackster.io: https://www.hackster.io/porrey/go-native-c-with-the-dht22-a8e8eb
 
@@ -124,6 +124,14 @@ var controller = new GpioController(PinNumberingScheme.Logical, driver);
 // This uses pin 4 in the logical schema so pin 7 in the physical schema
 var dht = new Dht11(4, gpioController: controller);
 ```
+
+**My DHT sensor using 1 wire protocol is not working on my Raspberry Pi with Windows 10 IoT Core, what can I do?**
+
+On the RPi with any of the DHT sensor, 1-Wire works using Raspian but not with Windows 10 IoT Core. The device has to switch the 1-wire pin between input and output and vice versa. It seems that Windows IoT Core OS can't switch the pin direction quick enough. There have been suggestions for using two pins; one for input and one for output. This solution has not been implemented here, but these are some handy links that may help setting that up:_
+- https://github.com/ms-iot/samples/tree/develop/GpioOneWire
+- And on Hackster.io: https://www.hackster.io/porrey/go-native-c-with-the-dht22-a8e8eb
+
+Now if your sensor is an I2C sensor, it should just work perfectly on Windows 10 IoT Core.
 
 ## References
 
--- a/src/devices/Dhtxx/README.md
+++ b/src/devices/Dhtxx/README.md
@@ -7,8 +7,8 @@ The DHT temperature and humidity sensors are very popular. This projects support
 |  | DHT10 | DHT11 | DHT12 | DHT21 | DHT22 |
 | :------: | :------: | :------: | :------: | :------: | :------: |
 | Image | <img src="imgs/dht10.jpg" height="60"/> | <img src="imgs/dht11.jpg" height="60"/> | <img src="imgs/dht12.jpg" height="60"/> | <img src="imgs/dht21.jpg" height="60"/> | <img src="imgs/dht22.jpg" height="60"/> |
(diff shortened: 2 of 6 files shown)


src/devices/Dhtxx/README.md
```diff
@@ -21,13 +21,29 @@ The DHT temperature and humidity sensors are very popular. This projects support
...
+            $"Dew point: {WeatherHelper.CalculateDewPoint(temperature, humidity).Celsius:0.#}\u00B0C");
+    }
+    else
+    {
+        Console.WriteLine("Error reading DHT sensor");
+    }
 }
 ```
-**Note:** _Specifically on the RPi with the DHT22, 1-Wire works using Raspian but not with IoT-Core. The device has to switch the 1-wire pin between input and output and vice versa. It seems that Windows IoT-Core OS can't switch the pin direction quick enough. There have been suggestions for using two pins; one for input and one for output. This solution has not been implemented here, but these are some handy links that may help setting that up:_
+**Note:** _Specifically on the RPi with the DHT22, 1-Wire works using Raspian but not with Windows 10 IoT Core. The device has to switch the 1-wire pin between input and output and vice versa. It seems that Windows IoT Core OS can't switch the pin direction quick enough. There have been suggestions for using two pins; one for input and one for output. This solution has not been implemented here, but these are some handy links that may help setting that up:_
```
pgrawehr, line 44: I think this applies for the DHT11, too. This also uses one wire for communication and switches between input and output.

src/devices/Dhtxx/samples/DhtSensor.sample.cs
```diff
@@ -15,26 +15,125 @@ internal class Program
         public static void Main(string[] args)
         {
             Console.WriteLine("Hello DHT!");
+            Console.WriteLine("Select the DHT sensor you want to use:");
+            Console.WriteLine(" 1. DHT10 on I2C");
+            Console.WriteLine(" 2. DHT11 on GPIO");
```
pgrawehr, line 20: :+1, Good example code

src/devices/Dhtxx/samples/DhtSensor.sample.cs
```diff
@@ -15,26 +15,125 @@ internal class Program
...
+            Console.WriteLine(" 3. DHT12 on GPIO");
+            Console.WriteLine(" 4. DHT21 on GPIO");
+            Console.WriteLine(" 5. DHT22 on GPIO");
+            var choice = Console.ReadKey();
+            Console.WriteLine();
+            if (choice.KeyChar == '1')
+            {
+                Console.WriteLine("Press any key to stop the reading");
+                Dht10II2c();
```
pgrawehr, line 29: Would be nice if this would be using the same code as below (except for the initialisation of course).

Ellerbach: Fully agree with you and at the same time, I was debating with myself. And I told myself that I would live it like this so it makes it easy for copy/paste. But I4m quite open. Happy to use the exact same code. Let's see other opinions.

src/devices/Dhtxx/README.md
Ellerbach: Yes, correct. It actually applies to all of them. I'll aslo then add to the Q&A

src/devices/Dhtxx/DhtBase.cs
```diff
@@ -281,6 +281,10 @@ public void Dispose()
             {
                 _controller?.Dispose();
             }
+            else
+            {
+                _controller.ClosePin(_pin);
```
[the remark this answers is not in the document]
Ellerbach: Yea, I have couple of others to do ;-) But nothing critical. So as soon as I have to update one, I then correct this :-)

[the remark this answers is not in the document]
pgrawehr: Crap, reviewing again suggests this needs to be fixed with a bit more logic. Since calling ClosePin with a pin that is not open throws an exception, this should likely reset the _controller to null, such that a repeated disposal does no harm. Repeatedly disposing the same instance shouldn't throw any exceptions.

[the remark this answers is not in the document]
Ellerbach: correct is controller is null. If it's not null, then if the pin is not open, an exception would have been raised much before. so fix is:
```csharp
_controller?.ClosePin(_pin);
```
Good catch!

src/devices/Dhtxx/samples/DhtSensor.sample.cs
Ellerbach: Debate with myself solved over night. Will just reuse. It's simple enough.

src/devices/Dhtxx/README.md
```diff
@@ -41,18 +57,82 @@ using (Dht12 dht = new Dht12(device))
...
+    {
+        Console.WriteLine($"Temperature: {tempValue.Celsius:0.#}\u00B0C");
+        Console.WriteLine($"Relative humidity: {humValue:0.#}%");
+
+        // WeatherHelper supports more calculations, such as saturated vapor pressure, actual vapor pressure and absolute humidity.
+        Console.WriteLine(
+            $"Heat index: {WeatherHelper.CalculateHeatIndex(tempValue, humValue).Celsius:0.#}\u00B0C");
+        Console.WriteLine(
+            $"Dew point: {WeatherHelper.CalculateDewPoint(tempValue, humValue).Celsius:0.#}\u00B0C");
+}   }
```
lindexi, line 70:
```suggestion
    }
```

Ellerbach: Good catch

src/devices/Dhtxx/DhtBase.cs
```diff
@@ -281,6 +281,10 @@ public void Dispose()
             {
                 _controller?.Dispose();
             }
+            else
+            {
+                _controller?.ClosePin(_pin);
+            }
 
             _i2cDevice?.Dispose();
```
pgrawehr, line 289:
```suggestion
            _i2cDevice?.Dispose();
            _controller = null;
```
+ Need to make _controller non-readonly.

Ellerbach: if it's readonly, what is the advantage of making it non readonly and setting it up to null? Also if setup to null, it will have to go to line 287, not here.

pgrawehr: Because if it is not reset to null, the change to `_controller?.ClosePin(_pin)` doesn't change anything. We need to make sure that ClosePin is not called a second time, and this requires the _controller to be null the second time. Since we Dispose the controller in the first case of the `if`, we can set it to null in both cases. (Additional advantage: Using the class after dispose will throw exceptions).

Ellerbach: But in this case, we want to close the pin and **not** dispose the controller

pgrawehr: True, I didn't say it should be disposed in the second case. It should be set to null, but in both cases.

src/devices/Dhtxx/DhtBase.cs
```diff
@@ -280,6 +280,7 @@ public void Dispose()
             if (_shouldDispose)
             {
                 _controller?.Dispose();
+                _controller = null;
```
pgrawehr, line 283: Sorry to insist: this needs to be outside the if (or in both cases). Otherwise, if `shouldDispose` is false, this still bombs on the second `Dispose` call.

Ellerbach: If ```_shouldDispose``` is false, the key principal is not to dispose. As an example, with the FT4222, you can only create 1 GPIO controller. If you have 2 devices which are using GPIO, and the first one dispose it, then the second one will raise an exception.
So it's important not to dispose at all the GPIO controller if the ```shouldDispose```` flag is false.

pgrawehr: True. I'm **not** talking about disposing, I'm talking about setting the controller to null. This doesn't do an implicit dispose or something.

Ellerbach: @pgrawehr I think I really miss something then. Can you please then write the full dispose class how you think it should be for me to understand what you mean.
Again, here is the principle:
- Whatever happens the I2C device is disposed
- the GPIO Controller is disposed only if _shouldDispose is true
- if _shouldDispose is flase, the open pins are closed and the controller is not disposed at all
To prevent some mistakes:
- _shouldDipose is set to true if the GPIO Controller is created in the class. So even if you pass false, if the controller = null in the class constructor, a controller will be created and _shouldDiposed will be set to true

pgrawehr: I'll do that. It really looks like we're talking different things. But give me some time,  as you probably know,  I'm traveling around,  and good WiFi is hard to find.

pgrawehr: Here's a patch file for your branch. I've also fixed a few minor other issues I just saw. In addition to your list above (which are all sattisfied), this
- does nothing if Dispose is called multiple times
- if an instance is used after Dispose, ObjectDisposedException is thrown

Ellerbach: Thanks, I understand now. So then, rather than raising an exception, why not then use the other pattern, for example present in some of the classes like the Explorer Hat: https://github.com/dotnet/iot/blob/3c4034781d158b26be0bc4866c34192d86bc4ef1/src/devices/ExplorerHat/Motors.cs#L87
So having a variable to avoid double disposing? Advantage: it doesn't raise an exception. I'm not much in favor of raising exceptions in the Dispose function. Those are exceptions usually no one really trap.

pgrawehr: The exception is not thrown in Dispose (thats what we try to avoid), but if the instance is otherwise used after Disposal, i.e when trying to read from the sensor after disposing. 
An ObjectDisposedException is rightly (almost) never handled in user code, since that is a programming error,  and they should stop the application typically in the debugger. 

Of course the same as what I did can be achieved by an extra bool variable,  but that doesn't change much conceptionally.

Ellerbach: OK, so I adjusted only the dispose part to make sure that is multiple dispose are called, it won't create any issue.
In case of a programming error and the class is called once disposed, an exception will be raised anyway. 
So it keeps the solution simple enough and easy to read. I hope it's a good compromise.

pgrawehr: I would still set it to null, but I'm ok with this as well.

pgrawehr (approved):

reddevilfirst:
May be it help.

Added at current DhtBase.cs.
   
```
/// <summary>
/// Read data from the bus
/// </summary>
/// <returns>is reading successful</returns>
public virtual bool TryGetData()
{
   ReadData();
   return IsLastReadSuccessful;
}
```


Test code:

```
        {
            Thread d22 = new Thread(new ThreadStart(dht22));
            d22.Start();
            Thread d21 = new Thread(new ThreadStart(dht21));
            d21.Start();
            Thread d11 = new Thread(new ThreadStart(dht11));
            d11.Start();

            lock (Thread.CurrentThread)
            {
                Monitor.Wait(Thread.CurrentThread);
            }
 }
```

```
private static void dht22()
        {
            using (Dhtxx.Devices.Dht22 dht = new Dht22(23, PinNumberingScheme.Logical))
            {
                while (true)
                {
                    if (dht.TryGetData())
                    {
                        Console.WriteLine($"Dht22. Temperature: {dht.Temperature.DegreesCelsius:0.0} ℃, Humidity: { dht.Humidity}");
                    }
                    Thread.Sleep(1000);
                }
            }
        }

        private static void dht21()
        {
            using (Dhtxx.Devices.Dht21 dht = new Dht21(18, PinNumberingScheme.Logical))
            {
                while (true)
                {
                    if (dht.TryGetData())
                    {
                        Console.WriteLine($"Dht21. Temperature: {dht.Temperature.DegreesCelsius:0.0} ℃, Humidity: { dht.Humidity}");
                    }
                    Thread.Sleep(1000);
                }
            }
        }

        private static void dht11()
        {
            using (Dhtxx.Devices.Dht11 dht = new Dht11(27, PinNumberingScheme.Logical))
            {
                while (true)
                {
                    if (dht.TryGetData())
                    {
                        Console.WriteLine($"Dht11. Temperature: {dht.Temperature.DegreesCelsius:0.0} ℃, Humidity: { dht.Humidity}");
                    }
                    Thread.Sleep(1000);
                }
            }
        }
```


result:

Dht11. Temperature: 18.7 ℃, Humidity: 51 %
Dht11. Temperature: 18.7 ℃, Humidity: 51 %
Dht21. Temperature: 20.1 ℃, Humidity: 53 %
Dht11. Temperature: 18.8 ℃, Humidity: 51 %
Dht21. Temperature: 20.3 ℃, Humidity: 53.1 %
Dht22. Temperature: 19.4 ℃, Humidity: 53.3 %
Dht11. Temperature: 18.8 ℃, Humidity: 51 %
Dht22. Temperature: 19.4 ℃, Humidity: 52.5 %
Dht21. Temperature: 20.0 ℃, Humidity: 53 %
Dht22. Temperature: 19.4 ℃, Humidity: 52.9 %
Dht11. Temperature: 18.9 ℃, Humidity: 51 %
Dht22. Temperature: 19.4 ℃, Humidity: 52.7 %
Dht22. Temperature: 19.4 ℃, Humidity: 52.6 %
Dht11. Temperature: 18.8 ℃, Humidity: 51 %
Dht21. Temperature: 20.0 ℃, Humidity: 53 %
Dht22. Temperature: 19.4 ℃, Humidity: 53.6 %
Dht21. Temperature: 20.0 ℃, Humidity: 53 %
Dht11. Temperature: 18.9 ℃, Humidity: 51 %
Dht21. Temperature: 20.0 ℃, Humidity: 53 %
Dht21. Temperature: 19.6 ℃, Humidity: 52.8 %
Dht22. Temperature: 19.4 ℃, Humidity: 52.6 %
Dht21. Temperature: 19.9 ℃, Humidity: 53 %
Dht22. Temperature: 19.4 ℃, Humidity: 52.4 %
Dht21. Temperature: 20.0 ℃, Humidity: 53 %
Dht21. Temperature: 20.0 ℃, Humidity: 53 %
Dht22. Temperature: 19.4 ℃, Humidity: 53.5 %

Data read without exceptions. .net core 3.1 raspberry pi 3 b+

Issue #637, opened 2019-07-29
System.Device.Gpio need Abstractions\HAL (Hardware Abstraction Layer)


Ordjony:
Propose to split **System.Device.Gpio** into different libs like **System.Device.Gpio.Abstractions** (e.g. asp.net core) without dependencies and **System.Device.Gpio.*** (OS or device-specific).

At the moment System.Device.Gpio has dependencies on Microsoft.Win32.Registry, System.Runtime.InteropServices.WindowsRuntime, System.Runtime.WindowsRuntime which looks like not necessary on Linux. Iot.Device.Bindings should use Abstractions library.

This Abstractions package gives the ability to use binding in different platforms (e.g. https://www.wildernesslabs.co/meadow).

joperezr:
Hey @Ordjony thanks a lot for logging the issue. Is the only concern that there are extra Windows-specific dependencies on the library package? If so, I don't think that this would warrant the fact of creating a new abstraction layer. All those three dependecies you found are only really used by the windows implementation, none of those dependecies are exposed in the API shape, and of course they aren't used at all in the Linux implementation. Also, those dependencies have packages that are supported in netstandard, so there is really no concern that the packages won't restore correctly on other non-Windows platforms. There is no good way in NuGet to add dependencies to a package that are specific only to one RID, which is why we have to add it to all of them but that doesn't mean that we need those dependencies everywhere. This should in no way limit where would IoT.Device.Bindings could run, as System.Device.Gpio library code should run anywhere where netstandard 2.0 is supported(provided that we are running in a board that does in fact expose GPIO pins)

Ordjony:
Hi, @joperezr thanks for the very fast answer. Now I looked at Nuget package inside and understood that it's no problem to target any platform which supports netstandard 2.0. This dependency was confused.

joperezr:
Thanks again for your input and suggestion. If there is a specific binding you were interested in let us know so that we can create an issue for it and start taking for PRs.

Ordjony:
I decided to reopen my issue and continue the discussion. I researched a little bit ARM MBED OS and watched to Rust lang HAL.
I think in .net we need our own HAL library which will contain only interfaces for GPIO without any methods to get this GPIO or settings.
```csharp
public interface II2CBus
{
    void Read(Span<byte> buffer);
    byte ReadByte();
    void Write(ReadOnlySpan<byte> buffer);
    void WriteByte(byte value);
[...]

Issue #383, opened 2019-04-23
Creating a platform neutral driver


raffaeler:
Not sure if this is already possible, but here is an idea.
Since the drivers already abstract different implementations of the local hardware peripherals such as Gpio, I2C and SPI, I find very valuable having an in-memory driver that lying below the existing drivers.

The goals for this proposal:
- ability to mock all the devices for testing purposes without having to use Moq or similar frameworks
- ability to log the sequence of I/O writes
- ability to create a framework to replay a sequence to simulate a complete I/O handshake without any connected device
- ability to run the application on a host which doesn't have any device connected. For example, this is useful to test the other parts of the application on Windows and later run the real one on the embedded board.

While it is possible to create a similar abstraction in the application, it would be less invasive at System.Devices.Gpio level.

Furthermore it would be useful to add this driver as an _additional_ layer so that it would be possible to log all the I/Os for later evaluation.

I would opt for a memory only driver in order to avoid any latency.

joperezr:
I'm not 100% sure that adding a new layer would be a good idea. Today the layering model goes something like: YourApp -> GpioController.DoSomething -> GpioDriver.DoSomething. Because of that, if we wanted to just add this extra layer for logging pourposes, I would suggest to add those at the GpioController level instead, and I would also be inclined to have them be optional since we don't want to affect perf by logging every I/O operation. For your other proposal (using it as a Moq), we could always add test drivers in this case, that would be possible by adding a testGpioDriver that would get called by the GpioController and can serve to fake I/O interactions. I also may have not understand your proposal correctly, so please correct me if that is the case.

raffaeler:
Thank you @joperezr 
I really didn't want to introduce a new layer for every operation. The proposal was to add an optional layer that you may want to plug-in only when desired.

One of the goals is interception that may need to either log or even patch a communication. There are times where the behavior is intermittently wrong. In these cases an intermediate driver can identify the problem better than the application.

The other goal is for testing purposes. All write-only devices can be tested this way and avoid developing a test driver. Reading-only devices would be very easy as well as you may load a sequence and then emulate the I/Os instead of the device. There are times where is not easy to reproduce the device behavior during the tests but you know from the datasheet how it can behave.

In any case I agree that this layer should be optional and plugged in only in specific case to avoid any overhead.

[...]

PR #1167, opened 2020-08-26
adding disposable, shared controller interface and gpiocontroller interface

colombod:
With this PR I am trying to propose a discussion around some interfaces used during the development of the piTop bindings

- introduction of a IGpioController interface to make it possible to have mocks and implement shared proxy
- share controller class to create a view over a controller so that devices and other part of applications can manage pins and callbacks with the ability to clean up just the portion of state interested
- disposable factory to create disposable object to simplify the closing of pins and the unregistration of callbacks

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/IGpioControllerExtensions.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/IGpioControllerExtensions.cs
@@ -1,0 +1,54 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Disposables;
+
+namespace System.Device.Gpio
+{
+    /// <summary>
+    /// Extensions for <see cref="IGpioController"/> interface.
+    /// </summary>
+    public static class IGpioControllerExtensions
+    {
+        /// <summary>
+        /// Opens a pin in order for it to be ready to use.
+        /// </summary>
+        /// <param name="controller">The GPioController.</param>
+        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        /// <returns>A disposable that will close the pin if disposed.</returns>
+        public static IDisposable OpenPinAsDisposable(this IGpioController controller, int pinNumber)
+        {
+            controller.OpenPin(pinNumber);
+            return Disposable.Create(() => controller.ClosePin(pinNumber));
+        }
+
+        /// <summary>
+        /// Opens a pin and sets it to a specific mode.
+        /// </summary>
+        /// <param name="controller">The GPioController.</param>
+        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        /// <param name="mode">The mode to be set.</param>
+        /// <returns>A disposable that will close the pin if disposed.</returns>
+        public static IDisposable OpenPinAsDisposable(this IGpioController controller, int pinNumber, PinMode mode)
+        {
+            controller.OpenPin(pinNumber, mode);
+            return Disposable.Create(() => controller.ClosePin(pinNumber));
+        }
+
+        /// <summary>
+        /// Adds a callback that will be invoked when pinNumber has an event of type eventType.
+        /// </summary>
+        /// <param name="controller">The GPioController.</param>
+        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        /// <param name="eventTypes">The event types to wait for.</param>
+        /// <param name="callback">The callback method that will be invoked.</param>
+        /// <returns>A disposable object that will remove the added callback </returns>
+        public static IDisposable RegisterCallbackForPinValueChangedEventAsDisposable(this IGpioController controller, int pinNumber,
+            PinEventTypes eventTypes, PinChangeEventHandler callback)
+        {
+            controller.RegisterCallbackForPinValueChangedEvent(pinNumber, eventTypes, callback);
+            return Disposable.Create(() => controller.UnregisterCallbackForPinValueChangedEvent(pinNumber, callback));
+        }
+    }
+}
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -92,6 +92,30 @@ namespace System.Device.Gpio.Tests
         }
 
         [Fact]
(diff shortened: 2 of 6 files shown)


pgrawehr:
Unfortunatelly, there's a big issue: System.Device.Gpio interfaces must be kept backwards compatible. Breaking changes must have _very_ good reasoning. Note also that some interface concept discussions have been open for a long time already (but where postponed due to other tasks and long absences of the responsible maintainers). See #1128 and #878.

colombod:
> Unfortunatelly, there's a big issue: System.Device.Gpio interfaces must be kept backwards compatible. Breaking changes must have _very_ good reasoning. Note also that some interface concept discussions have been open for a long time already (but where postponed due to other tasks and long absences of the responsible maintainers). See #1128 and #878.

Makes sense, the contract change on the OpenPin and RegisterCallBack is something that could be left out, would it still make the case of having IGpioController interface a big issue with the api backward compatibility?

colombod:
> Unfortunatelly, there's a big issue: System.Device.Gpio interfaces must be kept backwards compatible. Breaking changes must have _very_ good reasoning. Note also that some interface concept discussions have been open for a long time already (but where postponed due to other tasks and long absences of the responsible maintainers). See #1128 and #878.

Thank you for you feedback, I realized I really just need the IGpioController interface so I can build proxy and adaptors, the other api changes can be done as extension methods so not to make the type too big or break existing contract. Thank you

pgrawehr:
> 
> Makes sense, the contract change on the OpenPin and RegisterCallBack is something that could be left out, would it still make the case of having IGpioController interface a big issue with the api backward compatibility?

I guess so. Interfaces don't scale well (because _every_ change to an interface, including adding methods, is a breaking change). What we're probably going to do (is under discussion now), is to unseal the GpioController class and change its methods to virtual. This (besides other reasons) also allow the mocking support you want.

colombod:
@pgrawehr the unselaed class would be fine. Any time frame idea for such a change? That would actually enable all api addition I am doing to be an external package that user can adopt if they are interested in some additional pattern and utilities.

pgrawehr:
@colombod : Vote for/have a look at #1128. That contains quite a few concept updates, and it would be great to have a broad discussion on them.

colombod:
> @colombod : Vote for/have a look at #1128. That contains quite a few concept updates, and it would be great to have a broad discussion on them.

that PR is what I call Christmas present. It is absolutely brilliant with the controller and spi management and all the rest.

krwq:
This seem to overlap with some of the other conversations we have in other issues:
https://github.com/dotnet/iot/issues/383
https://github.com/dotnet/iot/issues/80
https://github.com/dotnet/iot/issues/1182
https://github.com/dotnet/iot/issues/637
https://github.com/dotnet/iot/issues/878

I'm hoping we will figure out the abstraction subject as we're having request from different directions about it

src/System.Device.Gpio/System/Device/Gpio/IGpioControllerExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    public static class IGpioControllerExtensions
+    {
+        /// <summary>
+        /// Opens a pin in order for it to be ready to use.
+        /// </summary>
+        /// <param name="controller">The GPioController.</param>
+        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        /// <returns>A disposable that will close the pin if disposed.</returns>
+        public static IDisposable OpenPinAsDisposable(this IGpioController controller, int pinNumber)
```
krwq, line 20: we're actually thinking of having GpioPin class instead, see https://github.com/dotnet/iot/issues/215

src/System.Device.Gpio/System/Device/Gpio/IGpioControllerExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        /// <summary>
+        /// Adds a callback that will be invoked when pinNumber has an event of type eventType.
+        /// </summary>
+        /// <param name="controller">The GPioController.</param>
+        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        /// <param name="eventTypes">The event types to wait for.</param>
+        /// <param name="callback">The callback method that will be invoked.</param>
+        /// <returns>A disposable object that will remove the added callback </returns>
+        public static IDisposable RegisterCallbackForPinValueChangedEventAsDisposable(this IGpioController controller, int pinNumber,
```
krwq, line 47: I'd rather have us introduce the real abstraction for this (i.e. CallbackForPinValueChanged) implementing IDisposable

krwq:
I do like confirmation from here that adding GpioPin is the right direction and also looking at this PR seems like another person trying to bypass GpioDriver abstraction which makes me thinking to perhaps just deprecate it and make everything implement  GpioController directly (so unseal and virtualize it as proposed by @pgrawehr as well and just not have GpioDriver - and of course leave compat GpioController to make old code still work)

pgrawehr:
> 
> 
> ...and just not have GpioDriver - and of course leave compat GpioController to make old code still work)

I think that extra driver class is fine to keep, even if the Controller is unsealed - it separates responsibilities. The Controller handles pin management, while the driver handles the low-level operations on individual pins.

Ellerbach:
> I do like confirmation from here that adding GpioPin is the right direction

Yes, this really help to clearly see the need of abstraction, separating drivers from the controller itself and having GpioPin as well.

> The Controller handles pin management, while the driver handles the low-level operations on individual pins.

I agree, this would perfectly work. 

> This seem to overlap with some of the other conversations we have in other issues:
> #383
> #80
> #1182
> #637
> #878

And the good news is that with all this, I feel we're all ready to move forward. Once the new model will be present, the trendy topic shouldn't be GpioController anymore on the various issues :-)

joperezr:
[Triage] Hey @colombod we have been making good progress on this PR: https://github.com/dotnet/iot/pull/1254 which might help you out with your needs and might be related to this one. Can you check that one to see if that would solve your needs?

colombod:
> [Triage] Hey @colombod we have been making good progress on this PR: #1254 which might help you out with your needs and might be related to this one. Can you check that one to see if that would solve your needs?

This pr looks good and enables quite few things for me

joperezr:
[Triage] @colombod do you think there is still value on keeping this PR open or would #1254 cover your needs? If the latter then we can go ahead and close this PR.

colombod:
Will close this or as I have different things based on your nee library

Issue #1123, opened 2020-07-02
Hardware PWM issue


fabus1989:
Hi,

I have problem with hardware PWM, cannot run it from .net core application - LED connected to GPIO12.
PWM is enabled in /boot/config.txt file (dtoverlay=pwm-2chan)
Sample python script works ok.

What I'm missing?

Here is the code sample:
```
using System;
using System.Device.Pwm;

namespace Sample
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");
            var pwm = PwmChannel.Create(0, 0, 100, 1);
            pwm.Start();
            Console.ReadKey();
        }
    }
}
```

 I'm using Raspberry PI 4 with 2GB of ram.

Best regards
Fabian

joperezr:
Hello thanks for logging the issue. I believe that the Raspberry Pi doesn't support frequency as low as 100, can you instead try something like `10000` for frequency and perhaps a `.5` for dutyCycle? dutyCycle will work fine with a value of 1 but the idea would be to dim it so that you can notice the change of brightness.

pgrawehr:
I've also observed issues when the frequency is below 10000 (or so). I'm not sure what's going on then, but it starts blinking with the given duty cycle, but at frequencies of like 1 Hz. I don't know whether this is a hardware or a software issue, though.

krwq:
is it possible you got multiple app instances running on the machine? I've had this happen to me couple of times when working over ssh

fabus1989:
@joperezr I've tried with multiple frequecy values (100,400,1000,10000) and with few duty cycle (0.25, 0.5, 0.75, 1.0). Unfortunatelly LED light is always off. I do not have multiple instances running. I also tried with different channels and different pins. In config.txt pwm setting is at the bottom of a file. Do you have other ideas what to check?

fabus1989:
One strange thing I observed, usually first time (after raspberry reboot) when I run program I'm receiving exception.
``Unhandled exception. System.UnauthorizedAccessException: Access to the path '/sys/class/pwm/pwmchip0/pwm0/duty_cycle' is denied.
 ---> System.IO.IOException: Permission denied
   --- End of inner exception stack trace ---
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
[...]

PR #1168, opened 2020-08-26
Adding PWM documentation

Ellerbach:
Adding specific step by step documentation for hardware PWM activation on Raspberry Pi
- Fixing issue #1123 and related to #1128 
- Detailed step by step for hardware PWM activation
- Permission step by step to run an application in the current user

### DIFF
--- a/Documentation/raspi-pwm.md
+++ b/Documentation/raspi-pwm.md
@@ -52,8 +52,7 @@ Unhandled exception. System.UnauthorizedAccessException: Access to the path '/sy
 
 In order to have the hardware PWM activated on the Raspberry Pi, you'll have to edit the /boot/config.txt file and add an overlay.
 
-Main https://github.com/raspberrypi/linux/blob/04c8e47067d4873c584395e5cb260b4f170a99ea/arch/arm/boot/dts/overlays/README#L925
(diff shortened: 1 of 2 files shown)


Documentation/raspi-pwm.md
```diff
@@ -0,0 +1,180 @@
...
+
+This is straight forward, you have to use ```sudo``` to run your application. Let's say your application is names ```yourapplication```, once in the same directory as your application, it will then be:
+
+```bash
+sudo ./yourapplication
+```
+
+### Adding your user to the right permission group
+
+You will have to create a [specific group in udev](https://raspberrypi.stackexchange.com/questions/66890/accessing-pwm-module-without-root-permissions).
```
pgrawehr, line 157: This should no longer be necessary. After the latest kernel updates, the lines below are by default part of 99-com.rules. I had asked for this some time ago, and meanwhile it was actually done! See https://github.com/RPi-Distro/raspberrypi-sys-mods/commit/eef0792e899ede22f641e017bae4bdda7e12b9d1

Ellerbach: It's not because it's in the latest version that it should not be present in this documentation. You may not always use last version. This is something to check in all the cases.

pgrawehr: True, but maybe explicitly mention that updating the Pi should fix this as well?

Ellerbach: Fair! Will adjust. Thanks as always for the great insights.

PR #1175, opened 2020-09-06
Add support for AHTxx temperature and humidity sensor devices and MH-Z19B CO2 gas sensor

fehdem:
Adding support for AHT20 and AHT10 temperature and humidity devices ([http://www.aosong.com/en/productslist-4.html]).
The devices can be connected to standard I2C bus.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1175) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1175) before we can accept your contribution.<br/><br/>:x: fehdem [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1175)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1175) it.</sub>

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,159 @@
...
+            {
+                (byte)Command.Measure,
+                0x33, // command parameters c.f. datasheet, version 1.1, ch. 5.4
+                0x00
+            };
+            _i2cDevice.Write(buffer);
+
+            while (IsBusy())
+            {
+                Thread.Sleep(10);
```
pgrawehr, line 95: Are you sure you need so long loop times? What is the time it usually takes for this command to execute? (Probably specified in the sheet)

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,159 @@
...
+            }
+
+            buffer = stackalloc byte[6];
+            _i2cDevice.Read(buffer);
+
+            Int32 rawHumidity = (buffer[1] << 12) | (buffer[2] << 4) | (buffer[3] >> 4);
+            _humidity = (rawHumidity * 100.0) / 0x100000;
```
pgrawehr, line 102: Looks a bit weird, converting to double and then do a hex division. This looks like you would do an integer division, only "by accident" the division is actually using double. Maybe add comment?

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,159 @@
...
+            _temperature = ((rawTemperature * 200.0) / 0x100000) - 50;
+        }
+
+        /// <summary>
+        /// Perform soft reset command sequence
+        /// </summary>
+        private void SoftReset()
+        {
+            _i2cDevice.WriteByte((byte)Command.SoftRest);
+            Thread.Sleep(20); // reset requires 20ms at most, c.f. datasheet version 1.1, ch. 5.5
```
pgrawehr, line 113: Please cross check that the datasheet really says 20 Milliseconds. Most sensors have wait times in the order of nanoseconds only.

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,159 @@
...
+            return (GetStatusByte() & (byte)StatusBit.Busy) == (byte)StatusBit.Busy;
+        }
+
+        private bool IsCalibrated()
+        {
+            return (GetStatusByte() & (byte)StatusBit.Calibrated) == (byte)StatusBit.Calibrated;
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
```
pgrawehr, line 150: Since this class is not sealed, it should use the correct dispose pattern (with an overloadable `Dispose(bool disposing)` member)

pgrawehr (changes requested):
Only minor comments from my end, otherwise LGTM.

src/devices/Ahtxx/Aht20.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Device.I2c;
+
+namespace Iot.Device.Ahtxx
+{
+    /// <summary>
+    /// Add documentation here
```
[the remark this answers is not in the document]
fehdem: Oh yes, I'll do so. Sorry.

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        /// <summary>
+        /// Gets the current temperature reading from the sensor.
+        /// </summary>
+        /// <returns>Temperature reading</returns>
+        public Temperature GetTemperature()
+        {
+            Measure();
+            return new Temperature(_temperature, UnitsNet.Units.TemperatureUnit.DegreeCelsius);
```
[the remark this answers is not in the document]
fehdem: Yes, that's a reasonable proposal.

src/devices/Ahtxx/AhtBase.cs
fehdem: Indeed I took the max. exec. times given in the datasheet. Probably I can shorten the cycle by querying I bit more often. I had this thought already, on the other hand speed wasn't my primary goal and at the same time I always try to avoid too much "polling" traffic on the bus. However, I think a 2ms interval wouldn't do any harm, but may speed up the whole reading process.

src/devices/Ahtxx/AhtBase.cs
fehdem: Yes, it does, at least as a max. exec. time. However, let me try to figure out if this could be sped up using the busy flag. I'll built a prototype this weekend and check.

src/devices/Ahtxx/AhtBase.cs
fehdem: I'll add documentation. This is indeed not self explaining.

src/devices/Ahtxx/AhtBase.cs
fehdem: Ok. I'll change this.

src/devices/Ahtxx/AhtBase.cs
fehdem: I double checked the datasheet - it states: "Wait for 80ms to wait for the measurement to be completed. If the read status word Bit [7] is 0, it indicates that the measurement is completed, ...."
So looping in 10ms intervals should be good. 
However I did some testing and profiling. 

```
GetTemperature
     [1,2]: 0
     [2,5]: 0
    [5,10]: 0
   [10,20]: 983
   [20,50]: 14
  [50,100]: 2
 [100,200]: 0
```
It seems that the measurement takes much less than 80ms. So, I think the loop is ok.

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,218 @@
...
+            return SendRequest(request);
+        }
+
+        /// <summary>
+        /// Switches the autmatic baseline correction on and off.
+        /// The sensor doesn't respond anything, so this is fire and forget.
+        /// </summary>
+        /// <param name="state">State of automatic correction</param>
+        /// <returns>true, if the command could be send</returns>
+        public bool AutomaticBaselineCorrection(AbmState state)
```
Ellerbach, line 142: Would it be easier if you are just using a property? 
AbmState is only on and off. That will make it simpler
I understand SendRequest returns a potential status but you can as well raise an IOException as if you have a problem with the serial port, none of the commands will work anyway. 
If there are good reasons why not having a bool property (or if you want to keep the enum a AbmState property), then please use a verb for the function like SetAutomaticBaselineCorrection or SwitchAutomaticBaselineCorrection

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            request[(int)MessageFormat.Checksum] = Checksum(request);
+
+            try
+            {
+                _serialPort.Open();
+                _serialPort.Write(request, 0, request.Length);
+                validity = true;
+            }
+            finally
```
Ellerbach, line 181: Would be better to raise an IOException rather than returning an issue? If there is an issue with the serial port, none of the other commands will work anyway.

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,218 @@
...
+        private bool _disposed = false;
+        private SerialPort _serialPort = null;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Mhz19b"/> class.
+        /// </summary>
+        /// <param name="uartDevice">Path to the UART device / serial port, e.g. /dev/serial0</param>
+        public Mhz19b(string uartDevice)
+        {
+            if (uartDevice == null)
```
Ellerbach, line 50: Maybe check with String.IsNullOrEmpty?

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            return (concentration, validity);
+        }
+
+        /// <summary>
+        /// Initiates a zero point calibration.
+        /// The sensor doesn't respond anything, so this is fire and forget.
+        /// </summary>
+        /// <returns>true, if the command could be send</returns>
+        public bool ZeroPointCalibration() => SendRequest(CreateRequest(Command.CalibrateZeroPoint));
```
Ellerbach, line 118: Please use a verb for this action like InitiateZeroPointCalibration()

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,218 @@
...
+
+        /// <summary>
+        /// Initiate a span point calibration.
+        /// The sensor doesn't respond anything, so this is fire and forget.
+        /// </summary>
+        /// <param name="span">span value, e.g. 2000[ppm]</param>
+        /// <returns>true, if the command could be send</returns>///
+        public bool SpanPointCalibration(int span)
```
Ellerbach, line 126: Please use a verb for this action like InitiateSpanPointCalibration
Please also use the propre UnitsNet for the span. There is one for concentrations
Please add in the comments to minimum and maximum value and check it at the beginning of the function. From what I can see, it's from 0 to 65K

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,218 @@
...
+            return SendRequest(request);
+        }
+
+        /// <summary>
+        /// Set the sensor detection range.
+        /// The sensor doesn't respond anything, so this is fire and forget
+        /// </summary>
+        /// <param name="detectionRange">Detection range of the sensor</param>
+        /// <returns>true, if the command could be send</returns>
+        public bool SensorDetectionRange(DetectionRange detectionRange)
```
Ellerbach, line 157: You can either use a property here or add an action verb like SetSensorDetectionRange. 
Having a property would a bit better. You can as well return the current value.

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,218 @@
...
+            {
+                throw new ArgumentNullException(nameof(uartDevice));
+            }
+
+            // create serial port using the setting acc. to datasheet, pg. 7, sec. general settings
+            _serialPort = new SerialPort(uartDevice, 9600, Parity.None, 8, StopBits.One);
+            _serialPort.Encoding = Encoding.ASCII;
+            _serialPort.ReadTimeout = 1000;
+            _serialPort.WriteTimeout = 1000;
+        }
```
Ellerbach, line 60: Do you have any way to check that the sensor is attached? A serial number to read or equivalent?

src/devices/Mhz19b/samples/Mhz19b.Sample.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// Samples for Mhz19b
+    /// </summary>
+    public class Program
+    {
+        /// <summary>
+        /// Main entry point
+        /// </summary>
+        public static void Main(string[] args)
+        {
+            Mhz19b sensor = new Mhz19b("/dev/serial0");
```
Ellerbach, line 21: Maybe add something for the calibration to show how to use it?

src/devices/Mhz19b/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+|Function|	Raspi pin|	MH-Z19 pin|
+|--------|-----------|------------|
+|Vcc +5V |2 (+5V)	 |6 (Vin)     |
+|GND	 |6 (GND)	 |7 (GND)     |
+|UART	 |8 (TXD0)	 |2 (RXD)     |
+|UART	 |10 (RXD0)	 |3 (TXD)     |
+Table: MH-Z19B to RPi 3 connection
+ 
+**Make sure that you read the datasheet carefully before altering the default calibration behaviour.
```
Ellerbach, line 19: Maybe add an example and a bit of details here?

src/devices/Mhz19b/README.md
```diff
@@ -0,0 +1,23 @@
+﻿# MH-Z19B CO2-Sensor
```
Ellerbach, line 1: Please also add a 'category.txt' file at the root of the project. see other sensors for the example. It will place the sensor in the right category .And I guess 'Gas sensors'

Ellerbach (commented):
Thanks for adding this sensor, couple of comments in this review

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the device connected through I2C interface.
+        /// </summary>
+        /// <paramref name="i2cDevice">Reference to the initialized I2C interface device</paramref>
+        public AhtBase(I2cDevice i2cDevice)
+        {
+            _i2cDevice = i2cDevice ?? throw new ArgumentNullException(nameof(i2cDevice));
+
+            SoftReset(); // even if not clearly stated in datasheet, start with a software reset to assure clear start conditions
```
Ellerbach, line 51: comment up the line would be better, not after the instructions. There are couple of places like this in the files

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,183 @@
...
+            return (GetStatusByte() & (byte)StatusBit.Calibrated) == (byte)StatusBit.Calibrated;
+        }
+
+        /// <inheritdoc cref="IDisposable" />
+        public void Dispose() => Dispose(true);
+
+        /// <inheritdoc cref="IDisposable" />
+        protected virtual void Dispose(bool disposing)
+        {
+            if (_disposed)
```
Ellerbach, line 166: Can you just check if i2cDevice is already null?

src/devices/Ahtxx/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Device Family
+The binding supports the following types:
+* AHT10 - http://www.aosong.com/en/products-40.html (not tested, yet)
+* AHT20 - http://www.aosong.com/en/products-32.html
+
+## Binding Notes
+
+The AHT-devices can be accessed as an I2C bus device. However, you can use only one device per bus as the address is fix.
+
+
```
Ellerbach, line 15: Maybe add a very simple few lines example on how to use the device? So makes it easy for anyone to use rather than going to the sample?

Ellerbach: And please don't forget as well the category.txt file!

src/devices/Mhz19b/Mhz19b.cs
fehdem: I wouldn't prefer a property over a function here, as the call requires a certain execution time. From my understanding a property access should be rather instantly be finished. In case everything is fine this would be (almost) the case, but if there is an issue with the serial port it takes until the timeout happens. Even if this would be reduced to, let's say 10 or 100ms, this isn't the timing I expect from a property. Furthermore there is no chance in reading back the current actual setting, therefore the property wouldn't be readable in the first place. 

Whether I use an exception or a return code for reporting the status is a matter of taste or project coding guideline.  
Usually I prefer the return code, but if preferred by the iot-project I could change this. Please advice.

src/devices/Mhz19b/Mhz19b.cs
fehdem: Whether I use an exception or a return code for reporting the status is a matter of taste or project coding guideline.
Usually I prefer the return code, but if preferred by the iot-project I could change this. Please advice.

src/devices/Mhz19b/Mhz19b.cs
fehdem: I'll change this.

src/devices/Mhz19b/Mhz19b.cs
fehdem: I'll change this.

src/devices/Mhz19b/Mhz19b.cs
fehdem: I'll change this. A range check is reasonable as I don't expect that the sensor specification will change in later revision which could lead to inconsistency w/ the implementation.

src/devices/Mhz19b/Mhz19b.cs
fehdem: Same comment as for AutomaticBaselineCorrection command. I would prefer a method over a property.

src/devices/Mhz19b/samples/Mhz19b.Sample.cs
fehdem: I could add this as code commented-out. Altering any of the calibration settings could make the device unusable for a user who can't perform a real calibration. I don't want to let someone run accidentally into this.

src/devices/Mhz19b/README.md
fehdem: I'll add.

src/devices/Ahtxx/AhtBase.cs
fehdem: Ok.

src/devices/Ahtxx/AhtBase.cs
fehdem: Yes, I think so.

src/devices/Ahtxx/README.md
fehdem: Yep. ;)

src/devices/Mhz19b/Mhz19b.cs
fehdem: No, unfortunately there is nothing that can be used to check availability and communication.
If the sensor is not recognizing the request it doesn't respond.

src/devices/Mhz19b/Mhz19b.cs
Ellerbach: In that case, as it's related to the SerialPort itsef, I would prefer an Exception. 
In the case of reading in a Try pattern, it's of course find to return a bool.
So ok then for the function, that said, you'll still need to rename it with an action verb like SetAutomaticBaselineCorrection or SwitchAutomaticBaselineCorrection

Ellerbach: btw, for the naming convention and Try pattern, see https://github.com/dotnet/iot/tree/master/src/devices#naming-of-functions

src/devices/Mhz19b/Mhz19b.cs
Ellerbach: See previous comment. When related to a serial port or I2C or SPI, then it's better to raise an IOException.
If it's related to a bad reading, to non possible data, then no exception.
And good point, I'll add this in the documentation as well so it will be more clear and consistent

src/devices/Mhz19b/Mhz19b.cs
Ellerbach: Then please adjust the name with an action verb

src/devices/Mhz19b/Mhz19b.cs
fehdem: Done (renamed method to SetAutomaticBaselineCorrection)

src/devices/Mhz19b/samples/Mhz19b.Sample.cs
Ellerbach: Yes, that's an option or then just into the Readme.md in this case. So someone advance who want to calibrate will have the correct few needed lines.

src/devices/Ahtxx/AhtBase.cs
Ellerbach: then use the pattern with the i2cDevice == null

src/devices/Mhz19b/Mhz19b.cs
fehdem: Done.
Changed to string.IsNullOrEmpty. Exception changed to ArgumentException to match semantic of check.

src/devices/Mhz19b/Mhz19b.cs
fehdem: Done. Range check added. Exceed the range results in ArgumentException. Method renamed to PerformSpanPointCalibration. Parameter type changed to VolumnConcentration.

src/devices/Mhz19b/Mhz19b.cs
fehdem: Done. Method renamed to "SetSensorDectectionRange".
Returning the current setting is not possible as there is no command to read out the current setting from the sensor device.

src/devices/Mhz19b/Mhz19b.cs
fehdem: Done. Method renamed to PerformZeroPointCalibration.

src/devices/Mhz19b/samples/Mhz19b.Sample.cs
fehdem: Examples added.

src/devices/Mhz19b/README.md
fehdem: Done

src/devices/Ahtxx/AhtBase.cs
fehdem: Done

src/devices/Ahtxx/AhtBase.cs
fehdem: Done

src/devices/Ahtxx/README.md
fehdem: Category.txt added.

src/devices/Mhz19b/Mhz19b.cs
fehdem: Makes sense to me. I'll change this to an exception.

fehdem: Done.

src/devices/Mhz19b/README.md
fehdem: I don't think that I could add any valuable here.

Ellerbach (approved):
Thanks! Looks good to me know! @pgrawehr do you want to have a look as well as you started a review? If you approve, I think we will be good to merge.

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,233 @@
...
+            DataHighRequest = 0x03,
+            DataLowRequest = 0x04,
+            DataHighResponse = 0x02,
+            DataLowResponse = 0x03,
+            Checksum = 0x08
+        }
+
+        private const int MessageSize = 9;
+
+        private SerialPort _serialPort = null;
```
pgrawehr, line 42: I would prefer to having a `Stream` here (maybe with an extra constructor creating it from a serial port). Serial ports might be routed trough some other hardware, and unit testing the class is also impossible like this.

src/devices/Mhz19b/Mhz19b.cs
pgrawehr: Why do you open and close the port each time? I guess that increases the risk of getting garbage data in the communication stream.

pgrawehr:
@Ellerbach : Looks mostly good to me. With the newly added CO2-Sensor, I would prefer a stream interface instead of directly opening a serial port. That's just very inflexible (and prevents unit testing).

src/devices/Mhz19b/Mhz19b.cs
fehdem: I like that idea....and I just implemented it. I added a second constructor taking a Stream, which could by SerialPort.BaseStream. I'll push that.

pgrawehr (approved):
Thanks for adding the stream support. Looks good to me now.

fehdem:
> Thanks for adding the stream support. Looks good to me now.

You're welcome. I'm happy to contribute!

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,251 @@
...
+        private Stream _serialPortStream = null;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Mhz19b"/> class using an existing (serial port) stream.
+        /// </summary>
+        /// <param name="stream">Existing stream</param>
+        public Mhz19b(Stream stream)
```
Ellerbach, line 49: Thanks for adding the stream. Just something that pops up in my mind when I was looking at the Dispose function: 
- should the stream just be disposed like you do now?
- should we use the same behavior as for GpioController, so adding a shouldDispose?
I'm more in favor of the second option. setting to null a stream that may be open thru a device than may have multiple streams or depends of some other elements seems a bit dangerous for me.

Ellerbach (commented):
Looks good, just the way to dispose the stream is in question to me.

src/devices/Mhz19b/Mhz19b.cs
fehdem: I've to admit that I'm not that deep in the details of streams and their optimal behavior on disposal.
However, I don't mind having a look at the GpioController and try to adapt my implementation accordingly.

Ellerbach: then, just do something like that:
```csharp
private bool _shouldDispose:

public Mhz19b(Stream stream, bool shouldDispose = true)
{
  // rest of your code
}

public Dispose()
{
  if (_shouldDispose)
  {
  }
}
```

fehdem: Done and pushed.

fehdem:
@Ellerbach Laurent, is there anything else I need to do to finish this?

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        /// <summary>
+        /// Gets the current humidity reading from the sensor.
+        /// Reading the humidity takes between 10 ms and 80 ms.
+        /// </summary>
+        /// <returns>Temperature reading</returns>
+        public Ratio GetHumidity()
+        {
+            Measure();
+            return new Ratio(_humidity, UnitsNet.Units.RatioUnit.Percent);
```
Ellerbach, line 81: maybe this should be better: `return Ratio.FromPercent(_humidity);`
Doesn't really change much but for the consistency with Temperature and how it's done on other bindings.

src/devices/Ahtxx/samples/Ahtxx.Sample.cs
```diff
@@ -0,0 +1,35 @@
...
+        public static void Main(string[] args)
+        {
+            const int I2cBus = 1;
+            I2cConnectionSettings i2cSettings = new I2cConnectionSettings(I2cBus, Aht20.DeviceAddress);
+            I2cDevice i2cDevice = I2cDevice.Create(i2cSettings);
+            Aht20 aht20Sensor = new Aht20(i2cDevice);
+
+            while (true)
+            {
+                Console.WriteLine($"{DateTime.Now.ToLongTimeString()}: {aht20Sensor.GetTemperature()}, {aht20Sensor.GetHumidity()}");
```
Ellerbach, line 30: As GetTemperature() returns a Temperature, you should do something like:
```csharp
aht20Sensor.GetTemperature().DegreeCelsius
```
similar for GetHumidity() with .Ratio so you'll display in the unit you want.

src/devices/Mhz19b/samples/Mhz19b.Sample.cs
```diff
@@ -0,0 +1,71 @@
...
+            // ---- Now change to target concentration for span point.
+            // sensor.PerformSpanPointCalibration(VolumeConcentration.FromPartsPerMillion(2000));
+
+            // Continously read current concentration
+            while (true)
+            {
+                try
+                {
+                    VolumeConcentration reading = sensor.GetCo2Reading();
+                    Console.WriteLine($"{reading}");
```
Ellerbach, line 58: Similar as for Temperature and  Ratio, do something like reading.unityouwanttodisplay for clarity (where unityouwanttodisplay is the proper unit)

Ellerbach (approved):
Thanks again, few minor changes. would be perfect if @krwq can have a quick look at it. And all good on my side!

src/devices/Ahtxx/samples/Ahtxx.Sample.cs
fehdem: For the sample I actually prefer using `GetTemperature()` instead of` GetTemperature().DegreesCelsius.` If using just `GetTemperature()` it makes use of the ToString()-method which gives me a nice formatting with a reasonable number of digits and a unit. Same holds for the humidity.
I think that a user would consider the UnitNet API when using the binding and get the desired unit and full double resolution, e.g. by using `.DegreesCelsius`.
Can you agree to that?

src/devices/Mhz19b/samples/Mhz19b.Sample.cs
fehdem: Same comment as for the Ahtxx sample.

src/devices/Ahtxx/AhtBase.cs
fehdem: Done.

src/devices/Ahtxx/samples/Ahtxx.Sample.cs
Ellerbach: It's always better to be explicit on the unit when displaying it.
You can't assume it's been created in a specific measure.  And you show the user how to use the actual units. Otherwise, it's a bit magic :-)

fehdem: I like magic... ;) However, I'll change it.
What do you think about this: aht20Sensor.GetTemperature().ToUnit(TemperatureUnit.DegreeCelsius).ToString() 
I think this meets you needs. Okay? If so, I would do commit/push soon.

fehdem: I did it, as you suggest as this is slightly shorter.
I'll push the change.

src/devices/Mhz19b/samples/Mhz19b.Sample.cs
fehdem: Done

Ellerbach (approved):
Looks good to me! Thanks! @krwq you may want to have a look as well.

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,180 @@
...
+            return (GetStatusByte() & (byte)StatusBit.Calibrated) == (byte)StatusBit.Calibrated;
+        }
+
+        /// <inheritdoc cref="IDisposable" />
+        public void Dispose() => Dispose(true);
+
+        /// <inheritdoc cref="IDisposable" />
+        protected virtual void Dispose(bool disposing)
+        {
+            if (_i2cDevice == null)
```
krwq, line 171: nit: this is not needed since `?.` will handle the null check

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,254 @@
...
+            _serialPortStream = stream ?? throw new ArgumentNullException(nameof(stream));
+            _shouldDispose = shouldDispose;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Mhz19b"/> class and creates a new serial port stream.
+        /// </summary>
+        /// <param name="uartDevice">Path to the UART device / serial port, e.g. /dev/serial0</param>
+        /// <exception cref="System.ArgumentException">uartDevice is null or empty</exception>
+        public Mhz19b(string uartDevice)
```
krwq, line 61: I think for consistency you might want to consider taking SerialPort as an input here

krwq: (optional since I don't have any use case of what potential problems it would solve)

src/devices/Ahtxx/AhtBase.cs
fehdem: Correct. Done.

src/devices/Mhz19b/Mhz19b.cs
fehdem: I would rather stay with the current constructors. The one you cited hides all the details from the user for convinience. The other one gives all flexibility by accepting an existing stream. 
I don't see any use case for the third variant at the moment.

src/devices/Ahtxx/AhtBase.cs
krwq: Thread.Sleep has resolution of about 12ms, IMO if this is expected time leave this as is, if this is much smaller then go with DelayHelper

src/devices/Ahtxx/Aht20.cs
```diff
@@ -0,0 +1,22 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Device.I2c;
+
+namespace Iot.Device.Ahtxx
+{
+    /// <summary>
+    /// AHT20 temperature and humidity sensor binding.
+    /// </summary>
+    public class Aht20 : AhtBase
```
krwq, line 12: I'd prefer base class was called Aht20 but if you prefer this way at least make the base abstract

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// Note: has been tested with AHT20 only, but should work with AHT10 and AHT15 as well.
+    /// Up to now all functions are contained in the base class, though there might be differences between the sensors types.
+    /// </summary>
+    public class AhtBase : IDisposable
+    {
+        /// <summary>
+        /// Address of AHTxx device (0x38). This address is fix and cannot be changed.
+        /// This implies that only one device can be attached to a single I2C bus at a time.
+        /// </summary>
+        public const int DeviceAddress = 0x38;
```
krwq, line 23: we call this constant DefaultI2cAddress everywhere

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        // datasheet version 1.1, table 10
+        private enum StatusBit : byte
```
krwq, line 26: subtypes should be defined in the bottom of the class

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            Calibrated = 0x08,
+            Busy = 0x80
+        }
+
+        private enum Command : byte
+        {
+            Calibrate = 0xbe,
+            SoftRest = 0xba,
```
krwq, line 35: SoftReset?

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,175 @@
...
+            Measure = 0xac
+        }
+
+        private I2cDevice _i2cDevice = null;
```
krwq, line 39: you shouldn't need to initialize it, especially this will actually never be null

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// </summary>
+        /// <returns>Temperature reading</returns>
+        public Temperature GetTemperature()
+        {
+            Measure();
+            return Temperature.FromDegreesCelsius(_temperature);
+        }
+
+        /// <summary>
+        /// Gets the current humidity reading from the sensor.
```
krwq, line 74:
```suggestion
        /// Gets the current relative humidity reading from the sensor.
```

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// Reading the humidity takes between 10 ms and 80 ms.
+        /// </summary>
+        /// <returns>Temperature reading</returns>
```
krwq, line 77:
```suggestion
        /// <returns>Relative humidity reading</returns>
```

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,175 @@
...
+        private void Measure()
+        {
+            Span<byte> buffer = stackalloc byte[3]
+            {
+                // command parameters c.f. datasheet, version 1.1, ch. 5.4
+                (byte)Command.Measure,
+                0x33,
+                0x00
+            };
+            _i2cDevice.Write(buffer);
```
krwq, line 96: enter after `};`

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+            buffer = stackalloc byte[6];
+            _i2cDevice.Read(buffer);
+
+            // data format: 20 bit humidity, 20 bit temperature
+            // 7               0 7              0 7             4           0 7          0 7         0
+            // [humidity 19..12] [humidity 11..4] [humidity 3..0|temp 19..16] [temp 15..8] [temp 7..0]
+            // c.f. datasheet ch. 5.4.5
+            Int32 rawHumidity = (buffer[1] << 12) | (buffer[2] << 4) | (buffer[3] >> 4);
```
krwq, line 112: this is probably the weirdest format I've seen, so much bit twiddling just to save a single byte

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,175 @@
...
+            Int32 rawTemperature = ((buffer[3] & 0xF) << 16) | (buffer[4] << 8) | buffer[5];
+            // RH[%] = Hraw / 2^20 * 100%, c.f. datasheet ch. 6.1
+            _humidity = (rawHumidity * 100.0) / 0x100000;
```
krwq, line 115: can you create Ratio directly here? Is it possible to avoid multiplying by floating point number?

src/devices/Ahtxx/AhtBase.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// Perform calibration command sequence
+        /// </summary>
+        private void Calibrate()
+        {
+            Span<byte> buffer = stackalloc byte[3]
+            {
+                (byte)Command.Calibrate,
+                0x08, // command parameters c.f. datasheet, version 1.1, ch. 5.4
+                0x00
+            };
```
krwq, line 140: enter after `};`

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,256 @@
...
+using UnitsNet;
+
+namespace Iot.Device.Mhz19b
+{
+    /// <summary>
+    /// MH-Z19B CO2 concentration sensor binding
+    /// </summary>
+    public sealed class Mhz19b : IDisposable
+    {
+        private enum Command : byte
```
krwq, line 19: those should be defined in the bottom of the class (style guidelines)

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,256 @@
...
+            SensorNum = 0x01,
+            Command = 0x02,
+            DataHighRequest = 0x03,
+            DataLowRequest = 0x04,
+            DataHighResponse = 0x02,
+            DataLowResponse = 0x03,
+            Checksum = 0x08
+        }
+
+        private const int MessageSize = 9;
```
krwq, line 40: `MessageBytes`/`MessageBits` make sure you got unit in the name

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,256 @@
...
+            try
+            {
+                // send read command request
+                var request = CreateRequest(Command.ReadCo2Concentration);
+                request[(int)MessageFormat.Checksum] = Checksum(request);
+                _serialPortStream.Write(request, 0, request.Length);
+
+                // read complete response (9 bytes expected)
+                byte[] response = new byte[MessageSize];
+                int timeout = 100;
```
krwq, line 99: timeout should be mesaured differently i.e. (written by hand may have errors):

```
long endTicks = DateTime.Now.AddMilliseconds(100).Ticks;
while (DateTime.Now.Ticks < endTicks && whateverotherConditionsYouGot)
{
  whateverYouAreDoingRightNowMinusTimeout();
}
```

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,256 @@
...
+                int bytesRead = 0;
+                while (timeout > 0 && bytesRead < MessageSize)
+                {
+                    bytesRead += _serialPortStream.Read(response, bytesRead, response.Length - bytesRead);
+                    Thread.Sleep(1);
+                    timeout--;
+                }
+
+                if (timeout == 0)
```
krwq, line 108: you should check bytesRead here

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,256 @@
...
+                {
+                    throw new IOException("Timeout");
```
krwq, line 110: TimeoutException?

krwq: if this is relatively probable event then you should change method to `bool TryGetCo2Reading(out VolumeConcentration volumeConcentration)`

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,256 @@
...
+                if (response[(int)MessageFormat.Checksum] == Checksum(response))
+                {
+                    return VolumeConcentration.FromPartsPerMillion((int)response[(int)MessageFormat.DataHighResponse] * 256 + (int)response[(int)MessageFormat.DataLowResponse]);
+                }
+                else
+                {
+                    throw new IOException("Invalid response message received from sensor");
+                }
+            }
+            catch (Exception e)
```
krwq, line 123: you should catch only what you expect, otherwise you should let anything throw as it was. Anything specific you're guarding against?

src/devices/Ahtxx/Aht20.cs
fehdem: Done

src/devices/Ahtxx/AhtBase.cs
fehdem: Changed. But I think this name is misleading, as the device supports exactly one address and 'default' suggests that there are more options available.

src/devices/Ahtxx/AhtBase.cs
fehdem: Done

src/devices/Ahtxx/AhtBase.cs
fehdem: Typo. Done.

src/devices/Ahtxx/AhtBase.cs
fehdem: Done

src/devices/Ahtxx/AhtBase.cs
fehdem: Done

src/devices/Ahtxx/AhtBase.cs
fehdem: Agreed. However, I didn't see any better way to convert the "highly packed" data word. The magic numbers are given by the datasheet as well. 
Do you want me to change anything or can we keep this?

src/devices/Ahtxx/AhtBase.cs
fehdem: Maybe...but for the sake of clarity and relation to the datasheet I would like to stick to this. 
Is there anything wrong with using the floating point here? Precision shouldn't be an issue.

src/devices/Ahtxx/AhtBase.cs
fehdem: Done

src/devices/Mhz19b/Mhz19b.cs
fehdem: Done and will be remembered in future

src/devices/Mhz19b/Mhz19b.cs
fehdem: Good point! Done.

src/devices/Mhz19b/Mhz19b.cs
fehdem: This is a very reasonable remark. Done.

src/devices/Mhz19b/Mhz19b.cs
fehdem: Done

src/devices/Mhz19b/Mhz19b.cs
fehdem: No, this doesn't happen relatively probable. Actually I never saw this except when my test system was some how heavily loaded or the hardware was somehow disturbed. I don't see the need for a Try...-method here.

I'll change the type of the exception. I had a broader sematic in mind when I chose IOException, but eventually it is a timeout and there is no further root cause available and guessing doesn't help.

src/devices/Mhz19b/Mhz19b.cs
fehdem: Well, no. I just wanted to hide all the details from user. In any case the communication with the sensor caused the exception therefore I mapped them all to IOException to keep the exceptions lean.
Alternatively I could remove the try...catch completely. How do you handle the documentation in this case? Add a description for each possible exception?

fehdem: I removed the try...catch so that any exception will passed trough as is.
Done.

src/devices/Mhz19b/Mhz19b.cs
fehdem: Done

src/devices/Mhz19b/Mhz19b.cs
fehdem: @krwq I just found out that the first call to ```long endTicks = DateTime.Now.AddMilliseconds(100).Ticks```takes always ~105ms. The sample program is executed on an RPi 3b. 
I set a timeout of 250ms to be on the safe side.

fehdem:
@krwq , @Ellerbach : I just pushed the changes for Krzysztof's review remarks. I only left out the conversation from the message bytes into readings as I think they're good (enough). I hope we can finalize this one now, so that I can focus on further contributions. ;)

fehdem:
@krwq I just figured out that there are minor, but relevant, differences between AHT10/15 and AHT20.
I suspected this, which was the reason why I introduced an AhtxxBase in the first place. I don't have any AHT10 or AHT15 device available. Even if I would implemented bindings for those types I wouldn't be able to test them. 
I really want to get this done, therefore I would like to get some advice from you how to continue.
I see basically two options: 
a) Remove the base class and make this an Aht20 only binding. If someone needs an AHT10 or AHT15 binding later the Aht20 binding could be reused.
b) I keep the base and prepare a derived binding for Aht10/15, but leave a remark in the class comments that this is not tested with a real device, yet.

I would be good if we could finalize this pull request then, because this prevents me from continuing with the Amg88xx binding pull request (because I messed up the git branches).

src/devices/Ahtxx/AhtBase.cs
Ellerbach: With the comments and looking at the datasheet, this is ok

Ellerbach (approved):
I'm good with the current implemetnation

fehdem:
@krwq  @Ellerbach 
I just ordered an AHT10 device; delivery expected on saturday. I'll add a class covering the specific properties of the AHT10/15 device. After that the binding will support AHT10/15 and AHT20. Most functionality will remain in the common base class AhtxxBase. I think I'll be able to commit the additions at the weekend after successful testing. I'll also rework the documentation to fit with the changes. I would appreciate if we could finalize this pull request afterwards.

Ellerbach:
> I just ordered an AHT10 device; delivery expected on saturday. I'll add a class covering the specific properties of the AHT10/15 device. After that the binding will support AHT10/15 and AHT20

That's great news! Thanks a lot! We will then review this PR fast after your next commit with the adjustments

fehdem:
@Ellerbach @krwq Support for Aht10/15 added and tested with real device.

src/devices/Ahtxx/README.md
```diff
@@ -0,0 +1,40 @@
...
+### Sensor classes
+You need to choose the class depending on the sensor type.
+
+|Sensor|Required class|
+|-----|---------------|
+|AHT10|Aht10          |
+|Aht15|Aht10          |
+|Aht20|Aht20          |
+
+
```
Ellerbach, line 32: too many CR and still no example of how to use it. Maybe the few lines were place holders?

src/devices/Mhz19b/Mhz19b.cs
```diff
@@ -0,0 +1,244 @@
...
+
+            if (bytesRead < MessageBytes)
+            {
+                throw new TimeoutException($"Communication with sensor failed.");
+            }
+
+            // check response and return calculated concentration if valid
+            if (response[(int)MessageFormat.Checksum] == Checksum(response))
+            {
+                return VolumeConcentration.FromPartsPerMillion((int)response[(int)MessageFormat.DataHighResponse] * 256 + (int)response[(int)MessageFormat.DataLowResponse]);
```
Ellerbach, line 92: you should use BinaryConvert for the (DataHigh * 256 + DalaLow)

Ellerbach (approved):
Looks all good, very minor comments.

src/devices/Ahtxx/README.md
fehdem: @Ellerbach 
> Maybe the few lines were place holders?
Nice hint! :relaxed: You won! I just added some more documentation.

src/devices/Mhz19b/Mhz19b.cs
fehdem: @Ellerbach Laurent, do you mean BitConverter?
If so, I'm not sure if I feel comfortable with it. The BitConverter assumes that the byte-order of the array is equal to the byte order of the executing CPU architecture (same Endianness). Is this a valid assumption considering the multi-platform nature of .net Core? Of course I could check the endianness of the actual platform and reverse the array if required, but this seems to me a solution beyond the problem.

Ellerbach: ups, I meant BinaryPrimitives: https://docs.microsoft.com/en-us/dotnet/api/system.buffers.binary.binaryprimitives?view=netcore-3.1
Now in this case as MessafeFormat is a parameter, that may not be super straight forward. So I4m ok with that

fehdem: @Ellerbach Laurent, yes, BinaryPrimitives could make sense. However in this case I don't think that it would better compared to my current solution. Actually my implementation is even "closer" to the datasheet, if someone should look into the code and be not familar with endianness. 
So, we agree to keep it as is, I think.
I'll push the doc change.

src/devices/Ahtxx/AhtBase.cs
krwq: looks good

src/devices/Ahtxx/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+
```
krwq, line 34: the line feeds still feel a bit broken, also please remove `<br/>`s from this file

krwq (approved):
LGTM, pleaase fix the line breaks in this PR if possible

src/devices/Ahtxx/README.md
fehdem: Good point. I actually missed the updated README in my last commit. Now its included.

Ellerbach:
Thanks @fehdem  for this great addition! Looking forward for more :-)

krwq:
Thank you @fehdem!

PR #1178, opened 2020-09-09
Common project cleanup

pgrawehr:
- Clean up the common project structure (adjust to the bindings)
- Add a small extra weather function

### DIFF
--- a/src/devices/Common/CommonHelpers.csproj
+++ b/src/devices/Common/CommonHelpers.csproj
@@ -3,10 +3,15 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <EnableDefaultItems>false</EnableDefaultItems>
+    <RootNamespace></RootNamespace>
   </PropertyGroup>
 
   <ItemGroup>
     <Compile Include="*.cs" />
+	<Compile Include="Iot/Device/Common/*.cs" />
+	<Compile Include="Iot/Device/Graphics/*.cs" />
+	<Compile Include="System/Device/*.cs" />
+	<Compile Include="System/Device/Gpio/*.cs" />
     <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
   </ItemGroup>
 
--- a/src/devices/Common/Iot/Device/Common/WeatherHelper.cs
+++ b/src/devices/Common/Iot/Device/Common/WeatherHelper.cs
@@ -180,7 +180,7 @@ namespace Iot.Device.Common
         /// <param name="pressure">The pressure at the point for which altitude is being calculated</param>
         /// <param name="seaLevelPressure">The sea-level pressure</param>
         /// <param name="airTemperature">The dry air temperature at the point for which altitude is being calculated</param>
-        /// <returns>The altitude in meters</returns>
+        /// <returns>The altitude</returns>
         public static Length CalculateAltitude(Pressure pressure, Pressure seaLevelPressure, Temperature airTemperature)
         {
             double meters = ((Math.Pow(seaLevelPressure.Pascals / pressure.Pascals, 1 / 5.255) - 1) * airTemperature.Kelvins) / 0.0065;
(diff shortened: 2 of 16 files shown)


Ellerbach (commented):
Looks good to me, just make sure no other project is using this CommonHelpers.csproj

pgrawehr:
I think I checked all. There shouldn't be any wrong references floating around.

Ellerbach (approved):
If all the references has been changed, I'm good to go.

src/devices/Common/CommonHelpers.csproj
```diff
@@ -3,11 +3,16 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
```
joperezr, line 4: NIT we don't use RootNamespace in our projects so lets just clean it up from here

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -138,11 +138,36 @@ public static Temperature CalculateDewPoint(Temperature airTemperature, Ratio re
         /// <remarks>
         /// Source https://de.wikipedia.org/wiki/Luftfeuchtigkeit#Absolute_Luftfeuchtigkeit
         /// </remarks>
-        public static double CalculateAbsoluteHumidity(Temperature airTemperature, Ratio relativeHumidity)
+        public static Density CalculateAbsoluteHumidity(Temperature airTemperature, Ratio relativeHumidity)
```
joperezr, line 141: I think there was an open issue to fix this to return units right?, if this was the only one missing lets make sure to close that issue with this PR

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -138,11 +138,36 @@ public static Temperature CalculateDewPoint(Temperature airTemperature, Ratio re
...
+        /// and another temperature-only sensor that gives more reasonable ambient temperature readings.
+        /// Do note that the relative humidity is dependent on the temperature, because it depends on how much water a volume of air
+        /// can contain, which increases with temperature.
+        /// </summary>
+        /// <param name="airTemperatureFromHumiditySensor">Temperature measured by the humidity sensor</param>
+        /// <param name="relativeHumidityMeasured">Humidity measured</param>
+        /// <param name="airTemperatureFromBetterPlacedSensor">Temperature measured by better placed sensor</param>
+        /// <returns>A corrected humidity. The value will be lower than the input value if the better placed sensor is cooler than
+        /// the "bad" sensor.</returns>
+        public static Ratio CorrectRelativeHumidityFromDifferentSensor(Temperature airTemperatureFromHumiditySensor,
```
joperezr, line 160: NIT I’m definitely not great at naming methods but this one seems a bit confusing. Should we do something like `GetRelativeHumidityFromActualAirTemperature` or something likethat?

src/devices/Media/Media.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Drawing.Common" Version="4.5.1" />
+    <PackageReference Include="System.Drawing.Common" Version="$(SystemDrawingCommonPackageVersion)" />
```
joperezr, line 11: Great catch!

joperezr (approved):
Left minor comments but looks great otherwise, thanks a bunch Patrick!

src/devices/Common/CommonHelpers.csproj
pgrawehr: @joperezr : Setting it to empty here just has the effect of automatically creating new files with the correct namespace, depending on their location in the directory tree. If this line is removed, the namespace of a file added to the "Common" folder will be something like `Iot.Common.Iot.Device.Common` instead of just `Iot.Device.Common`.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
pgrawehr: #1181 is actually about a different method. I think I shouldn't touch that, since somebody else is fixing that.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
pgrawehr: Yea, I guess that's a better name. I'll change it.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
krwq: if there is no open PRs there then you're good, make sure to update the issue 😄

krwq (approved):
LGTM after fixing comments

PR #1186, opened 2020-09-18
DC Motor upgrade

wsad4ryba:
Hi,

There is a small error with duty cycle, in DCMotor2PinNoEnable.cs. Please compare also to DCMotor3Pin.cs.


src/devices/DCMotor/DCMotor2PinNoEnable.cs
```diff
@@ -73,7 +73,7 @@ public override double Speed
                         Controller.Write(_pin1, PinValue.High);
                     }
 
-                    _pwm.DutyCycle = 1.0 + val;
+                    _pwm.DutyCycle = Math.Abs(val);
```
krwq, line 76: are you seeing any specific issue? this value is negative here and pin1 is opposite to what you have for positive values which means duty cycle has to be 1-Math.Abs(val)` but since we know value is negative we can just `1+val`

krwq (commented):
please describe issue you're seeing

wsad4ryba:
I noticed the problem when I tried something like this:
```cs
for (int i = 0; i < 360; i++)
{
    speedSet = Math.Sin(Math.PI * i / 180.0);
    System.Console.WriteLine($"{i}. Actual motor setting is: {speedSet}");
    motor.Speed = speedSet;
    Thread.Sleep(100);
}
```
I start with small positive values, accelerate the engine to the maximum and then slow it down, after passing negative values it should behave the same way, but turn the other way.
And this is where the problem begins, when small negative values appeared, the engine started to spin very quickly, then slowed down and then accelerated again, which is opposite to the positive values, and yet it should behave similarly but spin the other way.
The directional pin changes the direction in which the current flows through the motor, while the motor speed is proportional to the duty cycle of PWM, because the average output voltage is proportional to the duty cycle. So if the motor should turn slowly we set a small value in the duty cycle, if quickly this value should be high, regardless of the direction because the direction is changed with the direction pin.
Let's see what happens if we want to set a low negative value to motor.Speed, let's take -0.1 for example:
- if we use the old version we get 1 + (-0.1) = 0.9 so the engine spins fast which means bad performance.
- if we use the new version, -0.1 will turn into 0.1 so the motor will turn slowly, and by changing the directional pin it will turn in the opposite direction, so it does exactly what it should.

I tested it using this motor driver:
https://www.cytron.io/p-13amp-6v-30v-dc-motor-driver
and here you can find sample code to handle it:
https://github.com/CytronTechnologies/CytronMotorDriver/blob/master/CytronMotorDriver.cpp

I hope my answer is sufficient ;)

krwq:
can you show how you initialized the DCMotor?

pgrawehr:
Shouldn't `pwm.DutyCycle` throw an exception if the value is not within the range 0.0-1.0?

wsad4ryba:
@pgrawehr - yes, but in both cases, pwm.DutyCycle is positive but differs in value

@krwq - I think I understand where the problem is, I didn't look closely enough at your driver because I thought that "all H bridges are the same" - but not quite 😄 
In two pin mode you pull up EN input and you using second directional pin as PWM input, that why you need reversed pwm.DutyCycle, because in one direction the active signal is HIGH but in other direction the active signal is LOW.
In my driver two directional pins have been replaced by one because its creators have probably stated that people most often use simple movement in one direction or the other and probably no one will want to use additional modes like quick stop in your controller(or quick stop is the default mode in my controller, but since I have a large gearbox in my motor, it's hard for me to tell because it stops quickly anyway 😄  ). When I saw that in your class I can use pwm and one pin to control the directions I thought it was designed for my driver and I didn't ask any more questions 😄 

Going back to my layout, it is actually easier to control and understand, because the directional pin only changes the direction, while the pwm input controls the speed, the active state is always HIGH, so as I wrote before, low dutycycle values are low speed, high dutycycle values it's high speed, no matter the direction.

Initialization:
```csharp
DCMotor motor = DCMotor.Create(PwmChannel.Create(0, 1, 2000, 0.0), DIRpin, controler, false)
```

Wiring diagram:
DCmotor
I am using hardware pwm with dtoverlay = pwm-2chan.

What do you propose to keep functionality for both drivers? New parameter?

Or we can use your DCMotor3Pin class, if I use the same value in both directional pins, it should work with my driver, I can check it next week, although I'm afraid it won't be intuitive for everyone, for me it wasn't 😄 
```csharp
DCMotor motor = DCMotor.Create(PwmChannel.Create(0, 1, 2000, 0.0), DIRpin, DIRpin, controler, false)
```

krwq:
IMO (assuming we don't support your scenario already):
- change existing API to rename directionPin to pin2 or something not suggesting this is similar to your scenario
- create new class for your specific setup (or create a new mode in existing one - although for clarity might be better to create a new one)
- add new static API for Create, naming to be discussed (perhaps new bool parameter or something like CreateFromPwmChannelAndDirectionPin, TBD)
- update docs to explain how to use this setup and mention in the old one your scenario to make it less likely for someone to misuse it

src/devices/DCMotor/DCMotor.cs
```diff
@@ -69,16 +69,16 @@ protected virtual void Dispose(bool disposing)
         }
 
         /// <summary>
-        /// Creates <see cref="DCMotor"/> instance which allows to control speed in one direction.
+        /// Creates <see cref="DCMotor"/> instance with 3pin driver using only one pin which allows to control speed in one direction.
```
krwq, line 72: `3-pin` or `3 pin` but I'd refrain assuming someone is using driver as they could just use transistor or connect directly for small motors (`using only one pin` is fine)

src/devices/DCMotor/DCMotor.cs
```diff
@@ -89,18 +89,18 @@ public static DCMotor Create(PwmChannel speedControlChannel)
...
         /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
         /// <returns><see cref="DCMotor"/> instance</returns>
         /// <remarks>
         /// <paramref name="speedControlPin"/> can be connected to either enable pin of the H-bridge.
-        /// or directly to the input related with the motor (if H-bridge allows inputs to change frequently).
+        /// or directly to the on of two inputs related with the motor direction (if H-bridge allows inputs to change frequently).
         /// Connecting motor directly to GPIO pin is not recommended and may damage your board.
         /// </remarks>
-        public static DCMotor Create(int speedControlPin, GpioController controller = null, bool shouldDispose = true)
+        public static DCMotor Create3pinDriver(int speedControlPin, GpioController controller = null, bool shouldDispose = true)
```
krwq, line 103: same

src/devices/DCMotor/DCMotor.cs
```diff
@@ -116,20 +116,34 @@ public static DCMotor Create(int speedControlPin, GpioController controller = nu
         }
 
         /// <summary>
-        /// Creates <see cref="DCMotor"/> instance which allows to control speed in both directions.
+        /// Starts the motor by starting pwm with last speed set.
+        /// </summary>
+        public virtual void Start()
```
krwq, line 121: isn't this equivalent to `Speed = 0` and `Speed != 0`? For this functional API change I'd recommend creating a separate issue

src/devices/DCMotor/DCMotor.cs
```diff
@@ -116,20 +116,34 @@ public static DCMotor Create(int speedControlPin, GpioController controller = nu
...
         /// <remarks>
-        /// <paramref name="speedControlChannel"/> can be connected to either enable pin of the H-bridge.
-        /// or directly to the input related with the motor (if H-bridge allows inputs to change frequently).
+        /// <paramref name="speedControlChannel"/> should be connected to the one of two inputs
+        /// related with the motor direction (if H-bridge allows inputs to change frequently).
         /// <paramref name="directionPin"/> should be connected to H-bridge input corresponding to one of the motor inputs.
         /// Connecting motor directly to GPIO pin is not recommended and may damage your board.
         /// </remarks>
-        public static DCMotor Create(PwmChannel speedControlChannel, int directionPin, GpioController controller = null, bool shouldDispose = true)
+        public static DCMotor Create3pinDriver(PwmChannel speedControlChannel, int directionPin, GpioController controller = null, bool shouldDispose = true)
```
krwq, line 146: same

krwq: here specifically I was expecting you'd change description of directionPin and make name less confusing.. also another comment is that I'd personally consider this 2 pin (i.e. 1-wire protocol is actually using 2 wires as they don't consider  ground as a pin)

src/devices/DCMotor/DCMotor.cs
```diff
@@ -145,20 +159,20 @@ public static DCMotor Create(PwmChannel speedControlChannel, int directionPin, G
...
         /// <remarks>
-        /// PWM pin <paramref name="speedControlPin"/> can be connected to either enable pin of the H-bridge.
-        /// or directly to the input related with the motor (if H-bridge allows inputs to change frequently).
+        /// PWM pin <paramref name="speedControlPin"/> should be connected to the one of two inputs
+        /// related with the motor direction (if H-bridge allows inputs to change frequently).
         /// <paramref name="directionPin"/> should be connected to H-bridge input corresponding to one of the motor inputs.
         /// Connecting motor directly to GPIO pin is not recommended and may damage your board.
         /// </remarks>
-        public static DCMotor Create(int speedControlPin, int directionPin, GpioController controller = null, bool shouldDispose = true)
+        public static DCMotor Create3pinDriver(int speedControlPin, int directionPin, GpioController controller = null, bool shouldDispose = true)
```
krwq, line 175: same comment

src/devices/DCMotor/DCMotor.cs
```diff
@@ -194,7 +208,7 @@ public static DCMotor Create(int speedControlPin, int directionPin, GpioControll
         /// <paramref name="otherDirectionPin"/> should be connected to H-bridge input corresponding to the remaining motor input.
         /// Connecting motor directly to GPIO pin is not recommended and may damage your board.
         /// </remarks>
-        public static DCMotor Create(PwmChannel speedControlChannel, int directionPin, int otherDirectionPin, GpioController controller = null, bool shouldDispose = true)
+        public static DCMotor Create3pinDriver(PwmChannel speedControlChannel, int directionPin, int otherDirectionPin, GpioController controller = null, bool shouldDispose = true)
```
krwq, line 211: whatever name you pick for the new driver, this name should also change as it's similar concept except it's using 2 pins for direction rather than one

src/devices/DCMotor/DCMotor2Pin.cs
```diff
@@ -0,0 +1,104 @@
...
+        private double _speed;
+
+        public DCMotor2Pin(
+            PwmChannel pwmChannel,
+            int pin1,
+            GpioController controller,
+            bool shouldDispose)
+            : base(controller ?? ((pin1 == -1) ? null : new GpioController()), controller == null ? true : shouldDispose)
+        {
+            _pwm = pwmChannel;
```
krwq, line 25: please remove extra spacing

src/devices/DCMotor/DCMotor2Pin.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Device.Gpio;
+using System.Device.Pwm;
+using System.Device.Pwm.Drivers;
+
+namespace Iot.Device.DCMotor
+{
+    internal class DCMotor2Pin : DCMotor
+    {
+        private PwmChannel _pwm;
+        private int _pin1;
```
krwq, line 15: I assume this is a real direction pin now, if so please rename

src/devices/DCMotor/DCMotor2Pin.cs
```diff
@@ -0,0 +1,104 @@
...
+                    _pwm.DutyCycle = val;
+                }
+                else
+                {
+                    if (_pin1 != -1)
+                    {
+                        Controller.Write(_pin1, PinValue.Low);
+                    }
+
+                    _pwm.DutyCycle = Math.Abs(val);
```
krwq, line 86: why do you Abs if you know that value is negative and is `>= -1`

krwq (changes requested):
Thanks for doing the work! I'm glad you haven't updated the samples yet since I'm expecting we will need to adjust names. I suggest to start the discussion on names before updating the PR. Once we agree the new names are less confusing I suggest going ahead and changing. The overall shape looks good I only find new names maybe a bit more confusing (i.e. what you call 3-pin I'd consider 2 pin etc.; also I'd prefer to put purpose/disambiguity rather than setup in the public API names)

krwq:
Ok, let's try to move this forward:

IMO following things should happen in here:
- leave all previous methods with same name (Create) but adjust parameter name to not use phrase "direction" and "pin" in the same name unless it actually is controlling direction (I think there was one case where there were two pins responsible for direction and they should be negated at all times which counts as this case`*`)
- the new case you proposed here is a better version of existing overload (described above and marked with `*`), since this overload with just different argument name would be ambigious I suggest we describe this case in the remarks of the above case but call it differently, i.e.: `CreateWithDirectionControlPin` (name open for suggestions, other option is to allow for passing `-1` as one of the arguments and additional bool called `singleBiDirectionPin`), make sure to point back to other API if using different name (actually now writing this I like just the extra bool argument)
- remove extra APIs from this PR, I'm fine with creating an issue and discussing this again but this specific API has already been briefly discussed and we decided not to have it (fine to reconsider but outside of this PR as it has different concern)

src/devices/DCMotor/DCMotor.cs
wsad4ryba: While writing this I was just thinking about the L298N driver that was used in the samples.

src/devices/DCMotor/DCMotor.cs
```diff
@@ -69,16 +69,16 @@ protected virtual void Dispose(bool disposing)
...
         /// <param name="speedControlChannel"><see cref="PwmChannel"/> used to control the speed of the motor</param>
         /// <returns>DCMotor instance</returns>
         /// <remarks>
         /// PWM pin <paramref name="speedControlChannel"/> can be connected to either enable pin of the H-bridge.
-        /// or directly to the input related with the motor (if H-bridge allows inputs to change frequently).
+        /// or directly to the one of two inputs related with the motor direction (if H-bridge allows inputs to change frequently).
         /// Connecting motor directly to GPIO pin is not recommended and may damage your board.
         /// </remarks>
-        public static DCMotor Create(PwmChannel speedControlChannel)
+        public static DCMotor Create3pinDriver(PwmChannel speedControlChannel)
```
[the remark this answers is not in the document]
wsad4ryba: You are right, changing this name will cause backwards compatibility issues.

src/devices/DCMotor/DCMotor.cs
wsad4ryba: I will create a new issue and explain what I wanted to achieve.

src/devices/DCMotor/DCMotor.cs
wsad4ryba: As I mentioned, I took the 3 pin from the L298N driver because it has 3 pins to support one motor, how many of them we will use is another matter, I thought that this way it would be easier to distinguish which driver we are referring to (my has 2 active pins, I also don't consider ground as a pin).
In this case, directionPin does actually change direction, so I left the old name.
We can change the name to input2 but then it would consistently change speedControlChannel to input1, and the type would be a hint what signal will be there?

src/devices/DCMotor/DCMotor2Pin.cs
wsad4ryba: probably because i saw it in DCMotor3Pin 😄

wsad4ryba:
Forgive the late reply.
I agree that we should go back to the old name (Create). As for the renaming of the pins, it will be a bit tricky because apart from the one that carries the PWM signal, the rest of the pins (one or two), are actually responsible for the change of direction 😄 
I'm afraid that it will not be easy to avoid misunderstandings, but I like the idea with an additional parameter.
The case marked with * is in the DCMotor3Pin.cs file.
The idea with an additional parameter is similar to the one I mentioned at the beginning, using DCMotor3Pin:
```csharp
DCMotor motor = DCMotor.Create (PwmChannel.Create (0, 1, 2000, 0.0), DIRpin, DIRpin, controler, false)
```
in this form it will not work, but if we add (inside the constructor) checking if the pin has not been opened previously, it has a chance to work. if pin0 = pin1 its value will be changed twice but finally its value will be HIGH for val> 0 and LOW for val <0. What do you think about such solution?
If it is not intuitive for you, we will go ahead with an additional parameter, but I think in this case we should add it in DCMotor2PinNoEnable?

krwq:
@wsad4ryba I'd suggest simply adding new class and create different instance in the Create method, unless changing implementation is simply adding 1 if statement somewhere (but still I think new instance might be beneficial if that would make code more confusing). I do not want to use side effects even if this will make it work as it makes the code hard to maintain (there is a chance someone will find another weird case for another DC Motor and at this point we might forget it was used as a side effect and break it by accident)

src/devices/DCMotor/DCMotor.cs
```diff
@@ -116,20 +116,24 @@ public static DCMotor Create(int speedControlPin, GpioController controller = nu
...
         /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
         /// <returns><see cref="DCMotor"/> instance</returns>
         /// <remarks>
-        /// <paramref name="speedControlChannel"/> can be connected to either enable pin of the H-bridge.
-        /// or directly to the input related with the motor (if H-bridge allows inputs to change frequently).
+        /// <paramref name="speedControlChannel"/> should be connected to the one of two inputs
+        /// related with the motor direction (if H-bridge allows inputs to change frequently),
+        /// or to PWM input if a controller with one direction input is used.
         /// <paramref name="directionPin"/> should be connected to H-bridge input corresponding to one of the motor inputs.
+        /// or to DIR input if a controller with one direction input is used.
```
krwq, line 133:
```suggestion
        /// or to direction input if a controller with one direction input is used.
```

alternative use paramref with correct name

src/devices/DCMotor/DCMotor.cs
```diff
@@ -116,20 +116,24 @@ public static DCMotor Create(int speedControlPin, GpioController controller = nu
...
         /// Connecting motor directly to GPIO pin is not recommended and may damage your board.
         /// </remarks>
-        public static DCMotor Create(PwmChannel speedControlChannel, int directionPin, GpioController controller = null, bool shouldDispose = true)
+        public static DCMotor Create(PwmChannel speedControlChannel, int directionPin, bool singleBiDirectionPin, GpioController controller = null, bool shouldDispose = true)
```
krwq, line 136: new argument should be false by default and be last on the list to allow source compatibility (while we allow for some breaking changes we should not make them unless not necessary and it's not here)

src/devices/DCMotor/DCMotor.cs
```diff
@@ -141,24 +145,35 @@ public static DCMotor Create(PwmChannel speedControlChannel, int directionPin, G
...
-        /// or directly to the input related with the motor (if H-bridge allows inputs to change frequently).
+        /// <paramref name="speedControlPin"/> should be connected to the one of two inputs
+        /// related with the motor direction (if H-bridge allows inputs to change frequently),
+        /// or to PWM input if a controller with one direction input is used.
         /// <paramref name="directionPin"/> should be connected to H-bridge input corresponding to one of the motor inputs.
+        /// or to DIR input if a controller with one direction input is used.
         /// Connecting motor directly to GPIO pin is not recommended and may damage your board.
         /// </remarks>
-        public static DCMotor Create(int speedControlPin, int directionPin, GpioController controller = null, bool shouldDispose = true)
+        public static DCMotor Create(int speedControlPin, int directionPin, bool singleBiDirectionPin, GpioController controller = null, bool shouldDispose = true)
```
krwq, line 176: same comment: false by default and put it as last on the list

src/devices/DCMotor/DCMotor2Pin.cs
```diff
@@ -0,0 +1,94 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.Gpio;
+using System.Device.Pwm;
+using System.Device.Pwm.Drivers;
+
+namespace Iot.Device.DCMotor
+{
+    internal class DCMotor2Pin : DCMotor
```
krwq, line 12: please add a comment on top explaining the scenario

src/devices/DCMotor/DCMotor2Pin.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+        private PwmChannel _pwm;
+        private int _DIRpin;
```
krwq, line 15: naming convention says to use `_dirPin`

src/devices/DCMotor/DCMotor2Pin.cs
```diff
@@ -0,0 +1,94 @@
...
+        private double _speed;
+
+        public DCMotor2Pin(
+            PwmChannel pwmChannel,
+            int pin1,
```
krwq, line 20: I'd use `dirPin` as name to match field name

src/devices/DCMotor/DCMotor2Pin.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            _pwm = pwmChannel;
+
+            _DIRpin = pin1;
+
+            _speed = 0;
+
+            _pwm.Start();
+
+            if (_DIRpin != -1)
```
krwq, line 33: I'd say you should throw if it's not defined since we already have a class which handles that specific case where this is `-1`. You cannot have bi-directional pin and not have it at the same time

src/devices/DCMotor/DCMotor2Pin.cs
krwq: I'd suggest to put more explicit name as well, i.e. `DCMotor2PinWithBiDirectionalPin`

src/devices/ExplorerHat/Motors.cs
```diff
@@ -75,8 +75,8 @@ internal Motors(GpioController controller, bool shouldDispose = true)
 
             _motorArray = new List<DCMotor.DCMotor>()
             {
-                DCMotor.DCMotor.Create(MOTOR1_SPDPIN, MOTOR1_DIRPIN, _controller),
-                DCMotor.DCMotor.Create(MOTOR2_SPDPIN, MOTOR2_DIRPIN, _controller)
+                DCMotor.DCMotor.Create(MOTOR1_SPDPIN, MOTOR1_DIRPIN, false, _controller),
```
krwq, line 78: this code should not have to change (if you make it default to false and put it as last on the list that shouldn't be needed)

krwq:
@wsad4ryba awesome, we're getting close here! Main piece of feedback is to put new argument last on the list and make it default to false so that previous code still compiles and works without changing semantics. Now next steps minus feedback:
- add a simple sample code here explaining this new scenario
- (optionally but recommended) add a diagram to the sample code
- make sure to fix build errors (CI is not clean but build errors are valid)

and I think we're good to merge with this PR!

src/devices/DCMotor/DCMotor.cs
wsad4ryba: I wanted to make future users think twice about what driver they are using 😄, but you are right, compatibility is more important

src/devices/DCMotor/DCMotor.csproj
```diff
@@ -11,6 +11,7 @@
     <Compile Include="DCMotor.cs" />
     <Compile Include="DCMotor3Pin.cs" />
     <Compile Include="DCMotor2PinNoEnable.cs" />
+	<Compile Include="DCMotor2PinWithBiDirectionalPin.cs" />
```
krwq, line 14:
```suggestion
    <Compile Include="DCMotor2PinWithBiDirectionalPin.cs" />
```

(tab instead of spaces)

src/devices/DCMotor/DCMotor2PinWithBiDirectionalPin.cs
```diff
@@ -0,0 +1,99 @@
...
+            int dirpin,
+            GpioController controller,
+            bool shouldDispose)
+            : base(controller ?? ((dirpin == -1) ? null : new GpioController()), controller == null ? true : shouldDispose)
+        {
+            _pwm = pwmChannel;
+
+            _dirPin = dirpin;
+
+            _speed = 0;
```
krwq, line 32:
```suggestion
            _pwm = pwmChannel;
            _dirPin = dirpin;
            _speed = 0;
```

src/devices/DCMotor/DCMotor2PinWithBiDirectionalPin.cs
```diff
@@ -0,0 +1,99 @@
...
+        private PwmChannel _pwm;
+        private int _dirPin;
+        private double _speed;
+
+        public DCMotor2PinWithBiDirectionalPin(
+            PwmChannel pwmChannel,
+            int dirpin,
+            GpioController controller,
+            bool shouldDispose)
+            : base(controller ?? ((dirpin == -1) ? null : new GpioController()), controller == null ? true : shouldDispose)
```
krwq, line 26:
```suggestion
            : base(controller ?? ((dirpin == -1) ? null : new GpioController()), controller == null || shouldDispose)
```

krwq:
LGTM, there are some cosmetic changes left and please update the docs to mention this case

wsad4ryba:
I'm installing fritzing right now, so as soon as I have some time next week I'll try to complete the documentation.
Well, unless the paint scheme is enough? However, from what I've seen, you use fritzing everywhere.

krwq:
@wsad4ryba Fritzing is preferred but not required (it looks nicer if you use it). Unless you're very talented I'd avoid Paint for drawing circuits 😄 We do not require any picture but for APIs confusing as this one (since there are many ways to use this) I'll recommend it, I do want at minimum text description and a sample if there is no picture though.

krwq (approved):
LGTM

krwq:
@Ellerbach do you want to take an extra look or we can go ahead with the merge?

Ellerbach (approved):
@krwq I have followed the discussion all along. And it looks good to me ready for merge.

krwq:
Thanks @wsad4ryba!

wsad4ryba:
Thank you for all your tips.

Issue #1086, opened 2020-05-25
TLC1543 Device Binding


SaDiablo:
Hi,
I'm working on a student's science club [project](https://github.com/ACT-ZUT/AlphaBot2.NET) using [AlphaBot2](https://www.waveshare.com/wiki/AlphaBot2-Pi) robot. There are a few yet not implemented devices here.
One of them is TLC1543 - 11 channel 10 bit ADC [[Documentation]](https://www.ti.com/lit/ds/symlink/tlc1543.pdf)
In this robot it is used for converting analog reflective infrared sensor (ITR20001/T) and battery voltage

I already have a bit of binding working (fast mode 1, without using EOC pin) and it's my first time contributing to another project. How should I proceed? And also should I file rest of the missing bindings here?

pgrawehr:
There's a [template](https://github.com/dotnet/iot/blob/master/tools/templates/DeviceBindingTemplate/README.md) for how a binding could look like. Or just have a look at some of the existing bindings (i.e for an ADC it might be the ADS1115 binding). If you've got some .NET experience, you should be able to undestand the concept and general guidelines rather quickly by lookign at what's already there. 

I personally think just listing what's missing is not very useful, since that list will eventually be very long. Maybe list those that you think would be most helpful for others, too (i.e. not exotic chips), or those where you think you would be able to contribute yourself.

krwq:
As long as you implement functionality which will cover most common use cases you don't need to implement it all. As for mentioning what's missing: yes mention but not with too many details, just what might be interesting to people (i.e. if device supports both I2C/SPI but you only add 1 then mention it, if it's some implementation detail like interrupts or something then not necessarily).

Ellerbach:
@SaDiablo, have you made any progress? Do you need any help?

SaDiablo:
> @SaDiablo, have you made any progress? Do you need any help?

I need to comment out few things, add readme and few examples and then it should be ready to go. Thanks for asking :)

Ellerbach:
@SaDiablo , check as well the documentation update for couple of Do and Don't: https://github.com/dotnet/iot/tree/master/src/devices
That should help you as well!

Ellerbach:
[Triage] This has been added already. See https://github.com/dotnet/iot/tree/main/src/devices/Tlc1543

PR #1192, opened 2020-09-22
Add device binding for Tlc1543 binding

SaDiablo:
Added Tlc1543 binding as reported in #1086


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1192) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1192) before we can accept your contribution.<br/><br/>:x: SaDiablo [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1192)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1192) it.</sub>

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+    /// </summary>
+    public class Tlc1543 : IDisposable
+    {
+        #region Members
+
+        /// <summary>
+        /// Channel used between readings to charge up capacitors in ADC
+        /// </summary>
+        public Channel ChargeChannel = Channel.SelfTest512;
+        private readonly int _chipSelect;
```
Ellerbach, line 23: You have to start with the private readonly memebers
So the public property should be after

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+        #region Constructors
+
+        /// <summary>
+        /// Constructor for Tlc1543
+        /// </summary>
+        /// <param name="ADDR">Address</param>
+        /// <param name="CS">Chip Select</param>
+        /// <param name="DOUT">Data Out</param>
+        /// <param name="IOCLK">I/O Clock</param>
+        public Tlc1543(int ADDR, int CS, int DOUT, int IOCLK)
```
Ellerbach, line 43: You have to have as well a constructor where you pass the GpioController, can be null and use as well the shouldDispose pattern. See here: https://github.com/dotnet/iot/tree/master/src/devices#gpiocontroller

Please use names which are in line with the naming convention, so preffer Address to ADDR, ChipSelect to CS, etc.
Also, fir parameters, the first letter is lower case, so for ADDR, it will be address, for CS chipSelect (note that then you have to keep the uppercase for the next word).
This comment is valid for all the rest of this PR.

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+        /// Channel used between readings to charge up capacitors in ADC
+        /// </summary>
+        public Channel ChargeChannel = Channel.SelfTest512;
+        private readonly int _chipSelect;
+        private readonly int _dataOut;
+        private readonly int _address;
+        private readonly int _inputOutputClock;
+        private readonly int _endOfConversion;
+        private readonly int _channels = 11;
+        private GpioController _digital = new GpioController(PinNumberingScheme.Logical);
```
Ellerbach, line 29: See comment regarding the constructor. See https://github.com/dotnet/iot/tree/master/src/devices#gpiocontroller for the right pattern to use

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        /// <summary>
+        /// Mode 1 (Fast mode):
+        /// <br>10 clock transfer with !CS high </br>
+        /// <br>Figure 9 in datasheet</br>
+        /// </summary>
+        /// <param name="channelNumber">Channel to read</param>
+        /// <returns>10 bit value corresponding to relative voltage level on channel</returns>
+        public int Read(int channelNumber)
```
Ellerbach, line 136: What is the difference between this Read function and the ReadChannel function? (and by extention the ReadChannels).
Should this one be private maybe?

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+        {
+            int value = 0;
+            _digital.Write(_chipSelect, 0);
+            DelayHelper.DelayMicroseconds(1, false);
```
Ellerbach, line 140: Please refer to the documentation for the delays. see as well the recommendation for this topic: https://github.com/dotnet/iot/tree/master/src/devices#commenting-the-code

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+
+        /// <summary>
+        /// Mode 1 (Fast mode):
+        /// <br>10 clock transfer with !CS high </br>
+        /// <br>Figure 9 in datasheet</br>
+        /// </summary>
+        /// <param name="channelNumber">Channel to read</param>
+        /// <returns>10 bit value corresponding to relative voltage level on channel</returns>
+        public int Read(int channelNumber)
+        {
```
Ellerbach, line 137: And if you make this function public, you should check the validity of channelNumber. 
What will happen if I pass -3 for example? It should raise an ArgumentException.

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        #endregion
+
+        #region Enums
+
+        /// <summary>
+        /// Available Channels to poll from on Tlc1543
+        /// </summary>
+        public enum Channel
```
Ellerbach, line 180: please move this enum into a separate file

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+
+        #endregion
+
+        #region Disposing
+
+        /// <summary>
+        /// Dispose
+        /// </summary>
+        /// <param name="disposing">is disposing</param>
+        protected virtual void Dispose(bool disposing)
```
Ellerbach, line 261: please see the pattern as well regarding GpioController

src/devices/Tlc1543/samples/README.md
```diff
@@ -0,0 +1,4 @@
+# Sample
+This sample shows
```
Ellerbach, line 2: Maybe you can extract few code elements, it's always nice to have some in the documentation

src/devices/Tlc1543/README.md
```diff
@@ -0,0 +1,15 @@
...
+## Summary
+High-speed 10-bit switched-capacitor successive-approximation A/D Converter with 14 channels (11 inputs and 3 self-test channels)
+
+## Device Family
+**[TLC1542/3]**: [Datasheet](https://www.ti.com/lit/ds/symlink/tlc1543.pdf)
+
+## Binding Notes
+(For now) Implemented Fast Mode 1 which doesn't need EOC pin connected. Respective timing diagram can be seen on figure 9 in datasheet.
+
+It is possible to change ADC charge channel.
```
Ellerbach, line 12: Maybe add as well how to create the ADC element and how to read? I understand it's just super easy and straight forward but those little things helps a lot.

src/devices/Tlc1543/samples/Tlc1543.Sample.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Device.Gpio;
+using System.Device.I2c;
```
Ellerbach, line 8: Please clean the using you don't need

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+        {
+            if (!_disposedValue)
+            {
+                if (disposing)
+                {
+                    // TODO: dispose managed state (managed objects)
```
Ellerbach, line 267: I guess you can remove the TODO as seems that you've done it

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+                {
+                    value |= 0x01;
+                }
+
+                _digital.Write(_inputOutputClock, 1);
+                _digital.Write(_inputOutputClock, 0);
+            }
+
+            _digital.Write(_chipSelect, 1);
+            DelayHelper.DelayMicroseconds(100, true); // t(PZH/PZL)
```
Ellerbach, line 168: please place the comment up the line and please refer as well to the documentation for the delay

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+        private readonly int _channels = 11;
+        private GpioController _digital = new GpioController(PinNumberingScheme.Logical);
+        private bool _disposedValue;
+
+        #endregion
+
+        #region Constructors
+
+        /// <summary>
+        /// Constructor for Tlc1543
```
Ellerbach, line 37: Looking at the full code, it seems that this sensor can work as well with SPI. So would be nice as well to have a constructor and also the capacity to have it working with SPI if that's the case.

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+    /// Add documentation here
+    /// </summary>
+    public class Tlc1543 : IDisposable
+    {
+        #region Members
+
+        /// <summary>
+        /// Channel used between readings to charge up capacitors in ADC
+        /// </summary>
+        public Channel ChargeChannel = Channel.SelfTest512;
```
Ellerbach, line 22: please use { get; set; } for public properties like this one. Also, seems like this is more a readonly if I understand correctly or that if you change it, it you may not have all the channels possible, right? In this case, please adjust this property.

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,289 @@
...
+
+                // read 10-bit data
+                value <<= 1;
+                if (_digital.Read(_dataOut) == PinValue.High)
+                {
+                    value |= 0x01;
+                }
+
+                _digital.Write(_inputOutputClock, 1);
+                _digital.Write(_inputOutputClock, 0);
```
Ellerbach, line 164: Don't you have to wait a bit?

Ellerbach (changes requested):
Thanks for your first PR. Couple of comments and few changes required. But all up, it's a great start.

src/devices/Tlc1543/Tlc1543.cs
SaDiablo: I'm not experienced in SPI standard, so wasn't looking for that, but also I haven't seen any mention of proper SPI support in documentation. This may be something to consider in the future.

src/devices/Tlc1543/Tlc1543.cs
SaDiablo: I would have added delays there but simply rising edge (same with falling one) is 2 times as long as nominal time for the whole IOCLK period. Controller.Write is slow enough that adding any delay (even 1 tick long) doubles the length of IOCLK period and having that 10 to 20 times per conversions adds up quickly for performance hit that makes whole ADC polling one channel in ~1kHz speeds (instead of nominal 2MHz)

src/devices/Tlc1543/Tlc1543.cs
SaDiablo: ChargeChannel is only used between reading channel (or at the end of reading list of channels) because this device gives out data on the second cycle so having dummy value being written there helps in keeping it tidy - also helps with negating interferences between fast channel polling by reseting to one of three values 0 (SelfTest0), 512 (SelfTest512) or 1023(SelfTest1023) or if someone has a reason - there is also possibility to use one of the normal channels as a charge channel

src/devices/Tlc1543/Tlc1543.cs
Ellerbach: What would happen on a real time processor like an ESP32 if we migrate this code into nanoFramework?
And that's then one more element to use SPI as well, this then will be done automatically by the clock pin

src/devices/Tlc1543/Tlc1543.cs
Ellerbach: Fair enough. Let's keep then this binding with the GPIO only support so far.

src/devices/Tlc1543/Tlc1543.cs
Ellerbach: OK, so if only one of the 3 are usable, then please check the set and raise an ArugumentException if one of the 3 values are not used.
This would then looks like:
```csharp
private Channel _chargeChannel;
public Channel ChargeChannel
{
  get => _chargeChannel;  
  set
  {
    if ((value == Channel.SelfTest512) || (value == Channel.SelfTest0) || (value == Channel.SelfTest1023))
    {
      _chargeChannel = value;
    }
    else
    {
      thro new ArgumentException($"{ChargeChannel} can only be {Channel.SelfTest512} or {Channel.SelfTest0} or {Channel.SelfTest1023}");
    }
  }
}
```

SaDiablo: I've made it like this cause if someone has the reason to use normal channel as a charge channel they will be able to. It isn't forbidden to use them in this way, so i see no reason to limit them to only selftest channels
```c#
private Channel _chargeChannel = Channel.SelfTest512;
public Channel ChargeChannel
        {
            get => _chargeChannel;
            set
            {
                if ((int)value >= 0 & (int)value < 14)
                {
                    _chargeChannel = value;
                }
                else
                {
                    throw new ArgumentOutOfRangeException("ChargeChannel", "Out of range. \nMust be non-negative and less than 14. \nRefer to table 2 and 3 in TLC1543 documentation for more information");
                }
            }
        }
```

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,239 @@
...
+
+        /// <summary>
+        /// Constructor for Tlc1543
+        /// </summary>
+        /// <param name="address">Address</param>
+        /// <param name="chipSelect">Chip Select</param>
+        /// <param name="dataOut">Data Out</param>
+        /// <param name="endOfConversion">End of Conversion</param>
+        /// <param name="inputOutputClock">I/O Clock</param>
+        public Tlc1543(int address, int chipSelect, int dataOut, int inputOutputClock, int endOfConversion)
```
krwq, line 89: is this using SPI? if so please use SpiDevice (you can use SoftwareSpi for manual pins)

krwq: (see sample in Mcp3008 for how to do that)

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,239 @@
...
+            _address = address;
+            _chipSelect = chipSelect;
+            _dataOut = dataOut;
+            _endOfConversion = endOfConversion;
+            _inputOutputClock = inputOutputClock;
+
+            _digital.OpenPin(_address, PinMode.Output);
+            _digital.OpenPin(_chipSelect, PinMode.Output);
+            _digital.OpenPin(_dataOut, PinMode.InputPullUp);
+            _digital.OpenPin((int)_endOfConversion, PinMode.InputPullUp);
```
krwq, line 100: why does this need cast? should this support passing -1 if this is not supported?

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,239 @@
...
+        #region Members
+
+        private readonly int _chipSelect;
+        private readonly int _dataOut;
+        private readonly int _address;
+        private readonly int _inputOutputClock;
+        private readonly int? _endOfConversion;
+        private readonly int _channels = 11;
+        private readonly TimeSpan _conversionTime = new TimeSpan(210);
+        private GpioController _digital;
```
krwq, line 26: please call this `_controller` or `_gpioController`

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,239 @@
...
+            }
+
+            int value = 0;
+            _digital.Write(_chipSelect, 0);
+            for (int i = 0; i < 10; i++)
+            {
+                if (i < 4)
+                {
+                    // send 4-bit Address
+                    if ((channelNumber >> (3 - i) & 0x01) != 0)
```
krwq, line 131: please use brackets to make it easier to tell if `>>` or `&` has precedence

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,239 @@
...
+                    {
+                        _digital.Write(_address, 1);
```
krwq, line 133: you can simplify this whole `if` statement into:
`_digital.Write(_address, <yourconditionintheifstatement>)`

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,239 @@
...
+                        _digital.Write(_address, 0);
+                    }
+                }
+
+                // read 10-bit data
+                value <<= 1;
+                if (_digital.Read(_dataOut) == PinValue.High)
+                {
+                    value |= 0x01;
+                }
```
krwq, line 146:
```suggestion
                value |= _digital.Read(_dataOut) == PinValue.High;
```

perhaps even just `value |= _digital.Read(_dataOut);` will work too

src/devices/Tlc1543/Tlc1543.cs
krwq: interesting case... I think DelayHelper with no interrupts as parameter should work fine here though

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,239 @@
...
+        /// Reads sensor values into an List
+        /// <remarks>
+        /// <br>Fastest and most efficient to use</br>
+        /// <br>Reads channels in order in the list</br>
+        /// <br>Uses (list size + 1) cycles </br>
+        /// </remarks>
+        /// </summary>
+        /// <param name="channelList">List of channels to read</param>
+        /// <returns>List of 10 bit values corresponding to relative voltage level on specified device channels</returns>
+        public List<int> ReadChannels(List<Channel> channelList)
```
krwq, line 181: I don't really like this helper API, I could see adding something similar if the device supported reading all channels at once but then I suspect I'd ask you to change API into:

`public bool TryReadAllChannels(Span<...> outputs)`

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            values.Add(ReadChannel((int)ChargeChannel));
+            return values;
+        }
+
+        /// <summary>
+        /// Reads all sensor values into an List
+        /// </summary>
+        /// <returns>List of 10 bit values corresponding to relative voltage level on all channels</returns>
+        public List<int> ReadAll()
```
krwq, line 198: considering this is not hardware supported I'll vote to remove this helper. As I mentioned below we should not allocate in APIs like this (also this will be hard to support on nano since it doesn't support `List<T>`)

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,239 @@
...
+    /// </summary>
+    public class Tlc1543 : IDisposable
+    {
+        #region Members
+
+        private readonly int _chipSelect;
+        private readonly int _dataOut;
+        private readonly int _address;
+        private readonly int _inputOutputClock;
+        private readonly int? _endOfConversion;
```
krwq, line 23: this should not be nullable, use -1 as `does not have value`

src/devices/Tlc1543/Tlc1543.cs
SaDiablo: `Operator '|=' cannot be applied to operands of type 'int' and 'PinValue'`
It would need unnecessary cast every loop and in my opinion it would lower the readability of this part.

src/devices/Tlc1543/Tlc1543.cs
SaDiablo: As it is explained in summary this is most efficient (time and power wise) way to read listed channels, so instead of using 2 times as much cycles for reading them, it uses only one more cycle than channels being read.

src/devices/Tlc1543/Tlc1543.cs
SaDiablo: As have been said before let's have it with GPIO support only for now and consider SPI in future. 
And from what I have been testing in last few days I've noticed many problems in using SoftwareSpi as well as SpiDevice with this device (Interrupts while clocking or simply inconsistencies while reading data)

Example of such interrupt:

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// <remarks>
+        /// <br>Refer to table 2 and 3 in TLC1543 documentation for more information</br>
+        /// </remarks>
+        /// </summary>
+        public Channel ChargeChannel
+        {
+            get => _chargeChannel;
+            set
+            {
+                if ((int)value >= 0 & (int)value < 14)
```
krwq, line 41: `&&`

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// <remarks>
+        /// <br>1 cycle used - gets data on second call of this method</br>
+        /// <br>First cycle: Ask for value on the channel</br>
+        /// </remarks>
+        /// </summary>
+        /// <param name="channelNumber">Channel to read</param>
+        /// <returns>10 bit value corresponding to relative voltage level on channel</returns>
+        public int ReadChannel(int channelNumber)
+        {
+            if (channelNumber < 0 & channelNumber > 14)
```
krwq, line 102: you should create helper method like `IsValidChannelNumber`. Here you include 14 as valid while previously you checked for:
`if ((int)value >= 0 & (int)value < 14)` which made 14 invalid

so you should pick which one is true - in order to prevent this kind of errors it's better to have a method for that

src/devices/Tlc1543/Tlc1543.cs
krwq: my concern here is that there is no reason user wouldn't want to implement it themselves. Even faster way to do this is to not allocate the `List<T>`

krwq: Wait, you're not allocating here but you return shared structure. This is always problematic. All APIs which return a list should make sure user can modify it. Now if user writes:

```csharp
List<int> values = foo.ReadChannels(new List<Channel>() { ... });
values.Add(4);
foo.ReadChannels(new List<Channel>() { ... });
```

values will be cleared of content user added. This should not be a case. Ownership of objects should be clear by looking at the usage.

src/devices/Tlc1543/Tlc1543.cs
Ellerbach: If you are using the hardware SPI, it should just work as expected. For the software Spi, it depends of the frequency used and also you should make sure you have the precision timer. If you have precision timer, that should just work as well.
What were the parameters you used for this graph? (hardware or software, frequency, precision timer, etc)

src/devices/Tlc1543/Tlc1543.cs
SaDiablo:
```cs
List<int> values = new List<int>(16);
List<Channel> channelList = new List<Channel>
            {
                Channel.A0,
                Channel.A1,
                Channel.A2,
                Channel.A3,
                Channel.A4
            };
values = foo.ReadChannels(channelList);
values.Add(4);
foo.ReadChannels(channelList);
```
`channelList` in this case is simply list of which channel you want values read. `values` only returns read channel, there is no sharing in there. If you won't overwrite it with another `values = foo.ReadChannels(channelList);` nothing's gonna get changed. So I'm not sure if I understand the problem.

src/devices/Tlc1543/Tlc1543.cs
SaDiablo: SoftwareSpi
```cs
SpiConnectionSettings connectionSettingsSoft = new SpiConnectionSettings(0)
            {
                ClockFrequency = 2_100_000,
                Mode = SpiMode.Mode0,
                DataFlow = DataFlow.MsbFirst,
                ChipSelectLineActiveState = PinValue.Low,
            };
```
I have tested that so many times already and it's so inconsistent between runs that i got discouraged from implementing it further. HardwareSpi works good for what it's worth but that has its share of problems (not able to go for less than 16 clock cycle or using End Of Conversion pin for power/time saving without using both spi and gpio

Ellerbach: I understand that software SPI is not working and that hardware SPI is. So I would then add an SPI interface and mentioning that only hardware SPI is proven to work, not software SPI.
And adding some comments that hardware SPI may not be as efficient as GPIO. Now, some may prefer to use SPI over GPIO for different reasons and may not be too stressed about the few extra clocks and timing.

SaDiablo: Yes, I plan to add that later on. For now i want to have working module with GPIO, then implement rest of the 6 modes that the ADC supports (and proper EOC handling), next limited hardware SPI and some kind of software SPI (I may look into improving software SPI binding at that point too, or it may be improved already at this point in future)

joperezr:
[Triage] Sorry it has taken some time to take another look after your last round of changes, we'll take a look very soon. Looks like your latest changes broke the build though, do you mind having a look? You should be able to repro the issue by just building from the top by calling `build.cmd`/`build.sh`. We also recently enabled nullability on the repo, so it would be good if you could rebase this on top of latest master as that will likely show new warnings regarding nullability. Let us know if you need any help.

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,221 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
krwq, line 3: please remove the third line of the header files (we have converted them globally recently)

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,221 @@
...
+        private readonly bool _shouldDispose;
+        private readonly TimeSpan _conversionTime = new TimeSpan(210);
+        private GpioController _controller;
+        private Channel _chargeChannel = Channel.SelfTest512;
+        private List<int> _values = new List<int>(16);
+
+        /// <summary>
+        /// Channel used between readings to charge up capacitors in ADC
+        /// <remarks>
+        /// <br>Refer to table 2 and 3 in TLC1543 documentation for more information</br>
```
krwq, line 33: please remove openning and closing `<br>`

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,221 @@
...
+            get => _chargeChannel;
+            set
+            {
+                if (IsValidChannel((int)value))
+                {
+                    _chargeChannel = value;
+                }
+                else
+                {
+                    throw new ArgumentOutOfRangeException("ChargeChannel", "Out of range. \nMust be non-negative and less than 14. \nRefer to table 2 and 3 in TLC1543 documentation for more information");
```
krwq, line 47:
```suggestion
                    throw new ArgumentOutOfRangeException(nameof(ChargeChannel), "Value must be 1 and 14. Refer to table 2 and 3 in TLC1543 documentation for more information");
```

this makes it easier to rename as it will also rename the variable

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,221 @@
...
+        /// <returns></returns>
+        protected static bool IsValidChannel(int channelNumber)
+        {
+            if (channelNumber < 0 || channelNumber > 13)
+            {
+                return false;
+            }
+
+            return true;
+        }
```
krwq, line 104:
```suggestion
        protected static bool IsValidChannel(int channelNumber) => channelNumber >= 0 && channelNumber <= 13;
```

although the exception message in the constructor said "positive" which suggests (1-13) - please correct one or another

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,221 @@
...
+                }
+
+                _controller.Write(_inputOutputClock, 1);
+                DelayHelper.DelayMicroseconds(0, false);
+                _controller.Write(_inputOutputClock, 0);
+            }
+
+            _controller.Write(_chipSelect, 1);
+            // Max conversion time (21us) as seen in table on page 10 in TLC1543 documentation
+            DelayHelper.Delay(_conversionTime, false);
```
krwq, line 153: you should not implement SPI, rather you should take SPI in the constructor - for local testing you can use SoftwareSpi which takes GpioController and pin numbers

krwq: most likely for 10 bit data you can simply ReadWrite and pass zeros for ignored parts

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,221 @@
...
+        /// </summary>
+        /// <param name="address">Address pin</param>
+        /// <param name="chipSelect">Chip Select pin</param>
+        /// <param name="dataOut">Data Out pin</param>
+        /// <param name="inputOutputClock">I/O Clock pin</param>
+        /// <param name="endOfConversion">End of Conversion pin</param>
+        /// <param name="controller">The GPIO controller for defined external pins. If not specified, the default controller will be used.</param>
+        /// <param name="pinNumberingScheme">Pin Numbering Scheme</param>
+        /// <param name="shouldDispose">True to dispose the Gpio Controller</param>
+        public Tlc1543(int address, int chipSelect, int dataOut, int inputOutputClock, int endOfConversion = -1, GpioController controller = null, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical, bool shouldDispose = true)
```
krwq, line 67: this should take SpiDevice as input, so constructor should look like following:

```csharp
public Tlc1543(SpiDevice spiDevice)
```

endOfConversionPin - since you're not using it anywhere just do not require it since it doesn't matter anyway.

src/devices/Tlc1543/Tlc1543.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
krwq, line 4: you should now use net5.0

src/devices/Tlc1543/Tlc1543.cs
krwq: note: address will be MOSI pin, dataOut will be MISO pin, inputOutputClock will be CLK

src/devices/Tlc1543/Tlc1543.cs
krwq: BTW you can go for less than 16 cycles, change DataBitLength to 7 and send 2 bytes, that way you will go exactly 14 bits. I'm not in favor of adding just GPIO support because GPIO is already supported with SoftwareSpiDevice

krwq: See code: https://github.com/dotnet/iot/blob/master/src/devices/SoftwareSpi/SoftwareSpi.cs#L160

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,395 @@
...
+        #endregion
+
+        #region Methods
+
+        /// <summary>
+        /// Check if channelNumber is valid
+        /// </summary>
+        /// <param name="channelNumber">Channel Address</param>
+        /// <returns></returns>
+        protected static bool IsValidChannel(int channelNumber) => channelNumber >= 0 && channelNumber <= 13;
```
krwq, line 123: since you have an enum for channel you should use it everywhere where applies, otherwise please make the enum internal

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -23,9 +22,11 @@ public class Tlc1543 : IDisposable
         private readonly int _endOfConversion;
         private readonly bool _shouldDispose;
         private readonly TimeSpan _conversionTime = new TimeSpan(210);
-        private GpioController _controller;
+        private GpioController? _controller;
```
krwq, line 25: controller is never null in the lifetime so it should not be nullable (state after Dispose is not considered lifetime, you should just assign `_controller = null!;` in the Dispose method

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -116,42 +135,196 @@ public int ReadChannel(int channelNumber)
...
+                        ClockImpulse();
+                    }
 
-                _controller.Write(_inputOutputClock, 1);
-                DelayHelper.DelayMicroseconds(0, false);
-                _controller.Write(_inputOutputClock, 0);
+                    _controller.Write(_chipSelect, 1);
+                    break;
+                case Mode.Mode6:
+                    for (int i = 0; i < 16; i++)
```
krwq, line 278: you're repeating exactly the same code in the first 10 cycles and this is re-implementing SPI even when you go to https://www.ti.com/product/TLC1543 it says "Interface type: SPI".

krwq (changes requested):
please use SpiDevice for the implementation instead of re-implementing it.

src/devices/Tlc1543/samples/README.md
```diff
@@ -0,0 +1,76 @@
+# Simple Line finding algorithm with ADC usage
+
+This sample shows how to calculate position of line under matrix of 5 IR sensors with the help of TLC1543 ADC. In this case it's AlphaBot2 Pi from WaveShare with ITR-20001 IR sensors.
+
+## Initialization
+
+```c#
+Tlc1543 adc = new Tlc1543(24, 5, 23, 25);
```
krwq, line 8: please fix README to use changed APIs

krwq (approved):
LGTM, please update README page

joperezr:
@SaDiablo can you please update the README.md with the new API changes? That is the only thing now remaining before we can merge this PR.

src/devices/Tlc1543/Tlc1543.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// Dispose
+        /// </summary>
+        public void Dispose()
+        {
+            if (_controller != null)
+            {
+                if (_shouldDispose)
+                {
+                    _controller.Dispose();
+                    _spiDevice.Dispose();
```
Ellerbach, line 120: shouldDispose is for the GpioController only so the I2cDevice should be disposed

src/devices/Tlc1543/samples/Tlc1543.Samples.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net5.0;netcoreapp3.1</TargetFrameworks>
```
Ellerbach, line 5: net5.0 is enough for the sample

Ellerbach (approved):
Monir adjustments and looks good to me

Issue #1156, opened 2020-08-14
CI is unhealthy


krwq:
CI keeps on failing with random errors:

<details>
<summary>Windows (click to expand)</summary>

```
xUnit.net Console Runner v2.4.1 (32-bit .NET Core 4.6.27617.05)
  Discovering: System.Device.Gpio.Tests
  Discovered:  System.Device.Gpio.Tests
  Starting:    System.Device.Gpio.Tests
    System.Device.Gpio.Tests.ProtocolsTests.I2C_Bme280CanRead [STARTING]
    System.Device.Gpio.Tests.ProtocolsTests.I2C_Bme280CanRead [FAIL]
      System.IO.FileNotFoundException : The system cannot find the file specified.
      
      Slave address was not acknowledged.
      Stack Trace:
           at Windows.Devices.I2c.I2cDevice.Write(Byte[] buffer)
        /_/src/System.Device.Gpio/System/Device/I2c/Devices/Windows10I2cDevice.cs(85,0): at System.Device.I2c.Windows10I2cDevice.WriteByte(Byte value)
        /_/src/devices/Bmxx80/Bmxx80Base.cs(79,0): at Iot.Device.Bmxx80.Bmxx80Base..ctor(Byte deviceId, I2cDevice i2cDevice)
        /_/src/devices/Bmxx80/Bmx280Base.cs(47,0): at Iot.Device.Bmxx80.Bmx280Base..ctor(Byte deviceId, I2cDevice i2cDevice)
        /_/src/devices/Bmxx80/Bme280.cs(35,0): at Iot.Device.Bmxx80.Bme280..ctor(I2cDevice i2cDevice)
        /_/src/System.Device.Gpio.Tests/SetupHelpers.cs(56,0): at System.Device.Gpio.Tests.SetupHelpers.CreateBme280()
        /_/src/System.Device.Gpio.Tests/ProtocolTests.cs(55,0): at System.Device.Gpio.Tests.ProtocolsTests.I2C_Bme280CanRead()
    System.Device.Gpio.Tests.ProtocolsTests.I2C_Bme280CanRead [FINISHED] Time: 0.3997995s
    System.Device.Gpio.Tests.ProtocolsTests.SPI_Mcp3008CanRead [STARTING]
    System.Device.Gpio.Tests.ProtocolsTests.SPI_Mcp3008CanRead [FAIL]
      Assert.InRange() Failure
      Range:  (1018 - 1023)
      Actual: 0
      Stack Trace:
[...]

PR #1195, opened 2020-09-23
Make CI clean again

krwq:
Fixes https://github.com/dotnet/iot/issues/1156

Disable all flaky tests on CI. This PR needs to be re-run couple of times until we get couple of clean runs in the row.

### DIFF
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -270,16 +270,11 @@ namespace System.Device.Gpio.Tests
             }
         }
 
-        [Fact]
+        // [Fact] // https://github.com/dotnet/iot/issues/629
+        // Test seem to be failing with almost all drivers
         public void AddCallbackRemoveAllCallbackTest()
         {
             GpioDriver testDriver = GetTestDriver();
-            // Skipping the test for now when using the SysFsDriver or the RaspberryPi3Driver given that this test is flaky for those drivers.
-            // Issue tracking this problem is https://github.com/dotnet/iot/issues/629
-            if (testDriver is SysFsDriver || testDriver is RaspberryPi3Driver)
-            {
-                return;
-            }
 
             RetryHelper.Execute(() =>
             {


krwq:
`clean_runs++`

pgrawehr:
But still not all green :-(

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -270,16 +270,11 @@ void Callback(object sender, PinValueChangedEventArgs e)
             }
         }
 
-        [Fact]
+        // [Fact] // https://github.com/dotnet/iot/issues/629
+        // Test seem to be failing with almost all drivers
         public void AddCallbackRemoveAllCallbackTest()
```
joperezr, line 275: How did you determine which test to disable? I was going to do some Kusto queries when I got back in order to do some analysis to figure out which tests were the problematic ones and on which platforms did they fail

krwq: it keeps on failing on CI, I've never used Kusto queries, if you know an easy way to determine all failing tests that would be useful

krwq:
I need to figure out what to do with BME test as that one seems like some hardware issue (or perhaps just needs reboot)

pgrawehr:
@krwq: Not sure on the BME280, but for the BME680, if it locks up or enters (for whatever reason) SPI mode, it needs a full power cycle to come back up in I2C mode. I guess the 280 behaves the same.

krwq:
I've requested hard reboot of ddvsorpiw041 (windows machine with issues) since they're now in states while I'm in Europe

krwq:
I think `System.Device.Gpio.Tests.WindowsDriverTests.WaitForEventRisingEdgeTest` hanged but not sure if I'm interpreting output correctly.

joperezr:
To be honest I’d rather not check this in as is. We have very limited coverage on the library and this is disabling a test for all drivers while its not well understood where and when it fails. I can help this weekend with the Kusto queries to gather more info on the failures

krwq:
it fails on every driver except LibGpiod. I find the eventing currently not quite usable because of how unreliable it is but the fact it also gives false alarms on CI is super annoying. IMO we should start throwing PlatformNotSupportedException or similar error if we can't support a feature reliably rather than pretend it works and be unusable for pretty much any scenario.

Issue #1125, opened 2020-07-02
SPI on Raspi: Hardware CS select is not working


pgrawehr:
For each SPI bus (up to 6 on the Raspi4), there are 2 or 3 designated pins for Chip Select (CS)*. The underlying hardware driver normally expects that one of these is also used. For SPI0, pins 7 and 8 serve as CS1 and CS0 respectively, when set to mode ALT0. 

Problem 1: This doesn't work. When setting pin 8 to ALT0, the Chip never answers (to be investigated in detail with a scope, but I don't have mine at hand right now). 

Problem 2: Even though the SpiConnectionSettings structure takes this member and sets it in the driver, this has also no effect, because it doesn't set the pin mode. 

Problem 3: Many (all?) bindings fail to control the CS line at all, relying on the low-level driver, which can't work because of Problem 2. 

We should:
- Either make sure Problem 1 gets solved and/or 
- make clear that we're always using software CS control (by the binding or the SpiDevice?)

CC: @Ellerbach, as you've had some experience with SPI, I think

(*) See Alternate Pin Assignments in the BCM2711 ARM Peripherals Guide.

krwq:
@pgrawehr can you include exact repro steps? (what to put in config.txt; example device you are using and sample code)

Ellerbach:
I'll try but so far I did not experience this issue. I had to use software CS selection sometimes because of the way the end device was working. And yes, the config.txt and anything that will help to recreate the same hardware setup and software setup is needed in this case. Version of the OS you are running as well to recreate the full conditions.

pgrawehr:
I've got example code to reproduce the behavior, but it needs a bit of cleanup. I'll provide more information later today.

pgrawehr:
**How to reproduce the problem**
To reproduce, check out my branch https://github.com/pgrawehr/iot/tree/BoardApproach3 (PR #1128 ), then run the sample in devices/Board/samples

`boot/config.txt` only contains the relevant line `dtparam=spi=on`. 

After booting, we have the following pin configuration (relevant for this case are BCM pins 9, 10, 11 and 8):

```
pi@raspberrypi:~ $ gpio readall
 +-----+-----+---------+------+---+---Pi 4B--+---+------+---------+-----+-----+
 | BCM | wPi |   Name  | Mode | V | Physical | V | Mode | Name    | wPi | BCM |
 +-----+-----+---------+------+---+----++----+---+------+---------+-----+-----+
[...]

PR #1197, opened 2020-09-24
Adding SPI specific documentation for Raspberry Pi

Ellerbach:
Adding SPI specific documentation for Raspberry Pi
- Related to issue #1125
Fixing a typo in the PWM documentation

### DIFF
--- a/Documentation/raspi-pwm.md
+++ b/Documentation/raspi-pwm.md
@@ -20,7 +20,7 @@ namespace TestTest
             var pwm = PwmChannel.Create(0, 0, 400, 0.5);
             pwm.Start();
             Console.ReadKey();
-        }        
+        }
     }
 }
 ```
@@ -153,7 +153,7 @@ sudo ./yourapplication
 
 ### Adding your user to the right permission group
 
-If you're running, or just upgrated to a version published after Auguste 2020, this should be already done. 
+If you're running, or just upgraded to a version published after Auguste 2020, this should be already done. 
 You will have to create a [specific group in udev](https://raspberrypi.stackexchange.com/questions/66890/accessing-pwm-module-without-root-permissions).
 
 ```bash
--- a/Documentation/raspi-pwm.md
+++ b/Documentation/raspi-pwm.md
@@ -153,7 +153,7 @@ sudo ./yourapplication
 
 ### Adding your user to the right permission group
 
-If you're running, or just upgraded to a version published after Auguste 2020, this should be already done. 
+If you're running, or just upgraded to a version published after August 2020, this should be already done. 
 You will have to create a [specific group in udev](https://raspberrypi.stackexchange.com/questions/66890/accessing-pwm-module-without-root-permissions).
 
 ```bash
--- a/Documentation/raspi-spi.md
+++ b/Documentation/raspi-spi.md
@@ -6,7 +6,7 @@ This page will explain how to setup any SPI. Please refer to the [Raspberry Pi d
 
 ## Basic Hardware SPI usage
 
-The most simplest code you can build to use SPI is the following:
+The most simple code you can build to use SPI is the following:
 
 ```csharp
 using System;
--- a/Documentation/raspi-spi.md
+++ b/Documentation/raspi-spi.md
@@ -61,7 +61,7 @@ Then reboot:
 sudo reboot
 ```
 
-This will enable SP0 where those are the pins which will be seleted, only Software Chip Select is activated with the default pins:
+This will enable SPI0 where those are the pins which will be selected, only Software Chip Select is activated with the default pins:
 
 | SPI Function | Header Pin | GPIO # | Pin Name |
 | --- | --- | --- | --- |
--- a/Documentation/raspi-spi.md
+++ b/Documentation/raspi-spi.md
@@ -1,0 +1,135 @@
+# Enabling SPI on Raspberry Pi
+
+In most of the cases, it is easy and straight forward to enable SPI for your Raspberry Pi. The basic case can be [found here](https://www.raspberrypi-spy.co.uk/2014/08/enabling-the-spi-interface-on-the-raspberry-pi/).
+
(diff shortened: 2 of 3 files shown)


Documentation/raspi-spi.md
```diff
@@ -0,0 +1,135 @@
...
+
+Save the file with `ctrl + x` then `Y` then `enter`
+
+Then reboot:
+
+```bash
+sudo reboot
+```
+
+This will enable SP0 where those are the pins which will be seleted, only Software Chip Select is activated with the default pins:
```
pgrawehr, line 64: Thanks for adding this. Cost me a few hours to find out why hardware CS wasn't working...

Maybe add a sentence about the difference between hw and software cs? This isn't clear from this text.

Documentation/raspi-spi.md
```diff
@@ -0,0 +1,135 @@
...
+
+The following dtoverlay definition can be [found here](https://github.com/raspberrypi/firmware/blob/7b99da75f55a5ad7d572ec4ebe4e8f9573deaee7/boot/overlays/README#L2437).
+
+| SPI # | Chip Select # | Header Pin | Default GPIO | Pin Name | 
+| --- | --- | --- | --- | --- | --- | --- |
+| SPI0 | CE0 | 24 | GPIO08 | SPI0_CE0_N |
+| SPI0 | CE1 | 26 | GPIO07 | SPI0_CE1_N |
+
+
+If you want to change the default pins for Chip Select 0 to the GPIO pin 27 (hardware 13), and let's say GPIO pin 22 (hardware 15) for Chip Select 1, just add this line:
```
pgrawehr, line 90: Are you sure you can use any pin for hardware CS? The "alternate pin mode" table suggests that these are fixed.

Ellerbach: The table is the default configuration. Advance configuration is what I describe. So seems it is not super clear. How would you rephrase then?

Documentation/raspi-spi.md
Ellerbach: Seems that now there is no difference, once set it is by default activated. The hardware dtoverlay is depreciated.

krwq, line 64:
```suggestion
This will enable SPI0 where those are the pins which will be selevted, only Software Chip Select is activated with the default pins:
```

krwq (approved):

Documentation/raspi-spi.md
pgrawehr: I mean "software" as in "do it yourself". I've seen quite a few examples (don't remember where, though), where the CS pin is manually controlled directly in the application.

pgrawehr: "selected" is still spelled incorrectly ;-)

pgrawehr: Actually, just using `dtparam=spi=on` will not activate _any_ CS pin, you need to do it yourself (by manyually controlling the pin that is connected to the CS input line of the attached chip). My finding with adding `dtoverlay=spi0-hw-cs` will change the ALT setting to the one described in the manual and use that pin. 
So with your suggestion, we have actually three different possibilities for the CS pin: None, hardware and "driver controlled gpio". 
Can you check what "gpio readall" reports for i.e. pin 27 when your line `dtoverlay=spi0-2cs,cs0_pin=27,cs1_pin=22` is used?

Documentation/raspi-spi.md
pgrawehr: See above - I think we're talking about three different ways here. I'll have to think about how to describe this in an understandable and clear way.

Documentation/raspi-spi.md
Ellerbach: My understanding is that now you choose the pins you want, reason why the `dtoverlay)spi0-hw-cs` is deprecated. Also `gpio` is not available anymore. See the note from the WiringPi author: http://wiringpi.com/wiringpi-deprecated/
It seems, reading the doc, that `dtparam=spio=on`will activate both default pins as Chip Select. Reason why I linked to the latest doc. Now, for previous versions, it seems it was not the case.
If you have any working alternative for `gpio readall`, happy to try!

pgrawehr: Well, at least the spi0-hw-cs works... I'll need to redo my SPL tests when I find time - last time I tried (when I wrote that issue) `dtparam=spi=on` would definitelly not activate any CS pins. 

The latest version of WiringPi, 2.52, does work on the Pi4 (and is very helpful in analyzing pin multiplexing behavior). However you need to download it from the website, it's not available via apt. 

With #1128 in place, we would be able to extend the `DeviceApiTester` tool to completelly replace `gpio`.

Ellerbach: I fully support a board abstract. We discussed it with @krwq. With all the request to support various baords, we need anyway a HAL/PAL approach.
I can try to clone the #1128 and see what I get!

Documentation/raspi-pwm.md
```diff
@@ -153,7 +153,7 @@ sudo ./yourapplication
 
 ### Adding your user to the right permission group
 
-If you're running, or just upgrated to a version published after Auguste 2020, this should be already done. 
+If you're running, or just upgraded to a version published after Auguste 2020, this should be already done.
```
krwq, line 156:
```suggestion
If you're running, or just upgraded to a version published after August 2020, this should be already done. 
```

Documentation/raspi-spi.md
```diff
@@ -0,0 +1,135 @@
+# Enabling SPI on Raspberry Pi
+
+In most of the cases, it is easy and straight forward to enable SPI for your Raspberry Pi. The basic case can be [found here](https://www.raspberrypi-spy.co.uk/2014/08/enabling-the-spi-interface-on-the-raspberry-pi/).
+
+This page will explain how to setup any SPI. Please refer to the [Raspberry Pi documentation](https://www.raspberrypi.org/documentation/hardware/raspberrypi/spi/) to understand the different SPI available. You should be aware as well that for Raspberry Pi4, some of the configurations are different than for the other version especially for SPI3, 4 and 5. 
+
+## Basic Hardware SPI usage
+
+The most simplest code you can build to use SPI is the following:
```
krwq, line 9:
```suggestion
The most simple code you can build to use SPI is the following:
```

krwq (approved):
couple of typos but otherwise looks good to me, thank you!

krwq (approved):

Issue #1180, opened 2020-09-10
Hcsr04 should use UnitsNet 


Ellerbach:
**Describe the bug**

```Distance``` property returns a double which is actually coded as "cm". This should definitely use UnitsNet
https://github.com/dotnet/iot/blob/3c4034781d158b26be0bc4866c34192d86bc4ef1/src/devices/Hcsr04/Hcsr04.cs#L28

**Expected behavior**

Regarding the new rules for units, this should use the UnitsNet corresponding to meters

**Versions used**

Latest from the repo

krwq:
we should fix it in this release

Ellerbach:
This is now resolved by the merged PR

PR #1198, opened 2020-09-25
Fixing Hcsr04 issues

### LINKED ISSUE #1179: Hcsr04 TrickCount bug so waiting conditions not respected
**Describe the bug**

Miss usage of ```Environment.TickCount``` which returns a tick count and not milliseconds.
See here and all the following line from this function:
https://github.com/dotnet/iot/blob/3c4034781d158b26be0bc4866c34192d86bc4ef1/src/devices/Hcsr04/Hcsr04.cs#L93
Environment.TickCount:
https://docs.microsoft.com/en-us/dotnet/api/system.environment.tickcount?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(System.Environment.TickCount);k(DevLang-csharp)%26rd%3Dtrue&view=netcore-3.1

**Steps to reproduce**

So the code is actually not working depending on the platform used

**Expected behavior**

Should use ```(DateTime.Now.Ticks / TimeSpan.TicksPerMillisecond)``` or equivalent instead of ```Environment.TickCount```

**Versions used**

Version from repo with raw code

Ellerbach:
- Fixes #1179 and #1180
- Using DelayHelpers

### DIFF
--- a/src/devices/Hcsr04/Hcsr04.cs
+++ b/src/devices/Hcsr04/Hcsr04.cs
@@ -6,6 +6,8 @@ using System;
 using System.Diagnostics;
 using System.Device.Gpio;
 using System.Threading;
+using System.Device;
+using UnitsNet;
 
 namespace Iot.Device.Hcsr04
 {
@@ -20,12 +22,12 @@ namespace Iot.Device.Hcsr04
         private bool _shouldDispose;
         private Stopwatch _timer = new Stopwatch();
 
-        private int _lastMeasurment = 0;
+        private long _lastMeasurment = 0;
 
         /// <summary>
         /// Gets the current distance in cm.
         /// </summary>
-        public double Distance => GetDistance();
+        public Length Distance => GetDistance();
 
         /// <summary>
         /// Creates a new instance of the HC-SCR04 sonar.
@@ -66,7 +68,7 @@ namespace Iot.Device.Hcsr04
         /// <summary>
         /// Gets the current distance in cm.
         /// </summary>
-        private double GetDistance()
+        private Length GetDistance()
         {
             // Retry at most 10 times.
             // Try method will fail when context switch occurs in the wrong moment
@@ -77,7 +79,7 @@ namespace Iot.Device.Hcsr04
             {
                 if (TryGetDistance(out double result))
                 {
-                    return result;
+                    return Length.FromCentimeters(result);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -66,7 +68,7 @@ public Hcsr04(int triggerPin, int echoPin, PinNumberingScheme pinNumberingScheme
         /// <summary>
         /// Gets the current distance in cm.
```
krwq, line 69: unit should be removed from description since you change return type

src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -77,7 +79,7 @@ private double GetDistance()
             {
                 if (TryGetDistance(out double result))
```
krwq, line 80: nit: I'd vote for TryGetDistance to return Length directly

src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -90,39 +92,39 @@ private bool TryGetDistance(out double result)
...
-                Thread.Sleep(TimeSpan.FromMilliseconds(Environment.TickCount - _lastMeasurment));
+                Thread.Sleep(TimeSpan.FromTicks(DateTime.Now.Ticks - _lastMeasurment));
             }
 
+            _lastMeasurment = DateTime.Now.Ticks;
+
             // Trigger input for 10uS to start ranging
             _controller.Write(_trigger, PinValue.High);
-            Thread.Sleep(TimeSpan.FromMilliseconds(0.01));
+            DelayHelper.DelayMicroseconds(10, true);
```
krwq, line 109: 👍

src/devices/Hcsr04/Hcsr04.cs
```diff
@@ -90,39 +92,39 @@ private bool TryGetDistance(out double result)
...
             _controller.Write(_trigger, PinValue.Low);
 
             // Wait until the echo pin is HIGH (that marks the beginning of the pulse length we want to measure)
             while (_controller.Read(_echo) == PinValue.Low)
             {
-                if (Environment.TickCount - hangTicks > 0)
+                if (DateTime.Now.Ticks - hangTicks > 0)
```
krwq, line 115: perhaps we should add GetTicks (with inlining) to DelayHelpers in case we need to do this more often in the future and also have a way to convert ticks to seconds. The original code is fine here but seems it's not supported on nano. Also note that DateTime.Now will likely be broken if someone changes system time while app is running

src/devices/Hcsr04/samples/README.md
```diff
@@ -12,16 +12,21 @@
...
 ## Code
 ```C#
 using (var sonar = new Hcsr04(4, 17))
 {
-    while (true)
+    try
     {
-        Console.WriteLine($"Distance: {sonar.Distance} cm");
-        Thread.Sleep(1000);
+        Console.WriteLine($"Distance: {sonar.Distance.Centimeters} cm");
```
krwq, line 23: we might want to make TryGetDistance public just to avoid this anti-pattern of catching

src/devices/Hcsr04/Hcsr04.cs
Ellerbach: done

src/devices/Hcsr04/Hcsr04.cs
Ellerbach: I have a commit delay :-) That was fixed as well! forgot to save the file originally!

src/devices/Hcsr04/Hcsr04.cs
Ellerbach: If someone will change the time, it's fine, it will just break 1 measure. Not a big deal :-) None of the sensors are perfect. They all can return wrong values time to time. This sensor has multiple retry because of the quality of the measurement anyway.
And same, I already changed the element but didn't save before pushing.

src/devices/Hcsr04/samples/README.md
Ellerbach: yes, made the Try public and adjusted the sample. Will leave the GetDistance() and Distance public anyway

krwq (approved):

src/devices/Hcsr04/samples/README.md
```diff
@@ -12,16 +12,22 @@
...
-The fritz diagram above depicts how you should wire your RPi in order to run the program, the resistance for R1 is 1kOhm. (Optional)
+The fritz diagram above depicts how you should wire your RPi in order to run the program, the resistance for R1 is 1kOhm. (Optional).
 
 ## Code
 ```C#
 using (var sonar = new Hcsr04(4, 17))
 {
-    while (true)
+    var res = sonar.TryGetDistance(out Length distance);
+    if (res)
```
krwq, line 22:
```suggestion
    if (sonar.TryGetDistance(out Length distance))
```

krwq: (optional)

krwq:
LGTM, thanks @Ellerbach!

Issue #1177, opened 2020-09-09
GPIO Raspberry PI Compute Module 3


bPoller2810:
I found the https://github.com/dotnet/iot/issues/762 but its pretty silent in there. 
Also the https://github.com/dotnet/iot/pull/1034 is kinda dead.

Is there someone here who knows about the progress we can expect there?

in short we want to just be able to use the following 3 GPIO:
- 30 Read
- 37 Write
- 31 Write

pgrawehr:
Unfortunately, the responsible people for final review of the open PRs are very busy recently. I too hope this gets better soon.

bPoller2810:
Well the whole problem was reported 11 month ago (ish) and seemed fixed 5 month ago. 

maybe i need too look if there is other projects out there, if it issnt certain that problems will be fixed

krwq:
I believe this should be fixed with @pgrawehr's PR above

krwq:
(and sorry for not being able to take a look at the PRs sooner, I believe it should improve now)

PR #1034, opened 2020-04-02
Cm3 Driver

JTrotta:
`Fixes #762` 
- Adding Board class, to better  identify the board.

PR #1200, opened 2020-09-27
Updated board identification, CM3 support

### LINKED ISSUE #1145: GpioController is not correctly identifying Raspberry Pi after latest Raspbian kernel updates
**Describe the bug**

After running apt-get upgrade on a Raspberry Pi 4B, I couldn't set pinmodes to pullup or pulldown anymore. Instead, I was getting an error: Unhandled exception. System.InvalidOperationException: The pin does not support the selected mode.

After checking the System.Device.Gpio project I found that GpioController is trying to know if it's running on a Raspberry Pi via /proc/cpuinfo. There's a field in there called Hardware and it returns the processor type. Apparently in older Raspbian kernels it was returning the wrong value, since even on a Raspberry Pi 4 it would return "BCM2835". This is the value that GpioController is comparing against.

Since the kernel update however it seems to be returning the correct value, which for a Raspberry Pi 4 is "BCM2711". 

In fact, BCM2835 is only correct for the first Raspberry Pi, the other models are listed here: https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2711/README.md

Current workaround is to manually specify a GpioDriver and pass it to the GpioController constructor:

```csharp
GpioDriver driver = new RaspberryPi3Driver();
Controller = new GpioController(PinNumberingScheme.Logical, driver);
```

**Steps to reproduce**

Update Raspbian on Raspberry Pi 4b to latest kernel via:

sudo apt-get update
sudo apt-get upgrade

and reboot the Pi. Then the following code will fail:

```csharp
Controller = new GpioController();
Controller.OpenPin(26, PinMode.InputPullUp);
```

**Expected behavior**

The pin would be opened and the internal pullup resistor turned on.

**Actual behavior**

Exception is raised: System.InvalidOperationException: The pin does not support the selected mode.

**Versions used**
Target framework: netcoreapp3.1
System.Device.Gpio version: 1.1.0-prerelease.20276.1

Running self-contained on Raspberry Pi 4b, with kernel version 5.4.51-v7l+ (before the update it was 4.19.118-v7l+)

pgrawehr:
Fixes #1145, #1177. Replaces #1034 

Using the board model for identification, thanks @JTrotta. Supported Pi models for now: 3B, 3B+ and 4.

Also adds CM3 support (only for linux).


src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -381,29 +381,12 @@ private static GpioDriver GetBestDriverForBoard()
...
         {
-            string[] cpuInfoLines = File.ReadAllLines(CpuInfoPath);
-            Regex regex = new Regex(@"Hardware\s*:\s*(.*)");
-            foreach (string cpuInfoLine in cpuInfoLines)
+            try
+            {
+                return new RaspberryPi3Driver();
+            }
+            catch (PlatformNotSupportedException)
             {
```
Ellerbach, line 389: So you're catching but don't do anything here? Maybe just move the line `return UnixDriver.Create();` here

Ellerbach (commented):
Looks good! I like it. Would like to understand how to make it more generic to reuse on other boards. If not, then I'm good.

src/System.Device.Gpio/System/Device/Gpio/BoardIdentification .cs
```diff
@@ -0,0 +1,337 @@
+﻿using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+
+namespace System.Device.Gpio
+{
+    /// <summary>
+    /// Identification of Raspberry Pi board models
+    /// </summary>
+    internal class BoardIdentification
```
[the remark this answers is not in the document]
pgrawehr: Yes, I should change the name to become raspi-specific. 

This is on purpose internal for now, as krwq insists that we currently do not add new public apis to the core dll. Once we agree on the way forward with board identification and hardware abstraction layers, this will likely change again.

src/System.Device.Gpio/System/Device/Gpio/BoardIdentification .cs
```diff
@@ -0,0 +1,337 @@
...
+                case Model.RaspberryPiAPlus:
+                case Model.RaspberryPiBRev1:
+                case Model.RaspberryPiBRev2:
+                case Model.RaspberryPiBPlus:
+                case Model.RaspberryPiComputeModule:
+                case Model.RaspberryPiZero:
+                case Model.RaspberryPiZeroW:
+                    return Processor.Bcm2708;
+                case Model.RaspberryPi2B:
+                // TBC: B3(+) should be a BCM2710 processor ...
```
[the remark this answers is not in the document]
pgrawehr: I don't know, but the method is currently unused anyway. Not sure whether I should remove it for now, since it might still be useful later. Should probably rather use the Processor field now, since with the current kernel, that should work correctly.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -32,12 +32,27 @@ public RaspberryPi3Driver()
         {
             if (Environment.OSVersion.Platform == PlatformID.Unix)
             {
-                _internalDriver = new RaspberryPi3LinuxDriver();
-                RaspberryPi3LinuxDriver linuxDriver = _internalDriver as RaspberryPi3LinuxDriver;
+                RaspberryBoardIdentification identification = RaspberryBoardIdentification.LoadBoard();
+                RaspberryPi3LinuxDriver linuxDriver = null;
```
krwq, line 36: I don't believe the initialization with null is needed since you cover all paths

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -381,32 +381,14 @@ private static GpioDriver GetBestDriverForBoard()
...
-                        // Commenting out as HummingBoard driver is not implemented yet, will be added back after implementation
-                        // https://github.com/dotnet/iot/issues/76
-                        // if (match.Groups[1].Value == HummingBoardHardware)
-                        // {
-                        //     return new HummingBoardDriver();
-                        // }
-                        return UnixDriver.Create();
-                    }
-                }
+                return new RaspberryPi3Driver();
```
krwq, line 386: can you create some helper which you could use here and in the constructor to avoid catching here?

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
```diff
@@ -0,0 +1,337 @@
+﻿using System.Collections.Generic;
```
krwq, line 1: copyright header

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
```diff
@@ -0,0 +1,337 @@
...
+    {
+        /// <summary>
+        /// The Raspberry Pi model.
+        /// </summary>
+        public enum Model
+        {
+            /// <summary>
+            /// Unknown model.
+            /// </summary>
+            Unknown,
```
krwq, line 20: Currently this is not an issue but I suspect in the future this will become a bit problematic as adding new entry which is a variant of existing one will require updating all places which use the branching. This is fine for now since this is not an issue yet and code is internal but we should think of how to express these kinds of relationship.

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
```diff
@@ -0,0 +1,337 @@
...
+        /// Gets the processor name.
+        /// </summary>
+        /// <value>
+        /// The name of the processor.
+        /// </value>
+        public string ProcessorName
+        {
+            get
+            {
+                _settings.TryGetValue("Hardware", out string hardware);
```
krwq, line 206: instead of strings could you just make those `public string Foo { get; private set; }` and fill the properties on construction directly rather than going through string?

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
```diff
@@ -0,0 +1,337 @@
+﻿using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+
+namespace System.Device.Gpio
+{
+    /// <summary>
+    /// Identification of Raspberry Pi board models
+    /// </summary>
+    internal class RaspberryBoardIdentification
```
krwq, line 10: I'd rename this to RasberryPiBoardInfo

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
```diff
@@ -0,0 +1,337 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Detect the board CPU information from /proc/cpuinfo
+        /// </summary>
+        /// <returns>
+        /// The <see cref="RaspberryBoardIdentification"/>.
+        /// </returns>
+        internal static RaspberryBoardIdentification LoadBoard()
```
krwq, line 297: If you rename the class the more appropiate name for this would be `LoadBoardInfo`

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
```diff
@@ -0,0 +1,337 @@
...
+                    if (!string.IsNullOrWhiteSpace(line) && separator > 0)
+                    {
+                        var key = line.Substring(0, separator).Trim();
+                        var val = line.Substring(separator + 1).Trim();
+                        if (string.Equals(key, "processor", StringComparison.InvariantCultureIgnoreCase))
+                        {
+                            suffix = "." + val;
+                        }
+
+                        settings.Add(key + suffix, val);
```
krwq, line 320: I'd rather have us parse the names here - perhaps create a list of locals up front, initialize them with null/-1 or some "not initialized" value and assign them directly

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
```diff
@@ -0,0 +1,337 @@
...
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Get board model from firmware revision
+        /// See http://www.raspberrypi-spy.co.uk/2012/09/checking-your-raspberry-pi-board-version/ for information.
+        /// </summary>
+        /// <returns></returns>
+        internal Model GetBoardModel()
```
krwq, line 137: I'd rather have us do this once and this be a property

krwq:
Thanks @pgrawehr, overall this looks good to me, couple of cosmetic changes (for readability)

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
pgrawehr: Yea, I would have done the same, but for now I just did minimal changes to the existing PR. I can update it, though.

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
pgrawehr: Sure (it's not so many we need for now, anyway). But do note that these properties are initially strings (we're parsing the /proc/cpuinfo file, which is basically a string=string property list).

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
pgrawehr: Yea, true. Outside of the actual identification, I think the best way will be to use inheritance for the class features, (similar to the GpioDriver tree, but on a more generic level).

src/System.Device.Gpio/System/Device/Gpio/BoardIdentification .cs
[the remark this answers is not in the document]
krwq: I'd vote to remove it if it's not used. You can create an issue which this would solve and paste the snippet so it doesn't get lost

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
krwq: this shouldn't be necessary, I'd recommend just to construct a tree/graph (static field with `Dictionary<Model, Model>` with all related entries) which would describe how the things are related and then add a method to have a nicer way to get the relationship (i.e. you can describe who is parent of what, i.e. RaspberryPi3 is parent of RaspberryPi3A, RaspberryPi3B and then you can create method which would tell you if any of the values is RaspberryPi3), you can define even multilevel relationships. I'd start with brainstorming first rather than fixing it right now

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
krwq: I'd still go with parsing this, we need only specific information and not generic knowledge database

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
krwq: I'd prefer if we did it

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: Yea, done. Is a bit messy as well, since the ctor of the `RaspberryPi3Driver` is public and should work even when called directly (in which case an exception is ok).

pgrawehr: PS: I'm more and more in favor of getting the board identification stuff correctly done and dropping outside access to the drivers. We're doing more and more workarounds to keep the API static.

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardIdentification.cs
pgrawehr: Now parsing in ctor and setting properties.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
Ellerbach: I agree with you @pgrawehr. I'm sure very soon a discussion will pop up to address the Gpio API

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardInfo.cs
```diff
@@ -0,0 +1,288 @@
...
+            /// Pi 4 all versions.
+            /// </summary>
+            RaspberryPi4,
+        }
+
+        #region Fields
+
+        private readonly Dictionary<string, string> _settings;
+
+        private RaspberryBoardInfo(Dictionary<string, string> settings)
```
krwq, line 96: what I had in mind is to not have dictionary at all but this is fine I guess since this is something we should be doing once per execution

krwq (approved):
Thanks @pgrawehr!

krwq:
@Ellerbach do you also want to take a look?

krwq:
@Ellerbach feel free to add any post mortem review (create issue if anything blocking) 😄

krwq:
Thanks @pgrawehr!

PR #1204, opened 2020-10-04
Add SparkFun Qwiic Button binding

Esbob:
Device: https://www.sparkfun.com/products/15932


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1204) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1204) before we can accept your contribution.<br/><br/>:x: Esbob [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1204)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1204) it.</sub>

src/devices/QwiicButton/samples/FindI2cDevices.cs
```diff
@@ -0,0 +1,61 @@
...
+                try
+                {
+                    using (var button = new QwiicButton(i2cBusId, (byte)address))
+                    {
+                        // Try to read device ID - fails if no device on address
+                        button.GetDeviceId();
+                    }
+
+                    devicesFoundCount++;
+                    Console.WriteLine($"Found I2C device on address 0x{BitConverter.ToString(new[] { (byte)address })} ({address})");
```
pgrawehr, line 43: Just use "{address:X}"

pgrawehr: Also, be more precise in the text: You're only searching Qwiic buttons, there might be other devices on the bus.

pgrawehr (commented):
Just a bunch of initial comments, but looks quite good to me.
Nice set of examples!

Esbob:
Thank you so much for your excellent comments,  @pgrawehr ! They revealed to me some other inconsistencies, so I've made some refactorings and introduced your suggestions.

src/devices/QwiicButton/FlagsHelper.cs
```diff
@@ -0,0 +1,52 @@
+﻿//// Licensed to the .NET Foundation under one or more agreements.
+//// The .NET Foundation licenses this file to you under the MIT license.
+//// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.QwiicButton
+{
+    internal static class FlagsHelper
```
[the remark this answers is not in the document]
Esbob: Done, along with I2cRegisterAccess that I've made more generic. I believe others need to access registers...

src/devices/QwiicButton/FlagsHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        public static bool IsSet<T>(T flags, T flag)
```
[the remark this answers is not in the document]
Esbob: Correct - I've changed the implementation to use Enum.HasFlag, but I kept the IsSet method in the FlagsHelper to keep a consistent set of utilities in the class.

src/devices/QwiicButton/FlagsHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+            where T : struct
+        {
+            // Cast to object necessary due to C#'s restriction against a where T : Enum constraint
+            int flagsValue = (int)(object)flags;
+            int flagValue = (int)(object)flag;
+
+            return (flagsValue & flagValue) != 0;
+        }
+
+        public static T SetValue<T>(T flags, T flag, bool enabled)
```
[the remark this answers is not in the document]
Esbob: Done

src/devices/QwiicButton/I2cBusAccess.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Device.I2c;
+using System.Runtime.InteropServices;
+
+namespace Iot.Device.QwiicButton
+{
+    /// <summary>
+    /// Encapsulates low-level access to read from and write to the provided
+    /// Qwiic Button device.
+    /// </summary>
+    internal class I2cBusAccess : IDisposable
```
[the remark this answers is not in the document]
Esbob: Done

src/devices/QwiicButton/QueueStatusBitField.cs
```diff
@@ -0,0 +1,56 @@
+﻿//// Licensed to the .NET Foundation under one or more agreements.
+//// The .NET Foundation licenses this file to you under the MIT license.
+//// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.QwiicButton
+{
+    internal class QueueStatusBitField
```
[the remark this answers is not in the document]
Esbob: Done

src/devices/QwiicButton/QwiicButton.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            var success = _i2cBus.WriteSingleRegister(Register.I2cAddress, address);
+            if (success)
+            {
+                I2cAddress = address;
+                return true;
+            }
+
+            Console.WriteLine("Could not set I2C address");
```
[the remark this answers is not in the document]
Esbob: Good point - fixed.

src/devices/QwiicButton/QwiicButton.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </summary>
+        public bool SetI2cAddress(byte address)
+        {
+            if (address < 0x08 || address > 0x77)
+            {
+                Console.WriteLine("I2C input address is out of range");
+                return false;
+            }
+
+            var success = _i2cBus.WriteSingleRegister(Register.I2cAddress, address);
```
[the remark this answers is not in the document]
Esbob: Fixed

src/devices/QwiicButton/QwiicButton.cs
```diff
@@ -0,0 +1,102 @@
...
+using System;
+using System.Device.I2c;
+
+namespace Iot.Device.QwiicButton
+{
+    /// <summary>
+    /// SparkFun Qwiic Button is an I2C based button with a built-in LED.
+    /// Supported hardware version: 1.0.0
+    /// </summary>
+    public partial class QwiicButton : IDisposable
```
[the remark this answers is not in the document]
Esbob: Done

src/devices/QwiicButton/samples/FindI2cDevices.cs
Esbob: Hex notation: fixed.
As for the text: The way I detect buttons on the bus is by trying to read the register at address 0x00 and unfortunately other I2C devices could potentially contain something on that address as well, which means I cannot be completely sure that I found a button... That's why I use the words "I2C device".

src/devices/Common/Iot/Device/Common/FlagsHelper.cs
```diff
@@ -0,0 +1,66 @@
+﻿//// Licensed to the .NET Foundation under one or more agreements.
```
krwq, line 1: I think the comment has too many `/`

src/devices/Common/Iot/Device/Common/FlagsHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// </summary>
+    public static class FlagsHelper
+    {
+        /// <summary>
+        /// Determines whether one or more bit fields are set in the provided <paramref name="flags"/> bit field.
+        /// </summary>
+        /// <typeparam name="T"><see cref="Enum"/> decorated with the <see cref="FlagsAttribute"/>.</typeparam>
+        /// <param name="flags">Bit field to use as basis for the lookup.</param>
+        /// <param name="flag">One or more bit fields to lookup.</param>
+        public static bool IsSet<T>(T flags, T flag)
```
krwq, line 22: why not use HasFlags directly? It's longer this way and more instructions? (`foo.HasFlags(...)` vs `FlagsHelper.IsSet(foo, ...)`)

src/devices/Common/Iot/Device/Common/FlagsHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+            else
+            {
+                Unset(ref flags, flag);
+            }
+        }
+
+        private static void Set<T>(ref T flags, T flag)
+            where T : struct
+        {
+            int flagsValue = (int)(object)flags;
```
krwq, line 51: this will only work for `int` based enums, same comment below

src/devices/Common/Iot/Device/Common/FlagsHelper.cs
```diff
@@ -0,0 +1,66 @@
+﻿//// Licensed to the .NET Foundation under one or more agreements.
+//// The .NET Foundation licenses this file to you under the MIT license.
+//// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.Common
```
krwq, line 7: IMO we should not add helpers like this. `&=~` and `|=` are fine to use rather than this

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
```diff
@@ -0,0 +1,120 @@
+﻿//// Licensed to the .NET Foundation under one or more agreements.
```
krwq, line 1: too many `/`

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+               (byte)(object)register,
+               lower,
+               upper
+            }));
+        }
+
+        private static ReadOnlySpan<byte> ToReadOnlySpan(TRegisterMap registerValue)
+        {
+            return new ReadOnlySpan<byte>(new[] { (byte)(object)registerValue });
```
krwq, line 110: this is missing the purpose of Span. You use Span in order to avoid allocation, in here you neither have the benefit of the array nor span

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
```diff
@@ -0,0 +1,120 @@
...
+    /// enum RegisterMap : byte
+    /// {
+    ///     Register1 = 0x00,
+    ///     Register2 = 0x01,
+    ///     Register3 = 0x03
+    /// }
+    /// </code>
+    /// </example>
+    /// </summary>
+    public sealed class I2cRegisterAccess<TRegisterMap> : IDisposable
```
krwq, line 26: this would be fine if all registers were a single byte. Some devices may have larger registers

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
```diff
@@ -0,0 +1,120 @@
...
+            return MemoryMarshal.Read<ushort>(readBuffer);
+        }
+
+        /// <summary>
+        /// Reads a 32-bit integer from the provided register address.
+        /// </summary>
+        public uint ReadQuadRegister(TRegisterMap register)
+        {
+            Span<byte> readBuffer = new byte[4];
+            _device.WriteRead(ToReadOnlySpan(register), readBuffer);
```
krwq, line 76: I'm not a fan of this helper class due to couple of things:
- allocations: we have span in order to avoid allocations and it's not being used here this way, we still allocate
- it assumes how every device works - while most of the devices work similarly it's not true for all devices
- if this class is in fact making code better it should be done globally for all devices not just for this single one and since refactoring and adding new binding is different concern I prefer this be done as separate PR (if you plan to do global refactoring first send PR with single device and after you get agreement this is in fact making code cleaner then update remaining devices)

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
```diff
@@ -0,0 +1,120 @@
...
+            if (readBuffer.IsEmpty)
+            {
+                return 0;
+            }
+
+            return MemoryMarshal.Read<uint>(readBuffer);
```
krwq, line 82: This should be done using BinaryPrimitives (i.e. BinaryPrimitives.ReadUInt16LittleEndian)

src/devices/QwiicButton/QwiicButton.ButtonStatus.cs
```diff
@@ -0,0 +1,68 @@
...
+        public bool HasBeenClicked()
+        {
+            var status = new StatusRegisterBitField(_registerAccess.ReadSingleRegister(Register.ButtonStatus));
+            return status.HasBeenClicked;
+        }
+
+        /// <summary>
+        /// Returns whether a new button status event has occurred.
+        /// </summary>
+        public bool EventAvailable()
```
krwq, line 32: once you read this I assume you should call `ClearEventBits` immediatelly after, right? If so please make sure there is a remark in the xml doc about this

src/devices/QwiicButton/QwiicButton.ButtonStatus.cs
```diff
@@ -0,0 +1,68 @@
...
+                HasBeenClicked = false,
+                IsPressed = false
+            };
+            _registerAccess.WriteSingleRegister(Register.ButtonStatus, status.StatusRegisterValue);
+        }
+
+        /// <summary>
+        /// Returns the time in milliseconds that the button waits for the mechanical contacts to settle.
+        /// </summary>
+        public ushort GetDebounceTime()
```
krwq, line 55: this and below methods should be a property of type TimeSpan. Documentation should mention recommended/valid range for the property

krwq: If TimeSpan doesn't make sense or is making it hard to use then ushort is fine

src/devices/QwiicButton/QwiicButton.ButtonStatus.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            return _registerAccess.ReadDoubleRegister(Register.ButtonDebounceTime);
```
krwq, line 57: why is this called DoubleRegister while you actually read UInt16? Double is floating point

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
```diff
@@ -0,0 +1,92 @@
+﻿//// Licensed to the .NET Foundation under one or more agreements.
+//// The .NET Foundation licenses this file to you under the MIT license.
+//// See the LICENSE file in the project root for more information.
+
+using Iot.Device.QwiicButton.RegisterMapping;
+
+namespace Iot.Device.QwiicButton
+{
+    public sealed partial class QwiicButton
```
krwq, line 9: unless this class joined together is couple thousands line long I prefer this be in a single file

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
```diff
@@ -0,0 +1,92 @@
+﻿//// Licensed to the .NET Foundation under one or more agreements.
```
krwq, line 1: use double `/` instead of quadra `/` (here and everywhere)

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// then the interrupt will trigger on either a push or a click.
+        /// </summary>
+        public void EnablePressedInterrupt()
+        {
+            var interrupt =
+                new InterruptConfigBitField(_registerAccess.ReadSingleRegister(Register.InterruptConfig))
+                {
+                    PressedEnable = true
+                };
+            _registerAccess.WriteSingleRegister(Register.InterruptConfig, interrupt.InterruptConfigValue);
```
krwq, line 23: I assume this mean UInt8 (byte) rather than single which means `float`? If so please change name

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace Iot.Device.QwiicButton
+{
+    public sealed partial class QwiicButton
+    {
+        /// <summary>
+        /// When called, the interrupt will be configured to trigger when the button is pressed.
+        /// If <see cref="EnableClickedInterrupt"/> has also been called,
+        /// then the interrupt will trigger on either a push or a click.
+        /// </summary>
+        public void EnablePressedInterrupt()
```
krwq, line 16: This and below should be a property (i.e. `PressedInterruptEnabled`)

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
```diff
@@ -0,0 +1,92 @@
...
+                };
+            _registerAccess.WriteSingleRegister(Register.InterruptConfig, interrupt.InterruptConfigValue);
+        }
+
+        /// <summary>
+        /// When called, the interrupt will be configured to trigger when the button is clicked.
+        /// If <see cref="EnablePressedInterrupt"/> has also been called,
+        /// then the interrupt will trigger on either a push or a click.
+        /// </summary>
+        public void EnableClickedInterrupt()
```
krwq, line 46: this and below should be a property i.e. `ClickedInterruptEnabled`

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
```diff
@@ -0,0 +1,92 @@
...
+                new InterruptConfigBitField(_registerAccess.ReadSingleRegister(Register.InterruptConfig))
+                {
+                    ClickedEnable = false
+                };
+
+            _registerAccess.WriteSingleRegister(Register.InterruptConfig, interrupt.InterruptConfigValue);
+        }
+
+        /// <summary>
+        /// Resets the interrupt configuration back to defaults.
```
krwq, line 74: could you mention if default is enabled or disabled?

src/devices/QwiicButton/QwiicButton.LedConfiguration.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        public void LedOff()
+        {
+            LedConfig(0, 0, 0);
+        }
+
+        /// <summary>
+        /// Configures the onboard LED with the given max brightness, granularity, cycle time, and off time.
+        /// <param name="brightness">LED brightness value between 0 (off) and 255 (max). Default is 255.</param>
+        /// <param name="cycleTime">Total pulse cycle time (in ms). Does not include off time.</param>
```
krwq, line 31: I assume for this and the other one/two arguments `0` has special meaning of `do not touch this value`, right? if so please make sure documentation mentions that. If no then please make sure SetBrightness either uses previous values or at minimum mentions that it's not gonna blink anymore

src/devices/QwiicButton/QwiicButton.LedConfiguration.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <param name="offTime">Off time between pulses (in ms). Default is 500 ms.</param>
+        /// <param name="granularity">Amount of steps it takes to get to the set brightness level (1 is fine for most applications).</param>
+        /// </summary>
+        public void LedConfig(byte brightness, ushort cycleTime, ushort offTime, byte granularity = 1)
```
krwq, line 35: Could you add remark on how cycleTime and offTime work? I find these names confusing? Does this mean that led will be on for `cycleTime` and then off for `offTime` and out of `cycleTime` `granularity` time will be spend on making the smooth transition? How does it work when both `cycleTime` and `offTime` are both zeros?

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
```diff
@@ -0,0 +1,119 @@
...
+//// See the LICENSE file in the project root for more information.
+
+using Iot.Device.QwiicButton.RegisterMapping;
+
+namespace Iot.Device.QwiicButton
+{
+    public sealed partial class QwiicButton
+    {
+        /// <summary>
+        /// Returns whether the queue of button press timestamps is full.
```
krwq, line 12: does the register actually have a timestamp? is it public? if not then do not mention timestamp in the doc

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// </summary>
```
krwq, line 13: please add comment to mention what happens when queue is full and you get an event? Are you gonna lose new events or some error condition will happen?

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
```diff
@@ -0,0 +1,119 @@
...
+        public uint TimeSinceLastPress()
+        {
+            return _registerAccess.ReadQuadRegister(Register.PressedQueueFront);
+        }
+
+        /// <summary>
+        /// Returns how many milliseconds it has been since the first button press.
+        /// Since this returns a 32-bit unsigned int, it will roll over about every 50 days.
+        /// </summary>
+        public uint TimeSinceFirstPress()
```
krwq, line 42: can you mention if any method will call this to reset? What will it return if it wasn't pressed at all?

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            return _registerAccess.ReadQuadRegister(Register.PressedQueueBack);
```
krwq, line 44: I assume this means UInt32, if so please rename

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        /// <summary>
+        /// Returns the oldest value in the queue (milliseconds since first button press), and then removes it.
+        /// </summary>
+        public uint PopPressedQueue()
```
krwq, line 50: what happens when queue is empty and this is called?

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
```diff
@@ -0,0 +1,119 @@
...
+            // Remove the oldest value from the queue
+            _registerAccess.WriteSingleRegister(Register.PressedQueueStatus, pressedQueue.QueueStatusValue);
+
+            return timeSinceFirstPress; // Return the value we popped
+        }
+
+        /// <summary>
+        /// Returns whether the queue of button click timestamps is full.
+        /// </summary>
+        public bool IsClickedQueueFull()
```
krwq, line 69: what's the difference between press and click? Is Click Down & Up and Press just down? Usual terminology for that is `KeyDown` and `Click`, I'd stick to that even if the sensor uses different naming in its docs

src/devices/QwiicButton/QwiicButton.cs
```diff
@@ -0,0 +1,96 @@
...
+        private I2cRegisterAccess<Register> _registerAccess;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="QwiicButton"/> class.
+        /// </summary>
+        /// <param name="i2cBusId">I2C bus ID the button is connected to.</param>
+        /// <param name="i2cAddress">I2C bus address of the button (default=0x6F).</param>
+        public QwiicButton(int i2cBusId, byte i2cAddress = DefaultAddress)
+        {
+            I2cBusId = i2cBusId;
```
krwq, line 28: why do you need to store those two? You should at minimum provide constructor which takes I2cDevice - some I2CDevices are provided using i.e. Ft4222 and they can't be constructed using I2cDevice.Create

src/devices/QwiicButton/QwiicButton.cs
```diff
@@ -0,0 +1,96 @@
...
+            I2cAddress = i2cAddress;
+            var settings = new I2cConnectionSettings(i2cBusId, i2cAddress);
+            var device = I2cDevice.Create(settings);
+            _registerAccess = new I2cRegisterAccess<Register>(device);
+        }
+
+        /// <summary>
+        /// I2C bus ID the button is connected to.
+        /// </summary>
+        public int I2cBusId { get; set; }
```
krwq, line 38: please remove this and property below

src/devices/QwiicButton/QwiicButton.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            return _registerAccess.ReadSingleRegister(Register.Id);
+        }
+
+        /// <summary>
+        /// Returns the firmware version of the button as a 16-bit integer.
+        /// The leftmost (high) byte is the major revision number, and the rightmost (low) byte is
+        /// the minor revision number.
+        /// </summary>
+        public ushort GetFirmwareVersionAsInteger()
```
krwq, line 58: please return `Version`

src/devices/QwiicButton/QwiicButton.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            ushort version = (ushort)(_registerAccess.ReadSingleRegister(Register.FirmwareMajor) << 8);
+            version |= _registerAccess.ReadSingleRegister(Register.FirmwareMinor);
+            return version;
+        }
+
+        /// <summary>
+        /// Returns the firmware version of the button as a [major revision].[minor revision] string.
+        /// </summary>
+        public string GetFirmwareVersionAsString()
```
krwq, line 68: please remove, just returning `Version` is enough

src/devices/QwiicButton/QwiicButton.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            var major = _registerAccess.ReadSingleRegister(Register.FirmwareMajor);
+            var minor = _registerAccess.ReadSingleRegister(Register.FirmwareMinor);
+            return major + "." + minor;
+        }
+
+        /// <summary>
+        /// Configures the button to attach to the I2C bus using the specified address.
+        /// </summary>
+        public void SetI2cAddress(byte address)
```
krwq, line 78: Most of the devices become unusable after setting the address and need to be re-constructed, if that's the case here you should rename this to: `SetI2cAddressAndDispose` and dispose on usage

src/devices/QwiicButton/README.md
```diff
@@ -0,0 +1,124 @@
...
+        Initialize();
+        Console.ReadKey(); // Block the thread and wait for OnValueChanged to be invoked
+    }
+}
+
+private void Initialize()
+{
+    var gpioController = new GpioController(PinNumberingScheme.Board, new RaspberryPi3Driver());
+    gpioController.OpenPin(pinNumber: 23, PinMode.Input);
+    gpioController.RegisterCallbackForPinValueChangedEvent(
```
krwq, line 84: is this working fine for you when you're testing? did you see any missed events?

src/devices/QwiicButton/QwiicButton.ButtonStatus.cs
pgrawehr: Name is ambiguous, you're right. "Double" here means "two bytes" (same below). For clarity, maybe call it `ReadUint16Register` (and similarly in other places).

src/devices/Common/Iot/Device/Common/FlagsHelper.cs
Esbob: Fixed

src/devices/Common/Iot/Device/Common/FlagsHelper.cs
Esbob: I've tried to make a consistent mini API to work with flags, therefore the IsSet method.

src/devices/Common/Iot/Device/Common/FlagsHelper.cs
Esbob: Yes, that's why I moved the FlagsHelper back into the QwiicButton solution.

src/devices/Common/Iot/Device/Common/FlagsHelper.cs
Esbob: Bitwise and shift operators are not easy to understand when you haven't worked with them before, like myself. The FlagsHelper conveniently hides away such complexity.

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
Esbob: Fixed

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
Esbob: I've completely rewritten I2cRegisterAccess and made it into the two classes I2cRegisterAccess and I2cRegisterAccess<T>.
I also moved them back into the QwiicButton solution. If you like them, we can later move them into the CommonHelpers solution.

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
Esbob: I wrote a comment about this in I2cRegisterAccess/I2cRegisterAccess<T>.
It seems to me that I2cDevice.Write() does not support register addresses larger than a byte since you have no way of telling the method how many bytes in the input ReadOnlySpan<byte> are register address and how many are data.

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
Esbob: I completely rewrote I2cRegisterAccess - see comments above

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
Esbob: Done

src/devices/QwiicButton/QwiicButton.ButtonStatus.cs
Esbob: Done

src/devices/QwiicButton/QwiicButton.ButtonStatus.cs
Esbob: Done

src/devices/QwiicButton/QwiicButton.ButtonStatus.cs
Esbob: Fixed

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
Esbob: I'm a big fan of separation of concerns and keeping classes small. Since this conflicts with the wish of having all functionality on one device class, using partial classes is in my opinion a decent compromise.

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
Esbob: Fixed

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
Esbob: Fixed

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
Esbob: That would clash with MS' own design guidelines stating that: "Properties are meant to be used like fields, meaning that properties should not be computationally complex or produce side effects".
Read more here: https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100)?redirectedfrom=MSDN

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
Esbob: I disagree - see comment above

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
Esbob: Done

src/devices/QwiicButton/QwiicButton.LedConfiguration.cs
Esbob: Added more explanation.

src/devices/QwiicButton/QwiicButton.LedConfiguration.cs
Esbob: Added more explanation.

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
Esbob: Queue contains timestamps that you can pop.

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
Esbob: Interesting but I don't know - Sparkfun hasn't documented it anywhere and I don't intend on reading the firmware :-)

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
Esbob: Method does not reset anything. Added documentation about what happens if queue is empty.

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
Esbob: Fixed

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
Esbob: Added documentation

src/devices/QwiicButton/QwiicButton.QueueManipulation.cs
Esbob: I opted for `PressedDown` and `Clicked`. `KeyDown` is intended for keyboards. I also added some more explanation of the difference between PressedDown and Clicked.

src/devices/QwiicButton/QwiicButton.cs
Esbob: Removed

src/devices/QwiicButton/QwiicButton.cs
Esbob: Done

src/devices/QwiicButton/QwiicButton.cs
Esbob: Done

src/devices/QwiicButton/QwiicButton.cs
Esbob: Done

src/devices/QwiicButton/QwiicButton.cs
Esbob: Done

src/devices/QwiicButton/README.md
Esbob: On a rare occasion, I might have seen an event not firing, but most of the time it works just fine.

Esbob:
Thanks a lot @krwq for your insightful comments - I agree with most and have changed the code accordingly.

src/devices/Common/Iot/Device/Common/I2cRegisterAccess.cs
Esbob: I took a look at UnixI2cDevice, one of the implementations of I2cDevice, and there doesn't seem to be any references to register addresses or slicing of the data to write. OK, maybe I misunderstood and it is up to each device to know how many of the first bytes in the ReadOnlySpan constitute the register address - do you know if that is how register addressing works in this I2C implementation? If so, my I2cRegisterAccess classes are limited to my 8-bit register addressing use case.

src/devices/QwiicButton/FlagsHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+using System;
+
+namespace Iot.Device.QwiicButton
+{
+    /// <summary>
+    /// Contains functionality to operate on a bit field <see cref="Enum"/>,
+    /// which is an <see cref="Enum"/> decorated with the <see cref="FlagsAttribute"/>
+    /// and which represents a set of flags.
+    /// </summary>
+    public static class FlagsHelper
```
krwq, line 14: this should be internal

src/devices/QwiicButton/QwiicButton.ButtonStatus.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace Iot.Device.QwiicButton
+{
+    public sealed partial class QwiicButton
+    {
+        /// <summary>
+        /// Returns whether the button is pressed down.
+        /// </summary>
+        public bool IsPressedDown()
+        {
+            var status = new StatusRegisterBitField(_registerAccess.ReadRegister<byte>(Register.ButtonStatus));
```
krwq, line 17: I'm looking at all these RegisterAccess APIs and I honestly do not like them. If this was using the regular pattern every other device used this line would be just:
`new StatusRegisterBitField(ReadByte(Register.ButtonStatus));`

You need to ask yourself couple of questions:
- is this solution faster? No
- is this solution require writing less code? No/Maybe similar in optimistic case?
- is this solution clearer and easier to understand? No

given the above IMO you should be consistent with other devices and remove the extra abstraction layer

src/devices/QwiicButton/QwiicButton.InterruptConfiguration.cs
krwq: above comment is about partial classes, this is about this being a property

src/devices/QwiicButton/QwiicButton.ButtonStatus.cs
Esbob: I respectfully disagree - I see again and again the same nitty-gritty read/write byte(s) code if I open a random I2C device binding.
It belongs in a common library, and I don't pretend that my I2cRegisterAccess classes are necessarily good enough to be part of such a library - but it's a start.
As professional developers we have a responsibility for following good development pratices, which means among other to reduce code redundancy, ensure separation of concerns and create comprehensive API's.
In my opinion, when creating a new I2C device binding, developers should only have to focus on the actual "business" of the device, i.e. creating a register map, maybe bit field wrappers and then high-level methods that read/write to the registers defined in the register map. Figuring out how to stackalloc, do some bit shifting and other I2C protocol technicalities is not the real "business" of the device. Plus, it makes it harder for new contributers to create device bindings the right way, which is precisely what happened to me.
Anyway, if you don't allow me to keep my I2cRegisterAccess classes within my own device binding, thereby indirectly saying that you are not interested in trying to improve the quality of the codebase, then I don't think that I want my contribution to be part of this repo and I'll just close the pull request.

joperezr:
[Triage] @Esbob can you please also try rebasing your PR into latest master? We have made some changes in nullability that will likely break the build in your PR.

krwq:
@Esbob you will likely need to rebase and add nullable annotations in this binding. We've added them globally recently. Sorry for the extra work - we weren't able to enable them globally otherwise without breaking every PR

src/devices/QwiicButton/FlagsHelper.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
krwq, line 3: we now use two line copyright header (remove third line in each file)

src/devices/QwiicButton/QwiicButton.ButtonStatus.cs
Esbob: @krwq, before I spend any more time on updating this pull request, please answer my comment right above, so I can decide whether or not I abandon this PR altogether.

krwq:
@Esbob are you planning on re-creating the PR, it seems like a nice addition

Esbob:
@krwq, if you approve my PR including the I2cRegisterAccess classes (see comment from 1st of November), I might.

krwq:
@Esbob sorry, I missed the response (github collapsed it). In the end it's an implementation detail so I'm ok with that as long as all of that remains private/internal.

The reason I'm pushing back on that is we have most of the devices doing something one way and this doing it differently. There might be better ways to abstract this code, i.e. recently introduced C# code generators. I think we should discuss this first rather than experimenting in each binding but as I said in the end this is an implementation detail.

Every binding should be as simple as possible and expose as minimum elements as possible to simplify developer usage.
 
If we see patterns in some elements, which can include this implementation, we bring this into discussion and we try to solve it for all the repository, write documentation, propose pattern. A good example was around the units. There was a need, and we agreed to use UnitsNet to have a unified unit system allowing anyone to use any of their prefered unit.

Esbob:
@krwq, I've changed the code to comply with the new nullable annotation.

krwq:
Hey @Esbob there are still some comments which are not fixed and have no response (i.e. 2-line copyright header)

Esbob:
It should all be fixed by now.

src/devices/QwiicButton/FlagsHelper.cs
Esbob: Fixed

src/devices/QwiicButton/FlagsHelper.cs
Esbob: Fixed

krwq:
@Esbob let us know when this is ready for a review. I'd suggest to try to attempt 1 commit per iteration, that makes it easier to tell when we should review

Esbob:
Ready for review

src/devices/QwiicButton/FlagsHelper.cs
Ellerbach: @Esbob it doesn't seems fix for me. Are you sure you pushed the latest modficiations?

image

PR #1205, opened 2020-10-05
Fix doc for SPI return value

krwq:
found this gem by random 😄


Ellerbach (approved):
You mean when you reveiwed the port to nanoFramewrok ;-)

krwq:
@Ellerbach that too but I thought you copied from elsewhere 😄

PR #1211, opened 2020-10-14
Binding for Grove Bmi088 6 axis IMU

jlandheer:


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1211) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1211) before we can accept your contribution.<br/><br/>:x: jlandheer [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1211)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1211) it.</sub>

Ellerbach (changes requested):
Thanks for this PR, some cosmetic changes required and changing the directory name. All up, great addition!

jlandheer:
I committed the requested changes, should I then mark the conversations as resolved or is that something you do? My first PR so I still have some stuff to  learn.

src/devices/Bmi088/Bmi088.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// <summary>
+        /// Gets or sets the sensitivity of the accelerometer (3, 6, 12 or 24g)
+        /// </summary>
+        public AccelerometerScaleRange AccelerometerScaleRange
+        {
+            get => _accelerometerScaleRange;
+
+            set
+            {
+                if (value == AccelerometerScaleRange.Range3g)
```
Ellerbach, line 97: why not use a switch case?

src/devices/Bmi088/Bmi088.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// <summary>
+        /// Gets or sets the sensitivity of the gyroscope (125, 250, 500, 1000 or 2000 degrees/second)
+        /// </summary>
+        public GyroscopeScaleRange GyroscopeScaleRange
+        {
+            get => _gyroscopeScaleRange;
+
+            set
+            {
+                if (value == GyroscopeScaleRange.Range2000)
```
Ellerbach, line 128: same, why not using a switch case?

src/devices/Bmi088/Bmi088.cs
```diff
@@ -0,0 +1,301 @@
...
+        protected void ResetGyroscope()
+        {
+            WriteByte(_i2cGyroscope, (byte)GyroscopeRegister.SoftReset, 0xB6);
+        }
+
+        /// <summary>
+        /// Id of the accelerometer
+        /// </summary>
+        /// <returns></returns>
+        protected byte GetAccelerometerId()
```
Ellerbach, line 174: I asked in the previous review is this can be a Version? And if this is a specific number that is always the same?
In that case, would make sense to check them at initialization.

src/devices/Bmi088/Bmi088.cs
```diff
@@ -0,0 +1,301 @@
...
+        {
+            return ReadByte(_i2cAccelerometer, (byte)GyroscopeRegister.ChipId);
+        }
+
+        /// <summary>
+        /// Id of the gyroscope
+        /// </summary>
+        /// <returns></returns>
+        protected byte GetGyroscopeId()
```
Ellerbach, line 183: Same as for Accelerator

src/devices/Bmi088/Bmi088.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// </summary>
+        protected void ResetGyroscope()
+        {
+            WriteByte(_i2cGyroscope, (byte)GyroscopeRegister.SoftReset, 0xB6);
+        }
+
+        /// <summary>
+        /// Id of the accelerometer
+        /// </summary>
+        /// <returns></returns>
```
Ellerbach, line 173: missing comment

src/devices/Bmi088/Bmi088.cs
```diff
@@ -0,0 +1,301 @@
...
+        protected byte GetAccelerometerId()
+        {
+            return ReadByte(_i2cAccelerometer, (byte)GyroscopeRegister.ChipId);
+        }
+
+        /// <summary>
+        /// Id of the gyroscope
+        /// </summary>
+        /// <returns></returns>
```
Ellerbach, line 182: missing comment

src/devices/Bmi088/GyroscopeOutputDataRate.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace Iot.Device.Bmi088Device
+{
+    /// <summary>
+    /// Data rates for the gyroscope
+    /// </summary>
+    public enum GyroscopeOutputDataRate : byte
+    {
+        /// <summary>See datasheet</summary>
+        Odr2000_Bw532 = 0x00,
```
Ellerbach, line 9: There should not be any underscore. Also, better to use Bandwidth than Bw. Odr is fine in this case because it is present in the enum name and you can't start an enum with a number.

src/devices/Bmi088/README.md
```diff
@@ -0,0 +1,22 @@
+# Bosch Bmi088 6-Axis Accelerometer & Gyroscope
+The Bosch Bmi088 6-Axis Accelerometer & Gyroscope is a 6 DoF (degrees of freedom) High-performance Inertial Measurement Unit (IMU).
+This sensor is a high-performance IMU with high vibration suppression. The 6-axis sensor combines a 16 bit triaxial gyroscope and a 16 bit triaxial accelerometer.
+
```
Ellerbach, line 4: You can mention in the description you've been using the Grove version if you want.
Please mention that the implementation is I2C only. The sensor supports SPI as well but it's not implemented

src/devices/Bmi088/Bmi088.cs
```diff
@@ -0,0 +1,301 @@
...
+
+namespace Iot.Device.Bmi088Device
+{
+    /// <summary>
+    /// Based on the C++ implementation on https://github.com/Seeed-Studio/Grove_6Axis_Accelerometer_And_Gyroscope_BMI088
+    /// Product page: https://www.bosch-sensortec.com/products/motion-sensors/imus/bmi088.html
+    /// Data sheet: https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bmi088-ds001.pdf
+    /// </summary>
+    public class Bmi088 : IDisposable
+    {
```
Ellerbach, line 14: The sensor support as well a temperature sensor. Would be nice as well to get it. It's in the register of the Accelerometer 0x22 and 0x23. 
And if it's implemented, please use UnitsNet to return the temperature, see here for the pattern: https://github.com/dotnet/iot/tree/master/src/devices#using-units

src/devices/Bmi088/Bmi088.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// <summary>
+        /// Default i2c address for the accelerometer
+        /// </summary>
+        public const byte DEFAULT_ACCELEROMETER_ADDRESS = 0x19;
```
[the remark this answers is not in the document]
Ellerbach: I see as well in the documentation that there is as well the possibility to have 0x18, would be good then to have it as secondary.

src/devices/Bmi088/Bmi088.cs
```diff
@@ -0,0 +1,301 @@
...
+
+        /// <summary>
+        /// Default i2c address for the gyroscope
+        /// </summary>
+        public const byte DEFAULT_GYROSCOPE_ADDRESS = 0x69;
```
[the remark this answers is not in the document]
Ellerbach: Similar to Accelerometer with a secondary address to 0x68

Ellerbach (commented):
Thanks for the very fast commit and changes! Few more comments from my side, all up, looks very good now.

src/devices/Bmi088/AccelerometerOutputDataRate.cs
```diff
@@ -1,32 +1,32 @@
...
+﻿namespace Iot.Device.Bmi088Device
 {
     /// <summary>
     /// Data rate for the accelerometer
     /// </summary>
     public enum AccelerometerOutputDataRate : byte
     {
         /// <summary>See datasheet</summary>
-        ODR_12 = 0x05,
+        Odr12 = 0x05,
```
krwq, line 9: I think @Ellerbach meant to also expand this, also make sure to make it as similar as possible to other APIs:
https://github.com/dotnet/iot/blob/master/src/devices/Ads1115/DataRate.cs
https://github.com/dotnet/iot/blob/master/src/devices/Nrf24l01/DataRate.cs

also make sure units are clear (either included in the name or xml doc)

src/devices/Bmi088/AccelerometerOutputDataRate.cs
```diff
@@ -1,32 +1,32 @@
-﻿namespace Iot.Device.GroveBmi088Device
+﻿namespace Iot.Device.Bmi088Device
```
krwq, line 1: copyright header missing

src/devices/Bmi088/AccelerometerPowerMode.cs
```diff
@@ -1,14 +1,14 @@
-﻿namespace Iot.Device.GroveBmi088Device
+﻿namespace Iot.Device.Bmi088Device
```
krwq, line 1: copyright header here and everywhere where it's missing

src/devices/Bmi088/AccelerometerPowerMode.cs
```diff
@@ -1,14 +1,14 @@
...
     /// </summary>
     public enum AccelerometerPowerMode : byte
     {
         /// <summary>See datasheet</summary>
-        ACC_ACTIVE = 0x00,
+        Active = 0x00,
 
         /// <summary>See datasheet</summary>
-        ACC_SUSPEND = 0x03
+        Suspend = 0x03
```
krwq, line 12: Suspended

src/devices/Bmi088/AccelerometerRegister.cs
```diff
@@ -1,80 +1,80 @@
...
 {
     /// <summary>
     /// Registers for the accelerometer
     /// </summary>
-    public enum AccelerometerRegister : byte
+    internal enum AccelerometerRegister : byte
     {
         /// <summary>See datasheet</summary>
-        CHIP_ID = 0x00, // Default value 0x1E
+        ChipId = 0x00, // Default value 0x1E
```
krwq, line 9: for internal enums, especially those matching the spec we do not require renaming, it's fine both renamed or original spec names. We only require that for public APIs

src/devices/Bmi088/Bmi088.cs
```diff
@@ -0,0 +1,301 @@
...
+        private GyroscopeScaleRange _gyroscopeScaleRange;
+        private AccelerometerPowerMode _accelerometerPowerMode;
+        private GyroscopePowerMode _gyroscopePowerMode;
+        private AccelerometerOutputDataRate _accelerometerOutputDataRate;
+        private GyroscopeOutputDataRate _gyroscopeOutputDataRate;
+
+        /// <summary>
+        /// Initialize on default bus (1) with default I2C addresses.
+        /// </summary>
+        public Bmi088()
```
krwq, line 42: Consider doing following:
- create new classes:
  - Bmi088AccelerometerAndThermometer
  - Bmi088Gyroscope
- make each one of them have public `DefaultI2cAddress` const field
- make it possible to change I2cAddress (so ctor should be like `Bmi088Accelerometer(I2cDevice i2cDevice = null, bool shouldDispose = false)` and on null you should: create I2cDevice with default address, enforce disposal ([similar to what we do with GpioController](https://github.com/dotnet/iot/blob/master/src/devices/README.md#gpiocontroller))
- this implementation (Bmi088) should include capabilities of all 3 classes but use implementation provided by the split classes

see example for https://github.com/dotnet/iot/blob/master/src/devices/SenseHat/SenseHat.cs to see how can this be done

also lots of time when device has two i2c addresses it may consist of two separate sensors which may have different names, if you find that names then please create a new binding out of that and make this binding use it, if not then leave this in this folder (you might need to check spec or perhaps some forums - possibly we even already support it - you might i.e. compare registers of some accelerometers we have to see if there is overlap)

If you don't find names of the split devices treat this feedback as optional although still consider for clarity.

krwq: @Ellerbach as a side note this is the first time I'm seeing that I2cBus class would be useful for i.e. Ft4222 (you could create constructor which takes bus but leaves default addresses alone)

krwq: Quickly looking at 16-bit accelerometers bosch provides actually each one of them might work with this code, I suspect same will be with gyroscope. I.e. https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bma490l-ds000.pdf (I know at least couple of registers match here) - would be nice to support whole family of them rather than just this one sensor

krwq: I suspect this one as well will work here: https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BMA456-DS000.pdf

Ellerbach: > @Ellerbach as a side note this is the first time I'm seeing that I2cBus class would be useful for i.e. Ft4222 (you could create constructor which takes bus but leaves default addresses alone)

I It will work with multiple I2C. I tested this scenario. The I2cDevices just need to be created separately. And there is a constructor to pass both I2cDevices.

joperezr:
[Triage]: @jlandheer you get to choose if you want to mark the conversations as resolved, if it helps you to track it better that is fine so it is up to you. Looks like @Ellerbach and @krwq added more feedback here that you haven't addressed, any idea when you plan to check that?

joperezr:
[Triage] @jlandheer friendly ping, do you plan on addressing the feedback that is remaining?

PR #1212, opened 2020-10-14
Making Status property also internal to fix broken build

joperezr:
Fix borken build caused by https://github.com/dotnet/iot/pull/1067

cc: @krwq @famblard86


joperezr:
I’m merging this in as is for now to unblock the build but feel free to put up a subsequent PR if you wanted to fix this by making everything public instead.

PR #1213, opened 2020-10-15
Binding for PiJuice UPS HAT

### LINKED ISSUE #1096: PiJuice UPS Implementation
I currently have a C# .NET Standard library [(based on)](https://github.com/tresoft/PiJuice-Uwp) which I developed for the [PiJuice](https://github.com/PiSupply/PiJuice).

I was wondering if it would be of any use in the Iot.Device.Bindings project, or is it to specific because it is a Raspberry Pi only HAT?

DazFahy:
Fixes #1096

Add device binding for PiJuice UPS HAT

[PiJuice documentation](https://github.com/PiSupply/PiJuice)

EDIT (@krwq): `Fixes #1096` should not be quoted, github likely won't close issue automatically with that


src/devices/PiJuice/Models/LEDConfig.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    /// <summary>
+    /// LED configuration
+    /// </summary>
+    public class LEDConfig
+    {
+        /// <summary>
+        /// LED designator
+        /// </summary>
+        public LED LED { get; set; }
```
Ellerbach, line 17: Led Led

src/devices/PiJuice/Models/LEDConfig.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// Color for LED
+        /// If LedFunction is ChargeStatus
+        /// Red - parameter defines color component level of red below 15%
+        /// Green - parameter defines color component charge level over 50%
+        /// Blue - parameter defines color component for charging(blink) and fully charged states(constant)
+        /// Red LED and Green LED will show the charge status between 15% - 50%
+        /// </summary>
+        public Color RGB { get; set; }
```
Ellerbach, line 32: Rgb or just Color

src/devices/PiJuice/Models/LEDFunction.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.PiJuiceDevice.Models
+{
+    /// <summary>
+    /// LED function type
+    /// </summary>
+    public enum LEDFunction
```
Ellerbach, line 10: LedFunction

src/devices/PiJuice/Models/LEDFunction.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// For level greater than 15% and level less than or equal to 50% mix of red and green with configurable brightness
+        /// For level greater than 50% green with configurable brightness.
+        /// When battery is charging blinking blue with configurable brightness is added to current charge level color. For full buttery state blue component is steady on
+        /// </summary>
+        ChargeStatus,
+
+        /// <summary>
+        /// LED is configured as user LED
+        /// </summary>
+        UserLED
```
Ellerbach, line 29: Maybe `UserDefined`?

src/devices/PiJuice/Models/PiJuiceCommand.cs
```diff
@@ -0,0 +1,157 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.PiJuiceDevice.Models
+{
+    /// <summary>
+    /// PiJuice commands to read, write, setup pins and access special sensors
+    /// </summary>
+    public enum PiJuiceCommand
```
Ellerbach, line 10: Does this enum needs to be public? Maybe internal?

src/devices/PiJuice/Models/PiJuiceCommand.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        /// <summary>
+        /// Battery Profile
+        /// </summary>
+        BatteryProfile = 0x53,
+
+        /// <summary>
+        /// Battery extended profile
+        /// </summary>
+        BatteryExtProfile = 0x54,
```
Ellerbach, line 30: `BatteryExtendedProfile`

src/devices/PiJuice/Models/PiJuiceCommand.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        /// <summary>
+        /// Battery temperature sensor configuration
+        /// </summary>
+        BatteryTempSenseConfig = 0x5D,
```
Ellerbach, line 35: `BatteryTemperatureSensorConfig`

src/devices/PiJuice/Models/PiJuiceCommand.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        /// <summary>
+        /// Provide external power
+        /// </summary>
+        SystemPowerSwitch = 0x64,
+
+        /// <summary>
+        /// LED States
+        /// </summary>
+        LEDState = 0x66,
```
Ellerbach, line 125: `LedState`

src/devices/PiJuice/Models/PiJuiceCommand.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        /// <summary>
+        /// LED blink pattern
+        /// </summary>
+        LEDBlink = 0x68,
```
Ellerbach, line 130: `LedBlink`

src/devices/PiJuice/Models/PiJuiceCommand.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        /// <summary>
+        /// LED configuration
+        /// </summary>
+        LEDConfig = 0x6A,
```
Ellerbach, line 135: `LedConfig`

src/devices/PiJuice/Models/PowerInput.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        /// <summary>
+        /// Selects what power input will have precedence for charging and supplying VSYS output when both are present, PiJuice USB Micro Input, GPIO 5V Input. 5V_GPIO selected by default
+        /// </summary>
+        public PowerInputType Precedence { get; set; }
+
+        /// <summary>
+        /// Enables/disables powering PiJuice from 5V GPIO Input. Enabled by default
+        /// </summary>
+        public bool GPIOIn { get; set; }
```
Ellerbach, line 22: GpioIn

src/devices/PiJuice/Models/PowerInput.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// If enabled PiJuice will automatically power on 5V rail and trigger wake up as soon as power appears at USB Micro Input and there is no battery. Disabled by default
+        /// </summary>
+        public bool NoBatteryTurnOn { get; set; }
+
+        /// <summary>
+        /// Maximum current that PiJuice can take from USB Micro connected power source. 2.5A selected by default
+        /// </summary>
+        public ElectricCurrent USBMicroCurrentLimit { get; set; }
```
Ellerbach, line 32: UsbMicroCurrentLimit

src/devices/PiJuice/Models/PowerInput.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Minimum voltage at USB Micro power input for Dynamic Power Management Loop. 4.2V set by default
+        /// </summary>
+        public ElectricPotential USBMicroDPM { get; set; }
```
Ellerbach, line 37: UsbMicroDynamicPowerManagement

src/devices/PiJuice/Models/PowerInputType.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Power input type for charging and supplying VSYS output
+    /// </summary>
+    public enum PowerInputType
+    {
+        /// <summary>
+        /// USB micro input
+        /// </summary>
+        USBMicro,
```
Ellerbach, line 15: `UsbMicro`

src/devices/PiJuice/Models/PowerInputType.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        /// <summary>
+        /// GPIO 5V
+        /// </summary>
+        GPIO5V
```
Ellerbach, line 20: Gpio5Volt

src/devices/PiJuice/Models/PowerRegulatorMode.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        /// <summary>
+        /// This mode switches between DCDC switching mode and LDO mode with most of time in DCDC switching mode. This is mode has high efficiency but increased voltage ripple
+        /// </summary>
+        PowerSourceDetection = 0,
+
+        /// <summary>
+        /// This mode regulates 5V Rail voltage to 4.79V. In this mode output voltage has lowest output ripple
+        /// </summary>
+        LDO,
```
Ellerbach, line 20: Long name for LDO. what does it stand for?

src/devices/PiJuice/Models/PowerRegulatorMode.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        /// <summary>
+        /// This mode 5V Rail voltage is regulated to 5V with 2.5% tolerance, typically 5.07V at mid-loaded conditions. This is the most efficient operation mode
+        /// </summary>
+        DCDC
```
Ellerbach, line 25: Same as above

src/devices/PiJuice/Models/RSOCEstimationType.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.PiJuiceDevice.Models
+{
+    /// <summary>
+    /// Battery relative state-of-health
+    /// </summary>
+    public enum RSOCEstimationType
```
Ellerbach, line 10: RsocEstimationType or a long name rather than Rsoc. It's not ver know abbreviation

src/devices/PiJuice/Models/RSOCEstimationType.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// Let the PiJuice software determine RSOC estimation
+        /// </summary>
+        AutoDetect = 0,
+
+        /// <summary>
+        /// MCU used for RSOC estimation
+        /// </summary>
+        DirectByMCU
```
Ellerbach, line 20: DirectByMcu

src/devices/PiJuice/Models/Status.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        /// <summary>
+        /// Current USB Micro power input status
+        /// </summary>
+        public PowerInState PowerInput { get; set; }
+
+        /// <summary>
+        /// Current 5V GPIO power input status
+        /// </summary>
+        public PowerInState PowerInput5vIo { get; set; }
```
Ellerbach, line 35: I know I made a comment previously on this one or similar. Please refer to it and try to make it consistant, whatever is the choice :-)

src/devices/PiJuice/Models/WakeUpOnCharge.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// Is the wake up on charge disabled
+        /// </summary>
+        public bool Disabled { get; set; }
+
+        /// <summary>
+        /// Battery charge level percentage between [0 - 100] used to wake up the Raspberry Pi
+        /// </summary>
+        public short WakeUpPercentage { get; set; }
```
Ellerbach, line 20: I'm debating with myself for what is percent. For bindings with humidity and the rest, we use Ratio. I guess for consistency, would be better to use Ratio as well. Maybe @krwq have some opinion as well.

src/devices/PiJuice/PiJuice.cs
```diff
@@ -0,0 +1,240 @@
...
+        /// <summary>
+        /// Get the firmware version
+        /// </summary>
+        /// <returns>PiJuice firmware version</returns>
+        public Version GetFirmwareVerion()
+        {
+            var response = ReadCommand(PiJuiceCommand.FirmwareVersion, 2);
+
+            var major_version = response[0] >> 4;
+            var minor_version = (response[0] << 4 & 0xf0) >> 4;
```
Ellerbach, line 66: maybe `response[0] & 0x0F` is doing the same, right?

src/devices/PiJuice/PiJuice.cs
```diff
@@ -0,0 +1,240 @@
...
+
+            return new Version(major_version, minor_version);
+        }
+
+        /// <summary>
+        /// Restore PiJuice to its default settings
+        /// </summary>
+        public void SetDefaultConfiguration()
+        {
+            WriteCommand(PiJuiceCommand.ResetToDefault, new byte[] { 0xAA, 0x55, 0x0A, 0xA3 });
```
Ellerbach, line 76: if you can add a comment on top referencing from where the magic numbers are coming from like pointer on the page of the doc, it would be better

src/devices/PiJuice/PiJuiceConfig.cs
```diff
@@ -0,0 +1,414 @@
...
+{
+    /// <summary>
+    /// PiJuiceConfig class to support status of the PiJuice
+    /// </summary>
+    public class PiJuiceConfig
+    {
+        private readonly PiJuice _piJuice;
+
+        private readonly List<ElectricCurrent> _usbMicroCurrentLimits = new List<ElectricCurrent> { new ElectricCurrent(1.5, ElectricCurrentUnit.Ampere), new ElectricCurrent(2.5, ElectricCurrentUnit.Ampere) };
+        private readonly List<ElectricPotential> _usbMicroDPMs = new List<ElectricPotential>();
```
Ellerbach, line 22: _usbMicroDpm

Ellerbach (commented):
Thanks, look very promising and great addition!
Partial review so far, I'll continue, quite some cosmetic changes to start with.

src/devices/PiJuice/PiJuice.cs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        /// <summary>
+        /// Get the write protect of the EEPROM
+        /// </summary>
+        /// <returns>Whether the EEPROM is write protected</returns>
+        public bool GetIdEepromWriteProtect()
```
Ellerbach, line 83: You could make the `IdEepromWriteProtect` a property. There is a set and a get, it doesn't change over time

src/devices/PiJuice/PiJuice.cs
```diff
@@ -0,0 +1,240 @@
...
+        public void SetIdEepromWriteProtect(bool writeProtect)
+        {
+            WriteCommandVerify(PiJuiceCommand.IdEepromWriteProtect, new byte[] { (byte)(writeProtect ? 1 : 0) });
+        }
+
+        /// <summary>
+        /// Get the physical I2C address of the EEPROM
+        /// </summary>
+        /// <returns>The I2C Address of the EEPROM</returns>
+        public IdEepromAddress GetIdEepromAddress()
```
Ellerbach, line 103: Same here, a property would be perfect

src/devices/PiJuice/PiJuice.cs
```diff
@@ -0,0 +1,240 @@
...
+            IOException innerEx = null;
+            // When writing/reading to the I2C port, PiJuice doesn't respond on time in some cases
+            // So we wait a little bit before retrying
+            // In most cases, the I2C read/write can go thru without waiting
+            while (tries < MaxRetries)
+            {
+                try
+                {
+                    // create request
+                    var buffer = new byte[data.Length + 2];
```
Ellerbach, line 136: You shouls use a `Span<byte>` with stackalloc

src/devices/PiJuice/PiJuice.cs
```diff
@@ -0,0 +1,240 @@
...
+        /// Read data from PiJuice
+        /// </summary>
+        /// <param name="command">The PiJuice command</param>
+        /// <param name="length">The length of the data to be read</param>
+        /// <returns>Returns an array of the bytes read</returns>
+        internal byte[] ReadCommand(PiJuiceCommand command, byte length)
+        {
+            byte[] outArray = new byte[length + 1];
+            byte tries = 0;
+            IOException innerEx = null;
```
Ellerbach, line 188: why don't you just define it where it's raised?

src/devices/PiJuice/PiJuice.cs
```diff
@@ -0,0 +1,240 @@
...
+
+        /// <summary>
+        /// Write a PiJuice command
+        /// </summary>
+        /// <param name="command">The PiJuice command</param>
+        /// <param name="data">The data to write</param>
+        internal void WriteCommand(PiJuiceCommand command, byte[] data)
+        {
+            byte tries = 0;
+            IOException innerEx = null;
```
Ellerbach, line 127: why don't you just define it where it's raised?

src/devices/PiJuice/PiJuicePower.cs
```diff
@@ -0,0 +1,127 @@
...
+        public PiJuicePower(PiJuice piJuice)
+        {
+            _piJuice = piJuice;
+        }
+
+        /// <summary>
+        /// Gets the delay before the PiJuice removes power to the GPIO pins
+        /// </summary>
+        /// <returns>The delay in seconds</returns>
+        public byte GetPowerOff()
```
Ellerbach, line 30: A property would work `byte PowerOff` or even better `TimeSpan PowerOff`
For all time related delay, we tend to use TimeSpan to be consistent

src/devices/PiJuice/PiJuicePower.cs
```diff
@@ -0,0 +1,127 @@
...
+            return response[0];
+        }
+
+        /// <summary>
+        /// Sets the delay before the PiJuice removes power to the GPIO pins
+        /// </summary>
+        /// <param name="delaySeconds">The delay in seconds between 0 and 255</param>
+        public void SetPowerOff(byte delaySeconds)
+        {
+            if (delaySeconds < 0)
```
Ellerbach, line 43: with a byte, that should never happen :-) Now with a TimeSpan neither, but you should check the maximum number of seconds

src/devices/PiJuice/PiJuicePower.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+
+            _piJuice.WriteCommand(PiJuiceCommand.PowerOff, new byte[] { delaySeconds });
+        }
+
+        /// <summary>
+        /// Get the current state of the Wakeup on charge
+        /// </summary>
+        /// <returns>Current state of the wake up on charge function</returns>
+        public WakeUpOnCharge GetWakeUpOnCharge()
```
Ellerbach, line 55: Same, a property would perfectly work

src/devices/PiJuice/PiJuicePower.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+
+            _piJuice.WriteCommandVerify(PiJuiceCommand.WakeUpOnCharge, new byte[] { (byte)(wakeUpOnCharge.Disabled ? 0x7F : wakeUpOnCharge.WakeUpPercentage) });
+        }
+
+        /// <summary>
+        /// Gets the current watchdog timer time after which it will power cycle if it does not receive a heartbeat signal
+        /// </summary>
+        /// <returns>Time in minutes after which PiJuice will power cycle if it does not receive a heartbeat signal</returns>
+        public TimeSpan GetWatchdogTimer()
```
Ellerbach, line 84: Same, property would fit perfectly here as well

src/devices/PiJuice/PiJuicePower.cs
```diff
@@ -0,0 +1,127 @@
...
+            var minutes = (int)time.TotalMinutes & 0xFFFF;
+
+            _piJuice.WriteCommand(PiJuiceCommand.WatchdogActiviation, new byte[] { (byte)(minutes & 0xFF), (byte)((minutes >> 8) & 0xFF) });
+        }
+
+        /// <summary>
+        /// Gets the current state of system switch
+        /// </summary>
+        /// <returns>Current state of system switch</returns>
+        public SystemPowerSwitch GetSystemPowerSwitch()
```
Ellerbach, line 111: idem

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        /// Get battery temperature
+        /// </summary>
+        /// <returns>Battery temperature in celsius</returns>
+        public Temperature GetBatteryTemperature()
+        {
+            var response = _piJuice.ReadCommand(PiJuiceCommand.BatteryTemperature, 2);
+
+            return new Temperature((response[1] << 8) | response[0], TemperatureUnit.DegreeCelsius);
```
Ellerbach, line 123: BinaryPrimitives for the conversion?

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        /// Get battery voltage
+        /// </summary>
+        /// <returns>Battery voltage in millivolts</returns>
+        public ElectricPotential GetBatteryVoltage()
+        {
+            var response = _piJuice.ReadCommand(PiJuiceCommand.BatteryVoltage, 2);
+
+            return new ElectricPotential((response[1] << 8) | response[0], ElectricPotentialUnit.Millivolt);
```
Ellerbach, line 134: same, seemes like BinaryPrimitives can be used here

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        /// Get battery current
+        /// </summary>
+        /// <returns>Battery current in milliamps</returns>
+        public ElectricCurrent GetBatteryCurrent()
+        {
+            var response = _piJuice.ReadCommand(PiJuiceCommand.BatteryCurrent, 2);
+
+            var i = (response[1] << 8) | response[0];
```
Ellerbach, line 145: idem with BinaryPrimitives

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -0,0 +1,275 @@
...
+            }
+
+            return new ElectricCurrent(i, ElectricCurrentUnit.Milliampere);
+        }
+
+        /// <summary>
+        /// Get supplied voltage
+        /// </summary>
+        /// <returns>Voltage supplied from the GPIO power output from the PiJuice or when charging, voltage supplied in millivolts</returns>
+        public ElectricPotential GetIOVoltage()
```
Ellerbach, line 158: `GetIoVoltage` most likely. Ssomehow for tIO I'll be ok with either IO either Io. But it has to be consistent everyone. Still a preference for Io.

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            var response = _piJuice.ReadCommand(PiJuiceCommand.IOVoltage, 2);
+
+            return new ElectricPotential((response[1] << 8) | response[0], ElectricPotentialUnit.Millivolt);
```
Ellerbach, line 162: BinaryPrimitives?

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        /// Get supplied current in milliamps
+        /// </summary>
+        /// <returns>Current supplied from the GPIO power output from the PiJuice or when charging, current supplied in milliamps</returns>
+        public ElectricCurrent GetIOCurrent()
+        {
+            var response = _piJuice.ReadCommand(PiJuiceCommand.IOCurrent, 2);
+
+            var i = (response[1] << 8) | response[0];
```
Ellerbach, line 173: BinaryPrimitives?

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -0,0 +1,275 @@
...
+
+            return new ElectricCurrent(i, ElectricCurrentUnit.Milliampere);
+        }
+
+        /// <summary>
+        /// Get the color for a specific LED
+        /// </summary>
+        /// <param name="led">LED to get color for</param>
+        /// <returns>Color of LED</returns>
+        public Color GetLEDState(LED led)
```
Ellerbach, line 187: `GetLedState` consiste,t with `SetLedState`

Ellerbach (commented):
Still looks goo now I managed to get thru everything! Thanks for this great addition.

src/devices/PiJuice/Models/IdEepromAddress.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// EEPROM I2C Address 0x50
+        /// </summary>
+        Fifty = 0x50,
+
+        /// <summary>
+        /// Default EEPROM I2C Address 0x52
+        /// </summary>
+        FiftyTwo = 0x52
```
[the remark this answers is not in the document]
DazFahy: I think this should be First as 0x52 is the default I2C address

src/devices/PiJuice/Models/IdEepromAddress.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// EEPROM I2C Address
+    /// </summary>
+    public enum IdEepromAddress
+    {
+        /// <summary>
+        /// EEPROM I2C Address 0x50
+        /// </summary>
+        Fifty = 0x50,
```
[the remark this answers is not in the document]
DazFahy: I think this should be Second as 0x52 is the default I2C address

src/devices/PiJuice/PiJuice.cs
DazFahy: I've checked the docs pages and there is no mention of what the magic numbers are. I've even checked the c firmware code along with the python code for any comments but there are none.

src/devices/PiJuice/PiJuice.cs
DazFahy: It is inside a try catch which is itself inside a while loop for a specific number of retries. Also this code is based on the GrovePi.

I could change the catch to check if the attempts has exceeded the retries and change the while loop to a for loop?

src/devices/PiJuice/PiJuicePower.cs
DazFahy: Just for the GetWakeUpOnCharge or for the SetWakeUpOnCharge also?

Ellerbach: You have a setter and a getter, so both transforming to a property

src/devices/PiJuice/PiJuiceStatus.cs
DazFahy: I'm going to stick with IO on the bases that it is used in the exception IOException which comes from the GrovePi, which is what I based this project on.

Ellerbach: Fair enough. Go then for IO. Please check everywhere to be consistent then.

src/devices/PiJuice/PiJuiceStatus.cs
DazFahy: Have you an example of using BinaryPrimitives as I've never used the class before?

src/devices/PiJuice/Models/BatteryTempSense.cs
```diff
@@ -17,7 +17,7 @@ public enum BatteryTempSense
         /// <summary>
         /// Use batteries built-in NTC as per battery NTC terminal
         /// </summary>
-        NTC,
+        NegativeTemperatureCoefficientB,
```
Ellerbach, line 20: Are you sure about the B at the end?

src/devices/PiJuice/Models/LEDBlink.cs
```diff
@@ -29,21 +29,21 @@ public class LedBlink
...
         /// Color for first period of blink
         /// </summary>
-        public Color RGB1 { get; set; }
+        public Color ColorFirstPeriod { get; set; }
 
         /// <summary>
         /// Duration of first blink period in milliseconds between [10 – 2550]
         /// </summary>
-        public int Period1 { get; set; }
+        public int FirstPeriod { get; set; }
```
Ellerbach, line 37: Didn't catch this one before but if you have timing, on public classes, then please use TimeSpan.  Especially here with a range than is more than few milliseconds. And great that you put the range. You will have to adjust a bit the comment as well.

src/devices/PiJuice/Models/LEDBlink.cs
```diff
@@ -29,21 +29,21 @@ public class LedBlink
...
         /// Color for second period of blink
         /// </summary>
-        public Color RGB2 { get; set; }
+        public Color ColorSecondPeriod { get; set; }
 
         /// <summary>
         /// Duration of second blink period in milliseconds between [10 – 2550]
         /// </summary>
-        public int Period2 { get; set; }
+        public int SecondPeriod { get; set; }
```
Ellerbach, line 47: Same as for first period

src/devices/PiJuice/PiJuice.cs
```diff
@@ -63,57 +63,53 @@ public Version GetFirmwareVerion()
...
             var major_version = response[0] >> 4;
-            var minor_version = (response[0] << 4 & 0xf0) >> 4;
+            var minor_version = response[0] & 0x0F;
 
             return new Version(major_version, minor_version);
         }
 
         /// <summary>
-        /// Restore PiJuice to its default settings
+        /// Restore PiJuice to its default settings (https://github.com/PiSupply/PiJuice/blob/02c129c3aa884934c368936c9363df90690148a8/Software/Source/pijuice.py#L1531)
```
Ellerbach, line 72: just the url as a comment before the WriteCommand, it doesn't need to be in the description of what the function is doing

src/devices/PiJuice/PiJuice.cs
```diff
@@ -63,57 +63,53 @@ public Version GetFirmwareVerion()
...
         }
 
         /// <summary>
-        /// Get the write protect of the EEPROM
+        /// the write protectection of the EEPROM
         /// </summary>
-        /// <returns>Whether the EEPROM is write protected</returns>
-        public bool GetIdEepromWriteProtect()
+        /// <value>Whether the EEPROM is write protected</value>
+        public bool GetIdEepromWriteProtect
```
Ellerbach, line 83: As it's a property now, just name it `IdEepromWriteProtect`

src/devices/PiJuice/PiJuice.cs
```diff
@@ -63,57 +63,53 @@ public Version GetFirmwareVerion()
...
         }
 
         /// <summary>
-        /// Get the physical I2C address of the EEPROM
+        /// The physical I2C address of the EEPROM
         /// </summary>
-        /// <returns>The I2C Address of the EEPROM</returns>
-        public IdEepromAddress GetIdEepromAddress()
+        /// <value>The I2C Address of the EEPROM</value>
+        public IdEepromAddress GetIdEepromAddress
```
Ellerbach, line 101: same, the Get is not need, as a property, `IdEepromAddress` is enough

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -120,7 +121,7 @@ public Temperature GetBatteryTemperature()
         {
             var response = _piJuice.ReadCommand(PiJuiceCommand.BatteryTemperature, 2);
 
-            return new Temperature((response[1] << 8) | response[0], TemperatureUnit.DegreeCelsius);
+            return new Temperature(BinaryPrimitives.ReadInt16LittleEndian(new Span<byte>(response)), TemperatureUnit.DegreeCelsius);
```
Ellerbach, line 124: you don't need new Span<byte>, this is implicit with byte[]
Same everywhere you've used it

Ellerbach (commented):
Minor comments. I love it all up! You'll make me buy one!

src/devices/PiJuice/Models/BatteryTempSense.cs
DazFahy: Its a copy and paste error :-)

DazFahy:
@Ellerbach they are a really good product, just very hard to get hold of at the minute. I use mine to keep up my sensor box I have developed

src/devices/PiJuice/PiJuice.cs
Ellerbach: Yes, I would raise the exception in the catch part if the tries are higher than the maximum. But that's ok as well like this. It was really to avoid allocating the IOException for nothing

src/devices/PiJuice/PiJuice.cs
```diff
@@ -0,0 +1,240 @@
...
+
+        /// <summary>
+        /// Read data from PiJuice
+        /// </summary>
+        /// <param name="command">The PiJuice command</param>
+        /// <param name="length">The length of the data to be read</param>
+        /// <returns>Returns an array of the bytes read</returns>
+        internal byte[] ReadCommand(PiJuiceCommand command, byte length)
+        {
+            byte[] outArray = new byte[length + 1];
```
Ellerbach, line 186: This should be a Span<byte> with stackalloc. You're using it with I2cDevice.Read. So better use a Span<byte>

DazFahy: Should the ReadCommand still return a byte array or should it return a Span<byte> ?

Ellerbach: It's clear that something like `void ReadCommand(PiJuiceCommand command, Span<byte> readBuffer)` would be a better design. Same for the Write command. In this case, you would have to adjust the readBuffer size to match the length+1 size. Similar for the write command. Now those are internals, I saw how it's used for the rest, you just grab one or 2 bytes every time. I don't have a super strong opinion about it.
@krwq may have a stronger opinion.

DazFahy: As an example then

```csharp
WriteCommand(PiJuiceCommand.ResetToDefault, new byte[] { 0xAA, 0x55, 0x0A, 0xA3 });
   ```

would become

```csharp
Span<byte> buffer = stackalloc byte[4 + 2]
{
    0,
    0xAA,
    0x55,
    0x0A,
    0xA3,
    0
};

WriteCommand(PiJuiceCommand.ResetToDefault, buffer);
   ```

Comment (DazFahy) [reply src/devices/PiJuice/PiJuice.cs:186]: For the read command

```csharp
var response = ReadCommand(PiJuiceCommand.FirmwareVersion, 2);
```

Would become

```csharp
Span<byte> buffer = stackalloc byte[2 + 1];
buffer.Clear();

ReadCommand(PiJuiceCommand.FirmwareVersion, buffer);

var response = buffer.Slice(0, 2);
```

You would need to slice the buffer in the calling method as this is currently done in the ReadCommand

Ellerbach: As I said, I have no strong opinion on this. I always design API with Span<byte> when it's public. When private sometimes, it's with byte[] or Span<byte> when it does make sense.

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -0,0 +1,276 @@
...
+            data[8] = (byte)((int)(ledBlink.SecondPeriod.TotalMilliseconds / 10) & 0xFF);
+
+            _piJuice.WriteCommand(PiJuiceCommand.LedBlink + (byte)ledBlink.Led, data);
+        }
+
+        /// <summary>
+        /// Get button event type
+        /// </summary>
+        /// <param name="buttonEventTypeIndex">Button event type index</param>
+        private ButtonEventType GetButtonEventType(int buttonEventTypeIndex)
```
Ellerbach, line 266: I'm wondering if Enum.IsDefined is not doing exactly the same thing?

src/devices/PiJuice/samples/PiJuiceDevice.sample.cs
```diff
@@ -0,0 +1,48 @@
...
+using Iot.Device.PiJuiceDevice;
+using Iot.Device.PiJuiceDevice.Models;
+using UnitsNet;
+using UnitsNet.Units;
+
+namespace PiJuiceDevice.Sample
+{
+    internal class Program
+    {
+        public static void Main(string[] args)
```
Ellerbach, line 13: as it's a public class, I guess you should add a comment.

Ellerbach (approved):
Minor comments, all up after the changes will be good to me

src/devices/PiJuice/PiJuice.cs
DazFahy: I will leave it as is.

src/devices/PiJuice/PiJuice.cs
```diff
@@ -0,0 +1,252 @@
...
+            outArray.Clear();
+
+            // When writing/reading the I2C port, PiJuice doesn't respond on time in some cases
+            // So we wait a little bit before retrying
+            // In most cases, the I2C read/write can go thru without waiting
+            while (tries < MaxRetries)
+            {
+                try
+                {
+                    _i2cDevice.Write(new byte[] { (byte)command });
```
krwq, line 196: did you try joining read/write operation into ReadWrite? (note you will likely need to give 1 more byte on front in read buffer and ignore it after it finished reading) perhaps that's why you are getting timeouts and possibly retry won't be required

DazFahy: Hi @krwq, yes the original code used a ReadWrite but would fail about one third of the time. Since I have used the seperate read write commands i hav not seen a failure, i'm guessing it is just a timing issue.

src/devices/PiJuice/PiJuice.cs
```diff
@@ -0,0 +1,252 @@
...
+            }
+
+            return outArray.Slice(0, length).ToArray();
+        }
+
+        /// <summary>Gets the check sum.</summary>
+        /// <param name="data">The data.</param>
+        /// <param name="all">Whether the last byte in the data is included in the checksum</param>
+        /// <returns>Checksum</returns>
+        private byte GetCheckSum(ReadOnlySpan<byte> data, bool all = false)
```
krwq, line 240: you should rename `all` to `checkLastByte`

src/devices/PiJuice/PiJuicePower.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            get
+            {
+                var response = _piJuice.ReadCommand(PiJuiceCommand.WatchdogActiviation, 2);
+
+                return new TimeSpan(0, BinaryPrimitives.ReadInt16LittleEndian(response), 0);
+            }
+            set
+            {
+                if (value.TotalMinutes < 0 || value.TotalMinutes > 65535)
```
krwq, line 92: upper limit seems a bit high, it's 45 days

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -0,0 +1,262 @@
...
+            }
+
+            if (ledBlink.SecondPeriod.TotalMilliseconds < 10 || ledBlink.SecondPeriod.TotalMilliseconds > 2550)
+            {
+                throw new ArgumentOutOfRangeException(nameof(ledBlink.SecondPeriod));
+            }
+
+            var data = new byte[9];
+
+            data[0] = (byte)(ledBlink.Count & 0xFF);
```
krwq, line 249: should you throw if this is higher than 255?

krwq: you do throw actually, if so then why not just cast since it can't overflow?

krwq (approved):
overall looks good to me, there is a bit of allocations happening in some places but it's not in the bottleneck so it's fine for initial pass

src/devices/PiJuice/PiJuicePower.cs
DazFahy: Its in the original python code, https://github.com/PiSupply/PiJuice/blob/d3c193c9f40ffc2f45cdb9304cd7c59ee46952b2/Software/Source/pijuice.py#L955

src/devices/PiJuice/PiJuice.cs
DazFahy: Will do

src/devices/PiJuice/PiJuiceStatus.cs
DazFahy: I think it would make more sense to change Count from a short to a byte

krwq:
@DazFahy did you base your code on https://github.com/PiSupply/PiJuice or used spec/docs to implement it? If so was it done entirely or just some pieces? All pieces of code based on that project cannot be merged here because PiJuice is licensed on GPL which is not compatible with MIT which we ship here. cc: @richlander

Sorry to bring this up right now, I only realized that when you wrote this comment: https://github.com/dotnet/iot/pull/1213#discussion_r514208332 (and this piece of code we cannot merge here for sure)

DazFahy:
@krwq, it's a combination of both documentation and code. The documentation was the main source, but when it did not provide answers then I used the code.

krwq:
@DazFahy I do not know how to proceed with this in that case. I think the pieces which got based on code will need to be removed regardless, perhaps in the meantime you can ask authors of the PiJuice to update the docs. @richlander can you advice on what the next steps here should we so that we don't get into any legal pitfall?

DazFahy:
@krwq to be fair it will be the whole code because the only place they reference the commands is in the code, without these the code would be useless.

krwq:
@DazFahy I recommend you to open the issue in the PiJuice repo to ask them to add commands in the docs because you'd like to implement it in the non GPL compatible code and let's wait for @richlander's recommendation (although I suspect that unfortunately we will need to hold with merging this PR until are legal steps are satisfied). I think (but not sure) that reading GPL docs and implementing based on what you read is fine as long as you don't copy and paste the fragments but I'm not sure what about reading "what do I need to send" in the code.

So if I understand correctly you have not based the API shape just the internal protocol, correct?

DazFahy:
@krwq I've opened an issue in their outlining the issue and have asked them if it would be possible for them to update their documentation with the command addresses and the other bits of missing information.

The high level classes PiJuiceStatus etc are

richlander:
@krwq is giving good guidance.

It is fine to use docs, unless those docs are also GPL. It is best to ask for guidance from the people who maintain the other repo.

shawaj:
hey all, Aaron from Pi Supply here. Just wanted to step in and say hi and thanks for bringing this to our attention @DazFahy 

Our issue is here - https://github.com/PiSupply/PiJuice/issues/574

Can we just agree a one-time waiver of the GPL license with you and/or @DazFahy to enable you to publish this code under MIT License?

Happy to add to our documentation as well, and probably makes sense to do so for future, however that may take us some time and so is not particularly efficient for the purposes of getting this PR merged.

Let me know what you think @krwq @richlander

shawaj:
@krwq @richlander any thoughts on this?

krwq:
@shawaj I'm fine with this, thank you for being open to that option! We need @richlander's confirmation this is fine from legal perspective or if there is anything else we might need to do (he might need to double check with lawyers though). We do want this binding merged as well!

krwq:
@shawaj (this is my quick private research) according to https://www.revenera.com/blog/software-composition-analysis/open-source-software-licensing-not-everything-is-as-it-seems/ it suggests every contributor of the GPL project whose contribution remain in the code need to approve this (although article says about re-licensing the project not about using portions of software elsewhere so I suspect every contributor to the portion of the code being looked at would have to agree)

shawaj:
@krwq the code is here - https://github.com/PiSupply/PiJuice/blob/master/Software/Source/pijuice.py

However, all of the people who have contributed, apart from @tvoverbeek are (or were at the time) contracted employees of Pi Supply...so think it would just be Ton and Pi Supply that would need to sign off on that. Ton is a volunteer that helps us out and I am sure he would be willing to do that. And obviously I can arrange from Pi Supply side.

src/devices/PiJuice/PiJuiceStatus.cs
```diff
@@ -0,0 +1,262 @@
...
+        /// <returns>Battery current in milliamps</returns>
+        public ElectricCurrent GetBatteryCurrent()
+        {
+            var response = _piJuice.ReadCommand(PiJuiceCommand.BatteryCurrent, 2);
+
+            int i = BinaryPrimitives.ReadInt16LittleEndian(response);
+            if ((i & (1 << 15)) == (1 << 15))
+            {
+                i -= (1 << 16);
+            }
```
jstoker, line 150: Hi,
I'm just another .NET developer (unrelated to this project), and had a need for exactly this sort of library (saving me a bunch of work, thanks! 👍)
I've found when testing that ReadInt16LittleEndian appears to already have done the conversion that's being done here (resulting in a value 1<<16 less than expected).

```suggestion
```

This also exists in the GetIOCurrent() function a little bit lower down.

Thanks!

shawaj:
I am sending this as discussed via email with @richlander 

My company, [Nebra Ltd trading as Pi Supply](https://github.com/PiSupply) ([UK Company Number 06732600](https://find-and-update.company-information.service.gov.uk/company/06732600)), is the copyright holder of the [Python code](https://github.com/PiSupply/PiJuice/blob/master/Software/Source/pijuice.py) that was used to write this C# code. I grant you permission to make this code MIT.

krwq:
@richlander feel free to merge if the statement looks good, we've already signed off

src/devices/PiJuice/PiJuiceStatus.cs
krwq: @jstoker can you file an issue on that? also if you have the device to test I'd recommend to just send a PR with the fix

krwq: I've created https://github.com/dotnet/iot/issues/1333

Issue #1146, opened 2020-07-31
Enable Policheck as part of build pipeline


ericstj:
We should run Policheck on the IOT repo and fix up the things it flags.  @Anipik has the info on how to set this up.  It's semi-automated by arcade infra.

Anipik:
I can help set this up but do we want it to be part of our official builds ? we stopped doing that for the runtime repo and just run those as a part of the nightly builds and release pipeline

ericstj:
@Anipik same behavior as dotnet/runtime should be fine.  cc @joperezr @krwq

PR #1214, opened 2020-10-16
Adding sdl-tsa-vars.config for adding iot to arcade validation pipeline for running policheck

Anipik:
Fixes https://github.com/dotnet/iot/issues/1146
Copied the file from the runtime repo with repo name change and notification email change.
Bug Area path remains same bcoz i don't have appropriate rights to create a new one.

runtime copy https://github.com/dotnet/runtime/blob/master/eng/sdl-tsa-vars.config

### DIFF
--- a/eng/sdl-tsa-vars.config
+++ b/eng/sdl-tsa-vars.config
@@ -1,0 +1,11 @@
+-SourceToolsList @("policheck","credscan")
+-TsaInstanceURL https://devdiv.visualstudio.com/
+-TsaProjectName DEVDIV
+-TsaNotificationEmail joperezr@microsoft.com
+-TsaCodebaseAdmin REDMOND\danmose
+-TsaBugAreaPath "DevDiv\NET Runtime\Reliability\Docs"
+-TsaIterationPath DevDiv
+-TsaRepositoryName iot
+-TsaCodebaseName iot
+-TsaOnboard $True
+-TsaPublish $True


adiaaida:
File looks fine to me, but I'd probably reach out to folks who know how to to see if you want/need a different area path (I'd imagine yes, but I don't actually know).

pgrawehr:
Just curious: What is policheck? Do you have an example of what messages this generates?

Anipik:
> Just curious: What is policheck? Do you have an example of what messages this generates?

It scans over the code and flags bad words.

eng/sdl-tsa-vars.config
```diff
@@ -0,0 +1,11 @@
...
+-TsaInstanceURL https://devdiv.visualstudio.com/
+-TsaProjectName DEVDIV
+-TsaNotificationEmail joperezr@microsoft.com
+-TsaCodebaseAdmin REDMOND\danmose
+-TsaBugAreaPath "DevDiv\NET Runtime\Reliability\Docs"
+-TsaIterationPath DevDiv
+-TsaRepositoryName iot
+-TsaCodebaseName iot
+-TsaOnboard $True
+-TsaPublish $True
```
Ellerbach, line 11: you're most likely missing a CR at the end of the line

Ellerbach (changes requested):
I'm all find with policheck all up but you have to add as well a specific readme.MD with information on what is policheck, what it does, an example of the output and also where to find the list of "bad" words.
The added file is very "misterious" from an open source point of view as what it is pointing can't be accessed by anyone to check what it is. So minimum is to add documentation to explain all and a way to report any issue as well.

joperezr (commented):
Thanks for adding this @Anipik. File looks good to me but I do agree with @Ellerbach that if we could add some sort of documentation of what is being added it would be ideal, specifically since it is a check that might potentially break the build, so any docs around policheck and/or how to fix issues would be great 👍

adiaaida:
I don't know if this file would break the build. While this is not intuitive, this file enables us to run policheck/credscan as part of our external nightly validation pipeline (https://dev.azure.com/dnceng/internal/_build?definitionId=827), which is external to your builds. Credscan/policheck also do not fail the build when they are run if you pass -TsaPublish $True. They will open SDL issues that the team will need to investigate/fix, but it won't fail the build.

joperezr:
Got it! Thanks for the explanation

Anipik:
The policheck & credscan tools are internal and are part of a wide global readiness effort. I have reached out to the teams who are responsible for making the documentation public for these tools.

joperezr (approved):

PR #1215, opened 2020-10-17
Update GPIO library and samples to NRTs

richlander:
Updated the `System.Device.Gpio` and `samples` directories to enable nullable reference types (NRTs). The plan is to update the repo, step by step, to use NRTs (and then NRT only).

The repo builds cleanly after these changes.

The syntax required for the `!` operator when used with arrays surprised me. The analyzers that are enabled for this repo required a space where I didn't expect it. Example: https://github.com/dotnet/iot/pull/1215/files#diff-9739ebaef7d2de3f18125a9bdabe3efb1901e46a693e051d59a7283b82e8e573R404

This change deserves a thorough review. I've never applied NRT to libraries before, just to my own apps.


pgrawehr:
I'm a very novice when it comes to these nullables, but I would assume that the strange syntax is caused by lack of support for this particular scheme in StyleCop. It expects a space after a closing ], which is in most cases what you want.

joperezr:
Thanks a bunch for doing this work @richlander! As for the changes on System.Device.Gpio, I think they have to theoretically be reviewed by API review just to make sure that all annotations are correct, is that right @terrajobst ? If so, can we schedule some time slot to go over these? API surface area is small so shouldn't take much time

richlander:
Happy to adopt whatever process is desired/required. I'm quite sure that these changes could benefit from feedback. There were a few cases where I wasn't sure what to do.

For example, a really interesting case is what the right pattern is for variables that are non-null except for the dispose method. I adopted a non null pattern for that case, and then used `null!` for the dispose assignment. I figured that the object has non-deterministic behavior post-dispose, so trying to include this assignment in the general NRT model doesn't seem useful.

richlander:
I talked to the team. They pointed me to this doc on NRT: https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/api-guidelines/nullability.md

They said that I selected the right pattern for variables that are only null w/rt dispose. They also said that we should fix this line (remove the space) -- https://github.com/dotnet/iot/pull/1215/files#diff-9739ebaef7d2de3f18125a9bdabe3efb1901e46a693e051d59a7283b82e8e573R404 -- and disable style cop. They said most of our repos don't run it in CI.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -127,16 +132,16 @@ protected internal override void ClosePin(int pinNumber)
         {
             ValidatePinNumber(pinNumber);
 
-            bool isOpen = _pinModes[pinNumber] != null && _pinModes[pinNumber].InUseByInterruptDriver;
+            bool isOpen = _pinModes?[pinNumber] is object && (_pinModes[pinNumber]?.InUseByInterruptDriver ?? false);
```
krwq, line 135: since you have a null check later this probably could be:

```suggestion
            bool isOpen = _pinModes is object && (_pinModes[pinNumber]?.InUseByInterruptDriver ?? false);
```

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardInfo.cs
```diff
@@ -199,7 +196,7 @@ private Model GetBoardModel()
         /// <value>
         /// The name of the processor.
         /// </value>
-        public string ProcessorName
+        public string? ProcessorName
```
krwq, line 199: I'd vote to refactor a code a bit to make this non-null. Treat as optional considering it's internal code

krwq (approved):
LGTM with minor comments

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
richlander: Good call. That's a better pattern. Ooops. I forgot to accept your change. That wasn't intentional. Will accept next time. I very much appreciate people making "suggestions".

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardInfo.cs
richlander: I wasn't sure how to approach that one. I switched to assigning `string.Empty` as a fallback.

pgrawehr:
> They said that I selected the right pattern for variables that are only null w/rt dispose. They also said that we should fix this line (remove the space) -- [#1215 (files)](https://github.com/dotnet/iot/pull/1215/files#diff-9739ebaef7d2de3f18125a9bdabe3efb1901e46a693e051d59a7283b82e8e573R404) -- and disable style cop. They said most of our repos don't run it in CI.

I wouldn't just simply disable it - this has helped to prevent a lot of review effort just due to small formatting issues. Maybe there's even an update to solve this.

richlander:
I think this PR is ready to go.

pgrawehr:
Yes, that is an acceptable workaround. Note that you can also do that on a smaller level by putting 
```
#pragma warning disable...
// offending code
#pragma warning restore...
```

around just the lines that cause the problem. This should be done especially if the problem happens only on, say, one line of a large file.

pgrawehr:
I'll pick up the StyleCop issue later, but there's other stuff on my desk right now.

richlander:
I rebased this branch on master. This generated a lot of new NRT errors. They have all been resolved with the latest commit. My local branch build now runs cleanly.

richlander:
On the style cop issue, I'm leaving my change as-is (file-level pragma). We should put our energy towards adopting the latest unstable stylecop where this issue is resolved.

pgrawehr:
I'll pick up the stylecop issue later.

joperezr:
@richlander not sure if this is the stylecop issue you are referring too but build is broken now in Windows with:
`D:\a\1\s\src\System.Device.Gpio\System\Device\Gpio\Drivers\Windows10DriverPin.cs(79,43): error CS8600: Converting null literal or possible null value to non-nullable type. [D:\a\1\s\src\System.Device.Gpio\System.Device.Gpio.csproj]`

Are we missing a pragma there too or is it a warning we forgot to address? It is likely that you missed this locally if you didn't build for net5.0-windows10.0.17763.0

src/System.Device.Gpio/Interop/Windows/WinRT/Interop.WaitForCompletion.cs
```diff
@@ -8,7 +8,8 @@
 
 internal static partial class Interop
 {
-    public static TResult WaitForCompletion<TResult>(this IAsyncOperation<TResult> operation)
+    public static TResult? WaitForCompletion<TResult>(this IAsyncOperation<TResult> operation)
+        where TResult : class
```
joperezr, line 12: Just wondering, why did you have to add a constraint here?

richlander: It's to deal with nullables being a very different feature between classes and structs.

richlander: If you remove the constraint, you'll see the error. Would be interesting to see if you come to the same conclusion as me on the fix.

joperezr: I see yeah that's what I thought but I asked because I manually pulled down your PR branch and removed it and I was still able to build successfully, not sure why. Can you try removing it locally and rebuilding to see if you still get the warning?

richlander:
You are right, @joperezr. There were Windows errors remaining, and should now be resolved.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10Driver.cs
```diff
@@ -67,10 +67,13 @@ protected internal override void AddCallbackForPinValueChangedEvent(int pinNumbe
         /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
         protected internal override void ClosePin(int pinNumber)
         {
-            if (_openPins.TryGetValue(pinNumber, out Windows10DriverPin pin))
+            if (_openPins.TryGetValue(pinNumber, out Windows10DriverPin? pin))
             {
-                pin.ClosePin();
-                _openPins.Remove(pinNumber);
+                if (pin is object)
```
joperezr, line 72: I don't think you need this check, as we can guarantee that if TryGetVaue is true then pin won't be null. Where you getting a warning because of this?

joperezr: If you were, I would probably just use exclamation point rather than make that extra check since we know it will be defined

richlander:
I tried to recreate the constraint error. It no longer repros, so I removed that syntax.

src/System.Device.Gpio/System/Device/Gpio/Drivers/InterruptSysFsDriver.cs
```diff
@@ -23,6 +23,11 @@ public InterruptSysFsDriver(GpioDriver gpioDriver)
                 throw new PlatformNotSupportedException($"{GetType().Name} is only supported on Linux/Unix.");
             }
 
+            if (gpioDriver is null)
```
joperezr, line 26: NIT: Do you actually need this check? given gpioDriver is marked as non-nullable I think we can get rid of this, specially since we only call this internally

richlander: You are right. The constructor value cannot be null with NRTs enabled.

What happens if NRTs is not enabled, and the null is passed? Won't this code nullref?

```csharp
        protected internal override PinValue Read(int pinNumber)
        {
            return _gpioDriver.Read(pinNumber);
        }
```

joperezr: correct, except that this is an internal type so it is actually only called from within the library itself, which we are already ensuring we won't pass null. More specifically, there is only one place where we instantiate this driver, which is from a RaspberryPi3LinuxDriver, and in there we pass `this` as the driver which of course won't be null:

https://github.com/dotnet/iot/blob/be64b3e55e64401aae36004e9722001e928e387f/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs#L491

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -10,6 +10,8 @@
 using System.Threading;
 using System.Threading.Tasks;
 
+#pragma warning disable SA1011
```
joperezr, line 13: I know we've discussed about this but do you mind adding a comment above explaining briefly why we need this, as well as maybe a link to a new issue for removing it once we move to new analyzers

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -113,9 +115,14 @@ protected internal override void AddCallbackForPinValueChangedEvent(int pinNumbe
         {
             ValidatePinNumber(pinNumber);
 
-            _interruptDriver.OpenPin(pinNumber);
-            _pinModes[pinNumber].InUseByInterruptDriver = true;
+            _interruptDriver!.OpenPin(pinNumber);
```
joperezr, line 118: Now that I see this, I wonder if this is actually accurate. Seems like this will only be initialized if we called `Initialize` method which would have only happened if we called `OpenPin` which I'm not 100% sure if that will always be the case when calling `AddCallbackForPinValueChangedEvent`

joperezr: Actually nevermind, seems like we are already protected against that because the GpioController will only let you add a callback if you have also already opened the pin, so I think we are good.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -113,9 +115,14 @@ protected internal override void AddCallbackForPinValueChangedEvent(int pinNumbe
         {
             ValidatePinNumber(pinNumber);
 
-            _interruptDriver.OpenPin(pinNumber);
-            _pinModes[pinNumber].InUseByInterruptDriver = true;
+            _interruptDriver!.OpenPin(pinNumber);
+
+            if (_pinModes[pinNumber] is null)
```
joperezr, line 120: I think that because of my comment above of the GpioController already guarding against this, it is probably safe for us to remove this null check since we are already checking for it higher up in the stack.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -29,11 +31,11 @@ internal unsafe class RaspberryPi3LinuxDriver : GpioDriver
         private const string DeviceTreeRanges = "/proc/device-tree/soc/ranges";
         private const string ModelFilePath = "/proc/device-tree/model";
 
-        private readonly PinState[] _pinModes;
+        private readonly PinState?[] _pinModes;
```
joperezr, line 34: Why is this nullable? seems like we can switch it to non-nullable given we always instantiate it, and that way we can save some of the exclamation points you have to use bellow.

joperezr (approved):
Left a few comments in some places (some already addressed 😛) but this looks great overall. Thanks so much for doing this @richlander, we should probably log an issue to track making the device bindings package nullable enabled too once we are done with S.D.G

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
richlander: I also tried to track that back. The way that `_interruptDriver` is set and used was a bit confusing to me. I came to the conclusion that this `!` was sound, but it wasn't immediately obvious.

richlander:
Great feedback. As we found before, I over-annotated. I was able to remove several of the annotations, and the pragma is no longer needed. This was definitely a learning experience for me.

Unless I hear anything more, I'll merge this at lunchtime tomorrow.

richlander:
Actually, I think I mispoke. I think some of my earlier annotations were correct. Before I make any more changes, let's discuss this.

I think this line is wrong: https://github.com/dotnet/iot/pull/1215/files#diff-9739ebaef7d2de3f18125a9bdabe3efb1901e46a693e051d59a7283b82e8e573R137. What we decide about what line will decide the rest of the changes we need to make.

I think this line should assign `null` and not `null!`. As a result, the other line you commented on should be: `private readonly PinState?[] _pinModes;`. Note that the question mark after the type name is about nullability of the array element, and not of the _pinModes variable. A question mark after the brackets would be used to describe nullability of the _pinModes variable.

Make sense?

joperezr:
While your comments are absolutely correct and make sense, what I was referring to is that the only way to have that `_pinModes[x]` to be null is if Pin x is not open yet. The Code paths that are using `_pinModes[x]` in this file will get called by GpioController **after it performs a check to see if the pin is already opened**. For example, here is the code before we call the SetPinMode on the driver:

https://github.com/dotnet/iot/blob/be64b3e55e64401aae36004e9722001e928e387f/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs#L126-L129

My comment basically meant that if we know that we are already doing the check to make sure the pin is opened, we can tell the compiler on the driver side that "we know better" and that the element in the array won't be null.

richlander:
@joperezr -- I made the changes we discussed. If we merge that commit, then we can close #1230.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
richlander: Context: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2927

joperezr, line 13:
```suggestion
#pragma warning disable SA1011 // Closing square brackets should be spaced correctly https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2927
```
I would just add a comment directly in the code so that we remember to remove it when we don't need it any longer.

Issue #1193, opened 2020-09-23
Add .net5.0 asset in System.Device.Gpio


richlander:
The System.Device.Gpio package contains old-style WinRT assets. Those are no longer compatible with the 5.0 runtime. I propose we fix this in an iterative fashion:

1. Add a new `net5.0` asset that is Linux-only.
2. Add CSWinRT-generated code for the WinRT APIs that we are using.

Context: https://github.com/dotnet/runtime/issues/42478

pgrawehr:
That's allready been prepared. Also, we should definitelly keep the assembly platform-independent, as it is now.

Issue #1103, opened 2020-06-15
Rationalize WinRT dependencies


richlander:
I am working on a sample. I noticed that the WinMD assets are included when publishing for a Linux RID. I wouldn't expect that to happen. In addition, we're making [changing WinRT interop in .NET 5.0](https://github.com/dotnet/runtime/issues/37672).

Here's what I'd expect:

- No Windows-specific assets when publishing for Linux.
- Provide CSWinRT assets for .NET 5.0+

Demonstration of what I see:

```console
PS D:\git\iot\src\devices\Sn74hc595\samples> type .\Sn74hc595-test.csproj
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <RuntimeIdentifier>linux-arm64</RuntimeIdentifier>
    <SelfContained>false</SelfContained>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
    <ProjectReference Include="../Sn74hc595.csproj" />
  </ItemGroup>

</Project>
PS D:\git\iot\src\devices\Sn74hc595\samples> dir .\bin\Debug\netcoreapp3.1\linux-arm64\


    Directory: D:\git\iot\src\devices\Sn74hc595\samples\bin\Debug\netcoreapp3.1\linux-arm64

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---           6/15/2020 10:52 AM           7680 Sn74hc595.dll
-a---           6/15/2020 10:52 AM           2504 Sn74hc595.pdb
-a---           6/15/2020 10:52 AM           6810 Sn74hc595.xml
-a---           6/15/2020  1:12 PM         135944 Sn74hc595-test
-a---           6/15/2020  1:12 PM          11216 Sn74hc595-test.deps.json
-a---           6/15/2020  1:12 PM           7680 Sn74hc595-test.dll
-a---           6/15/2020  1:12 PM           1808 Sn74hc595-test.pdb
-a---           6/15/2020  1:12 PM            212 Sn74hc595-test.runtimeconfig.dev.json
-a---           6/15/2020  1:12 PM            154 Sn74hc595-test.runtimeconfig.json
-a---           6/15/2020  1:12 PM            440 Sn74hc595-test.xml
[...]

Issue #1091, opened 2020-05-28
Replace references to WinRT with Microsoft.Windows.SDK.NET


krwq:
Per https://github.com/dotnet/iot/issues/1089#issuecomment-635586758 WinRT will not have a builtin runtime support and will cause our library to likely stop working on Windows after 5.0 ships.

Per @jkoritzinsky:

> The way to migrate is to remove all references to winmds and System.Runtime.WindowsRuntime and System.Runtime.InteropServices.WindowsRuntime and instead just reference the Microsoft.Windows.SDK.NET package.

joperezr:
I tried doing this and found a blocker because of issue https://github.com/microsoft/CsWinRT/issues/338 since that Windows SDK package is not currently strong-named signed. Once they fix that issue, we can come back and continue this work

pgrawehr:
@joperezr : https://github.com/microsoft/CsWinRT/issues/338 is fixed. So you might want to give it another try.

PR #1217, opened 2020-10-19
Adding crosstargeting in order to be able to call WinRT APIs in net5.0

joperezr:
Fixes https://github.com/dotnet/iot/issues/1091
Fixes https://github.com/dotnet/iot/issues/1103
Fixes https://github.com/dotnet/iot/issues/1193

cc: @richlander @pgrawehr @Ellerbach @krwq @ericstj @terrajobst @jkoritzinsky @dsplaisted 

This PR will add cross-targeting to System.Device.Gpio package so that it will now have 3 different assets in the package:
- One netstandard2.0 asset: This will be used by consuming apps that are targeting netstandard2.0 compatible frameworks **that are not net5.0**. Nothing has changed from this asset from what we shipped before.
- One net5.0 asset: This asset will be for people targeting net5.0 that **don't intend to target Windows**. This implementation is very similar to the netstandard2.0 one, except that it will now throw platform not supported exceptions when running on Windows and trying to use types that depend on WinRT APIs.
- One net5.0-windows10.0.17763.0 asset: This asset will be for people targeting net5.0 that **want to run in Windows**. It will be using the new CSWinRT project to call into WinRT APIs and don't depend on System.Runtime.* package any longer. In order for folks to get this asset, they will need to change the target frameworks section in their projects and add `net5.0-windows10.0.17763.0` so that the package will work at runtime. **This is a breaking change**

As explained above, when running on net5.0 and targeting Windows it will be required to edit the target frameworks section of the project. This is the only way to get this working for now, and we are looking into some embedding scenarios (once CSWinRT supports them) so that we can change the way we distribute our package so that consuming projects won't require to cross-target. As part of this change, we are also adding a targets file with the package that will try to detect cases where cross-targeting is required, in order to catch those cases early and give helpful messages to consumers so that they can change their project accordingly. Here is how the package will look now:
image

And here are the package dependencies after this change:
image

Any feedback is appreciated 😄

### DIFF
--- a/src/System.Device.Gpio/System.Device.Gpio.csproj
+++ b/src/System.Device.Gpio/System.Device.Gpio.csproj
@@ -1,8 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
-    <!--<TargetFrameworks>netstandard2.0;net5.0</TargetFrameworks>
-    <TargetFrameworks Condition="$(NETCoreSdkRuntimeIdentifier.StartsWith('win'))">$(TargetFrameworks);net5.0-windows10.0.17763.0</TargetFrameworks>-->
+    <TargetFrameworks>netstandard2.0;net5.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(NETCoreSdkRuntimeIdentifier.StartsWith('win'))">$(TargetFrameworks);net5.0-windows10.0.17763.0</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <LangVersion>preview</LangVersion>
     <IsPackable>true</IsPackable>
--- a/src/System.Device.Gpio/System.Device.Gpio.csproj
+++ b/src/System.Device.Gpio/System.Device.Gpio.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net5.0;net5.0-windows10.0.17763.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net5.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(NETCoreSdkRuntimeIdentifier.StartsWith('win'))">$(TargetFrameworks);net5.0-windows10.0.17763.0</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <LangVersion>preview</LangVersion>
     <IsPackable>true</IsPackable>
--- a/src/System.Device.Gpio/System.Device.Gpio.csproj
+++ b/src/System.Device.Gpio/System.Device.Gpio.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net5.0;net5.0-windows10.0.17763.0</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <LangVersion>preview</LangVersion>
     <IsPackable>true</IsPackable>
@@ -10,17 +10,19 @@
     <Compile Remove="winmd\**" />
     <EmbeddedResource Remove="winmd\**" />
     <None Remove="winmd\**" />
+    <None Include="build\net5.0\System.Device.Gpio.targets" Pack="true" PackagePath="\build\net5.0" />
     <PackageReference Include="Microsoft.Win32.Registry" Version="$(MicrosoftWin32RegistryPackageVersion)" /> <!-- This is Windows specific -->
-    <PackageReference Include="System.Runtime.WindowsRuntime" Version="$(SystemRuntimeWindowsRuntimePackageVersion)" /> <!-- This is Windows specific -->
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.3" />
     <PackageReference Include="System.Memory" Version="4.5.3" />
-    <PackageReference Include="System.Runtime.InteropServices.WindowsRuntime" Version="4.3.0" /> <!-- This is Windows specific -->
     <PackageReference Include="Microsoft.DotNet.GenAPI" Version="$(MicrosoftDotNetGenApiPackageVersion)">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
   </ItemGroup>
   
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Runtime.WindowsRuntime" Version="$(SystemRuntimeWindowsRuntimePackageVersion)" />
+    <PackageReference Include="System.Runtime.InteropServices.WindowsRuntime" Version="4.3.0" />
+
     <Reference Include="Windows.Devices.DevicesLowLevelContract">
       <HintPath>winmd\Windows.Devices.DevicesLowLevelContract.winmd</HintPath>
       <IsWinMDFile>true</IsWinMDFile>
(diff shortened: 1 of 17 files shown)


src/System.Device.Gpio/build/net5.0/System.Device.Gpio.targets
```diff
@@ -0,0 +1,7 @@
+<Project>
+  <Target Name="EnsureProjectIsTargetingWindowsPlatformIfRequired"
+          BeforeTargets="Build"
+          Condition="'$(TargetFramework)' == 'net5.0' And $(RuntimeIdentifier.StartsWith('win'))">
```
ericstj, line 4: You only want to cause runnable applications to fail, not libraries.  It's ok for another library to target just `net5.0`.  Can you make sure this check only applies to apps?  Also it's best practice to have a way to disable such errors.  Consider adding a property to suppress this.

jkoritzinsky (approved):

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10Driver.notSupported.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Windows10Driver"/> class.
+        /// </summary>
+        public Windows10Driver()
+        {
+            // If we land in this method it means the console application is running on Windows and targetting net5.0 (without specifying Windows platform)
+            // In order to call WinRT code in net5.0 it is required for the application to target the specific platform
+            // so we throw the bellow exception with a detailed message in order to instruct the consumer on how to move forward.
+            throw new PlatformNotSupportedException("In order to use Windows10Driver on Windows with .NET 5.0 it is required for your application to target net5.0-windows10.0.17763.0. Please add that to your target frameworks in your project file.");
```
ericstj, line 22: Can you factor this to a common resource string with parameter (`{0}`) for type name to avoid duplication?  Perhaps you might also want to parameterize that windows version?  Not sure how you plan to keep that in sync with TargetFramework.

src/System.Device.Gpio/build/net5.0/System.Device.Gpio.targets
```diff
@@ -0,0 +1,7 @@
...
+    <Error Text="In order to use System.Device.Gpio package when running on Windows and targeting net5.0, it is required to add 'net5.0-windows10.0.17763.0' to your project's target frameworks." />
```
dsplaisted, line 5: It doesn't have to be exactly 17763, it could be that version or higher.  Also, I'm not sure if the RuntimeIdentifier will necessarily be set in all scenarios where you'd want this check.

joperezr: I'll change the message to say that 17763 is just the lower bound. What would you recommend instead of using RuntimeIdentifier? I know that we might still have some cases where we won't be able to catch problems, like for example shared framework applications, but that's why for those we are relying on the PlatformNotSupported exception messages to help figure out how to fix the targeting. Of course, if there are better ways to catch it at compile time instead of runtime that would be much prefered.

joperezr:
@dsplaisted looking at the CI failures seemed like the problem is that the net5.0-windows build was failing on Linux and OSX as apparently, you can only build for that TFM with a Windows SDK. Is that by design or will support for targeting windows from a non-windows machine will be added eventually? I just added a change that disables Windows-specific build on non-Windows but this scenario seems kind of busted as in theory you would only need the Windows targeting pack and we have supported in the past targeting TFMs that might not necessarily run on the machine you are in, this also means for us that our Linux and OSX contributors won't be able to make changes and build windows specific files.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10Driver.notSupported.cs
joperezr: The repo doesn't currently use ResourceManager and it is in our plans to add that support, but ideally, when added I would rather do it to the whole library and not just this string. I can centralize the format of this string for now as some sort of constant format, and then for vNext work on moving everything into resources instead, does that sound good?

src/System.Device.Gpio/build/net5.0/System.Device.Gpio.targets
dsplaisted: Good point about not generating the error for libraries.  Relatedly, you probably want to use the `buildTransitive` folder instead of the `build` folder so that if an app transitively references this package it will get the build error.

ericstj: Also per offline conversation you might be able to use `AppHostRuntimeIdentifier` to catch the case where someone is building a framework-dependent app on Windows.

dsplaisted:
> @dsplaisted looking at the CI failures seemed like the problem is that the net5.0-windows build was failing on Linux and OSX as apparently, you can only build for that TFM with a Windows SDK. Is that by design or will support for targeting windows from a non-windows machine will be added eventually? I just added a change that disables Windows-specific build on non-Windows but this scenario seems kind of busted as in theory you would only need the Windows targeting pack and we have supported in the past targeting TFMs that might not necessarily run on the machine you are in, this also means for us that our Linux and OSX contributors won't be able to make changes and build windows specific files.

It's by design for now.  Hopefully when we have optional workload support we'll eventually make the Windows optional workload work on other operating systems.

If you just want to use the types, you may be able to add an `Update` statement to set the `IsWindowsOnly` metadata to false for the `KnownFrameworkReference` items for Windows.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10Driver.notSupported.cs
ericstj: Even though you aren't using ResX here you can still de-dup the string to a single helper method.

joperezr: yeah, that's what I meant with centralizing the format. I'll do that then.

joperezr:
> If you just want to use the types, you may be able to add an Update statement to set the IsWindowsOnly metadata to false for the KnownFrameworkReference items for Windows.

I actually expect very few people to actually want to change things on the Windows-specific build for now, so I'm fine with just not building it on Linux and OSX at the moment, but my 2 cents here are just that it would be great to either document the `isWindowsOnly` metadata workaround for other folks or to allow people to target different platform than the one they run in.

src/System.Device.Gpio/build/net5.0/System.Device.Gpio.targets
joperezr: If I do that, then folks creating framework-dependent apps on Windows with the intention of running them on Linux will get the error, which I actually believe will be one of the most common scenarios. Would it be ok to not protect against that case and instead let the runtime exception catch such scenarios given they should, in theory, be the least common case?

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10Driver.notSupported.cs
ericstj: Yep, sorry I had replied here before seeing your reply.  GH didn't refresh for some reason.

src/System.Device.Gpio/build/net5.0/System.Device.Gpio.targets
ericstj: > creating framework-dependent apps on Windows with the intention of running them on Linux

We could teach these folks to use `AppHostRuntimeIdentifier` but maybe that's overkill.  I'm fine with whatever you think is best here.  We could even forgo the whole targets if we don't think it'll help enough to warrant the added complexity.

joperezr: I think that as long as it catches some of the cases it is worth keeping, as a build time error is always preferred over runtime error, but just want to make sure that we don't catch false-positives by making our safety net too general.

src/System.Device.Gpio/build/net5.0/System.Device.Gpio.targets
pgrawehr: Not sure this should cause an error at all. Even if I'm running on windows, I might not need to use the windows specific implementation, especially if I'm not using the WindowsGpioDriver (but maybe only an Arduino driver).

joperezr: While most (if not all) of this library would render useless if we can call WinRT in Windows, I do agree with you that perhaps Error is too aggressive even though it is suppressable. @ericstj do you think switching to Warning would be appropriate here?

pgrawehr:
I do want to fully understand what's going on here (so far I honestly don't get it yet). Please do not merge before I can take an adequate review.

joperezr:
@pgrawehr see the description where I added info on the three assets distributed by the package and when would each apply. In there I provide a high level description of what is changing with this PR, but if you have any follow up questions, I’m happy to answer.

src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -35,4 +38,27 @@
...
+    <Compile Remove="System\Device\Gpio\Drivers\Windows10Driver.notSupported.cs" />
+    <Compile Remove="System\Device\I2c\I2cDevice.nonWindows.cs" />
+    <Compile Remove="System\Device\Pwm\PwmChannel.nonWindows.cs" />
+    <Compile Remove="System\Device\Spi\SpiDevice.nonWindows.cs" />
+    <Compile Remove="System\Device\CommonHelpers.cs" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net5.0'">
+    <!--Remove Windows Code-->
+    <Compile Remove="Interop\Windows\WinRT\*.cs" />
```
pgrawehr, line 52: I think the PR lacks files. This directory is empty for me (except for "Interop.WaitForCompletion.cs", which only contains a helper method). 

Besides, I don't understand why "net5.0" or not would have anything to do with Windows or not.

joperezr: > I think the PR lacks files. This directory is empty for me (except for "Interop.WaitForCompletion.cs", which only contains a helper method).

I know this is the only WinRT file today, but I used wildcards as any file that goes in it in the future would have to be removed, and because everything is added to all configurations by default, I wanted to make sure that we guarded against new files or renames.

> Besides, I don't understand why "net5.0" or not would have anything to do with Windows or not.

Our net5.0 configuration is effectively a non-Windows build since we want everyone targeting net5.0 and Windows to use the net5.0-windows version instead. That is why net5.0 version now throws anytime you are trying to use anything that relies on WinRT code.

src/System.Device.Gpio/build/net5.0/System.Device.Gpio.targets
joperezr: Chatted offline, I'll keep as error for now since we want to make sure that users won't miss it, but keep it suppressible so that users can disable if they think they are a false positive.

joperezr:
I'll merge this now to unblock progress on getting the repo ready for shipping our 1.1 release but feel free to let me know if there are any additional comments/questions around this change and I'm happy to address/answer them.

PR #1218, opened 2020-10-20
Adding device binding for AMG88xx infrared array sensor family

fehdem:
This PR adds a device binding for the AMG88xx infrared array sensor family. 
(Second attempt after fixing GIT branching issue. Initial PR closed).


src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+        }
+        #endregion
+
+        #region Operating Mode / Power Control
+
+        /// <summary>
+        /// Gets the current operating mode
+        /// </summary>
+        /// <returns>Operating mode</returns>
+        public OperatingMode GetOperatingMode()
```
Ellerbach, line 260: Same `OperatingMode` as a property would be perfect as well

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+        }
+        #endregion
+
+        #region Reset
+
+        /// <summary>
+        /// Performs an initial reset of the sensor. The flags and all configuration registers
+        /// are reset to default values.
+        /// </summary>
+        public void InitialReset()
```
Ellerbach, line 281: Maybe just `Rest`? would be consistent with other bindings

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+
+        #endregion
+
+        #region Interrupt control, levels and pixel flags
+
+        /// <summary>
+        /// Gets the interrupt mode
+        /// </summary>
+        /// <returns>Interrupt mode</returns>
+        public InterruptMode GetInterruptMode()
```
Ellerbach, line 305: Would go for a property as well `InterruptMode`

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+        public void SetInterruptMode(InterruptMode mode)
+        {
+            SetBit(Register.INTC, (byte)InterruptModeBit.INTMODE, mode == InterruptMode.AbsoluteMode);
+        }
+
+        /// <summary>
+        /// Enables the interrupt pin of the AMG88xx sensor.
+        /// The pin is pulled down if an interrupt is active.
+        /// </summary>
+        public void EnableInterruptPin()
```
Ellerbach, line 323: Same here. `public bool InterruptPin` with the get and set function

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+        public void DisableInterruptPin()
+        {
+            SetBit(Register.INTC, (byte)InterruptModeBit.INTEN, false);
+        }
+
+        /// <summary>
+        /// Gets the lower level interrupt temperature
+        /// </summary>
+        /// <returns>Temperature level</returns>
+        public Temperature GetInterruptLowerLevel()
```
Ellerbach, line 341: Same here, property would be great `Temperature InterruptLowerLevel`

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+            (byte tl, byte th) = Amg88xxUtils.ConvertFromTemperature(temperature);
+            SetRegister(Register.INTLL, tl);
+            SetRegister(Register.INTLH, th);
+        }
+
+        /// <summary>
+        /// Gets the upper level interrupt temperature
+        /// </summary>
+        /// <returns>Temperature level</returns>
+        public Temperature GetInterruptUpperLevel()
```
Ellerbach, line 363: +1 for this one as well :-) `InterruptUpperLevel`

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+            (byte tl, byte th) = Amg88xxUtils.ConvertFromTemperature(temperature);
+            SetRegister(Register.INTHL, tl);
+            SetRegister(Register.INTHH, th);
+        }
+
+        /// <summary>
+        /// Gets the hysteresis level interrupt temperature
+        /// </summary>
+        /// <returns>Temperature level</returns>
+        public Temperature GetInterruptHysteresisLevel()
```
Ellerbach, line 385: Same here, property

src/devices/Amg88xx/README.md
```diff
@@ -0,0 +1,164 @@
+﻿# AMG88xx Infrared Array Sensor Family
```
Ellerbach, line 1: Thanks you for the very detailed and comprehensive readme!

src/devices/Amg88xx/Register.cs
```diff
@@ -0,0 +1,953 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Iot.Device.Amg88xx
+{
+    /// <summary>
+    /// Defines the addresses of the AMG88xx registers
+    /// </summary>
+    public enum Register : byte
```
Ellerbach, line 12: Any reason why you make the Register public? Should it be internal? I don't really see any method using it.

src/devices/Amg88xx/Register.cs
```diff
@@ -0,0 +1,953 @@
...
+        /// Pixel 64 value register (high byte)
+        /// </summary>
+        T64H = 0xff
+    }
+
+    /// <summary>
+    /// Defines the status bits of the status register (addr: 0x04)
+    /// </summary>
+    [Flags]
+    public enum StatusFlagBit : byte
```
Ellerbach, line 814: Same for this one.
Also, would be good to have separated file for each enum

src/devices/Amg88xx/Register.cs
```diff
@@ -0,0 +1,953 @@
...
+        /// revision of the sensor.
+        /// </summary>
+        OVF_THS = 3
+    }
+
+    /// <summary>
+    /// Defines the status bits of the status clear register (addr: 0x05)
+    /// </summary>
+    [Flags]
+    public enum StatusClearBit : byte
```
Ellerbach, line 839: same here

src/devices/Amg88xx/Register.cs
```diff
@@ -0,0 +1,953 @@
...
+        /// <summary>
+        /// Thermistor output overflow occured
+        /// </summary>
+        OVFTHCLR = 3
+    }
+
+    /// <summary>
+    /// Defines the bit(s) of the moving average mode register (addr: 0x07)
+    /// </summary>
+    public enum MovingAverageModeBit : byte
```
Ellerbach, line 860: And here and all the other enums

src/devices/Amg88xx/tests/Amg88xxTests.cs
```diff
@@ -0,0 +1,629 @@
...
+
+using System;
+using Iot.Device.Amg88xx;
+using UnitsNet;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Iot.Device.Amg88xx.Tests
+{
+    public class Amg88xxTests
```
Ellerbach, line 13: Thanks for the Tests as well!

src/devices/Amg88xx/tests/Amg88xxTests.cs
```diff
@@ -0,0 +1,629 @@
...
+                {
+                    Assert.Equal(col == expectedColumn - 1 && row == expectedRow - 1, flags[col, row]);
+
+                }
+            }
+        }
+
+        #endregion
+    }
+}
```
Ellerbach, line 629: you're missing a CRLF

src/devices/Amg88xx/tests/Amg88xxTests.cs
```diff
@@ -0,0 +1,629 @@
...
+            {
+                Assert.Equal(address, i2cDevice.DataWritten.Dequeue());
+            }
+
+            for (int col = 0; col < Amg88xx.Columns; col++)
+            {
+                for (int row = 0; row < Amg88xx.Rows; row++)
+                {
+                    Assert.Equal(col == expectedColumn - 1 && row == expectedRow - 1, flags[col, row]);
+
```
Ellerbach, line 622: extra enter

src/devices/Amg88xx/tests/Amg88xxUtilsTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        [InlineData(-25, 0x9c, 0x0f)]
+        [InlineData(-55, 0x24, 0x0f)]
+        public void ConvertFromTemperatureTest(double temperature, byte expectedTl, byte expectedTh)
+        {
+            (byte tl, byte th) = Amg88xxUtils.ConvertFromTemperature(Temperature.FromDegreesCelsius(temperature));
+            Assert.Equal(expectedTl, tl);
+            Assert.Equal(expectedTh, th);
+        }
+    }
+}
```
Ellerbach, line 39: missing enter

src/devices/Amg88xx/tests/I2cTestDevice.cs
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+
+        public override void WriteRead(ReadOnlySpan<byte> writeBuffer, Span<byte> readBuffer)
+        {
+            Write(writeBuffer);
+            Read(readBuffer);
+        }
+    }
+}
```
Ellerbach, line 55: missing enter

src/devices/Amg88xx/tests/Amg88xx.Tests.csproj
```diff
@@ -0,0 +1,21 @@
...
+  
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../Amg88xx.csproj" />
+
```
Ellerbach, line 15: extra enter

src/devices/Amg88xx/samples/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+**Note:** There are also boards available with additional interfaces or even with an integrated Arduino or compatible circuit. You can use this binding only if the boards gives you access to the I2C interface only.
+
+
+## Wiring
+For demonstration purpose the INT-pin of the sensor is connected to GPIO PIN 5 of the RPi. Additionally an LED is connected to GPIO PIN 6 of the RPi. The LED signals the occurrence of an interrupt.
+The resistor depends on the LED type; however 150R-220R is a safe choice for a standard red LED. Or precisely: R = (3,3V -  U LED,forward) / I LED, forward
+
+![Wiring of a sensor breakout and LED for the sample](./AMG88xxSample.png)
+
```
Ellerbach, line 20: extra enter

src/devices/Amg88xx/samples/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## References
+You should refer to the README.md of the binding for more useful information as well as to the reference specification of the AMG88xx sensor family.
+
```
Ellerbach, line 24: extra enter

src/devices/Amg88xx/samples/README.md
```diff
@@ -0,0 +1,28 @@
...
+The sample application demonstrates the key functions of the sensor and the binding
+* thermal image readout
+* interrupt triggering based on temperature levels incl. hysteresis
+* sensor states
+* noise reduction by using the sensor's moving average function
+
+There are AMG88xx breakout boards available from a variety of vendors. You can use any of them as long as it provides access to the I2C interface of the sensor.
+
+**Note:** There are also boards available with additional interfaces or even with an integrated Arduino or compatible circuit. You can use this binding only if the boards gives you access to the I2C interface only.
+
```
Ellerbach, line 13: extra enter

src/devices/Amg88xx/samples/Amg88xx.Sample.cs
```diff
@@ -0,0 +1,131 @@
...
+
+using System;
+using System.Device.Gpio;
+using System.Device.I2c;
+using System.Threading;
+using UnitsNet;
+
+namespace Iot.Device.Amg88xx.Samples
+{
+    /// <summary>
```
Ellerbach, line 13: Thanks as well for the very detailed sample

Ellerbach (commented):
Thanks for the nice adition. Most comments are about adjusting some functions to properties and about the enums that should be split over different files and internal and few cosmetic adjustments.

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+
+        #endregion
+
+        #region Moving average
+
+        /// <summary>
+        /// Get the state of the moving average mode
+        /// </summary>
+        /// <returns>True, if moving average mode is on</returns>
+        public bool GetMovingAverageModeState()
```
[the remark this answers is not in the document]
fehdem: @Ellerbach: I don't think that a property is the right chioce for this, because the reading the current state doesn't work with all revisions of the sensor. The current datasheet doesn't say that the state can be read while older revisions can. At least for the device that I used for testing I can confirm that reading the state isn't possible.
From my point of view a propery should always return the value that has been set using the same property. 
I also don't think that it would be a good idea trackig the set state in the binding as this would imply that I would make an assumption about the default state. This may change in a future revision of the device or there might be revisions out there which actually allow reading the current state. All this could end up in inconcsistency. 
That's the reason why I chose a Get-/Set-API. Waat's your advice?

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+
+        #endregion
+
+        #region Frame Rate
+
+        /// <summary>
+        /// Get the current frame rate.
+        /// </summary>
+        /// <returns>Frame rate (either 1 or 10fps) </returns>
+        public int GetFrameRate()
```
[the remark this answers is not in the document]
fehdem: @Ellerbach: For the sake of consistency the the moving average I made this a Get-/Set-API. Would you change this to a property even if MovingAverageModeState would be a Get-/Set-API?

src/devices/Amg88xx/tests/Amg88xx.Tests.csproj
fehdem: Done

src/devices/Amg88xx/samples/README.md
fehdem: Done

src/devices/Amg88xx/samples/README.md
fehdem: Done

src/devices/Amg88xx/samples/README.md
fehdem: Done

src/devices/Amg88xx/samples/Amg88xx.Sample.cs
fehdem: You're welcome!

src/devices/Amg88xx/tests/Amg88xx.Tests.csproj
fehdem: done

src/devices/Amg88xx/samples/README.md
fehdem: done

src/devices/Amg88xx/samples/README.md
fehdem: done

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+        /// Standard device address
+        /// (AD_SELECT pin is low, c.f. reference specification, pg. 11)
+        /// </summary>
+        public const int DefaultDeviceAddress = 0x68;
+
+        /// <summary>
+        /// Alternative device address
+        /// (AD_SELECT pin is high, c.f. reference specification, pg. 11)
+        /// </summary>
+        public const int AlternativeDeviceAddress = 0x69;
```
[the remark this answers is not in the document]
fehdem: done

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+    /// <summary>
+    /// Binding for the AMG88xx family of infrared array sensors
+    /// </summary>
+    public class Amg88xx : IDisposable
+    {
+        /// <summary>
+        /// Standard device address
+        /// (AD_SELECT pin is low, c.f. reference specification, pg. 11)
+        /// </summary>
+        public const int DefaultDeviceAddress = 0x68;
```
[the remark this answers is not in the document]
fehdem: Done

src/devices/Amg88xx/Amg88xx.cs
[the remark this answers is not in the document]
Ellerbach: OK, then fair enough to keep it like this with the Get and Set. And you should add a comment about fact it may not run on al revisions properly as well.

src/devices/Amg88xx/Amg88xx.cs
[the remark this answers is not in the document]
Ellerbach: This one seems closer to a property. At least my understanding. Now is there are only 2 choices for the framerate, why not having a public enum for the 2 possible options? 1 and 10 fps?

src/devices/Amg88xx/FrameRateBit.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Amg88xx
+{
+    /// <summary>
+    /// Defines the bit of the frame rate register (addr: 0x02)
+    /// </summary>
+    public enum FrameRateBit : byte
```
Ellerbach, line 10: So you may use this one for the FrameRate potential property and have something like
```csharp
public enum FrameRateBit : byte
{
   Rate10FramePerSecond = 0,
   Rate1FramePerSecond = 1,
}
```

src/devices/Amg88xx/InterruptModeBit.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Amg88xx
+{
+    /// <summary>
+    /// Defines the interrupt controls bits of the interrupt control register (addr: 0x03)
+    /// </summary>
+    public enum InterruptModeBit : byte
```
Ellerbach, line 10: Does this one needs to be public?
If yes, then use proper naming convention

src/devices/Amg88xx/MovingAverageModeBit.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Amg88xx
+{
+    /// <summary>
+    /// Defines the bit(s) of the moving average mode register (addr: 0x07)
+    /// </summary>
+    public enum MovingAverageModeBit : byte
```
Ellerbach, line 10: Does this one needs to be public?
If yes, then use proper naming convention

src/devices/Amg88xx/OperatingMode.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// sleep mode
+        /// </summary>
+        Sleep = 0x10,
+
+        /// <summary>
+        /// stand-by mode, 10s intermittence
+        /// </summary>
+        StandBy10 = 0x21,
```
Ellerbach, line 25: maybe StandBy10Seconds same for the 60 one?

src/devices/Amg88xx/Register.cs
Ellerbach: This seems to be not need to be exposed in the public surface. Internal should be enough

src/devices/Amg88xx/ResetType.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Iot.Device.Amg88xx
+{
+    /// <summary>
+    /// Defines the reset types of the reset register (addr: 0x01)
+    /// </summary>
+    public enum ResetType : byte
```
Ellerbach, line 10: same comment as for all enums, does it need to be publically exposed?

src/devices/Amg88xx/StatusClearBit.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+
+namespace Iot.Device.Amg88xx
+{
+    /// <summary>
+    /// Defines the status bits of the status clear register (addr: 0x05)
+    /// </summary>
+    [Flags]
+    public enum StatusClearBit : byte
```
Ellerbach, line 13: same as everywhere for the enums, I don't think it should be exposed as you have methods for that. If it needs to be exposed, then use proper naming convention

src/devices/Amg88xx/StatusFlagBit.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System;
+
+namespace Iot.Device.Amg88xx
+{
+    /// <summary>
+    /// Defines the status bits of the status register (addr: 0x04)
+    /// </summary>
+    [Flags]
+    public enum StatusFlagBit : byte
```
Ellerbach, line 13: +1 for visibility

Ellerbach (commented):
Comments mainly on visibility of the enums

src/devices/Amg88xx/Amg88xx.cs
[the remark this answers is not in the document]
fehdem: @Ellerbach Yes, I do generally agree that this could be a property. I was just thinking about consistency. As Get/SetMovingAverageModeState can't be a property I thought it would be "cleaner" to keep all a discrete getter/setter combinations. However, the majority of functions could be expressed as a property.
So I leave it your choice - consistency or properties where possible?

src/devices/Amg88xx/ResetType.cs
fehdem: @Ellerbach The only reason for exposing the enums is that I use them in the unit tests and the unit tests have a different namespace. Any advice?

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+            }
+
+            return image;
+        }
+
+        /// <summary>
+        /// Gets the current raw thermal image from the sensor.
+        /// </summary>
+        /// <returns>Thermal image in two's complement representation</returns>
+        public int[,] GetThermalRawImage()
```
krwq, line 93: should this be Uint16/Int16?

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+            _i2cDevice.Read(buffer);
+
+            var image = new int[Columns, Rows];
+
+            int idx = 0;
+            for (int r = 0; r < Rows; r++)
+            {
+                for (int c = 0; c < Columns; c++)
+                {
+                    byte tl = buffer[idx++];
```
krwq, line 108: would be better to use BinaryPrimitives

src/devices/Amg88xx/Amg88xx.cs
krwq, line 93: since you're already using Span consider allowing for users to pass the buffer in and also making Stride public

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+
+        #endregion
+
+        #region Status
+
+        /// <summary>
+        /// Gets the temperature overflow flag from the status register
+        /// </summary>
+        /// <returns>Temperature overflow flag</returns>
+        public bool HasTemperatureOverflow()
```
krwq, line 141: can you make the docs clearer about when is this useful? Does this mean internal sensor is overheated or that temperature of measured thing is too high/low? The implementation detail (overflow flag) is not very useful from user perspective

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+        {
+            return GetBit(Register.STAT, (byte)StatusFlagBit.OVF_IRS);
+        }
+
+        /// <summary>
+        /// Clears the temperature overflow flag in the status register
+        /// </summary>
+        public void ClearTemperatureOverflow()
```
krwq, line 149: consider collapsing this and previous method into one and doing the clearing automatically, start with writing a sample showing the usage and based on that sample figure out if it makes more sense to have one or two methods.

Also if you keep two methods make sure to describe this scenario more. When should user call this? What happens if they don't (i.e. will reading temperature not work at all?)

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+        }
+
+        /// <summary>
+        /// Gets the thermistor overflow flag from the status register.
+        /// Note: the bit is only menthioned in early versions of the reference specification.
+        /// It is not clear whether this is a specification error or a change in a newer
+        /// revision of the sensor.
+        /// </summary>
+        /// <returns>Thermistor overflow flag</returns>
+        public bool HasThermistorOverflow()
```
krwq, line 162: how is this different from TemperatureOverflow? Can you describe the docs more? Also consider collapsing this and next method into one

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+        {
+            // only the bit to be cleared is set, the other bits need to be 0
+            SetRegister(Register.SCLR, 1 << (byte)StatusClearBit.OVFTHCLR);
+        }
+
+        /// <summary>
+        /// Gets the interrupt flag from the status register
+        /// </summary>
+        /// <returns>Interrupt flag</returns>
+        public bool HasInterrupt()
```
krwq, line 180: when and how should users use this? Should you have an API to make this usage better instead of making this public directly? I.e. WaitForXYZAsync with while loop waiting for interrupt and and clearing it?

src/devices/Amg88xx/Amg88xx.cs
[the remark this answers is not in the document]
krwq: Consider calling it `UseMovingAverage` or similar, `MovingAverageModeState` suggests it will return some kind of enum

src/devices/Amg88xx/Amg88xx.cs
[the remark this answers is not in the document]
krwq: If you accept only few values (i.e. 1 or 10) consider creating an enum (Amg88xxFrameRate or similar)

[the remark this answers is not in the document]
krwq: (you should still check for OutOfRange though)

src/devices/Amg88xx/Amg88xx.cs
krwq: I agree, we usually call it Reset or SoftwareReset

src/devices/Amg88xx/Amg88xx.cs
```diff
@@ -0,0 +1,475 @@
...
+        public void InitialReset()
+        {
+            // an initial reset (factory defaults) is initiated by writing 0x3f into the reset register (RST)
+            SetRegister(Register.RST, (byte)ResetType.Initial);
+        }
+
+        /// <summary>
+        /// Performs a reset of all flags
+        /// </summary>
+        public void FlagReset()
```
krwq, line 290: Ideally this should not be public as is, either rename this to something which suggest why user would want to call this or use it internally automatically when you detect some erroneous condition

src/devices/Amg88xx/Amg88xxUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+// See the LICENSE file in the project root for more information.
+
+using UnitsNet;
+
+namespace Iot.Device.Amg88xx
+{
+    /// <summary>
+    /// This class contains utilities for working with AMG88xx devices.
+    /// </summary>
+    public static class Amg88xxUtils
```
krwq, line 12: should this be internal?

src/devices/Amg88xx/FrameRateBit.cs
krwq: I agree, and for values either use what sensor uses internally or logically the same value (so Rate1FramePerSecond = 1, Rate10FramePerSecond = 10)

src/devices/Amg88xx/MovingAverageModeBit.cs
krwq: doesn't seem like it should be to me

src/devices/Amg88xx/ResetType.cs
krwq: for testing you shouldn't have need to use it at all but if you strongly feel you should or makes it much easier then you can:
- InternalsVisibleToAttribute
- include source files directly in the tests

you should never design APIs based on the tests and you should only think about what is useful for users

src/devices/Amg88xx/tests/Amg88xxTests.cs
```diff
@@ -0,0 +1,629 @@
...
+            sensor.ClearInterrupt();
+
+            Assert.Equal(2, i2cDevice.DataWritten.Count);
+            Assert.Equal((byte)Register.SCLR, i2cDevice.DataWritten.Dequeue());
+            Assert.Equal(1 << (byte)StatusClearBit.INTCLR, i2cDevice.DataWritten.Dequeue());
+        }
+
+        [Fact]
+        public void ClearAllFlagsTest()
```
krwq, line 212: i.e. if APIs were designed a bit differently you could test for i.e.:
`OverflowTriggeredARetry`, this way user don't need to know about overflow at all (treat this as an example you might have a better idea about how to use this properly)

krwq:
@fehdem thanks for the PR I'm excited you're adding this sensor and extra points for the tests!

src/devices/Amg88xx/ResetType.cs
fehdem: @krwq I strongly agree to 
> never design APIs based on what you need for testing

I can't agree to 
> for testing you shouldn't have need to use it 

I don't want to duplicate enums for testing or work with any kind of "magic" number in my tests. An enum defining registers or bits is a static entity and usually won't change at any time in future,  As they're rather simple, too, I just do a code review and use them in both application and test. At least this is my experience.  
So, probably the InternalsVisibleToAttribute approach would be a good idea.

src/devices/Amg88xx/InterruptModeBit.cs
fehdem: Can be internal, though renamed to InterrruptModeBits, too.
I use the InternalsVisibleToAttribute to expose it to the unit test assembly.

src/devices/Amg88xx/MovingAverageModeBit.cs
fehdem: Can be internal, though renamed to MovingAverageModeBits, too.
I use the InternalsVisibleToAttribute to expose it to the unit test assembly.

src/devices/Amg88xx/StatusFlagBit.cs
fehdem: Can be internal. I use the InternalsVisibleToAttribute to expose it to the unit test assembly.

src/devices/Amg88xx/StatusClearBit.cs
fehdem: Can be internal. I use the InternalsVisibleToAttribute to expose it to the unit test assembly.

src/devices/Amg88xx/FrameRateBit.cs
fehdem: @Ellerbach @krwq This enum does not define the frame rate itself. It defines the bit position in the corresponding register (as all the other xxxBit enums). So your suggestion doesn't work. However, it could be reasonable to define an additional enum for the actual frame rate setting, as you suggested.

src/devices/Amg88xx/Amg88xx.cs
fehdem: @krwq I'm uncertain about this. The pixel value is a 12-bit two's complement (I should add the 12-bits information to the document). So this is a signed value. However, it feels semantically imprecise to me using either UInt16 or Int16. The '16' gives a wrong hint, from my point of view. That's the reason why I chose the rather "neutural" int-type.
Am I wrong with this?

src/devices/Amg88xx/Amg88xx.cs
fehdem: Done

src/devices/Amg88xx/Amg88xx.cs
fehdem: @krwq Same as for the `HasTemperatureOverflow` method... I don't know much about its use in applications. I can only tell that this is a kind of error indicator which needs to be reset manually. An automatic reset doesn't make sense to me. 
I'll extend the documentation accordingly.

src/devices/Amg88xx/Amg88xxUtils.cs
fehdem: Yes. Done.

src/devices/Amg88xx/tests/Amg88xxTests.cs
fehdem: @krwq As stated in my comment for HasTemperatureOverflow. If there is any use for the overflow indication it would be up to the user to deal with it. The sensor will still work as usual, if the indicator is not reset. It doesn't require a retry. It is a kind of "history" information that can be evaluated by the application. 
I don't want to make any assumptions about its actual use as I don't know the requirements of the sensor users.

src/devices/Amg88xx/Amg88xx.cs
fehdem: Done

src/devices/Amg88xx/Amg88xx.cs
fehdem: @krwq I actually thought about this, but I rather prefer the current API. If I would expect the user to provide a Span I would have to reveal that each pixel requires 2 bytes. From my point of view this is a technical detail the user shouldn't need to know. Can you agree to this?

src/devices/Amg88xx/OperatingMode.cs
fehdem: Done

src/devices/Amg88xx/ResetType.cs
Ellerbach: > I don't want to duplicate enums for testing or work with any kind of "magic" number in my tests

I guess it's not about replicating, it's about direct file reference rather than project reference. So you get access to the internal elements with a compatible namespace.

> InternalsVisibleToAttribute approach would be a good idea.

One or the other approach should work. Looking forward to see which one you choose :-)

joperezr: Even though we usually don’t encourage the use of InternalsVisibleTo, in this case it is totally fine specially because we never end up shipping the individual binding assembly but instead ship the all up iot.device.bindings package which won’t have the attribute

src/devices/Amg88xx/Amg88xx.cs
fehdem: I changed this into a property.

Ellerbach: Good. Any new push of your code to see all the changes you've made so far?

src/devices/Amg88xx/Amg88xx.cs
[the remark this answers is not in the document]
fehdem: I changed this into a property, too, to keep consistency across all the properties related to sensor configuration. I added a comment regarding the quirky behaviour, though, as reading the property may result always ```false``` even if it has just been set to ```true```.

src/devices/Amg88xx/Amg88xx.cs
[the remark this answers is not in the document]
fehdem: I changed this to a property, too.

src/devices/Amg88xx/Amg88xx.cs
fehdem: @Ellerbach I just pushed it, though, I didn't complete all remarks, yet. A few are still to do or to think about. Maybe you can comment on some of the open questions I commented.

src/devices/Amg88xx/Amg88xx.cs
fehdem: @krwq I think you suggest a slightly higher level of abstraction from the sensor details. 
I still see a value in making the "pure" interrupt information available to the user as this allows to check at any time whether the sensor has seen a temperature which exceeded the configured lower or uppter threshold of any pixel. 
Though, I'll think about adding some "semantics" to the API naming. 
I'll also think about adding a WaitForSomethingAsync in addition.

src/devices/Amg88xx/Amg88xx.cs
fehdem: I renamed the method to ResetAllFlags and added further documentation describing when the user wants to use this. 
I kept it public as there is no reasonable automatism, because the use of a flag reset depends on the actual use case, from my point of view.

src/devices/Amg88xx/Amg88xx.cs
krwq: why does it feel imprecise? Int16 allows you to store 15 bits + 1 sign bit and you got only 12. Also you say before you need 2 bytes per color

krwq: @fehdem yes and no, it is a technical detail but all graphics library reveal this detail for perf reasons. You will allocate much more data the way you have it right now. Other 2 alternatives which allow you to avoid allocations on each frame and allow similar APIs:
- create a struct/class (i.e. AMG88xxImage) which has a `byte[]` internally which you can pass in here and will be filled in this method and also have indexer for that class to hide internal data format
- inside of this class have a `byte[]` which can store the whole image, have a method Read/ReadImage or similar which does not return anything (or perhaps bool if it can fail but then you call it TryRead) and have an indexer here implemented i.e.:

```csharp
public class MyClass
{
    public const int Width = 3;
    public const int Height = 5;
    private const int BytesPerPixel = 2;
    private byte[] _bytes = new byte[BytesPerPixel * Width * Height];

    private UInt16 ConvertToTemperature(ReadOnlySpan<byte> rawTemperature)
    {
        // whatever the format for reading is, consider inlining this method
        // if this is exactly as below then consider using ReadOnlySpan<UInt16> directly
        // in the previous method and not use this at all
        return BinaryPrimitives.ReadUInt16LittleEndian(rawTemperature);
    }

    public void Read()
    {
        // Fill the _bytes here, something along the lines of:
        // _i2cDevice.Read(_bytes);
    }

    public UInt16 this[int x, int y]
    {
        get
        {
            if (x < 0 || x >= Width)
                throw new ArgumentOutOfRangeException(nameof(x));
            if (y < 0 || y >= Height)
                throw new ArgumentOutOfRangeException(nameof(y));

            // fix the byte mapping to whatever the actual is
            Span<byte> buffer = _bytes;
            Span<byte> temperature = buffer.Slice(BytesPerPixel * (Width * y + x), BytesPerPixel);
            return ConvertToTemperature(temperature);
        }
    }
}
```

src/devices/Amg88xx/Amg88xx.cs
krwq: fair enough for now, if we have users asking and clearer idea of API shape we can improve in the future.

src/devices/Amg88xx/Amg88xx.cs
krwq: this seems to be missed comment

src/devices/Amg88xx/Amg88xx.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <None Include="README.md" />
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
+  </ItemGroup>
+
+  <!-- Make the internal classes visible to the unit test assembly -->
+  <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleToAttribute">
+    <_Parameter1>$(AssemblyName).Tests</_Parameter1>
+    </AssemblyAttribute>
+  </ItemGroup>
```
krwq, line 20: neat

src/devices/Amg88xx/Amg88xx.cs
fehdem: @krwq Yes, with the coffee of today, I do agree, ;)

src/devices/Amg88xx/Amg88xx.cs
fehdem: @krwq Even though I can't promisse anything, I'll keep an eye on that and may pick up future improvement requests.

src/devices/Amg88xx/Amg88xx.cs
fehdem: @krwq I got your point of reducing the (re-)allocations of memory for each frame.
I think about another adjustment to the API out of this.
I like the indexer approach you suggested. But, I have the temperature-converted image and the raw image. I want to keep both options, because both have reasonable use cases (I use the temperatures for display purpose and I feed the raw values into a neural network.) 
However, I think I could make the raw image even more raw. Instead of using an x/y-index I could use the pixel number (0..63) only. This would meet even better the neuronal network use case. 
So I would end up with two indexers:
```
public Temperature this[int x, int y]
public UInt16 this[int pixel]
```
What do you think about this idea?

Ellerbach: > What do you think about this idea?

I personally like the idea! Makes sense

src/devices/Amg88xx/Amg88xx.cs
fehdem: Yes, I missed it. I renamed it to "purely" ```Reset```. The datasheet calls it "initial reset", but this doesn't add anything useful for the user.

src/devices/Amg88xx/Amg88xx.cs
krwq: @fehdem sounds good but to improve perf you can also make pixel array user provided, i.e.:

```csharp
public void ReadPixels(Span<UInt16> pixels)
```

src/devices/Amg88xx/Amg88xxUtils.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
Ellerbach, line 3: We've just moved to 2 lines headers. If you can adjust, that would be perfect.

Ellerbach (approved):
Thanks, looks good to me. If you can adjust to the new header, would be perfect, also missing enter at the end of 2 files.

src/devices/Amg88xx/Amg88xxUtils.cs
krwq: (remove third line, no change on the first two)

fehdem: Done for all .cs files.

Ellerbach (approved):
I though I already approved it, seems that are are very few minor things to fix from various feedback.

fehdem:
@Ellerbach Thank you for the quick re-review. Is there a chance, that we finish this PR soon? 
I would like to start working on the next device binding, but complete this first.

Ellerbach:
> Is there a chance, that we finish this PR soon?

We're going thru changes to support .NET 5 and that does involve all the bindings and all the repo. I let @joperezr or @krwq approve the PR and see if we merge it before or after the changes.

> I would like to start working on the next device binding

Nothing prevent you to start :-) Just use another branch. Most changes that will have to be adjusted are related to .NET 5 nullable. You can check in #1243

fehdem:
@Ellerbach 
> Nothing prevent you to start :-) Just use another branch. Most changes that will have to be adjusted are related to .NET 5 nullable. You can check in #1243

As I'm working on the master branch of my fork again I'm hesitating - I don't want to mess up the git branches again. I'm not really experienced with using git in this kind of setup.

src/devices/Amg88xx/FrameRateBit.cs
krwq: Should this enum be internal in that case?

krwq: also since this has only a single value consider making it constant

krwq (approved):
LGTM, please fix/respond to unresolved conversations (I've reopened one with a question and also seen one from @Ellerbach)

krwq:
@richlander this PR is ready to merge, please let me know if you're ok with merging this before nullability changes or prefer to wait (we'll likely need to help with adding annotations here)

src/devices/Amg88xx/Amg88xx.cs
fehdem: Done

src/devices/Amg88xx/Amg88xx.cs
fehdem: Done

src/devices/Amg88xx/Amg88xx.cs
fehdem: Done

src/devices/Amg88xx/Amg88xx.cs
fehdem: Done

src/devices/Amg88xx/FrameRateBit.cs
fehdem: yes, this should be internal. I missed that.
I would keep it an enum for code consistency, though this is my taste.

src/devices/Amg88xx/Amg88xx.cs
fehdem: I think the API is good with the indexers and the internally kept pixel buffer.

fehdem:
@krwq I just fixed the open remarks. Hopefully I didn't miss any this time. 
BTW: I'm very exited having .net 5 finally available and the IOT stuff migrated as well.

joperezr:
@fehdem I took the liberty of rebasing and fixing the new nullability warnings directly on your PR so that you don't have to do this. @krwq is not clear if there is anything left to address here or if this is good to go from your two last comments, do you mind advising here?

fehdem:
@joperezr Thank you! Though I did already prepare myself to do it ;)

joperezr:
Great! since we were actually the ones that introduced these breaks to your PR by merging nullability support I thought it would be correct for us to also fix your PR

fehdem:
@krwq @Ellerbach @joperezr Is there any chance that we get this PR to master and closed soon?

joperezr:
That's the plan yes, we are tracking it as a needs one final round of review, so I'll take a look at this soon and once it is done we'll go ahead and merge.

fehdem:
@joperezr, @Ellerbach Is there any progress with the review? It's a little bit demotivating spending time with a lot of enthusiasm for contributions, if the final step is pending so long.
I know you're working on many things, but this one went through multiple reviews already and I think it could and should be done now. 
I really appriciate this project and I would really like to invest further time in near future for further contributions.

joperezr:
> Is there any progress with the review? It's a little bit demotivating spending time with a lot of enthusiasm for contributions, if the final step is pending so long.

I'm sorry to hear that, as you suspect we have been working really hard to burn down a lot of the debt we had on the repo with many PRs that had been open for a long time so it is taking us a bit more than we expected but I think we are making great progress and we are getting the repo into a healthy state. I'm working on this review as I'm writing this and the idea is for future PRs to respond much faster and have the iteration cycles be much better than in the past.

> I really appriciate this project and I would really like to invest further time in near future for further contributions.

I'm glad to hear this and I'm looking forward to collaborating on more great contributions to make the device collection even richer.

joperezr (approved):

Issue #1190, opened 2020-09-21
Move samples projects to .NET 5.0


richlander:
We should decide when we can do this. Can we do it now? Wait until .NET 5.0 RTM?

It would be really nice to move all of the samples to top-level programs, for example.

What do people think that use this repo? Do you use the samples (both at https://github.com/dotnet/iot/tree/master/samples and the ones that exist in samples directories at https://github.com/dotnet/iot/tree/master/src/devices)?

Ellerbach:
> Can we do it now?

You can always prepare everything...

> Wait until .NET 5.0 RTM?

...and push the PR once .NET 5.0 is final.

And if you update, better update everything for consistency. At least it will be easier.

shaggygi:
> It would be really nice to move all of the samples to top-level programs, for example.

Don't quite understand the comment.  Do you mean move all binding samples to https://github.com/dotnet/iot/tree/master/samples?  If so, I disagree. It seems to be better to keep the binding samples within respective binding folder.  That allows it to be discovered easier and updated when changes or new features occur.

As for https://github.com/dotnet/iot/tree/master/samples, this is a good location for newbies to get a quick sense of using APIs.

krwq:
I agree with @shaggygi that it's much easier to find when they are local to the project. What we can do is perhaps generate part of https://github.com/dotnet/iot/blob/master/samples/README.md to point to samples/READMEs from devices

pgrawehr:
@richlander : Due to #1099, the entire library currently is not working with .NET Core 5, preview 5 or later. So unless you use preview 4, you'll have a hard time converting anything to .NET 5.

richlander:
I wasn't proposing to move the samples from the binding folder. I also agree that we should not do that and they are best where they are.

@pgrawehr -- this is due to a breaking change w/rt to WinRT, related to https://github.com/dotnet/iot/issues/1193. 5.0 samples should now be resolved.

Once my open NRTs people gets merged, I'll create a PR for updating the samples.

PR #1224, opened 2020-10-26
Update samples to .NET 5.0

richlander:
- Updated the TFM.
- Updated most of the samples to top-level programs.
- Leverages target-type new() in a few places.

I think we can merge this at any time. We can wait until 11/10 if people want, but that doesn't seem necessary given the "go live" nature of the release.

Fixes https://github.com/dotnet/iot/issues/1190


samples/force-sensitive-resistor/FsrWithCapacitorSample.cs
```diff
@@ -8,12 +8,11 @@ namespace force_sensitive_resistor
 {
     class FsrWithCapacitorSample
     {
-        private GpioController _controller;
+        private GpioController _controller = new();
```
joperezr, line 11: Nifty :smile: I wasn't aware of this C#9 feature

joperezr: I know that this is just a sample but given you are touching this line we might as well just add the readonly to _controller 😆

samples/force-sensitive-resistor/Program.cs
```diff
@@ -5,54 +5,46 @@
 using System;
 using System.Threading;
 using Iot.Device.Adc;
+using force_sensitive_resistor;
 
-namespace force_sensitive_resistor
```
joperezr, line 9: I understand we are trying to showcase functional programming for c#9 here, but I wonder if we should stick with object oriented language given that these samples are supposed to be entry-level sort of examples?

joperezr: Also, I'm open for discussion here, if you think it's best to keep it as you have it I'm fine too.

joperezr:
I guess we don't have any of the samples in which the Readme tells you how to run it in Windows right? Because if we do, I think it would be good if in that sample we instead targeted net5.0-windows10.0.17763.0 in order to show folks how to use our library from the Windows side.

joperezr (approved):
Left some minor comments but this looks great otherwise. Thanks @richlander

pgrawehr:
> 
> 
> I guess we don't have any of the samples in which the Readme tells you how to run it in Windows right? Because if we do, I think it would be good if in that sample we instead targeted net5.0-windows10.0.17763.0 in order to show folks how to use our library from the Windows side.

I could provide a binding that reads CPU temperatures, Fan speeds etc. for Windows (using an external library). This one is really Windows only (without actually needing the GPIO stuff, though) and is also a good example for the usage of units. Want to take a look?

joperezr:
While it would be good to have either that sample or a pointer to it somewhere in the repo, I think I would like to keep the samples section of the repo to specific usages of System.Device.Gpio. The idea with my question was to have one of these samples to target Windows, so having the right TFM to target that.

samples/force-sensitive-resistor/FsrWithCapacitorSample.cs
richlander: done

samples/force-sensitive-resistor/Program.cs
richlander: Seems like keeping that line would significantly reduce the value of top-level projects. Also seems like we're not losing much. I also don't think we're losing anything OO. It would be nice if you could declare the namespace with a single line that didn't require nesting the rest of the file.

richlander:
I responded to the feedback and found some more opportunities for simplification (and uniformity) with dispose/using.

I took an initial take at documenting windows support. We probably need a central document along the lines of `targeting-windows.md`. Where do you think that document should live?

PR #1236, opened 2020-10-29
Add support for Meddler Toledo devices

frogcrush:
This would add basic support for Mettler Toledo MT-SICS devices, which are a line of scales/balances. It has NOT been tested on real devices yet, but will be in the coming weeks. 

It is based on the [official documentation,](https://www.mt.com/mt_ext_files/Editorial/Generic/7/MT-SICS_for_Excellence_Balances_BA_Editorial-Generic_1116311007471_files/Excellence-SICS-BA-e-11780711B.pdf) which I'm not sure we're allowed to include in the repo, or where I would link it.


src/devices/MettlerToledo/Exceptions/MettlerToledoException.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Iot.Device.MettlerToledo.Exceptions
+{
+    /// <summary>
+    /// A generic Mettler Toledo exception
+    /// </summary>
+    public abstract class MettlerToledoException : System.Exception
```
Ellerbach, line 6: I don't feel it's really necessary to have specific exceptions for specific bindings. A simple IOException with the command as a parameter is doing the exact same effect.
IOException are widely used in the repo when there is an issue communicating with the devices.
I've been thru the all code and I'm really not sure that it brings a lot of value to the developer. The difference between the fact a command has been executed badly or create an issue is just that you can't get a proper result. To make it properly, it requires to handle different type of exceptions. And at the end, a simple retry mechanism should be enough.

src/devices/MettlerToledo/Exceptions/BalanceOutOfRangeException.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
```
Ellerbach, line 1: Missing licensing headers everywhere. Please see on other bindings. It needs to be added everywhere.

src/devices/MettlerToledo/Messages.cs
```diff
@@ -0,0 +1,130 @@
...
+            internal const string DYNAMIC_SUCCESS = "D";
+        }
+    }
+
+    /// <summary>
+    /// Unit of weight as expressed by the Mettler Toledo balance.
+    /// </summary>
+    public enum MettlerToledoWeightUnit
+    {
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
Ellerbach, line 106: Why do you disable the XML comments? Seems like this is a public enum, so comments are needed. 
And also, why this needs to be public while you're only using it internally with the dictionary?

src/devices/MettlerToledo/Messages.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            internal const string STABLE_SUCCESS = "S";
+            internal const string DYNAMIC_SUCCESS = "D";
+        }
+    }
+
+    /// <summary>
+    /// Unit of weight as expressed by the Mettler Toledo balance.
+    /// </summary>
+    public enum MettlerToledoWeightUnit
```
Ellerbach, line 104: Would be better to have this one in a separate file. We tend to have 1 file per class especially for public. Few exceptions.

src/devices/MettlerToledo/Messages.cs
```diff
@@ -0,0 +1,130 @@
...
+using System.Text;
+
+namespace Iot.Device.MettlerToledo.Messages
+{
+    internal static class Errors
+    {
+        /// <summary>
+        /// The balance has not recognized the received command.
+        /// </summary>
+        internal const string SYNTAX_ERROR = "ES";
```
Ellerbach, line 12: Please adjust the constant to be compliant with the naming convention. it should be `SyntaxError`. The exception are for register to match the documentation. In the documentation, the only names are the ES, ET, etc, so the naming convention should apply everywhere.

src/devices/MettlerToledo/MettlerToledoDevice.cs
```diff
@@ -0,0 +1,362 @@
...
+        /// Retrieves the last internal weight value (stable or dynamic) before receipt of this command.
+        /// </summary>
+        /// <returns>Current weight reading</returns>
+        public MettlerToledoWeightReading GetImmediateWeight() => GetWeightReading(true);
+
+        /// <summary>
+        /// Opens a connection to the MT-SICS scale.
+        /// </summary>
+        /// <exception cref="UnauthorizedAccessException"></exception>
+        public void Open()
```
Ellerbach, line 88: Should that be hidden? You may open when you create and Close when you dispose. Any reason to leave those 2 public?

src/devices/MettlerToledo/MettlerToledoDevice.cs
```diff
@@ -0,0 +1,362 @@
...
+        {
+            return SendCommandWithResponse(command, command.Split(' ')[0]);
+        }
+
+        private string[] ReadResponse(string command, string responseCode, TimeSpan? timeout = null)
+        {
+            var start = DateTime.Now;
+            while (timeout == null || DateTime.Now - start < timeout)
+            {
+                // TODO: this timeout isn't accurate when timeout < _serialPort.ReadTimeout, because time may have elapsed already
```
Ellerbach, line 264: So is there a simple way to fix this?

src/devices/MettlerToledo/MettlerToledoDevice.cs
```diff
@@ -0,0 +1,362 @@
...
+
+    /// <summary>
+    /// Provides an interface for interacting with MT-SICS devices, like scales.
+    /// </summary>
+    public class MettlerToledoDevice : IDisposable
+    {
+        // TODO: we need to test on a real scale to determine what the response for other weights are when they are set to Unit 1
+        private static readonly (string mtAbbreviation, MettlerToledoWeightUnit mtUnit, UnitsNet.Units.MassUnit unitsNetUnit)[] _unitMapping =
+        {
+            ("g", MettlerToledoWeightUnit.Grams, UnitsNet.Units.MassUnit.Gram),
```
Ellerbach, line 36: Why are there only 3 enums mapped while you have more? Also seems that all should be able to be mapped with the MassUint.
The abbreviation can be gathered from the UnitsNet as well.

src/devices/MettlerToledo/README.md
```diff
@@ -0,0 +1,40 @@
+﻿# Mettler Toledo
+
+## Summary
+This binding provides a higher-level interface for interacting with [Mettler Toledo](https://www.mt.com)-brand scales.
```
Ellerbach, line 4: missing enter before to have a proper markdown. I can recommend to use a markdown linter. You haver quite some choice. Most fix automatically those kinds of nit.

src/devices/MettlerToledo/samples/README.md
```diff
@@ -0,0 +1,4 @@
+# TODO: This needs to be determined
+
+Help Wanted Please
```
Ellerbach, line 3: What kind of help do you need? :-) You can refer to the main readme as you placed the example and detailed usage there

src/devices/MettlerToledo/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Device Family
+This binding should work with any MT-SICS-compatible scale that uses MT-SICS Level 0. Some Level 1 commands are also available.
```
Ellerbach, line 7: Is there a way to get a list of the functions implemented and not implemented at the end of this readme at least for the level 0 and 1? 
I see 3 levels of commands. so if none from level 2 and level 3 are implemented, fine to say none and don't go into the details. That will make the life easier for anyone who wants to improve later on and add missing commands.

src/devices/MettlerToledo/samples/MettlerToledo.Sample.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            // Now fetch a reading, when the scale is stable. This means the scale is confident in the weight.
+            MettlerToledoWeightReading reading = scale.GetStableWeight();
+
+            // Print reading to console
+            Console.WriteLine($"Received a weight of {reading.Weight.Grams} grams.");
+
+            scale.WeightUpdated += (s, e) =>
+            {
+                Console.WriteLine($"Scale reported a new weight of {reading.Weight.Grams} grams.");
```
Ellerbach, line 37: isn't it e rather than reading?

src/devices/MettlerToledo/MettlerToledoDevice.cs
```diff
@@ -0,0 +1,362 @@
...
+                    throw new BalanceOutOfRangeException(false, true, command);
+            }
+
+            return ParseWeightReading(response);
+        }
+
+        private MettlerToledoWeightReading ParseWeightReading(string[] response)
+        {
+            var value = double.Parse(response[2]);
+            var unit = _unitMapping.First(x => x.mtAbbreviation == response[3]).unitsNetUnit;
```
Ellerbach, line 337: Even if I love Linq, I would avoid using it and just browse the collection the normal way. Reason is because we're working on porting automatically bindings to the nanoFramework world to have those kind of bindings working on tiny devices such as ESP32. And there generics and Linq are not supported.

Ellerbach (commented):
Thanks for this nice addition and great start. I've added couple of comments for the naming conventions, missing headers, some thoughts about Exceptions and few more.

src/devices/MettlerToledo/MettlerToledoDevice.cs
frogcrush: So these are abbreviations given by the MT-SICS scales, which seem to be different than some of the normal abbreviations. Unfortunately without access to the physical scale I can't confirm what the abbreviations are yet (which is why only 3 are provided)

src/devices/MettlerToledo/samples/README.md
frogcrush: Oops, this is part of the template

joperezr:
[Triage]: @frogcrush have you had a chance to test against the real device? Also, looks like @Ellerbach added some feedback a while ago, do you know when you will be able to address it?

frogcrush:
Unfortunately no to both. I no longer have any sort of ability to access the machinery so it's probably good to consider this dead unfortunately.

PR #1237, opened 2020-10-29
Fixes #1079: Creates a basic device binding for generic relays.

### LINKED ISSUE #1079: Relay device
I would like to add support for some different relay boards. These devices are super simple and support for this would be to add consistency for code in projects more than anything. 

This device would simply keep track of what port a relay is on, and provide methods for turning it on and off by setting that pin high and low. 

image

frogcrush:
Creates a basic device binding for generic relay devices, which is helpful for keeping track of said devices and managing their state. This is especially true when said devices differ in being NormallyOpen and NormallyClosed relays. 

This PR has been tested for basic functionality, but a double-check of the inversion for NormallyOpen and NormallyClosed relays would be much appreciated.


src/devices/RelayBoard/RelayType.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace RelayBoard
```
Ellerbach, line 5: IoT.Device.RelayBoard

src/devices/RelayBoard/RelayBoard.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        /// <summary>
+        /// Create and initialize a relay on this board.
+        /// </summary>
+        /// <param name="pin">Pin to use.</param>
+        public void CreateRelay(int pin)
+        {
+            _relays.Add(new Relay(pin, _controller, Type));
+        }
+
```
Ellerbach, line 58: I would add as well a RelayBoard.Set(int pin, bool state) to avoid having to get a specific Relay

src/devices/RelayBoard/samples/RelayBoard.Sample.cs
```diff
@@ -0,0 +1,45 @@
...
+            board.CreateRelays(2, 3, 4);
+
+            // Go through all the relays and set them to on
+            foreach (var relay in board)
+            {
+                relay.On = true;
+            }
+
+            // Get a specific relay
+            var r = board.GetRelay(1);
```
Ellerbach, line 41: See my comment in the RelayBoard, maybe a RelayBoard.Set(int pin, bool state) can be usefull as well

src/devices/RelayBoard/samples/RelayBoard.Samples.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
Ellerbach, line 5: Can be 3.1 now

src/devices/RelayBoard/Relay.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+        #endregion
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Relay"/> class.
+        /// </summary>
+        /// <param name="pin">The pin this relay is on.</param>
+        /// <param name="controller">The GPIO controller to use.</param>
+        /// <param name="relayType">The type of relay this is.</param>
+        public Relay(int pin, GpioController controller, RelayType relayType = RelayType.NormallyClosed)
```
Ellerbach, line 55: This can be used just like this, without the RealyBoard. So same as for the RelayBoard, you can adjust the controller with the same _shloudDispose pattern. And in this case when you call the constructor, you of course pass the existing GpioController and don't dispose.
so If I need 1 relay, I can just use this directly as well with the rigth patten https://github.com/dotnet/iot/tree/master/src/devices#gpiocontroller

Ellerbach (commented):
Thanks for this addition, simple and efficient, few comments.

src/devices/RelayBoard/RelayType.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Describes what type of relay a device is.
+    /// </summary>
+    public enum RelayType
```
HumJ0218, line 10: Relay type in addition to normal open and normal closure, but also a large number of such as SPDT, DPDT and so on. How should this be handled?

Ellerbach: I think that the only importance here is the fact the relay is normally open or normally close. The technology behind has no importance for the code side. At least that's how I've always used relay.

krwq: for the extra parameters you should create them only in the device which has such option but they should not be part of generic abstraction

src/devices/RelayBoard/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+You can either use the ``RelayBoard`` class to help organize multiple relays (especially if they are of the same type), or use the individual ``Relay`` class directly.
+
+If you use ``Relay`` directly, you *must* manage the lifetime of the GPIO controller.
+
+An example of how you would use a ``RelayBoard``:
+
+```csharp
+// Create a relay board, using default values
+var board = new RelayBoard();
```
krwq, line 19: in a sample you should dispose the relay as well to avoid teaching anti-patterns

src/devices/RelayBoard/Relay.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using System;
+using System.Device.Gpio;
+using System.Device.I2c;
+using System.Device.Spi;
+
+namespace RelayBoard
+{
+    /// <summary>
+    /// A relay.
```
krwq, line 13: please put something more descriptive here (at minimum follow other classes)

src/devices/RelayBoard/RelayBoard.cs
```diff
@@ -0,0 +1,138 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Device.Gpio;
+using System.Device.I2c;
+using System.Device.Spi;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
```
krwq, line 12: X509Certificates?

src/devices/RelayBoard/RelayBoard.cs
```diff
@@ -0,0 +1,138 @@
...
+        public void CreateRelay(int pin)
+        {
+            _relays.Add(new Relay(pin, _controller, Type));
+        }
+
+        /// <summary>
+        /// Create and initialize relays on the specified pins.
+        /// </summary>
+        /// <param name="pins">List of pins to create relays on.</param>
+        public void CreateRelays(params int[] pins)
```
krwq, line 63: I feel like this should be a constructor parameter and we should not have this API directly - you should know what relays you have up front - I do not think it will be a common scenario that someone will plug & play relays, and if we find in the future that it is then we should consider doing it

src/devices/RelayBoard/RelayBoard.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                CreateRelay(pin);
+            }
+        }
+
+        /// <summary>
+        /// Remove a relay from this board.
+        /// </summary>
+        /// <param name="pin">Pin the relay is on.</param>
+        public void RemoveRelay(int pin)
```
krwq, line 75: same, while API seems useful for plug & play scenarios I do not think this is very common and I do not see why user would ever call this

src/devices/RelayBoard/RelayBoard.cs
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            _relays.RemoveAt(relay);
+        }
+
+        /// <summary>
+        /// Remove a relay from this board.
+        /// </summary>
+        /// <param name="relay">The relay to remove.</param>
+        public void RemoveRelay(Relay relay)
```
krwq, line 90: AddRelay/RemoveRelay taking a Relay class I'm fine with leaving

src/devices/RelayBoard/RelayBoard.cs
```diff
@@ -0,0 +1,138 @@
...
+    /// A board containing some <see cref="Relay"/> devices.
+    /// </summary>
+    /// <example>
+    /// Create a relay board, with a relay:
+    /// <code>
+    /// RelayBoard board = new RelayBoard();
+    /// board.CreateRelay(pin: 1);
+    /// </code>
+    /// </example>
+    public class RelayBoard : IDisposable, IEnumerable<Relay>
```
krwq, line 28: consider implementing `IList<Relay>` (or IReadOnlyList)

src/devices/RelayBoard/RelayBoard.cs
```diff
@@ -0,0 +1,138 @@
...
+using System.Collections.Generic;
+using System.Device.Gpio;
+using System.Device.I2c;
+using System.Device.Spi;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+
+using RelayBoard;
+
+namespace Iot.Device.RelayBoard
```
krwq, line 16: I'm having a hard time figuring out what's the benefit of RelayBoard class (based on GpioController) - if this was an I2C/SPI device with multiple relays I can see the interface being somewhat useful but in case of GpioController why not just use array directly? The reason I bring this up is we go ahead with this class there is no reason why we shouldn't have: ButtonBoard/KeyBoard, LedBoard/LedStrip (or if you go dimension further LedMatrix), PwmBoard, AdcBoard, DHT11Board and many other single pin devices... Hope this is making sense why I'm pushing back on this API

src/devices/RelayBoard/RelayType.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
```
krwq, line 1: copyright header

src/devices/RelayBoard/RelayBoard.cs
frogcrush: My thought was if you are dynamically configuring a device - eg the Pi has a web portal, where you can add or remove components/relays/etc. or dynamically set what each pin does through said UI. However, we can change this if you want.

src/devices/RelayBoard/Relay.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.Gpio;
+using System.Device.I2c;
+using System.Device.Spi;
+
+namespace RelayBoard
```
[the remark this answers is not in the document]
frogcrush: Good catch! I thought I had changed it in this project.

src/devices/RelayBoard/Relay.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        /// <summary>
+        /// Gets the type of relay this device is.
+        /// </summary>
+        public RelayType Type { get; }
+
+        /// <summary>
+        /// Get or set if the relay is currently on
+        /// </summary>
+        public bool On
```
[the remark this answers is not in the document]
frogcrush: I took the naming from the issue (#1079) and the discussion there. I feel like Enable might be confusing/unclear but I agree that "On" might not be the best choice either.

src/devices/RelayBoard/RelayBoard.cs
frogcrush: Not a bad idea!

src/devices/RelayBoard/RelayBoard.cs
frogcrush: Odd, I have no idea why that got added

src/devices/RelayBoard/RelayBoard.cs
frogcrush: The reason for this is due to the ability to abstract away the construction of the relays; for each relay board they are going to be the same type (Normally Open or Normally Closed) and use the same GpioController. With the RelayBoard class, you only have to worry about setting that property once. 

I went back and forth a lot on whether or not this was a good enough reason to include RelayBoard, and in the end I felt like there was enough of a benefit to add this while still allowing users to create the Relay directly (and then they could keep their own list/reference to each one).

src/devices/RelayBoard/Relay.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            _controller?.ClosePin(Pin);
+
+            if (_shouldDispose)
+            {
+                _controller?.Dispose();
+            }
```
matthewethomas, line 76: Looking at some other implementations in the Iot.Device.* namespace, it seems like the convention is to null out the controller instance after disposing it. Also, it looks like GpioController [closes any open pins on dispose](https://github.com/dotnet/iot/blob/master/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs#L343), so manually closing before disposal isn't necessary

I suggest this implementation from [DhtBase](https://github.com/dotnet/iot/blob/06fc3a0e78f74e430e26183ad342b24c1c92c368/src/devices/Dhtxx/DhtBase.cs#L287)
```suggestion
            if (_shouldDispose)
            {
                _controller?.Dispose();
                _controller = null;
            }
            else if(_controller?.IsPinOpen(Pin) ?? false)
            {
                _controller.ClosePin(Pin); 
            }
```

src/devices/RelayBoard/Relay.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Relay"/> class.
+        /// </summary>
+        /// <param name="pin">The pin this relay is on.</param>
+        /// <param name="gpioController">The GPIO controller to use.</param>
+        /// <param name="relayType">The type of relay this is.</param>
+        /// <param name="pinNumberingScheme">Numbering scheme to use for <paramref name="gpioController"/>, if one is not passed.</param>
+        /// <param name="shouldDispose">Whether this class should dispose the <paramref name="gpioController"/>.</param>
+        public Relay(int pin, GpioController? gpioController = null, RelayType relayType = RelayType.NormallyClosed, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical, bool shouldDispose = true)
```
matthewethomas, line 59: Is it possible to not default relay type? If this is going to become part of the framework, I suspect whatever is the default is always going to be the opposite of what is needed in someone's project

src/devices/RelayBoard/RelayBoard.cs
```diff
@@ -0,0 +1,171 @@
...
+
+        /// <summary>
+        /// Gets the type of relay board.
+        /// </summary>
+        public RelayType Type { get; init; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RelayBoard"/> class.
+        /// </summary>
+        public RelayBoard(RelayType relayType = RelayType.NormallyClosed, GpioController? gpioController = null, PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical, bool shouldDispose = true)
```
matthewethomas, line 37: Same comments from Relay.cs:
Is it possible to not default relay type? If this is going to become part of the framework, I suspect whatever is the default is always going to be the opposite of what is needed in someone's project

src/devices/RelayBoard/RelayBoard.cs
matthewethomas: @krwq I could see this being very useful in an initializer function or stage

If this is a constructor argument, in a DI scenario, you'll need to provide a custom resolver func for the board which pulls your pin configuration, and passes the pins as appropriate. Not a lot of overhead, but just enough to muddy up type registration.

I'm using IHostedService as an example, which provides a StartAsync function - if registered correctly, this can serve as an initializer before your web server or any other hosted services are started and begin to . Other implementations will use different means.

In this initializer stage, you can configure your hardware - open the pins on gpio controller to use the desired mode, ensure connectivity between devices, etc. And in this case create the appropriate relays from configuration - all before it's actually used by application code.

I get this is probably a matter of semantics, but if we're voting, I'd like to see this stay.

joperezr:
[Triage]: @frogcrush looks like the build is broken here, are you able to reproduce locally on what is going on? Also, there seems to be un-addressed feedback, do you know when you will be able to take  a look?

joperezr:
[Triage]: @frogcrush friendly ping 😄. Let us know if you are still interested in working on this so that we can keep it open.

PR #1238, opened 2020-10-30
CPU Temperature (and similar values) support for Windows

pgrawehr:
This adds basic CPU temperature support for windows (unfortunately, it is not very generic, so it might work on your hardware or not). 

A separate binding to get values from OpenHardwareMonitor (https://openhardwaremonitor.org/) is provided which gets a full set of whatever values you might want to know about your PC. This leverages the requirement to do anything hardware-dependent on our side. It is also a good example on how to work with UnitsNet.


src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
```diff
@@ -0,0 +1,740 @@
...
+        /// Adds some special derived sensors.
+        /// - For each power sensor, this adds another sensor that integrates power over time and so generated the energy used in W/h or
+        /// more conveniently, Kilowatthours (this is the unit the electricity bill bases on)
+        /// - Gives the heat flux for the primary CPU, using the given CPU die size (or a default value)
+        /// </summary>
+        /// <param name="cpuDieSize">Die size of your CPU, optional. Find your CPU on https://en.wikichip.org/ to find out. Note: This
+        /// value is usually much smaller than the size of the physical CPU.</param>
+        public void EnableDerivedSensors(Area cpuDieSize = default)
+        {
+            if (cpuDieSize.Value == 0)
```
Ellerbach, line 434: that's minimal but you may want to test == default as your default initialized value is default.

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
```diff
@@ -0,0 +1,740 @@
...
+
+        /// <summary>
+        /// Adds some special derived sensors.
+        /// - For each power sensor, this adds another sensor that integrates power over time and so generated the energy used in W/h or
+        /// more conveniently, Kilowatthours (this is the unit the electricity bill bases on)
+        /// - Gives the heat flux for the primary CPU, using the given CPU die size (or a default value)
+        /// </summary>
+        /// <param name="cpuDieSize">Die size of your CPU, optional. Find your CPU on https://en.wikichip.org/ to find out. Note: This
+        /// value is usually much smaller than the size of the physical CPU.</param>
+        public void EnableDerivedSensors(Area cpuDieSize = default)
```
Ellerbach, line 432: I really love this concept showing the power of having UnitNet and being able to make advance scenarios!

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
```diff
@@ -0,0 +1,740 @@
...
+            Sensor cpuPower = null;
+
+            foreach (var sensor in GetSensorList())
+            {
+                if (sensor.SensorType == SensorType.Power)
+                {
+                    // Energy usage (integration of power over time)
+                    var managementInstance = new EnergyManagementInstance();
+                    Sensor newSensor = new Sensor(managementInstance, sensor.Name + " Energy", sensor.Identifier + "/energy", sensor.Identifier, SensorType.Energy);
+                    newSensor.Job = StartMonitoring(sensor, TimeSpan.FromMilliseconds(500), (s, value, timeSinceUpdate) =>
```
Ellerbach, line 462: any reason why 500 ms?

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
```diff
@@ -0,0 +1,740 @@
...
+                    });
+                    _derivedSensors.Add(newSensor);
+                }
+
+                // For CPU package, calculate heat flux
+                if (sensor.SensorType == SensorType.Power && sensor.Name.IndexOf("CPU Package", StringComparison.OrdinalIgnoreCase) >= 0)
+                {
+                    var managementInstance = new EnergyManagementInstance();
+                    var newSensor = new Sensor(managementInstance, sensor.Name + " HeatFlux", sensor.Identifier + "/heatflux", sensor.Identifier, SensorType.HeatFlux);
+                    newSensor.Job = StartMonitoring(sensor, TimeSpan.FromMilliseconds(500), (s, value, timeSinceUpdate) =>
```
Ellerbach, line 478: Same, any reason why 500 ms? Can this be adjusted? Having shortest period would give more precise value if I understand correctly the logic

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
```diff
@@ -0,0 +1,740 @@
...
+                    return false;
+                }
+
+                if (typeof(T) != elem.Type)
+                {
+                    value = default(T);
+                    return false;
+                }
+
+                float newValue = Convert.ToSingle(_instance["Value"]);
```
Ellerbach, line 619: With this pattern, how do we ensure that the unit is created with the right default unit? Are _instance["Value"] always returning the default international unit? I may have miss something here, I would just like to understand

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
Ellerbach, line 4: This may raise a question as so far netcore2.1 is used. I'm personally very happy to see netstandard2.0 btw

src/devices/OpenHardwareMonitor/README.md
```diff
@@ -0,0 +1,20 @@
...
+This binding works on Windows only. It requires that OpenHardwareMonitor (https://openhardwaremonitor.org/) is running in the background. While that tool requires elevated permissions to work, 
+the binding (and the application using it) does not. 
+
+The binding supports some additional, "virtual" sensor measuments that are derived from other values. The following extra values are provided:
+
+- For each sensor returning power, another is generated which calculates energy consumed (by integrating the values over time).
+- From the CPU Package power sensor, the CPU heat flux is calculated (estimating the size of the die).
+- If both a power and a voltage sensor are available for CPU Package, the current is calculated.
+
+## References
```
Ellerbach, line 20: seems like you are missing the references

src/devices/OpenHardwareMonitor/SensorType.cs
```diff
@@ -0,0 +1,20 @@
+﻿#pragma warning disable CS1591
```
Ellerbach, line 1: header

Ellerbach: And why deactivation the /// comments?  I don't see any reason why

Ellerbach (commented):
Thanks for this addition, really like it all up. Great usage of the UnitsNet especially to add calculated additional sensors. That makes life so easy.

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -25,6 +25,7 @@
     <PackageReference Include="System.IO.Ports" Version="$(SystemIOPortsPackageVersion)" />
     <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
     <PackageReference Include="UnitsNet" Version="$(UnitsNetPackageVersion)" />
+    <PackageReference Include="System.Management" Version="5.0.0-alpha.1.19618.1" />
```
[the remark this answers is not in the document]
pgrawehr: I think there should be a new version out now. I wrote this binding some time ago.

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
```diff
@@ -0,0 +1,740 @@
...
+        }
+
+        /// <summary>
+        /// Constructs a new instance of this class.
+        /// The class can be constructed even if no sensors are available or OpenHardwareMonitor is not running (yet).
+        /// </summary>
+        /// <exception cref="PlatformNotSupportedException">The operating system is not Windows.</exception>
+        public OpenHardwareMonitor()
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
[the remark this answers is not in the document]
pgrawehr: I will check what it takes to use it on linux (and at least document that).

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
pgrawehr: That's true, the shorter the period is, the better the integration works. But relying on the user to know what an integral is, might be a bit demanding, too. I'll think of something.

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.csproj
pgrawehr: I think most projects where changed some time ago, and apparently I just copied this over from an older project state. But since it's only the test package, it doesn't really matter (these dll's are not shipped).

src/devices/OpenHardwareMonitor/SensorType.cs
pgrawehr: I customary add this pragma during development, so that I don't need to write the documentation before I even know whether the API works the way I start it. But of course, this should never be in a PR ;-)

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
Ellerbach: So maybe a default TimeSpan in the function with this default time? Anyone else who'd like something more precise would be able to?

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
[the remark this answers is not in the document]
krwq: should this be Windows-only dependency? @joperezr for how to do it

src/devices/CpuTemperature/CpuTemperature.cs
```diff
@@ -3,29 +3,64 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
```
krwq, line 6: I think in this case because implementation is completely different I'd recommend to create platform specific files: CpuTemperature.Linux.cs, CpuTemperature.Windows.cs

Also you should not have public API differences between Windows/Linux, extra stuff can be available through Windows-only library though

src/devices/CpuTemperature/CpuTemperature.csproj
```diff
@@ -10,4 +10,8 @@
     <None Include="README.md" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Management" Version="5.0.0-alpha.1.19618.1" />
```
krwq, line 14: I'd recommend to make it Windows-only dependency

src/devices/CpuTemperature/README.md
```diff
@@ -1,11 +1,13 @@
 # Cpu Temperature
 
-Device bindings for the CPU Temperature Sensor
+Device bindings for the CPU Temperature Sensor on Linux and Windows
```
krwq, line 3: since we now cover all OS-es (I assume OSX also works through Linux implementation) I'd just not change this

src/devices/CpuTemperature/samples/CpuTemperature.Sample.cs
```diff
@@ -13,17 +13,29 @@ internal class Program
...
+            Console.WriteLine("Press any key to quit");
 
-            while (true)
+            while (!Console.KeyAvailable)
             {
                 if (cpuTemperature.IsAvailable)
                 {
-                    double temperature = cpuTemperature.Temperature.DegreesCelsius;
-                    if (!double.IsNaN(temperature))
+                    var temperature = cpuTemperature.ReadTemperatures();
```
krwq, line 22: I'd recommend to leave the old API alone and throw if there are no thermometrs/data and add new API on top of it (on Linux enumerate just 1)

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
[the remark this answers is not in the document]
pgrawehr: @Ellerbach While I can see that there are newer versions available on the dev servers, I cannot seem to use them. Possibly because none of the never versions have been published to the official nuget feed.

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
```diff
@@ -0,0 +1,740 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
```
krwq, line 3: should Windows implementation of CpuTemperature use this class instead of doing queries directly?

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
```diff
@@ -0,0 +1,740 @@
...
+namespace Iot.Device.OpenHardwareMonitor
+{
+    /// <summary>
+    /// This class connects to a running instance of OpenHardwareMonitor and reads out all available values.
+    /// This works only if OpenHardwareMonitor (https://openhardwaremonitor.org/) is currently running.
+    /// While the tool needs to be run with elevated permissions, the application using this binding does not.
+    /// </summary>
+    public sealed class OpenHardwareMonitor : IDisposable
+    {
+        private static readonly TimeSpan MonitorInterval = TimeSpan.FromMilliseconds(100);
```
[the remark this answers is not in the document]
pgrawehr: There is only one thread for all the monitoring jobs, so all the intervals need to be a multiple of this value. I fixed the calculation now, so it's possible to have arbitrary values. (+ changed to property)

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
pgrawehr: Added optional argument

src/devices/CpuTemperature/CpuTemperature.cs
pgrawehr: The Iot.Devices library has no platform dependency (and shouldn't have one). Therefore the platform test is done at runtime, so that two files don't really add value, I think. 

The API is exactly the same for all cases.

src/devices/CpuTemperature/CpuTemperature.csproj
pgrawehr: See above. There's no platform dependency on this dll.

pgrawehr: Also, any platform dependency introduced at this level would apply to the _source computer_, not the one running it.

src/devices/CpuTemperature/samples/CpuTemperature.Sample.cs
pgrawehr: I think I left the old API in place, and just updated the test to use the new one, but I'll check.

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
pgrawehr: I considered this as well, but since this class requires an external application to be running, I tought that it might be good to not add that dependency to the basic CpuTemperature class. But since the current implementation of CpuTemperature often doesn't work (depending on hardware), it may be better to change it anyway. 

So we have a solution (current implementation of CpuTemperature) that often doesn't work and needs elevation vs a solution that requires external components. Which do you prefer?

krwq: I think whichever one is more useful. We should start with making the code work with good APIs, next steps are perf and size

src/devices/CpuTemperature/samples/CpuTemperature.Sample.cs
krwq: sounds good in that case

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
pgrawehr: Yes, the values returned are always SI units. Which unit belongs to which sensor type is defined in the global _typeMap array.

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
pgrawehr: I'll try to use the OpenHardwareMonitor implementation on Windows now, and use the direct approach only as fallback.

src/devices/OpenHardwareMonitor/OpenHardwareMonitor.cs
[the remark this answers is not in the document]
pgrawehr: There seems to be some sort of support on linux, using mono. I haven't tried that, though. There's an alternate connection api that creates a small HTTP server that could be used, but this would need to be investigated separately.

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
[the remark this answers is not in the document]
pgrawehr: Now switched to the latest version that seems to install directly. I expect a final version to be uploaded to nuget.org when .NET 5.0 ships.

src/devices/CpuTemperature/CpuTemperature.csproj
```diff
@@ -9,5 +9,12 @@
     <Compile Include="CpuTemperature.cs" />
     <None Include="README.md" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="../HardwareMonitor/HardwareMonitor.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Management" Version="5.0.0-preview.5.20224.13" />
```
joperezr, line 17: We don't want to use static versions here, instead we should do what we are doing with with the rest of package dependencies from the framework so that autmoatic infrastructure updates will update the version everytime there is a new one on NuGet.org. See what we do here:

[https://github.com/dotnet/iot/blob/aa2ba37e8eeff3ea1e91c4ba97f83abba637c3ba/eng/Version.Details.xml#L19-L34](https://github.com/dotnet/iot/blob/aa2ba37e8eeff3ea1e91c4ba97f83abba637c3ba/eng/Version.Details.xml#L19-L34)

And here:

[https://github.com/dotnet/iot/blob/aa2ba37e8eeff3ea1e91c4ba97f83abba637c3ba/eng/Versions.props#L7-L10](https://github.com/dotnet/iot/blob/aa2ba37e8eeff3ea1e91c4ba97f83abba637c3ba/eng/Versions.props#L7-L10)

I actually need to update those to point to dotnet/runtime as opposed to dotnet/corefx in order to get the right updates for the versions, so I'll take care of that. Let me know if you want me to take care of the ones you are introducing as well.

src/devices/HardwareMonitor/HardwareMonitor.csproj
```diff
@@ -0,0 +1,18 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="OpenHardwareMonitor.cs" />
+    <Compile Include="SensorType.cs" />
+    <None Include="README.md" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Management" Version="5.0.0-preview.5.20224.13" />
```
joperezr, line 15: See my other comment, basically this versio should be comming from eng\Versions.props

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -25,6 +25,7 @@
     <PackageReference Include="System.IO.Ports" Version="$(SystemIOPortsPackageVersion)" />
     <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
     <PackageReference Include="UnitsNet" Version="$(UnitsNetPackageVersion)" />
+    <PackageReference Include="System.Management" Version="5.0.0-preview.5.20224.13" />
```
joperezr, line 28: See my other comment, basically versions should be comming from Versions.props instead of this project defining them. If you want me to do that let me know. Also, to answer @krwq question: there is no such thing as a platform specific dependency on a NuGet package.

joperezr:
Sorry @pgrawehr I didn't have time yesterday to push the change but I just did, so that should now use the right 5.0 versions of the packages that will ship in 5.0

pgrawehr:
@joperezr No problem, I've got other loose ends to work on ;-) 

Unfortunately, it looks as if you've broken the build now (something about package downgrading). Did you miss an occurence?

joperezr:
Oops, my bad, looking into it now.

joperezr:
I'll rebase to fix the merge conflict

joperezr:
@richlander given you are the nullable expert in the repo, let me know if you have any feedback on the last commit changes: 855a8e6

richlander:
Would be nice to see this marked in some way to remove the `!` when we stop targeting `netcoreapp2.1`. We want as much compiler checking as we can get.

https://github.com/dotnet/iot/commit/855a8e6924c8fa1c60f0939c96c8977317d277a0#diff-962c5fd4c57ced02999e704a7a51c628ec345eacb38dcd0bc3fb83a2d2056eafR461

An `#if` and `#elif` would certainly resolve that.

joperezr (approved):
PR looks good to me, thanks for the contribution @pgrawehr, I'll go ahead and merge once this is green in order to also get the update on the package dependency versions in.

PR #1242, opened 2020-11-02
Simplify license header

richlander:
- We decided that a two line scheme is better.
- It is simply a simplification. It has no actual or intended legal meaning/change.
- Same as https://github.com/dotnet/runtime/pull/38793


Ellerbach (approved):
I was now so used to the 3 lines :-) Great it's simplified!

pgrawehr:
I guess the license header in each file has no legal impact anyway. Whoever uses the code must check the license details anyway. (If in doubt, not specifying a license would mean "no rights granted").

krwq:
LGTM, @richlander why do we need license header at all? Isn't root LICENSE file enough?

joperezr (approved):
Thanks a bunch for doing this @richlander

PR #1243, opened 2020-11-03
Update repo to Nullable Reference Types (NRTs)

richlander:
The intention of this PR is to transition this repo to be Nullable Reference Types (NRT) enabled. That means that all future bindings and changes would need be to be NRT-enabled.

I also updated almost all of the samples to .NET 5.0 and C# 9 (including top-level projects).


krwq:
@richlander I'm ok with locking the repo for couple of days to get this merged, let me know if you have an ETA so that I know not to merge any more PRs

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -1,15 +1,15 @@
...
+    <TargetFrameworks>netcoreapp2.1;net5.0</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <EnableDefaultItems>false</EnableDefaultItems>
     <IsPackable>true</IsPackable>
     <DeviceRoot>$(MSBuildThisFileDirectory)../devices/</DeviceRoot>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
     <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage)</TargetsForTfmSpecificContentInPackage>
-    <LangVersion>8</LangVersion>
+    <LangVersion>9</LangVersion>
```
krwq, line 12: preview is fine with me as well

src/devices/Ads1115/samples/Program.cs
```diff
@@ -9,158 +9,147 @@
 using Iot.Device.Ads1115;
 using UnitsNet;
 
-namespace Ads1115.Samples
+// set I2C bus ID: 1
+// ADS1115 Addr Pin connect to GND
+I2cConnectionSettings settings = new I2cConnectionSettings(1, (int)I2cAddress.GND);
```
krwq, line 14: this got me confused a bit - I forgot we added top-level expressions, neat

src/devices/Apa102/Apa102.cs
```diff
@@ -28,8 +28,7 @@ public class Apa102 : IDisposable
         /// <param name="length">Number of LEDs</param>
         public Apa102(SpiDevice spi, int length)
         {
-            _spi = spi ?? throw new ArgumentNullException(nameof(spi));
```
krwq, line 31: we don't do validation anymore? I thought the [guidelines](https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/api-guidelines/nullability.md#annotation-guidance) said to retain validation:

```
DO NOT remove existing argument validation when annotating existing public/protected APIs. If through the course of annotating it becomes clear that internal/private surface area is unnecessarily checking for nulls when nulls aren't possible (i.e. you found dead code), such use can be removed or replaced by asserts.
```

krwq: (I'm fine either way but would be weird that if people didn't opt in for the NRT feature they would start getting NRE because of the feature which was supposed to prevent them from getting it)

Ellerbach: please keep the validation, one good reason: nanoFramework does not support nullable.

richlander: I wasn't sure about that. Yes, we should follow the guidelines.

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
richlander: I'm good with either. I personally think that the integer value is better if it is correctly descriptive. We can always moved to preview if we start using newer features. That won't be for a while since C# 10 won't be usable for some time.

richlander:
Good call on locking the repo. I agree. We should focus on making this change.

src/devices/Apa102/Apa102.cs
richlander: I fixed this. I also found other cases like this and restored them.

src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardInfo.cs
```diff
@@ -99,7 +99,7 @@ private RaspberryBoardInfo(Dictionary<string, string> settings)
             ProcessorName = _settings.TryGetValue("Hardware", out string? hardware) && hardware is object ? hardware : string.Empty;
 
             if (_settings.TryGetValue("Revision", out string? revision)
-                && !string.IsNullOrEmpty(revision)
+                && revision is { Length: > 0 }
```
krwq, line 102: I personally find this less readable

richlander: We can certainly change it back.

I like the new syntax for three reason:

- It is centered on the object, as opposed to a helper. I do, however, agree that the name of helper very clearly describes the check being performed. This is clearly the reason why people prefer this approach. The mental cost to figure out the nature of the check is 0.
- The syntax is symmetric with `revision is object` (or `revision is != null`). The `{ }` is the same as and takes the place of `object` in the check, with the added benefit that you can perform property tests. I think this becomes more natural when you start using `is` generally instead of `==` and `!=`. Related: https://twitter.com/jaredpar/status/1115019017297596416.
- It is shorter.

PR #1246, opened 2020-11-03
Add ! operator to PinValue

pgrawehr:
Split away some parts of larger PRs, so that hopefully we can merge quicker.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/PinValue.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/PinValue.cs
@@ -82,6 +82,21 @@ namespace System.Device.Gpio
         /// </summary>
         public static bool operator !=(PinValue a, PinValue b) => !a.Equals(b);
 
+        /// <summary>
+        /// Returns the inverse of the value
+        /// </summary>
+        /// <param name="v">Input value</param>
+        /// <returns>High, when the input is low, low when the input is high</returns>
+        public static PinValue operator !(PinValue v)
+        {
+            if (v._value == 0)
+            {
+                return High;
+            }
+
+            return Low;
+        }
+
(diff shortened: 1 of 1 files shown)


src/System.Device.Gpio/System/Device/Gpio/PinValue.cs
```diff
@@ -82,6 +82,21 @@ public override bool Equals(object? obj)
...
+        /// <returns>High, when the input is low, low when the input is high</returns>
+        public static PinValue operator !(PinValue v)
+        {
+            if (v._value == 0)
+            {
+                return High;
+            }
+
+            return Low;
+        }
```
krwq, line 98:
```suggestion
        public static PinValue operator !(PinValue v) => (v._value == 0) ? High : Low;
```

krwq (approved):
LGTM

krwq:
Thanks, @joperezr @Ellerbach since this is technically public API we require second approver, PTAL

Ellerbach (approved):
I prefer the short form

src/System.Device.Gpio/System/Device/Gpio/PinValue.cs
pgrawehr: Personally, I prefer the long form (especially, if it includes logic, like here). Are there any recommendations on this?

joperezr (approved):
The change itself LGTM. Seems like we currently don't have unit tests for this struct but we should definitely add coverage for it. If you could add some unit tests for them on src/System.Device.Gpio.Tests project that would be ideal, if not please log an issue so that we add coverage soon.

pgrawehr:
@joperezr : Just setting up a bunch of unit tests. I'll include this as well.

PR #1254, opened 2020-11-04
Unseal GpioController

pgrawehr:
As discussed in several tickets and PRs already: this unseals the GpioController and changes its public API to virtual methods. 

In addition, I fixed the PinNumberingScheme.Board use case. Now all methods do a proper translation to the logical scheme the driver is using. The bug #974 is still open, though. This requires further thought and possibly an interface change. 

Unit testing for all methods provided as well (using a bit of a hacky approach to mocking, but there's no other way given that all the methods of the GpioDriver class are internal protected). At least the MockableGpioDriver class is reusable.

### DIFF
--- a/src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
@@ -1,0 +1,307 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using Iot.Device.Mcp25xxx.Register.MessageReceive;
+using Moq;
+using Xunit;
+
+namespace System.Device.Gpio.Tests
+{
+    public class GpioControllerSoftwareTests : IDisposable
+    {
+        private Mock<MockableGpioDriver> _mockedGpioDriver;
+
+        public GpioControllerSoftwareTests()
+        {
+            _mockedGpioDriver = new Mock<MockableGpioDriver>(MockBehavior.Default);
+            _mockedGpioDriver.CallBase = true;
+        }
+
+        public void Dispose()
+        {
+            _mockedGpioDriver.VerifyAll();
+        }
+
+        [Fact]
+        public void PinCountReportedCorrectly()
+        {
+            var ctrl = new GpioController(PinNumberingScheme.Logical, _mockedGpioDriver.Object);
+            Assert.Equal(28, ctrl.PinCount);
+        }
+
+        [Fact]
+        public void OpenTwiceThrows()
+        {
+            var ctrl = new GpioController(PinNumberingScheme.Logical, _mockedGpioDriver.Object);
+            _mockedGpioDriver.Setup(x => x.OpenPinEx(1));
+            _mockedGpioDriver.Setup(x => x.ClosePinEx(1));
+            ctrl.OpenPin(1);
+            Assert.Throws<InvalidOperationException>(() => ctrl.OpenPin(1));
+            ctrl.ClosePin(1);
+            Assert.Throws<InvalidOperationException>(() => ctrl.ClosePin(1));
+        }
+
+        [Fact]
+        public void WriteInputPinThrows()
+        {
+            _mockedGpioDriver.Setup(x => x.OpenPinEx(1));
+            _mockedGpioDriver.Setup(x => x.IsPinModeSupportedEx(1, It.IsAny<PinMode>())).Returns(true);
+            _mockedGpioDriver.Setup(x => x.SetPinModeEx(1, It.IsAny<PinMode>()));
+            _mockedGpioDriver.Setup(x => x.GetPinModeEx(1)).Returns(PinMode.Input);
+            var ctrl = new GpioController(PinNumberingScheme.Logical, _mockedGpioDriver.Object);
+
+            ctrl.OpenPin(1, PinMode.Input);
+            Assert.Throws<InvalidOperationException>(() => ctrl.Write(1, PinValue.High));
+        }
+
+        [Fact]
+        public void GpioControllerCreateOpenClosePin()
+        {
+            _mockedGpioDriver.Setup(x => x.OpenPinEx(1));
+            _mockedGpioDriver.Setup(x => x.IsPinModeSupportedEx(1, PinMode.Output)).Returns(true);
+            _mockedGpioDriver.Setup(x => x.GetPinModeEx(1)).Returns(PinMode.Output);
+            _mockedGpioDriver.Setup(x => x.WriteEx(1, PinValue.High));
+            _mockedGpioDriver.Setup(x => x.ClosePinEx(1));
+            var ctrl = new GpioController(PinNumberingScheme.Logical, _mockedGpioDriver.Object);
+            Assert.NotNull(ctrl);
+            ctrl.OpenPin(1, PinMode.Output);
+            Assert.True(ctrl.IsPinOpen(1));
+            ctrl.Write(1, PinValue.High);
+            ctrl.ClosePin(1);
+            Assert.False(ctrl.IsPinOpen(1));
+        }
+
+        [Fact]
+        public void IsPinModeSupported()
+        {
+            _mockedGpioDriver.Setup(x => x.IsPinModeSupportedEx(1, PinMode.Input)).Returns(true);
+            var ctrl = new GpioController(PinNumberingScheme.Logical, _mockedGpioDriver.Object);
+            Assert.NotNull(ctrl);
+            Assert.True(ctrl.IsPinModeSupported(1, PinMode.Input));
+        }
+
+        [Fact]
+        public void GetPinMode()
+        {
+            _mockedGpioDriver.Setup(x => x.OpenPinEx(1));
+            _mockedGpioDriver.Setup(x => x.GetPinModeEx(1)).Returns(PinMode.Output);
+            var ctrl = new GpioController(PinNumberingScheme.Logical, _mockedGpioDriver.Object);
+            Assert.NotNull(ctrl);
+            // Not open
+            Assert.Throws<InvalidOperationException>(() => ctrl.GetPinMode(1));
+            ctrl.OpenPin(1);
+            Assert.Equal(PinMode.Output, ctrl.GetPinMode(1));
+        }
+
+        [Fact]
+        public void UsingBoardNumberingWorks()
+        {
+            // Our mock driver maps physical pin 2 to logical pin 1
+            _mockedGpioDriver.Setup(x => x.ConvertPinNumberToLogicalNumberingSchemeEx(2)).Returns(1);
+            _mockedGpioDriver.Setup(x => x.OpenPinEx(1));
+            _mockedGpioDriver.Setup(x => x.SetPinModeEx(1, PinMode.Output));
+            _mockedGpioDriver.Setup(x => x.IsPinModeSupportedEx(1, PinMode.Output)).Returns(true);
+            _mockedGpioDriver.Setup(x => x.GetPinModeEx(1)).Returns(PinMode.Output);
+            _mockedGpioDriver.Setup(x => x.WriteEx(1, PinValue.High));
+            _mockedGpioDriver.Setup(x => x.ReadEx(1)).Returns(PinValue.High);
+            _mockedGpioDriver.Setup(x => x.ClosePinEx(1));
+            var ctrl = new GpioController(PinNumberingScheme.Board, _mockedGpioDriver.Object);
+            ctrl.OpenPin(2, PinMode.Output);
+            ctrl.Write(2, PinValue.High);
+            Assert.Equal(PinValue.High, ctrl.Read(2));
+            ctrl.ClosePin(2);
+            ctrl.Dispose();
+        }
+
+        [Fact]
+        public void UsingLogicalNumberingDisposesTheRightPin()
+        {
+            _mockedGpioDriver.Setup(x => x.OpenPinEx(1));
+            _mockedGpioDriver.Setup(x => x.ClosePinEx(1));
+            _mockedGpioDriver.Setup(x => x.IsPinModeSupportedEx(1, PinMode.Output)).Returns(true);
+            _mockedGpioDriver.Setup(x => x.GetPinModeEx(1)).Returns(PinMode.Output);
+            var ctrl = new GpioController(PinNumberingScheme.Logical, _mockedGpioDriver.Object);
+            ctrl.OpenPin(1, PinMode.Output);
+            ctrl.Write(1, PinValue.High);
+            // No close on the pin here, we want to check that the Controller's Dispose works correctly
+            ctrl.Dispose();
+        }
+
+        [Fact]
+        public void UsingBoardNumberingDisposesTheRightPin()
+        {
+            // Our mock driver maps physical pin 2 to logical pin 1
+            _mockedGpioDriver.Setup(x => x.ConvertPinNumberToLogicalNumberingSchemeEx(2)).Returns(1);
+            _mockedGpioDriver.Setup(x => x.OpenPinEx(1));
+            _mockedGpioDriver.Setup(x => x.SetPinModeEx(1, PinMode.Output));
+            _mockedGpioDriver.Setup(x => x.ClosePinEx(1));
+            _mockedGpioDriver.Setup(x => x.IsPinModeSupportedEx(1, PinMode.Output)).Returns(true);
+            var ctrl = new GpioController(PinNumberingScheme.Board, _mockedGpioDriver.Object);
+            ctrl.OpenPin(2, PinMode.Output);
+            // No close on the pin here, we want to check that the Controller's Dispose works correctly
+            ctrl.Dispose();
+        }
+
+        [Fact]
+        public void CallbackOnEventWorks()
+        {
+            // Our mock driver maps physical pin 2 to logical pin 1
+            _mockedGpioDriver.Setup(x => x.OpenPinEx(1));
+            _mockedGpioDriver.Setup(x => x.AddCallbackForPinValueChangedEventEx(1,
+                PinEventTypes.Rising, It.IsAny<PinChangeEventHandler>()));
+            var ctrl = new GpioController(PinNumberingScheme.Logical, _mockedGpioDriver.Object);
+            ctrl.OpenPin(1); // logical pin 1 on our test board
+            bool callbackSeen = false;
+            PinChangeEventHandler eventHandler = (sender, args) =>
+            {
+                callbackSeen = true;
+                Assert.Equal(1, args.PinNumber);
+                Assert.Equal(PinEventTypes.Falling, args.ChangeType);
+            };
+
+            ctrl.RegisterCallbackForPinValueChangedEvent(1, PinEventTypes.Rising, eventHandler);
+
+            _mockedGpioDriver.Object.FireEventHandler(1, PinEventTypes.Falling);
+
+            Assert.True(callbackSeen);
+
+            ctrl.UnregisterCallbackForPinValueChangedEvent(1, eventHandler);
+        }
+
+        [Fact]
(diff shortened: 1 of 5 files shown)


src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
```diff
@@ -0,0 +1,304 @@
+﻿using System;
```
krwq, line 1: copyright header

src/System.Device.Gpio.Tests/PinValueTests.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            Assert.True(PinValue.High == 1);
+            Assert.True(PinValue.Low == 0);
+        }
+
+        [Fact]
+        public void VerifyInequality()
+        {
+            Assert.True(PinValue.High != 0);
+            Assert.True(PinValue.Low != 1);
```
krwq, line 28: verify also != 2

src/System.Device.Gpio.Tests/PinValueTests.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        [Fact]
+        public void Inverse()
+        {
+            Assert.Equal(PinValue.Low, !PinValue.High);
```
krwq, line 34: also `Assert.Equal(PinValue.High, !PinValue.Low);`

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -59,25 +69,26 @@ public GpioController(PinNumberingScheme numberingScheme, GpioDriver driver)
...
 
         /// <summary>
         /// Gets the logical pin number in the controller's numbering scheme.
         /// </summary>
-        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        /// <param name="pinNumber">The pin number</param>
+        /// <param name="givenScheme">The scheme for the pin number</param>
         /// <returns>The logical pin number in the controller's numbering scheme.</returns>
-        private int GetLogicalPinNumber(int pinNumber)
+        protected virtual int GetLogicalPinNumber(int pinNumber, PinNumberingScheme givenScheme)
```
krwq, line 80: why cannot this use NumberingScheme property instead of argument?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -310,11 +333,17 @@ public void UnregisterCallbackForPinValueChangedEvent(int pinNumber, PinChangeEv
...
 
-        private void Dispose(bool disposing)
+        /// <summary>
+        /// Disposes this instance and closes all open pins associated with this controller.
+        /// </summary>
+        /// <param name="disposing">True to dispose all instances, false to dispose only unmanaged resources</param>
+        protected virtual void Dispose(bool disposing)
         {
-            foreach (int pin in _openPins)
+            var tempList = new List<int>(_openPins); // Because ClosePin modifies this list
```
krwq, line 342: should you instead create ClosePinInternal or similar method which doesn't modify the list?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -353,16 +383,11 @@ public void Read(Span<PinValuePair> pinValuePairs)
...
-        /// <param name="numberingScheme">The numbering scheme used to represent pins provided by the controller.</param>
-        public GpioController(PinNumberingScheme numberingScheme)
-            : this(numberingScheme, GetBestDriverForBoard())
-        {
-        }
-
-        private static GpioDriver GetBestDriverForBoard()
+        /// <returns>An instance of a GpioDriver that best matches the current hardware</returns>
+        /// <exception cref="PlatformNotSupportedException">No matching driver could be found</exception>
+        protected static GpioDriver GetBestDriverForBoard()
```
krwq, line 390: I think perhaps this should be just GpioDriver.Create

krwq: cc: @joperezr

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: Because it is sometimes required to do a transform from a scheme that is not necessarily the default, see the `Dispose` method. It would also be possible (and has been suggested already) to add overloads to `Open(), Write() or Read()` that take the numbering scheme as extra argument, so that the default can be overriden for individual calls.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: Not so easy, because ClosePin is virtual and any inheritors shouldn't have to worry about a distinction there.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: How would we know which driver to create?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: Discussed directly and proposed alternate approach

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -353,15 +382,10 @@ public void Read(Span<PinValuePair> pinValuePairs)
...
         /// </summary>
-        /// <param name="numberingScheme">The numbering scheme used to represent pins provided by the controller.</param>
-        public GpioController(PinNumberingScheme numberingScheme)
-            : this(numberingScheme, GetBestDriverForBoard())
-        {
-        }
-
+        /// <returns>An instance of a GpioDriver that best matches the current hardware</returns>
+        /// <exception cref="PlatformNotSupportedException">No matching driver could be found</exception>
         private static GpioDriver GetBestDriverForBoard()
```
pgrawehr, line 389: Having this private prevents adding new APIs, but requires copy-pasting the implementation to the Board infrastructure. That's not so big of a deal for now, except that it requires a few ugly try-catch attempts in testing which driver is the right one. 
I think we need the identification logic to become public somehow (or at least provide each driver with a IsCompatible check)

krwq: Let's refrain from fixing this in this PR and perhaps suggest GpioDriver.Create in a separate PR/issue

pgrawehr: I agree to do that later. We should just not forget about it.

krwq:
@pgrawehr can you squash and rebase to take nullability changes?

src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
```diff
@@ -0,0 +1,307 @@
...
+            Assert.NotNull(ctrl);
+            ctrl.OpenPin(1, PinMode.Input);
+
+            var result = ctrl.WaitForEvent(1, PinEventTypes.Falling | PinEventTypes.Rising, TimeSpan.FromSeconds(0.01));
+            Assert.False(result.TimedOut);
+            Assert.Equal(PinEventTypes.Falling, result.EventTypes);
+        }
+
+        // TODO: This is still broken. See #974
+        ////[Fact]
```
Ellerbach, line 283: Only 2 // should be enougth :-) Not blocking, and thanks for adding the comment that this code even if dead should stay here

Ellerbach (approved):
Looks very good to me. thanks!

src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
pgrawehr: If only using two // you get StyleCop errors, because then the comment must be immediatelly followed by code. And a /// comment on the next function (if any) must not, on the other hand, be preceeded by a // comment. So //// is the only way to properly comment out an entire function.

pgrawehr:
> 
> 
> @pgrawehr can you squash and rebase to take nullability changes?

Yes, will do.

pgrawehr:
@joperezr Can you please restart the build here?

src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
krwq: /*perhaps there are other ways*/

krwq: (I'm fine with quadra comments though)

src/System.Device.Gpio.Tests/MockableGpioDriver.cs
```diff
@@ -0,0 +1,109 @@
+﻿using System;
```
krwq, line 1: copyright header missing

krwq (approved):

colombod:
This pr looks very good and promising

colombod:
merging soon?

pgrawehr:
@colombod Yes, this is just awaiting another quick review from joperezr, because he's most familiar with these pieces.

colombod:
Perfect!

krwq:
@joperezr have time to review?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -75,14 +85,23 @@ private int GetLogicalPinNumber(int pinNumber)
...
-            int logicalPinNumber = GetLogicalPinNumber(pinNumber);
-            if (_openPins.Contains(logicalPinNumber))
+            if (IsPinOpen(pinNumber))
             {
-                throw new InvalidOperationException($"Pin {logicalPinNumber} is already open.");
+                throw new InvalidOperationException($"Pin {pinNumber} is already open.");
             }
 
+            OpenPinCore(pinNumber);
+            _openPins.Add(pinNumber);
```
joperezr, line 94: Should the opertaion of adding it to the openPins list live instead on OpenPinCore method? Or is the idea of having it here so that this part will never be overriden by subclasses?

joperezr:
>             _driver.SetPinMode(logicalPinNumber, mode);

Is there a reason why we added the OpenPinCore and ClosePinCore (where we are doing the validation on the non-virtual ones) in some cases but not added a SetPinModeCore as well that did the validation above? I'm basically just wondering if we should've just made OpenPin and ClosePin virtual as well and remove the Core ones.

---
Refers to: src/System.Device.Gpio/System/Device/Gpio/GpioController.cs:161 in 41b1c1a. [](commit_id = 41b1c1a5e60e7c8cb3c432608423e8fdf13a1e72, deletion_comment = False)

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -166,7 +193,7 @@ public bool IsPinOpen(int pinNumber)
         /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
         /// <param name="mode">The mode to check.</param>
         /// <returns>The status if the pin supports the mode.</returns>
-        public bool IsPinModeSupported(int pinNumber, PinMode mode)
+        public virtual bool IsPinModeSupported(int pinNumber, PinMode mode)
```
joperezr, line 196: The implementation of this is using _openPins list which is private (instead of protected) as well as _driver, which is also private. Should those two fields be protected instead so that subclasses can use them in their implementation?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -323,6 +351,7 @@ private void Dispose(bool disposing)
         public void Dispose()
         {
             Dispose(true);
+            GC.SuppressFinalize(this);
```
joperezr, line 354: NIT: I know this is the standard implementation of the Dispose pattern, but the class has no finalizer so this will not have any effect. I would just remove the line.

joperezr:
Left some small comments/questions but this is mostly ready to go as soon as those points are finalized.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: Yes, that was exactly the reason to do it here. Getting this wron in an overriden class would confuse the drivers significantly. And the list of open pins is not public.

pgrawehr:
> 
> 
> > ```
> >         _driver.SetPinMode(logicalPinNumber, mode);
> > ```
> 
> Is there a reason why we added the OpenPinCore and ClosePinCore (where we are doing the validation on the non-virtual ones) in some cases but not added a SetPinModeCore as well that did the validation above? I'm basically just wondering if we should've just made OpenPin and ClosePin virtual as well and remove the Core ones.
> 
> Refers to: src/System.Device.Gpio/System/Device/Gpio/GpioController.cs:161 in [41b1c1a](https://github.com/dotnet/iot/commit/41b1c1a5e60e7c8cb3c432608423e8fdf13a1e72). [](commit_id = [41b1c1a](https://github.com/dotnet/iot/commit/41b1c1a5e60e7c8cb3c432608423e8fdf13a1e72), deletion_comment = False)

Except for the initial validation that the pin is open, the other methods don't do anything except calling into the driver, so adding the Core variant seems a bit to much, but I could life with that as well. 

> The implementation of this is using _openPins list which is private (instead of protected) as well as _driver, which is also private. Should those two fields be protected instead so that subclasses can use them in their implementation?

I think this is not necessary, since the (base) implementation of `IsPinOpen()` is always accessible and returns the information you need. Making the _openPins list protected would allow derived classes to tamper with it. And `IsPinModeSupported()` might on some hardware even work without the pin being open. 

The `_driver` member can^t be made protected, since a derived class can't do anything with it. There are no public members on `GpioDriver`. If required, a derived class would keep a reference to its special driver instead.

joperezr:
Left one last small comment, but that and the finalizer are the only remaining things to clear out.

pgrawehr:
@joperezr : Removed the finalizer line, but which other comment do you mean?

joperezr (commented):
Sorry forgot to click send 😀

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
colombod: I also think that we need to have access to the collection maybe via protected methods instead of giving access to the collection itself? But is we can override things like the pincount property and ispinopen there is also the need to have a way to check the collection

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -138,14 +166,14 @@ public void SetPinMode(int pinNumber, PinMode mode)
...
-            int logicalPinNumber = GetLogicalPinNumber(pinNumber);
-            if (!_openPins.Contains(logicalPinNumber))
+            if (!IsPinOpen(pinNumber))
             {
-                throw new InvalidOperationException($"Can not get the mode of pin {logicalPinNumber} because it is not open.");
+                throw new InvalidOperationException($"Can not get the mode of pin {pinNumber} because it is not open.");
             }
 
+            int logicalPinNumber = GetLogicalPinNumber(pinNumber);
             return _driver.GetPinMode(logicalPinNumber);
```
colombod, line 177: Maybe all this drive access can be exposed via protected methods like *FromDriver, eg `GetPinModeFromDriver`, `OpenPinFromDriver` or at least give access to the driver instance via a protected property?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -12,7 +12,7 @@ namespace System.Device.Gpio
     /// <summary>
     /// Represents a general-purpose I/O (GPIO) controller.
     /// </summary>
-    public sealed class GpioController : IDisposable
+    public class GpioController : IDisposable
```
colombod, line 15: I don't think that this class can satisfy the scenario that I am using in the pi-top libraries. I need a controller that is just a proxy to a real controller. With this implementation anything conforming with the controller interface will have to call base class constructors that will try to create a drive. This part of the class contracts seems a little bit too restrictive.

colombod:
> Left one last small comment, but that and the finalizer are the only remaining things to clear out.

I have some questions about constructor requirements  too

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
krwq: can you pass null driver and override all methods?

colombod: sounds little bit aggressive if null is allowed then then internal calls should be something like `bool TryOpenPin()`, that should wrap the driver access. Such protected methods will then potentially address the previous questions about the driver.

I like your suggestion, if we can then protected the access to the driver api via a Try* set of methods that handles the null driver that would be an easier pattern for implementors to adopt.

pgrawehr: @colombod Can you elaborate a bit more on your use case?

It is possible to pass null as the driver, or a DummyDriver() [will be provided in one of my next PRs]. All methods accessing the `_driver` member are virtual, so if you override all of them (without calling the base implementation, of course), you should be fine.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: That adds a lot of complexity to the interface. Why would you need that?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
colombod: in the pi-top libraries we use a gpio controller factory, this returns a wrapper that tracks it's onw pin open/close logic and also events subscriptions. All work si delegated to the concrete controller.
on dispose such controller "projections" dispose all things that where opened via the proxy. That controller will require to have no driver and track the pins open requests. If passing null is safe in this PR then is good for me. And will be able to adopt this new class and finally get rid of the interfaces we introduced. This is the [shared controller class](https://github.com/pi-top/pi-top-4-.NET-SDK/blob/master/src/PiTop.Abstractions/SharedGpioController.cs)

pgrawehr: Yea, I think this should all fit now. You do not need any of your intermediate classes. Even `SharedGpioController` doesn't seem to do anything more than what is aleady implemented.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -154,10 +182,9 @@ public PinMode GetPinMode(int pinNumber)
         /// </summary>
         /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
         /// <returns>The status if the pin is open or closed.</returns>
-        public bool IsPinOpen(int pinNumber)
+        public virtual bool IsPinOpen(int pinNumber)
```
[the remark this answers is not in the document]
pgrawehr: I guess either way is ok.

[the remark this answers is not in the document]
joperezr: I would rather keep it non-overridable for now if we don't see a good reason to do it, and in the future change it if we think it is necessary. That would be the non-breaking way to do it. I think if we allow people to override this, they will 100% of the time either call base implementation only, or try to do something different and get it wrong.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
joperezr: Yeah, I agree that those would just add complexity and don't see a good use case for them.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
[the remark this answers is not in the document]
pgrawehr: done

joperezr (approved):

PR #1257, opened 2020-11-05
Fix Sense HAT sample

### LINKED ISSUE #1248: SenseHat sample throws runtime exceptions related to formatting
I'm currently writing a quickstart that uses this sample, but it's currently broken due to formatting exceptions at runtime.

I've already fixed it locally, but the file the fix is in is getting renamed in https://github.com/dotnet/iot/pull/1243 so I'm going to wait until that's merged before I submit a PR so as not to create a merge issue.

CamSoper:
Fixes #1248 

The values from the device binding are already formatted, so formatting them as numbers was throwing an exception. This PR removes the extra formatting.

### DIFF
--- a/src/devices/SenseHat/samples/Program.cs
+++ b/src/devices/SenseHat/samples/Program.cs
@@ -8,14 +8,6 @@ using Iot.Device.Common;
 using Iot.Device.SenseHat;
 using UnitsNet;
 
-// Sample for each device separately
-// LedMatrix.Run();
-// Joystick.Run();
-// AccelerometerAndGyroscope.Run();
-// Magnetometer.Run();
-// TemperatureAndHumidity.Run();
-// PressureAndTemperature.Run();
-
 // set this to the current sea level pressure in the area for correct altitude readings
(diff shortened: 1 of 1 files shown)


joperezr (approved):
Thanks for fixing @CamSoper!

richlander:
What is the reason for removing the units, like `hPa`? Surely that wasn't throwing an exception?

joperezr:
I assumed that it was because the formatted string already has them, but if not, I do agree we should add them back.

CamSoper:
I'll be honest, I was being lazy because I didn't want to do any debugging. I've had a long week. 🤷🏻‍♂️

The specific problem was that at some point the binding was changed to return `Pressure` and `Altitude` objects instead of numerical values and the sample wasn't changed to match.

I've noticed similar problems in the BME280 sample. I'm guessing there was a breaking change in UnitsNet or something at some point?

Anyway, it's *actually* fixed the right way now. Sorry about that, mea culpa.

richlander:
That looks good to me now. Thanks!

richlander (approved):

PR #1258, opened 2020-11-06
Use records with Bh1745

richlander:
First test of using records in this codebase. This binding seemed like a perfect use case for records.

I made a few choices:

- Defined the record to the primary code file. It doesn't warrant its own file. IMO, this is a best practice.
- Defined the record first (before the primary type). If there were >2 records, I'd pull them out into their own file.
- Defined `IsExternalInit` as part of the `netstandard2.0` asset (required to use records pre .NET 5.0).
- Unrelated, I also moved the (single) extension to the main file. That seemed like a poor practice, IMO.

These are all a first take. What do you think?

`///` doc comments don't currently work for records due to https://github.com/dotnet/roslyn/issues/44571. I don't think this should stop us from getting started with records. 

Going forward, I think we should convert a small set of data-oriented classes to records, and see if we get any feedback on their usage. For this first case, I've gone with the (very pretty) immutable version of records. That seems to fit this case well. We may find that we want to enable mutable properties for some cases. That said, the `with` syntax is pretty compelling.


src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -39,6 +39,10 @@
     </Reference>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
```
joperezr, line 42: Shouldn't this condition be the other way around? Right now it is compiling the file on net5.0 but not on NS2.0

src/devices/Bh1745/Bh1745.cs
```diff
@@ -35,10 +42,11 @@ public class Bh1745 : IDisposable
         public Bh1745(I2cDevice device)
         {
             _i2cDevice = device;
-            ChannelCompensationMultipliers = new ChannelCompensationMultipliers
-            {
-                Red = 2.2, Green = 1.0, Blue = 1.8, Clear = 10.0
-            };
+            ChannelCompensationMultipliers = new (
```
joperezr, line 45: While I don't have a big preference, I suppose that object initializer still works with records right? Like:
```c#
ChannelCompensationMultipliers = new ()
            {
                Red = 2.2, Green = 1.0, Blue = 1.8, Clear = 10.0
            };
```

src/System.Device.Gpio/System.Device.Gpio.csproj
joperezr: Discussed offline, we will instead move this type to iot.device.bindings and the binding itself and try to make it internal.

src/devices/Bh1745/Bh1745.cs
richlander: If I do that, then the constructor isn't satisfied, and then I cannot use the pretty records syntax.

src/devices/Bh1745/Bh1745.csproj
```diff
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <Compile Include="*.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)/../Common/System/Runtime/CompilerServices/IsExternalInit.cs"/>
```
joperezr, line 11:
```suggestion
    <Compile Include="$(MSBuildThisFileDirectory)/../Common/System/Runtime/CompilerServices/IsExternalInit.cs">
      <Link>Common/System/Runtime/CompilerServices/IsExternalInit.cs</Link>
    </Compile>
```
Nit: Add a Link here for VS Experience.

richlander: If you switch the order of the TFMs in `Bh1745.csproj` or load these projects on a machine within .NET 5.0, you will get the following experience, with `IsExternalInit.cs` listed. Just ignore it.

image

joperezr (approved):

PR #1277, opened 2020-11-11
Utilize .NET Core 3.0+ System.Net.Sockets.Socket in SocketCan

thecatkitty:
**This change needs retargeting of the project because it relies on `AddressFamily.ControllerAreaNetwork` which has been added in .NET Core 3.0**

Currently, reading CAN frames in a loop will block the thread until some frame arrives, thus making it hard to gracefully exit the program.

This change enables (opt-in) non-blocking read of CAN frames by adding `Blocking` property. When set `true`, if there is no frame to read, `TryReadFrame` returns `false` instead of blocking the thread.

~~It utilizes `SetLastError` to prepare `SocketException` on `read` and `write` syscalls. In `CanRaw` it checks if the exception error code is `WouldBlock` and (in order to preserve API compatibility) packs it into `IOException` otherwise.~~

It does so by utilizing .NET Core `System.Net.Sockets.Socket` class with its `Blocking` property.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1277) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1277) before we can accept your contribution.<br/><br/>:x: thecatkitty [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1277)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1277) it.</sub>

src/devices/SocketCan/Interop.cs
```diff
@@ -61,7 +67,7 @@ public static unsafe void Write(SafeHandle handle, ReadOnlySpan<byte> buffer)
                     int bytesWritten = Interop.SocketWrite((int)handle.DangerousGetHandle(), b, buffer.Length);
                     if (bytesWritten < 0)
                     {
-                        throw new IOException("`write` operation failed");
+                        throw new SocketException();
```
krwq, line 70: can we instead create TryWrite to avoid throwing and catching? Also soon I believe we should be able to reuse socket class.

Even better option might be perhaps to use Socket class directly and remove most of the interop code since https://github.com/dotnet/runtime/issues/862 and https://github.com/dotnet/runtime/issues/29684 are now closed.

krwq: (note https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.addressfamily?view=net-5.0 now has ControllerAreaNetwork value which we can use)

thecatkitty: I am aware of `ControllerAreaNetwork` value. If we want to use it, I can go with replacing Interop with the standard `Socket` instead of current *manual* approach.

krwq: It would be perfect! I've been asking for these changes when I wrote this code. Please try not to remove any public surface when doing it though (new APIs are fine)

thecatkitty: Okay, I've started implementing :)

But the only issue I have encountered is that in order to use `AddressFamily.ControllerAreaNetwork`, we need to target *at least* `netcoreapp3.0`...

krwq: I think we should eventually re-target as we can't be stuck with old version forever 😄 Before you proceed though let's double check with @joperezr @richlander @Ellerbach they don't have any objections to that so that your work doesn't go in vain 😃

Ellerbach: I'm open for retargeting. Rational:
- If anyone needs an update on an "older" version, there is always the source code, there is always git history
- it does allow to have access for advance bindings to more features like here and that's not blocking for the ones which should be able to get ported to nano and other smaller platforms
- it's like OS and tools, install updates :-) So move to the last version :-)

thecatkitty: We can either retarget to `netcoreapp3.1` **or** become `net5.0`-only.

(Fun fact, the last time I've checked .NET Core installation instructions was 3 days ago. 2 days ago .NET 5.0 has been released.)

src/devices/SocketCan/Interop.cs
```diff
@@ -30,106 +29,54 @@ internal class Interop
...
-        private static extern int CloseSocket(int fd);
-
-        [DllImport("libc", EntryPoint = "write", CallingConvention = CallingConvention.Cdecl)]
-        private static unsafe extern int SocketWrite(int fd, byte* buffer, int size);
-
-        [DllImport("libc", EntryPoint = "read", CallingConvention = CallingConvention.Cdecl)]
-        private static unsafe extern int SocketRead(int fd, byte* buffer, int size);
-
         [DllImport("libc", EntryPoint = "setsockopt", CallingConvention = CallingConvention.Cdecl)]
         private static unsafe extern int SetSocketOpt(int fd, int level, int optName, byte* optVal, int optlen);
```
thecatkitty, line 37: As much as I wish that these `extern`s were gone, we can't use `Socket.IOControl` (it is almost Windows-only) and `Socket.SetSocketOption` (in `SocketOptionLevel` there's no definition for CAN bus, and PAL verifies provided values, so we cannot write them manually).

krwq: can we convert rest of the code but leave this one? That still should be much better than re-implementing socket 😄

src/devices/SocketCan/SocketCan.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net5.0;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net5.0;netcoreapp3.1</TargetFrameworks>
```
richlander, line 4: Projects need to maintain these two targets: 5.0 or 2.1. 

That's what IoT.Device.Bindings targets: https://github.com/dotnet/iot/blob/master/src/Iot.Device.Bindings/Iot.Device.Bindings.csproj

Certainly, a project could add a third target, but it seems like a very niche need. In the general case, bindings should provide an implementation within the oldest targeted version and feel free to light up for the newest one.

Fair?

src/devices/SocketCan/Interop.cs
krwq: let's start with minimum you require to complete this. Note that projects for bindings can target different version than Iot.Device.Bindings, this way we can ensure that core nano bindings will still work

src/devices/SocketCan/SocketCan.csproj
krwq: can we dump 2.1 support for vNext and start with 3.1? I'd prefer we didn't get stuck in the past. Perhaps we should consider splitting the Iot.Device.Bindings into individual packages so this problem is scoped

richlander: Given: .NET Core 2.1 goes EOL in August 2021 ([source](https://github.com/dotnet/core/blob/master/microsoft-support.md)].

I have a new idea for how to approach this.

- IoT.Device.Binding should target 2.1, 3.1 and 5.0, starting now.
- New bindings are free to pick their lowest supported version (with some justification as to why).
- In August, we drop the 2.1 build / package asset as a kind of non-event.

The benefit of this plan is that the 2.1 -> 3.1 transition (as the lowest supported version) is a long process, not a single moment. It enables us to take advantage of new functionality sooner, while offering 2.1 support until the end of that release's support period.

joperezr: I would like to discuss that a bit further before we make any decision here. Even if 2.1 goes EOL, we are a library which is why in principle we want to continue targeting the lowest possible framework whenever possible. I also don't think that bringing in a build for 3.1 would be beneficial unless the 3.1 implementation is different than the 5.0 one. Basic principles should be:
- Keep targeting the lowest possible framework you can, which today is 2.1
- Only add new TFM targets on Iot.Device.Bindings if we have a) bindings that can only be supported on higher TFMs than the ones we compile against today, or b) bindings that will have a drasticly different implementation on a higher TFM given to new types that are available.

Given that, and in order to take this PR, I would probably just add a new 3.1 target on Iot.Device.Bindings for now, and condition all of the code added in this PR to be 3.1 specific, and keep the 2.1 code as it was before.

krwq: @joperezr people really needing 2.1 can use older version of the package, the old code continues to work. If you write new code you should use newer stack and newer version of the package. I'd rather announce up to which version we support 2.1

joperezr: We'll discuss this further in our next triage meeting. There are no current downsides of multi-targeting, and there are MANY downsides for dropping support.

krwq:
@thecatkitty here are the conclusions:
- we will target netcoreapp2.1;netcoreapp3.1;net5.0 here but we will need for @joperezr to prepare the build for multi-targetting (ETA next week) and you will need to rebase after that
- you should split the code into the old and new:
  - put all old code (which is not shared with new code, shared code leave as is) into 2.1 folder somewhere under source (copy & paste) - this code should not be changed further
  - change old code to use new APIs in Socket
  - csproj file (both in Iot.Device.Binding and the one for SocketCan) should be modified so that when building for 2.1 shared and sources from 2.1 folder are used, when not building for 2.1 it should use new code

something along the lines of:

```xml
<ItemGroup>
<!-- shared sources here -->
</ItemGroup>
<ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
<!-- sources for 2.1 only here -->
</ItemGroup>
<ItemGroup Condition="'$(TargetFramework)' != 'netcoreapp2.1'">
<!-- sources for 3.1/5.0 only here -->
</ItemGroup>
```

@joperezr please verify this is correct (the part about splitting won't change so you can start with that, the details about TargetFrameworks we should target and specific syntax might change as per @joperezr reply)

joperezr:
Your summary looks right to me with just one difference, for now we plan to only continue targeting netcoreapp2.1 and net5.0 with iot.device.bindings, no plans on adding netcoreapp3.1 yet since that build and the net5.0 one would be the same for now. So I would say that this specific project binding should probably target either `netcoreapp2.1;net5.0` or `netcoreapp2.1;netcoreapp3.1`, no need to target all three

thecatkitty:
FYI: I haven't abandoned this change, I'm just currently focused on finishing my diploma :)

krwq:
Good luck, take your time 😄 Let us know the ETA so we can ping you at appropriate time

krwq:
@thecatkitty sorry for the delay, this is now unblocked, you should be able to multi-target

PR #1278, opened 2020-11-12
Add dependabot manifest

richlander:
Enable dependency updates for projects with PackageReference dependencies (with actual version numbers).

Context: 
- https://github.com/dotnet/core/pull/5578
- https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-dependencies-updated-automatically

This is what it will look like. It is noisy at first and then calms down. It doesn't have an option to batch PRs, which is unfortunate.

- https://github.com/richlander/iot/network/updates
- https://github.com/richlander/iot/pulls

image


richlander:
I've got an idea to make this better. Will create a new PR shortly.

Issue #1270, opened 2020-11-09
GpioController is not correctly identifying Raspberry Pi with Revision c03112


Semour81:
I would like to use Dht11 module on Raspberry PI 4 Revision 1.2 (Model C03112). 

**Steps to reproduce**

```csharp
using (Dht11 dht = new Dht11(17))
{
    var temperature = dht.Temperature;
}
```

**Expected behavior**

Accessing dht.Temperature must work without getting an exception. Used PI model supports InputPulUp pin mode.

**Actual behavior**

Exception occurs when accessing dht.Temperature.

```
An unhandled exception of type 'System.InvalidOperationException' occurred in System.Device.Gpio.dll: 'Pin 17 does not support mode InputPullUp.'
```

**Further analysis**

In [Issue 1145](https://github.com/dotnet/iot/issues/1145) there is a workaround described:

```csharp
GpioDriver driver = new RaspberryPi3Driver();
GpioController controller = new GpioController(PinNumberingScheme.Logical, driver);
```

Unfortunately this also does not help because PI model could not be detected and RaspberryPi3Driver() cannot be initialized.

```
An unhandled exception of type 'System.PlatformNotSupportedException' occurred in System.Device.Gpio.dll: 'Not a supported Raspberry Pi type: Unknown'
at System.Device.Gpio.Drivers.RaspberryPi3Driver..ctor() at MyConsoleApp.Program.ReadFromDHT11()
```
The actual reason of that behavior is, that enum Model in System.Device.Gpio.RaspberryBoardInfo does not contain C03112 value.

Model C03112 is officially listed on [Raspberry.org](https://www.raspberrypi.org/documentation/hardware/raspberrypi/revision-codes/README.md), model C03114 seems also not supported.

**Versions used**

_Raspberry:_

/proc/cpuinfo:
Hardware: BCM2711
Revision: c03112
Serial : 10000000683cf6ea
Model : Raspberry Pi 4 Model B Rev 1.2

_OS:_
Linux 5.4.0-1022-raspi #25-Ubuntu SMP PREEMPT Thu Oct 15 14:22:53 UTC 2020

_Dotnet:_

.NET Core SDK (reflecting any global.json):
 Version:   3.1.403
 Commit:    9e895200cd

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  20.04
 OS Platform: Linux
 RID:         ubuntu.20.04-arm
 Base Path:   /usr/bin/dotnet/sdk/3.1.403/

[...]

PR #1293, opened 2020-11-14
Fixing DHT InputPullUp to Input only

Ellerbach:
Changing InputPullUp to Input only
Rational:
- The InputPullUp is not supported on all platforms
- The InputPullUp is not required if the hardware includes a pull up
  - Most hardware come with the pull up already build in
  - Our recommendation from the sample is to use a physical resistor pull up. See https://github.com/dotnet/iot/tree/master/src/devices/Dhtxx/samples
  - In general, it's better to have real physical hardware pull up as you're sure it's always there

Should fix as well #1270

### DIFF
--- a/src/devices/Dhtxx/DhtBase.cs
+++ b/src/devices/Dhtxx/DhtBase.cs
@@ -162,7 +162,7 @@ namespace Iot.Device.DHTxx
             // wait 20 - 40 microseconds
             DelayHelper.DelayMicroseconds(30, true);
 
-            _controller.SetPinMode(_pin, PinMode.InputPullUp);
+            _controller.SetPinMode(_pin, PinMode.Input);
 
             // DHT corresponding signal - LOW - about 80 microseconds
             count = _loopCount;
--- a/src/devices/Dhtxx/DhtBase.cs
+++ b/src/devices/Dhtxx/DhtBase.cs
@@ -145,6 +145,7 @@ namespace Iot.Device.DHTxx
 
             byte readVal = 0;
             uint count;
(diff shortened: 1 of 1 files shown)


pgrawehr:
For #1270 this is more a workaround, is it?

pgrawehr (approved):

Ellerbach:
> For #1270 this is more a workaround, is it?

Yes and no. It's really not needed to use Pull Up. In general, our bindings should stick to Output and Input only. Nothing else. This is the only way to make sure it works everywhere. Hardware pull up is the way to go for any scenario.

krwq (approved):

src/devices/Dhtxx/DhtBase.cs
```diff
@@ -162,7 +162,7 @@ internal virtual void ReadThroughOneWire()
             // wait 20 - 40 microseconds
             DelayHelper.DelayMicroseconds(30, true);
 
-            _controller.SetPinMode(_pin, PinMode.InputPullUp);
+            _controller.SetPinMode(_pin, PinMode.Input);
```
krwq, line 165: the other way to ensure this works everywhere is have a field somewhere in this class:
PinMode _pinMode; with value `_pinMode = _controller.IsPinModeSupported(PinMode.InputPullUp) ? PinMode.InputPullUp : PinMode.Input;`

Ellerbach: Yes, that's actually a good idea. Made the changes.

krwq (approved):

PR #1296, opened 2020-11-17
Update SenseHat Joystick sample

richlander:
The goal of this update is to simplify the Joystick sample.

### DIFF
--- a/src/devices/SenseHat/SenseHatJoystick.cs
+++ b/src/devices/SenseHat/SenseHatJoystick.cs
@@ -59,7 +59,7 @@ namespace Iot.Device.SenseHat
         /// </summary>
         public void Read()
         {
-            JoystickState state = ReadState();
+            JoystickState state = GetState();
             HoldingLeft = state.HasFlag(JoystickState.Left);
             HoldingRight = state.HasFlag(JoystickState.Right);
             HoldingUp = state.HasFlag(JoystickState.Up);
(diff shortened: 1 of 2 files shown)


src/devices/SenseHat/SenseHatJoystick.cs
```diff
@@ -80,20 +80,45 @@ public void Dispose()
             _i2c = null!;
         }
 
-        private JoystickState ReadState()
+        /// <summary>
+        /// Read joystick state
+        /// </summary>
+        public JoystickState GetState()
```
joperezr, line 86: I'm fine if we want to keep it like this but in general I'd rather not have two ways of doing the same thing public ina binding, unless there is a good reason to do it.

src/devices/SenseHat/samples/Joystick.Sample.cs
```diff
@@ -17,33 +17,17 @@ public static void Run()
...
-                    Console.Write("D");
-                }
-
-                if (j.HoldingLeft)
-                {
-                    Console.Write("L");
-                }
-
-                if (j.HoldingRight)
+                char state = j.GetState() switch
```
joperezr, line 20: State is a flags attribute, doesn't that mean you could have more than one holding* at the same time? I don't have he device but what happens if you hold the joystick south-east for example? Shouldn't it be true for holding down and holding right in that case?

richlander: I thought of that. Wouldn't this could produce the same result as before?

joperezr: I don't think so. The way you have it it will only apply on one result per read, while before it might get you multiple per read. Also, if the read value has more than one state (two flags turned on) then you will now go to the default case.

richlander: I was assuming that because it was a while loop that it would be fast. I see your point now. You'd have to pass through the switch expression n times to get the same result.

I new before I made this update that there might be a difference, but couldn't quite express it. I'll close the PR.

joperezr: Yeah it would even not be the sane result even when entering twice to the while as it would likely just print @ several times. https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUBgBjwwGYACAbzzJrI3UutuYFkBPAUSggFsyAjMgF4yACnZdeAShIBuJsxoBnAO4BLYAGMAFmP5SFiqrkWmymgIZKApmQnceAOgAiAexVQQhs4tSYAnKIARHDuUEFS8iY+Zvxg1hYA1lExtJY2dpwOjgCqAA5e0anMfoFBEHkRKcXMcQnJ3mZw1gBmFhAANqCNqaXB/K7A2lU9MXVJ1aYAvoYzuHN4ANoAYh0WAOZKALrE5NYOmZJ8IAJswNZ4xjT5wmSYMExuHrfoeFNAA=

PR #1301, opened 2020-11-20
Adding support to multi-target device binding projects

joperezr:
cc: @richlander @krwq @Ellerbach @pgrawehr 

Even though device bindings have individual project files, we have always shipped them as one single library (Iot.Device.Bindings). The way we are doing that currently is by manually specifying in the Iot.Device.Bindings a static list of files to be compiled and excluding the ones that we don't want to, which won't be easy to scale in the future once we want to start adding bindings that are only supported in higher frameworks (for example, only in net5.0). The changes on this PR will make it so that instead of having one static list defined, the Iot.Device.Bindings project will now dynamically call into each individual device binding project to get out the right compile items for a specific framework, and that way it will now be easy moving forward to add new bindings projects that are supported only in some TFMs. At the moment, we target two TFMs with Iot.Device.Bindings (netcoreapp2.1, net5.0) but that is now able to scale in the future if we want to.

### DIFF
--- a/src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
+++ b/src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
@@ -14,12 +14,13 @@
     <!--Disabling default items so samples source won't get build by the main library-->
   </PropertyGroup>
   
+  <!-- The following ItemGroup is in charge of getting the source files we will compile on each TFM -->
   <ItemGroup>
-    <_ExcludeFromCompile Include="$(DeviceRoot)**/samples/**/*.cs" />
-    <_ExcludeFromCompile Include="$(DeviceRoot)**/tests/**/*.cs" />
-    <_ExcludeFromCompile Include="$(DeviceRoot)**/obj/**/*.cs" />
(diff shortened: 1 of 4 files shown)


krwq:
@thecatkitty once this is merged please rebase your PR

pgrawehr:
Question: Will this also work for other than source files (i.e. resources)? Just curious, because I have a PR pending that requires a file to be included as resource.

joperezr:
Good question, it won’t the way it is but we can add it very easily. We would’ve had to do it as well without this change as currently we are only globally adding compile files. I suppose your resources needed to be added is not for resx right? (String resources)

pgrawehr:
No, it will be an ordinary text file that serves as font map for a display. But yes, already now this has required an extra change in the iot.devices.csproj file. I'll get back to you once that PR is ready (I want to complete some of the other stuff first).

Ellerbach:
My understanding looking at the structure is that it will take anything that you reference in your csproj as this is what is added to the main build definition. So it should include any file that you would have added into the csproj.

joperezr:
That is true for source files, but not for resources, so as @pgrawehr points out it will require adding one extra line on device bindings on his pr, but that should be straight forward.

richlander (approved):

krwq:
@joperezr what's left here?

PR #1302, opened 2020-11-20
Fixing build due to new stylecop rule

joperezr:
FYI: @fehdem @krwq


joperezr:
Will merge as soon as CI passes in order to unblock the build

Issue #802, opened 2019-10-15
Documentation should include instructions on how to use/enable I2C/SPI/PWM/1-wire/...


krwq:
Currently we are lacking gathered place where we describe how to plug in and setup Raspberry PI with I2C/SPI/PWM and other protocols. We should write instructions and start linking samples to those docs to make people easier to start.

Work left here:
- [x] I2C instructions for Raspbian
- [x] SPI instructions for Raspbian
- [x] PWM instructions for Raspbian
- [ ] 1-wire instructions for Raspbian
- [x] Links in the samples

let me know if the list is missing anything

@MarkCiliaVincenti @shaggygi

Frankenslag:
@krwq 
Perhaps be more specific and say that it's a Pi running Raspbian. I have noticed that 64bit Debian, for example, doesn't have the same device tree configuration and so the SPI configuration instructions used for Raspian aren't appropriate.

krwq:
@Frankenslag good point, ideally we should have couple of other common OSes but I guess that's optional here

krwq:
Should we just point people to `/boot/overlays/README` for now? It describes how to enable everything on the PI using `/boot/config.txt`

larivierec:
I'm running this on Ubuntu 19.10 and it's not enough. It's a start because it allowed me to enable pwm through the overlays however, hardware PWM is currently unsupported within dotnet as certain pins need to be in different modes in order to operate on them.

krwq:
@larivierec I'm not sure you can change the mode without root access or if it would be appropriate for .NET to change that (although if we can do it so that you can switch between PWM/GPIO without issues then I think we should allow it). For PWM without changing settings you can use SoftwarePwm

larivierec:
> @larivierec I'm not sure you can change the mode without root access or if it would be appropriate for .NET to change that (although if we can do it so that you can switch between PWM/GPIO without issues then I think we should allow it). For PWM without changing settings you can use SoftwarePwm

I guess it's totally an opinion but I believe it's important for .NET to be able to change that also, if permissions are properly set, a user could request changing the pins state. Thanks for your input!

krwq:
@larivierec can you please file a separate issue on that? We can figure out if we should close this one there once we have enough data. If you have any info how to actually do that with i.e. SysFs please put additional info there. I don't expect that will work for all possible cases - we can likely get the GpioController which has direct memory access to do that, not sure about the rest of them.

[...]

PR #1305, opened 2020-11-22
Adding I2C documentation for Raspberry Pi

Ellerbach:
- Adding I2C documentation for Raspberry Pi
-One more from the #802 list

### DIFF
--- a/Documentation/raspi-i2c.md
+++ b/Documentation/raspi-i2c.md
@@ -40,6 +40,8 @@ Unhandled exception. System.IO.IOException: Error 121 performing I2C data transf
 Aborted
 ```
 
+**Note**: In rare cases, you might see the above exception during normal operation of a device as well. Adding retries might solve the issue.
+
 See at the end if you want to write your own I2C scanner and find the correct device address.
 
 ## Enabling I2C
@@ -66,7 +68,7 @@ Then reboot:
 sudo reboot
 ```
 
-More information on dtoverlay and how to select specific elements of I2C buses are [available here](https://github.com/raspberrypi/firmware/blob/bff705fffe59ad3eea33999beb29c3f26408de40/boot/overlays/README#L1387). 2 busses are available on any Raspberry Pi. More can be available on specific models like those base out of BCM2711.
+More information on dtoverlay and how to select specific elements of I2C buses are [available here](https://github.com/raspberrypi/firmware/blob/bff705fffe59ad3eea33999beb29c3f26408de40/boot/overlays/README#L1387). 2 busses are available on any Raspberry Pi. More can be available on specific models like those base out of BCM2711 such as Raspberry Pi 4.
 
 ### Enabling I2C with advance parameters
 
@@ -127,6 +129,12 @@ SUBSYSTEM=="i2c-dev", GROUP="i2c", MODE="0660"
 If you don't have it or if you want to adjust the permissions, this is what you'll need to add/adjust, as always save through `ctrl + x` then `Y` then `enter` and then reboot.
 
 ## Bonus: write your own I2C scanner
+
+Raspberry Pi OS comes with a tool to scan the I2C bus. This tool is called `i2cdetect`. It is not the case in all systems and all OS. Usage for bus 1 (default on Raspberry Pi) is:
+
+```bash
+i2cdetect -y 1
+```
 
 I2C devices are available from the bus address 8 (0x08) to 127 (0x7F). If a device is present, it will be ok to be read. So you just need to loop and check if you can read a device. Note that this code will only work if you have previously activated I2C. The following code require C#9.0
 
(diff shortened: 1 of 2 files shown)


Documentation/raspi-i2c.md
```diff
@@ -0,0 +1,162 @@
...
+
+Save the file with `ctrl + x` then `Y` then `enter`
+
+Then reboot:
+
+```bash
+sudo reboot
+```
+
+More information on dtoverlay and how to select specific elements of I2C buses are [available here](https://github.com/raspberrypi/firmware/blob/bff705fffe59ad3eea33999beb29c3f26408de40/boot/overlays/README#L1387). 2 busses are available on any Raspberry Pi. More can be available on specific models like those base out of BCM2711.
```
pgrawehr, line 69: I suggest adding "such as Raspberry Pi 4" somewhere. That's what people know better than the CPU it runs with.

Documentation/raspi-i2c.md
```diff
@@ -0,0 +1,162 @@
...
+
+You should find a line like this one:
+
+```text
+SUBSYSTEM=="i2c-dev", GROUP="i2c", MODE="0660"
+```
+
+If you don't have it or if you want to adjust the permissions, this is what you'll need to add/adjust, as always save through `ctrl + x` then `Y` then `enter` and then reboot.
+
+## Bonus: write your own I2C scanner
```
pgrawehr, line 129: Maybe add a hint that there's already an existing tool to do this?

Documentation/raspi-i2c.md
```diff
@@ -0,0 +1,162 @@
...
+
+```text
+Unhandled exception. System.IO.IOException: Error 121 performing I2C data transfer.
+   at System.Device.I2c.UnixI2cDevice.ReadWriteInterfaceTransfer(Byte* writeBuffer, Byte* readBuffer, Int32 writeBufferLength, Int32 readBufferLength)
+   at System.Device.I2c.UnixI2cDevice.Transfer(Byte* writeBuffer, Byte* readBuffer, Int32 writeBufferLength, Int32 readBufferLength)
+   at System.Device.I2c.UnixI2cDevice.WriteByte(Byte value)
+   at TestTest.Program.Main(String[] args) in C:\tmp\TestI2C\Program.cs:line 14
+Aborted
+```
+
```
pgrawehr, line 42: Add: "In rare cases, you might see the above exception during normal operation of a device as well. Adding retries might solve the issue."

pgrawehr (approved):

Documentation/raspi-i2c.md
Ellerbach: Yes, good point. It does not exist on all systems. But right, it does on some! Will add!

Documentation/raspi-i2c.md
Ellerbach: Yes, good point. It does not exist on all systems. But right, it does on some! Will add!

Documentation/raspi-i2c.md
Ellerbach: ok!

Documentation/raspi-i2c.md
```diff
@@ -0,0 +1,170 @@
...
+```bash
+sudo reboot
+```
+
+More information on dtoverlay and how to select specific elements of I2C buses are [available here](https://github.com/raspberrypi/firmware/blob/bff705fffe59ad3eea33999beb29c3f26408de40/boot/overlays/README#L1387). 2 busses are available on any Raspberry Pi. More can be available on specific models like those base out of BCM2711 such as Raspberry Pi 4.
+
+### Enabling I2C with advance parameters
+
+The general pattern looks like the following:
```
krwq, line 75: you should also mention you can do `sudo raspi-config nonint do_i2c 0` on raspbian

krwq (approved):

Documentation/raspi-i2c.md
Ellerbach: Added. 1 is the default, so changed for 1 with a note

krwq (approved):

Documentation/raspi-i2c.md
```diff
@@ -0,0 +1,162 @@
...
+
+Then reboot:
+
+```bash
+sudo reboot
+```
+
+More information on dtoverlay and how to select specific elements of I2C buses are [available here](https://github.com/raspberrypi/firmware/blob/bff705fffe59ad3eea33999beb29c3f26408de40/boot/overlays/README#L1387). 2 busses are available on any Raspberry Pi. More can be available on specific models like those base out of BCM2711.
+
+### Enabling I2C with advance parameters
```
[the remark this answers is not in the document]
Ellerbach: I have to say I prefer all in one. Makes it easier. The first part may solve 99% of the problems, knowing you have all in 1 page makes you feel that you'll find a solution whatever happens.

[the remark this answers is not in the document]
krwq: Yes I agree with @Ellerbach, we can maybe restructure doc a bit to cover 99% cases in the first paragraph but I think it's not really that large. What I think we should do next is to add a category in each device using I2cDevice/SpiDevice (perhaps even automate that) and add links to these instructions in each sample so people can easily find it

Issue #1326, opened 2020-11-27
Ft4222 is throwing exceptions on Dispose


krwq:
```csharp
            var i2c = new Ft4222I2c(new I2cConnectionSettings(0, Bme280.DefaultI2cAddress));

            using (Bme280 bme = new Bme280(i2c))
            {
            }
```

```
Unhandled exception. System.ObjectDisposedException: Safe handle has been closed.
Object name: 'SafeHandle'.
   at System.Runtime.InteropServices.SafeHandle.DangerousAddRef(Boolean& success)
   at System.StubHelpers.StubHelpers.SafeHandleAddRef(SafeHandle pHandle, Boolean& success)
   at Iot.Device.Ft4222.FtFunction.FT4222_UnInitialize(SafeFtHandle ftHandle)
   at Iot.Device.Ft4222.SafeFtHandle.ReleaseHandle()
   at System.Runtime.InteropServices.SafeHandle.InternalRelease(Boolean disposeOrFinalizeOperation)
   at System.Runtime.InteropServices.SafeHandle.Dispose(Boolean disposing)
   at System.Runtime.InteropServices.SafeHandle.Dispose()
   at Iot.Device.Ft4222.Ft4222I2c.Dispose(Boolean disposing)
   at System.Device.I2c.I2cDevice.Dispose()
   at Iot.Device.Bmxx80.Bmxx80Base.Dispose(Boolean disposing)
   at Iot.Device.Bmxx80.Bmxx80Base.Dispose()
   at Ft4222Test.Program.Main(String[] args) in Program.cs:line 52
```

PR #1327, opened 2020-11-27
Fix exceptions on Ft4222's Dispose methods

krwq:
Fixes: #1326


Ellerbach (approved):
Looks all good to me. Thanks for the fix!

PR #1332, opened 2020-11-30
Add a generic logging infrastructure

pgrawehr:
Allows to have a common base for logging
within all bindings.

Simple console logger provided as well, but
in a larger application, you might want to
forward the logs to some logging framework
such as NLog.

Note: Some bindings that use their own logging
right now should be updated afterwards.

### DIFF
--- a/src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
+++ b/src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
@@ -29,6 +29,7 @@
     <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
     <PackageReference Include="UnitsNet" Version="$(UnitsNetPackageVersion)" />
     <PackageReference Include="System.Management" Version="$(SystemManagementPackageVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
     <PackageReference Include="SixLabors.ImageSharp" Version="$(SixLaborsImageSharpPackageVersion)" />
   </ItemGroup>
 
--- a/src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
+++ b/src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
@@ -29,7 +29,7 @@
     <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
     <PackageReference Include="UnitsNet" Version="$(UnitsNetPackageVersion)" />
     <PackageReference Include="System.Management" Version="$(SystemManagementPackageVersion)" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractionsPackageVersion)" />
     <PackageReference Include="SixLabors.ImageSharp" Version="$(SixLaborsImageSharpPackageVersion)" />
   </ItemGroup>
 
--- a/src/devices/Common/Iot/Device/Common/SimpleConsoleLogger.cs
+++ b/src/devices/Common/Iot/Device/Common/SimpleConsoleLogger.cs
@@ -1,0 +1,72 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Iot.Device.Common
+{
+    /// <summary>
+    /// A simple console logger - logs all incoming log messages to the console
+    /// </summary>
+    public class SimpleConsoleLogger : ILogger
+    {
+        /// <summary>
+        /// Creates console output with color support
+        /// </summary>
+        public SimpleConsoleLogger(string loggerName)
+        {
+            LoggerName = loggerName;
+            MinLogLevel = LogLevel.Information;
+        }
+
+        /// <summary>
+        /// Specifies the minimum log level that is printed. Default is <see cref="LogLevel.Information"/>
+        /// </summary>
+        public LogLevel MinLogLevel
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// Name of the logger
+        /// </summary>
+        public string LoggerName { get; }
+
+        /// <inheritdoc />
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            var previousColor = Console.ForegroundColor;
+            switch (logLevel)
+            {
+                case LogLevel.Error:
+                case LogLevel.Critical:
+                    Console.ForegroundColor = ConsoleColor.Red;
+                    break;
+                case LogLevel.Warning:
+                    Console.ForegroundColor = ConsoleColor.Yellow;
+                    break;
+                case LogLevel.Information:
+                    Console.ForegroundColor = ConsoleColor.Blue;
+                    break;
+            }
+
+            string msg = formatter(state, exception);
+            Console.WriteLine($"{logLevel} - {msg}");
+            Console.ForegroundColor = previousColor;
+        }
+
+        /// <inheritdoc />
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return logLevel >= MinLogLevel;
+        }
+
+        /// <inheritdoc />
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            return new LogDispatcher.ScopeDisposable();
+        }
+    }
+}
--- a/src/devices/Common/Iot/Device/Common/SimpleConsoleLoggerFactory.cs
+++ b/src/devices/Common/Iot/Device/Common/SimpleConsoleLoggerFactory.cs
@@ -1,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Iot.Device.Common
+{
+    /// <summary>
+    /// Provides a very simple console logger that does not require a reference to Microsoft.Extensions.Logging.dll
+    /// </summary>
+    public class SimpleConsoleLoggerFactory : ILoggerFactory
+    {
+        /// <summary>
+        /// The console logger is built-in here
+        /// </summary>
+        /// <param name="provider">Argument is ignored</param>
+        public void AddProvider(ILoggerProvider provider)
+        {
+        }
+
+        /// <inheritdoc/>
+        public ILogger CreateLogger(string categoryName)
+        {
+            return new SimpleConsoleLogger(categoryName);
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
+        }
+    }
+}
--- a/src/devices/Common/README.md
+++ b/src/devices/Common/README.md
@@ -22,7 +22,7 @@ The `LogDispatcher` class contains static methods to enable a binding to provide
 
 Logging is disabled by default. To enable logging to the console in an application, you can use this code: 
 
-```
+```csharp
 using var loggerFactory = new LoggerFactory();
 loggerFactory.AddProvider(new ConsoleLoggerProvider((s, level) => true, false));
 
@@ -33,7 +33,8 @@ LogDispatcher.LoggerFactory = loggerFactory;
 Other loggers are available in the `Microsoft.Extensions.Logging` Nuget-package. Some help is (here)[https://docs.microsoft.com/en-us/dotnet/core/extensions/logging#non-host-console-app]. Alternatively, you can also connect the log output to your favorite logging framework, such as NLog, Log4Net and others.
 
 To enable a binding to use logging, you can call `GetCurrentClassLogger()` in the constructor to get a logger that automatically identifies the class.
-```
+
+```csharp
     private ILogger _logger;
 
     public MyTestComponent()
@@ -43,7 +44,8 @@ To enable a binding to use logging, you can call `GetCurrentClassLogger()` in th
 ```
 
 Then, logging can be performed using the methods on the obtained `ILogger` interface.
-```
+
+```csharp
 _logger.LogInformation("An informative message");
 _logger.LogError("An error situation");
 _logger.LogWarning(new PlatformNotSupportedException("Something is not supported"), "With exception context");
--- a/src/devices/Common/README.md
+++ b/src/devices/Common/README.md
@@ -23,8 +23,11 @@ The `LogDispatcher` class contains static methods to enable a binding to provide
 Logging is disabled by default. To enable logging to the console in an application, you can use this code: 
 
 ```csharp
-using var loggerFactory = new LoggerFactory();
-loggerFactory.AddProvider(new ConsoleLoggerProvider((s, level) => true, false));
+using var loggerFactory = LoggerFactory.Create(builder =>
+{
+	builder
+		.AddConsole();
+});
 
 // Statically register our factory. Note that this must be done before instantiation of any class that wants to use logging.
 LogDispatcher.LoggerFactory = loggerFactory;
--- a/src/devices/Common/README.md
+++ b/src/devices/Common/README.md
@@ -1,0 +1,51 @@
+# Common classes
+
+This contains a set of classes that can be used accross different sensors and bindings.
+
+## Weather helpers
+
+The `WeatherHelper` class contains a set of static functions to calculate meteorological data based on different input sets.
+
+Some of the functions include:
+
+- CalculateDewPoint: Calculates the [dew point](https://en.wikipedia.org/wiki/Dew_point), given a measured temperature and relative humidity. Simply put, the dew point indicates at what temperature water will condense if temperature drops.
+- CalculateAltitude: (Different overloads) Calculate observer altitude given current pressure and temperature
+- CalculateBarometricPressure: Calculate the sea-level equivalent of the current pressure, given the observers altitude. Used to make pressure readings from different locations comparable.
+
+## Number Helper
+
+The `NumberHelper` class contains a few methods to convert BCD numbers (binary coded decimal, uses 4 bits for each decimal digit) to binary and vice-versa.
+
+## Logging 
+
+The `LogDispatcher` class contains static methods to enable a binding to provide log output. This can be helpful for debugging or analysis.
+
+Logging is disabled by default. To enable logging to the console in an application, you can use this code: 
+
+```
+using var loggerFactory = new LoggerFactory();
+loggerFactory.AddProvider(new ConsoleLoggerProvider((s, level) => true, false));
+
+// Statically register our factory. Note that this must be done before instantiation of any class that wants to use logging.
+LogDispatcher.LoggerFactory = loggerFactory;
+```
+
+Other loggers are available in the `Microsoft.Extensions.Logging` Nuget-package. Some help is (here)[https://docs.microsoft.com/en-us/dotnet/core/extensions/logging#non-host-console-app]. Alternatively, you can also connect the log output to your favorite logging framework, such as NLog, Log4Net and others.
+
+To enable a binding to use logging, you can call `GetCurrentClassLogger()` in the constructor to get a logger that automatically identifies the class.
+```
+    private ILogger _logger;
+
+    public MyTestComponent()
+    {
+        _logger = this.GetCurrentClassLogger();
+    }
+```
+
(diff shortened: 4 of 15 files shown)


krwq:
cc: @colombod

krwq:
@pgrawehr I've talked with @colombod offline and he has suggested we might want to use https://github.com/jonsequitur/PocketLogger/blob/master/docs/readme.md instead

colombod:
Pocket logger is a library we have been using for a while and I like that it comes as code as opposed to binary dependency, so all the work is based on internally compiled types that can be extended very easily without bringing additional load to the dependency graph. The part I really like of pocket logger is that the logging is completely customizable and custom subscriber can be added for specific set of assemblies. There are already few custom subscribers able to pump over to serilog and App Insight, but here I could see also feeding in monitoring channels in azure iot or other platforms that supports semantic logging or even tracing events in distributed deployment. The log entries in pocket logger all contains correlation ids that allows for tracing and reconstruction of events and their timing out of the box.

Ellerbach:
I've been using this one for the NFC devices and related elements: https://github.com/dotnet/iot/blob/master/src/devices/Card/LogInfo.cs
We have to keep in mind we **must** keep things super small and super fast on embedded devices. The less overhead the best. So as much as I like existing logger, as much as I think if we select one, it should be as simple and straight forward as the one I have for the NFC elements.

pgrawehr:
I know that there are great logging libraries out there, and this proposal clearly should not replace any of those. The interface is intended to be linked to these, but without adding a dependency. This allows every app to use its own logging framework. @colombod Can you explain how that Pocket logger works if it's a source dependency only? The documentation doesn't seem to mention that. 

@Ellerbach The logger also needs to be generic and it must be able to forward events to an external listener. I think the approach I took is about the smallest possible, or where would you cut even more?

colombod:
cc @jonsequitur as he is the main maintainer of pocket logger.
When you add pocket logger to a project it brings in csharp files as oppose to a dll to link. this means that all logging code is compile and relates to your types. The logging is done as semantic logging as opposed to simple strings. The api takes advantage of IDisposable to start and close. The platform is the fully configurable and flexible. You can even express the intent to deliver logging entries from specific assemblies over dedicated sinks.

krwq:
@Ellerbach good point about embedded devices - we need to think about generating code for nano framework so it can't use too fancy types as they won't work in there (i.e. preferably no generic types). I haven't looked into the pocket logger to see what exactly it uses but code seemed relatively small

jonsequitur:
> we need to think about generating code for nano framework so it can't use too fancy types as they won't work in there (i.e. preferably no generic types). I haven't looked into the pocket logger to see what exactly it uses but code seemed relatively small

PocketLogger is source code only and defines its own types, which are internal so that the dependency graph of your library isn't affected. We're also looking at replacing the use of runtime reflection with source generation, which should give a significant performance boost. We can definitely look into making these improvements with an eye to these scenarios.

pgrawehr:
I just had a quick look into PocketLogger, but I think it's not suitable for the purpose we need it here. The fact that it's all internal hides the reference from any consumers of the library, but that's exactly not what I think we should be doing. The behavior of the logging shall be defined by the consuming application (i.e. whether it wants to log to the console, or to a file, to a network service or whatever). PocketLogger would have every component (or binding) to define its logging output. 

Additionally, the source dependency creates an issue: Because the logger is included as source, it requires to be conformant to this projects StyleCop rules, which it is not. This could certainly be leveraged somehow, but is an additional topic to consider.

krwq:
@pgrawehr style rules are needed to help read the code, normally you don't read generated code which means style rules doesn't apply there. Also internal APIs can be published through public APIs. I think this should be a matter of:
- how easily can the code be converted to work on nano framework? (does it even need any special post processing?)
- how small it is?
- if we ever decide to split bindings into separate packages will we have to carry copies of logger in each of them?

Pocket logger will tell you the execution time of methods which you decide to log and can show you the structure of execution which is nice for debugging but if it will prevent us working on nano framework or adds significant work I'd rather go with simple logging. I think this is something which is worth investigating before we set up the guidelines.

pgrawehr:
@krwq: The code is not _generated_ in a sense, but directly included in the project. I agree that you don't normally need to read external code and so this is not an usability problem, but rather a technical one (without extra effort to exclude StyleCop for these files, a project which references PocketLogger just won't build).

When we externalise the interface to the logger, we'll need an interface such as the one I provided here, so this is required anyway (in some form or other).

For me, the main requirements were
- Clients of Iot.Devices should decide where the logging goes to (File, Console, Network...) and not the bindings themselves
- Simple to add to the individual bindings
- There should be just one log stream combined from all the sources

> if we ever decide to split bindings into separate packages will we have to carry copies of logger in each of them?
I think this answers to true for PocketLogger, and already with our distributed project infrastructure (with the bindings having a separate assembly for development purposes), this would cause conflicts.

Ellerbach:
The Log I put in place for the NFC reader is extremely basic and can be activated when developing with cards as you sometimes needs to understand the various messages going from and to the cards. That's the reason it's there.
For other bindings, what are those which may require a similar level of log when they are in use and when the developer using them needs more in depth data?
Otherwise, I feel like the log should be part of the application itself. So do we really need a logger for the purpose of having one if there is almost no bindings which needs to use it?
Don't get me wrong, I have nothing against :-) Just want to make sure we take the right decision. And as @krwq mention, we need to find one which can be fully deactivated or supports a small target like nanoFramrwork.

pgrawehr:
@Ellerbach The proposal I make here should support all your requirements. By default, logging is disabled - you have to register a logger to see a result. As already stated above, I'm fully with you that logging should be done by the application. The classes here just provide one common callback for all possible logging sources. The very basic `ConsoleLogger` is just provided for convenience. It is not enabled by default.

At the moment, there are in fact not so many bindings using logging, but there could (and possibly should) be more. I've seen more than one case where new bindings included Console.WriteLine commands where logging would be just what you need.

matthewethomas:
Maybe I missed something, but how does this differ from ILogger in [Microsoft.Extensions.Logging.Abstractions](https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions/), and then wiring up the logging framework of your choice to forward entries to the console ([example](https://docs.microsoft.com/en-us/dotnet/core/extensions/logging#non-host-console-app)) (or whatever destination you need)? 

I don't want to discount what looks to be good work, but "add a generic logging infrastructure" seems a little out of scope for the IOT space, and even then it seems that this problem is already solved in another Microsoft library. 

Assuming that this PR is merged, a few things things:
- Developer education might prove to be difficult - M.E.L.A ILogger has YUUUGE mindshare due to it being the [recommended means of logging](https://docs.microsoft.com/en-us/dotnet/core/extensions/logging) in .Net. The broader community's going to need a clear, simple reason of "why this, and not that" in order to see adoption
     - Building on that, anything that's not M.E.L.A. ILogger in new Microsoft stuff seems weird. "I" need a reason to use weird stuff.
- I as a consumer am going to want this interface to integrate with and forward to the logging framework of my choice (serilog, nlog, log4net, pocketlogger, etc) - As is, it looks like that's still on "me" to do. That is code "I" really don't want to have to write, much less maintain.
- For consistency's sake, [Iot.Device.Card.LogInfo](https://github.com/dotnet/iot/blob/master/src/devices/Card/LogInfo.cs) should be removed, and everything affected should be updated to use this method of logging,

Again - I concede that I may be missing a use case, or just not seeing the forest for the trees.

> I've seen more than one case where new bindings included Console.WriteLine commands where logging would be just what you need

May I suggest instead of adding a logging framework or abstraction into the IOT/System.Devices namespace, the powers that be take a stance on logging ("Console.WriteLine is ok/not ok", "Logging is done via X", etc.) and then follow up with some guidance document and/or PRs to have the existing console writelines and logging conform to the new recommendations?

pgrawehr:
@matthewethomas There are quite a few points you bring up here. First of all: I just didn't know this Microsoft.Extensions.Logging package. Despite that it is part of the core runtime, it is very poorly documented - there's not a single example of how you're supposed to use that abstraction layer or how you would link it to other frameworks. But yes, in general, that seems to be more or less what we need here. 

> 
>     * I as a consumer am going to want this interface to integrate with and forward to the logging framework of my choice (serilog, nlog, log4net, pocketlogger, etc) - As is, it looks like that's still on "me" to do. That is code "I" really don't want to have to write, much less maintain.

You'd have to do that either way. This library cannot provide the integration, outerwise it would have to link to the logging framework, which it absolutely should not. 

> 
>     * For consistency's sake, [Iot.Device.Card.LogInfo](https://github.com/dotnet/iot/blob/master/src/devices/Card/LogInfo.cs?rgh-link-date=2020-12-05T04%3A27%3A51Z) should be removed, and everything affected should be updated to use this method of logging,

Sure. This part is purposedly missing from this PR and will be done once a concept is in place.

> 
>     * If NanoFramework is a compilation target, something that outputs to Debug.Write is going to be a needed change either as part of this PR, or in the future, as it looks like they've flagged [Console.WriteLine as obsolete](https://github.com/nanoframework/lib-CoreLibrary/blob/develop/nanoFramework.CoreLibrary/System/Console.cs?rgh-link-date=2020-12-05T04%3A27%3A51Z#L24)

That shall be the integrator's concern. 

> 
> 
> Again - I concede that I may be missing a use case, or just not seeing the forest for the trees.
> 
> > I've seen more than one case where new bindings included Console.WriteLine commands where logging would be just what you need
> 
> May I suggest instead of adding a logging framework or abstraction into the IOT/System.Devices namespace, the powers that be take a stance on logging ("Console.WriteLine is ok/not ok", "Logging is done via X", etc.) and then follow up with some guidance document and/or PRs to have the existing console writelines and logging conform to the new recommendations?
That's what we try to do here. There's consensus (I think) that Console.WriteLine shall not be used directly within the library, since most applications built against this library are currently console applications. One of the main hardware targets is the Raspberry Pi running linux. And arbitrary low-level log messages within a console-application are not desirable.

matthewethomas:
> Despite that it is part of the core runtime, it is very poorly documented - there's not a single example of how you're supposed to use that abstraction layer or how you would link it to other frameworks. But yes, in general, that seems to be more or less what we need here.

As for examples, it's buried, but there is documentation on [how to use it](https://docs.microsoft.com/en-us/dotnet/core/extensions/logging#create-logs)

Additionally, f you need to implement your own provider, [there's documentation on how to do that](https://docs.microsoft.com/en-us/dotnet/core/extensions/custom-logging-provider), but many other logging frameworks have already implemented a provider - [PocketLogger](https://www.nuget.org/packages/PocketLogger.For.MicrosoftExtensionsLogging/), [Serilog](https://github.com/serilog/serilog-extensions-logging), [NLog](https://github.com/NLog/NLog.Extensions.Logging), [Log4Net](https://github.com/huorswords/Microsoft.Extensions.Logging.Log4Net.AspNetCore), to name a few

If MELA Ilogger were used, log events could always be emitted with a level of trace or debug, and then filtered out as the project needs, as in this [simple example that emits to the console](https://docs.microsoft.com/en-us/dotnet/core/extensions/logging#non-host-console-app)

pgrawehr:
@matthewethomas I now updated this to use the interfaces from Microsoft.Extensions.Logging.Abstractions - that's really quite similar to what we need here. Remains to say that the documentation is in fact very poor, I'm not used to that from Microsoft. The example that you mentioned [here](https://docs.microsoft.com/en-us/dotnet/core/extensions/logging#non-host-console-app) doesn't work at all - there's no method `LoggerFactory.Create()`. In other places, methods are documented with plain wrong arguments.

src/devices/Common/Iot/Device/Common/SimpleConsoleLogger.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Iot.Device.Common
+{
+    /// <summary>
+    /// A simple console logger - logs all incoming log messages to the console
+    /// </summary>
+    public class SimpleConsoleLogger : ILogger
```
krwq, line 12: Should this be in a sample project rather than part of library?

src/devices/Common/Iot/Device/Common/SimpleConsoleLoggerFactory.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
```
krwq, line 1: 1. missing copyright header
2. should this be showed in the sample rather than part of the library? (I'd suggest to use root level sample folder to showcase logging)

src/devices/Common/Iot/Device/Common/SimpleConsoleLogger.cs
pgrawehr: Either is possible. I added it to the library so it's easier to reuse, but one could, of course, also reuse it by copy-pasting.

src/devices/Common/Iot/Device/Common/SimpleConsoleLoggerFactory.cs
pgrawehr: Yea, I need to add a readme

pgrawehr: Added documentation on usage. We can discuss, but we might want to keep this in the library, to avoid code duplication in all our examples when more and more bindings support logging.

src/devices/Common/Iot/Device/Common/SimpleConsoleLogger.cs
matthewethomas: @krwq, agreed - implementations of ILogger shouldn't ship as part of the framework - we can use them in example projects to show concepts, but it shouldn't be part of the system.devices.* namespace

pgrawehr: This isn't going to the System.Devices namespace, just the Iot.Devices namespace. I'd rather not replicate the implementation in all examples

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -28,6 +28,7 @@
     <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
     <PackageReference Include="UnitsNet" Version="$(UnitsNetPackageVersion)" />
     <PackageReference Include="System.Management" Version="$(SystemManagementPackageVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
```
Ellerbach, line 31: I check the source code and we will have hard time to make this working for nano as it's using various generics <T> but we may just rewrite using your initial code couple of high level wrappers which make the tricks for nano

src/devices/Common/Iot/Device/Common/SimpleConsoleLogger.cs
Ellerbach: Agree that I would like to see it in Iot.Device namespace. Yes it's an implementation, but bindings are as well implementations of hardware drivers :-) So why not having it there like the rest of the bindings, easy and fast to use. But clearly not in Sytem.Device

src/devices/Common/README.md
```diff
@@ -0,0 +1,51 @@
...
+    private ILogger _logger;
+
+    public MyTestComponent()
+    {
+        _logger = this.GetCurrentClassLogger();
+    }
+```
+
+Then, logging can be performed using the methods on the obtained `ILogger` interface.
+```
```
Ellerbach, line 46: add csharp after the 3 quotes so it can be properly displayed.
And you next an extra enter before

src/devices/Common/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+The `NumberHelper` class contains a few methods to convert BCD numbers (binary coded decimal, uses 4 bits for each decimal digit) to binary and vice-versa.
+
+## Logging 
+
+The `LogDispatcher` class contains static methods to enable a binding to provide log output. This can be helpful for debugging or analysis.
+
+Logging is disabled by default. To enable logging to the console in an application, you can use this code: 
+
+```
```
Ellerbach, line 25: add csharp after the 3 quotes so it can be properly displayed.

src/devices/Common/README.md
```diff
@@ -0,0 +1,51 @@
...
+loggerFactory.AddProvider(new ConsoleLoggerProvider((s, level) => true, false));
+
+// Statically register our factory. Note that this must be done before instantiation of any class that wants to use logging.
+LogDispatcher.LoggerFactory = loggerFactory;
+```
+
+Other loggers are available in the `Microsoft.Extensions.Logging` Nuget-package. Some help is (here)[https://docs.microsoft.com/en-us/dotnet/core/extensions/logging#non-host-console-app]. Alternatively, you can also connect the log output to your favorite logging framework, such as NLog, Log4Net and others.
+
+To enable a binding to use logging, you can call `GetCurrentClassLogger()` in the constructor to get a logger that automatically identifies the class.
+```
```
Ellerbach, line 36: add csharp after the 3 quotes so it can be properly displayed.
And you next an extra enter before

Ellerbach (approved):
We will have to think how to best implement this for nano without generics and that will require a bit of work but we can derive it from the simple logger I wrote originally for the NFC readers. We will have to align the functions and remove the generics but should be quite ok

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
pgrawehr: It looks weird, but I think it shouldn't be such a big deal, since their usage of generics serves - as far as I have seen - no real purpose. They mostly use generics to forward the class types of the logging sources. So the generic arguments should be easily replaced by an additional string or type argument.

pgrawehr: Note that the implementation I use tries to reduce the visible generics from the imported library. If you redefine the ILogger interface for nano without generics, there will be very few changes required.

Ellerbach: Yes, we will manage a way to adjust. Not too stressed about it!

krwq:
@pgrawehr I've tested PocketLogger and seems to me if we add that in the Iot.Device.Bindings then you will get logging without adding any abstraction.

In the exe app by default you will not see any logging but if you add PocketLogger to your project you can subscribe to which assemblies you want to see the logging and opt-in to see them from Iot.Device.Bindings. My main drawback from PocketLogger is nano support as likely this will currently be an issue but other than that I find it very convenient

pgrawehr:
Ok, so it would work for our scenario, but apparently trough heavy use of reflection, which is certainly problematic for Nano, but could also be a downside for other hardware, because of the performance impact. 

My larger drawback on PocketLogger though is that we should avoid depending on a specific logging framework and require the clients to use that. The Microsoft.Extensions.Logging package is designed just to be able to connect any logging framework you want to the interfaces. If one has NLog in his project, he probably doesn't want to have to write a PocketLogger->Nlog forwarder in his app.

jonsequitur:
I'm not pushing for any particular outcome here but I'd like to add a few clarifications about PocketLogger:

> Ok, so it would work for our scenario, but apparently trough heavy use of reflection, which is certainly problematic for Nano, but could also be a downside for other hardware, because of the performance impact.

The reflection cost is a one-time startup cost. That will be removed in an upcoming version, which would make this more suitable for Nano.

> My larger drawback on PocketLogger though is that we should avoid depending on a specific logging framework and require the clients to use that. The Microsoft.Extensions.Logging package is designed just to be able to connect any logging framework you want to the interfaces. If one has NLog in his project, he probably doesn't want to have to write a PocketLogger->Nlog forwarder in his app.

Avoiding a dependency on a specific logging framework is the main goal of PocketLogger. It was written to allow logging code within libraries that doesn't force a dependency on the user. Instead of an assembly dependency, the client embeds the PocketLogger.Subscribe source code via a source-only NuGet package. Then users can choose their own logging infrastructure. If they're not interested in logs, there's nothing to do at all, and no dependency.

The adapter for Microsoft.Extensions.Logging is [107 LoC](https://github.com/jonsequitur/PocketLogger/blob/master/Pocket.Logger/For.MicrosoftExtensionsLogging/LoggerFactoryExtensions.cs). For Serilog, [8 LoC](https://github.com/jonsequitur/interactive/blob/c44fed791093db19261c49b39d7d901e7983a3af/src/dotnet-interactive/Program.cs#L47-L54).

krwq:
cc: @joperezr

joperezr:
@tarekgh This PR is adding a generic logging infrastructure based on Microsoft.Extensions.Logging. I mentioned that it might be good to get onboard with the new work happening in System.Diagnositcs instead. What is your input regarding that?

tarekgh:
@joperezr the doc https://docs.microsoft.com/en-us/dotnet/core/diagnostics/logging-tracing is very useful to review and then decide what the IoT libraries need to do. Need to decide if the libraries need to support logging vs. tracing? Also, the doc talk about the perf consideration. 

In .NET 6.0 we are gong to support `Metrics` logging which I believe would be super useful for the IoT libraries. `Metrics` will allow publishing numeric data (e.g. counters or gauges). I had some offline chatting with @krwq about that and I think he is planning to do something later after we release the `Metrics`.

joperezr (approved):

Issue #1335, opened 2020-12-01
I2cBus


krwq:
Some sensors use multiple I2C devices. Example of that is SenseHat. Currently in order to swap I2C implementation with different one (i.e. Ft4222) it's a bit of a hassle: device need to take 6 I2C Devices as inputs and construct all of them. If we had I2cBus this could be solved much simpler:

```csharp
public abstract class I2cBus
{
  public abstract I2cDevice CreateDevice(int deviceAddress);
  public static I2cBus Create(int bus)
  {
    return new DefaultI2cBus(bus);
  }
}

internal class DefaultI2cBus : I2cBus
{
  private int _bus;

  public DefaultI2cBus(int bus)
  {
    _bus = bus;
  }

  public override I2cDevice CreateDevice(int deviceAddress)
  {
    return I2cDevice.Create(new I2cConnectionSettings(bus, deviceAddress));
  }
}
```

in that case sense hat could take bus as an input (and null as default which would use the correct bus). The Ft4222 could implement a bus and could be provided directly to SenseHat.

- open question: should bus manage lifetime of devices or just be a factory (factory would be fine for most of the cases)

Ellerbach:
This is similar but kind of reverse way of the GpioController and GpioPin discussion :-)
In the I2C case, we have the equivalent of the GpioPin and we are missing the equivalent of the GpioController.
In all cases, we need both, they have their usage. So I'm very much in favor of having an I2cBus as described.
You'll most likely need a bit more:

```csharp
public override void RemoveDevice(int deviceAddress)
```

shaggygi:
Somewhat related #214

krwq:
@shaggygi actually wouldn't be really bad idea to add Write/Read/WriteRead to I2cBus with the device address to allow for direct writing to the device. For things like I2C scanner this would remove overhead of having to have create instances and also make it easier to implement those few weird devices. It would also be on pair with GpioController and potential GpioPin we add later

Ellerbach:
+1

pgrawehr:
I've also thought of this before, but IIRC it was considered to much overhead. But I agree that an i2cbus instance would be helpful. It could provide functions such as bus scanning or also bus reset (send a 0x6 to addres 0x0 I think that was).

joperezr:
[Triage] This is approved and we are ready for a PR. We will want the bus to handle device management.

pgrawehr:
[...]

PR #1341, opened 2020-12-06
Change Ft4222 to work with multiple I2C devices (I2cBus prototype)

krwq:
Testing idea of I2C bus from: https://github.com/dotnet/iot/issues/1335

also fixing Ft4222 to work with multiple devices. What I just learned is that SenseHat is using same I2cDevice address for both Joystick and LED matrix but it uses 2 devices.

### DIFF
--- a/src/devices/Ft4222/Ft4222I2cBus.cs
+++ b/src/devices/Ft4222/Ft4222I2cBus.cs
@@ -1,0 +1,170 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Device.I2c;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+
+namespace Iot.Device.Ft4222
+{
+    /// <summary>
+    /// FT4222 I2C Device
+    /// </summary>
+    public class Ft4222I2cBus : IDisposable
+    {
+        private const uint I2cMasterFrequencyKbps = 400;
+
+        private SafeFtHandle _ftHandle;
+        private HashSet<int> _usedAddresses = new HashSet<int>();
+
+        /// <summary>
+        /// Store the FTDI Device Information
+        /// </summary>
+        public DeviceInformation DeviceInformation { get; private set; }
+
+        /// <summary>
+        /// Create a FT4222 I2C Device
+        /// </summary>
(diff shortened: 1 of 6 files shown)


src/devices/Ft4222/samples/Program.cs
```diff
@@ -31,13 +31,21 @@
     Console.WriteLine($"Device type: {device.Type}");
 }
 
+if (devices.Count == 0)
+{
+    Console.WriteLine("No devices to connected to run tests.");
```
krwq, line 36:
```suggestion
    Console.WriteLine("No devices connected to run tests.");
```

src/devices/Ft4222/samples/Program.cs
```diff
@@ -55,11 +63,11 @@
...
 
-void TestI2c()
+void TestI2c(DeviceInformation device)
 {
-    using Ft4222I2c ftI2c = new(new I2cConnectionSettings(0, Bno055Sensor.DefaultI2cAddress));
-
-    Bno055Sensor bno055Sensor = new(ftI2c);
+    using Ft4222I2cBus ftI2c = new(device);
+    using I2cDevice i2cDevice = ftI2c.CreateDevice(Bno055Sensor.DefaultI2cAddress);
+    Bno055Sensor bno055Sensor = new(i2cDevice);
```
Ellerbach, line 70: You should add how to create multiple devices maybe. Just to show it's possible

src/devices/Ft4222/Ft4222I2cBus.cs
```diff
@@ -0,0 +1,170 @@
...
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+
+namespace Iot.Device.Ft4222
+{
+    /// <summary>
+    /// FT4222 I2C Device
+    /// </summary>
+    public class Ft4222I2cBus : IDisposable
```
Ellerbach, line 16: So related to the issue about I2cBus, should you directly then propose an I2cBus implementation and use FT4222 as an example of implementation?
On what I can see here, you have a good list of functions :-)

Ellerbach (approved):
Seems great! And a good base to setup I2cBus then!

pgrawehr:
@krwq: The build is broken.

src/devices/Ft4222/Ft4222I2cBus.cs
krwq: do not want to do this as part of this PR

Issue #215, opened 2019-02-10
Create New GpioPin and IGpioPin


shaggygi:
It would be nice to have the ability to provide actual GPIO pin objects to bindings instead of pin numbers.  Currently, we usually have to provide a GPIO controller along with pin numbers or just the pin numbers and the binding will create the controller and configure the pins.  This is a limitation as the pins are usually all on the same controller where I might want to mix it up between dev board, bindings, etc.

### Examples
```csharp
// Can remove the GpioDriver in constructor as each pins knows what controller/driver it uses.
public Mcp23xxx(int deviceAddress, IGpioPin? reset = null, IGpioPin? interruptA = null, IGpioPin? interruptB = null)
```

```csharp
// public class GpioPort(params IGpioPin[] bitIndexes)
GpioPort port = new GpioPort(IGpioPin one, IGpioPin two, IGpioPin three, IGpioPin four);
```

Referencing https://github.com/dotnet/iot/issues/125#issuecomment-460009493
Also referencing #204 as it appears we would want something similar to PwmPins.

krwq:
@shaggygi I actually thought about it a bit and we don't need to change existing design. We will need to add new types of controllers which will map to other controllers.

Example would be something like VirtualController where you can say:
- map this IGpioController to this pin address space
- you will need to have a way of getting virtual address given sub-controller and address in sub-controller address space
- we will need to make sure all devices have a way of providing controller
- settings should be specified up-front meaning address space cannot be modified after making it (this will allow us to make optimizations so that VirtualController with sub VirtualControllers can flatten the graph which technically will make it always a linear mapping to physical controllers except when you ask for virtual addresses of virtual sub-controllers)


but if we are ever going to provide a pin abstraction I think this should be a simple struct of IGpioController and a number

UncleSamSwiss:
Is there any decision/progress on this proposal?

In our application code handling multiple GPIOs in different (unrelated) classes, we see a pattern evolving which would be much easier to implement with `GpioPin` and `IGpioPin`. Also, this API could probably help #347 as you could dispose or not dispose the `IGpioPin` object depending on your needs.

[...]

PR #1342, opened 2020-12-06
WIP: GpioController: LibGpiodDriver: Add methods to handle pins by name

microhobby:
From Linux Kernel docs[1]:

> "Optionally, a GPIO controller may have a "gpio-line-names" property.
> This is an array of strings defining the names of the GPIO lines going
> out of the GPIO controller. This name should be the most meaningful
> producer name for the system, such as a rail name indicating the usage."

`gpiod_line_find` searches for a gpio line that was defined in
`gpio-line-names` and return its handle. This is very useful to avoid
having to search in the datasheets the pin bank/line. The user can
check the labels with the pin names on the PCB and use it instead.

1: https://www.kernel.org/doc/Documentation/devicetree/bindings/gpio/gpio.txt

Example:

The Toradex has a family of computer on modules that are pin compatible. But among the board models, we can have different SoCs. For example the [Apalis Family](https://www.toradex.com/computer-on-modules/apalis-arm-family):

Accesing the `MXM3_3/GPIO2` pin on a [Apalis iMX8QM](https://www.toradex.com/computer-on-modules/apalis-arm-family/nxp-imx-8):

image

Accessing the `MXM3_3/GPIO2` pin on [Apalis iMX6Q](https://www.toradex.com/pt-br/computer-on-modules/apalis-arm-family/nxp-freescale-imx-6):

image

The pin in the form factor (hardware) is the same for the two computer on modules but inside the SoC the gpio bank/line numbers changes because of course they are different (arm32v7 for iMX6 and arm64v8 for iMX8 and etc ...).

Using the  `gpio-line-names` the hardware vendor supply a BSP that is also pin-compatible at the software level.

With the changes proposed in this merge request we can write the follow:

```csharp
// Note that libgpiod bank number is not necessary this will be abstracted by the gpiod_line_find
GpioController gpioController = new GpioController();
gpioController.OpenPin("MXM3_3/GPIO2", PinMode.Output);
gpioController.Write("MXM3_3/GPIO2", PinValue.Low);
```

This will work on both Apalis iMX8 and iMX6, we can use a single code base that will work on all hardware compatible with a family of form factor that is compatible.
 
Raspberry Pi also define `gpio-line-names` on their BSP, the pins can also be accessed by their respective names: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm/boot/dts/bcm2837-rpi-3-b.dts#n61

### DIFF
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -51,6 +51,14 @@ internal partial class Interop
         internal static extern SafeLineHandle gpiod_chip_get_line(SafeChipHandle chip, int offset);
 
         /// <summary>
+        /// Find a GPIO line by its name.
+        /// </summary>
+        /// <param name="name">The name of the GPIO line</param>
+        /// <returns>Handle to the GPIO line or <see langword="null" /> if the line could not be found or an error occurred.</returns>
+        [DllImport(LibgpiodLibrary, SetLastError = true)]
+        internal static extern SafeLineHandle gpiod_line_find(string name);
+
+        /// <summary>
         /// Reserve a single line, set the direction to input.
         /// </summary>
         /// <param name="line">GPIO line handle</param>
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
@@ -20,7 +20,9 @@ namespace System.Device.Gpio.Drivers
         private static string s_consumerName = Process.GetCurrentProcess().ProcessName;
         private readonly object _pinNumberLock;
         private readonly ConcurrentDictionary<int, SafeLineHandle> _pinNumberToSafeLineHandle;
+        private readonly ConcurrentDictionary<string, SafeLineHandle> _pinNameToSafeLineHandle;
         private readonly ConcurrentDictionary<int, LibGpiodDriverEventHandler> _pinNumberToEventHandler;
+        private readonly ConcurrentDictionary<string, LibGpiodDriverEventHandler> _pinNameToEventHandler;
         private readonly int _pinCount;
         private SafeChipHandle _chip;
 
@@ -77,10 +79,25 @@ namespace System.Device.Gpio.Drivers
                 _pinCount = Interop.libgpiod.gpiod_chip_num_lines(_chip);
                 _pinNumberToEventHandler = new ConcurrentDictionary<int, LibGpiodDriverEventHandler>();
                 _pinNumberToSafeLineHandle = new ConcurrentDictionary<int, SafeLineHandle>();
+                _pinNameToEventHandler = new ConcurrentDictionary<string, LibGpiodDriverEventHandler>();
+                _pinNameToSafeLineHandle = new ConcurrentDictionary<string, SafeLineHandle>();
             }
             catch (DllNotFoundException)
             {
                 throw ExceptionHelper.GetPlatformNotSupportedException(ExceptionResource.LibGpiodNotInstalled);
+            }
+        }
+
+        private void AddCallbackForPinValueChangedEvent(LibGpiodDriverEventHandler eventHandler, PinEventTypes eventTypes, PinChangeEventHandler callback)
+        {
+            if ((eventTypes & PinEventTypes.Rising) != 0)
+            {
+                eventHandler.ValueRising += callback;
+            }
+
+            if ((eventTypes & PinEventTypes.Falling) != 0)
+            {
+                eventHandler.ValueFalling += callback;
             }
         }
 
@@ -90,16 +107,21 @@ namespace System.Device.Gpio.Drivers
             if ((eventTypes & PinEventTypes.Rising) != 0 || (eventTypes & PinEventTypes.Falling) != 0)
             {
                 LibGpiodDriverEventHandler eventHandler = _pinNumberToEventHandler.GetOrAdd(pinNumber, PopulateEventHandler);
+                AddCallbackForPinValueChangedEvent(eventHandler, eventTypes, callback);
+            }
+            else
+            {
+                throw ExceptionHelper.GetArgumentException(ExceptionResource.InvalidEventType);
+            }
+        }
 
-                if ((eventTypes & PinEventTypes.Rising) != 0)
-                {
-                    eventHandler.ValueRising += callback;
-                }
-
-                if ((eventTypes & PinEventTypes.Falling) != 0)
-                {
-                    eventHandler.ValueFalling += callback;
-                }
+        /// <inheritdoc/>
+        protected internal override void AddCallbackForPinValueChangedEvent(string pinName, PinEventTypes eventTypes, PinChangeEventHandler callback)
+        {
+            if ((eventTypes & PinEventTypes.Rising) != 0 || (eventTypes & PinEventTypes.Falling) != 0)
+            {
+                LibGpiodDriverEventHandler eventHandler = _pinNameToEventHandler.GetOrAdd(pinName, PopulateEventHandler);
+                AddCallbackForPinValueChangedEvent(eventHandler, eventTypes, callback);
             }
             else
             {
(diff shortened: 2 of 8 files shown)


pgrawehr:
While I think it is a good idea to reduce platform-dependency as far as possible, I'm not sure we should add so much extra code for this to the core. What about just creating a simple module that does the translation? Is there a way to do a string->int transformation by querying some data from libgpiod? 

Additionally, your change breaks the existing tests, likely because the new methods are being used already. Snip of test log:
```
System.Device.Gpio.Tests.RaspberryPiDriverTests.PinValueReadAndWrite [FAIL]
      System.PlatformNotSupportedException : Open pin by name is supported only on Libgpiod driver
      Stack Trace:
        /_/src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs(51,0): at System.Device.Gpio.GpioDriver.ClosePin(String pinName)
        /_/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs(199,0): at System.Device.Gpio.GpioController.ClosePinCore(String pinName)
        /_/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs(584,0): at System.Device.Gpio.GpioController.Dispose(Boolean disposing)
        /_/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs(61,0): at System.Device.Gpio.Tests.GpioControllerTestBase.PinValueReadAndWrite()
    System.Device.Gpio.Tests.RaspberryPiDriverTests.PinValueReadAndWrite [FINISHED] Time: 0.0415625s
    System.Device.Gpio.Tests.RaspberryPiDriverTests.WaitForEventFallingEdgeTest [STARTING]
    System.Device.Gpio.Tests.RaspberryPiDriverTests.WaitForEventFallingEdgeTest [FAIL]
```

microhobby:
> While I think it is a good idea to reduce platform-dependency as far as possible, I'm not sure we should add so much extra code for this to the core. What about just creating a simple module that does the translation? Is there a way to do a string->int transformation by querying some data from libgpiod?
> 

The problem is that it is not as out of the box as just using `OpenPin`, a translation would give us a pair of bank/line, and then we would have to instantiate a new driver to set the bank reference and then do the operations on the pin. Maybe it's a matter of improving the code here to try to reduce duplicate code and work with `SafeLineHandle` reference. 

The ideal would be a redesign of the API to work with a pin object that would have the bank/pin reference instead of just work with the number reference of the pin (something that was inherited from sysfs that doesn't make sense on libgpiod unless your hardware has only one gpio bank)

> Additionally, your change breaks the existing tests, likely because the new methods are being used already. Snip of test log:
> 
> ```
> System.Device.Gpio.Tests.RaspberryPiDriverTests.PinValueReadAndWrite [FAIL]
>       System.PlatformNotSupportedException : Open pin by name is supported only on Libgpiod driver
>       Stack Trace:
>         /_/src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs(51,0): at System.Device.Gpio.GpioDriver.ClosePin(String pinName)
>         /_/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs(199,0): at System.Device.Gpio.GpioController.ClosePinCore(String pinName)
>         /_/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs(584,0): at System.Device.Gpio.GpioController.Dispose(Boolean disposing)
>         /_/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs(61,0): at System.Device.Gpio.Tests.GpioControllerTestBase.PinValueReadAndWrite()
>     System.Device.Gpio.Tests.RaspberryPiDriverTests.PinValueReadAndWrite [FINISHED] Time: 0.0415625s
>     System.Device.Gpio.Tests.RaspberryPiDriverTests.WaitForEventFallingEdgeTest [STARTING]
>     System.Device.Gpio.Tests.RaspberryPiDriverTests.WaitForEventFallingEdgeTest [FAIL]
> ```

Thanks for pointing it @pgrawehr I'm converting this to draft to check the tests. 

I would also like the opinion of @joperezr and @buyaa-n on this.
Thanks

pgrawehr:
> The ideal would be a redesign of the API to work with a pin object that would have the bank/pin reference instead of just work with the number reference of the pin (something that was inherited from sysfs that doesn't make sense on libgpiod unless your hardware has only one gpio bank)

We're considering this, see #215. It's currently not top-prio, though.

krwq:
If you put that directly on LibGpiodDriver and not on generic GpioController I'm fine with adding extra utilities. How would you see the APIs for this?

buyaa-n:
This is a useful feature for LibGpiod driver but seems not supported for other drivers, i don't think we can add it without adding on generic GpioController. Not sure if we want to just throw for the other drivers when accessing string overload, probably just wait for https://github.com/dotnet/iot/issues/215 support added

microhobby:
> This is a useful feature for LibGpiod driver but seems not supported for other drivers, i don't think we can add it without adding on generic GpioController. Not sure if we want to just throw for the other drivers when accessing string overload, probably just wait for #215 support added

Yes, totally agree. Unfortunately, I don't have the time to work on the #215 right now, and it has to be done calmly and think carefully. So, I'm thinking of just adding new methods to the `LibGpiodDriver` to handle this specific behavior:

```csharp
LibGpiodDriver gpioController = new LibGpiodDriver(6);
gpioController.OpenPinByName("MXM3_3/GPIO2", PinMode.Output);
gpioController.WriteByName("MXM3_3/GPIO2", PinValue.Low);
gpioController.ReadByName("MXM3_3/GPIO2");
gpioController.CloseByName("MXM3_3/GPIO2");
```

Summarizing, the idea is to have the controls methods with "ByName" to handle this.
@pgrawehr @krwq @buyaa-n @joperezr let me know what you guys think.

microhobby:
> > This is a useful feature for LibGpiod driver but seems not supported for other drivers, i don't think we can add it without adding on generic GpioController. Not sure if we want to just throw for the other drivers when accessing string overload, probably just wait for #215 support added
> 
> Yes, totally agree. Unfortunately, I don't have the time to work on the #215 right now, and it has to be done calmly and think carefully. So, I'm thinking of just adding new methods to the `LibGpiodDriver` to handle this specific behavior:
> 
> ```cs
> LibGpiodDriver gpioController = new LibGpiodDriver(6);
> gpioController.OpenPinByName("MXM3_3/GPIO2", PinMode.Output);
> gpioController.WriteByName("MXM3_3/GPIO2", PinValue.Low);
> gpioController.ReadByName("MXM3_3/GPIO2");
> gpioController.CloseByName("MXM3_3/GPIO2");
> ```
> 
> Summarizing, the idea is to have the controls methods with "ByName" to handle this.
> @pgrawehr @krwq @buyaa-n @joperezr let me know what you guys think.

Ok, updating it. I forgot a important detail, we will need these on the generic controller right. @buyaa-n  I know it is boring to have this throwing `Not supported` to the others drivers but I think it is the easiest and shortest solution 🙄.

Another thing I thought would be to add maybe another Controller for Libgpiod:

```
// first argument the gpio bank
LibGpiodController controller = new LibGpiodController(6, PinNumberingScheme.Logical);
```

But I'm not sure if it would be worth it. Or maybe add it to the bindings, well, thinking out loud here ... 😁

joperezr:
I wouldn't add an extra controller for libgpiod driver. I think that the way to go about it here would be to just add this functionality only the libgpiodriver, and to have to be called via the driver directly, similar to the way we have some registers exposed via the RaspberryPiDriver directly. Also, regarding the "ByName" suffix, I think that instead of that, a more adecuate solution would be to drop the suffix and instead just create new overloads that take in strings instead of ints to identify pins for that driver. Also, these methods that get implemented directly in the driver should do some of the validation that the controller does today before passing the calls to the driver as you would be skipping that layer. Basically this is more or less what I'm thinking:

```c#
LibGpiodDriver myDriver = new LibGpiodDriver(6);
GpioController myController = new GpioController(PinNumberingScheme.Logical, myDriver);

// When not using the byName overloads, you can just call the controller directly like before:
myController.Open(16, PinMode.Output);
myController.Write(16, PinValue.Low);
myController.Read(16);
myController.Close(16);

// When using the byName overloads, then call via the Driver directly:
myDriver.Open("MXM3_3/GPIO2", PinMode.Output);
myDriver.Write("MXM3_3/GPIO2", PinValue.Low);
myDriver.Read("MXM3_3/GPIO2");
myDriver.Close("MXM3_3/GPIO2");
```

pgrawehr:
Since the GpioController is no longer sealed, adding a LibgpiodController could be an option. Adding public methods to the drivers should only be used if there's a very specific use case where the given methods are used (typically from within a single class). This approach should IMHO not be used for things that are possibly used accross an entire application.

joperezr:
That's a good point @pgrawehr! I think perhaps the best way to go here would be to create a LibGpiodController that exposed those string overloads. If the list of pin names is static, we could even add an Enum so that it is strongly typed.

joperezr:
@microhobby do you think you'll get a chance any time soon to try out the suggestions that @pgrawehr and I had above?

PR #1343, opened 2020-12-06
Added MAX31856 sensor to devices library

amgarnier:
- Created a MAX31856 device utilizing the [Technical Documentation](https://datasheets.maximintegrated.com/en/ds/MAX31856.pdf), Code from other devices in the dotnet-iot library, and inspiration from [MIT Open License Python MAX31856 code found here]( https://github.com/johnrbnsn/Adafruit_Python_MAX31856)
- I configured this device for single shot reads. So that the user can request one temperature read at a time. If you think using an automatic conversion method is needed that can easily be configured.
- I added multiple thermocouple device capabilities
- I only added a throw exception for when the thermocouple is damaged or not connected. I can add an over and under temperature throw exception but felt that it would just unnecessarily complicate the code. I can add it though if you feel it will be useful.
- The fault output pin is not utilized.
- I added code to read the Cold Junction on the device. This is for users that want the temperature on the device and on the thermocouple. For instance if you are using a thermocouple outside on a grill you can also read the ambient temperature outside on the device by the Cold Junction Temperature reading.
- I appreciate any feedback and help. I am a MECHE that develops as a hobby.
Thanks


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1343) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1343) before we can accept your contribution.<br/><br/>:x: amgarnier [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1343)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1343) it.</sub>

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// </summary>
+        public const SpiMode SpiMode = System.Device.Spi.SpiMode.Mode1;
+
+        #endregion
+
+        /// <summary>
+        ///  Reads the temperature from the sensor
+        /// </summary>
+        /// <returns>
+        ///  Temperature in degrees celsius
```
Ellerbach, line 41: Temperature from UnitsNet is a Temperature, so you can then use it as you like and have Celsius or Kelvin or any other. So for the comment, better not to mention it. Or you can give the usual range. Some sensors may have a range, let say -2C to +2C, in this case, better to mention.
Same for the next one.

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Initializes the device and then reads the data next it converts the bytes to thermocouple temperature  and throws an exception if there is an error
+        /// </remarks>
+        private double ReadTemperature()
+        {
+            var spiOutputData = WriteRead(Register.READ_CR0, 16);
+            if ((spiOutputData[16] & (byte)Register.ERROR_OPEN) == (byte)Register.ERROR_OPEN)
+            {
+                throw new Exception("Thermocouple is not connected properly");
```
Ellerbach, line 101: We tend to use IOException for errors reading or writing to sensors

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+            Write(configurationSetting0);
+            Write(configurationSetting1);
+        }
+
+        /// <summary>
+        /// Command to Get Temperature from the Device
+        /// </summary>
+        /// <remarks>
+        /// Initializes the device and then reads the data next it converts the bytes to thermocouple temperature  and throws an exception if there is an error
+        /// </remarks>
```
Ellerbach, line 95: Bonus: You can use the '/// <exception'  to describe the exception

src/devices/MAX31856/Register.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <summary>
+        /// Write Register for CR1
+        /// </summary>
+        WRITE_CR1 = 0x81,
+
+        /// <summary>
+        /// Error code for an open thermocouple line which could be from a broken wire or disconnected device
+        /// </summary>
+        ERROR_OPEN = 0x01 // error code for when the thermocouple is open read on the SR register
+
```
Ellerbach, line 35: extra line

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+        private double ReadTemperature()
+        {
+            var spiOutputData = WriteRead(Register.READ_CR0, 16);
+            if ((spiOutputData[16] & (byte)Register.ERROR_OPEN) == (byte)Register.ERROR_OPEN)
```
Ellerbach, line 99: So what happen when the SPI device just receive 0x00 on the 17 bytes? You won't then detect an error and the message will be seen as valid. Is there any other way to check that the WriteRead has been successfull?

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        #endregion
+
+        /// <summary>
+        ///  Reads the temperature from the sensor
+        /// </summary>
+        /// <returns>
+        ///  Temperature in degrees celsius
+        /// </returns>
+        public Temperature GetTemperature() => Temperature.FromDegreesCelsius(ReadTemperature());
```
Ellerbach, line 43: As the ReadTemperature() function can raise exceptions, you should implement a 'TryGetTemperature' pattern. You'll find this pattern. See here: https://github.com/dotnet/iot/tree/master/src/devices#naming-of-functions

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            var temperature = ConvertspiOutputDataTemp(spiOutputData);
+            return temperature;
+        }
+
+        /// <summary>
+        /// Command to Get Temperature from the Device
+        /// </summary>
+        /// <remarks>
+        /// Initializes the device and then reads the data for the cold junction temperature also checks for errors to throw
```
Ellerbach, line 112: I don't see where you throw an exception in the flow :-)

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// Takes the spi data as an input and outputs the Thermocouple Temperature Reading
+        /// </remarks>
+        /// <param name="spiOutputData">Spidata read from the device as 16 bytes</param>
+        private double ConvertspiOutputDataTemp(byte[] spiOutputData)
+        {
+            double tempRaw = (((spiOutputData[13] & 0x7F) << 16) + (spiOutputData[14] << 8) + (spiOutputData[15]));
+            double temperature = tempRaw / 4096; // temperature in degrees C
+            if ((spiOutputData[13] & 0x80) == 0x80) // checks if the temp is negative
+            {
+                temperature = temperature * -1;
```
Ellerbach, line 134: Maybe `temperature = -temperature;`?

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// Takes the spi data as an input and outputs the Cold Junction Temperature Reading
+        /// </remarks>
+        /// <param name="spiOutputData">Spidata read from the device as 16 bytes</param>
+        private double ConvertspiOutputDataTempColdJunction(byte[] spiOutputData)
+        {
+            var tempRaw = (((spiOutputData[11] & 0x7F) << 8) + spiOutputData[12]);
+            double temperatureColdJunction = tempRaw / 256; // convert decimal ouput to temperature using a constant from the Technical Manual
+            if ((spiOutputData[11] & 0x80) == 0x80) // checks if the temp is negative
+            {
+                temperatureColdJunction = temperatureColdJunction * -1;
```
Ellerbach, line 153: Same here, maybe `temperatureColdJunction = -temperatureColdJunction;`?

src/devices/MAX31856/README.md
```diff
@@ -0,0 +1,51 @@
+﻿# MAX31856
+
+## Summary
```
Ellerbach, line 3: Markdown 101 tip: there is a new line before and after all ## titles (and any other title except the first # which has only 1 line after and must be unique)

src/devices/MAX31856/README.md
```diff
@@ -0,0 +1,51 @@
...
+The MAX31856 device is a SPI interface cold-junction compensated thermocouple to digital converter.
+
+## Sensor Image
+![](device.jpg)
```
Ellerbach, line 7: Place a description of the picture here
This comment is used for accessibility

src/devices/MAX31856/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Usage
+The MAX31856.samples file contains a sample usage of the device. Note that this reads two temperatures. One is a connected thermocouple reading called ```temp``` and the other is the temperature of the device itself called ```tempCJ``` which is used internally to increase the accuracy of the thermocouple but can also be read if you find a use for it.
+
+Create a new ```SpiConnectionSettings``` Class if using a raspberry pi3 do not change these settings.
```
Ellerbach, line 12: `Raspberry Pi` (version doesn't matter) and then the cast is important as it's a brand.

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+            return temperature;
+        }
+
+        /// <summary>
+        /// Command to Get Temperature from the Device
+        /// </summary>
+        /// <remarks>
+        /// Initializes the device and then reads the data for the cold junction temperature also checks for errors to throw
+        /// </remarks>
+        public double ReadCJTemperature()
```
Ellerbach, line 114: Maybe private as you have a property?

src/devices/MAX31856/README.md
```diff
@@ -0,0 +1,51 @@
...
+using MAX31856 sensor = new(device, ThermocoupleType.K);
+```
+
+Now read the temperature from the device. Using the UnitsNet nuget you can convert from DegreesCelsius to DegreesFahrenheit or any other unit by changing ```.GetTemperature().DegreesFahrenheit``` to another unit of your choice.
+
+```csharp
+while (true)
+{
+    // read temperature
+    var temp = sensor.GetTemperature().DegreesFahrenheit;
```
Ellerbach, line 35: As the GetTemperature raises exception, you should use a Try Catch. And that's the reason why it's better to use the TryGetTemperature pattern

src/devices/MAX31856/samples/MAX31856.Sample.cs
```diff
@@ -0,0 +1,30 @@
...
+    Mode = MAX31856.SpiMode,
+    DataFlow = 0
+};
+
+using SpiDevice device = SpiDevice.Create(settings);
+using MAX31856 sensor = new(device, ThermocoupleType.K);
+while (true)
+{
+    // read temperature
+    var temp = sensor.GetTemperature().DegreesFahrenheit;
```
Ellerbach, line 21: As pointed in the Readme, catch the exception or better, use a TryGetTemperature pattern

src/devices/MAX31856/samples/README.md
```diff
@@ -0,0 +1,4 @@
+# TODO: This needs to be determined
```
Ellerbach, line 1: TODO: remove this file :-) It's not needed as you've placed everything into the main read me

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// </returns>
+        public Temperature GetTemperature() => Temperature.FromDegreesCelsius(ReadTemperature());
+
+        /// <summary>
+        ///  Reads the temperature from the sensor
+        /// </summary>
+        /// <returns>
+        ///  Temperature in degrees celsius
+        /// </returns>
+        public Temperature GetCJTemperature() => Temperature.FromDegreesCelsius(ReadCJTemperature());
```
Ellerbach, line 51: We try to avoid abbreviations if they are not well know. So even if a bit long, I would go for GetColdJunctionTemperature

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// <summary>
+        ///  Reads the temperature from the sensor
+        /// </summary>
+        /// <returns>
+        ///  Temperature in degrees celsius
+        /// </returns>
+        public Temperature GetTemperature() => Temperature.FromDegreesCelsius(ReadTemperature());
+
+        /// <summary>
+        ///  Reads the temperature from the sensor
```
Ellerbach, line 46: Maybe add it's the Cold-Junction temperature that you read here

Ellerbach (commented):
Thanks a lot for this nice addition. Looks all up very good, few comments to make it even better!

src/devices/MAX31856/MAX31856.cs
amgarnier: yes that would be better

src/devices/MAX31856/MAX31856.cs
amgarnier: left over code from when I was going to have it throw an error

src/devices/MAX31856/MAX31856.cs
amgarnier: updated code to reflect IOException

src/devices/MAX31856/MAX31856.cs
amgarnier: I added this but am not sure if I am implementing it correctly.

src/devices/MAX31856/MAX31856.cs
amgarnier: I have added code to read CR0 after initialization of it. If it reads the same value it initializes to then the device is wired correctly. If not then it throws an exception. This is under the new TryGetTemperature class I have a separate exception for the thermocouple being wired properly.

src/devices/MAX31856/MAX31856.cs
amgarnier: Implemented, I'm not sure if I am error handling correctly. Also if it outputs bool I have have it read true if the device is working properly but I am not sure how to set the bool false if there is an exception

src/devices/MAX31856/samples/MAX31856.Sample.cs
```diff
@@ -17,13 +18,15 @@
...
 {
-    // read temperature
-    var temp = sensor.GetTemperature().DegreesFahrenheit;
-    temp = Math.Round(temp, 7); // round temp output to seven significant figures from resolution in Technical Documentation
-    // read cold junction temperature of device
-    var tempCJ = sensor.GetCJTemperature().DegreesFahrenheit;
-    tempCJ = Math.Round(tempCJ, 2); // round temp output to two significant figures from resolution in Technical Documentation
-    Console.WriteLine($"Temp: {temp} ColdJunctionTemp: {tempCJ} ");
+    // Reads temperature if the device is not reading properly
+    sensor.TryGetTemperature(out Temperature temperature);
```
Ellerbach, line 22: with the TryReadTemperature pattern, it should then be like:

```csharp
var tempColdJunction = sensor.GetCJTemperature();
if (sensor.TryGetTemperature(out Temperature temperature))
{
  Console.WriteLine($"Temperature: {temperature.DegreesFahrenheit:0.0000000} °F, Cold Junction: {tempColdJunction.DegreesFahrenheit:0.00} °F");
}
else
{
  Console.WriteLine($"Error reading temperature, Cold Junction temperature: {tempColdJunction.DegreesFahrenheit:0.00}");
}
```

Ellerbach (commented):
Thanks for adjusting, few comments left.

src/devices/MAX31856/MAX31856.cs
```diff
@@ -35,12 +36,29 @@ public class MAX31856 : IDisposable
...
+        {
+            try
+            {
+                temperature = ReadTemperature(out byte[] spiOutputData);
+                return true;
+            }
+            catch (IOException e)
+            {
+                // The tempearture is set to 1 million if there is an exception
+                temperature = Temperature.FromDegreesCelsius(1_000_000);
```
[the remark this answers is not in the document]
amgarnier: How would you recommend setting a temperature default? I'm sorry not sure. The issue im having is I cant set unitnet Temperature to null

src/devices/MAX31856/samples/MAX31856.Sample.cs
amgarnier: This is great thank you so much.

src/devices/MAX31856/MAX31856.cs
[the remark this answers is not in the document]
Ellerbach: Just `new Temperature()`

src/devices/MAX31856/MAX31856.cs
```diff
@@ -48,18 +48,17 @@ public bool TryGetTemperature(out Temperature temperature)
...
                 return false;
             }
         }
 
         /// <summary>
-        ///  Reads the temperature from the sensor
+        ///  Reads the temperature from the Cold-Junction sensor
         /// </summary>
         /// <returns>
         ///  Temperature in degrees celsius
```
Ellerbach, line 64: I would got for the long name here as well GetColdJunctionTemperature()

src/devices/MAX31856/MAX31856.cs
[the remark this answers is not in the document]
amgarnier: I updated this thank you

src/devices/MAX31856/MAX31856.cs
amgarnier: updated thank oyu

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,211 @@
...
+        public bool TryGetTemperature(out Temperature temperature)
+        {
+            try
+            {
+                temperature = ReadTemperature(out byte[] spiOutputData);
+                return true;
+            }
+            catch (IOException)
+            {
+                // The tempearture is set to 1 million if there is an exception
```
Ellerbach, line 53: I guess this is not true anymore :-D

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,211 @@
...
+                // Write error message to council
+                return false;
+            }
+        }
+
+        /// <summary>
+        ///  Reads the temperature from the Cold-Junction sensor
+        /// </summary>
+        /// <returns>
+        ///  Temperature in degrees celsius
```
Ellerbach, line 64: It's just a temperature now :-) But you can add the precision and range. So would be something like (adjust to the real values):
/// Temperature, precision +- 0.01 Celsius range from -100 Celsius to +200 Celsius

src/devices/MAX31856/README.md
```diff
@@ -0,0 +1,51 @@
...
+Create a new ```SpiDevice``` with the settings from above. Then create a new MAX31856 device with the ```SpiDevice``` as well as the correct ```ThermocoupleType``` (see note below)
+```csharp
+using SpiDevice device = SpiDevice.Create(settings);
+using MAX31856 sensor = new(device, ThermocoupleType.K);
+```
+
+Now read the temperature from the device. Using the UnitsNet nuget you can convert from DegreesCelsius to DegreesFahrenheit or any other unit by changing ```.GetTemperature().DegreesFahrenheit``` to another unit of your choice.
+
+```csharp
+while (true)
```
Ellerbach, line 32: You should adjust the documentation as this has changed now

Ellerbach (approved):
Looks good to me. Little minor changes in documentation still needed

src/devices/MAX31856/samples/MAX31856.Sample.cs
```diff
@@ -18,8 +18,8 @@
 using MAX31856 sensor = new(device, ThermocoupleType.K);
 while (true)
 {
-    // Reads temperature if the device is not reading properly
     var tempColdJunction = sensor.GetColdJunctionTemperature();
+    // Attempts to read temperature if an error is thrown the temperature will read as the default value of temperature in the Temperature class
```
Ellerbach, line 22: You should adjust this comment: there won't be any error thrown as the TryGetTemperature pattern prevent this. You'll just get false as a result and a Temperature that is not a correct one

Ellerbach (approved):
Thanks, looks good all up. I'll let @krwq or @joperezr having a look and giving potentially additional comments.

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        /// <summary>
+        /// Spi Clock Frequency from the Technical Manual of the device
+        /// </summary>
+        public const int SpiClockFrequency = 5_000_000;
+
+        /// <summary>
+        /// Spi Clock Frequency from the Technical Manual of the device
+        /// </summary>
+        public const int DataFlow = 0;
```
krwq, line 29: This should be:

`public const DataFlow SpiDataFlow = DataFlow.<whatever>;`

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        #region SpiSettings
+
+        /// <summary>
+        /// Spi Clock Frequency from the Technical Manual of the device
+        /// </summary>
+        public const int SpiClockFrequency = 5_000_000;
+
+        /// <summary>
+        /// Spi Clock Frequency from the Technical Manual of the device
```
krwq, line 27: description is same as for frequency

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// </summary>
+        public const int DataFlow = 0;
+
+        /// <summary>
+        /// MAX31856 SPI Mode
+        /// </summary>
+        public const SpiMode SpiMode = System.Device.Spi.SpiMode.Mode1;
```
krwq, line 34: `using System.Device.Spi;` and try avoiding full namespace unless you got name collision

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// Command to Get Temperature from the device
+        /// </summary>
+        /// <remarks>
+        /// Initializes the device and then reads the data next it converts the bytes to thermocouple temperature
+        /// </remarks>
+        public bool TryGetTemperature(out Temperature temperature)
+        {
+            try
+            {
+                temperature = ReadTemperature(out byte[] spiOutputData);
```
krwq, line 48: this should be implemented reverse - the other method instead of throwing should be returning false, and GetTemperature should throw when TryGetTemperature returns false so that you avoid any throwing/catching

krwq: If you want to preserve extra information about error then either implement it twice or have an internal overload which returns some kind of enum with Ok, ErrorWhatever1, ErrorWhatever2

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// Temperature, precision +- 0.7 Celsius range from -20 Celsius to +85 Celsius
+        /// </returns>
+        public Temperature GetColdJunctionTemperature() => Temperature.FromDegreesCelsius(ReadCJTemperature());
+
+        /// <summary>
+        /// Creates a new instance of the max31856.
+        /// </summary>
+        /// <param name="spiDevice">The communications channel to a device on a SPI bus</param>
+        /// <param name="thermocoupleType">Thermocouple Type T,K,R</param>
+        public MAX31856(SpiDevice spiDevice, ThermocoupleType thermocoupleType)
```
krwq, line 71: Should you default thermocoupleType to something sensible so that user doesn't have to know about it? I personally have no clue what that means 😄

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        #endregion
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            _spiDevice?.Dispose();
+            _spiDevice = null!;
+        }
+
```
krwq, line 207: nit: extra enter

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// <summary>
+        /// Full Duplex Read of the Data on the Device
+        /// </summary>
+        /// <remarks>
+        /// Writes the read address of the register and outputs a byte list of the length provided
+        /// </remarks>
+        /// /// <param name="register">Register location to write to which starts the device reading</param>
+        /// /// <param name="readbytesize">Number of bytes being read</param>
+        private byte[] WriteRead(Register register, int readbytesize)
```
krwq, line 188: optional: consider changing this so that you do not need to allocate an output buffer (i.e. take Span as input instead of int and stackalloc at the callsite)

src/devices/MAX31856/README.md
```diff
@@ -0,0 +1,53 @@
+﻿# MAX31856
```
krwq, line 1:
```suggestion
﻿# MAX31856 - cold-junction compensated thermocouple to digital converter
```

we used this specific line to generate device listing (src/devices/README.md)

src/devices/MAX31856/README.md
```diff
@@ -0,0 +1,53 @@
...
+    ClockFrequency = MAX31856.SpiClockFrequency,
+    Mode = MAX31856.SpiMode,
+    DataFlow = 0
+};
+```
+
+Create a new ```SpiDevice``` with the settings from above. Then create a new MAX31856 device with the ```SpiDevice``` as well as the correct ```ThermocoupleType``` (see note below)
+```csharp
+using SpiDevice device = SpiDevice.Create(settings);
+using MAX31856 sensor = new(device, ThermocoupleType.K);
```
krwq, line 26: I suggest to put only general info about the sensor in this README.md and move sample code to samples/README.md (treat as optional feedback but we do that in most of the devices)

src/devices/MAX31856/ThermocoupleType.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Iot.Device.MAX31856
+{
+    /// <summary>
+    /// Sets the Thermocouple Type
+    /// </summary>
+    public enum ThermocoupleType : byte
+    {
+        /// <summary>
+        /// Type B thermocouple
```
krwq, line 12: If this is known domain knowledge leave this as is otherwise I'd recommend to write a bit more about what this means

src/devices/MAX31856/samples/MAX31856.Sample.cs
```diff
@@ -0,0 +1,34 @@
...
+    Mode = MAX31856.SpiMode,
+    DataFlow = 0
+};
+
+using SpiDevice device = SpiDevice.Create(settings);
+using MAX31856 sensor = new(device, ThermocoupleType.K);
+while (true)
+{
+    var tempColdJunction = sensor.GetColdJunctionTemperature();
+    // Attempts to read temperature if it is in error "false" will be returned. The Temperature that is shown will default to the UnitsNet Default value and will not be the correct value of the thermocouple.
```
krwq, line 22: the comment is not necessary since you print "Error" when this returns false

src/devices/MAX31856/samples/MAX31856.Sample.cs
```diff
@@ -0,0 +1,34 @@
...
+    if (sensor.TryGetTemperature(out Temperature temperature))
+    {
+        Console.WriteLine($"Temperature: {temperature.DegreesFahrenheit:0.0000000} °F, Cold Junction: {tempColdJunction.DegreesFahrenheit:0.00} °F");
+    }
+    else
+    {
+        Console.WriteLine($"Error reading temperature, Cold Junction temperature: {tempColdJunction.DegreesFahrenheit:0.00}");
+    }
+
+    // wait for 2000ms
```
krwq, line 32: also line below is self explanatory and doesn't need a comment

src/devices/MAX31856/samples/MAX31856.Sample.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Device.Spi;
+using System.Threading;
+using Iot.Device.MAX31856;
+using UnitsNet;
+
+SpiConnectionSettings settings = new(0, 0)
+{
+    ClockFrequency = MAX31856.SpiClockFrequency,
+    Mode = MAX31856.SpiMode,
+    DataFlow = 0
```
krwq, line 14: MAX31856.SpiDataFlow

krwq:
Overall LGTM, nothing major

src/devices/MAX31856/MAX31856.cs
amgarnier: Its typical with this device to use other thermocouple types such as K. Most common is T.

src/devices/MAX31856/MAX31856.cs
amgarnier: Sorry this is too advanced for me :/

src/devices/MAX31856/ThermocoupleType.cs
amgarnier: This is domain knowledge  (it is a type of thermocouple probe) https://en.wikipedia.org/wiki/Thermocouple#Types

src/devices/MAX31856/MAX31856.cs
```diff
@@ -46,12 +46,12 @@ public bool TryGetTemperature(out Temperature temperature)
             try
             {
                 temperature = ReadTemperature(out byte[] spiOutputData);
-                return true;
+                return false;
```
krwq, line 49: that's not what I meant when I said "Reverse" I meant to avoid try..catch at all by not throwing and having TryGetTemperature method doing the communication and ReadTemperature method calling into TryReadTemperature and throwing based on what it returns.

The previous `true`/`false` behavior was correct (true on successful reads)

src/devices/MAX31856/samples/README.md
```diff
@@ -0,0 +1,39 @@
+## Usage
+The MAX31856.samples file contains a sample usage of the device. Note that this reads two temperatures. One is a connected thermocouple reading which can be read using the  ```TryGetTemperature``` command and the other is the temperature of the device itself which can be read using the ```GetColdJunctionTemperature``` command. The Cold Junction Temperature is used internally to increase the accuracy of the thermocouple but can also be read if you find a use for it.
```
krwq, line 2: - extra enter after header
- header should be `#` rather than `##`

src/devices/MAX31856/MAX31856.cs
amgarnier: I attempted to fix this but am not sure if I am doing it correctly. Sorry Im new to C# and doing my best to learn it! (Mechanical engineering background).

I am also getting this error all of a sudden in VS 2019:
Error	NU1007	The runtime.json specified in the project 'C:\Program Files\dotnet\sdk\5.0.100\RuntimeIdentifierGraph.json' could not be found.	System.Device.Gpio	C:\Dev\iot2\src\System.Device.Gpio\System.Device.Gpio.csproj

I couldn't figure out a way to fix this.

src/devices/MAX31856/samples/README.md
amgarnier: updated

amgarnier (commented):
Thanks for all of your help

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,217 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Device.Spi;
+using System.IO;
+using UnitsNet;
+
+namespace Iot.Device.MAX31856
```
krwq, line 9: Per our [coding guidelines](https://github.com/dotnet/iot/blob/master/Documentation/Coding-guidelines.md#coding-guideliness) this should be spelled as `Iot.Device.Max31856`

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,217 @@
...
+{
+    /// <summary>
+    /// Documentation https://datasheets.maximintegrated.com/en/ds/MAX31856.pdf
+    /// </summary>
+    public class MAX31856 : IDisposable
```
krwq, line 14:
```suggestion
    public class Max31856 : IDisposable
```

src/devices/MAX31856/MAX31856.cs
krwq: actually we can just create a single `Temperature GetTemperature()` method which throws on error, no other `TryGetTemperature` etc.  that will simplify it

src/devices/MAX31856/MAX31856.cs
krwq: no worries. in here we do not require having TryGetTemperature method, we can just to a single GetTemperature() or even property `Temperature`. TryGetTemperature method only makes sense if device is error prone (i.e. 1-wire thermomers are not working reliably as they are very time sensitive which we can't do correctly because OS can interrupt our process which causes error in transmission. In cases like that throwing and having to catch is a frequent operation and try .. catching is slower than not throwing at all. Easy way to explain is i.e.:

```csharp
boo TryGetTemperature(out Temperature temperature)
{
   temperature = ReadTemperatureFromRegister();
   if (DeviceStatuseShowsError())
   {
      temperature = default(Temperature);
      return false;
   }
  return true;
}
```

rather than throwing and try catching

krwq: (as I said, in this case in this device errors are not expected so not having Try.. method is acceptable)

src/devices/MAX31856/MAX31856.cs
amgarnier: I updated all code to reflect Max31856 vice MAX31856

src/devices/MAX31856/MAX31856.cs
amgarnier: I just hve the get temperature to simplify it. I agree with you that it was overly complicated for a simple device.

src/devices/MAX31856/MAX31856.cs
amgarnier: Ok did that makese sense

src/devices/MAX31856/MAX31856.cs
amgarnier: updated

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,190 @@
...
+
+using System;
+using System.Device.Spi;
+using System.IO;
+using UnitsNet;
+
+namespace Iot.Device.Max31856
+{
+    /// <summary>
+    /// Documentation https://datasheets.maximintegrated.com/en/ds/MAX31856.pdf
```
krwq, line 12:
```suggestion
    /// MAX31856  - cold-junction compensated thermocouple to digital converter
```

Datasheet link should be in the README.md

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,190 @@
...
+    /// </summary>
+    public class Max31856 : IDisposable
+    {
+        private readonly byte _thermocoupleType;
+        private SpiDevice _spiDevice;
+
+        #region SpiSettings
+
+        /// <summary>
+        /// Spi Clock Frequency from the Technical Manual of the device
```
krwq, line 22:
```suggestion
        /// SPI Clock Frequency
```

Note that users of this library do not need to know where this came from when reading docs

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// </summary>
+        public const int SpiClockFrequency = 5_000_000;
+
+        /// <summary>
+        /// Set the SPI data flow to MSB First
```
krwq, line 27:
```suggestion
        /// SPI data flow
```

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// </summary>
+        public const DataFlow SpiDataFlow = DataFlow.MsbFirst;
+
+        /// <summary>
+        /// Max31856 SPI Mode
```
krwq, line 32:
```suggestion
        /// SPI Mode
```

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,190 @@
...
+        public const SpiMode SpiModeSetup = SpiMode.Mode1;
+
+        #endregion
+
+        /// <summary>
+        /// Command to Get Temperature from the device
+        /// </summary>
+        /// <remarks>
+        /// Initializes the device and then reads the data converting the bytes to thermocouple temperature
+        /// </remarks>
```
krwq, line 43: This is an implementation detail

```suggestion
```

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,190 @@
...
+        public Temperature GetTemperature()
+        {
+            Temperature temperature = ReadTemperature(out byte[] spiOutputData);
+            return temperature;
```
krwq, line 47:
```suggestion
            return ReadTemperature(out byte[] spiOutputData);
```

krwq: (you can use also one liner: `=> ReadTemperature(...`

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// <summary>
+        ///  Reads the temperature from the Cold-Junction sensor
+        /// </summary>
+        /// <returns>
+        /// Temperature, precision +- 0.7 Celsius range from -20 Celsius to +85 Celsius
+        /// </returns>
+        public Temperature GetColdJunctionTemperature() => Temperature.FromDegreesCelsius(ReadCJTemperature());
+
+        /// <summary>
+        /// Creates a new instance of the max31856.
```
krwq, line 59:
```suggestion
        /// Creates a new instance of the MAX31856.
```

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// </summary>
+        /// <param name="spiDevice">The communications channel to a device on a SPI bus</param>
+        /// <param name="thermocoupleType">Thermocouple Default Type is T can change to B,E,J,K,N,R,S</param>
```
krwq, line 62:
```suggestion
        /// <param name="thermocoupleType">Thermocouple Type. It default to T.</param>
```

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,190 @@
...
+        public Max31856(SpiDevice spiDevice, ThermocoupleType thermocoupleType)
```
krwq, line 63:
```suggestion
        public Max31856(SpiDevice spiDevice, ThermocoupleType thermocoupleType = ThermocoupleType.T)
```

(the docs say it defaults to T but it doesn't in the code)

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// </remarks>
+        /// <param name="spiOutputData">Spidata read from the device as 16 bytes</param>
+        private double ConvertspiOutputDataTempColdJunction(byte[] spiOutputData)
+        {
+            var tempRaw = (((spiOutputData[11] & 0x7F) << 8) + spiOutputData[12]);
+            double temperatureColdJunction = tempRaw / 256; // convert decimal ouput to temperature using a constant from the Technical Manual
+            if ((spiOutputData[11] & 0x80) == 0x80) // checks if the temp is negative
+            {
+                temperatureColdJunction = -temperatureColdJunction;
+            }
```
krwq, line 144: This should be something along the lines of:

```csharp
Span<byte> reading = new Span<byte>(spiOutputData, 11, 2);
short tempRaw = BinaryPrimitives.ReadInt16BigEndian(reading);
return tempRaw / 256.0;
```

note: `256.0` not `256`, dividing integer by integer will lose everything after decimal separator

krwq (commented):
LGTM after fixing requested changes

src/devices/MAX31856/MAX31856.cs
amgarnier: I updated the code. The only issue is that the previous code removes the first two digits of byte 11 so I have to code that in to read the proper temperature. 

reading[0] = (byte)(spiOutputData[11] & 0x3F);

src/devices/MAX31856/MAX31856.cs
```diff
@@ -136,14 +133,15 @@ private Temperature ReadTemperature(out byte[] spiOutputData)
...
+            reading[0] = (byte)(spiOutputData[11] & 0x3F);
+            short tempRaw = BinaryPrimitives.ReadInt16BigEndian(reading);
             if ((spiOutputData[11] & 0x80) == 0x80) // checks if the temp is negative
             {
-                temperatureColdJunction = -temperatureColdJunction;
+                return -tempRaw / 256.0;
             }
 
-            return temperatureColdJunction;
+            return tempRaw / 256.0;
```
krwq: ReadInt16BigEndian will do this part for you (you can just remove it), now you will double negate.

src/devices/MAX31856/MAX31856.cs
```diff
@@ -136,14 +133,15 @@ private Temperature ReadTemperature(out byte[] spiOutputData)
         /// <param name="spiOutputData">Spidata read from the device as 16 bytes</param>
         private double ConvertspiOutputDataTempColdJunction(byte[] spiOutputData)
         {
-            var tempRaw = (((spiOutputData[11] & 0x7F) << 8) + spiOutputData[12]);
-            double temperatureColdJunction = tempRaw / 256; // convert decimal ouput to temperature using a constant from the Technical Manual
+            Span<byte> reading = new Span<byte>(spiOutputData, 11, 2);
+            reading[0] = (byte)(spiOutputData[11] & 0x3F);
```
krwq, line 137: I'm not sure what are you doing here... with this you will never get negative number below

src/devices/MAX31856/MAX31856.cs
krwq: Looking at the documentation I have some questions (https://datasheets.maximintegrated.com/en/ds/MAX31856.pdf page 24):
- why 11 and 12, the doc says about 0xA and 0xB which maps to 10 and 11
- the way spec describes this the code should be more like (written by hand, not tested):

```csharp
byte msb = ...; // 0xA
byte lsb = ...; // 0xB

bool sign = (msb >> 7) == 1;
int raw = ((msb & 0x7F) << 6) | (lsb >> 2);
if (sign)
{
    raw = -raw;
}

return raw / 64.0;
```

src/devices/MAX31856/MAX31856.cs
krwq: disregard, per docs this looks correct

src/devices/MAX31856/MAX31856.cs
krwq: should this 0x7F?

src/devices/MAX31856/MAX31856.cs
amgarnier: the var spiOutputData = WriteRead(Register.READ_CR0, 16); reads 16 bits starting at 1. So it shifts the reading by 1

src/devices/MAX31856/MAX31856.cs
amgarnier: Yes good catch! The first bits on byte 11(0ah)  is for temperature sign. We have to strip it out and combine the remaining 7 bits of byte 11 with byte 12 to get temperature. My mind jumped and I thought it was the first two hence the 0x3f it should be 0x7f

krwq (approved):

krwq:
Seems your build is failing on Linux, looking at the fact that only Linux is failing and the fact it cannot find a project my guess is that you have casing wrong in the ProjectReference in the csproj in the sample project

src/devices/MAX31856/MAX31856.cs
```diff
@@ -0,0 +1,188 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
krwq, line 1: also please remove  ConsoleApp1 file from the PR

amgarnier: I cant find where this file is thats weird

amgarnier: I figured it out it has to do with the files name changing to Max31856 vice MAX31856 changed all files to reflect Max31856

amgarnier: One last thing the main src folder of MAX31856 is wrong I dont know how to change the folder name in src to Max31856 and push it

eng/stylecop.json
```diff
@@ -16,4 +16,3 @@
         }        
     }
   }
-
```
krwq, line 19: revert

src/devices/MAX31856/MAX31856.cs
krwq: do two commits:
- rename to MAX31856-2
- commit
- rename to whatever

eng/stylecop.json
amgarnier: I can not figure out why vs 2019 keeps addinig that blank line

krwq: don't open VS when you revert :)

krwq (approved):

krwq:
@amgarnier I went ahead and reverted the change in the stylecop.json file in this PR. once CI is green we can merge

krwq:
Thanks @amgarnier!

amgarnier:
Thank you guys for all of your help I learned alot

PR #1355, opened 2020-12-09
only throw exception if glyph data not found

### LINKED ISSUE #1344: BdfFont class cannot ever successfully call GetCharData()
**Describe the bug**

When calling [BdfFont.GetCharData()](https://github.com/dotnet/iot/blob/6537bbabf7424b0e0d561bd4a3bf98de35505b8f/src/devices/Common/Iot/Device/Graphics/BdfFont.cs#L135), the code will always throw an exception

**Steps to reproduce**
1. Load a BdfFont
2. Call GetCharData()

or

1. Run the [RGBLedMatrix sample](https://github.com/dotnet/iot/tree/master/src/devices/RGBLedMatrix/samples)
2. Press 1 to switch to scenario 1

**Expected behavior**

It should return the data for the character and the sample should run without throwing an exception. It looks like the throw should be in an else clause for that if statement. I made that change and the sample seems to run fine.

**Actual behavior**

It always throws an exception.

**Versions used**
 Version:   5.0.100
 Commit:    5044b93829

Runtime Environment:
 OS Name:     raspbian
 OS Version:  10
 OS Platform: Linux
 RID:         linux-arm
 Base Path:   /home/pi/dotnet/sdk/5.0.100/

Host (useful for support):
  Version: 5.0.0
  Commit:  cf258a14b7

.NET SDKs installed:
  5.0.100 [/home/pi/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 5.0.0 [/home/pi/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 5.0.0 [/home/pi/dotnet/shared/Microsoft.NETCore.App]

negativeeddy:
Fixes #1344
- Problem: All calls to BdfFontGet.GetCharData() will throw an exception, regardless of whether the font was loaded properly or not.
- Fix: only throw an exception if the glyph data cannot be located

### DIFF
--- a/src/devices/Common/Iot/Device/Graphics/BdfFont.cs
+++ b/src/devices/Common/Iot/Device/Graphics/BdfFont.cs
@@ -131,8 +131,10 @@ namespace Iot.Device.Graphics
             {
                 charData = GlyphUshortData.AsSpan().Slice(index, Height);
             }
-
-            throw new InvalidDataException("Couldn't get the glyph data");
+            else
+            {
+                throw new InvalidDataException("Couldn't get the glyph data");
+            }
         }
 
         /// <summary>


krwq (approved):

krwq:
Thanks for fixing this!

Issue #214, opened 2019-02-09
I2C APIs should take address on Read/Write


shaggygi:
I started a binding for the [At24Cxx Serial EEPROM](http://ww1.microchip.com/downloads/en/devicedoc/doc0180.pdf) and wanted to ping for thoughts before proceeding.

The problem I noticed while reading the datasheets is the I2C addresses vary based on the size of memory.  For example, the AT24C01/AT24C02 uses the hardware pins for addressing A0-A2 and the page addresses are in the first byte read/written (as in most cases).

However, as you work your way up to the AT24C16, it begins using the address bits A0-A2 instead to address the pages.  **Figure 7. Device Address** in datasheet link above explains.  This is somewhat of a hindrance in the I2cDevice implementation as you must assign the I2C address to the I2cConnectionSettings and then instantiate the I2cDevice.  The I2cDevice object is assigned the I2C address for the lifetime of the object.

I'm hesitantly okay with this as you could design the binding where certain devices would have to use multiple I2cDevice objects to access all available memory.  Basically, the one device would look like multiple devices on the bus.  For example, the AT24C16 would need 8 I2cDevice objects to address all memory using addresses 0x50 - 0x57 where each object could access 16 pages.

The problem I see with this approach, is there will be some users that want to interact with the device/memory as a whole.  I've personally seen where an IoT device has a 4K size of configuration data that is loaded into EEPROM.  Ideally, I would use a method call that provides starting address and bytes of data to write.  In general use, the binding API would need some internal logic to determine the paging (no matter what AT24C device) and would offer a little more appeasing API surface for Reads/Writes.  However, I'm not sure of a good solution being forced to instantiate multiple I2cDevice objects to accomplish this.

/cc @joperezr @JeremyKuhne @krwq  I'm copying you as you might have some thoughts how to get around this with current implementation or ideas on how to improve the II2cController that is in the works.

Since most of our current bindings pass in an I2cDevice object to instantiate the binding, I'm not sure how we could recreate another I2cDevice object with the new address to use when attempting to access related pages.

Would it be beneficial to add the following methods where it would override the initial address and use one provided?  I could see this confusing some users as they'll see this as an option and think they have to use instead of how the current method includes the address.

```csharp
public abstract byte ReadByte(int deviceAddress);
public abstract void Read(int deviceAddress, Span<byte> buffer);
public abstract void WriteByte(int deviceAddress, byte data);
public abstract void Write(int deviceAddress, Span<byte> data);
```

[...]

PR #1360, opened 2020-12-10
Add I2cBus

krwq:
Fixes: https://github.com/dotnet/iot/issues/1335

- also: SenseHat now takes I2cBus as input (not required arg)
- DeviceInformation from Ft4222 is now FtDevice. Previous FtDevice is now FtDeviceType (that seemed like logical change given their purpose)
- Ft4222I2cBus is removed and now only accessible through `FtDevice.CreateI2cBus()`

cc: @pgrawehr @Ellerbach @joperezr


src/devices/Ft4222/FtDevice.cs
```diff
@@ -1,81 +1,72 @@
...
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
+using System.Device.I2c;
+
 namespace Iot.Device.Ft4222
 {
     /// <summary>
-    /// List of FTDI device types
+    /// FT4222 device information
```
krwq, line 10: This file got moved to FtDeviceType and was replaced with DeviceInformation

src/devices/SenseHat/SenseHat.cs
```diff
@@ -133,23 +147,32 @@ public SenseHat()
         /// <inheritdoc/>
         public void Dispose()
         {
-            _ledMatrix?.Dispose();
-            _ledMatrix = null!;
+            if (_shouldDispose)
+            {
+                _i2cBus?.Dispose();
```
krwq, line 152: note we don't need to dispose of devices separately here since this will be handled by bus

krwq:
Seems there is something wrong with BME280, Since I added second test testing the same thing on both OS-es we now get:

```
    System.Device.Gpio.Tests.ProtocolsTests.I2C_Bme280CanRead [FAIL]
      Assert.InRange() Failure
      Range:  (892 - 1084)
      Actual: 657.193881275765
      Stack Trace:
        /_/src/System.Device.Gpio.Tests/ProtocolTests.cs(83,0): at System.Device.Gpio.Tests.ProtocolsTests.TestBme280Reading(Bme280 bme280)
        /_/src/System.Device.Gpio.Tests/ProtocolTests.cs(56,0): at System.Device.Gpio.Tests.ProtocolsTests.I2C_Bme280CanRead()
```

src/System.Device.Gpio/System.Device.Gpio.csproj
```diff
@@ -43,6 +43,7 @@
     <!--Remove the Linux-specific code-->
     <Compile Remove="System\Device\Gpio\Drivers\Windows10Driver.notSupported.cs" />
     <Compile Remove="System\Device\I2c\I2cDevice.nonWindows.cs" />
+    <Compile Remove="System\Device\I2c\I2cBus.nonWindows.cs" />
```
joperezr, line 46: We should consider also removing UnixI2cBus, UnixI2cDevice and UnicI2cFileTransferBus from the Windows-specific build

joperezr: No need to do that on this PR though, we can do that later.

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// </summary>
+        /// <param name="deviceAddress">Device address to create</param>
+        public abstract void RemoveDevice(int deviceAddress);
+
+        /// <summary>
+        /// Reads data from the specified device.
+        /// </summary>
+        /// <param name="deviceAddress">Device address to read from.</param>
+        /// <param name="buffer">Buffer to read data to.</param>
+        public abstract void Read(int deviceAddress, Span<byte> buffer);
```
joperezr, line 46: Do you mind explaining why you need Read, Write and WriteRead methods on the Bus? I had envisioned the Bus being more of a I2cDevice manager, that you would use to create and remove devices, and then use the devices you got back to actually perform the read, write and writeRead operations. WIth this now you have read, write and writeRead both on the bus and on the device.

joperezr: Basically my comment here is, the way this PR is today, about 80% of the API in i2cBus is also present on I2cDevice, so it would be good to separate a bit more the responsabilities of  both. Today, I2cDevice class can Create i2c devices, Read, Write and Dispose them. I2cBus with your changes can do exactly that as well, plus  manage i2c Devices (Remove devices). Why not make I2cBus all about I2cDevice management (possibly even consider deprecating I2cDevice.Create so folks try to use I2cBus.CreateDevice instead) and leave the read/write operations on the device?

krwq: i.e.: I2C scanner, sending reset to I2C Bus, devices using multiple I2C addresses (see the issue in the description)

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
```diff
@@ -0,0 +1,83 @@
...
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Disposes this instance
+        /// </summary>
+        /// <param name="disposing"><see langword="true"/> if explicitly disposing, <see langword="false"/> if in finalizer</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            // Nothing to do in the base class.
```
joperezr, line 80: Won't we want to dispose the devices that where created with this?

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
joperezr: GpioController is different as we don't have a GpioPin yet, so this is not an issue there. My concern is that  we will now have 2 classes that are more than 80% the same. It would be better to separate responsabilities. From the issue seems like having read/write/readwrite on bus is just  a convinience thing, I think we should prioritize having separate concepts for each class over convinience.

src/System.Device.Gpio/System/Device/I2c/I2cDevice.cs
```diff
@@ -70,7 +70,7 @@ public static I2cDevice Create(I2cConnectionSettings settings)
             }
             else
             {
-                return new UnixI2cDevice(settings);
+                return new UnixI2cDevice(UnixI2cBus.Create(settings.BusId), settings.DeviceAddress, shouldDisposeBus: true);
```
joperezr, line 73: Perhaps we should just deprecate this API in favor of the bus managing the creation of devices, what do you think?

src/System.Device.Gpio/System/Device/I2c/UnixI2cBus.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+
+            if (buffer.Length > ushort.MaxValue)
+            {
+                throw new ArgumentException($"{nameof(buffer)} length is too long.", nameof(buffer));
+            }
+
+            fixed (byte* readBufferPointer = buffer)
+            {
+                WriteReadCore((ushort)deviceAddress, null, readBufferPointer, 0, (ushort)buffer.Length);
```
joperezr, line 107: >WriteReadCore [](start = 16, length = 13)

This is another problem of having the read/write/readwrite methods here. With this we now have two almost identical implementations of all of these methods, one here and one on UnixI2cDevice so code duplication will just be prone to issues when fixing bugs as we may fix it in one place but not the other. I would strongly recommend to just remove all these operations from here and leave them on I2cDevice, or worst case scenario keep a list of I2cDevices opened on the bus class and simply calling i2cDevice.Read from here.

krwq: UnixI2cDevice doesn't duplicate this code

joperezr: oh nvm, now I see that you also changed the I2cDevice to now use the bus to perform the operations, so at least we won't have code duplication. I still have the concern of having to classes doing almost the same in terms of API surface.

src/System.Device.Gpio/System/Device/I2c/I2cDevice.cs
krwq: No, I think you should use I2cDevice.Create when you have one device and I2cBus when you have multiple I2C devices

joperezr: Basically we should ask ourselves this question: What is the point of having I2cDevice class after we introduce I2cBus the way you have it on this PR? The only benefit I see is so that you don't have to remember the I2cAddress, which to me makes very little of an argument to keep it. I'm not against having I2cBus having all t hese methods, but if that's the case then perhaps we should just deprecate the whole I2cDevice if we don't see any benefit of it other than remembering which address it's plugged into.

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
krwq: this doesn't create any devices, only the implementations do

src/System.Device.Gpio/System.Device.Gpio.csproj
pgrawehr: Nope. `!= 'net5.0'` here actually means "net5.0-windows*". And this _should_ include the linux stuff, otherwise crosscompilling is no longer possible.

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
pgrawehr: I thought the same, but I can see the use cases @krwq points out. Could we at least have the read/write/writeread methods as extension methods somewhere else (doesn't even need to be in System.Devices)?

src/System.Device.Gpio/System/Device/I2c/Devices/Windows10I2cBus.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                throw new ArgumentException($"Device with address 0x{deviceAddress,0X2} was not open.", nameof(deviceAddress));
+            }
+
+            RemoveDeviceNoCheck(deviceAddress, device);
+        }
+
+        internal void RemoveDeviceNoCheck(int deviceAddress, Windows10I2cBusDevice device)
+        {
+            _devices?.Remove(deviceAddress);
```
pgrawehr, line 51: This looks quite risky. If the device is removing itself from the bus on Dispose (which it probably should), this is calling for an infinite recursion.

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// </summary>
+        /// <param name="deviceAddress">Device address related with the device to create.</param>
+        /// <returns>I2cDevice instance.</returns>
+        public abstract I2cDevice CreateDevice(int deviceAddress);
+
+        /// <summary>
+        /// Removes I2C device.
+        /// </summary>
+        /// <param name="deviceAddress">Device address to create</param>
+        public abstract void RemoveDevice(int deviceAddress);
```
pgrawehr, line 39: This should probably have an overload taking an I2cDevice.

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
krwq: The extension methods won't be a good idea. I.e. in case of Ft4222 or Unix implementation they only have a concept of a bus and I2cDevice is not present at all, in case of Windows implementation they don't have a concept of a bus. In case of Windows 10 extension method would need to create a device just for the sake of doing this operation while in the case of Unix/Ft4222 you create a device for the sake of having device. I do not feel strongly about having these on the bus but it feels to me like depending on the device either one or the other one will make more sense

src/System.Device.Gpio/System/Device/I2c/I2cDevice.cs
krwq: I kind of think of this the same as XML APIs, they have different purpose even though you can achieve the same things with them: there is XmlDocument which is optimized for DOM manipulation, XDocument optimized for quick extraction of the data and XPathDocument optimized for making complex queries. Here I2cDevice is for manipulating on a single address, I2cBus is giving you way of creating devices and operating on the spectrum. Once we remove Read/Write/ReadWrite methods there should be easier distinction but I still think there is a value in having it in both places. Also see discussion in here: https://github.com/dotnet/iot/issues/214 the device mention in there uses different I2C address depending on the size of the memory.

src/System.Device.Gpio/System/Device/I2c/UnixI2cBus.cs
krwq: I2cDevice is a convenience and mostly recommended one. I2cBus should only be used when you get very specialized APIs. I.e.:
- SenseHat which has 6 devices and chaining it with Ft4222 is a pure pain without I2cBus
- Ft4222 only has a concept of a bus which means it needs an object to represent it, while you could create I2cBus just for Ft4222 it just makes it easier to use in context like SenseHat
- https://github.com/dotnet/iot/issues/214 uses different address depending on memory size
- I2C scanning

for case where you use a single address you should only use I2cDevice. I2cBus overload could be provided for convenience so that address doesn't have to be passed

src/System.Device.Gpio/System/Device/I2c/Devices/Windows10I2cBus.cs
krwq: Note that this is internal and the only reason this method exists is to prevent infinite recursion.

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
krwq: I think we should do one or the other, not both. In the issue I've proposed both solutions and address made more sense as it was similar to GPIO APIs

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
pgrawehr: I have the impression we got a clash of partially contradicting concepts (or requirements) here. What you are implementing here is a `I2cBus_Driver_`, while what I was more expecting is a `I2cBus_Manager_`, that would be similar in use to the GpioController. 

Your requirements/implementation:
- Implement a hardware-dependent component that is used to talk to devices, actually implementing the read/write operations.
- Create either a bus or a device to forward to a binding, whichever is more convenient

My requirements/expectations:
- The I2cBus manages devices, but is not hardware-dependent, that part is left to the devices
- Typically, I2cBus is a singleton instance. Only one instance per physical bus exists
- Supports per-bus operations, such as scanning or resetting, using the driver for this purpose
- I2cBus shall be used to create devices

I guess we'll have to find consensus on which path to go, maybe what we really need is an `I2cBus` (as management instance) and an `I2cBusDriver` (as mostly what this PR proposes). I would not mix the two.

src/System.Device.Gpio/System/Device/I2c/Devices/Windows10I2cBus.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            Windows10I2cBusDevice device = new Windows10I2cBusDevice(this, new I2cConnectionSettings(BusId, deviceAddress));
+            _devices[deviceAddress] = device;
+
+            return device;
+        }
+
+        public override void Read(int deviceAddress, Span<byte> buffer)
+        {
+            using (Windows10I2cDevice i2cDevice = new Windows10I2cDevice(new I2cConnectionSettings(BusId, deviceAddress)))
```
Ellerbach, line 33: So in terms of performance, I guess this will be quite slow. And the preferred way will be to get the device. Especially if you are intensively trying to read/write using the I2cBus rather than the I2cDevice
You should definitely mention this in the documentation at least for the Windows part.

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
Ellerbach: I agree with @krwq and prefer the device address as it's what's used and how the I2C is managed all up. This makes for me the moist sense.

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
Ellerbach: Another element in favor of the read and write on the I2cBus is support for SMBus which is very close to I2C with alerting and few other elements.
I agree that it can be a bit confusing having both read and write on both the I2cDevice and the I2cBus but it's a matter of scenario. And so far, we only allow devices in bindings having I2cDevice and we should continue like this except for specific scenarios like sense hat where having a bus will make sense but is not necessary needed.
Documentation and guidelines and PR review should be strict about this.

src/System.Device.Gpio/System/Device/I2c/I2cDevice.cs
Ellerbach: I2cDevice makes it easy when you have only 1 device. Similar somehow with GpioPin vs GpioController. What's the purpose of having a controller when you need only 1 pin? Same for me on the I2cDevice. 
Bindings so far are using only I2cDevice and it's good. We should continue to encourage this and continue to push for this except where there are multiple devices like sense hat, where reading and writing on the bus to initialize, send alerts (SMBus scenarios) is needed.

src/System.Device.Gpio/System/Device/I2c/UnixI2cBus.cs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        internal bool RemoveDeviceNoCheck(int deviceAddress)
+        {
+            return _usedAddresses?.Remove(deviceAddress) ?? false;
+        }
+
+        public override unsafe void Read(int deviceAddress, Span<byte> buffer)
+        {
+            if (deviceAddress < 0 || deviceAddress > ushort.MaxValue)
```
Ellerbach, line 90: for I2C the max address is 2014 or 128 but support for 128+ address space is not supported (yet) in .NET IoT. See https://www.i2c-bus.org/addressing/. And for 10 bits addresses here: https://www.i2c-bus.org/addressing/10-bit-addressing/ and that's something that can be done on a higher level.
So max value checked should be either 128 or 1024 or just 256 to be safe and allow the 1024 scenario to happen on a higher level.

src/System.Device.Gpio/System/Device/I2c/UnixI2cBus.cs
Ellerbach: Add to this: 1024 addressing support, SMBus specific support (some will still be missing but getting there), broadcast to the bus, hardware I2C programming, more information here: https://www.i2c-bus.org/addressing/

Ellerbach (approved):
Looks good. Impatient to play with it!

src/System.Device.Gpio/System/Device/I2c/I2cBus.cs
pgrawehr: I'm not opposing the creation of I2cBus, but I think the proposed approach needs a conceptual review. I believe this is mostly a move of functions from one class to another, but does not provide the abstraction we really want. As I noted above, I think the requirements are not clear.

Ellerbach: @pgrawehr what would be then the classes abstract for what you describe (even if not complete) to have a better understanding of the relations between all of them?

pgrawehr: @Ellerbach : I hope I'll be able to provide a schema of what I mean by tomorrow.

pgrawehr: I2cBusPR
I2cBusWithDriver

The first schema shows how things look with this PR, the second schema is somewhat what I would prefer. For further extension there should be a separation between adding features that affect devices or busses in general (such as bus scanning, resetting, ref-counting devices, writing exactly 7 bytes or whatever) and the actual I2C low-level transport implementation. 

Rationale: Extension will be different for the two cases, and often you wouldn't want to do both at the same time. I.e. for the board PR, I'll be adding pin and (eventually) device managing capabilities, which I could add by inheriting from I2cBus and/or I2cDevice. For adding a bus scanning function, one could inherit from I2cBus, or even write it as an extension function. However, for the Arduino PR, I'll be writing an I2c Driver, but leave the bus and device classes probably untouched.

Ellerbach: So if I summarize correctly, you propose to:

- Add another layer I2cBusDriver
- Create either the I2cBus either I2cDevice from the I2cBusDriver

So if I understand correctly that will imply:

- To have the I2cBusDriver exposed => rather than have I2cBus and I2cDevice exposed only, we will have then the 3. I guess that would bring even more confusions

So I would prefer to go for I2cBus and I2cDevice exposed only and hide some of the plumbing behind. Some systems have a notion of bus (Linux and in general all systems like FT4222, Arduino, etc), only exception is Windows (so far). So the I2cBus seems more suitable to create instances of I2cDevice than having an intermediate driver.

I may have miss something in your proposal @pgrawehr

pgrawehr: @Ellerbach Yes, that sums up my proposal quite well. I think however, that the additional publicly visible interface would cause very little harm, because most users won't have to deal with it, similar to the existing GpioDriver abstract base. However, it would make deriving instances for the board, FT4222, Arduino etc, much more clean. We should talk this over in tomorrow's call.

src/System.Device.Gpio/System/Device/I2c/UnixI2cBus.cs
krwq: I did ushort.MaxValue because that's the max we can pass to the native code. I prefer on letting native code handle the range part as there may be some more scenarios we're missing here and native can handle.

krwq:
I have removed I2cBus.Read/Write/WriteRead per offline conversation with @pgrawehr and @joperezr. I'll create a separate issue to add it as we couldn't get to the agreement if it's useful but we agree on the remainder of the APIs.

src/System.Device.Gpio/System/Device/I2c/UnixI2cBus.cs
Ellerbach: OK, makes sense then.

krwq:
I'm expecting there will be merge conflict with new changes (MFRC522) and CI will fail. This will likely need rebase

krwq:
I've rebased and added a fix

krwq:
@pgrawehr @joperezr please review

pgrawehr (approved):
This looks good to me now.

PR #1361, opened 2020-12-11
A GPIO has to be writeable even in drive mode input

mstroppel:
...because one wants to set the value used when switching to output mode.

Detailed explanation:

CPU's GPIO pins are set to input mode upon system start. Via pull-up or pull-down resistor the state can be determined on schematic/electronic side.
If the pin is used as an output, the software can set the mode to output. But before it has to be possible to set the output state to ensure that the output state is where it should be upon switching from input to output state.

A simplified schematic of GPIO internals:

```text
                           Vcc
                          /
                   ------X2 high / low switch
                  /       \
                 /         GND
                /
GPIO pin ------ X1 input / output switch or output activate switch (input will still be there)
                \
                 ------- Input logic
```

So before switching X1 we should define in which state X2 is.

I had the case with an Apollo Lake CPU where:

```csharp
_gpioController.SetDriveMode(1, PinMode.Output)
_gpioController.Write(1, PinValue.Low)
```
Lead to a short high peak in the output that I have to avoid. That's why I had to switch to use `Windows.Devices.Gpio` via the NuGet `Microsoft.Windows.SDK.Contracts`, which allows to write before setting to output.

### DIFF
--- a/src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
@@ -46,7 +46,7 @@ namespace System.Device.Gpio.Tests
         }
 
         [Fact]
-        public void WriteInputPinThrows()
+        public void WriteInputPinDoesNotThrow()
         {
             _mockedGpioDriver.Setup(x => x.OpenPinEx(1));
             _mockedGpioDriver.Setup(x => x.IsPinModeSupportedEx(1, It.IsAny<PinMode>())).Returns(true);
@@ -55,7 +55,7 @@ namespace System.Device.Gpio.Tests
             var ctrl = new GpioController(PinNumberingScheme.Logical, _mockedGpioDriver.Object);
 
             ctrl.OpenPin(1, PinMode.Input);
-            Assert.Throws<InvalidOperationException>(() => ctrl.Write(1, PinValue.High));
+            ctrl.Write(1, PinValue.High);
         }
 
         [Fact]
--- a/src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
@@ -51,7 +51,6 @@ namespace System.Device.Gpio.Tests
             _mockedGpioDriver.Setup(x => x.OpenPinEx(1));
             _mockedGpioDriver.Setup(x => x.IsPinModeSupportedEx(1, It.IsAny<PinMode>())).Returns(true);
             _mockedGpioDriver.Setup(x => x.SetPinModeEx(1, It.IsAny<PinMode>()));
-            _mockedGpioDriver.Setup(x => x.GetPinModeEx(1)).Returns(PinMode.Input);
             var ctrl = new GpioController(PinNumberingScheme.Logical, _mockedGpioDriver.Object);
 
             ctrl.OpenPin(1, PinMode.Input);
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1361) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1361) before we can accept your contribution.<br/><br/>:x: mstroppel [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1361)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1361) it.</sub>

pgrawehr:
I've seen the same behavior on other hardware as well, so I think the proposed change is a good idea.

mstroppel:
Guys, as far as I can see, the failing build is not my fault. Did someone cut the connection between the input and output pin?

krwq:
Seems like SysFs implementation doesn't like this new capability:

```
    System.Device.Gpio.Tests.SysFsDriverTests.AllowWritingOnInputPin [FAIL]
      System.UnauthorizedAccessException : Reading a pin value requires root permissions.
      ---- System.UnauthorizedAccessException : Access to the path '/sys/class/gpio/gpio6/value' is denied.
      -------- System.IO.IOException : Operation not permitted
```

ddvsorpil022 might have some hardware issues though

pgrawehr:
If sysfs really doesn't like this feature, we could also just move the exception to the driver. I think we shouldn't prevent this for all drivers if only one of them fails to support it, since sysfs is mostly deprecated, anyway.

krwq:
@pgrawehr seems they now recommend to use chardev devices: https://www.kernel.org/doc/Documentation/gpio/consumer.txt instead

krwq:
@joperezr @Ellerbach are you ok with SysFs throwing NotSupportedException while other GPIO drivers would allow for this scenario?

pgrawehr:
Alternatively, we could also have sysfs emulate this (by storing the desired value). It wouldn't fix the original issue when using sysfs, but the behavior would be otherwise the same.

krwq:
or change mode, write value, and change mode back (not sure if that would work though)

microhobby:
Well, this will not have any side effects in `LibGpioDriver` but otherwise, it will be useless. Once configured the pin handle is locked. If there is an attempt to change the `PinMode` we will have an `EBUSY` exception. The only way to switch to another `PinMode` is to call `ClosePin` first, which will lose the previously written status anyway.

For `LibGpioDriver` we must add an API to call the `int gpiod_line_set_direction_output(struct gpiod_line *line, int value)` that enables to drive the pin to output with a default value.

Or maybe add the:

```csharp
OpenPin(int pinNumber, PinMode mode, PinValue initialValue);
SetPinMode(int pinNumber, PinMode mode, PinValue initialValue);
```

pgrawehr:
@microhobby Are you saying that with the libgpiodriver, it is already not possible to use SetPinMode to change from input to output or vice-versa? So the existing implementation would throw an exception when trying to do so?

microhobby:
> @microhobby Are you saying that with the libgpiodriver, it is already not possible to use SetPinMode to change from input to output or vice-versa? So the existing implementation would throw an exception when trying to do so?

@pgrawehr yes, I wrote an attempt to evaluate it and:

9zS6ZOJH8L

pgrawehr:
Bad... I suppose we should fix that with a workaround. There are some sensors (i.e. DHT11) that require quick switching of single lines from input to output and back.

Ellerbach:
@krwq and @pgrawehr I would prefer to implement something even slower on SysFs than raising a not implemented exception. The reason is just because if someone will want to use this and doesn't really have the knowledge, they would not understand the issue. Now, I understand it's an advance scenario, still.

joperezr:
[Triage]: Closing this PR as it is now superseded by https://github.com/dotnet/iot/pull/1473. Commits from here have been cherry-picked.

PR #1367, opened 2020-12-12
Adding Tsl256x

Ellerbach:
- Adding TSL2560 and TSL2561
- Full support of all features for this sensor
- Documentation
- Full example including interruption sample

Documentation: https://cdn-shop.adafruit.com/datasheets/TSL2561.pdf

### DIFF
--- a/src/devices/Tsl256x/PackageType.cs
+++ b/src/devices/Tsl256x/PackageType.cs
@@ -1,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace IoT.Device.Tsl256x
+{
+    /// <summary>
+    /// The type of package for TSL256x. This is used for the Lux calculation
+    /// </summary>
+    public enum PackageType
+    {
+        /// <summary>
+        /// PAckage type CS
+        /// </summary>
+        PackageCs = 0,
+
+        /// <summary>
+        /// Package type CL, T and FN
+        /// </summary>
+        Other,
+    }
+}
--- a/src/devices/Tsl256x/PackageType.cs
+++ b/src/devices/Tsl256x/PackageType.cs
@@ -9,7 +9,7 @@ namespace IoT.Device.Tsl256x
     public enum PackageType
     {
         /// <summary>
-        /// PAckage type CS
+        /// Package type CS
         /// </summary>
         PackageCs = 0,
 
--- a/src/devices/Tsl256x/README.md
+++ b/src/devices/Tsl256x/README.md
@@ -27,8 +27,8 @@ Console.WriteLine($"Illuminance is {lux.Lux} Lux");
 
 - Be aware, there are 2 types of packaging the CS and the Others T, FN and CL. Refer to the documentation to understand which package you have on the board. This is an argument when creating the sensor as the calculation for the illuminance is different.
 - There are 3 different possible I2C addresses for this device depending on how the address pin is setup:
-  - DefaultI2cAddress = 0x29: When address pin is to the ground
-  - SecondI2cAddress = 0x39: When address pin is floating
+  - DefaultI2cAddress = 0x39: When address pin is to the ground
+  - SecondI2cAddress = 0x29: When address pin is floating
   - ThirdI2cAddress = 0x49: When address pin is to VDD
 
 ## Check the version
--- a/src/devices/Tsl256x/README.md
+++ b/src/devices/Tsl256x/README.md
@@ -37,7 +37,7 @@ You can determine if you have a TSL2560 or TSL2561 version:
 
 ```csharp
 var ver = tsl256X.Version;
-string msg = ver.Major == 1 ? $"This is a TSL2561, revision {ver.Minor}" : $"This is a TSL2560, revision {ver.Minor}";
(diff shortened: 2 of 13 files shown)


src/devices/Tsl256x/PackageType.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace IoT.Device.Tsl256x
+{
+    /// <summary>
+    /// The type of package for TSL256x. This is used for the Lux calculation
+    /// </summary>
+    public enum PackageType
+    {
+        /// <summary>
+        /// PAckage type CS
```
krwq, line 12:
```suggestion
        /// Package type CS
```

src/devices/Tsl256x/README.md
```diff
@@ -0,0 +1,110 @@
...
+Tsl256x tsl256X = new(i2cDevice, PackageType.Other);
+tsl256X.IntegrationTime = IntegrationTime.Integration402Milliseconds;
+tsl256X.Gain = Gain.Normal;
+var lux = tsl256X.MeasureAndGetIlluminance();
+Console.WriteLine($"Illuminance is {lux.Lux} Lux");
+```
+
+**notes**:
+
+- Be aware, there are 2 types of packaging the CS and the Others T, FN and CL. Refer to the documentation to understand which package you hav on the board. This is an argument when creating the sensor as the calculation for the illuminance is different.
```
krwq, line 28:
```suggestion
- Be aware, there are 2 types of packaging the CS and the Others T, FN and CL. Refer to the documentation to understand which package you have on the board. This is an argument when creating the sensor as the calculation for the illuminance is different.
```

src/devices/Tsl256x/README.md
```diff
@@ -0,0 +1,110 @@
...
+- OutOfRangeXXIntegrationTimePeriods: where XX is from 2 to 15, will interrupt only if the value remains outside more than XX times
+
+The following example shows how to setup a basic interruption using test mode:
+
+```csharp
+Console.WriteLine("Set interruption to test. Read the interrupt pin");
+GpioController controller = new();
+controller.OpenPin(PinInterrupt, PinMode.Input);
+tsl256X.InterruptControl = InterruptControl.TestMode;
+tsl256X.Power = true;
```
krwq, line 66: I'm confused what Power does? Could you rename it to something less confusing? (i.e. Disabled if it means to turn on/off device)

src/devices/Tsl256x/README.md
```diff
@@ -0,0 +1,110 @@
...
+while (controller.Read(PinInterrupt) == PinValue.High)
+{
+    Thread.Sleep(1);
+}
+
+Console.WriteLine($"Interrupt detected, read the value to clear the interrupt");
+tsl256X.GetRawChannels(out ushort ch0, out ushort ch1);
```
krwq, line 73: can you add sample how to convert that to illuminance or how to get it directly?

src/devices/Tsl256x/Tsl256x.cs
```diff
@@ -0,0 +1,353 @@
...
+        private IntegrationTime _integrationTime;
+        private Gain _gain;
+        private InterruptControl _interruptControl;
+        private InterruptPersistence _interruptPersistence;
+        private PackageType _packageType;
+
+        /// <summary>
+        /// When the address select pin is to ground
+        /// </summary>
+        public const int SecondI2cAddress = 0x29;
```
krwq, line 29: I think you should swap the order so that Default is first then second, then third 😄

src/devices/Tsl256x/Tsl256x.cs
```diff
@@ -0,0 +1,353 @@
...
+
+            scale = _gain == Gain.Normal ? scale : scale * 16;
+
+            double channel0 = (ch0 * scale) / ChannelScale;
+            double channel1 = (ch1 * scale) / ChannelScale;
+
+            double lux = 0;
+            if (_packageType == PackageType.PackageCs)
+            {
+                if (ratio <= 0.52)
```
krwq, line 274: can you point to spec about these calculations?

krwq (approved):
Couple of nits

src/devices/Tsl256x/README.md
Ellerbach: Yes, will rename. Wanted to keep close to datasheet but Enabled will be much better

src/devices/Tsl256x/README.md
Ellerbach: There is a function for that: `MeasureAndGetIlluminance`. This is just reading the raw data.

src/devices/Tsl256x/Tsl256x.cs
Ellerbach: The default is the 0x39, the second and third has no importance :-)

src/devices/Tsl256x/Tsl256x.cs
Ellerbach: It's done line 233. : `See documentation page 23 and following for the constant and the calculation`

src/devices/Tsl256x/Tsl256x.cs
krwq: I meant the field order :)

Ellerbach: :-)))) I can do that, yes :-)

krwq (approved):

PR #1375, opened 2020-12-20
added mlx90640 📷🕯 to devices library

garciaolais:

### DIFF
--- a/src/devices/Mlx90640/Mlx90640.cs
+++ b/src/devices/Mlx90640/Mlx90640.cs
@@ -687,7 +687,7 @@ namespace Iot.Device.Mlx90640
 
                     ktaTemp[p] = ktaTemp[p] * (1 << ktaScale2);
                     ktaTemp[p] = ktaRC[split] + ktaTemp[p];
-                    ktaTemp[p] = (float)(ktaTemp[p] / Math.Pow(2, ktaScale1));                    
+                    ktaTemp[p] = (float)(ktaTemp[p] / Math.Pow(2, ktaScale1));
                 }
             }
 
--- a/src/devices/Mlx90640/Mlx90640.cs
+++ b/src/devices/Mlx90640/Mlx90640.cs
@@ -1,0 +1,1072 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Ported from https://github.com/adafruit/Adafruit_CircuitPython_MLX90640
+// Formulas and code examples can also be found in the datasheet https://github.com/melexis/mlx90640-library/blob/master/MLX90640%20driver.pdf
+using System;
+using System.Buffers.Binary;
+using System.Device.I2c;
+using UnitsNet;
+
+namespace Iot.Device.Mlx90640
+{
+    /// <summary>
+    /// Infrared Camera MLX90640
+    /// </summary>
+    public sealed class Mlx90640 : IDisposable
+    {
+        private const int OPENAIR_TA_SHIFT = 8;
+        private const double SCALEALPHA = 0.000001;
+
+        private I2cDevice _i2cDevice;
+        private ushort[] _eeData = new ushort[832];
+
+        private int[] _alpha = new int[768];
+        private double _alphaPTAT = 0;
+        private byte _alphaScale = 0;
+        private ushort[] _brokenPixels = new ushort[5];
+        private int _calibrationModeEE = 0;
+        private float[] _cpAlpha = new float[2];
+        private float _cpKta = 0;
+        private float _cpKv = 0;
+        private int[] _cpOffset = new int[2];
+        private int[] _ct = new int[5];
+        private short _gainEE = 0;
+        private float[] _ilChessC = new float[3];
+        private float _ksTa = 0;
+        private double[] _ksTo = new double[5];
+        private byte[] _kta = new byte[768];
+        private byte _ktaScale = 0;
+        private float _ktPTAT = 0;
+        private short _kVdd = 0;
+        private sbyte[] _kv = new sbyte[768];
+        private float _kvPTAT = 0;
+        private byte _kvScale = 0;
+        private int[] _offset = new int[768];
+        private ushort[] _outlierPixels = new ushort[5];
+        private byte _resolutionEE = 0;
+        private float _tgc = 0;
+        private short _vdd25 = 0;
+        private short _vPTAT25 = 0;
+
+        /// <summary>
+        /// MLX90640 Default I2C Address
+        /// </summary>
+        public const byte DefaultI2cAddress = 0x33;
+
+        /// <summary>
+        /// Creates a new instance of the MLX90640
+        /// </summary>
+        /// <param name="i2cDevice">The I2C device used for communication.</param>
+        public Mlx90640(I2cDevice i2cDevice)
+        {
+            _i2cDevice = i2cDevice ?? throw new ArgumentNullException(nameof(i2cDevice));
+            DumpEE();
+            ExtractVDDParameters();
+            ExtractPTATParameters();
+            ExtractVDDParameters();
+            ExtractPTATParameters();
+            ExtractGainParameters();
+            ExtractTgcParameters();
+            ExtractResolutionParameters();
+            ExtractKsTaParameters();
+            ExtractKsToParameters();
+            ExtractCPParameters();
+            ExtractAlphaParameters();
+            ExtractOffsetParameters();
+            ExtractKtaPixelParameters();
+            ExtractKvPixelParameters();
+            ExtractCILCParameters();
+            ExtractDeviatingPixels();
+        }
+
+        /// <summary>
+        /// Sets sampling to the given value
+        /// </summary>
+        /// <param name="mode">Sampling Mode</param>
+        public void SetSampling(Sampling mode)
+        {
+            ushort[] controlRegister1 = new ushort[1];
+            int value;
+
+            value = ((byte)mode & 0x07) << 7;
+
+            controlRegister1 = ReadRegisters16(0x800D, 1);
+            value = (controlRegister1[0] & 0xFC7F) | value;
+            Write16(0x800D, value);
+        }
+
+        /// <summary>
+        /// Request both 'halves' of a frame from the sensor, merge them
+        /// and calculate the temperature in C for each of 32x24 pixels.
+        /// </summary>
+        public float[] GetFrame()
+        {
+            float[] framebuf = new float[768];
+            float emissivity = 0.95f;
+            ushort[] mlx90640Frame = new ushort[834];
+
+            for (int i = 0; i < 2; i++)
+            {
+                if (GetFrameData(ref mlx90640Frame) < 0)
+                {
+                    throw new Exception("Frame data error");
+                }
+
+                float tr = GetTa(mlx90640Frame) - OPENAIR_TA_SHIFT;
+                CalculateTo(mlx90640Frame, emissivity, tr, ref framebuf);
+            }
+
+            return framebuf;
+        }
+
+        private void CalculateTo(ushort[] frameData, float emissivity, float tr, ref float[] result)
+        {
+            float vdd;
+            float ta;
+            float ta4;
+            float tr4;
+            float taTr;
+            float gain;
+            float[] irDataCP = new float[2];
+            float irData;
+            float alphaCompensated;
+            byte mode;
+            sbyte ilPattern;
+            byte chessPattern;
+            byte pattern;
+            byte conversionPattern;
+            float sx;
+            float to;
+            float[] alphaCorrR = new float[4];
+            byte range;
+            ushort subPage;
+            float ktaScale;
+            float kvScale;
+            float alphaScale;
+            float kta;
+            float kv;
+
+            subPage = frameData[833];
+            vdd = GetVdd(frameData);
+            ta = GetTa(frameData);
+
+            ta4 = (float)(ta + 273.15);
+            ta4 = ta4 * ta4;
+            ta4 = ta4 * ta4;
+            tr4 = (float)(tr + 273.15);
+            tr4 = tr4 * tr4;
+            tr4 = tr4 * tr4;
+            taTr = tr4 - (tr4 - ta4) / emissivity;
+
+            ktaScale = (float)(Math.Pow(2, _ktaScale));
+            kvScale = (float)(Math.Pow(2, _kvScale));
+            alphaScale = (float)(Math.Pow(2, _alphaScale));
+
+            alphaCorrR[0] = (float)(1 / (1 + _ksTo[0] * 40));
+            alphaCorrR[1] = 1;
+            alphaCorrR[2] = (float)((1 + _ksTo[1] * _ct[2]));
+            alphaCorrR[3] = (float)(alphaCorrR[2] * (1 + _ksTo[2] * (_ct[3] - _ct[2])));
+
+            //------------------------- Gain calculation -----------------------------------
+            gain = frameData[778];
+            if (gain > 32767)
+            {
+                gain = gain - 65536;
+            }
+
+            gain = (float)(_gainEE / gain);
+
+            //------------------------- To calculation -------------------------------------
+            mode = (byte)((frameData[832] & 0x1000) >> 5);
+
+            irDataCP[0] = frameData[776];
+            irDataCP[1] = frameData[808];
+            for (int i = 0; i < 2; i++)
+            {
+                if (irDataCP[i] > 32767)
+                {
+                    irDataCP[i] = irDataCP[i] - 65536;
+                }
+
+                irDataCP[i] = irDataCP[i] * gain;
+            }
+
+            irDataCP[0] = (float)(irDataCP[0] - _cpOffset[0] * (1 + _cpKta * (ta - 25)) * (1 + _cpKv * (vdd - 3.3)));
+            if (mode == _calibrationModeEE)
+            {
+                irDataCP[1] = (float)(irDataCP[1] - _cpOffset[1] * (1 + _cpKta * (ta - 25)) * (1 + _cpKv * (vdd - 3.3)));
+            }
+            else
+            {
+                irDataCP[1] = (float)(irDataCP[1] - (_cpOffset[1] + _ilChessC[0]) * (1 + _cpKta * (ta - 25)) * (1 + _cpKv * (vdd - 3.3)));
+            }
+
+            for (int pixelNumber = 0; pixelNumber < 768; pixelNumber++)
+            {
+                ilPattern = (sbyte)(pixelNumber / 32 - (pixelNumber / 64) * 2);
+                chessPattern = (byte)(ilPattern ^ (pixelNumber - (pixelNumber / 2) * 2));
+                conversionPattern = (byte)(((pixelNumber + 2) / 4 - (pixelNumber + 3) / 4 + (pixelNumber + 1) / 4 - pixelNumber / 4) * (1 - 2 * ilPattern));
+
+                if (mode == 0)
+                {
+                    pattern = (byte)ilPattern;
+                }
+                else
+                {
+                    pattern = chessPattern;
+                }
+
+                if (pattern == frameData[833])
+                {
+                    irData = frameData[pixelNumber];
+                    if (irData > 32767)
+                    {
+                        irData = irData - 65536;
+                    }
+
+                    irData = irData * gain;
+
+                    kta = _kta[pixelNumber] / ktaScale;
+                    kv = _kv[pixelNumber] / kvScale;
+                    irData = (float)(irData - _offset[pixelNumber] * (1 + kta * (ta - 25)) * (1 + kv * (vdd - 3.3)));
+
+                    if (mode != _calibrationModeEE)
+                    {
+                        irData = irData + _ilChessC[2] * (2 * ilPattern - 1) - _ilChessC[1] * conversionPattern;
+                    }
+
+                    irData = irData - _tgc * irDataCP[subPage];
+                    irData = irData / emissivity;
+
+                    alphaCompensated = (float)(SCALEALPHA * alphaScale / _alpha[pixelNumber]);
+                    alphaCompensated = alphaCompensated * (1 + _ksTa * (ta - 25));
+
+                    sx = alphaCompensated * alphaCompensated * alphaCompensated * (irData + alphaCompensated * taTr);
+                    sx = (float)(Math.Sqrt(Math.Sqrt(sx)) * _ksTo[1]);
+
+                    to = (float)(Math.Sqrt(Math.Sqrt(irData / (alphaCompensated * (1 - _ksTo[1] * 273.15) + sx) + taTr)) - 273.15);
+
+                    if (to < _ct[1])
+                    {
+                        range = 0;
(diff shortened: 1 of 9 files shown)


src/devices/Mlx90640/Mlx90640.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net5.0;netcoreapp2.1</TargetFrameworks>
+    <EnableDefaultItems>false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="Mlx90640.cs" />
+    <Compile Include="Sampling.cs" />
```
krwq, line 10:
```suggestion
    <Compile Include="*.cs" />
```

src/devices/Mlx90640/Mlx90640.csproj
```diff
@@ -0,0 +1,29 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="category.txt" />
```
krwq, line 14:
```suggestion
    <None Include="category.txt" />
```

krwq: you should merge it with section below

src/devices/Mlx90640/Mlx90640.csproj
```diff
@@ -0,0 +1,29 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include="README.md" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Drawing.Common" Version="5.0.0" />
```
krwq, line 22:
```suggestion
    <PackageReference Include="System.Drawing.Common" Version="$(SystemDrawingCommonPackageVersion)" />
```

src/devices/Mlx90640/Mlx90640.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <PackageReference Include="System.Drawing.Primitives" Version="4.3.0" />
```
krwq, line 23: do you need to include this directly? If so please add the version to https://github.com/dotnet/iot/blob/master/eng/Versions.props and use the variable

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+            ExtractDeviatingPixels();
+        }
+
+        /// <summary>
+        /// Sets sampling to the given value
+        /// </summary>
+        /// <param name="mode">Sampling Mode</param>
+        public void SetSampling(Sampling mode)
+        {
+            ushort[] controlRegister1 = new ushort[1];
```
krwq, line 89: use Span + stackalloc to avoid allocations

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+            controlRegister1 = ReadRegisters16(0x800D, 1);
+            value = (controlRegister1[0] & 0xFC7F) | value;
+            Write16(0x800D, value);
+        }
+
+        /// <summary>
+        /// Request both 'halves' of a frame from the sensor, merge them
+        /// and calculate the temperature in C for each of 32x24 pixels.
+        /// </summary>
+        public float[] GetFrame()
```
krwq, line 103: `Temperature[]`?

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+            ushort subPage;
+            float ktaScale;
+            float kvScale;
+            float alphaScale;
+            float kta;
+            float kv;
+
+            subPage = frameData[833];
+            vdd = GetVdd(frameData);
+            ta = GetTa(frameData);
```
krwq, line 152: if this returned Temperature instead you could use `.Kelvin` rather than adding 273.15

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+
+            ta4 = (float)(ta + 273.15);
```
krwq, line 154: would it make sense to make this:

```csharp
float ta = Math.Pow(GetTa(frameData).Kelvin, 4);
float tr = Math.Pow(tr.Kelvin, 4);
float taTr = tr - (tr - ta) / emissivity;
```

krwq: you'd need to use Temperature rather than passing floats around

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+
+                if (pattern == frameData[833])
+                {
+                    irData = frameData[pixelNumber];
+                    if (irData > 32767)
+                    {
+                        irData = irData - 65536;
+                    }
+
+                    irData = irData * gain;
```
krwq, line 228:
```suggestion
                    irData = gain * (short)frameData[pixelNumber];
```

also I suggest to use local variables closer to where you use them and do not reuse them later

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+            float vdd;
+            float ta;
+
+            vdd = GetVdd(frameData);
+
+            ptat = frameData[800];
+            if (ptat > 32767)
+            {
+                ptat = ptat - 65536;
+            }
```
krwq, line 287:
```suggestion
            float ptat = (short)frameData[800];
```

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+
+            ptatArt = frameData[768];
+            if (ptatArt > 32767)
+            {
+                ptatArt = ptatArt - 65536;
+            }
```
krwq, line 293:
```suggestion
            float ptatArt = (short)frameData[768];
```

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+
+            ptatArt = (float)((ptat / (ptat * _alphaPTAT + ptatArt)) * Math.Pow(2, 18));
+
+            ta = (float)((ptatArt / (1 + _kvPTAT * (vdd - 3.3)) - _vPTAT25));
+            ta = ta / _ktPTAT + 25;
+
+            return ta;
```
krwq, line 300:
```suggestion
            return Temperature.FromCelsius(ta);
```

assuming this is returning temperature in ceslius

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+            float vdd;
+            float resolutionCorrection;
+
+            int resolutionRAM;
+
+            vdd = (float)frameData[810];
+            if (vdd > 32767)
+            {
+                vdd = vdd - 65536;
+            }
```
krwq, line 314:
```suggestion
            float vdd = (short)frameData[810];
```

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+            resolutionRAM = (frameData[832] & 0x0C00) >> 10;
+            resolutionCorrection = (float)(Math.Pow(2, _resolutionEE) / Math.Pow(2, resolutionRAM));
+            vdd = (float)(((resolutionCorrection * vdd - _vdd25) / _kVdd) + 3.3);
+
+            return vdd;
+        }
+
+        private ushort GetFrameData(ref ushort[] frameData)
+        {
+            byte dataReady = 0;
```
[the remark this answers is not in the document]
krwq: actually you shouldn't need variable at all

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+            ushort[] statusRegister = new ushort[0];
+            byte[] data = new byte[64];
+            byte cnt = 0;
+            ushort[] framedata = new ushort[832];
+
+            while (dataReady == 0)
+            {
+                statusRegister = ReadRegisters16(0x8000, 1);
+                dataReady = (byte)(statusRegister[0] & 0x0008);
+            }
```
krwq, line 336:
```suggestion
            do
            {
                statusRegister = ReadRegisters16(0x8000, 1);
            }
            while ((byte)(statusRegister[0] & 0x0008) == 0);
```

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+                }
+
+                float tr = GetTa(mlx90640Frame) - OPENAIR_TA_SHIFT;
+                CalculateTo(mlx90640Frame, emissivity, tr, ref framebuf);
+            }
+
+            return framebuf;
+        }
+
+        private void CalculateTo(ushort[] frameData, float emissivity, float tr, ref float[] result)
```
krwq, line 123: can you expand names, i.e. `To` `Ta`

src/devices/Mlx90640/Mlx90640.csproj
krwq: do you actually use those two in the library? I only saw them in the sample

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+            _i2cDevice.Write(new[]
+            {
+                 (byte)(writeAddress >> 8),
+                 (byte)(writeAddress & 0x00FF),
+                 (byte)(data >> 8),
+                 (byte)(data & 0x00FF)
+            });
+        }
+
+        internal ushort[] ReadRegisters16(int addr, int length)
```
krwq, line 1044: private

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+
+            if (pixPosDif > 30 && pixPosDif < 34)
+            {
+                return -6;
+            }
+
+            return 0;
+        }
+
+        internal void Write16(int writeAddress, int data)
```
krwq, line 1033: private

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+                 (byte)(writeAddress >> 8),
+                 (byte)(writeAddress & 0x00FF),
+                 (byte)(data >> 8),
+                 (byte)(data & 0x00FF)
+            });
+        }
+
+        internal ushort[] ReadRegisters16(int addr, int length)
+        {
+            Span<byte> readReg = stackalloc byte[length * 2];
```
krwq, line 1046: you should not pass variable length to stackalloc

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+        {
+            int p = 0;
+            sbyte[] ktaRC = new sbyte[4];
+            sbyte ktaRoCo;
+            sbyte ktaRoCe;
+            sbyte ktaReCo;
+            sbyte ktaReCe;
+            byte ktaScale1;
+            byte ktaScale2;
+            byte split;
```
krwq, line 637: general comment: you should declare variables as close to where they're actually used instead of C-like code...

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+            float[] ktaTemp = new float[768];
+            float temp;
+
+            ktaRoCo = (sbyte)((_eeData[54] & 0xFF00) >> 8);
+            if (ktaRoCo > 127)
```
krwq, line 642: this will never be true since sbyte.MaxValue is 127

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,1072 @@
...
+            {
+                ktaRoCo = (sbyte)(ktaRoCo - 256);
+            }
+
+            ktaRC[0] = ktaRoCo;
+
+            ktaReCo = (sbyte)(_eeData[54] & 0x00FF);
+            if (ktaReCo > 127)
```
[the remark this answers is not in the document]
krwq: and below

src/devices/Mlx90640/Mlx90640.csproj
garciaolais: It's only used in the sample, I'm going to delete it in the lib

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+            ExtractKvPixelParameters();
+            ExtractCILCParameters();
+            ExtractDeviatingPixels();
+        }
+
+        /// <summary>
+        /// Sets sampling to the given value
+        /// </summary>
+        /// <param name="mode">Sampling Mode</param>
+        public void SetSampling(Sampling mode)
```
Ellerbach, line 102: Why not create a property with it?
Also, why not directly having the correct values to Sampling? 
like `Sampling_64_Hz = 0x07 << 7`
So you won't have to adjust in the code. You'll need to change the type to ushort (see the comment as well on Write16) for the enum but it doesn't really matter.

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+
+            if (pixPosDif > 30 && pixPosDif < 34)
+            {
+                return -6;
+            }
+
+            return 0;
+        }
+
+        private void Write16(int writeAddress, int data)
```
Ellerbach, line 918: as your value is a ushort, maybe just use ushort

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+        {
+            _i2cDevice.Write(new[]
+            {
+                 (byte)(writeAddress >> 8),
+                 (byte)(writeAddress & 0x00FF),
+                 (byte)(data >> 8),
```
Ellerbach, line 924: and for the conversion of data, you can use as well BinaryPrimitives

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+        /// <summary>
+        /// Request both 'halves' of a frame from the sensor, merge them
+        /// and calculate the temperature in C for each of 32x24 pixels.
+        /// </summary>
+        /// <returns>Temperature of the specified pixel</returns>
+        public Temperature[] GetFrame()
+        {
+            Temperature[] framebuf = new Temperature[PixelCount];
+            float emissivity = 0.95f;
+            ushort[] mlx90640Frame = new ushort[834];
```
Ellerbach, line 122: Why 834?

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+            int resolutionRAM = (frameData[832] & 0x0C00) >> 10;
+            float resolutionCorrection = (float)(Math.Pow(2, _resolutionEE) / Math.Pow(2, resolutionRAM));
+            vdd = (float)(((resolutionCorrection * vdd - _vdd25) / _kVdd) + 3.3);
+
+            return vdd;
+        }
+
+        private ushort GetFrameData(ref ushort[] frameData)
+        {
+            Span<ushort> statusRegister = stackalloc ushort[0];
```
Ellerbach, line 278: You don't need to allocate if you're reading it right after

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+            do
+            {
+                statusRegister = ReadRegisters16(0x8000, 1);
+            }
+            while ((byte)(statusRegister[0] & 0x0008) == 0);
```
Ellerbach, line 283: if there a chance that statusRegister[0] doesn't exist?

Ellerbach: And is there a change that there will be some timeout? Because this may just loop forever. So a timeout pattern may be a good idea

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+
+            byte cnt = 0;
+            Span<ushort> framedata = stackalloc ushort[832];
+
+            do
+            {
+                Write16(0x8000, 0x0030);
```
Ellerbach, line 290: Would be good to have a reference in the documentation for those 2 magic numbers

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+                framedata = ReadRegisters16(0x0400, 832);
+                statusRegister = ReadRegisters16(0x8000, 1);
+                cnt += 1;
```
Ellerbach, line 293: where are you using cnt? I may be tired but I don't see before the end of the functions

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+            }
+            while ((byte)(statusRegister[0] & 0x0008) == 0);
```
Ellerbach, line 295: Same, would be good to have a timeout pattern somehow

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+
+            Span<ushort> controlRegister = stackalloc ushort[1];
```
Ellerbach, line 297: no need to allocate if you just get it allocated after.
Also having a Span<ushort> doesn't really bring value here as you don't slice of do anythins special with it. Same for the statusRegister btw

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+                Temperature tr = Temperature.FromDegreesCelsius(GetTa(mlx90640Frame).DegreesCelsius - OPENAIR_TA_SHIFT);
+                CalculateTo(mlx90640Frame, emissivity, tr, ref framebuf);
+            }
+
+            return framebuf;
+        }
+
+        private void CalculateTo(ushort[] frameData, float emissivity, Temperature tr, ref Temperature[] result)
+        {
+            ushort subPage = frameData[833];
```
Ellerbach, line 140: is there a page somewhere on the documentation that explains how to make this math?

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+                framedata = ReadRegisters16(0x0400, 832);
+                statusRegister = ReadRegisters16(0x8000, 1);
+                cnt += 1;
+            }
+            while ((byte)(statusRegister[0] & 0x0008) == 0);
+
+            Span<ushort> controlRegister = stackalloc ushort[1];
+            controlRegister = ReadRegisters16(0x800D, 1);
+
+            for (int i = 0; i < framedata.Length; i++)
```
Ellerbach, line 300: Are you always sure the destination buffer is large enough? And why not just use the CopyTo function?

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+
+            frameData[832] = controlRegister[0];
+            frameData[833] = (ushort)(statusRegister[0] & 0x0001);
+
+            return frameData[833];
+        }
+
+        private void DumpEE()
+        {
+            _eeData = ReadRegisters16(0x2400, 832);
```
Ellerbach, line 313: for all magic numbers like this, would be really great to have a pointer to the documentation

src/devices/Mlx90640/Mlx90640.cs
```diff
@@ -0,0 +1,957 @@
...
+            _ct[1] = 0;
+            _ct[2] = (_eeData[63] & 0x00F0) >> 4;
+            _ct[3] = (_eeData[63] & 0x0F00) >> 8;
+            _ct[2] *= step;
+            _ct[3] = _ct[2] + _ct[3] * step;
+
+            int ksToScale = (_eeData[63] & 0x000F) + 8;
+            ksToScale = 1 << ksToScale;
+
+            _ksTo[0] = _eeData[61] & 0x00FF;
```
Ellerbach, line 398: for those types of conversion, you may want to have a look at BinaryPrimitives.

src/devices/Mlx90640/README.md
```diff
@@ -0,0 +1,2 @@
+﻿# MLX90640 - Infra Red Camera
+The MLX90614 is an 32x24 pixel Infra Red camera for noncontact temperature measurements
```
Ellerbach, line 2: you should add the usage, reference on the documentation, example of usage (even simple)
Also, in Markdown, there is always an extra line after all titles

src/devices/Mlx90640/Sampling.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Iot.Device.Mlx90640
+{
+    /// <summary>
+    /// MLX90640 sampling
+    /// </summary>
+    public enum Sampling : byte
```
Ellerbach, line 12: see comment about making is ushort and adding directly << 7 in the enum numbers.

Ellerbach (commented):
Couple of comments, al up, seems very interesting sensor!

src/devices/Mlx90640/Mlx90640.cs
garciaolais: That's is the allocation memory recommended in the datasheet.

src/devices/Mlx90640/Mlx90640.cs
garciaolais: check **11.2.2.9. Calculating To for basic temperature range**

https://www.melexis.com/-/media/files/documents/datasheets/mlx90640-datasheet-melexis.pdf

src/devices/Mlx90640/Mlx90640.cs
garciaolais: Ok, in this case, `0x2400` it's the start address of the **EEPROM**

src/devices/Mlx90640/Mlx90640.cs
garciaolais: You are right, `cnt` it's useless 🤦‍♂️

src/devices/Mlx90640/Mlx90640.cs
garciaolais: Ok, I will take your suggestion of `CopyTo `

Ellerbach:
[Triage] @garciaolais do you think you'll have time to finalize the PR and make the few changes requested?

garciaolais:
@Ellerbach 
Hi, I hope to have time in the next few weeks 🗓️

joperezr:
That's great to hear @garciaolais! We are now using a bot that helps us with bookkeeping on the inactive open PRs so depending on when you are able to come back, the bot may flag or even close this PR before you come back. If that happens, just feel free to re-open it whenever you are ready to jumb back into the contribution 😄

PR #1392, opened 2020-12-23
Adding PCD8544, famous Nokia 5110 LCD screen

Ellerbach:
- Adding full support for PCD8544 screen
- Adding documentation
- Adding basic point, line and rectangle primitive
- Complete example using all the features of the screen


src/devices/Pcd8544/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+Supply voltage can vary, 3.3 or 5V, up to 7V is acceptable. Make sure that there is a current limiting resistor between the brightness led and the controller you'll plug. Most of the kits you buy have it. If you don't add one, you will damage the external brightness leds.
+
+The following example shows how to setup the screen with a PWM to control the brightness:
+
+```csharp
+SpiConnectionSettings spiConnection = new(0, 1) { ClockFrequency = 5_000_000, Mode = SpiMode.Mode0, DataFlow = DataFlow.MsbFirst, ChipSelectLineActiveState = PinValue.Low };
+PwmChannel pwmChannel = PwmChannel.Create(0, 0);
+SpiDevice spi = SpiDevice.Create(spiConnection);
+lcd = new(27, 22, spi, pwmChannel);
```
pgrawehr, line 27: You need to provide the type here.

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+using System.Threading;
+using System.Drawing;
+using Iot.Device.Display.Pcd8544Enums;
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// PCD8544 - 48 × 84 pixels matrix LCD
+    /// </summary>
+    public class Pcd8544 : IDisposable
```
pgrawehr, line 17: I think it should be possible to implement `ICharacterLcd` (from the CharacterLCD package). This will allow using the LcdCharacterEncoding capabilities, which provides help to work with different cultures.

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+        /// Change the back light from 0 to 100
+        /// </summary>
+        public float BacklightBrightness
+        {
+            get => _backlightVal;
+            set
+            {
+                if (_pwmBacklight != null)
+                {
+                    _backlightVal = value > 1 ? 1 : value;
```
pgrawehr, line 130: Handle negative values?

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: I was hesitating as the screen does not embedded all the capabilities. But I will look at it.

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: yes, good catch. I had a byte and change to float  because of the PWM. Didn't adjust. Thanks!

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: There is no blinking cursor for example in the screen or no real concept of cursor.
And there is no concept of font at all, reason why I just embed the original one.
So that's the reason I only picked few of the functions. But I'll be happy to understand how to make this better integrated if you have an idea.
This device is actually closer to Ssd13xx than the ones in CharacterLcd

src/devices/Pcd8544/FunctionSet.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
```
krwq, line 5:
```suggestion
```

src/devices/Pcd8544/FunctionSet.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
```
krwq, line 4: unnecessary using

src/devices/Pcd8544/NokiaFont.cs
```diff
@@ -0,0 +1,114 @@
...
+        public static readonly byte[][] Ascii = new byte[][]
+        {
+        new byte[] { 0x00, 0x00, 0x00, 0x00, 0x00 }, // 20
+        new byte[] { 0x00, 0x00, 0x5f, 0x00, 0x00 }, // 21 !
+        new byte[] { 0x00, 0x07, 0x00, 0x07, 0x00 }, // 22 "
+        new byte[] { 0x14, 0x7f, 0x14, 0x7f, 0x14 }, // 23 #
+        new byte[] { 0x24, 0x2a, 0x7f, 0x2a, 0x12 }, // 24 $
+        new byte[] { 0x23, 0x13, 0x08, 0x64, 0x62 }, // 25 %
+        new byte[] { 0x36, 0x49, 0x55, 0x22, 0x50 }, // 26 &
+        new byte[] { 0x00, 0x05, 0x03, 0x00, 0x00 }, // 27 '
```
krwq, line 23: indent seems off

src/devices/Pcd8544/NokiaFont.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// The specific PCD8544 font for Nokia 5110
+    /// </summary>
+    public static class NokiaCharacters
+    {
+        /// <summary>
+        /// ASCII Font specific to the PCD8544 Nokia 5110 screen
+        /// </summary>
+        public static readonly byte[][] Ascii = new byte[][]
+        {
+        new byte[] { 0x00, 0x00, 0x00, 0x00, 0x00 }, // 20
```
krwq, line 16: where do these come from? are those columns in binary?

src/devices/Pcd8544/Pcd8544.cs
krwq: I think we should not try to do that as part of this PR. I'd rather start thinking on display abstraction again and then think of some sort of solution to make all displays implement character lcd (i.e. create some class which adds the functionality to any display)

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+        /// </summary>
+        public const int ScreenSizeBytes = 504;
+
+        /// <summary>
+        /// The size of the screen in terms of characters
+        /// </summary>
+        public static Size Size => new Size(84, 6);
+
+        /// <summary>
+        /// The size of the screen in terms of characters
```
krwq, line 30:
```suggestion
        /// The size of the screen in terms of pixels
```

krwq: I'm actually confused here. The size looks similar to what you have for characters but at the same time name suggests pixels

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+{
+    /// <summary>
+    /// PCD8544 - 48 × 84 pixels matrix LCD
+    /// </summary>
+    public class Pcd8544 : IDisposable
+    {
+        /// <summary>
+        /// Size of the screen 48 x 84 / 8
+        /// </summary>
+        public const int ScreenSizeBytes = 504;
```
krwq, line 22: ScreenBufferByteSize

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+
+        /// <summary>
+        /// The size of the screen in terms of characters
+        /// </summary>
+        public static Size PixelScreenSize => new Size(84, 48);
+
+        /// <summary>
+        /// Number of color per pixel
+        /// </summary>
+        public const int ColorPerPixel = 1;
```
krwq, line 37: I'm not sure what this means.

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+        /// <summary>
+        /// Create Pcd8544
+        /// </summary>
+        /// <param name="dataCommandPin">The data command pin.</param>
+        /// <param name="resetPin">The reset pin. Use a negative number if you don't want to use it</param>
+        /// <param name="spiDevice">The SPI device.</param>
+        /// <param name="pwmBacklight">The PWM channel for the back light</param>
+        /// <param name="gpioController">The GPIO Controller.</param>
+        /// <param name="shouldDispose">True to dispose the GPIO controller</param>
+        public Pcd8544(int dataCommandPin, int resetPin, SpiDevice spiDevice, PwmChannel? pwmBacklight = null, GpioController? gpioController = null, bool shouldDispose = true)
```
krwq, line 64: could the backlight be also a pin?

krwq: (instead of gradient you would just get on/off)

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+            }
+
+            _dataCommandPin = dataCommandPin;
+            _pwmBacklight = pwmBacklight;
+            _pwmBacklight?.Start();
+            _spiDevice = spiDevice ?? throw new ArgumentNullException(nameof(spiDevice));
+            _shouldDispose = gpioController == null || shouldDispose;
+            _controller = gpioController ?? new();
+            _resetPin = resetPin;
+            if (resetPin > 0)
```
krwq, line 78: if reset pin is not required then I suggest to make it default to -1 and also check directly for -1 (I think pin 0 is perfectly valid number and you block it here)

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+                _enabled = value;
+                byte enab = (byte)(_enabled ? FunctionSet.PowerOn : FunctionSet.PowerOff);
+                SpiWrite(false, new byte[] { enab });
+            }
+        }
+
+        /// <summary>
+        /// Change the back light from 0 to 100
+        /// </summary>
+        public float BacklightBrightness
```
krwq, line 123: after this PR is merged we should think of joining abstraction of https://github.com/dotnet/iot/tree/master/src/devices/BoardLed and this into one. I think it would be better to take Led directly as an input in the constructor here.

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+                    _backlightVal = _backlightVal < 0 ? 0 : _backlightVal;
+                    _pwmBacklight.DutyCycle = _backlightVal;
+                }
+            }
+        }
+
+        /// <summary>
+        /// The bias for 0 to 7
+        /// </summary>
+        public byte Bias
```
krwq, line 140: Can you explain what is this biasing?

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+            {
+                _bias = value < 8 ? value : throw new ArgumentException($"Bias can't be more than 7");
+                SpiWrite(false, new byte[] { (byte)(_enabled ? FunctionSet.PowerOn | FunctionSet.ExtendedMode : FunctionSet.PowerOff | FunctionSet.ExtendedMode), (byte)(0x10 | _bias) });
+            }
+        }
+
+        /// <summary>
+        /// True to inverse the screen color
+        /// </summary>
+        public bool InverseMode
```
krwq, line 153: I'd suggest `InvertedColors`

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+
+        #region Text
+
+        /// <summary>
+        /// Write text
+        /// </summary>
+        /// <param name="text">The text to write</param>
+        public void Write(string text)
+        {
+            byte[] letter = new byte[6];
```
krwq, line 233: nit: stackalloc

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+            foreach (char c in text.ToCharArray())
```
krwq, line 234: 1. you don't need to call ToCharArray, you can enumerate chars directly
2. You can use `ReadOnlySpan<char>` rather than string. String will cast correctly

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+            y = y > Size.Height ? Size.Height : y;
+            SetCursorPosition(0, y);
+        }
+
+        /// <summary>
+        /// Write a raw byte stream to the display.
+        /// Used if character translation already took place.
+        /// </summary>
+        /// <param name="text">Text to print</param>
+        public void Write(ReadOnlySpan<byte> text)
```
krwq, line 273: I'd refrain from calling this `text`, I suggest `buffer` instead.

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+
+        /// <summary>
+        /// Draw a line
+        /// </summary>
+        /// <param name="x1">The first point X coordinate.</param>
+        /// <param name="y1">The first point Y coordinate.</param>
+        /// <param name="x2">The second point X coordinate.</param>
+        /// <param name="y2">The second point Y coordinate.</param>
+        /// <param name="isOn">True if the line has pixels on, false for off.</param>
+        public void DrawLine(int x1, int y1, int x2, int y2, bool isOn)
```
krwq, line 347: I generally think utilities like this should be on generic abstractions but since we don't have display abstraction I'm fine with that as is for now.

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Draw a line
+        /// </summary>
+        /// <param name="p1">First point coordinate.</param>
+        /// <param name="p2">Second point coordinate.</param>
+        /// <param name="isOn">True if the line has pixels on, false for off.</param>
+        public void DrawLine(Point p1, Point p2, bool isOn) => DrawLine(p1.X, p1.Y, p2.X, p2.Y, isOn);
```
krwq, line 384: I'd suggest only this API to be public.

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,487 @@
...
+        private void SpiWrite(bool isData, byte[] toSend)
+        {
+            _controller.Write(_dataCommandPin, isData ? PinValue.High : PinValue.Low);
+            _spiDevice.Write(toSend);
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            if (_shouldDispose)
```
krwq, line 466: null checks missing

krwq: (and null assignment)

src/devices/Pcd8544/README.md
```diff
@@ -0,0 +1,156 @@
+# PCD8544 - 48 × 84 pixels matrix LCD
```
krwq, line 1: I'd suggest put Nokia 5110 screen somewhere in the title to make it easier to find

src/devices/Pcd8544/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+Point, line and rectangle primitives are available:
+
+```csharp
+lcd.DrawPoint(5, 5, true);
+lcd.DrawLine(0, 0, 15, 35, true);
+lcd.DrawRectangle(10, 30, 10, 20, true, true);
+lcd.DrawRectangle(12, 32, 6, 16, false, true);
+// You should not forget to refresh to draw everything
+lcd.Refresh();
```
krwq, line 64: if drawing doesn't happen without calling this, I'd recommend to call it Draw

src/devices/Pcd8544/README.md
```diff
@@ -0,0 +1,156 @@
...
+```csharp
+lcd.InverseMode = true;
+```
+
+### Enabled
+
+You can switch on or off the screen fully with the `Enabled` property:
+
+```csharp
+lcd.Enabled = true;
```
krwq, line 151: I'd suggest renaming this because Enabled doesn't suggest it will turn it off (perhaps just `On`)

src/devices/Pcd8544/Temperature.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Display.Pcd8544Enums
+{
+    /// <summary>
+    /// Temperature set for the PCD8544
+    /// </summary>
+    public enum Temperature
```
krwq, line 9: Should this be internal? Also I'd suggest to change name of the enum because it will collide with UnitsNet

src/devices/Pcd8544/samples/Pcd8544.samples.csproj
```diff
@@ -0,0 +1,31 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="$(MainLibraryPath)System.Device.Gpio.csproj" />
+    <ProjectReference Include="..\Pcd8544.csproj" />
+    <ProjectReference Include="..\..\Ft4222\Ft4222.csproj" />
+    <ProjectReference Include="..\..\SoftPwm\SoftwarePwm.csproj" />    
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="me.bmp">
```
krwq, line 20: Contributor's privilege :shipit: 🚀

src/devices/Pcd8544/samples/Pcd8544.samples.csproj
```diff
@@ -0,0 +1,31 @@
...
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="nokia_bw.bmp">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="nonbmp.jpg">
```
krwq, line 26: consider making this file tiny bit smaller if possible

src/devices/Pcd8544/Pcd8544.cs
pgrawehr: I'm fine with delaying that for later, but actually that interface is exactly intended to be implemented by various displays to provide support for common functionality (such as text wrapping, culture dependent character translations etc).

src/devices/Pcd8544/Pcd8544.cs
pgrawehr: Having a simple LED class would certainly be fine, but consider that a) here a pin with PWM is preferred, which is for a simple LED already an advanced case and b) the BoardLed has a different purpose (it's actually related to a hardware feature of the Pi).

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: And what about for screens aligning with the API of Azure GUIX: https://docs.microsoft.com/fr-fr/azure/rtos/guix/about-guix?
So having something that is inspired from that?
Couple of comments on the naming in there are related as well to this choice.

krwq: my suggestion would be to go with ImageSharp and then build on top of that (although haven't read about GUIX that much just ImageSharp is recommended drawing library now) but let's discuss this outside of this PR

Ellerbach: By the way, this will be an issue: https://docs.sixlabors.com/articles/imagesharp/pixelformats.html#is-it-possible-to-store-a-pixel-on-a-single-bit-for-monochrome-images
If we can't support 1 bit for monochrome screens, almost none of those b&w screens will be supported. While having ImageSharp for high level design may work, we will still always have to rework on the generated buffer.
Also, they don't provide any interface or anything like this for screens

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: in the readme it is already explained

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: I just aligned on the interface for the LCD screens. so we should consider adjusting it

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: @pgrawehr btw if I implement fully the ICharacterLcd , then there are function which will throw a not implemented, is that acceptable?

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: I'll rename `ColorBitPerPixel`

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: yes, that's possible. So I'll produce another constructor for that and more logic inside

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: Yes, will do as well. That's actually good idea.

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: yep, will fix

src/devices/Pcd8544/README.md
Ellerbach: I was debating with myself :-) Thanks, `Draw` it will be!

src/devices/Pcd8544/README.md
Ellerbach: For all other bindings we have `Enabled` for those scenarios, so just wanted to align. I can clearly rename `On` but that would not align with our general approach for other bindings.

src/devices/Pcd8544/FunctionSet.cs
Ellerbach: needed as [Flags] is used

src/devices/Pcd8544/NokiaFont.cs
Ellerbach: Yes, adding

src/devices/Pcd8544/NokiaFont.cs
Ellerbach: fixing

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: good catch

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: And yes for >=0

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: It's not directly a board led as the BoardLed class would have to be adjusted for such scenarios. And if we adjust it to be constructed with a Pin or a PWM, that can clearly work.

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: And will add here as well.

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: This function is coming from the ICharacterLcd. Fine to adjust anything but it should be consistant.

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: moved the byte array to Span, it should be byte as then I have to convert it to a byte array anyway

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: It doesn't really cost anything to have few more options.

src/devices/Pcd8544/Temperature.cs
Ellerbach: I will rename it, it has to be public. It's actually called Temperature in the docs but will rename `ScreenTemperature`

src/devices/Pcd8544/samples/Pcd8544.samples.csproj
Ellerbach: adjusted

Ellerbach: adjusted

src/devices/Pcd8544/Pcd8544.cs
pgrawehr: @Ellerbach I suppose it would be better to just ignore unsupported calls, especially if there's no side effect to be expected for the caller if it doesn't work. I.e. if your display doesn't support enabling/disabling the backlight, then the getter would just always return true and the setter wouldn't do anything. A cursor (although not blinking) could be implemented by printing a _ letter.

pgrawehr: We could also be discussing about removing these properties from the interface or move them to a derived one, to make the thing more generic.

src/devices/Pcd8544/Pcd8544.cs
krwq: fair enough, we can clean it up once we get the actual abstraction

src/devices/Pcd8544/NokiaFont.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// The specific PCD8544 font for Nokia 5110
+    /// </summary>
+    public static class NokiaCharacters
+    {
+        /// <summary>
+        /// ASCII Font specific to the PCD8544 Nokia 5110 screen.
+        /// Font characters are column bit mask.
+        /// Font size is 5 pixels width and 8 pixels height. Each byte represent a vertical column for the character.
+        /// </summary>
+        public static readonly byte[][] Ascii = new byte[][]
```
krwq, line 16: note that only the outer array is readonly but you can still edit the content. I don't expect any action here as I can imagine users might want that.

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,544 @@
...
+        /// Enable the screen
+        /// </summary>
+        public bool Enabled
+        {
+            get => _enabled;
+            set
+            {
+                _enabled = value;
+                byte enab = (byte)(_enabled ? FunctionSet.PowerOn : FunctionSet.PowerOff);
+                SpiWrite(false, new byte[] { enab });
```
krwq, line 159: consider span

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,544 @@
...
+        /// <summary>
+        /// The bias for 0 to 7. Bias represent the voltage applied to the LCD. The highest, the darker the screen will be.
+        /// </summary>
+        public byte Bias
+        {
+            get => _bias;
+            set
+            {
+                _bias = value < 8 ? value : throw new ArgumentException($"Bias can't be more than 7");
+                SpiWrite(false, new byte[] { (byte)(_enabled ? FunctionSet.PowerOn | FunctionSet.ExtendedMode : FunctionSet.PowerOff | FunctionSet.ExtendedMode), (byte)(0x10 | _bias) });
```
krwq, line 195: consider span

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,544 @@
...
+        public void Write(string text)
+        {
+            Span<byte> letter = stackalloc byte[6];
+            foreach (char c in text)
+            {
+                // We only display specific characters and ignore the rest
+                // And only if it's in the screen
+                if (_position <= ScreenBufferByteSize - 5)
+                {
+                    if (c >= 0x20 && c <= 0x7F)
```
krwq, line 289: consider throwing when they're out of range. also for some characters like backspace you might consider moving back 1

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,544 @@
...
+        /// <param name="isOn">True if the point has pixels on, false for off.</param>
+        /// <returns>True if success</returns>
+        public bool DrawPoint(int x, int y, bool isOn)
+        {
+            if (x < 0 || x >= 84 || y < 0 || y >= 48)
+            {
+                return false;
+            }
+
+            int index = ((x % 84) + (int)(y * 0.125) * 84);
```
krwq, line 364: any reason you picked to do `(int)(y * 0.125)` rather than `y / 8`?

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,544 @@
...
+        /// <summary>
+        /// Draw a rectangle
+        /// </summary>
+        /// <param name="x">The X coordinate.</param>
+        /// <param name="y">The Y coordinate.</param>
+        /// <param name="width">The width of the rectangle.</param>
+        /// <param name="height">The height of the rectangle.</param>
+        /// <param name="isOn">True if the rectangle has pixels on, false for off.</param>
+        /// <param name="isFilled">If it's filled or not.</param>
+        public void DrawRectangle(int x, int y, int width, int height, bool isOn, bool isFilled)
```
krwq, line 444: Technically speaking this is called axis aligned rectangle. Fine as is though

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,544 @@
...
+                    _controller.ClosePin(_resetPin);
+                }
+
+                if (_controller.IsPinOpen(_backlightPin))
+                {
+                    _controller.ClosePin(_backlightPin);
+                }
+            }
+
+            _spiDevice.Dispose();
```
krwq, line 538:
```suggestion
            _spiDevice?.Dispose();
```

src/devices/Pcd8544/Pcd8544.cs
```diff
@@ -0,0 +1,544 @@
...
+        public void Dispose()
+        {
+            if (_shouldDispose)
+            {
+                _controller?.Dispose();
+                _controller = null!;
+            }
+            else
+            {
+                if (_controller.IsPinOpen(_dataCommandPin))
```
krwq, line 522: consider `if (_controller is not null)` for all these

src/devices/Pcd8544/ScreenTemperature.cs
```diff
@@ -0,0 +1,23 @@
...
+        Coefficient0 = 0b0000_0100,
+
+        /// <summary>Temperature Coefficient 1</summary>
+        Coefficient1 = 0b0000_0101,
+
+        /// <summary>Temperature Coefficient 2</summary>
+        Coefficient2 = 0b0000_0110,
+
+        /// <summary>Temperature Coefficient 3</summary>
+        Coefficient3 = 0b0000_0111,
```
krwq, line 21: If there is more info in the docs what those coefficients mean consider adding those info in the docs. Currently it's hard to tell what this means

krwq (approved):
nothing blocking, couple of nits/suggestions

src/devices/Pcd8544/Pcd8544.cs
pgrawehr: ... or maybe an overload that takes a single byte. Using SpiDevice.WriteByte() would not require any object instantiation.

src/devices/Pcd8544/ScreenTemperature.cs
Ellerbach: I would love to have more info! The doc just give them and recommend to use the default I'm using and mentioned in the readme. The sample goes thru all modes and covers everything. So everyone can understand the impact of each element.

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: yes, if I use /8, I will not get a proper rounding.

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: it's ok if it writes out of the screen, it's like for drawing. But great catch for backspace!

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: I was clearly a bit reluctant to create Span and then stackalloc and then put everything in there for 1 or 2 bytes. I understand the logic with the stack allocation but still.
Any advantage to use Span? 
And yes @pgrawehr , that would help to have a WriteByte(), there are quite some scenarios like this.

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: @pgrawehr, let me give a try. And let see how it can look like.
@krwq I fully migrated it to ImageSharp. Was quite straight forward. See this commit

src/devices/Pcd8544/NokiaFont.cs
Ellerbach: yes. But even if a developer want to change a specific part of the font, that's kind of acceptable if they know what they are doing!

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: And @krwq my primitive internal function takes byte[], not Span. I can always adjust but does that makes sense?

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: @pgrawehr, I gave it a try and I'm bock because now I4m using ImageSharp, the ICharacterLcd is not and it's still using System.Drawing. See [for example](https://github.com/dotnet/iot/blob/ca828c3d28d34976a8e07d24357e029e66b87905/src/devices/CharacterLcd/ICharacterLcd.cs#L38)
For the rest I can manage. That's more than doable. For the cursor, it will be a bit weird as it's not blinking, underlining should be manageable. 
For custom Characters, here is the question: technically, you can really create almost as many as you want, it's just about having a list of byte[] in the binding and that's it. Now because the Characters can be any format but so far are aligned with the Nokia font embedded, they are byte[] of 5.
The doc says byte[] of 8. [See](https://github.com/dotnet/iot/blob/ca828c3d28d34976a8e07d24357e029e66b87905/src/devices/CharacterLcd/ICharacterLcd.cs#L91)
So I can then just return 0, meaning no special character can be allowed. Also, then when displaying them, I don't know how to properly handle this.
So all up, if you remove the custom character map from the interface and move to ImageSharp, then I'm sure that can be actually a pretty nice interface for screens as well. At least for a basic usage, that will perfectly work!

Ellerbach:
> nothing blocking, couple of nits/suggestions

@krwq any reason why the ImageSharp nudget is not restored? I guess because it should be added for the all up project but just wondering.

src/devices/Pcd8544/Pcd8544.cs
pgrawehr: @Ellerbach The Type `System.Drawing.Size` is in `System.Drawing.Primitives.dll`, not in `System.Drawing.Common.dll`, which seems to be part of the framework. Are you sure this should also be removed? Not a big deal, but still this is a different assembly.

The custom characters for the LCD displays currently use row-oriented byte[8] characters. The characters are 5x8 pixels, each byte represents a row (only the lower 5 bits used). The documentation on the interface is actually wrong, as only the implementation defines the expected size, not the interface. So if your implementation takes byte[5], that's basically fine. 

Since the font size in the end is also 5x8, you could (automatically) transpose those bitmasks to the same format. Then you could basically declare the `NumberOfCustomCharacters` as 256 and implement the font as an encoding. This would even (since the font sizes are equal) allow the use of the existing font maps for the custom characters I have added already. I can understand if that last part sounds a bit complicated, but I can have a look at it myself if you want.

Ellerbach: Regarding `System.Drawing`,from the issue #1403 and thid: https://docs.microsoft.com/en-us/dotnet/api/system.drawing?view=dotnet-plat-ext-5.0#remarks, I guess we should really move out fully.
I know it's not much but if we move, we have to move all. @krwq any thoughts?

Regarding the cataracter set, ok, got it, so that can be done. I can then have this implemented. The new characters will replace the default ones if one is added with the same number. This will allow fully customize fonts.
Got it for the font orientation. I can even detect if it's byte[5] or byte[8], assuming if it's byte[8], it's then 5 rows and transforms it.

Thought on this?

pgrawehr: Sounds good. Just be aware that the custom characters so far do not have fixed numbers. Their place in the character map is determined by the culture and the fixed font.

Ellerbach: > Their place in the character map is determined by the culture and the fixed font.

Can you give me an example so it's more clear to me?

src/devices/Pcd8544/Pcd8544.cs
krwq: there is less GC pressure overall. It might affect larger apps (it's about many tiny allocations all over not about this particular one). I do not think we currently have a need for super good perf. I'm ok with this as is and changing that in the future when we have a specific perf goals (that's why I wrote "consider" in the first place, nothing blocking here)

src/devices/Pcd8544/Pcd8544.cs
krwq: Can you give specific example? I've never had to use floating points in routines like this. (I'm ok with this as is, just curious)

krwq:
@Ellerbach I suspect it is restored but you also need to add the PackageReference to Iot.Device.Bindings.csproj. Also consider moving version of that package to: https://github.com/dotnet/iot/blob/main/eng/Versions.props

src/devices/Pcd8544/Pcd8544.cs
pgrawehr: You'll find relevant samples in `LcdConsole.Samples.cs` (if you've done everything right<sup>tm</sup>, that example should work against your interface implementation with a minor change only). 

Basically the workflow is as follows, given an instance of `ICharacterLcd`:

```csharp
ICharacterLcd lcd=new Pcd8544(...)
LcdConsole console = new LcdConsole(lcd, "NoFontRom", false); // Use a descriptive string for your "empty" character rom
console.Write(...); // *
LcdCharacterEncoding encoding = LcdConsole.CreateEncoding(CultureInfo.CreateSpecificCulture("de-CH"), "NoFontRom", '?', 256); // Provide Rom name again, and number of available custom characters. This knows which letters the given language requires (i.e. in this case äöüÄÖÜ) and internally maps these to the free slots. So for instance, "ä" might become character value 0.
console.LoadEncoding(encoding); // Load the encoding. This also loads all required custom characters.
console.Write(...) // Now this uses the address translation from the encoding to know that the string "äää" shall actually print 3 times the letter with character value 0.

// * That would now do nothing visible, since there's no default font. - case to consider separately, because using the display without an explicit character set will lead to unexpected behavior, also with the current implementation. Currently, it defaults to a simple cast of char to byte, but I consider changing this to an exception instead.
```

While writing the above, I found that it seems the `LcdCharacterEncoding` class is missing (at least) a few `virtual` declarations to make the scenario work with a fully customized font. I think we should have a call on the details.

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: ho, I think this is what I originally wanted to do: int index = (int)((x % 84) + (y * 0.125) * 84);
and that's different. Need to test again that it's the expected behavior with the actual code, in that case, yes, the /8 will work perfectly!

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: OK, moved all to Span<byte>, thanks for the explanation, that helps.

src/devices/Pcd8544/Pcd8544.cs
Ellerbach: @pgrawehr I implemented the ICharacterLcd and all the cursor part plus some needed functions, had to do some refactoring/transformation to fully and properly support cursors, now working all good. Moved the CharacterLcd namespace to ImageSharp.
Still some work on the font side now. Trying to understand the best way to handle all this, based as well on our discussion. That's next on the list!

Ellerbach: @krwq and @pgrawehr I would go for merge for this PR as it implements all the basics for the CharacterLcd and a certain level of compatibility. Now to support fonts and few other advanced scenarios, that would require more changes in both classes (we've discussed with @pgrawehr some of the implication). That's then a work I can do later on as I won't really have time to make this soon.
By merging this one, it can be on the 1.4 and we have a decent support so the screen can be used like a LCD as well.

Ellerbach: @krwq and @pgrawehr I added 2 special fonts 5x7 and 5x8 containing a lot of unicode characters to cover almost all scenarios. Those had been autogenerated.
I added as well the support for those specific characters for the screen. Example updated.
I had to update the ILcdCharacter for this additional support. I don't have any small screen to test on it. So a deep test should be done before merge. Rest seems to work on my Nokia screen.

src/devices/CharacterLcd/Font5x7.cs
```diff
@@ -0,0 +1,10186 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// <autogenerated />
+using System.Collections.Generic;
```
krwq, line 5: how is this data generated? any chance we could generate this on the fly rather than having this dictionary?

Ellerbach: it is generated out of bdf files, various ones but we will have to adjust all this. Most of the characters are not the proper size. In all cases, a similar size or even large file will have to be included. It can be as well a raw stack of bytes, that doesn't really matter.
There is still work here to go because of the size that is not exactly what we need.

krwq: Can this use BdfFont similarly as LedMatrix rather than including this?

Ellerbach: We can try problem is that the bdf fonts are not proper ones. The one 5x7 is in reality 5x6, we're looking for a real 5x7 and real 5x8 with all proper unicode. That's the key challenge so far.

Ellerbach: And I'm all ok to, remove the font and the dome associated to it part of the program.cs and move forward this binding. We can work in parallel on this font issue. Just let me know (I won't have much time this week anyway)

src/devices/CharacterLcd/LcdInterface.I2c.cs
```diff
@@ -115,6 +117,31 @@ public override void SendData(ReadOnlySpan<byte> values)
...
+                Span<byte> bufferData = buffer.Slice(1);
+
+                while (values.Length > 0)
+                {
+                    ReadOnlySpan<char> buff = values.Slice(0, values.Length > MaxCopy ? MaxCopy : values.Length);
+                    // As we are in a while loop, we can't use stackalloc
+                    Span<byte> currentValues = new byte[buff.Length];
+                    for (int i = 0; i < buff.Length; i++)
+                    {
+                        currentValues[i] = (byte)buff[i];
```
krwq, line 137: you seem to be copying these without checking the range, you might get some characters randomly. Wouldn't it make more sense to ask people wanting to call this to use Encoding.ASCII.GetBytes or similar instead of this new method?

src/devices/CharacterLcd/LcdInterface.cs
```diff
@@ -33,6 +33,12 @@ public abstract partial class LcdInterface : IDisposable
         /// <param name="values">Bytes to be send to the device</param>
         public abstract void SendData(ReadOnlySpan<byte> values);
 
+        /// <summary>
+        /// Sends data to the LCD device
+        /// </summary>
+        /// <param name="values">Bytes to be send to the device</param>
```
krwq, line 39: Those are not really bytes... I don't find this very intuitive how you should implement this in the derived class? Should this instead be just some helper method like:

```csharp`
public void SendAscii(ReadOnlySpan<char> values)
{
    byte[] bytes = Encoding.ASCII.GetBytes(values);
    SendData(bytes);
}
```

src/devices/Pcd8544/NokiaFont.cs
```diff
@@ -0,0 +1,116 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// The specific PCD8544 font for Nokia 5110
```
krwq, line 7: as we talked, would it make sense for this to be provided as an external file and be shown as an example in the sample rather than having it as part of the library?

src/devices/CharacterLcd/LcdInterface.cs
Ellerbach: Will rephrase, it's actually char to be send. Good catch. Thanks.

src/devices/CharacterLcd/LcdInterface.I2c.cs
Ellerbach: The all purpose of LcdCharacterEncoding is to do this job. We add additional characters, not aligned with the traditional ASCII way. The class is a acutally doing this. It was already existing and working code, adjusted the definition only from byte to char to be compatible with devices having larger capacities than 255 characters

src/devices/Pcd8544/NokiaFont.cs
Ellerbach: yes and no. If this is not embedded, then there won't be any font at all in the screen, so it will not work at all to display any test at all.
You'll always has to load this font or another one to even display simple things like temperature or humidity. Which is for me extremely annoying. All LCD screen has some sort of embedded fonts always, at least minimal ones.

pgrawehr:
Did a quick test with the LCD display, nothing broken it seems. As discussed with @Ellerbach, the fonts provided so far have not the size we should have for those displays (they're generally 1 pixel to narrow and 1 pixel to small). But proper unicode support for all displays (or support for multiple different fonts for graphic displays) can be done in a separate PR.

src/devices/CharacterLcd/LcdInterface.I2c.cs
krwq: should we remove byte overload in that case?

Ellerbach: The byte one is the one used directly in the small LCD, the byte ones avoid having to reallocate a new buffer and avoid some over complicated part on the interface. @pgrawehr your feedback here would help as we chatted about this over the weekend.

pgrawehr: I would keep it, as it is used if the LcdInterface is used directly or if the character translation is already done. The char overload is only to prevent another buffer copy in case we do a char->char mapping instead.

krwq: but why can't char overload provide sane default implementation and be overriden only if it makes sense?

Ellerbach:
In the last commit, I also fixed #1441

src/devices/CharacterLcd/CharacterLcd.csproj
```diff
@@ -13,8 +13,8 @@
     <Compile Include="CharacterLcdExtensions.cs" />
     <Compile Include="Lcd1602.cs" />
     <Compile Include="Lcd2004.cs" />
-    <Compile Include="LcdCharacterEncoding.cs" />
-    <Compile Include="LcdCharacterEncodingFactory.cs" />
+    <Compile Include="..\Common\Iot\Device\Graphics\LcdCharacterEncoding.cs" />
```
pgrawehr, line 16: Reference the common project instead of individual files.

src/devices/Pcd8544/Pcd8544.csproj
```diff
@@ -7,6 +7,8 @@
 
   <ItemGroup>
     <Compile Include="*.cs" />
+    <Compile Include="..\Common\Iot\Device\Graphics\BdfFont.cs" />
```
pgrawehr, line 10: Same here. Don't include files from other tree locations directly (might cause problems if another project needs to import this one)

pgrawehr (approved):

src/devices/CharacterLcd/LcdInterface.cs
```diff
@@ -33,6 +33,12 @@ public abstract partial class LcdInterface : IDisposable
         /// <param name="values">Bytes to be send to the device</param>
         public abstract void SendData(ReadOnlySpan<byte> values);
 
+        /// <summary>
+        /// Sends data to the LCD device
+        /// </summary>
+        /// <param name="values">Char to be send to the device</param>
+        public abstract void SendData(ReadOnlySpan<char> values);
```
krwq, line 40: I thought you were going to make this non-abstract and call into the byte version

krwq (approved):
I'd prefer we made `SendData(ReadOnlySpan<char>)` overload not public but I'm ok with merging as is if you don't have time - it's protected API and it's unlikely users will find this problematic

Ellerbach:
> I'd prefer we made `SendData(ReadOnlySpan<char>)` overload not public but I'm ok with merging as is if you don't have time - it's protected API and it's unlikely users will find this problematic

I have to admit I don't really have time right now. And as you write, it's normally no real issue.

joperezr:
Hey @Ellerbach seems like there is a merge conflict here, do you mind fixing and then merging?

Issue #567, opened 2019-07-05
Add more metadata about devices


krwq:
In https://github.com/dotnet/iot/pull/554 I've added simple category.txt files next to devices which allow generating categorized device bindings index.

In https://github.com/dotnet/iot/pull/554#issuecomment-508040445 @shaggygi has proposed we should investigate Digital Twin Definition Language (DTDL) format. I.e. `implements` section is very similar to categories.

DTDL would allow us not only to generate bulk part of README.md in a consistent way but also to possibly use those files in more orthogonal setups provided by different services.

Work here:
- investigate specific applications of this feature (i.e. plug & play capability)
- assuming we find it useful:
  - generate first version of DTDL file from category.txt + README.md (or at least placeholder for that)
  - implement min-value product/PoC

PR #1393, opened 2020-12-23
Attributes for devices

krwq:
Read README.md for description.

Contributes to: #567, #418

cc: @pgrawehr @colombod

### DIFF
--- a/src/devices/Ags01db/Ags01db.cs
+++ b/src/devices/Ags01db/Ags01db.cs
@@ -5,14 +5,22 @@ using System;
 using System.Buffers.Binary;
 using System.Threading;
 using System.Device.I2c;
+using System.Device.Model;
+using UnitsNet;
 
 namespace Iot.Device.Ags01db
 {
     /// <summary>
     /// MEMS VOC Gas Sensor ASG01DB
     /// </summary>
+    [Interface("MEMS VOC Gas Sensor ASG01DB")]
     public class Ags01db : IDisposable
     {
+        /// <summary>
+        /// ASG01DB Default I2C Address
+        /// </summary>
+        public const byte DefaultI2cAddress = 0x11;
+
         // CRC const
         private const byte CRC_POLYNOMIAL = 0x31;
         private const byte CRC_INIT = 0xFF;
@@ -23,17 +31,13 @@ namespace Iot.Device.Ags01db
         /// <summary>
         /// ASG01DB VOC (Volatile Organic Compounds) Gas Concentration (ppm)
         /// </summary>
-        public double Concentration => GetConcentration();
+        [Telemetry]
+        public Ratio Concentration => Ratio.FromPartsPerMillion(GetConcentration());
 
         /// <summary>
         /// ASG01DB Version
         /// </summary>
         public byte Version => GetVersion();
-
-        /// <summary>
-        /// ASG01DB Default I2C Address
-        /// </summary>
-        public const byte DefaultI2cAddress = 0x11;
 
         /// <summary>
         /// Creates a new instance of the ASG01DB
--- a/src/devices/Ags01db/Ags01db.cs
+++ b/src/devices/Ags01db/Ags01db.cs
@@ -37,6 +37,7 @@ namespace Iot.Device.Ags01db
         /// <summary>
         /// ASG01DB Version
         /// </summary>
+        [Property]
         public byte Version => GetVersion();
 
         /// <summary>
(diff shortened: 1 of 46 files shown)


src/devices/System.Device.Model/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Type serialization
+
+Only simple types can be serialized:
+- enums (without Flags attribute)
+  - values out of enum range are not permitted (i.e. bitwise combination)
+- UnitsNet units
+- basic C# types
+- Vector2, Vector3, Vector4
+- Color
```
colombod, line 111: Maybe this should be fully qualified type names to make it easier to user what we refer to

colombod: Should we do the switch to portable library like ImageSharp at this point?

krwq, line 111:
```suggestion
- System.Numerics.Vector2, System.Numerics.Vector3, System.Numerics.Vector4
- System.Drawing.Color
```

colombod (approved):

src/devices/System.Device.Model/README.md
krwq: this doesn't really depend on it yet, once we convert Iot.Device.Bindings to ImageSharp that's when we should deprecate serialization of Color, that's why it's still experimental. Also only serializer will really depend on System.Drawing not the library itself

src/devices/Ina219/Ina219.cs
```diff
@@ -6,20 +6,22 @@
...
-    /// Binding that exposes an INA219 Bidirectional Current/Power monitor.
+    /// INA219 Bidirectional Current/Power monitor.
     ///
     /// The INA219 is a high-side current shunt and power monitor with an I2C interface.
     /// The INA219 monitors both shunt drop and supply voltage, with programmable conversion
     /// times and filtering. A programmable calibration value, combined with an internal multiplier,
     /// enables direct readouts in amperes. An additional multiplying register calculates power in watts.
     /// <see href="http://www.ti.com/lit/ds/symlink/ina219.pdf"/>
     /// </summary>
+    [Interface("INA219 Bidirectional Current/Power monitor")]
```
Ellerbach, line 24: are there any special character not allowed or which needs to be escaped? in this case the `/` ?

src/devices/SenseHat/SenseHat.cs
```diff
@@ -53,96 +83,96 @@ public SenseHat(I2cBus? i2cBus = null, bool shouldDispose = false)
         /// Fills LED matrix with specified color
         /// </summary>
         /// <param name="color">Color to fill LED matrix with</param>
-        public void Fill(Color color = default(Color)) => _ledMatrix.Fill(color);
+        public void Fill(Color color = default(Color)) => LedMatrix.Fill(color);
```
Ellerbach, line 86: Can those type of functions be `[Command]` ?

src/devices/System.Device.Model/README.md
```diff
@@ -0,0 +1,111 @@
...
+For properties providing name of the `Telemetry` is optional as it can be deduced from the property name.
+
+Telemetry can be put on:
+- properties
+- methods returning value but not taking any arguments
+- methods returning bool and taking one `out` argument
+  - multiple `out` arguments are currently out of scope but are considered
+
+```csharp
+        [Telemetry]
```
Ellerbach, line 40: remove the extra space, that's not needed, can do everywhere in the samples from this readme

Ellerbach (approved):
Few comments, great work! Looking forward for the next steps!

src/devices/SenseHat/SenseHat.cs
krwq: I wasn't sure about annotating any of the displays yet so I left them out of scope for now.

src/devices/Ags01db/Ags01db.cs
```diff
@@ -23,18 +31,14 @@ public class Ags01db : IDisposable
...
         /// ASG01DB VOC (Volatile Organic Compounds) Gas Concentration (ppm)
         /// </summary>
-        public double Concentration => GetConcentration();
+        [Telemetry]
+        public Ratio Concentration => Ratio.FromPartsPerMillion(GetConcentration());
 
         /// <summary>
         /// ASG01DB Version
         /// </summary>
         public byte Version => GetVersion();
```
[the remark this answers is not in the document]
krwq: I might have missed some on first couple of sensors when I still wasn't sure if these should be properties, I've fixed couple of more cases.

src/devices/SenseHat/SenseHat.cs
Ellerbach: Fair, let's see first with the first sensors. Impatient to see all this coming together :-)

PR #1396, opened 2020-12-27
Add device binding - Sensorion SGP30 TVOC / eCO2 Gas Sensor

PurpleGranite:
Adds a device binding for the Sensorion SGP30 TVOC / eCO2 Gas Sensor.  All device features are implemented with the exception of the 'Measure Test' command, which is used during device manufacture and test only.

### DIFF
--- a/src/devices/Sgp30/ChecksumFailedException.cs
+++ b/src/devices/Sgp30/ChecksumFailedException.cs
@@ -1,0 +1,11 @@
+﻿using System;
+
+namespace Iot.Device.Sgp30
+{
+    /// <summary>
+    /// Exception raised where an SGP30 Checksum Validation fails during a device read operation.
+    /// </summary>
+    public class ChecksumFailedException : Exception
+    {
+    }
+}
--- a/src/devices/Sgp30/README.md
+++ b/src/devices/Sgp30/README.md
@@ -1,0 +1,18 @@
+﻿# Sgp30
+
+## Summary
+Provide a brief description on what the component is and its functionality.
+
+## Device Family
+Provide a list of component names and link to datasheets (if available) the binding will work with.
+
+**[Family Name Here]**: [Datasheet link here]
+
+## Binding Notes
+
+Provide any specifics related to binding API.  This could include how to configure component for particular functions and example code.
+
+**NOTE**:  Don't repeat the basics related to System.Device.API* (e.g. connection settings, etc.).  This helps keep text/steps down to a minimum for maintainability.
+
+## References 
+Provide any references to other tutorials, blogs and hardware related to the component that could help others get started.
--- a/src/devices/Sgp30/README.md
+++ b/src/devices/Sgp30/README.md
@@ -1,18 +1,8 @@
-﻿# Sgp30
+﻿# Sensorion SGP30 TVOC + eCO2 Gas Sensor
+This is a device binding for the Sensorion SGP30 TVOC + eCO2 Gas Sensor.
 
-## Summary
-Provide a brief description on what the component is and its functionality.
-
-## Device Family
-Provide a list of component names and link to datasheets (if available) the binding will work with.
-
-**[Family Name Here]**: [Datasheet link here]
+## Datasheet
+[SGP30](https://cdn.shopify.com/s/files/1/0174/1800/files/Sensirion_Gas_Sensors_SGP30_Datasheet_EN-1148053.pdf)
 
 ## Binding Notes
-
-Provide any specifics related to binding API.  This could include how to configure component for particular functions and example code.
-
-**NOTE**:  Don't repeat the basics related to System.Device.API* (e.g. connection settings, etc.).  This helps keep text/steps down to a minimum for maintainability.
-
-## References 
-Provide any references to other tutorials, blogs and hardware related to the component that could help others get started.
+**NOTE**: Once initialised, this sensor **must** be queried at least once per second or re-initialisation will be required.
--- a/src/devices/Sgp30/README.md
+++ b/src/devices/Sgp30/README.md
@@ -5,4 +5,41 @@ This is a device binding for the Sensorion SGP30 TVOC + eCO2 Gas Sensor.
 [SGP30](https://cdn.shopify.com/s/files/1/0174/1800/files/Sensirion_Gas_Sensors_SGP30_Datasheet_EN-1148053.pdf)
 
 ## Binding Notes
-**NOTE**: Once initialised, this sensor **must** be queried at least once per second or re-initialisation will be required.
+**NOTE**: Once initialised, this sensor should be queried at least once per second. If more than several seconds pass,
+the initialisation routine should be called again to ensure the device is returning valid data.
+
+```cs
+// Setup Device Instance
+I2cDevice sgp30Device = I2cDevice.Create(new I2cConnectionSettings(1, Sgp30.DefaultI2cAddress));
+Sgp30 sgp30 = new Sgp30(sgp30Device);
+
+// Initialise the SGP30 device
+Console.WriteLine("Wait for initialisation (upto 20 seconds).");
+try
+{
+    Sgp30Measurement measurement = sgp30.InitialiseMeasurement();
+    Console.WriteLine($"TVOC: {measurement.Tvoc.ToString()} ppb, eCO2: {measurement.Eco2.ToString()} ppm.");
+}
+catch (ChecksumFailedException)
+{
+    Console.WriteLine("Checksum was invalid when initialising SGP30 measurement operation.");
+    throw;
+}
+```
+
+After initialisation, getting sensor readings is straightforward.
+```cs
+try
+{
+    Sgp30Measurement measurement = sgp30.GetMeasurement();
+    Console.WriteLine($"TVOC: {measurement.Tvoc.ToString()} ppb, eCO2: {measurement.Eco2.ToString()} ppm.");
+}
+catch (ChecksumFailedException)
+{
+    Console.WriteLine("Checksum was invalid when reading SGP30 measurement.");
+    throw;
+}
+```
+
+Note that all device commands may throw `ChecksumFailedException` if the data and checksum read from the device do not
+match. This will typically indicate that the connected device is not an SGP30 sensor.
--- a/src/devices/Sgp30/Sgp30.cs
+++ b/src/devices/Sgp30/Sgp30.cs
@@ -48,7 +48,7 @@ namespace Iot.Device.Sgp30
         /// </summary>
         /// <returns>Device Serial ID.</returns>
         /// <exception cref="ChecksumFailedException">Thrown if checksum validation fails.</exception>
-        public ushort[]? GetSerialId()
+        public ushort[] GetSerialId()
         {
             Span<byte> resultBuffer = stackalloc byte[10];
             _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[] { (byte)((SGP30_GET_SERIAL_ID & 0xFF00) >> 8), (byte)(SGP30_GET_SERIAL_ID & 0x00FF) }));
@@ -57,16 +57,16 @@ namespace Iot.Device.Sgp30
 
             ushort[] serialValues = new ushort[]
             {
-                        (ushort)(resultArray[1] << 8 | resultArray[2]),
-                        (ushort)(resultArray[4] << 8 | resultArray[5]),
-                        (ushort)(resultArray[7] << 8 | resultArray[8])
+                (ushort)(resultArray[1] << 8 | resultArray[2]),
+                (ushort)(resultArray[4] << 8 | resultArray[5]),
+                (ushort)(resultArray[7] << 8 | resultArray[8])
             };
 
             byte[] checksums = new byte[]
             {
-                        CalculateChecksum(serialValues[0]),
-                        CalculateChecksum(serialValues[1]),
-                        CalculateChecksum(serialValues[2]),
+                CalculateChecksum(serialValues[0]),
+                CalculateChecksum(serialValues[1]),
+                CalculateChecksum(serialValues[2]),
             };
 
             if (checksums[0] != resultArray[3] || checksums[1] != resultArray[6] || checksums[2] != resultArray[9])
--- a/src/devices/Sgp30/Sgp30.cs
+++ b/src/devices/Sgp30/Sgp30.cs
@@ -137,21 +137,21 @@ namespace Iot.Device.Sgp30
         public Sgp30Measurement GetMeasurement()
         {
             Span<byte> resultBuffer = stackalloc byte[7];
-            _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[] { (byte)((SGP30_INITIALISE_AIR_QUALITY & 0xFF00) >> 8), (byte)(SGP30_INITIALISE_AIR_QUALITY & 0x00FF) }));
+            _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[] { (byte)((SGP30_MEASURE_AIR_QUALITY & 0xFF00) >> 8), (byte)(SGP30_MEASURE_AIR_QUALITY & 0x00FF) }));
             Thread.Sleep(12);
             _i2cDevice.Read(resultBuffer.Slice(1));
             byte[] resultArray = resultBuffer.ToArray();
 
             Sgp30Measurement result = new Sgp30Measurement
             {
-                Tvoc = (ushort)(resultArray[1] << 8 | resultArray[2]),
-                Eco2 = (ushort)(resultArray[4] << 8 | resultArray[5])
+                Eco2 = (ushort)(resultArray[1] << 8 | resultArray[2]),
+                Tvoc = (ushort)(resultArray[4] << 8 | resultArray[5])
             };
 
             byte[] checksums = new byte[]
             {
-                CalculateChecksum(result.Tvoc),
-                CalculateChecksum(result.Eco2)
+                CalculateChecksum(result.Eco2),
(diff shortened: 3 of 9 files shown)


src/devices/Sgp30/ChecksumFailedException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Iot.Device.Sgp30
+{
+    /// <summary>
+    /// Exception raised where an SGP30 Checksum Validation fails during a device read operation.
+    /// </summary>
+    public class ChecksumFailedException : Exception
```
Ellerbach, line 8: Is it really necessary to add a specific exception? Will a IOException works?
We recommend not to create new exception except if really needed. For anythin related to communication, we prefer a simple IOExcetion.
On what I've seen in the code, I'm not convienced that this bring a lot of value.

src/devices/Sgp30/README.md
```diff
@@ -0,0 +1,45 @@
+﻿# Sensorion SGP30 TVOC + eCO2 Gas Sensor
+This is a device binding for the Sensorion SGP30 TVOC + eCO2 Gas Sensor.
```
Ellerbach, line 2: you need an extra enter after all titles in Markdown.

src/devices/Sgp30/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Datasheet
+[SGP30](https://cdn.shopify.com/s/files/1/0174/1800/files/Sensirion_Gas_Sensors_SGP30_Datasheet_EN-1148053.pdf)
+
+## Binding Notes
+**NOTE**: Once initialised, this sensor should be queried at least once per second. If more than several seconds pass,
```
Ellerbach, line 8: Is there a minim time you should call the device. Or can I call it every millisecond? Would be great if you have a range to give a range.

src/devices/Sgp30/README.md
```diff
@@ -0,0 +1,45 @@
...
+the initialisation routine should be called again to ensure the device is returning valid data.
+
+```cs
```
Ellerbach, line 11: `csharp` will render better. Github is quite tolerant but an other site or IDe may not be

src/devices/Sgp30/README.md
```diff
@@ -0,0 +1,45 @@
...
+// Setup Device Instance
+I2cDevice sgp30Device = I2cDevice.Create(new I2cConnectionSettings(1, Sgp30.DefaultI2cAddress));
+Sgp30 sgp30 = new Sgp30(sgp30Device);
+
+// Initialise the SGP30 device
+Console.WriteLine("Wait for initialisation (upto 20 seconds).");
```
Ellerbach, line 17: Can you mention that the wait is done in the sensor code? so it's clear that the developer does not have to wait themselves.

src/devices/Sgp30/README.md
```diff
@@ -0,0 +1,45 @@
...
+}
+catch (ChecksumFailedException)
+{
+    Console.WriteLine("Checksum was invalid when initialising SGP30 measurement operation.");
+    throw;
+}
+```
+
+After initialisation, getting sensor readings is straightforward.
+```cs
```
Ellerbach, line 31: csharp and extra line before a code block in Markdown

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+namespace Iot.Device.Sgp30
+{
+    /// <summary>
+    /// Device binding for Sensorion SGP30 TVOC + eCO2 Gas Sensor.
+    /// </summary>
+    public class Sgp30 : IDisposable
+    {
+        // SGP30 Air Quality Sensor I2C Commands
+        // NOTE: The 'Measure Test' command is used for production verification only and is not included here
+        private const ushort SGP30_INITIALISE_AIR_QUALITY = 0x2003;
```
Ellerbach, line 19: Are thos constants you can find in the docs? If yes, fine to leave them in upper case, if not, better to have them with the normal naming convention we are using so `Sgp30InitialiseAirQuality` for example.
Also, you may want o tranform this as an internal enum called Register, see the pattern in other bindings.

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+        public ushort[] GetSerialId()
+        {
+            Span<byte> resultBuffer = stackalloc byte[10];
+            _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[] { (byte)((SGP30_GET_SERIAL_ID & 0xFF00) >> 8), (byte)(SGP30_GET_SERIAL_ID & 0x00FF) }));
+            _i2cDevice.Read(resultBuffer.Slice(1));
+            byte[] resultArray = resultBuffer.ToArray();
+
+            ushort[] serialValues = new ushort[]
+            {
+                (ushort)(resultArray[1] << 8 | resultArray[2]),
```
Ellerbach, line 60: For all those kind of convertion, you may want to use `BinaryPrimitives`: https://docs.microsoft.com/en-us/dotnet/api/system.buffers.binary.binaryprimitives?view=net-5.0
Placing this comment here but you're using this in many places, so please consider everywhere

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+        {
+            _i2cDevice = i2cDevice;
+        }
+
+        /// <summary>
+        /// Gets the Serial ID of the SGP30 sensor.
+        /// </summary>
+        /// <returns>Device Serial ID.</returns>
+        /// <exception cref="ChecksumFailedException">Thrown if checksum validation fails.</exception>
+        public ushort[] GetSerialId()
```
Ellerbach, line 51: Is the serial ID unique? or is it more like a version number?

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+            }
+
+            return serialValues;
+        }
+
+        /// <summary>
+        /// Get the device featureset version.
+        /// </summary>
+        /// <returns>Device Features.</returns>
+        public ushort GetFeaturesetVersion()
```
Ellerbach, line 84: If it's a version, then maybe use Version? And in this case, a property would work perfectly as well

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        /// <summary>
+        /// Gets the Serial ID of the SGP30 sensor.
+        /// </summary>
+        /// <returns>Device Serial ID.</returns>
+        /// <exception cref="ChecksumFailedException">Thrown if checksum validation fails.</exception>
+        public ushort[] GetSerialId()
+        {
+            Span<byte> resultBuffer = stackalloc byte[10];
+            _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[] { (byte)((SGP30_GET_SERIAL_ID & 0xFF00) >> 8), (byte)(SGP30_GET_SERIAL_ID & 0x00FF) }));
```
Ellerbach, line 54: I see you're using multiple times this pattern. Maybe an internal function just for this would make sense?

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        /// <summary>
+        /// Get the device featureset version.
+        /// </summary>
+        /// <returns>Device Features.</returns>
+        public ushort GetFeaturesetVersion()
+        {
+            Span<byte> resultBuffer = stackalloc byte[4];
+            _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[] { (byte)((SGP30_GET_FEATURESET_VERSION & 0xFF00) >> 8), (byte)(SGP30_GET_FEATURESET_VERSION & 0x00FF) }));
+            _i2cDevice.Read(resultBuffer.Slice(1));
```
Ellerbach, line 88: Why do you allocate 4 and slide the first one you never use?
You do this everywhere and it doesn't seems to be needed. In general, avoid allocating what's you'll not use

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+            byte[] resultArray = resultBuffer.ToArray();
```
Ellerbach, line 89: Why to you convert the resultBuffer to an Array? You can use it like an array, so this allocate a new array with no extra value on what I can see.

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+        /// Gets the Serial ID of the SGP30 sensor.
+        /// </summary>
+        /// <returns>Device Serial ID.</returns>
+        /// <exception cref="ChecksumFailedException">Thrown if checksum validation fails.</exception>
+        public ushort[] GetSerialId()
+        {
+            Span<byte> resultBuffer = stackalloc byte[10];
+            _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[] { (byte)((SGP30_GET_SERIAL_ID & 0xFF00) >> 8), (byte)(SGP30_GET_SERIAL_ID & 0x00FF) }));
+            _i2cDevice.Read(resultBuffer.Slice(1));
+            byte[] resultArray = resultBuffer.ToArray();
```
Ellerbach, line 56: Why to you convert the resultBuffer to an Array? You can use it like an array, so this allocate a new array with no extra value on what I can see.

src/devices/Sgp30/Sgp30.cs
Ellerbach: This function will look like that if you apply all the comments. You may consider for the rest of the code as well:

```csharp
Span<byte> resultBuffer = stackalloc byte[3];
WriteRegister(SGP30_GET_FEATURESET_VERSION);
_i2cDevice.Read(resultBuffer);
ushort result = BinaryPrimitives.ReadUInt32BigEndian(resultBuffer.Slice(0,2));
byte checksum = CalculateChecksum(result);
return checksum == resultArray[3] ? result : throw new IOException("Checksum failed");
```

Ellerbach: The internal function will looks like:

```csharp
internal void WriteRegister(ushort reg) => _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[] { (byte)((ushort & 0xFF00) >> 8), (byte)(ushort & 0x00FF) }));
```

You can as well use the BinaryPrimitives:
```csharp
internal void WriteRegister(ushort reg)
{
  Span<byte> toSend = stackalloc byte[2];
  BinaryPrimitives.WriteUInt16BigEndian(toSend, reg);
  _i2cDevice.Write(toSend);
}
```

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        /// <summary>
+        /// Initialises the SGP30 for measurement. NOTE: This call will block for up to 20 seconds.
+        /// </summary>
+        /// <returns>Initial SGP30 measurement result.</returns>
+        public Sgp30Measurement InitialiseMeasurement()
+        {
+            _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[] { (byte)((SGP30_INITIALISE_AIR_QUALITY & 0xFF00) >> 8), (byte)(SGP30_INITIALISE_AIR_QUALITY & 0x00FF) }));
+            Thread.Sleep(10);
```
Ellerbach, line 108: Can you point in the documentation where you have to wait for 10 milliseconds? this helps later on on the debug or changes that may have to be done on this code

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+        {
+            return GetData(SGP30_GET_BASELINE);
+        }
+
+        /// <summary>
+        /// Generalised data read for getting Measurement, Raw Measurement and Baseline data.
+        /// </summary>
+        /// <param name="command">Command to send to the SGP30.</param>
+        /// <returns>Resulting data returned from SGP30.</returns>
+        private Sgp30Measurement GetData(ushort command)
```
Ellerbach, line 167: As mentionned before, would really mke sense to have an internal enum with all the possible commands.

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+        {
+            Span<byte> resultBuffer = stackalloc byte[7];
+            _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[] { (byte)((command & 0xFF00) >> 8), (byte)(command & 0x00FF) }));
+            Thread.Sleep(12);
```
Ellerbach, line 171: Same as previous comment, please add a comment from where all delays are coming from

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        /// <summary>
+        /// Set SGP30 baseline values.
+        /// </summary>
+        /// <param name="tvoc">TVOC Baseline.</param>
+        /// <param name="eco2">eCO2 Baseline.</param>
+        public void SetBaseline(ushort tvoc, ushort eco2)
+        {
+            _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[]
```
Ellerbach, line 202: You may want to use the BinaryPrimitives as well here and allocate a normal Span<byte> using stackalloc rather than a new byte array

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+        {
+            double absoluteHumidity = WeatherHelper.CalculateAbsoluteHumidity(temperature, humidity).GramsPerCubicMeter;
+
+            // First byte contains the whole number part, from 0 to 256 g/m³
+            byte firstByte = (byte)((int)(Math.Floor(absoluteHumidity)) & 0xFF);
+
+            // Second byte contains the decimal part, expressed as the nearest whole number of 256ths of a g/m³
+            byte secondByte = (byte)((int)Math.Floor(((absoluteHumidity - Math.Floor(absoluteHumidity)) / (1 / 256))) & 0xFF);
+
+            _i2cDevice.Write(new ReadOnlySpan<byte>(new byte[]
```
Ellerbach, line 231: same comment as previously

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+                    }
+                }
+            }
+
+            return (byte)(crc & 0xFF);
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
```
Ellerbach, line 276: you should dispose the _i2cDevice here

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+                        crc = (byte)((crc << 1) ^ 0x31);
+                    }
+                    else
+                    {
+                        crc = (byte)(crc << 1);
+                    }
+                }
+            }
+
+            return (byte)(crc & 0xFF);
```
Ellerbach, line 271: as crc is a byte & 0xFF does not bring anything, you can remove it.

src/devices/Sgp30/Sgp30.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        /// <summary>
+        /// Calculates the SGP30 CRC / Checksum for a given input word.
+        /// </summary>
+        /// <param name="data">Data word to checksum.</param>
+        /// <returns>Resulting CRC checksum byte.</returns>
+        private byte CalculateChecksum(ushort data)
+        {
+            byte crc = 0xFF;
+            byte[] bytes = new byte[]
```
Ellerbach, line 249: You may want to use a Span<byte> as well and BinaryPrimitives.

src/devices/Sgp30/Sgp30.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net5.0;netcoreapp2.1</TargetFrameworks>
+    <!--Disabling default items so samples source won't get build by the main library-->
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <LangVersion>9</LangVersion>
```
Ellerbach, line 7: it's by default in the repository, so it's not needed to add

src/devices/Sgp30/Sgp30Measurement.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// SGP30 Gss sensor measurement result.
+    /// </summary>
+    public struct Sgp30Measurement
+    {
+        /// <summary>
+        /// Total VOC result (parts per billion).
+        /// /// </summary>
+        public ushort Tvoc;
```
Ellerbach, line 11: Please use the UnitsNet. And you can adjust the comment giving the usual range of the sensor

src/devices/Sgp30/Sgp30Measurement.cs
```diff
@@ -0,0 +1,18 @@
...
+
+        /// <summary>
+        /// Equivalent CO2 result (parts per million).
+        /// </summary>
+        public ushort Eco2;
```
Ellerbach, line 16: Same

Ellerbach (commented):
Thanks a lot for your sumission, couple of comments about binaryPrimitives, and how to beter handle the registers plus few cosmetic elements and usage of UnitsNet as well.
It's a really interesting sensor!

Ellerbach:
[Triage] @PurpleGranite it's been a month since you last pushed something and there are comments awaiting. Do you think you'll have time to move this PR forward? Any help needed?

PR #1397, opened 2020-12-30
Add possibility to set the alternate pin mode on Raspberry Pis

pgrawehr:
This is a split from PR #1128 

CC: @krwq, @joperezr

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
@@ -157,12 +157,6 @@ namespace System.Device.Gpio.Drivers
         /// <inheritdoc/>
         protected internal override int PinCount => _internalDriver.PinCount;
 
-        /// <summary>
-        /// The Raspberry Pi does support extended pin modes, but the this may depend on the actual driver implementation.
-        /// </summary>
-        /// <remarks>This returns true on Linux, false on Windows</remarks>
-        protected internal override bool ExtendedPinModeSupported => _internalDriver.ExtendedPinModeSupported;
-
         /// <inheritdoc/>
         protected internal override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEventTypes eventTypes, PinChangeEventHandler callback) => _internalDriver.AddCallbackForPinValueChangedEvent(pinNumber, eventTypes, callback);
 
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
@@ -13,6 +13,7 @@ namespace System.Device.Gpio.Drivers
     public class RaspberryPi3Driver : GpioDriver
     {
         private GpioDriver _internalDriver;
+        private RaspberryPi3LinuxDriver? _linuxDriver;
 
         /* private delegates for register Properties */
         private delegate void Set_Register(ulong value);
@@ -24,39 +25,52 @@ namespace System.Device.Gpio.Drivers
         private readonly Get_Register _getClearRegister;
 
         /// <summary>
-        /// Default Gpio mode. Use <see cref="SetPinMode"/> to switch between input and output.
+        /// Used to set the Alternate Pin Mode on Raspberry Pi 3/4.
+        /// The actual pin function for anything other than Input or Output is dependent
+        /// on the pin and can be looked up in the Raspi manual.
         /// </summary>
-        public static readonly ExtendedPinMode GpioMode = new RaspberryPiExtendedPinMode("GPIO", 0);
+        public enum AltMode
+        {
+            /// <summary>
+            /// Gpio mode input
+            /// </summary>
+            Input,
 
-        /// <summary>
-        /// Alternate mode 0 for Raspberry Pi
-        /// </summary>
-        public static readonly ExtendedPinMode Alt0Mode = new RaspberryPiExtendedPinMode("Alt0", 0b100);
+            /// <summary>
+            /// Gpio mode output
+            /// </summary>
+            Output,
 
-        /// <summary>
-        /// Alternate mode 1 for Raspberry Pi
-        /// </summary>
-        public static readonly ExtendedPinMode Alt1Mode = new RaspberryPiExtendedPinMode("Alt1", 0b101);
+            /// <summary>
+            /// Mode ALT0
+            /// </summary>
+            Alt0,
 
-        /// <summary>
-        /// Alternate mode 2 for Raspberry Pi
-        /// </summary>
-        public static readonly ExtendedPinMode Alt2Mode = new RaspberryPiExtendedPinMode("Alt2", 0b110);
+            /// <summary>
+            /// Mode ALT1
+            /// </summary>
+            Alt1,
 
-        /// <summary>
-        /// Alternate mode 3 for Raspberry Pi
-        /// </summary>
-        public static readonly ExtendedPinMode Alt3Mode = new RaspberryPiExtendedPinMode("Alt3", 0b111);
+            /// <summary>
+            /// Mode ALT2
+            /// </summary>
+            Alt2,
 
-        /// <summary>
-        /// Alternate mode 4 for Raspberry Pi
-        /// </summary>
-        public static readonly ExtendedPinMode Alt4Mode = new RaspberryPiExtendedPinMode("Alt4", 0b011);
+            /// <summary>
+            /// Mode ALT3
+            /// </summary>
+            Alt3,
 
-        /// <summary>
-        /// Alternate mode 5 for Raspberry Pi
-        /// </summary>
-        public static readonly ExtendedPinMode Alt5Mode = new RaspberryPiExtendedPinMode("Alt5", 0b010);
+            /// <summary>
+            /// Mode ALT4
+            /// </summary>
+            Alt4,
+
+            /// <summary>
+            /// Mode ALT5
+            /// </summary>
+            Alt5,
+        }
 
         /// <summary>
         /// Creates an instance of the RaspberryPi3Driver.
@@ -66,18 +80,18 @@ namespace System.Device.Gpio.Drivers
         {
             if (Environment.OSVersion.Platform == PlatformID.Unix)
             {
-                RaspberryPi3LinuxDriver? linuxDriver = CreateInternalRaspberryPi3LinuxDriver(out RaspberryBoardInfo boardInfo);
+                _linuxDriver = CreateInternalRaspberryPi3LinuxDriver(out RaspberryBoardInfo boardInfo);
 
-                if (linuxDriver == null)
+                if (_linuxDriver == null)
                 {
                     throw new PlatformNotSupportedException($"Not a supported Raspberry Pi type: " + boardInfo.BoardModel);
                 }
 
-                _setSetRegister = (value) => linuxDriver.SetRegister = value;
-                _setClearRegister = (value) => linuxDriver.ClearRegister = value;
-                _getSetRegister = () => linuxDriver.SetRegister;
-                _getClearRegister = () => linuxDriver.ClearRegister;
-                _internalDriver = linuxDriver;
+                _setSetRegister = (value) => _linuxDriver.SetRegister = value;
+                _setClearRegister = (value) => _linuxDriver.ClearRegister = value;
+                _getSetRegister = () => _linuxDriver.SetRegister;
+                _getClearRegister = () => _linuxDriver.ClearRegister;
+                _internalDriver = _linuxDriver;
             }
             else
             {
@@ -177,16 +191,38 @@ namespace System.Device.Gpio.Drivers
         /// <inheritdoc/>
         protected internal override void Write(int pinNumber, PinValue value) => _internalDriver.Write(pinNumber, value);
 
-        /// <inheritdoc />
-        protected internal override ExtendedPinMode GetExtendedPinMode(int pinNumber)
+        /// <summary>
+        /// Retrieve the current alternate pin mode for a given logical pin.
+        /// This works also with closed pins.
+        /// </summary>
+        /// <param name="pinNumber">Pin number in the logical scheme of the driver</param>
+        /// <returns>Current pin mode</returns>
+        public AltMode GetExtendedPinMode(int pinNumber)
         {
-            return _internalDriver.GetExtendedPinMode(pinNumber);
+            if (_linuxDriver == null)
+            {
+                throw new NotSupportedException("This operation is not supported with the current driver.");
+            }
+
+            return _linuxDriver.GetExtendedPinMode(pinNumber);
         }
 
-        /// <inheritdoc />
-        protected internal override void SetExtendedPinMode(int pinNumber, ExtendedPinMode altMode)
+        /// <summary>
+        /// Set the specified alternate mode for the given pin.
+        /// Check the manual to know what each pin can do.
+        /// </summary>
+        /// <param name="pinNumber">Pin number in the logcal scheme of the driver</param>
+        /// <param name="altPinMode">Alternate mode to set</param>
+        /// <exception cref="NotSupportedException">This mode is not supported by this driver (or by the given pin)</exception>
+        /// <remarks>The method is intended for usage by higher-level abstraction interfaces. User code should be very careful when using this method.</remarks>
+        public void SetExtendedPinMode(int pinNumber, AltMode altPinMode)
         {
-            _internalDriver.SetExtendedPinMode(pinNumber, altMode);
+            if (_linuxDriver == null)
+            {
+                throw new NotSupportedException("This operation is not supported with the current driver.");
+            }
+
+            _linuxDriver.SetExtendedPinMode(pinNumber, altPinMode);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -350,6 +356,96 @@ private void SetInputPullModePi4(int pinNumber, PinMode mode)
             }
         }
 
+        /// <inheritdoc />
+        protected internal override void SetExtendedPinMode(int pinNumber, ExtendedPinMode altMode)
```
krwq, line 360: argument name should be `extendedPinMode` here and in any similar place like this

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPiExtendedPinMode.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
```
krwq, line 1: copyright header

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -350,6 +356,96 @@ private void SetInputPullModePi4(int pinNumber, PinMode mode)
...
+            int shift = (pinNumber % 10) * 3;
+            // Gets a pointer to the register that holds the mode for the pin
+            uint* registerPointer = &_registerViewPointer->GPFSEL[pinNumber / 10];
+            uint register = *registerPointer;
+            // get the three bits of the register
+            register = (register >> shift) & 0b111;
+
+            switch (register)
+            {
+                case 0b000:
```
krwq, line 426: since you have exactly 8 values it's faster to create an array with all entries:
`ExtendedPinMode[] s_pinModes = new[8] { ... };`

and then just use indexer here

krwq: you can put this array in the RaspberryPiExtendedPinMode and have a method in there which resolves int into the instance

src/System.Device.Gpio/System/Device/Gpio/ExtendedPinMode.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Text;
+
+namespace System.Device.Gpio
+{
+    /// <summary>
+    /// Platform-dependent description of an alternate pin mode (i.e. a pin that is used for I2C or SPI transfer).
+    /// Depending on the hardware in use, the use will reflect the effective mode ("PWM", "SPI") or the state of the
+    /// mode multiplexer of the CPU ("ALT0", "ALT1").
+    /// </summary>
+    public class ExtendedPinMode : IEquatable<ExtendedPinMode>
```
krwq, line 15: I'd refrain from doing IEquatable for now unless we have a clear use case.

krwq, line 15: I'd suggest abstract

src/System.Device.Gpio/System/Device/Gpio/ExtendedPinMode.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        private readonly string _modeName;
+
+        /// <summary>
+        /// The mode is not known
+        /// </summary>
+        public static readonly ExtendedPinMode Unknown = new ExtendedPinMode("Unknown");
```
krwq, line 22: for unknown just create an internal class UnknownExtendedPinMode with single instance

src/System.Device.Gpio/System/Device/Gpio/ExtendedPinMode.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        /// <summary>
+        /// Creates a new mode.
+        /// </summary>
+        /// <param name="modeName">Common name of the mode</param>
+        protected ExtendedPinMode(string modeName)
```
krwq, line 28: I'd suggest to do this argumentless and make implementation choose if they want to use string or something else to get a name (i.e. in RaspberryPiExtendedMode you don't really need any string)

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -236,6 +236,35 @@ public virtual void Write(int pinNumber, PinValue value)
             _driver.Write(logicalPinNumber, value);
         }
 
+        /// <summary>
+        /// Returns the currently set pin mode by directly reading the hardware
+        /// </summary>
+        /// <param name="pinNumber">Pin number</param>
+        /// <returns>(Alternate) Pin mode. 0 = Alt0, 1= Alt1... -1 Gpio Input, -2 Gpio Output</returns>
+        public virtual ExtendedPinMode GetAlternatePinMode(int pinNumber)
```
krwq, line 244: GetExtendedPinMode (same comment for Set*)

src/System.Device.Gpio/System/Device/Gpio/ExtendedPinMode.cs
pgrawehr: I need this to be public, since clients (in this case the board class) needs to compare against it.

src/System.Device.Gpio/System/Device/Gpio/ExtendedPinMode.cs
pgrawehr: I need to create the "Unknown" instance below, therefore this is not abstract. Since the ctor is protected, the effect is the same.

src/System.Device.Gpio/System/Device/Gpio/ExtendedPinMode.cs
pgrawehr: I know that the string is not absolutely necessary, but it helps in keeping the things testable. Also, otherwise the instance wouldn't necessarily have any identity at all and comparing them was impossible.

src/System.Device.Gpio/System/Device/Gpio/ExtendedPinMode.cs
pgrawehr: The user needs to do comparisons (i.e. `if (mode == Modes.GPIO)...`). Without implementing IEquatable, things get needlessly complicated.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
pgrawehr: Not sure this is a good idea. For one, this would cause a static initializer depenceny between this class and RaspberryPi3Driver, and then it may be confusing, because the order is not sequential.

src/System.Device.Gpio/System/Device/Gpio/ExtendedPinMode.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (ReferenceEquals(other, null))
+            {
+                return false;
+            }
+
+            return Equals(other);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
```
Ellerbach, line 59: super minor but you may want to use the condense notation => for anything that just do a return or call a single function. Apply to this and the next 3 ones

Ellerbach (approved):
I'm fine with this implementation. Would be good to add a readme somewhere to demonstrate to use and a sample code. That would make things easier. 
I'm thinking of checking if a pin is PWM, changing one it to PWM and creating a hardware PWM then changing things back.

src/System.Device.Gpio/System/Device/Gpio/ExtendedPinMode.cs
krwq: This still can be public, UnknownExtendedPinMode  would be internal and ExtendedPinMode decoupled from `string`

src/System.Device.Gpio/System/Device/Gpio/ExtendedPinMode.cs
krwq: There is no problem with having static readonly members on abstract class

krwq: This will still work considering you return the same reference

pgrawehr: Sure, the problem is not having the static readonly member, but constructing its instance. The `UnknownExtendedPinMode` class would work around this, but I don't really see the benefit of an additional class just to make this abstract.

pgrawehr: You are right, that would work with reference comparison as well. But considering that the class is inheritable, it should be made clear that instances of it are meant to be compared against each other. Maybe some derived implementations will not use statics, but dynamically create the instances (i.e. because the set of possible modes is only known at runtime).

pgrawehr:
> 
> 
> I'm fine with this implementation. Would be good to add a readme somewhere to demonstrate to use and a sample code. That would make things easier.
> I'm thinking of checking if a pin is PWM, changing one it to PWM and creating a hardware PWM then changing things back.

While it is of course possible to use this directly, use of the higher-level board abstraction to access these functions is clearly preferred. There should be documentation about all this there already, iirc.

krwq:
cc: @microhobby (ref: https://github.com/dotnet/iot/pull/1128#discussion_r550335051)

I'm curious on how else can we solve issues like https://github.com/dotnet/iot/issues/874

@microhobby Do you also suggest to not have memory based implementation of GpioDriver? I personally like being able to change mode of the pin in the app rather than having changing global config file and reboot.

microhobby:
> cc: @microhobby (ref: [#1128 (comment)](https://github.com/dotnet/iot/pull/1128#discussion_r550335051))
> 
> I'm curious on how else can we solve issues like #874
>

I don't know how the tests are implemented, but I believe for the RPI we are using  `/dev/mem` and so they are changing the alternative mode to GPIO, overwriting the mode that is defined on device tree and configured by the Kernel on the boot, which should lead to PWMs not working. But I'm speculating I should take a look at the implementation to say for sure. This is something that for example `gpiolib` is concerned about, will we change the state of a pin? Ok, but as soon as the process that changed the state is done or the pin is put into disuse it goes back to its reset state.
 
> @microhobby Do you also suggest to not have memory based implementation of GpioDriver? I personally like being able to change mode of the pin in the app rather than having changing global config file and reboot.

I agree that it is easier to be able to change the pin alternative modes in user space, without the need for a reboot or possible driver reload. But this is not the recommended way. It can be useful for development but should not be encouraged for production. For production the use of devices trees or devices trees overlays are the recommended way. The Kernel knows the hardware by the description given by the device tree, it is the Kernel that knows and will best handle these resources. For this case `gpiolib` will list the pin with an alternate function set as different from GPIO as used by the kernel and will not attempt to change it.

I am also curious if for example we change the alternative function of an i2c pin without it having been previously defined in the device tree, if it will be listed in the i2c sysfs and if after changing the registers by `/dev/mem` the driver will continue to work or whether we will have to write our own i2c implementations. 🤷‍♂️ I've never tried anything like this.

pgrawehr:
> I agree that it is easier to be able to change the pin alternative modes in user space, without the need for a reboot or possible driver reload. But this is not the recommended way. It can be useful for development but should not be encouraged for production. For production the use of devices trees or devices trees overlays are the recommended way. The Kernel knows the hardware by the description given by the device tree, it is the Kernel that knows and will best handle these resources. For this case `gpiolib` will list the pin with an alternate function set as different from GPIO as used by the kernel and will not attempt to change it.
> 
> I am also curious if for example we change the alternative function of an i2c pin without it having been previously defined in the device tree, if it will be listed in the i2c sysfs and if after changing the registers by `/dev/mem` the driver will continue to work or whether we will have to write our own i2c implementations. 🤷‍♂️ I've never tried anything like this.

I haven't tested it with I2C, but from my tests with PWM I don't expect anything "bad" to happen. I'm actually expecting that you could just enable all drivers you might possibly need in `boot/config.txt` and then switch at runtime to whichever you need. Of course, if you enable i.e. UART and I2C on pins 0&1, you'll need an external multiplexer to actually use them both.

Just tested this scenario. I added these two lines to `boot/config.txt`:
```
dtoverlay=i2c0 # uses pins 0 & 1, mode ALT0
dtoverlay=uart2 # also uses pins 0 & 1, mode ALT4
```

So now I could switch the pins between modes 0 and 4 to change the modes, the kernel was all fine with it it seems, and `i2cdetect -y 0` would work mostly fine. There was, however, some interference to be observed. When I changed the pins to ALT0 and ran that command, it would print all addresses as used despite there's nothing connected, really. That seems to be more of a hardware issue than a driver problem though and is out of scope here.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -23,6 +23,41 @@ public class RaspberryPi3Driver : GpioDriver
         private readonly Set_Register _setClearRegister;
         private readonly Get_Register _getClearRegister;
 
+        /// <summary>
+        /// Default Gpio mode. Use <see cref="SetPinMode"/> to switch between input and output.
+        /// </summary>
+        public static readonly ExtendedPinMode GpioMode = new RaspberryPiExtendedPinMode("GPIO", 0);
```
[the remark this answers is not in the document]
krwq: - make them public rather than protected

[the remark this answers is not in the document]
pgrawehr: @krwq, @joperezr Updated PR according to our discussion. I did have to add a property to query the support, because it doesn't currently work on Windows, and I don't necessarily want to rely on exceptions in this case. (And Windows 10 IOT support on the Raspi is dying out it seems, anyway)

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -97,6 +157,12 @@ private static GpioDriver CreateWindows10GpioDriver()
         /// <inheritdoc/>
         protected internal override int PinCount => _internalDriver.PinCount;
 
+        /// <summary>
+        /// The Raspberry Pi does support extended pin modes, but the this may depend on the actual driver implementation.
+        /// </summary>
+        /// <remarks>This returns true on Linux, false on Windows</remarks>
+        protected internal override bool ExtendedPinModeSupported => _internalDriver.ExtendedPinModeSupported;
```
krwq, line 164: This should be public as well, otherwise user can't check it

krwq: actually you akready have AlternatePinModeSettingSupported, this should be removed in that case

pgrawehr: True, this was just forgotten.

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -70,6 +125,11 @@ internal RaspberryPi3Driver(RaspberryPi3LinuxDriver linuxDriver)
             }
         }
 
+        /// <summary>
+        /// True if the driver supports <see cref="SetAlternatePinMode"/> and <see cref="GetAlternatePinMode"/>.
+        /// </summary>
+        public bool AlternatePinModeSettingSupported => _linuxDriver != null;
```
krwq, line 131: please only leave this one

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
krwq: I'm still seeing this

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -23,6 +24,59 @@ public class RaspberryPi3Driver : GpioDriver
...
+        /// Used to set the Alternate Pin Mode on Raspberry Pi 3/4.
+        /// The actual pin function for anything other than Input or Output is dependent
+        /// on the pin and can be looked up in the Raspi manual.
+        /// </summary>
+        public enum AltMode
+        {
+            /// <summary>
+            /// The mode is unknown
+            /// </summary>
+            Unknown,
```
krwq, line 37: Since you never return this or allow this to set, perhaps consider removing this value

krwq:
@pgrawehr still seeing ExtendedPinModeSupported.

krwq (approved):
LGTM after 1-2 fixes:
- consider removing AltMode.Unknown since you never use it
- remove ExtendedPinModeSupported

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
pgrawehr: I found that it will be helpful for the client, therefore I added it.

krwq:
@microhobby just FYI we have changed the implementation to limit this functionality to only the driver which already uses direct memory access, we do not plan to have a generic functionality like was initially proposed.

krwq (approved):

PR #1398, opened 2020-12-30
Dc motor start stop

### LINKED ISSUE #1207: DCMotor - safe stop
In #1186 I wanted to add ```Start()``` and ```Stop()``` to disable and enable pwm. 
Of course their effect would be the same as for the ```Speed = 0``` and ```Speed = <not a zero>```, but when the motor control would take place in separate threads, ```Speed = 0``` may not be sufficient. A good example would be the use of limit switches (my motor rotates the screw that moves the platform, which has limited movement, that is why I need limit switches). If we control the motor in a simple loop, e.g.
```csharp
for (int i = 0; i < 360; i++)
{
    speedSet = Math.Sin(Math.PI * i / 180.0);
    System.Console.WriteLine($"{i}. Actual motor setting is: {speedSet}");
    motor.Speed = speedSet;
    Thread.Sleep(100);
}
```
and use ```RegisterCallbackForPinValueChangedEvent``` for limit switch with ```Speed = 0``` will not break the loop so the motor will run again in the next iteration. You can check the condition of the limit switches in each control loop, but the use of events seemed safer to me. You only need to start monitoring events once and you don't have to worry about breaking something by accident.
I understand that checking the status of limit switches will have a similar effect, but you can accidentally forget about it when you try something new, and I just wanted to add additional protection, which of course is not obliged to use, as pwm would be enabled by default, and you can still only use Speed to control the engine.

wsad4ryba:
Fixes #1207
Adds a wrapper that allows the engine to stop or start, regardless of the Speed value.

### DIFF
--- a/src/devices/DCMotor/DCMotorWithStartStop.cs
+++ b/src/devices/DCMotor/DCMotorWithStartStop.cs
@@ -1,18 +1,12 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-using System;
-using System.Device.Gpio;
-using System.Device.Pwm;
-using System.Device.Pwm.Drivers;
 
 namespace Iot.Device.DCMotor
 {
     /// <summary>
     /// Direct current (DC) motor with Start/Stop
     /// </summary>
-    public class DCMotorWithStartStop : DCMotor
+    public class DCMotorStartStop : DCMotor
     {
         private DCMotor _inner;
         private bool _stopped = false;
@@ -22,7 +16,7 @@ namespace Iot.Device.DCMotor
         /// Constructs instance with added Start() and Stop() as additional protection
         /// </summary>
         /// <param name="innerMotor">Crate DCMotor instance</param>
-        public DCMotorWithStartStop(DCMotor innerMotor)
+        public DCMotorStartStop(DCMotor innerMotor)
             : base(null, false)
         {
             _inner = innerMotor;
@@ -60,6 +54,25 @@ namespace Iot.Device.DCMotor
                 if (!_stopped)
                 {
                     _inner.Speed = _speed;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Get or Set motor status.
+        /// </summary>
+        public bool Enabled
+        {
+            get => !_stopped;
+            set
+            {
+                if (value)
+                {
+                    Start();
+                }
+                else
+                {
+                    Stop();
                 }
             }
         }
--- a/src/devices/DCMotor/DCMotorWithStartStop.cs
+++ b/src/devices/DCMotor/DCMotorWithStartStop.cs
@@ -1,0 +1,67 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.Gpio;
+using System.Device.Pwm;
+using System.Device.Pwm.Drivers;
+
+namespace Iot.Device.DCMotor
+{
+    /// <summary>
+    /// Direct current (DC) motor with Start/Stop
+    /// </summary>
+    public class DCMotorWithStartStop : DCMotor
+    {
+        private DCMotor _inner;
+        private bool _stopped = false;
+        private double _speed;
+
+        /// <summary>
+        /// Constructs instance with added Start() and Stop() as additional protection
+        /// </summary>
+        /// <param name="innerMotor">Crate DCMotor instance</param>
+        public DCMotorWithStartStop(DCMotor innerMotor)
+            : base(null, false)
+        {
+            _inner = innerMotor;
+            _speed = innerMotor.Speed;
+        }
+
+        /// <summary>
+        /// Enable motor operation.
+        /// </summary>
+        public void Start()
+        {
+            _stopped = false;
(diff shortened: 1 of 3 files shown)


src/devices/DCMotor/DCMotorWithStartStop.cs
```diff
@@ -0,0 +1,67 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Device.Gpio;
```
Ellerbach, line 6: not sure you need all the System.Device namespaces here

src/devices/DCMotor/DCMotorWithStartStop.cs
```diff
@@ -0,0 +1,67 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
Ellerbach, line 3: The new header is only 2 lines vs 3 lines in the past

src/devices/DCMotor/DCMotorWithStartStop.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// Disable motor operation.
+        /// </summary>
+        public void Stop()
+        {
+            _stopped = true;
+            _inner.Speed = 0.0;
+        }
+
```
Ellerbach, line 49: Maybe you want as well a property called `Enabled` which would be like this:
```csharp
public bool Enabled
{
  get => !_stopped;
  set => value ? Start() : Stop();
}
```

src/devices/DCMotor/DCMotorWithStartStop.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Device.Gpio;
+using System.Device.Pwm;
+using System.Device.Pwm.Drivers;
+
+namespace Iot.Device.DCMotor
+{
+    /// <summary>
+    /// Direct current (DC) motor with Start/Stop
+    /// </summary>
+    public class DCMotorWithStartStop : DCMotor
```
Ellerbach, line 15: For the name, I'm not super sure. It feels like a bit long at the same time it describe swell what this does

Ellerbach (commented):
Looks good to me. I understand some may need this function. I would add a property as well called `Enabled`. We do have quite a few bindings with this pattern Start()/Stop()/Enabled.

src/devices/DCMotor/DCMotorWithStartStop.cs
wsad4ryba: you're right

src/devices/DCMotor/DCMotorWithStartStop.cs
wsad4ryba: corrected

src/devices/DCMotor/DCMotorWithStartStop.cs
wsad4ryba: I got the error "Only assignment, call, increment, decrement, await, and new object expressions can be used as a statement" so I changed it to:
```csharp
public bool Enabled
{
    get => !_stopped;
    set
    {
        if (value)
        {
            Start();
        }
        else
        {
            Stop();
        }
    }
}
```

src/devices/DCMotor/DCMotorWithStartStop.cs
wsad4ryba: So maybe DCMotorStartStop?

src/devices/DCMotor/DCMotorWithStartStop.cs
Ellerbach: You're right! Did that without testing.

src/devices/DCMotor/DCMotorWithStartStop.cs
Ellerbach: Let's see once @krwq or someone else will do a review.

src/devices/DCMotor/DCMotorWithStartStop.cs
```diff
@@ -0,0 +1,80 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.DCMotor
+{
+    /// <summary>
+    /// Direct current (DC) motor with Start/Stop
+    /// </summary>
+    public class DCMotorStartStop : DCMotor
```
krwq, line 9: please add a sample somewhere in the bottom of the sample/README.md, otherwise this API might not be used too much

krwq (approved):

src/devices/DCMotor/DCMotorWithStartStop.cs
```diff
@@ -0,0 +1,80 @@
...
+            get => !_stopped;
+            set
+            {
+                if (value)
+                {
+                    Start();
+                }
+                else
+                {
+                    Stop();
```
krwq, line 75: You should also implement Dispose and Dispose of the inner motor (with shouldDispose in the constructor set to true by default)

src/devices/DCMotor/DCMotorWithStartStop.cs
krwq: I'd rather have long descriptive name. The documentation is most important as I expect people to copy & paste samples

krwq (approved):

PR #1406, opened 2021-01-13
Fix unreliable code branch

hackey9:
I fix an unreliable code branch in example app.

Prev behaviour:
- I catch exception if enter 'Q'
- Return from function if enter any

Observe: 
```
while (true)
{
    string? command = Console.ReadLine()?.ToLower();
    if (command?[0] is not 'q') // 'c' or <double> will return
    {
        return;
    }
    if (command[0] == 'c') // command[0] is 'q'
    {
        CalibrateServo(servo);
        PrintServoDemoHelp();
    }
    else
    {
        double value = double.Parse(command); // Catch an exception because parse 'q'
        servo.WriteAngle(value);
        Console.WriteLine($"Angle set to {value}. PWM duty cycle = {pca9685.GetDutyCycle(channel)}");
    }
}
```

Solution:
- Remove 'not'

### DIFF
--- a/src/devices/Pca9685/samples/Program.cs
+++ b/src/devices/Pca9685/samples/Program.cs
@@ -115,7 +115,12 @@ void ServoDemo(Pca9685 pca9685, int channel)
     while (true)
     {
         string? command = Console.ReadLine()?.ToLower();
-        if (command?[0] is 'q')
+        if (command is null)
+        {
+            return;
+        }
+
+        if (command[0] is 'q')
         {
             return;
         }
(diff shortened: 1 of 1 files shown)


hackey9:
Dont bite me, this is my first PR in my life😅

krwq:
@hackey9 good find! LGTM, you will need to do null check for `command`, otherwise you will get the nullability warnings. I suspect we introduced this errors when converting code base to nullable.

src/devices/Pca9685/samples/Program.cs
```diff
@@ -115,7 +115,7 @@ void ServoDemo(Pca9685 pca9685, int channel)
     while (true)
     {
         string? command = Console.ReadLine()?.ToLower();
-        if (command?[0] is not 'q')
+        if (command?[0] is 'q')
```
krwq, line 118:
```suggestion
        if (command is null)
        {
            return;
        }

        if (command[0] is 'q')
```

krwq (approved):

krwq:
failures are unrelated, merging

krwq:
thanks!

PR #1420, opened 2021-01-26
Binding for the Qmc5883l 3-Axis sensor

Speedo69:
Binding for the Qmc5883l 3-Axis sensor.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1420) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1420) before we can accept your contribution.<br/><br/>:x: Speedo69 [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1420)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1420) it.</sub>

src/devices/Qmc5883l/FieldRange.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// can be used.The field range goes hand in hand with the sensitivity of the magnetic sensor. The lowest field range
+    /// has the highest sensitivity, therefore, higher resolution.
+    /// </summary>
+    [Flags]
+    public enum FieldRange : byte
+    {
+       /// <summary>
+       /// Field range of 2 Gauss
+       /// </summary>
+        GAUSS_2 = 0x00,
```
Ellerbach, line 17: For public enums, we use a naming convention that is Camel oriented.
So, this enum will be: `Gauss2`, the next one `Gauss8`
We do accept other conventions for internal, register based enums to be close to the documentation and help in debugging. 
This comment applies to all the enums in the same condition.

src/devices/Qmc5883l/FieldRange.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
```
Ellerbach, line 1: You are missing the header in many files. Please add it:
```csharp
// Licensed to the .NET Foundation under one or more agreements.

```

Please add it in all the cs files.

src/devices/Qmc5883l/Mode.cs
```diff
@@ -0,0 +1,27 @@
...
+    [Flags]
+    public enum Mode : byte
+    {
+        /// <summary>
+        /// Standby mode is the default state of QMC5883L upon POR and soft reset, only few function blocks are activated
+        /// in this mode, which keeps power consumption as low as possible. In this state, register values are hold on by an
+        /// ultra-low power LDO, I2C interface can be woken up by reading or writing any registers.There is no
+        /// magnetometer measurement in the Standby mode.Internal clocking is also halted.
+      /// </summary>
+      STAND_BY = 0x00,
```
Ellerbach, line 17: This one would be `Standby` as it's a normal name (not StandBy).

src/devices/Qmc5883l/OutputRate.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// 100Hz and 200Hz.For most of compassing applications, we recommend 10 Hz for low power consumption.For
+    /// gaming, the high update rate such as 100Hz or 200Hz can be used.
+    /// </summary>
+    [Flags]
+    public enum OutputRate : byte
+    {
+        /// <summary>
+        /// Output rate of 10 Hz
+        /// </summary>
+        RATE_10HZ = 0x00,
```
Ellerbach, line 16: For those ones, something like `Rate10Hz` is a good fit

src/devices/Qmc5883l/Oversampling.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// Larger OSR valueleads to smaller filter bandwidth, less in-band noise and higher power consumption.It could be used to reach a
+    /// good balance between noise and power. Four over sample ratio can be selected, 64, 128, 256 or 512.
+    /// </summary>
+    [Flags]
+    public enum Oversampling : byte
+    {
+        /// <summary>
+        /// Over sample rate of 512
+        /// </summary>
+        OS512 = 0x00,
```
Ellerbach, line 16: We tend not to use abbreviation, so for those one, `Rate512` may work as the enum name already contains `Oversampling`, so it makes it easy to understand `Oversampling.Rate512`

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,145 @@
...
+    {
+        /// <summary>
+        /// QMC5883L Default I2C Address
+        /// </summary>
+        public const byte DefaultI2cAddress = 0x0D;
+
+        /// <summary>
+        /// QMC5883L Chip id.
+        /// </summary>
+        public const byte Chip_Id = 0xff;
```
Ellerbach, line 26: Same with the const and anything public and private, the same naming convention applies, so `ChipId` or `ChipIdentifier` as we prefer not to use abbreviation

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <summary>
+        /// Calculate heading
+        /// </summary>
+        /// <param name="vector">QMC5883L Direction Vector</param>
+        /// <returns>Heading (DEG)</returns>
+        private static double VectorToHeading(Vector3 vector)
+        {
+            double deg = Math.Atan2(vector.Y, vector.X) * 180 / Math.PI;
+
+            if (deg < 0)
```
Ellerbach, line 130: you simplify by `deg = deg < 0 ? deg + 360 : deg;` then `return deg;`

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,145 @@
...
+                               BinaryPrimitives.ReadInt16BigEndian(yRead),
+                               BinaryPrimitives.ReadInt16BigEndian(zRead));
+        }
+
+        /// <summary>
+        /// Calculate heading
+        /// </summary>
+        /// <param name="vector">QMC5883L Direction Vector</param>
+        /// <returns>Heading (DEG)</returns>
+        private static double VectorToHeading(Vector3 vector)
```
Ellerbach, line 126: We're using UnitsNet for anything related to units. See https://github.com/dotnet/iot/tree/master/src/devices#using-units
You should use it for angles. So you the user then can decide what unit of angle they want.

src/devices/Qmc5883l/README.md
```diff
@@ -0,0 +1,34 @@
+﻿# Qmc5883l
+
+## Summary
+QMC5883l is a 3-Axis Magnetic Sensor commonly used as a substitute for the QMC5883l. Easiest way to tell them apart is by checking the I2c address. HMC5883L = 0x1E while QMC5883L = 0x0D
```
Ellerbach, line 4: in Markdown, you need an extra enter after any title, same for all of them.

src/devices/Qmc5883l/Qmc5883l.csproj
```diff
@@ -0,0 +1,17 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>net5.0;netcoreapp2.1</TargetFrameworks>
+    <!--Disabling default items so samples source won't get build by the main library-->
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <LangVersion>9</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <None Include="Datasheet-QMC5883L-1.0 .pdf" />
```
Ellerbach, line 12: I guess, this is not needed as you're not checking it and should not check this file.

src/devices/Qmc5883l/Registers.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Iot.Device.Qmc5883l
+{
+    /// <summary>
+    /// registeries used by QMC5883L
+    /// </summary>
+    internal enum Registers : byte
+    {
+        QMC_CONFIG_REG_1_ADDR = 0x09,
```
Ellerbach, line 8: And here, the naming are ok because it's internal register, from the documentation.

src/devices/Qmc5883l/samples/Qmc5883l.Sample.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
Ellerbach, line 1: file name should be program.cs

src/devices/Qmc5883l/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+    while (true)
+    {
+        // If you aren't using an interrupt PIN, then always make sure that the data is ready.
+        if (sensor.IsReady())
+        {
+            // read heading
+            Console.WriteLine($"Heading: {sensor.Heading.ToString("0.00")} °");
+            // read vectors
+            Console.WriteLine(sensor.DirectionVector.X + " : " + sensor.DirectionVector.Y + " : " + sensor.DirectionVector.Z);
```
Ellerbach, line 27: This is not the right way of accessing the property. It will read 3 time the sensor. Proper way is:
```csharp
// Store the value into a variable
var direction = sensor.DirectionVector
Console.WriteLine(direction .X + " : " + direction .Y + " : " + direction .Z);
```

src/devices/Qmc5883l/README.md
```diff
@@ -0,0 +1,34 @@
...
+    //It is possible to also change the sensor mode after initialization. (Ex. Set the mode to STAND_BY to save power)
+    sensor.SetMode(outputRate: OutputRate.RATE_200HZ, fieldRange: FieldRange.GAUSS_8, oversampling: Oversampling.OS256);
+
+    while (true)
+    {
+        // If you aren't using an interrupt PIN, then always make sure that the data is ready.
+        if (sensor.IsReady())
+        {
+            // read heading
+            Console.WriteLine($"Heading: {sensor.Heading.ToString("0.00")} °");
```
Ellerbach, line 25: You will have to adjust this once you'll use UnitsNet

src/devices/Qmc5883l/samples/Qmc5883l.Sample.cs
```diff
@@ -0,0 +1,29 @@
...
+
+I2cConnectionSettings settings = new(1, Qmc5883l.DefaultI2cAddress);
+using I2cDevice device = I2cDevice.Create(settings);
+using (Qmc5883l sensor = new(device))
+{
+    sensor.SetMode(outputRate: OutputRate.RATE_200HZ, fieldRange: FieldRange.GAUSS_8, oversampling: Oversampling.OS256);
+
+    while (true)
+    {
+        // If you aren't using an interrupt PIN, then always make sure that the data is ready.
```
Ellerbach, line 17: See the comment in the README.md, same apply of course in the code here

src/devices/Qmc5883l/samples/README.md
```diff
@@ -0,0 +1,4 @@
+# TODO: This needs to be determined
```
Ellerbach, line 1: What kind of help do you need? :-D

Ellerbach (commented):
Thanks a lot for this new binding! Some comments, mainly about naming conventions, using of UnitsNet, missing headers.

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// </summary>
+        public const byte Chip_Id = 0xff;
+
+        private I2cDevice _i2cDevice;
+
+        /// <summary>
+        /// QMC5883L Direction Vector
+        /// </summary>
+        [Telemetry]
+        public Vector3 DirectionVector => ReadDirectionVector();
```
Ellerbach, line 34: Why not just `Direction`?

Speedo69:
Thank you for a fast reply. I have taken into account all of your suggestions and improved my code to fit the standard. ❤️

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <summary>
+        /// QMC5883L Direction Vector
+        /// </summary>
+        [Telemetry]
+        public Vector3 Direction => ReadDirectionVector();
+
+        /// <summary>
+        /// QMC5883L Heading (DEG)
+        /// </summary>
+        public Angle Heading => VectorExtentsion.GetHeading(ReadDirectionVector());
```
pgrawehr, line 34: I would consider changing these to methods. Properties should not be expensive or return different values if called multiple times. You might even want to change them to the Try... pattern. So this would be `bool TryGetHeading(out Angle heading);`

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        /// <summary>
+        /// Reads status register to determine if data is avaliable.
+        /// </summary>
+        /// <returns>1 == read is ready. 0 == read is not ready</returns>
+        public bool IsReady()
+        {
+            _i2cDevice.WriteByte((byte)Registers.QMC_STATUS_REG_ADDR);
+            byte status = _i2cDevice.ReadByte();
+            Console.WriteLine(status);
```
pgrawehr, line 90: Do not use the console in bindings. We'll (hopefully soon) be providing some generic logging infrastructure for such messages.

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <summary>
+        /// Sets the sensors mode with additional parameters.
+        /// </summary>
+        /// <param name="mode">Sensor mode</param>
+        /// <param name="interrupt">Controlls the Interrupt PIN</param>
+        /// <param name="rollPointer">When the point roll-over function is enabled, the I2C data pointer automatically rolls between 00H ~ 06H</param>
+        /// <param name="outputRate">Typical Data Output Rate (Hz)</param>
+        /// <param name="oversampling">Bandwidth of an internal digital filter</param>
+        /// <param name="fieldRange">The field range goes hand in hand with the sensitivity of the magnetic sensor</param>
+        public void SetMode(Mode mode = Mode.Continuous,
```
Ellerbach, line 53: So as all those modes are available, why not propose a property for each one?
This will be super easy to adjust only one mode for the developer using this biding. With the specific decoration, that will allow as well to adjust the mode easily. You still can keep this function public. You'll need to call it anyway to set up any change in the modes.

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,123 @@
...
+            };
+            _i2cDevice.Write(interruptCommand);
+            _i2cDevice.Write(resetCommand);
+            _i2cDevice.Write(setMainRegisteryCommand);
+        }
+
+        /// <summary>
+        /// Reads status register to determine if data is avaliable.
+        /// </summary>
+        /// <returns>1 == read is ready. 0 == read is not ready</returns>
```
Ellerbach, line 85: I think it returns True of it's ready. The description you placed here is what happened internally.

src/devices/Qmc5883l/Status.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System;
+
+namespace Iot.Device.Qmc5883l
+{
+    /// <summary>
+    /// Status flags for QMC5883L first register
+    /// </summary>
+    [Flags]
+    public enum Status : byte
```
Ellerbach, line 12: It doesn't seem to be needed to be public. I guess internal is enough in this case as if I read correctly, you're not using it outside of the class

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,123 @@
...
+            return (status & 0x01) == 1;
+        }
+
+        /// <summary>
+        /// Read raw data from QMC5883L
+        /// </summary>
+        /// <returns>Raw Data</returns>
+        private Vector3 ReadDirectionVector()
+        {
+            Span<byte> xRead = stackalloc byte[2];
```
pgrawehr, line 100: This can be called before SetMode() was used. Either call `SetMode()` in the constructor or throw an exception here if the device is not initialized (or not ready)

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <summary>
+        /// Reads status register to determine if data is avaliable.
+        /// </summary>
+        /// <returns>1 == read is ready. 0 == read is not ready</returns>
+        public bool IsReady()
+        {
+            _i2cDevice.WriteByte((byte)Registers.QMC_STATUS_REG_ADDR);
+            byte status = _i2cDevice.ReadByte();
+            Console.WriteLine(status);
+            return (status & 0x01) == 1;
```
Ellerbach, line 91: as you have a nice enum, rather than 0x01, you may want to use `(byte)Status.DRDY` instead

src/devices/Qmc5883l/VectorExtentsion.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Numerics;
+
+using UnitsNet;
+
+namespace Iot.Device.Qmc5883l
+{
+    /// <summary>
+    /// Extentsions class for getting a heading from X and Y Vectors.
+    /// </summary>
+    public static class VectorExtentsion
```
pgrawehr, line 14: This has a typo

src/devices/Qmc5883l/samples/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    sensor.SetMode(outputRate: OutputRate.Rate200Hz, fieldRange: FieldRange.Gauss8, oversampling: Oversampling.Rate512);
+    while (true)
+    {
+        // If you aren't using an interrupt PIN, then always make sure that the data is ready.
+        if (sensor.IsReady())
+        {
+            // Make sure not to access sensor.Direction to many times, as it will read the sensor every time you do it.
+            // Store the value in a variable instead.
+            // Here is an example of what not to do :D
```
Ellerbach, line 21: :-D and please reflect this in the README as well. You still have the "old" code.

Ellerbach (commented):
Few more comments, starts to look really good all up!
Key question is on offering properties to get and set all modes. I think it can be a super nice addition.

Speedo69:
Thank you again for a quick response. I have again taken into account all of your suggestions and changed my code accordingly.

Ellerbach (approved):
Looks good to me, minor details and we should be good to merge. Thanks again!

src/devices/Hmc5883l/Hmc5883l.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>net5.0;netcoreapp2.1</TargetFrameworks>
     <EnableDefaultItems>false</EnableDefaultItems>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
```
krwq, line 6: why is this needed?

src/devices/Qmc5883l/FieldRange.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System;
+
+namespace Iot.Device.Qmc5883l
+{
+    /// <summary>
+    /// Field ranges of the magnetic sensor can be selected through the register RNG. The full scale field range is
+    /// determined by the application environments.For magnetic clear environment, low field range such as +/- 2gauss
+    /// can be used.The field range goes hand in hand with the sensitivity of the magnetic sensor. The lowest field range
+    /// has the highest sensitivity, therefore, higher resolution.
```
krwq, line 12:
```suggestion
    /// Field ranges of the magnetic sensor can be selected through the register RNG. The full scale field range is
    /// determined by the application environments.For magnetic clear environment, low field range such as +/- 2gauss
    /// can be used.The field range goes hand in hand with the sensitivity of the magnetic sensor. The lowest field range
    /// has the highest sensitivity, therefore, higher resolution.
```

RNG=Random Number Generator? Can you expand that, I do not see relationship? Is the information about any registers useful for most of the users?

src/devices/Qmc5883l/Interrupt.cs
```diff
@@ -0,0 +1,23 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Qmc5883l
+{
+    /// <summary>
+    /// Interrupt enabling is controlled by register INT_ENB in control register 2. Once the interrupt is enabled, it will flag
+    /// when new data is in Data Output Registers.
+    /// INT_ENB: “0”: enable interrupt PIN, “1”: disable interrupt PIN
+    /// </summary>
+    public enum Interrupt : byte
```
krwq, line 11: could you use bool instead and change fields into InterruptEnabled?

src/devices/Qmc5883l/Interrupt.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Qmc5883l
+{
+    /// <summary>
+    /// Interrupt enabling is controlled by register INT_ENB in control register 2. Once the interrupt is enabled, it will flag
```
krwq, line 7: information about register is only useful for developers of the binding not for consumer of this class. I'd recommend removing register specific information

src/devices/Qmc5883l/Mode.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// magnetometer measurement in the Standby mode.Internal clocking is also halted.
+      /// </summary>
+      Standby = 0x00,
+
+        /// <summary>
+        /// During the continuous-measurement mode (mode bits= 01), the magnetic sensor continuously makes
+        /// measurements and places measured data in data output registers. The field range (or sensitivity) and data output
+        /// rate registers are also located in the control register (09H), they should be set up properly for your applications in
+        /// the continuous-measurement mode.
+        /// </summary>
```
krwq, line 26: indentation of docs is off

src/devices/Qmc5883l/Mode.cs
```diff
@@ -0,0 +1,29 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+using System;
+
+namespace Iot.Device.Qmc5883l
+{
+    /// <summary>
+    /// Modes of Qmc5883l
+    /// </summary>
+    [Flags]
+    public enum Mode : byte
```
krwq, line 11: could you convert this into bool and call it just `StandbyMode` or `LowPowerMode`?

src/devices/Qmc5883l/OutputRate.cs
```diff
@@ -0,0 +1,36 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Iot.Device.Qmc5883l
+{
+    /// <summary>
+    /// Output data rate is controlled by ODR registers. Four data update frequencies can be selected: 10Hz, 50Hz,
+    /// 100Hz and 200Hz.For most of compassing applications, we recommend 10 Hz for low power consumption.For
+    /// gaming, the high update rate such as 100Hz or 200Hz can be used.
```
krwq, line 11: - you should avoid register specific information
- extra info about recommendations should be in remarks sections
- summary should just contain generic info like `Output data rate of the device` or `Number of samples device can produce per second`
- use spaces between dot ending the sentence and first word

src/devices/Qmc5883l/Oversampling.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Iot.Device.Qmc5883l
+{
+    /// <summary>
+    /// Over sample Rate (OSR) registers are used to control bandwidth of an internal digital filter.
```
krwq, line 9: ditto

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// </summary>
+        public const byte DefaultI2cAddress = 0x0D;
+        private I2cDevice _i2cDevice;
+        private bool _isDeviceInitialized = false;
+
+        /// <summary>
+        /// Enables or disables the Interrupt PIN.
+        /// </summary>
+        /// <value>Disabled by default.</value>
+        public Interrupt Interrupt { get; set; } = Interrupt.Disable;
```
krwq, line 31: `bool InterruptPinEnabled { get { ... } set { ... } }`

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// When the point roll-over function is enabled, the I2C data pointer automatically rolls between 00H ~ 06H.
+        /// </summary>
+        /// <value>Disabled by default.</value>
+        public RollPointer RollPointer { get; set; } = RollPointer.Disable;
```
krwq, line 37: this doesn't seem to be updating anything in the device, most of the devices are updating registers here. You are requiring people to call SetMode instead, most people won't remember to do it and will likely cause them to work differently than they wanted

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// </summary>
+        public Qmc5883l(I2cDevice i2cDevice)
+        {
+            _i2cDevice = i2cDevice ?? throw new ArgumentNullException(nameof(i2cDevice));
+        }
+
+        /// <summary>
+        /// Sets the sensors mode.
+        /// </summary>
+        public void SetMode()
```
krwq, line 86: this should be an implementation detail

krwq: i.e. call from the setter of each property. I recommend to update/read a flags related to a mode you're changing without any extra variables directly in the getter and setter

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,150 @@
...
+            _i2cDevice.Write(resetCommand);
+            _i2cDevice.Write(setMainRegisteryCommand);
+            _isDeviceInitialized = true;
+        }
+
+        /// <summary>
+        /// Reads status register to determine if data is avaliable.
+        /// </summary>
+        /// <returns>State of the register as a boolean</returns>
+        public bool IsReady()
```
krwq, line 114: You should rename to `IsDataReady`. My first thought was that this was about the device being ready for usage

src/devices/Qmc5883l/README.md
```diff
@@ -0,0 +1,52 @@
...
+            // Calculates the heading from a previously stored value.
+            Console.WriteLine($"Heading: {direction.GetHeading()}");
+        }
+
+        // wait for a second
+        Thread.Sleep(1000);
+    }
+}
+
+
```
krwq, line 51:
```suggestion
}
```

src/devices/Qmc5883l/Qmc5883l.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// Sensors operation mode.
+        /// </summary>
+        /// <value>In Standby mode by default. Make sure to change it to Continuous before trying to get any data!</value>
+        public Mode DeviceMode { get; set; } = Mode.Standby;
+
+        /// <summary>
+        /// QMC5883L Direction Vector
+        /// </summary>
+        [Telemetry]
+        public Vector3 GetDirection() => _isDeviceInitialized ? ReadDirectionVector() : throw new IOException("Sensor has not yet been initialized. Please call SetMode() to initialize it.");
```
krwq, line 67: the `_isDeviceInitialized` should not be required. Do required initialization in the constructor

src/devices/Qmc5883l/README.md
krwq: (remove extra enters at the end - github doesn't seem to display the suggested change correctly)

src/devices/Qmc5883l/RollPointer.cs
```diff
@@ -0,0 +1,23 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Qmc5883l
+{
+    /// <summary>
+    /// Pointer roll-over function is controlled by ROL_PNT register. When the point roll-over function is enabled, the I2C
+    /// data pointer automatically rolls between 00H ~ 06H, if I2C read begins at any address among 00H~06H.
+    /// ROL_PNT: “0”: Normal, “1”: Enable pointer roll-over function
+    /// </summary>
+    public enum RollPointer : byte
```
krwq, line 11: -I'd suggest to change binary enums into bool unless it's super unobvious what true/false will mean. Also I suggest to use the name `RollI2cPointer`
- what does this mean for user? when should they use it? should this be an implementation detail?

src/devices/Qmc5883l/VectorExtension.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <param name="vector">Direction Vector</param>
+        /// <returns>Heading (Angle)</returns>
+        public static Angle GetHeading(this Vector3 vector)
+        {
+            double deg = Math.Atan2(vector.Y, vector.X) * 180 / Math.PI;
+
+            deg = deg < 0 ? deg + 360 : deg;
+            return Angle.FromDegrees(deg);
+        }
+    }
```
krwq, line 27: I'd recommend to only add this in the sample. Especially this has high likelihood of someone doing exactly same in their code which would make it very annoying to use if that happens.

src/devices/Qmc5883l/VectorExtension.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// Calculate heading.
+        /// </summary>
+        /// <param name="vector">Direction Vector</param>
+        /// <returns>Heading (Angle)</returns>
+        public static Angle GetHeading(this Vector3 vector)
+        {
+            double deg = Math.Atan2(vector.Y, vector.X) * 180 / Math.PI;
+
+            deg = deg < 0 ? deg + 360 : deg;
+            return Angle.FromDegrees(deg);
```
krwq, line 25: as a side note you could construct it directly from radians and then use Degrees in the code

src/devices/Qmc5883l/samples/Program.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Numerics;
+using System.Threading;
+using Iot.Device.Qmc5883l;
+
+I2cConnectionSettings settings = new(1, Qmc5883l.DefaultI2cAddress);
+using I2cDevice device = I2cDevice.Create(settings);
+using (Qmc5883l sensor = new(device))
+{
+    // Configure the sensor to match our needs.
+    // Make sure to set the device in Continuous mode if you plan on reading any data.
```
krwq, line 15: I'd suggest to make it a default in the constructor and optionally (only if you find it useful) provide a flag to not touch the mode or make it standby by default

src/devices/Qmc5883l/samples/Program.cs
```diff
@@ -0,0 +1,45 @@
...
+    sensor.DeviceMode = Mode.Continuous;
+    sensor.FieldRange = FieldRange.Gauss8;
+    sensor.Interrupt = Interrupt.Disable;
+    sensor.OutputRate = OutputRate.Rate200Hz;
+    sensor.RollPointer = RollPointer.Disable;
```
krwq, line 20: ideally simple sample should not require anything to be set, only show those for advanced users

Ellerbach:
[Triage] @Speedo69 do you think you'll have to time address the latest comments and move this PR forward?

Issue #1381, opened 2020-12-22
We should enable SA1201, SA1202, SA1203, SA1204, SA1214 and SA1215


krwq:
See https://stackoverflow.com/questions/150479/order-of-items-in-classes-fields-properties-constructors-methods

This would improve readability in the code in general

Ellerbach:
I'm not super in favor of those rules. For the bindings I've developed and in general when I'm coding, I regroup elements logically and not by their access type. If you look at the long bindings, they all have regions, grouping things logically. It makes it millions times easier to maintain than if all those functions, properties would be grouped by access level.
I'm find with the constant on the top, the global variables on the top, the constructor on the top but for the rest, I would let the logical flow of grouping by features.

krwq:
fair enough, we should only pick those we all feel comfortable with

pgrawehr:
I think rules SA1201, 1203 and eventually 1214 and 1215 are fine, but SA1202 and SA1204 really don't help in adding readability. They force that all static methods must come before all non-static methods, and private methods must come after public methods etc. Besides that one generally places a private helper method next to where it's used, this has also the downside of requiring that a method must be moved within the file when it's visibility changes.

Ellerbach:
[Triage] We will investigate the following:
* SA1201 - ? depends on number of changes
* SA1203 - ok
* SA1204 - ? depends on number of changes
* SA1214 - ok
* SA1215 - ? depends on number of changes

krwq:
* [x] SA1201 - Not convinced about it, seems more noisy than worth it, I didn't find the order super obvious
* [x] SA1203 - ok - already enabled
* [x] SA1204 - I found the code easier to read in most cases
* [x] SA1214 - ok - already enabled
* [x] SA1215 - ? depends on number of changes - no errors so assuming this was intuitively done correctly

PR #1444, opened 2021-02-19
Enable SA1204

krwq:
Fixes: https://github.com/dotnet/iot/issues/1381

* [x] SA1201 - Not convinced about it, seems more noisy than worth it, I didn't find the order super obvious
* [x] SA1203 - ok - already enabled
* [x] SA1204 - I found the code easier to read in most cases
* [x] SA1214 - ok - already enabled
* [x] SA1215 - enabled, no errors so assuming this was intuitively done correctly or it doesn't exist

cc: @pgrawehr

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1444)


pgrawehr:
I'm not convinced, especially when it comes to the method ordering. You normally have similar methods (be they static or not) close to each other (or private helper methods). This tears that appart.

krwq:
@pgrawehr look at the diff, in most cases this enforced nicer field order so that static data is on top along with const. Only some of them were methods - in most of them the enforced order made sense as it grouped the static methods together. I think overall there is more benefit, only ocassionally it is a mild inconvenience. As a bonus I also found one method which should have been static and was not (it was in the middle of other static methods)

Ellerbach:
I have to say I'm like @pgrawehr I'm not fully convinced. This doesn't bring much more clarity. It mixes a bit the logic. While I understand the logic, the static on the top is not always the best to read. Especially when the variables are declared after.

pgrawehr:
Unfortunately, there's no way of enforcing the ordering only for the fields (where I too find the rule useful) but not for the methods.

Ellerbach (approved):

PR #1446, opened 2021-02-20
Fixing SPI LSB mode not supported to have software emulation

### LINKED ISSUE #1308: Ensure SPI's DataFlow.Lsb works
Per https://github.com/dotnet/iot/pull/1306#discussion_r529242750

SPI's DataFlow.Lsb might not be working on Raspberry PI. We should double check we use driver correctly and if driver doesn't support it we should reverse bits in software.

Once this work is done we should enforce this mode in PN532 and ensure it still works

Ellerbach:
Support LSB mode for SPI when it's not supported by the hardware
- This fixes #1308 where LSB and Mode0 is not supported by Raspberry Pi
- Uses software emulation for this purpose and reverse the bytes
- Tested on PN532 and fully operational on a Raspberry Pi
- Added LSB support for Windows driver as well

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1446)


src/System.Device.Gpio/System/Device/Spi/Devices/UnixSpiDevice.cs
```diff
@@ -71,7 +72,17 @@ private unsafe void Initialize()
                 int result = Interop.ioctl(_deviceFileDescriptor, (uint)SpiSettings.SPI_IOC_WR_MODE, nativePtr);
                 if (result == -1)
                 {
-                    throw new IOException($"Error {Marshal.GetLastWin32Error()}. Can not set SPI mode to {_settings.Mode}.");
+                    // We should try the other mode and invert the buffers manually
+                    // RPI can't open Mode0 in LSB for example
+                    mode = _settings.DataFlow == DataFlow.LsbFirst ? mode &= ~UnixSpiMode.SPI_LSB_FIRST : mode |= UnixSpiMode.SPI_LSB_FIRST;
```
krwq, line 77: can you change this line to regular `if`? it's super confusing and you also double assign the value to `mode`

src/System.Device.Gpio/System/Device/Spi/Devices/UnixSpiDevice.cs
```diff
@@ -182,9 +203,22 @@ public override unsafe void Write(ReadOnlySpan<byte> buffer)
         {
             Initialize();
 
-            fixed (byte* dataPtr = buffer)
+            if (_isInverted)
             {
-                Transfer(dataPtr, null, buffer.Length);
+                Span<byte> toSend = stackalloc byte[buffer.Length];
```
krwq, line 208: You should not `stackalloc` variable buffer length as you can easily stackoverflow here - instead cap stackalloc to some number of bytes and affter that do regular allocation. See this as an example:
https://github.com/dotnet/runtime/blob/1d9e50cb4735df46d3de0cee5791e97295eaf588/src/libraries/System.Text.Json/src/System/Text/Json/Reader/JsonReaderHelper.Unescaping.cs#L17-L19

If you can use ArrayPool that's good, if not then regular `new byte[...]` is fine

src/System.Device.Gpio/System/Device/Spi/Devices/UnixSpiDevice.cs
```diff
@@ -202,11 +236,29 @@ public override unsafe void TransferFullDuplex(ReadOnlySpan<byte> writeBuffer, S
                 throw new ArgumentException($"Parameters '{nameof(writeBuffer)}' and '{nameof(readBuffer)}' must have the same length.");
             }
 
-            fixed (byte* writeBufferPtr = writeBuffer)
+            if (_isInverted)
+            {
+                Span<byte> toSend = stackalloc byte[writeBuffer.Length];
```
krwq, line 241: same comment

src/System.Device.Gpio/System/Device/Spi/Devices/Windows10SpiDevice.cs
```diff
@@ -14,6 +14,7 @@ internal class Windows10SpiDevice : SpiDevice
     {
         private readonly SpiConnectionSettings _settings;
         private WinSpi.SpiDevice _winDevice;
+        private bool _isInverted = false;
```
krwq, line 17: similar comments for this file

src/System.Device.Gpio/System/Device/Spi/Devices/Windows10SpiDevice.cs
```diff
@@ -112,7 +123,17 @@ public override void WriteByte(byte value)
...
         public override void Write(ReadOnlySpan<byte> buffer)
         {
-            _winDevice.Write(buffer.ToArray());
+            if (_isInverted)
+            {
+                Span<byte> toSend = stackalloc byte[buffer.Length];
+                buffer.CopyTo(toSend);
+                ReverseByte(toSend);
+                _winDevice.Write(toSend.ToArray());
+            }
```
krwq, line 132: actually here it might be better to always allocate regular array rather than stackallocating, reversing and converting to array

src/System.Device.Gpio/System/Device/Spi/Devices/Windows10SpiDevice.cs
```diff
@@ -112,7 +123,17 @@ public override void WriteByte(byte value)
         /// </param>
         public override void Write(ReadOnlySpan<byte> buffer)
         {
-            _winDevice.Write(buffer.ToArray());
+            if (_isInverted)
+            {
+                Span<byte> toSend = stackalloc byte[buffer.Length];
+                buffer.CopyTo(toSend);
+                ReverseByte(toSend);
+                _winDevice.Write(toSend.ToArray());
```
krwq, line 131: You can simplify this whole method to:

```csharp
byte[] toSend = buffer.ToArray();
if (_isInverted)
{
    ReverseByte(toSend);
}

_winDevice.Write(toSend);
```

src/System.Device.Gpio/System/Device/Spi/Devices/Windows10SpiDevice.cs
```diff
@@ -127,9 +148,20 @@ public override void TransferFullDuplex(ReadOnlySpan<byte> writeBuffer, Span<byt
...
-            byte[] byteArray = new byte[readBuffer.Length];
-            _winDevice.TransferFullDuplex(writeBuffer.ToArray(), byteArray);
-            byteArray.CopyTo(readBuffer);
+            if (_isInverted)
+            {
+                byte[] byteArray = new byte[readBuffer.Length];
+                ReverseByte(byteArray);
+                _winDevice.TransferFullDuplex(writeBuffer.ToArray(), byteArray);
+                byteArray.CopyTo(readBuffer);
+                ReverseByte(readBuffer);
```
krwq, line 157: similar comment as with Write, it can be greatly simplified since we always use array anyway

src/System.Device.Gpio/System/Device/Spi/SpiDevice.cs
```diff
@@ -82,5 +82,45 @@ protected virtual void Dispose(bool disposing)
...
+        /// </summary>
+        /// <param name="span">The Span of byte to reverse</param>
+        public static void ReverseByte(Span<byte> span)
+        {
+            for (int i = 0; i < span.Length; i++)
+            {
+                span[i] = ReverseByte(span[i]);
+            }
+        }
+
```
krwq, line 124: extra enter

src/devices/Pn532/README.md
```diff
@@ -182,8 +182,9 @@ It is possible to emulate any Type A, Type B and Felica cards.
 Communication support:
 - [X] HSU serial port: fully supported
 - [X] I2C: supported
-- [X] SPI: supported but using a specific chip select pin as well as using LSB with reverse bytes function rather than built in function. This is due to current limitation in the SPI implementations for Raspberry Pi. **Important**: when using Chip Select, it must be a pin which is different from the SPI defined one.
-- [ ] Hardware reset pin: This can be done with the user code.
+- [X] SPI: supported but using a specific chip select pin
```
krwq, line 185: since you're changing this file you can also edit line 9: `## USage` => `## Usage`

krwq:
Overall looks good to me, couple of improvements and security considerations.

src/System.Device.Gpio/System/Device/Spi/Devices/UnixSpiDevice.cs
pgrawehr: The number of bytes typically sent is very small. Isn't there a length limit for SPI transfers?

Ellerbach: I'll go for a normal new byte[], no need of a Span anyway, what is sent if a normal buffer then.

src/System.Device.Gpio/System/Device/Spi/Devices/Windows10SpiDevice.cs
Ellerbach: ho yes and the rest as well. Perfect catch!

src/System.Device.Gpio/System/Device/Spi/Devices/UnixSpiDevice.cs
Ellerbach: Reworked all of them as it's only byte[] based. Simplified and make it more clear.

src/System.Device.Gpio/System/Device/Spi/Devices/UnixSpiDevice.cs
```diff
@@ -181,8 +210,13 @@ public override unsafe void WriteByte(byte value)
         public override unsafe void Write(ReadOnlySpan<byte> buffer)
         {
             Initialize();
+            byte[] toSend = buffer.ToArray();
```
krwq, line 213: I'm ok with this code when we're inverting but in case when we're not we used to not allocate at all here and now we copy the buffer

krwq: same comment for other similar methods in UnixSpiDevice, on Windows we always allocated anyway

src/System.Device.Gpio/System/Device/Spi/Devices/Windows10SpiDevice.cs
```diff
@@ -23,9 +24,14 @@ internal class Windows10SpiDevice : SpiDevice
...
-            if (settings.DataFlow != DataFlow.MsbFirst || settings.ChipSelectLineActiveState != PinValue.Low)
+            if (settings.ChipSelectLineActiveState != PinValue.Low)
             {
-                throw new PlatformNotSupportedException($"Changing {nameof(settings.DataFlow)} or {nameof(settings.ChipSelectLineActiveState)} options is not supported on the current platform.");
+                throw new PlatformNotSupportedException($"Changing{nameof(settings.ChipSelectLineActiveState)} options is not supported on the current platform.");
+            }
+
+            if (settings.DataFlow == DataFlow.MsbFirst)
+            {
+                _isInverted = true;
```
krwq, line 34: I think this should be `== DataFlow.LsbFirst`, previously we used to throw for Lsb and now we will invert

src/System.Device.Gpio/System/Device/Spi/SpiDevice.cs
```diff
@@ -82,5 +82,44 @@ protected virtual void Dispose(bool disposing)
...
+        /// Reverse the bytes
+        /// </summary>
+        /// <param name="span">The Span of byte to reverse</param>
+        public static void ReverseByte(Span<byte> span)
+        {
+            for (int i = 0; i < span.Length; i++)
+            {
+                span[i] = ReverseByte(span[i]);
+            }
+        }
```
krwq, line 123: Those should be private/internal

krwq (approved):
LGTM after changing visibility to internal/private

src/System.Device.Gpio/System/Device/Spi/SpiDevice.cs
Ellerbach: Yes, I can make that internal.

krwq (approved):

krwq:
@joperezr I think the bot merged this before the CI picked up the job (I don't expect any failures but still...)

PR #1460, opened 2021-02-25
Add nuget.org feed

richlander:
Required for dependencies.

Blocking #1451 and #1457.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1460)

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,6 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
+    <add key="nuget-org" value="https://api.nuget.org/v3/index.json" />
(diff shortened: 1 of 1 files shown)


joperezr (changes requested):
We specifically and intentionally removed nuget.org feed here, and should not add it back. If we need any external dependencies we should push thise to our internal feeds instead.

richlander:
This is what @krwq told me to do. Either way.

Who is responsible for posting the dependencies so that we can move forward on the dependabot PRs?

PR #1466, opened 2021-03-04
Introduce wait states for some commands

pgrawehr:
This improves the stability of the initialization sequence for the CharacterLcd-Displays. Previously, it sometimes happened that the init sequence was executing to fast, which caused the display to enter an undefined state (displaying random characters and not properly updating)

The problem mostly showed up on a 20x4 Display connected trough I2C. The patch is verified to work with this and also a 16x2 Display connected using GPIOs.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1466)


Ellerbach (approved):
Looks good to me. You're missing a using for the Tread somewhere.

Issue #1222, opened 2020-10-24
Unhandled exception trying to re-set PinMode with LibGpiodDriver


DanieleNeo:
**Bug description**

I was implementing DHT22 sensor and I got a strange error trying to read temperature and humidity values. The exception was thrown in the ReadThroughOneWire method, exactly when the method try to set the pin in InputPullup mode

``` 
internal virtual void ReadThroughOneWire()
        {
            byte readVal = 0;
            uint count;

            // keep data line HIGH
            _controller.SetPinMode(_pin, PinMode.Output);
            _controller.Write(_pin, PinValue.High);
            DelayHelper.DelayMilliseconds(20, true);

            // send trigger signal
            _controller.Write(_pin, PinValue.Low);
            // wait at least 18 milliseconds
            // here wait for 18 milliseconds will cause sensor initialization to fail
            DelayHelper.DelayMilliseconds(20, true);

            // pull up data line
            _controller.Write(_pin, PinValue.High);
            // wait 20 - 40 microseconds
            DelayHelper.DelayMicroseconds(30, true);

            _controller.SetPinMode(_pin, PinMode.InputPullUp);
```  

This is the exception

``` 
The pin does not support the selected mode.

   at System.Device.Gpio.GpioController.SetPinMode(Int32 pinNumber, PinMode mode)
   at ConsoleApp3.DhtBase.ReadThroughOneWire() in C:\Users\daniele\Source\Repos\ConsoleApp3\ConsoleApp3\DhtBase.cs:line 162
   at ConsoleApp3.DhtBase.ReadData() in C:\Users\daniele\Source\Repos\ConsoleApp3\ConsoleApp3\DhtBase.cs:line 130
   at ConsoleApp3.DhtBase.get_Temperature() in C:\Users\daniele\Source\Repos\ConsoleApp3\ConsoleApp3\DhtBase.cs:line 68
   at ConsoleApp3.Program.Main(String[] args) in C:\Users\daniele\Source\Repos\ConsoleApp3\ConsoleApp3\Program.cs:line 64
``` 
So I called the IsPinModeSupported method and I found that PinMode.InputPullUp is not supported (by LibGpiodDriver driver? I don't know...), not a big problem, my interface has hardware pullup, so I modified the PinMode.InputPullUp with PinMode.Input, discovering the more general issue

``` 
Error setting pin mode, for pin: 8, error: 16

[...]

PR #1473, opened 2021-03-07
A GPIO has to be writeable even in drive mode input because one wants to set the value used when switching to output mode.

### LINKED ISSUE #721: Output initialization problem
I have a Raspberry Pi 3B and I'm using an output to control a relay board. The relay board requires and active-low signal
Using the code below to initialize the output causes it to go low for long enough for the relay to briefly close before the output goes high. 

```
GpioController controller = new GpioController(PinNumberingScheme.Logical);
controller.OpenPin(18, PinMode.Output);
controller.Write(18, PinValue.High);
```

What is needed is a way of setting the output state of the pin before configuring it as an output. Calling GpioController.Write before opening the pin throws an exception.

pgrawehr:
Fixes #721 
Fixes https://github.com/dotnet/iot/issues/1222
Superseedes #1361 

Adds two new methods to GpioController:
```csharp
OpenPin(int pinNumber, PinMode mode, PinValue initialValue);
SetPinMode(int pinNumber, PinMode mode, PinValue initialValue);
```

Writing a value before changing the mode, so that the pin immediatelly enters that state (or stays where it is), seems to work just fine on the Raspi. The other boards will currently emulate the feature as good as possible (libgpiod seems to have a separate problem that changing modes always fails). 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1473)

### DIFF
--- a/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
+++ b/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
@@ -54,7 +54,7 @@ namespace System.Device.Gpio.Tests
         }
 
         [Fact]
-        public void HighPulledPinDoesNotGlitchToLowWhenChangedToOutput()
+        public void HighPulledPinDoesNotChangeToLowWhenChangedToOutput()
         {
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
--- a/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
+++ b/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
@@ -56,11 +56,25 @@ namespace System.Device.Gpio.Tests
         [Fact]
         public void HighPulledPinDoesNotGlitchToLowWhenChangedToOutput()
         {
+            ////while (!System.Diagnostics.Debugger.IsAttached)
+            ////{
+            ////    Thread.Sleep(100);
+            ////}
+
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 bool didTriggerToLow = false;
-                int testPin = OutputPin; // TODO: Change to OpenPin
-                controller.OpenPin(testPin, PinMode.InputPullUp);
+                int testPin = OutputPin;
+                // Set value to low prior to test, so that we have a defined start situation
+                controller.OpenPin(testPin, PinMode.Output);
+                controller.Write(testPin, PinValue.Low);
+                controller.ClosePin(testPin);
+                // For this test, we use the input pin as an external pull-up
+                controller.OpenPin(InputPin, PinMode.Output);
+                controller.Write(InputPin, PinValue.High);
+                Thread.Sleep(2);
+                // If we were to use InputPullup here, this would work around the problem it seems, but it would also make our test pass under almost all situations
+                controller.OpenPin(testPin, PinMode.Input);
                 Thread.Sleep(50);
                 controller.RegisterCallbackForPinValueChangedEvent(testPin, PinEventTypes.Falling, (sender, args) =>
                 {
@@ -74,6 +88,9 @@ namespace System.Device.Gpio.Tests
                 controller.SetPinMode(testPin, PinMode.Output);
                 Thread.Sleep(50);
                 Assert.False(didTriggerToLow);
+
+                controller.ClosePin(OutputPin);
+                controller.ClosePin(InputPin);
             }
         }
     }
--- a/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
+++ b/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
@@ -56,11 +56,6 @@ namespace System.Device.Gpio.Tests
         [Fact]
         public void HighPulledPinDoesNotGlitchToLowWhenChangedToOutput()
         {
-            ////while (!System.Diagnostics.Debugger.IsAttached)
-            ////{
-            ////    Thread.Sleep(100);
-            ////}
-
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 bool didTriggerToLow = false;
--- a/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
+++ b/src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
@@ -52,5 +52,29 @@ namespace System.Device.Gpio.Tests
                 Assert.Equal(PinValue.Low, controller.Read(OpenPin));
             }
         }
+
+        [Fact]
+        public void HighPulledPinDoesNotGlitchToLowWhenChangedToOutput()
+        {
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+            {
+                bool didTriggerToLow = false;
+                int testPin = OutputPin; // TODO: Change to OpenPin
+                controller.OpenPin(testPin, PinMode.InputPullUp);
+                Thread.Sleep(50);
+                controller.RegisterCallbackForPinValueChangedEvent(testPin, PinEventTypes.Falling, (sender, args) =>
+                {
+                    if (args.ChangeType == PinEventTypes.Falling)
+                    {
+                        didTriggerToLow = true;
+                    }
+                });
+
+                controller.Write(testPin, PinValue.High);
+                controller.SetPinMode(testPin, PinMode.Output);
+                Thread.Sleep(50);
+                Assert.False(didTriggerToLow);
+            }
+        }
     }
 }
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -111,6 +111,13 @@ internal partial class Interop
         internal static extern void gpiod_line_release(IntPtr lineHandle);
 
         /// <summary>
+        /// Release a previously reserved line.
+        /// </summary>
+        /// <param name="lineHandle">GPIO line handle</param>
+        [DllImport(LibgpiodLibrary)]
+        internal static extern void gpiod_line_release(SafeLineHandle lineHandle);
+
+        /// <summary>
         /// Request all event type notifications on a single line.
         /// </summary>
         /// <param name="line">GPIO line handle</param>
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -107,15 +107,11 @@ internal partial class Interop
         /// Release a previously reserved line.
         /// </summary>
         /// <param name="lineHandle">GPIO line handle</param>
+        /// <remarks>
+        /// This does NOT invalidate the line handle. This only releases the lock, so that a gpiod_line_request_input/gpiod_line_request_output can be called again.
+        /// </remarks>
         [DllImport(LibgpiodLibrary)]
         internal static extern void gpiod_line_release(IntPtr lineHandle);
-
-        /// <summary>
-        /// Release a previously reserved line.
-        /// </summary>
-        /// <param name="lineHandle">GPIO line handle</param>
-        [DllImport(LibgpiodLibrary)]
-        internal static extern void gpiod_line_release(SafeLineHandle lineHandle);
 
         /// <summary>
         /// Request all event type notifications on a single line.
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -74,9 +74,10 @@ internal partial class Interop
         /// </summary>
         /// <param name="line">GPIO line handle</param>
         /// <param name="consumer">Name of the consumer.</param>
+        /// <param name="default_val">Initial value of the line</param>
         /// <returns>0 if the line was properly reserved, -1 on failure.</returns>
         [DllImport(LibgpiodLibrary, SetLastError = true)]
-        internal static extern int gpiod_line_request_output(SafeLineHandle line, string consumer);
+        internal static extern int gpiod_line_request_output(SafeLineHandle line, string consumer, int default_val);
 
         /// <summary>
         /// Set the value of a single GPIO line.
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
@@ -232,7 +232,9 @@ namespace System.Device.Gpio.Drivers
         {
             if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle? pinHandle))
             {
-                Interop.libgpiod.gpiod_line_release(pinHandle);
+                // This call does not release the handle. It only releases the lock on the handle. Without this, changing the direction of a line is not possible.
+                // Line handles cannot be freed and are cached until the chip is closed.
+                pinHandle.ReleaseLock();
                 int requestResult = mode switch
                 {
                     PinMode.Input => Interop.libgpiod.gpiod_line_request_input(pinHandle, s_consumerName),
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
@@ -153,14 +153,14 @@ namespace System.Device.Gpio.Drivers
         {
             lock (_pinNumberLock)
             {
-            if (!_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle? pinHandle))
-            {
+                if (!_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle? pinHandle))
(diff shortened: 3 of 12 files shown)


joperezr:
[Triage]: @pgrawehr please let us know once this PR is ready to review 😄

pgrawehr:
@microhobby Please have a look at my changes to libgpiodriver. It seems to be working this way, but you have more experience with this driver.

pgrawehr:
@ZhangGaoxing Consider upgrading your drivers as well, once this is merged. If your devices support it, you should add an overload for `void SetPinMode(int pinNumber, PinMode mode, PinValue initialValue)void SetPinMode(int pinNumber, PinMode mode, PinValue initialValue)` that writes the pin value before changing the mode.

microhobby:
@pgrawehr sorry for the delay. I'll take a closer look at it yet. I will fork your branch to do a test because I believe that libgpiod will return an error if we write without first setting the direction.

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -110,6 +110,13 @@ internal partial class libgpiod
         [DllImport(LibgpiodLibrary)]
         internal static extern void gpiod_line_release(IntPtr lineHandle);
 
+        /// <summary>
+        /// Release a previously reserved line.
+        /// </summary>
+        /// <param name="lineHandle">GPIO line handle</param>
+        [DllImport(LibgpiodLibrary)]
+        internal static extern void gpiod_line_release(SafeLineHandle lineHandle);
```
joperezr, line 118: Looks like above we already have this method which takes in an IntPtr. Can we modify that one to instead take a SafeLineHandle and fix the callsites to use that instead of IntPtr?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -26,7 +26,10 @@ public class GpioController : IDisposable
         private const string HummingBoardProduct = "HummingBoard-Edge";
         private const string HummingBoardHardware = @"Freescale i.MX6 Quad/DualLite (Device Tree)";
 
-        private readonly HashSet<int> _openPins;
+        /// <summary>
+        /// If a pin element exists, that pin is open. Uses current controller's numbering scheme
+        /// </summary>
+        private readonly Dictionary<int, PinValue?> _openPins;
```
joperezr, line 32: NIT: I know this wasn't introduced by your change, but since you are changing this, should we make this a ConcurrentDictionary instead? That way it would be safer across threads.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -115,6 +118,20 @@ public void OpenPin(int pinNumber, PinMode mode)
...
+        /// <summary>
+        /// Opens a pin and sets it to a specific mode and value.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        /// <param name="mode">The mode to be set.</param>
+        /// <param name="initialValue">The initial value to be set if the mode is output. The driver will attempt to set the mode without causing glitches to the other value.
+        /// (if <paramref name="initialValue"/> is <see cref="PinValue.High"/>, the pin should not glitch to low during open)</param>
+        public void OpenPin(int pinNumber, PinMode mode, PinValue initialValue)
+        {
+            OpenPin(pinNumber);
```
joperezr, line 130: Should we instead have an OpenPin in the driver that takes in an initial value? That way if there is a driver that supports setting an initial value during opening they can do it right there, while other drivers that don't can make the two operations separately like you are doing here.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -115,6 +118,20 @@ public void OpenPin(int pinNumber, PinMode mode)
...
+            Write(pinNumber, initialValue);
```
joperezr, line 131: Do you know what is the initial pin mode when opening one for a driver? If it is output, then this method would actually light the LED even when called with mode=input right? If that's the case, perhpas we should call setPinMode first?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -158,7 +175,27 @@ public virtual void SetPinMode(int pinNumber, PinMode mode)
...
+        }
+
+        /// <summary>
+        /// Sets the mode to a pin and sets an initial value for an output pin.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        /// <param name="mode">The mode to be set.</param>
+        /// <param name="initialValue">The initial value if the <paramref name="mode"/> is output. The driver will do it's best to prevent glitches to the other value when
+        /// changing from input to output.</param>
+        public void SetPinMode(int pinNumber, PinMode mode, PinValue initialValue)
```
joperezr, line 195: mm I'm not so sure about this one. I'm fine with OpenPin having an initialValue as that makes sense always, but initialValue when changing a mode only makes sense when doing output. Since we allow to write on input pins now, I would expect folks wanting to have a immediate value would be calling Write() and then changing the mode from input to output.

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
```diff
@@ -43,6 +43,23 @@ public abstract class GpioDriver : IDisposable
...
+        /// <summary>
+        /// Sets the mode to a pin and sets an initial value for an output pin.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+        /// <param name="mode">The mode to be set.</param>
+        /// <param name="initialValue">The initial value if the <paramref name="mode"/> is output. The driver will do it's best to prevent glitches to the other value when
+        /// changing from input to output.</param>
+        protected internal virtual void SetPinMode(int pinNumber, PinMode mode, PinValue initialValue)
+        {
+            // TODO: Find out which driver support this atomically
```
joperezr, line 55: Haven't you already figured this out on this PR? If so we can remove the comment.

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
pgrawehr: The other overload is used internally by the SafeLineHandle itself, therefore I think both are needed.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: I considered this, actually. However, the way it's used right now is not thread safe, even if we use ConcurrentDictionary, because there's always a sequence such as 
```
            if (IsPinOpen(pinNumber))
            {
                throw new InvalidOperationException($"Pin {pinNumber} is already open.");
            }

            OpenPinCore(pinNumber);
            _openPins.Add(pinNumber);
```

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: I think that's not required, because `OpenPin(pinNumber)` should not change the pin mode or the value on open. This would be different if we called `OpenPin(pinNumber, mode)` here.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: That's a good point. The expectation is that opening leaves the mode as-is (but that seems to be inconsistent between drivers, see also #1479), but then yes, if it was output, and the new mode is input, it would change the value first.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: You are right, but I thought I would add the convenience method anyway (the overload is also suggested in #721)

Ellerbach:
[Triage] @krwq please have a look and if all ok for you, we're good to merge.

src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
```diff
@@ -52,5 +52,41 @@ public void InputPullResistorsWork()
                 Assert.Equal(PinValue.Low, controller.Read(OpenPin));
             }
         }
+
+        [Fact]
+        public void HighPulledPinDoesNotGlitchToLowWhenChangedToOutput()
```
krwq, line 57:
```suggestion
        public void HighPulledPinDoesNotChangeToLowWhenChangedToOutput()
```

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
krwq: I agree with @joperezr on this. Since we allow to change value in Input mode I don't believe there is much benefit of doing this and it's confusing

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
```diff
@@ -230,10 +230,10 @@ protected internal override void RemoveCallbackForPinValueChangedEvent(int pinNu
         /// <inheritdoc/>
         protected internal override void SetPinMode(int pinNumber, PinMode mode)
         {
-            int requestResult = -1;
             if (_pinNumberToSafeLineHandle.TryGetValue(pinNumber, out SafeLineHandle? pinHandle))
             {
-                requestResult = mode switch
+                Interop.libgpiod.gpiod_line_release(pinHandle);
```
krwq, line 235: why do you even need the interop? Should you just Dispose the handle?

I find this confusing you're releasing native resource and then use handle further...

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
joperezr: [Triage]: SetPinMode with initial value is just a convinience method so that you don't have to call Write() and then SetPinMode(). Given it is really only saving one call, It's better to remove this so as to not to overcomplicate the API on the controller.

pgrawehr: Removed overload, as discussed.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
joperezr: I agree with @krwq here that this indeed looks very fishy. Can you point to the documentation you found where the right thing to do is to call release on the handle and later using it again? The docs I found do point to saying that running this operation will invalidate the handle which then gets used again so I don't think this is the way to go. Also, even if we do have to call the release method, I would never call it directly from here, and instead dispose the safehandle by calling Dispose() and letting the release happen there.

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
joperezr: I think this method should only ever be used by the Dispose method in SafeHandle, and never called upopn directly. Take a look at my comment on @krwq's comment where you are using this. We should either not be calling release from there as it could be a bug, or if we decide that that is correct, then we should be calling it via disposing the handle.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
joperezr: ConcurrentDictionary won't fail when you call Add more than once across threads. I understand most of this class isn't thread safe, but the idea is to move towards thread-safety not away from it.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
pgrawehr: I can't find unambiguous documentation on it, unfortunately. But looking at the code [(here)](https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/tree/lib/core.c) it does seem like the right thing to do. `gpiod_line_release` does not free the line handle. Actually, no handles are ever freed until `gpiod_line_close` is called, it seems.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: It's not about ConcurrentDictionary failing, but about calling OpenPinCore twice with the same pin in that scenario. If we want thread safety here, we should use a lock and do it right. Should not hurt much as long as we don't need the lock on `Read` or `Write` operations.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
joperezr: If so then that would be a potential bug in our implementation of libgpiod driver because we could have a memory leak by not appropriately freeing the handle on Dispose.

pgrawehr: No, because it seems the line handle _cannot_ be properly freed. So we could even remove the SafeLineHandle class, as the line handle is cached in the library. Repeatedly calling `gpiod_chip_get_line` always returns the same value as long as the chip is not released. Let's discuss in todays call.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -115,6 +119,27 @@ public void OpenPin(int pinNumber, PinMode mode)
...
+        /// Opens a pin and sets it to a specific mode and value.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        /// <param name="mode">The mode to be set.</param>
+        /// <param name="initialValue">The initial value to be set if the mode is output. The driver will attempt to set the mode without causing glitches to the other value.
+        /// (if <paramref name="initialValue"/> is <see cref="PinValue.High"/>, the pin should not glitch to low during open)</param>
+        public void OpenPin(int pinNumber, PinMode mode, PinValue initialValue)
+        {
+            OpenPin(pinNumber);
+            if (GetPinMode(pinNumber) == PinMode.Output)
```
joperezr, line 132: Seems like we need an else clause here so that in pinmode was input, we still save the initial value so that when the mode is changed to output, this initial value is respected. Seems like the way it is now, we are only caring about initial value when mode is output.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
```diff
@@ -244,14 +246,25 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
...
-                throw ExceptionHelper.GetIOException(ExceptionResource.SetPinModeError, Marshal.GetLastWin32Error(),
-                    pinNumber);
-            }
+            throw new InvalidOperationException($"Pin {pinNumber} is not open");
+        }
+
+        /// <inheritdoc />
+        protected internal override void SetPinMode(int pinNumber, PinMode mode, PinValue initialValue)
+        {
+            // On the Raspberry Pi, we can Write the out value even if the mode is something other than out. It will take effect once we change the mode
```
joperezr, line 265: This comment says that this behavior is specific to Raspberry Pi, but does this apply as well to other boards? Libgpiod driver could be used in other boards so we should double check to make sure that other boards won't break with this.

src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
```diff
@@ -52,5 +52,41 @@ public void InputPullResistorsWork()
                 Assert.Equal(PinValue.Low, controller.Read(OpenPin));
             }
         }
+
+        [Fact]
+        public void HighPulledPinDoesNotChangeToLowWhenChangedToOutput()
```
joperezr, line 57: Are there other drivers that also won't trigger to Low? for example libgpiod?

joperezr:
Left a few comments to address, but other than that this looks good. Once those are addressed I'm good with us merging this. Thanks for all the work @pgrawehr

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
pgrawehr: Well, it works on the Raspberry Pi (both with the Raspberry Pi driver and the Libgpiod driver), whether other boards would fail on this I cannot tell. At least the libgpiod code does not include any test that could throw here, except standard error handling on an Ioctl call.

src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
pgrawehr: Yes, libgpiod seems to work as well. I need to check why I didn't add the test there as well.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: The whole condition can be removed, we just need to set the desired value.

joperezr (approved):
LGTM. The only remaining feedback to address before merging is to add the test not only to the RPI driver but also to the libgpiod driver as we discussed in one previous comment.

pgrawehr:
@joperezr: Good point you brought that up. The test cannot be directly copied for libgpiod, since it doesn't allow setting the pin mode while an event is registered. So to really test this, we would need a third pin connected to our input and output pins on the test board. So I tested with the scope instead and found that we were missing the third argument to `gpiod_line_request_output` which is exactly for the purpose we need it. So I used this and tested it with the scope.

PR #1476, opened 2021-03-08
Add IOutputSegment interface

richlander:
Take 2 on https://github.com/dotnet/iot/pull/1107. 

The goal is to establish an abstraction over the multiplexing APIs that I've added to this repo.

- [ShiftRegister](https://github.com/dotnet/iot/tree/main/src/devices/ShiftRegister)
- [CharlieplexSegment](https://github.com/dotnet/iot/tree/main/src/devices/Charlieplex)

Description of the architecture:

- IOutputSegment is the new interface that abstracts updating LEDs from the various wiring and address schemes and controllers.
- GpioOutputSegment is the implementation for direction GPIO wiring.
- ShiftRegister has an implementation of the interface.
- CharlieplexSegment will have an implementation shortly. I'll add it before switching the PR from draft mode.
- VirtualOutputputSegment is a helper that makes writing an implementation easier and makes the behavior more consistent. I wrote tests for this type to ensure it has the behavior I intended.
- I wrote a sample app using GpioOutputSegment. I'll test it over ShiftRegister and CharlieplexSegment shortly (again, before switching from draft). So far, it works well.
- The part that took me the longest (design) time was managing time delays and cancellation. I think the model I have now is workable, and even good. I wonder if you think the same.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1476)


src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,45 @@
...
+    public interface IOutputSegment : IDisposable
+    {
+        /// <summary>
+        /// Length of segment (number of outputs)
+        /// </summary>
+        int Length { get; }
+
+        /// <summary>
+        /// Writes a PinValue to a multiplexed output.
+        /// Does not perform a latch.
```
joperezr, line 32: NIT: Can we clarify here what does performing or not a latch means? It would be good since theoretically new classes might want to implement this in the future, and if this is something that should be applied to all implementations, it may require a more detailed explanation

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        void Write(int output, PinValue value);
```
joperezr, line 34: NIT: I don't think naming here is correct regarding `output`. It is not clear from here that output really means `position` of the LED to turn on on the segment. It is specially  confusing given you have another overload of Write that also takes an int but it means a different thing.

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Writes an int (processed as one of more bytes) to a multiplexed output.
+        /// Written one byte at a time, left to right. Least significant bit will written to index 0.
+        /// Does not perform a latch.
+        /// </summary>
+        void Write(int value);
```
joperezr, line 41: Could a segment be more than 32 bits? Perhaps a better API for this woudl be `void Write(ReadOnlySpan<byte>)`?

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Turns off all outputs.
+        /// Performs a latch.
+        /// </summary>
+        void Clear();
```
joperezr, line 47: Should this be a bit more descriptive like `TurnOffAllOutputs` or something? Clear (specially in a segment which technically is a collection) has the connotation in C# of removing things from the collection.

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Displays segment until token receives a cancellation signal, possibly due to a specificated duration.
+        /// As appropriate for a given implementation, performs a latch.
+        /// </summary>
+        void Display(TimeSpan time);
```
joperezr, line 53: Same here, perhaps have an API that is a bit more descriptive with just the name? I guess that even by reading the summary here I'm not sure if the idea here is that not until you call Display then all previous Writes haven't really been applied?

src/devices/Common/Iot/Device/Multiplexing/GpioOutputSegment.cs
```diff
@@ -0,0 +1,115 @@
...
+        private readonly bool _shouldDispose;
+        private readonly CancellationToken _token;
+        private readonly VirtualOutputSegment _segment;
+        private GpioController _controller;
+
+        /// <summary>
+        /// IOutputSegment implementation that uses GpioController.
+        /// </summary>
+        /// <param name="pins">The GPIO pins that should be used and are connected.</param>
+        /// <param name="token">Cancellation token to use to notify cancelling the output segment.</param>
```
joperezr, line 28: I don't think CancellationToken should be a parameter that gets passed in to the constructor. Seems like it is really only used by the Dispay method, so I think the right way to do it would be to have the method `void Display()` instead be `Task DisplayAsync(CancellationToken ct)`

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Segment values.
+        /// </summary>
+        PinValue this[int index] { get; }
+
+        /// <summary>
+        /// CancellationToken for segment.
+        /// </summary>
+        CancellationToken CancellationToken { get; }
```
joperezr, line 28: Why is this public?

src/devices/Common/Iot/Device/Multiplexing/GpioOutputSegment.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            _segment.Clear();
+            Display();
+        }
+
+        /// <summary>
+        /// Displays segment for a given duration.
+        /// Alternative to Thread.Sleep
+        /// </summary>
+        public void Display(TimeSpan time)
```
joperezr, line 89: How can I call this if I want the display to stay lit forever? in other words, why is the other Display method not public

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        /// <summary>
+        /// Length of segment (number of outputs)
+        /// </summary>
+        int Length { get; }
+
+        /// <summary>
+        /// Segment values.
+        /// </summary>
+        PinValue this[int index] { get; }
```
pgrawehr, line 23: Why no setter?

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// Turns off all outputs.
+        /// Performs a latch.
+        /// </summary>
+        void Clear();
+
+        /// <summary>
+        /// Displays segment until token receives a cancellation signal, possibly due to a specificated duration.
+        /// Publishes (latches) values.
+        /// </summary>
+        void Display(CancellationToken token);
```
pgrawehr, line 48: Maybe this should be Async or have an Async overload?

src/devices/Common/Iot/Device/Multiplexing/Multiplexing.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
pgrawehr, line 4: Use `<TargetFrameworks>netcoreapp2.1;net5.0</TargetFrameworks>`

src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
```diff
@@ -0,0 +1,102 @@
...
+using System;
+using System.Device.Gpio;
+using System.Threading;
+
+namespace Iot.Device.Multiplexing.Utility
+{
+    /// <summary>
+    /// Interface that abstracts multiplexing over a segment of outputs.
+    /// </summary>
+    public class VirtualOutputSegment : IOutputSegment
```
pgrawehr, line 13: First: Description is missleading, this is not an interface. 
Second: It seems this should rather be named `NullOutputSegment` as it does nothing except discarding all inputs.
Third: Seal classes not designed for inheritance.

src/devices/Common/Iot/Device/Multiplexing/samples/OutputSegment-driver.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
```
pgrawehr, line 5: Same as above. Unless there's a reason, projects should target both netcoreapp2.1 and net5.0

samples/led-animate/AnimateLeds.cs
```diff
@@ -0,0 +1,167 @@
...
+
+public class AnimateLeds : IDisposable
+{
+    private readonly int[] _leds;
+    private readonly int[] _ledsRev;
+    private IOutputSegment _segment;
+    public int LitTimeDefault = 200;
+    public int DimTimeDefault = 50;
+    public int LitTime = 200;
+    public int DimTime = 50;
```
pgrawehr, line 19: Public non-const fields? Is this really intended? Prefer either constants or properties (if changing should be possible).

src/devices/Common/Iot/Device/Multiplexing/Multiplexing.csproj
pgrawehr: Actually, I think you can drop this .csproj entirely. There's a project file in src/devices/Common that should include all files from all its subfolders.

src/devices/Common/Iot/Device/Multiplexing/GpioOutputSegment.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            Display();
+            _segment.Display(token);
+        }
+
+        private void Display()
+        {
+            for (int i = 0; i < _pins.Length; i++)
+            {
+                _controller.Write(_pins[i], _segment[i]);
```
pgrawehr, line 91: If you use the `void Write(ReadOnlySpan<PinValuePair> pinValuePairs)` overload, the driver can attempt to synchronize pin writes.

src/devices/Common/Iot/Device/Multiplexing/GpioOutputSegment.cs
```diff
@@ -0,0 +1,107 @@
...
+namespace Iot.Device.Multiplexing
+{
+    /// <summary>
+    /// IOutputSegment implementation that uses GpioController.
+    /// </summary>
+    public class GpioOutputSegment : IOutputSegment, IDisposable
+    {
+        private readonly int[] _pins;
+        private readonly bool _shouldDispose;
+        private readonly VirtualOutputSegment _segment;
```
pgrawehr, line 20: What purpose does the `VirtualOutputSegment` have here? No value is ever read from it.

src/devices/Common/Iot/Device/Multiplexing/tests/UnitTest1.cs
```diff
@@ -0,0 +1,89 @@
+﻿using System;
+using System.Threading;
+using Iot.Device.Multiplexing.Utility;
+using Xunit;
+
+namespace Iot.Device.Multiplexing
+{
+    public class UnitTest1
```
pgrawehr, line 8: Please use a better name.

pgrawehr (changes requested):

krwq:
@richlander maybe this is a bit dummy question but what's the benefit of sing IOutputSegment over a regular list or IEnumerable? Would it make more sense to add GpioPin class so that it could be easier combined with other classes? (see https://github.com/dotnet/iot/issues/215)

src/devices/Common/Iot/Device/Multiplexing/GpioOutputSegment.cs
richlander: That's not true. It is used extensively.

richlander:
Look at the implementations:

- Charlieplex: https://github.com/dotnet/iot/pull/1476/files#diff-1522f58990241b68d30f9ba5b13c3d75c27fa7cc18d4f3b5461fc2bb1c9333d5R215
- ShiftRegister: https://github.com/dotnet/iot/pull/1476/files#diff-95e80221d0adc25b15ceea0e6baba6d912ee897e559d87045c3e62a77f516f1bR217
- Gpio: https://github.com/dotnet/iot/pull/1476/files#diff-ef995fe1c4ddda472f2d78e74d4938ab8c1bf375a3913c5ba53f0699659a79b5R16

How would those implementations be handled with an IEnumerable?

src/devices/Common/Iot/Device/Multiplexing/GpioOutputSegment.cs
richlander: `VirtualOutputSegment` maintains state for all the writes.

src/devices/Common/Iot/Device/Multiplexing/samples/OutputSegment-driver.csproj
richlander: This is for the sample app. I'm thinking that samples can all target .NET latest. If you look, you can see that's already the case. For bindings, I totally agree.

src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
richlander: Will make it sealed. `Virtual` seems better than `Null`. It is the backing store for all the implementations I wrote. Is there a better name for that?

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,50 @@
...
+using System;
+using System.Device.Gpio;
+using System.Threading;
+
+namespace Iot.Device.Multiplexing
+{
+    /// <summary>
+    /// Interface that abstracts multiplexing over a segment of outputs.
+    /// </summary>
+    public interface IOutputSegment : IDisposable
```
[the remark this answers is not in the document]
richlander: It's to collect pins or nodes that are exposed by a multiplexer / different sources. In addition, those sources have very different interaction models, addressing schemes, and can expose differing numbers of pins.

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
richlander: How's this?

```csharp
        /// <summary>
        /// Writes a PinValue to a virtual output.
        /// Does not display output.
        /// </summary>
        void Write(int output, PinValue value);
```

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
richlander: Renamed to `index`. Does that work?

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
richlander: Good call. An `int` is a bad plan. I'll fix that.

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
richlander: Right. Would `Update` be a better term than `Write`?

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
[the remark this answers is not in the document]
richlander: I updated the text: `Abstracts a segment of outputs over multiplexing sources (like a shift register).`

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
richlander: Good call. Adding.

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
richlander: Adding overload.

src/devices/Common/Iot/Device/Multiplexing/GpioOutputSegment.cs
richlander: Adding that method.

src/devices/Common/Iot/Device/Multiplexing/Multiplexing.csproj
richlander: Deleted project.

samples/led-animate/AnimateLeds.cs
richlander: Fixed

src/devices/Common/Iot/Device/Multiplexing/tests/UnitTest1.cs
richlander: Done

src/devices/Charlieplex/CharlieplexSegment.cs
```diff
@@ -137,15 +141,16 @@ public void Write(int node, PinValue value, int duration = 0)
 
             if (duration > 0)
             {
-                DisplaySegment(TimeSpan.FromMilliseconds(duration));
+                using CancellationTokenSource cts = new(duration);
```
joperezr, line 144: NIT: I know you are not adding this, but duration should be a timespan, not an int

src/devices/Charlieplex/CharlieplexSegment.cs
```diff
@@ -210,5 +213,88 @@ public void Dispose()
...
             }
         }
+
+        // IOutputSegment Implementation
+        // Only supported when shift register is connected with GPIO
+
+        /// <summary>
+        /// The length of the segment; the number of GPIO pins it exposes.
+        /// </summary>
+        int IOutputSegment.Length => _segment.Length;
```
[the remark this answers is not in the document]
richlander: Do you mean they are `Interface.` or why I'm relying on `VirtualOutputSegment` for the implementation?

src/devices/Charlieplex/CharlieplexSegment.cs
richlander: Will fix that.

src/devices/Charlieplex/CharlieplexSegment.cs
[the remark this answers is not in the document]
joperezr: Yeah, the way they are right now you need to cast the class to the interface in order to call the methods which is less than ideal. Meaning:

```c#
CharlieplexSegment myCPS = new CharlieplexSegment();
var theLength = myCPS.Length; // <-- This line won't compile
var theLength = ((IOutputSegment)myCPS).Length; // <-- This line does compile
```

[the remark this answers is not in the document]
joperezr: To fix, simply change these to look like:
```suggestion
        public int Length => _segment.Length;
```

src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Device.Gpio;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Iot.Device.Multiplexing.Utility
+{
+    /// <summary>
+    /// Interface that abstracts multiplexing over a segment of outputs.
+    /// </summary>
+    public sealed class VirtualOutputSegment : IOutputSegment
```
joperezr, line 14: To be honest, I'm not sure I understand if there is enough value of having a separate class for this as opposed to just have a PinValue array on the GpioOutputSegment class that preserves state. The only thing I see is so that the WriteByteAsValue methods can get reused by CharlieplexSegment and GpioOutputSegment but do let me know if I'm missing something else. Also, if we do decide to keep it around, can we make it internal? I'm not sure I see the point of having this type public as it is really just an implementation detail.

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,63 @@
...
+        int Length { get; }
+
+        /// <summary>
+        /// Segment values.
+        /// </summary>
+        PinValue this[int index] { get; set; }
+
+        /// <summary>
+        /// Writes a PinValue to a virtual segment.
+        /// Does not display output.
```
joperezr, line 28: One more small nit here, perhaps say

```suggestion
        /// Does not display output until calling Display() or Display(CancellationToken ct) methods.
```

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// </summary>
+        void Write(byte value);
+
+        /// <summary>
+        /// Writes discrete underlying bits to a virtual output.
+        /// Writes each byte, left to right. Least significant bit will written to index 0.
+        /// Does not display output.
+        /// </summary>
+        void Write(ReadOnlySpan<byte> value);
```
joperezr, line 44: Sorry, now that I saw this perhaps instead of ReadOnlySpan<byte> this should be ReadOnlySpan<PinValue>? The advantage of changing it is that PinValue can be implicitly casted from byte, bool, or int.

src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
richlander: You are right. It's borderline. How about we go with this for the PR, and then decide what to do with it. Perhaps this was a good tool for prototyping quickly, since I didn't have to make changes in multiple places. Certainly, it could be collapsed, but it's an implementation detail so can be updated later.

src/devices/Charlieplex/CharlieplexSegment.cs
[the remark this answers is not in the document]
richlander: That's what I thought you meant. I did it like that because each of the APIs have their own concepts and I didn't want this set of APIs to get confused with those. For this one (`Length`) however, it might make sense to make it generic. For `ShiftRegister`, I went with `BitLength` originally. I don't recall why. I'll make this change to `Length` on each of the implementations.

src/devices/Common/Iot/Device/Multiplexing/IOutputSegment.cs
richlander: I enjoy using binary literals because it is such a compact syntax. Also, the important states have such easy to remember (even more terse) representations: `0`, `255`. How about we sit on this one a little longer.  I know we've discussed it before. I'm torn.  It probably makes sense to look at some Python examples to see how representing multiple bits is handled.

src/devices/Charlieplex/CharlieplexSegment.cs
```diff
@@ -210,5 +215,88 @@ public void Dispose()
...
+
+        /// <summary>
+        /// The length of the segment; the number of GPIO pins it exposes.
+        /// </summary>
+        public int Length => _segment.Length;
+
+        /// <summary>
+        /// Segment values.
+        /// </summary>
+        PinValue IOutputSegment.this[int index]
```
joperezr, line 230: Sorry I only added a comment on Length property, but I really meant on all the methods of the interface. Looks like you implemented all of them through the interface, where we really want them to be implemented through the class so that they can be called without casting.

src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
pgrawehr: If it's a backing store, `Virtual` is probably ok, or something like `MemoryMapped`. However, what confused me more is that seemingly nothing is ever read back from this class (initially, it didn't even have a way to get values back), so maybe you implemented it differently than what your intent was.

src/devices/Common/Iot/Device/Multiplexing/samples/OutputSegment-driver.csproj
pgrawehr: True, that's fine with me

src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
pgrawehr: Ok, found it. It's hidden a bit in that Write() loop, but got it now.

src/devices/Charlieplex/CharlieplexSegment.cs
richlander: Ya, that is intentional. These methods and then ones on the class are not intended to be used together. As implemented, they won't play nicely in all cases.  This is a layer on top, which is why I went with the interface methods like that. I could have made a new class for each one, but that seemed like overkill.

joperezr (approved):
I think we could still change one or two things here but given the time constraint I'm fine with merging this as is and revising whatever is needed later.

Issue #1483, opened 2021-03-15
Kernel SPI buffer size limitation with SpiDevice.Write()


thecaptncode:
I am a contributor on a Waveshare E-Paper device driver for the Raspberry Pi 2 or later that uses the SPI bus.  The solution uses System.Device.Gpio v1.3.0 and targets .NET Standard 2.0.  In trying to boost performance, one of the things I am doing is moving to writing to the SPI device using ```SpiDevice.Write(ReadOnlySpan<Byte>)``` since a single screen on the Waveshare 7.5inch E-Paper (B) driver is over 120KB of data and must be written in full with every update.

On my Raspberry Pi 3B+ running Raspbian 10 (buster) I see that as long as I write slices of 4KB or smaller to the SpiDevice, this works fine, but as soon as a slice is at all larger than 4KB, it throws the exception:
```
Unhandled exception. System.IO.IOException: Error 90 performing SPI data transfer.
   at System.Device.Spi.UnixSpiDevice.Transfer(Byte* writeBufferPtr, Byte* readBufferPtr, Int32 buffersLength)
   at System.Device.Spi.UnixSpiDevice.Write(ReadOnlySpan`1 buffer)
   at Waveshare.Common.EPaperDisplayHardware.Write(ReadOnlySpan`1 buffer) in C:\Users\captn\source\repos\thecaptncode\Waveshare.EPaperDisplay\Waveshare\Common\EPaperDisplayHardware.cs:line 186
   at Waveshare.Common.EPaperDisplayBase.SendData(Byte[] data) in C:\Users\captn\source\repos\thecaptncode\Waveshare.EPaperDisplay\Waveshare\Common\EPaperDisplayBase.cs:line 275
   at Waveshare.Common.EPaperDisplayBase.DisplayImage(Bitmap image) in C:\Users\captn\source\repos\thecaptncode\Waveshare.EPaperDisplay\Waveshare\Common\EPaperDisplayBase.cs:line 187
   at Waveshare.Example.Program.Main() in C:\Users\captn\source\repos\thecaptncode\Waveshare.EPaperDisplay\Waveshare.Example\Program.cs:line 54

```

By the way, this is a self-contained deployment with no SDK installed on the RPi.

[...]

PR #1496, opened 2021-03-19
SpiBusInfo stub added (#1483)

thecaptncode:
New class created for SpiBusInfo
This is currently a stub/placeholder for getting the SPI Bus buffer size.  Additional work will be needed.  
See #1483

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1496)

### DIFF
--- a/src/devices/Common/System/Device/Spi/SpiBusInfo.cs
+++ b/src/devices/Common/System/Device/Spi/SpiBusInfo.cs
@@ -1,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Device.Spi
+{
+    public static class SpiBusInfo
+    {
+        private static int s_bufferSize = GetBufferSize();
+
+        // we would need to decide if this should throw when size can't be found or just return -1 or other placeholder
+        public static int BufferSize => s_bufferSize;
+
+        private static int GetBufferSize()
+        {
+            // non-throwing version of int.Parse(File.ReadAllText("/sys/module/spidev/parameters/bufsiz"))
+            // return -1 on error
+        }
+    }
+}
--- a/src/devices/Common/System/Device/Spi/SpiBusInfo.cs
+++ b/src/devices/Common/System/Device/Spi/SpiBusInfo.cs
@@ -1,19 +1,41 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.IO;
+using System.Runtime.InteropServices;
+
 namespace System.Device.Spi
 {
+    /// <summary>
+    /// Base class for SPI bus information.
+    /// </summary>
     public static class SpiBusInfo
     {
         private static int s_bufferSize = GetBufferSize();
 
-        // we would need to decide if this should throw when size can't be found or just return -1 or other placeholder
+        /// <summary>
+        /// Buffer size assigned to the SPI bus or -1 if the buffer size is not available.
+        /// </summary>
         public static int BufferSize => s_bufferSize;
 
         private static int GetBufferSize()
         {
-            // non-throwing version of int.Parse(File.ReadAllText("/sys/module/spidev/parameters/bufsiz"))
-            // return -1 on error
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                const string parameterPath = "/sys/module/spidev/parameters/bufsiz";
+                if (File.Exists(parameterPath) && int.TryParse(File.ReadAllText(parameterPath), out int buferSize))
+                {
+                    return buferSize;
+                }
+                else
+                {
+                    return -1;
+                }
+            }
+            else
+            {
+                throw new PlatformNotSupportedException("This value is only supported on Linux operating systems");
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1496) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1496) before we can accept your contribution.<br/><br/>:x: thecaptncode [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1496)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1496) it.</sub>

src/devices/Common/System/Device/Spi/SpiBusInfo.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        private static int s_bufferSize = GetBufferSize();
+
+        // we would need to decide if this should throw when size can't be found or just return -1 or other placeholder
+        public static int BufferSize => s_bufferSize;
+
+        private static int GetBufferSize()
+        {
+            // non-throwing version of int.Parse(File.ReadAllText("/sys/module/spidev/parameters/bufsiz"))
+            // return -1 on error
```
krwq, line 16: you will need to put actual implementation here 😄
- handle non-existing file
- handle badly formatted file (i.e. use int.TryParse)

thecaptncode: @krwq I completely misunderstood what you were asking about a pull request.  I can go ahead and make it a functional implementation.  Yes, file existence check and TryParse should be there as well as as check for RuntimeInformation.IsOSPlatform(OSPlatform.Linux) I believe.  Should I delete the previous pull request?

krwq: you can just add a commit in this PR (push to your branch, it will auto-update)

src/devices/Common/System/Device/Spi/SpiBusInfo.cs
```diff
@@ -1,19 +1,41 @@
...
+                    return buferSize;
+                }
+                else
+                {
+                    return -1;
+                }
+            }
+            else
+            {
+                throw new PlatformNotSupportedException("This value is only supported on Linux operating systems");
```
pgrawehr, line 37: This should not throw, never. This would now throw a TypeInitalizationException when just referencing this type on Windows. This is especially critical since this code executes in a static constructor.

krwq: yes, if we return -1 when info is not available let's stick to that

src/devices/Common/System/Device/Spi/SpiBusInfo.cs
```diff
@@ -0,0 +1,34 @@
...
+        private static int s_bufferSize = GetBufferSize();
+
+        /// <summary>
+        /// Buffer size assigned to the SPI bus or -1 if the buffer size is not available.
+        /// </summary>
+        public static int BufferSize => s_bufferSize;
+
+        private static int GetBufferSize()
+        {
+            const string parameterPath = "/sys/module/spidev/parameters/bufsiz";
```
krwq, line 23: can you also add a test similar to https://github.com/dotnet/iot/blob/main/src/System.Device.Gpio.Tests/Mcp3008Tests.cs in the same folder? In the test just do something similar to:

```csharp
if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
{
    Assert.True(SpiBusInfo.GetBufferSize() > 0, "SpiBusInfo.GetBufferSize() returned 0 or less");
}
else
{
    Assert.Equal(-1, SpiBusInfo.GetBufferSize());
}
```

All our Linux tests run on Raspbian so that should always work. This is to make sure that i.e. file doesn't contain any whitespace such as newline character and int.TryParse isn't returning false because of that

Ellerbach:
[Triage] @thecaptncode looks good! Please add the test and if all goes right, we should be good to go!

krwq:
Interesting I'm seeing this test fail on Debug but not on release - perhaps SPI is misconfigured on that machine...

```
    System.Device.Gpio.Tests.SpiBusInfoTests.VerifyBufferSize [FAIL]
      SpiBusInfo.BufferSize returned 0 or less
      Expected: True
      Actual:   False
```

thecaptncode:
@krwq Are you needing anything additional from me at this time?

src/devices/Common/System/Device/Spi/SpiBusInfo.cs
```diff
@@ -0,0 +1,34 @@
...
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) &&
+                File.Exists(parameterPath) &&
+                int.TryParse(File.ReadAllText(parameterPath), out int buferSize))
+            {
+                return buferSize;
```
pgrawehr, line 28:
```suggestion
                int.TryParse(File.ReadAllText(parameterPath), out int bufferSize))
            {
                return bufferSize;
```

pgrawehr:
@thecaptncode I just had a minor finding, but otherwise I think this is good to be merged this week.

pgrawehr (approved):

krwq (approved):

PR #1520, opened 2021-04-15
Opening/Closing a pin should not affect its state if possible

### LINKED ISSUE #1222: Unhandled exception trying to re-set PinMode with LibGpiodDriver

### LINKED ISSUE #1479: Opening an output pin with libpgiod driver changes the pin value
**Describe the bug**

I want to read the value from an output pin, but by just opening an output pin it gets set to high. So reading it always returns high of course.

But using gpioget to read the pin value returns it as low, if that was what it was.

**Steps to reproduce**

Create GPIO controller
Open pin (for one that is low)
It is set to high, without an explicit write

```csharp
var led1 = 19;

using var gpio = new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver(2));

gpio.OpenPin(led1, PinMode.Output);
// gpio.Write(led1, PinValue.High);
gpio.ClosePin(led1);
```

**Expected behavior**

I expected to be able to open the pin without its state being affected, so I could read the state afterwards.

**Actual behavior**

Opening the pin changes state.

**Versions used**

Debian 10 (arm64)
System.Device.Gpio package: 1.3.0
.NET Core: 3.1

Dockerfile
```
FROM mcr.microsoft.com/dotnet/sdk:3.1 as build

WORKDIR /src
COPY GpioApp.csproj ./
RUN dotnet restore GpioApp.csproj

COPY Program.cs ./
RUN dotnet build GpioApp.csproj -c Release -o /app

FROM mcr.microsoft.com/dotnet/runtime:3.1-alpine-arm64v8 as final

RUN apk update && apk add libgpiod-dev

WORKDIR /app
COPY --from=build /app .
ENTRYPOINT ["dotnet", "GpioApp.dll"]
```

docker run --privileged --rm dotnet-gpio:3.1

pgrawehr:
Fixes #1222 
Fixes #1479 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1520)

### DIFF
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -23,8 +23,10 @@ namespace System.Device.Gpio.Tests
             _testOutputHelper = testOutputHelper;
         }
 
-        [Fact]
-        public void PinValueStaysSameAfterDispose()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void PinValueStaysSameAfterDispose(bool closeAsHigh)
         {
             var driver = GetTestDriver();
             if (driver is SysFsDriver)
@@ -37,15 +39,15 @@ namespace System.Device.Gpio.Tests
             {
                 controller.OpenPin(OutputPin, PinMode.Output);
                 controller.OpenPin(InputPin, PinMode.Input);
-                controller.Write(OutputPin, PinValue.High);
+                controller.Write(OutputPin, closeAsHigh);
                 Thread.SpinWait(100);
-                Assert.Equal(PinValue.High, controller.Read(InputPin));
+                Assert.Equal(closeAsHigh, controller.Read(InputPin));
             }
 
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 controller.OpenPin(InputPin, PinMode.Input);
-                Assert.Equal(PinValue.High, controller.Read(InputPin));
+                Assert.Equal(closeAsHigh, controller.Read(InputPin));
             }
         }
 
@@ -173,9 +175,11 @@ namespace System.Device.Gpio.Tests
             }
         }
 
-        [Fact]
+        [Theory]
+        [InlineData(PinMode.Output)]
+        [InlineData(PinMode.Input)]
         [Trait("SkipOnTestRun", "Windows_NT")] // Currently, the Windows Driver is defaulting to InputPullDown, and it seems this cannot be changed
-        public void OpenPinDefaultsModeToLastMode()
+        public void OpenPinDefaultsModeToLastMode(PinMode modeToTest)
         {
             var driver = GetTestDriver();
             if (driver is SysFsDriver)
@@ -188,12 +192,17 @@ namespace System.Device.Gpio.Tests
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), driver))
             {
                 controller.OpenPin(OutputPin);
-                controller.SetPinMode(OutputPin, PinMode.Output);
-                Assert.Equal(PinMode.Output, controller.GetPinMode(OutputPin));
+                controller.SetPinMode(OutputPin, modeToTest);
+                Assert.Equal(modeToTest, controller.GetPinMode(OutputPin));
                 controller.ClosePin(OutputPin);
+            }
+
+            // Close controller, to make sure we're not caching
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), driver))
+            {
                 controller.OpenPin(OutputPin);
                 Thread.Sleep(100);
-                Assert.Equal(PinMode.Output, controller.GetPinMode(OutputPin));
+                Assert.Equal(modeToTest, controller.GetPinMode(OutputPin));
             }
         }
 
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -44,7 +44,6 @@ namespace System.Device.Gpio.Tests
 
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
-                controller.OpenPin(OutputPin, PinMode.Output);
                 controller.OpenPin(InputPin, PinMode.Input);
                 Assert.Equal(PinValue.High, controller.Read(InputPin));
             }
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -26,7 +26,14 @@ namespace System.Device.Gpio.Tests
         [Fact]
         public void PinValueStaysSameAfterDispose()
         {
-            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+            var driver = GetTestDriver();
+            if (driver is SysFsDriver)
+            {
+                // This check fails on the SysFsDriver, because it always sets the value to 0 when the pin is opened (but on close, the value does stay high)
+                return;
+            }
+
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), driver))
             {
                 controller.OpenPin(OutputPin, PinMode.Output);
                 controller.OpenPin(InputPin, PinMode.Input);
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -182,6 +182,7 @@ namespace System.Device.Gpio.Tests
             {
                 controller.OpenPin(OutputPin);
                 controller.SetPinMode(OutputPin, PinMode.Output);
+                Assert.Equal(PinMode.Output, controller.GetPinMode(OutputPin));
                 controller.ClosePin(OutputPin);
                 controller.OpenPin(OutputPin);
                 Thread.Sleep(1000);
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -171,10 +171,10 @@ namespace System.Device.Gpio.Tests
         [Trait("SkipOnTestRun", "Windows_NT")] // Currently, the Windows Driver is defaulting to InputPullDown, and it seems this cannot be changed
         public void OpenPinDefaultsModeToLastMode()
         {
+            // This works for input/output on most systems, but not on pullup/down, since sometimes the hardware doesn't have read possibilities for those (ie. the Raspi 3)
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 controller.OpenPin(OutputPin);
-                Assert.Equal(PinMode.Input, controller.GetPinMode(OutputPin));
                 controller.SetPinMode(OutputPin, PinMode.Output);
                 controller.ClosePin(OutputPin);
                 controller.OpenPin(OutputPin);
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -177,15 +177,22 @@ namespace System.Device.Gpio.Tests
         [Trait("SkipOnTestRun", "Windows_NT")] // Currently, the Windows Driver is defaulting to InputPullDown, and it seems this cannot be changed
         public void OpenPinDefaultsModeToLastMode()
         {
+            var driver = GetTestDriver();
+            if (driver is SysFsDriver)
+            {
+                // See issue #1581. There seems to be a library-version issue or some other random cause for this test to act differently on different hardware.
+                return;
+            }
+
             // This works for input/output on most systems, but not on pullup/down, since sometimes the hardware doesn't have read possibilities for those (ie. the Raspi 3)
-            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), driver))
             {
                 controller.OpenPin(OutputPin);
                 controller.SetPinMode(OutputPin, PinMode.Output);
                 Assert.Equal(PinMode.Output, controller.GetPinMode(OutputPin));
                 controller.ClosePin(OutputPin);
                 controller.OpenPin(OutputPin);
-                Thread.Sleep(1000);
+                Thread.Sleep(100);
                 Assert.Equal(PinMode.Output, controller.GetPinMode(OutputPin));
             }
         }
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -184,6 +184,7 @@ namespace System.Device.Gpio.Tests
                 controller.SetPinMode(OutputPin, PinMode.Output);
                 controller.ClosePin(OutputPin);
                 controller.OpenPin(OutputPin);
+                Thread.Sleep(50);
                 Assert.Equal(PinMode.Output, controller.GetPinMode(OutputPin));
             }
         }
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -24,8 +24,7 @@ namespace System.Device.Gpio.Tests
         }
 
         [Fact]
-        [Trait("SkipOnTestRun", "Windows_NT")] // The WindowsDriver is kept as High when disposed.
-        public void PinValueReturnsToLowAfterDispose()
+        public void PinValueStaysSameAfterDispose()
         {
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
@@ -40,7 +39,7 @@ namespace System.Device.Gpio.Tests
             {
                 controller.OpenPin(OutputPin, PinMode.Output);
                 controller.OpenPin(InputPin, PinMode.Input);
-                Assert.Equal(PinValue.Low, controller.Read(InputPin));
+                Assert.Equal(PinValue.High, controller.Read(InputPin));
             }
         }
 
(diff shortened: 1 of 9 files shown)


joperezr:
This PR depends on https://github.com/dotnet/iot/pull/1473 which is why it is marked as no merge for now. Once that one goes in this will be rebased and will be ready to review.

pgrawehr:
@joperezr I do not know why SysFsDriverTests.OpenPinDefaultsModeToLastMode fails on the build machine. It works just fine for me, and even if I execute the sysfs commands manually in the shell, I get the expected result. Maybe something is different between the Pi3 and Pi4 (I only have a Pi4) or a recent software update changed that behavior.

joperezr:
This is ready for review, and I'll take a look at the failure in my RPi 3 to see if I can repro and investigate what is going on.

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -169,17 +174,18 @@ public void ThrowsIfWritingClosedPin()
...
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 controller.OpenPin(OutputPin);
-                Assert.Equal(PinMode.Input, controller.GetPinMode(OutputPin));
                 controller.SetPinMode(OutputPin, PinMode.Output);
                 controller.ClosePin(OutputPin);
                 controller.OpenPin(OutputPin);
-                Assert.Equal(PinMode.Input, controller.GetPinMode(OutputPin));
+                Thread.Sleep(50);
+                Assert.Equal(PinMode.Output, controller.GetPinMode(OutputPin));
```
joperezr, line 188: I think the fact that this test is failing now makes sense to me. SysFSDriver wasn't changed, nor did the Controller, but we are changing the expected value in the test, which now makes it fail so that seems reasonable to me. What this tells me is that the sys fs driver on the Pi will always set mode to Output by default, so perhaps what we need to do is to emmulate the desired behavior on the sysfs driver by keeping track of the pinmodes values.

joperezr: That said, I just tried in my Pi and that doesn't seem to be the case if I manually make the change by invoking the file directly.  Can you try adding a larget Sleep just to see if that is what is doing it?

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -169,17 +174,18 @@ public void ThrowsIfWritingClosedPin()
...
+        [Trait("SkipOnTestRun", "Windows_NT")] // Currently, the Windows Driver is defaulting to InputPullDown, and it seems this cannot be changed
+        public void OpenPinDefaultsModeToLastMode()
         {
+            // This works for input/output on most systems, but not on pullup/down, since sometimes the hardware doesn't have read possibilities for those (ie. the Raspi 3)
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
                 controller.OpenPin(OutputPin);
-                Assert.Equal(PinMode.Input, controller.GetPinMode(OutputPin));
                 controller.SetPinMode(OutputPin, PinMode.Output);
                 controller.ClosePin(OutputPin);
```
joperezr, line 185:
```suggestion
                Assert.Equal(PinMode.Output, controller.GetPinMode(OutputPin));
                controller.ClosePin(OutputPin);
```

Ellerbach (approved):

Ellerbach:
[Triage] As there is one test failing on the hardware pipeline build/test, and it does not fail on real hardware we have with us, we will temporary disable (skip) the failing test:

```text
    System.Device.Gpio.Tests.SysFsDriverTests.OpenPinDefaultsModeToLastMode [FAIL]
      Assert.Equal() Failure
      Expected: Output
      Actual:   Input
      Stack Trace:
        /_/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs(188,0): at System.Device.Gpio.Tests.GpioControllerTestBase.OpenPinDefaultsModeToLastMode()
```
And an issue will be open for further investigation @pgrawehr .

pgrawehr:
@Ellerbach The build finally passed!
@krwq Can you sign off, so this can be merged? (Needs two approvals)

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -38,9 +44,8 @@ public void PinValueReturnsToLowAfterDispose()
 
             using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
             {
-                controller.OpenPin(OutputPin, PinMode.Output);
                 controller.OpenPin(InputPin, PinMode.Input);
-                Assert.Equal(PinValue.Low, controller.Read(InputPin));
+                Assert.Equal(PinValue.High, controller.Read(InputPin));
```
krwq, line 48: I recommend to make this test a theory and test it with both values

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -169,17 +174,26 @@ public void ThrowsIfWritingClosedPin()
...
+                return;
+            }
+
+            // This works for input/output on most systems, but not on pullup/down, since sometimes the hardware doesn't have read possibilities for those (ie. the Raspi 3)
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), driver))
             {
                 controller.OpenPin(OutputPin);
-                Assert.Equal(PinMode.Input, controller.GetPinMode(OutputPin));
                 controller.SetPinMode(OutputPin, PinMode.Output);
+                Assert.Equal(PinMode.Output, controller.GetPinMode(OutputPin));
```
krwq, line 192: same comment about testing this for input and output

krwq: in here might be a good idea to also write something to make sure it's not just a cached value of mode

src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
```diff
@@ -27,6 +28,20 @@ public RaspberryPiDriverTests(ITestOutputHelper testOutputHelper)
 
         protected override PinNumberingScheme GetTestNumberingScheme() => PinNumberingScheme.Logical;
 
+        private bool IsRaspi4()
+        {
+            if (File.Exists("/proc/device-tree/model"))
```
krwq, line 33: I'd prefer we used APIs exposed in System.Device for this, if not available at least check for type name for the driver - I do not want to duplicate logic here. I feel this might become an issue once we Raspi5/Raspi6 shows up or some other board shows up. Even better you can [check if InputPullDown mode is supported](https://docs.microsoft.com/en-us/dotnet/api/system.device.gpio.gpiocontroller.ispinmodesupported?view=iot-dotnet-1.3#System_Device_Gpio_GpioController_IsPinModeSupported_System_Int32_System_Device_Gpio_PinMode_)

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
```diff
@@ -187,6 +187,30 @@ protected internal override void OpenPin(int pinNumber)
...
 
+                int mode = Interop.libgpiod.gpiod_line_direction(pinHandle);
+                if (mode == 1)
+                {
+                    pinHandle.PinMode = PinMode.Input;
+                }
+                else if (mode == 2)
+                {
+                    pinHandle.PinMode = PinMode.Output;
+                }
```
krwq, line 198: should you throw on different value?

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
```diff
@@ -187,6 +187,30 @@ protected internal override void OpenPin(int pinNumber)
...
+                    int bias = Interop.libgpiod.gpiod_line_bias(pinHandle);
+                    if (bias == (int)RequestFlag.GPIOD_LINE_REQUEST_FLAG_BIAS_PULL_DOWN)
+                    {
+                        pinHandle.PinMode = PinMode.InputPullDown;
+                    }
+
+                    if (bias == (int)RequestFlag.GPIOD_LINE_REQUEST_FLAG_BIAS_PULL_UP)
+                    {
+                        pinHandle.PinMode = PinMode.InputPullUp;
+                    }
```
krwq, line 211: what other options there are? Should we throw or do something if it's out of the known range?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -239,6 +236,56 @@ protected internal override void SetPinMode(int pinNumber, PinMode mode)
...
+                    2 => PinMode.InputPullDown,
+                    _ => PinMode.Input,
+                };
+            }
+            else
+            {
+                // Pi3. We can't detect the pull mode, since it cannot be read back according to the documentation
+            }
+
+            if (_pinModes[pinNumber] is object)
```
krwq, line 277: `is not null`

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -85,6 +85,7 @@ protected virtual int GetLogicalPinNumber(int pinNumber)
 
         /// <summary>
         /// Opens a pin in order for it to be ready to use.
+        /// If possible, the driver attempts to open the pin without changing its mode or value.
```
krwq, line 88: I'd remove "If possible" - it might be possible but we might not support it

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
```diff
@@ -26,6 +26,7 @@ public abstract class GpioDriver : IDisposable
 
         /// <summary>
         /// Opens a pin in order for it to be ready to use.
+        /// The driver should attempt to open the pin without affecting its current state.
```
krwq, line 29:
```suggestion
        /// The driver attempts to open the pin without changing its mode or value.
```

krwq: (try be consistent about the docs and use the same sentences if possible)

krwq (approved):
Couple of nits

src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
pgrawehr: We do not expose this API, therefore it needs duplication here. I'm also not very happy about this. We have the following possibilities:

- Use InternalsVisibleTo and change the implementation in System.Device to internal.
- Make the implementation public
- Make the implementation public, but with a more generic API.

I would prefer 3, that means having some generic hardware detection API. But this is way beyond this PR.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
pgrawehr: The value of PinMode defaults to input. This shouldn't ever give something else, but I don't like throwing here because if the library does something unexpected, this would make the whole driver unusable.

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
pgrawehr: Currently, only these values are known. See also above

src/System.Device.Gpio.Tests/RaspberryPiDriverTests.cs
krwq: I guess we can fix this when it becomes a problem

krwq:
Thank you @pgrawehr!

PR #1525, opened 2021-04-21
Added TryWrite() and TryRead() methods to I2cDevice

### LINKED ISSUE #1519: Support Ack Polling for I2C
When writing to an EEPROM (Microchip 24LC256), the device goes into a "busy" state for a few milliseconds when a page of data has been written. It is possible to "poll" the device to determine when it has finished by essentially performing a zero byte write, essentially meaning that the device address is sent on the bus, then if the device is ready it will acknowledge (ACK) and if it is not ready it will not acknowledge (NACK).

The issue is that there needs to be a way to poll the device to determine whether it is ready for the next command. 

Ideal solution: add the following method to I2cDevice: `bool AckPoll()`, which sends only the device address on the bus and returns a boolean matching the ACK/NACK state. This can then be read in a tight loop to determine when the device is ready.
Could probably also be implemented as an awaitable async method with a timeout to do background polling, but the simple implementation is fine.

Note: I am not using the underlying OS support for I2c via `I2cDevice.Create()`, I have derived from I2cBus and I2cDevice to implement a Serial to I2C converter that works on any OS, so all my testing so far has been done using this running on Windows 10. The trivial implementation below _works_, in that a zero byte write that is not acknowleged by the device throws an exception stating "NACK on address" - it just takes too long to throw/catch the exception longer than the busy time of the device), hence the need for a *proper* AckPoll method that doesnt rely on catching exceptions (I am _assuming_ that the OS support for I2C works in the same way): 

```
bool AckPoll()
{
    try
    {
        i2cDevice.Write(new byte[0]);
        return true;
    }
    catch
    {
        return false;
    }
}
```

philrawlings:
Fixes #1519 

This adds provides `I2cDevice.TryRead()` and `I2cDevice.TryWrite()` methods which do no not throw exceptions on IO errors. 

This allows for the following scenarios:
- **Acknowledge polling of devices**: Typically used on EEPROMs, when a write is being performed, the EEPROM goes into a busy state for a few milliseconds and will ignore subsequent commands. The solution is to poll the device for acknowledgements by sending the address with the write bit set, but no data bytes. Simply using the standard `Write` method and catching exceptions is too slow due to the overhead involved (takes longer than the EEPROM write busy time).
- **Bus scanning**: Provides a faster means to do a bus scan where you know that the majority of addresses will not respond - meaning the overhead of catching each exception is removed.

Added a unit test, which utilises the BME280. This sends the address with the write bit set, but with no data bytes - then checks that the response is true (no error).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1525)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1525) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1525) before we can accept your contribution.<br/><br/>:x: philrawlings [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1525)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1525) it.</sub>

krwq:
We've only approved this API shape: https://github.com/dotnet/iot/issues/1519#issuecomment-820540408

we will need to discuss these APIs separately.

pgrawehr:
@philrawlings : We discussed your approach this week and we are not sure it really needs both methods in the public API. It may be useful to have them both internally (and maybe try both, or one of them, depending on the address), but we didn't see a use case for having both the methods in the public API. Does your use case with the EEPROM really require one of the methods or can it use either?

philrawlings:
@pgrawehr: I only need the ability to write zero bytes. 
In other words, the message sent would be as follows: 
- Start condition
- device address
- write bit set
- stop condition
The device will acknowledge (or not) and I need to know whether the device acknowledged successfully.

I'm happy to slim back the API if you let me know what you want the method(s) to look like. 
I noticed that my unit test fails with the BME280 as part of the CI actions. I need to set up my Raspberry Pi locally to check to see if it works with an EEPROM, to see if it is an issue with my code or whether the BME280 doesn't like it. However, if you let me know what you want the API to look like I'll make those modifications at the same time. There is more discussion about this on the [issue thread](https://github.com/dotnet/iot/issues/1519) which covers why I moved to the more generic `TryWrite` method rather than `IsDeviceReady` - however I'm happy to implement whatever you want.

pgrawehr:
After having read that large post in the issue again, I understand that it's not so easy. I'll have to think about the best approach once again.

src/System.Device.Gpio/System/Device/I2c/UnixI2cBus.cs
```diff
@@ -152,12 +172,12 @@ internal unsafe void WriteRead(int deviceAddress, ReadOnlySpan<byte> writeBuffer
...
                 fixed (byte* readBufferPointer = readBuffer)
                 {
-                    WriteReadCore((ushort)deviceAddress, writeBufferPointer, readBufferPointer, (ushort)writeBuffer.Length, (ushort)readBuffer.Length);
+                    WriteReadCore((ushort)deviceAddress, writeBufferPointer, readBufferPointer, (ushort)writeBuffer.Length, (ushort)readBuffer.Length, true);
                 }
             }
         }
 
-        protected virtual unsafe void WriteReadCore(ushort deviceAddress, byte* writeBuffer, byte* readBuffer, ushort writeBufferLength, ushort readBufferLength)
+        protected virtual unsafe bool WriteReadCore(ushort deviceAddress, byte* writeBuffer, byte* readBuffer, ushort writeBufferLength, ushort readBufferLength, bool throwExceptionOnIOError)
```
pgrawehr, line 180: Technically, this is a breaking change, but I guess there won't be many derived classes that are affected.

pgrawehr (commented):
@philrawlings I guess I'll suggest this to be approved. But can you please rebase? There are some unit tests failing and we need to analyze the cause (some may be due to problems with the test hardware).

philrawlings:
@pgrawehr - I'll try to update later this week

Ellerbach:
[Triage] We are more leading to a solution similar how [.NET nanoFramework is doing](https://github.com/nanoframework/System.Device.I2c/blob/40a88d6ffb53b46c37fc19f0fcf6e0e1fa388030/System.Device.I2c/I2cDevice.cs#L57) and we are waiting for an API review on the best way to do.
In short, we don't want to change the existing Read and Write API, but not neither add a TryRead/TryWrite as this would be confusing as well. Reason why originally the the proposition was a IsReady.
The changes would then to add a new Read and new Write, something like (API review dependent) `public I2cTransferResult ReadEx(Span<byte> buffer)`. This one will return the structure with the full details (number of byte sent/received and the error if any).

cc @bartonjs

krwq:
I've got guidelines that ReadWithXyz name is acceptable as long as Xyz is the result. (example options: ReadWithErrorInfo, ReadWithStatus, ...)

josesimoes:
> I've got guidelines that ReadWithXyz name is acceptable as long as Xyz is the result. (example options: ReadWithErrorInfo, ReadWithStatus, ...)

This conversation has a high probability to derail very, very quickly jumping into naming standards and become philosophical 😉 but here it goes.

My 2 cents: if the method names besides telling you what the method does, will also start including the result type, the API is going to lose _readability_ and become hard to read/understand.
I understand that this is meant to address breaking changes, but... sometimes those are needed or easy to remedy.

I get it and it's clear when one sees **Try**Parse or Read**Async**, those are "generic" and convey a sense on what the method it's meant to do and how it will perform. 
The vast majority of methods in .NET API return something (either an error, an expected value or a feedback about the execution) and we rarely see any hint about what _that_ is in the name. And that's OK.

My suggestion is to choose the path of simplicity and keep the names Read and Write and "just" start returning a struct with the operation outcome for the caller to handle as it deems appropriate.

pgrawehr:
@josesimoes The problem is that we currently have `public abstract void Read(Span<byte> buffer);`, and this is throwing exceptions in case of an error. Now if we change this to return a structure instead of throwing an exception, this is a breaking change - and one that is not even causing a build failure but instead will result in undefined runtime behavior after an update.

philrawlings:
Some more information which may (or may not) help with the naming convention:
I have been working on some firmware using STM32 HAL and there is a function there called IsDeviceReady()

```C
for (uint16_t addr = 0x03; addr <= 0x77; addr++) {
      status = HAL_I2C_IsDeviceReady(&hi2c1, addr << 1, 1, 1000);
}
```
The IsDeviceReady function always does a write (address with no data bytes), regardless of the address value and returns _OK_ if the device acknowledges or _ERROR_ if it does not. While this is a bit different to the way `i2cDetect` works under Linux (which does a write for all addresses except 0x30 to 0x3F and 0x50 to 0x5F, where actually performs a one byte read) - for my use case this is exactly what I want. Note that replicating the `i2cDetect` functionality _wont_ work for me as I need to do a zero byte write to address 0x50.

So personally, if this makes the decision any easier - I'd be happy with @joperezr 's original suggestion of:

```C#
    public abstract partial class I2cBus : IDisposable
    {
       //..
+      public virtual bool IsDeviceReady(int deviceAddress);
    }

    public abstract partial class I2cDevice : IDisposable
    {
       //..
+      public virtual bool IsDeviceReady();
    }
```
then make the documentation clear as to what the function does. While it may not be as flexible as the TryWrite/TryRead option, it does solve the primary use case for this feature and doesnt introduce breaking changes.

pgrawehr:
@philrawlings We had a quite lengthy discussion on this issue in our last team call, and we came to the conclusion that it would probably be the best way to go forward by adding an overload to Read/Write that returns a status. The driver implementation for nanoframework is already doing that, and we're trying to allign APIs as best as possible. 

So for this PR, this would basically mean you rename TryRead/TryWrite to ReadWithResult/WriteWithResult and return a structure similar to `I2CTransferResult` from [here](https://github.com/nanoframework/System.Device.I2c/blob/develop/System.Device.I2c/I2cTransferResult.cs). Maybe you find a better name. We can't change the implementation of Read/Write from void to a return value, since that would break clients expecting an exception. This solution would solve your problem (I suppose), allow both reading and writting and offer extra error information even for the "normal" transmission modes (in case the underlying OS supports it).

josesimoes:
@pgrawehr got it and I understand the predicament of not wanting to add breaking changes to the API.

May I suggest that:
1) Add those new `ReadWithResult` and `WriteWithResult`
2) Mark the existing `Read` and `Write` with and Obsolete attribute
3) In the next version where that's possible, remove them
4) In version following that one, add `Read` and `Write` again _but_ with the return result and mark the `ReadWithResult` and `WriteWithResult` as Obsolete. 
5) Remove `ReadWithResult` and `WriteWithResult` in the version next to that one.

It's a lengthy process but it would serve the purposes of keeping everyone happy and (eventually) align the APIs.

krwq:
@josesimoes fair point, we'll figure out the plan for that. For now seems that regardless of which path we take we will have some sort of ReadWithResult. Obsoletion can wait until we have a clearance on making the breaking change and we should open the issue about this before we merge this PR. I'm not convinced having just one method is also good. Read/Write throwing an exception seem easier to use than checking result on every call (on the other hand most of the time you call it once or twice per library and have some sort of higher level reading routine)

Ellerbach:
[Triage] @philrawlings we're waiting then for you to implement the few changes discussed in this thread. Let us know if anything unclear. Currently, we only expect you to do the step 1:

> 1. Add those new `ReadWithResult` and `WriteWithResult`

Thanks!

src/DeleteMe/DeleteMe.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
krwq, line 1: deleteme 🙄

philrawlings: Yeah - sorry, that is temporary. I'm sure theres better ways to do this but needed to sync across to a Linux PC with a logic analyser attached, but remote debug from my Windows laptop running VS - will get rid of that shortly!

src/System.Device.Gpio/System/Device/I2c/Devices/Windows10I2cDevice.cs
```diff
@@ -77,6 +77,26 @@ public override void Read(Span<byte> buffer)
...
         }
 
+        /// <summary>
+        /// Attempts to read data from the I2C device. If the read fails for any reason, returns false rather than throwing an exception.
+        /// </summary>
+        /// <param name="buffer">
+        /// The buffer to read the data from the I2C device.
+        /// The length of the buffer determines how much data to read from the I2C device.</param>
+        /// <returns>true: read succeeded, false if the read failed for any reason.</returns>
+        public override bool ReadWithResult(Span<byte> buffer)
```
krwq, line 87: This should match https://github.com/nanoframework/System.Device.I2c/blob/develop/System.Device.I2c/I2cTransferResult.cs as closely as possible

philrawlings: Yep - started implementing that but then realised I'm not sure we have access to the enough level of detail to set those properties. I realise now that pushing to origin was probably a bad idea at this stage as it triggers a review process plus automated tests - nice to know the system works though! 
Am currently cloning the repo on my Linux PC so I can remotely debug with hardware attached, then planning to implement I2cTransferResult with as much info as can be captured from the low level calls

philrawlings:
OK - having had some time to debug this, I'm not sure it is going to work (on Linux at least).
I'm hoping I can be proved wrong but I appear to have hit a dead end.

Essentially, when `Interop.ioctl(BusFileDescriptor, (uint)I2cSettings.I2C_RDWR, new IntPtr(&msgset))` is called in `UnixI2cBus.WriteReadCore()`, if the write and read buffer lengths are zero, then **nothing is sent on the bus and the function returns -1**. Also `Marshal.GetLastWin32Error()` returns 22.

Presumably related to this: 
https://stackoverflow.com/questions/1195659/i2c-write-acknowledge-polling-in-linux-kernel

Any thoughts from anyone?

pgrawehr:
[Triage] @philrawlings We don't fully understand what you mean now, because you earlier said that the PR as currently standing works for your use case? Maybe you need at least to write the address?

philrawlings:
@pgrawehr - The previous PR would have worked for me from an _API perspective_ (as would have `IsDeviceReady` or `WriteWithResult`) - however when I pushed the change - it failed the unit test. It is only now that I have been able to investigate (by setting up real hardware connected to a logic analyser) - and unfortunately the low level call to send/receive data does not work if you have tx/rx buffer sizes set to zero. I had _hoped_ it would just send the address on the bus - but it doesnt, there is no I/O and the function just returns -1.

It would seem that you have to send or receive at least one byte. Unless someone else knows how to work around this limitation at the lowest level (and have something similar that works for Windows) - then I'm out of ideas.

krwq:
@philrawlings for Windows if it doesn't work you can just skip the test there. For Linux it might be worth filing and issue there if it's possible or refer to the their docs, worst case scenario we will wait until the kernel is updated.

Ellerbach:
[Triage] @philrawlings is this PR still valid and you still think this should move forward?

scalablecory:
I hope this is still in development!

This API would have been useful in https://github.com/dotnet/iot/pull/1668 -- Sensirion also uses read address NACK to indicate "device busy". Without it, I need to perform double the I/Os to read data.

PR #1530, opened 2021-05-05
Add .NET Standard 2.0 as target

pgrawehr:
I was attempting to use a binding in a project that still uses .NET Framework (and cannot be migrated just yet) and found that this was not possible. This adds cross-build support for most bindings with only minor tweaks. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1530)

### DIFF
--- a/src/devices/BoardLed/BoardLed.cs
+++ b/src/devices/BoardLed/BoardLed.cs
@@ -119,15 +119,7 @@ namespace Iot.Device.BoardLed
 
         private void SetBrightness(int value)
         {
-            if (value < 0)
-            {
-                value = 0;
-            }
-
-            if (value > 255)
-            {
-                value = 255;
-            }
+            value = MathExtensions.Clamp(value, 0, 255);
 
             _brightnessWriter.BaseStream.SetLength(0);
             _brightnessWriter.Write(value);
--- a/src/devices/BoardLed/BoardLed.cs
+++ b/src/devices/BoardLed/BoardLed.cs
@@ -66,7 +66,7 @@ namespace Iot.Device.BoardLed
         {
             Name = name;
             Initialize();
-#if NETCOREAPP2_1
+#if !NET5_0_OR_GREATER
             if (_brightnessReader is null ||
                 _brightnessWriter is null ||
                 _maxBrightnessReader is null ||
@@ -119,7 +119,16 @@ namespace Iot.Device.BoardLed
 
         private void SetBrightness(int value)
         {
-            value = Math.Clamp(value, 0, 255);
+            if (value < 0)
+            {
+                value = 0;
+            }
+
+            if (value > 255)
+            {
+                value = 255;
+            }
+
             _brightnessWriter.BaseStream.SetLength(0);
             _brightnessWriter.Write(value);
             _brightnessWriter.Flush();
@@ -145,7 +154,7 @@ namespace Iot.Device.BoardLed
             _triggerWriter.Flush();
         }
 
-#if !NETCOREAPP2_1
+#if NET5_0_OR_GREATER
         [MemberNotNull(nameof(_brightnessReader), nameof(_brightnessWriter), nameof(_triggerReader), nameof(_triggerWriter), nameof(_maxBrightnessReader))]
 #endif
         private void Initialize()
--- a/src/devices/Common/FrameworkCompatibilityExtensions.cs
+++ b/src/devices/Common/FrameworkCompatibilityExtensions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.IO;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
@@ -45,6 +46,19 @@ namespace Iot.Device
             {
                 buffer[i] = (byte)random.Next();
             }
+        }
+
+        public static void Write(this Stream stream, Span<byte> data)
+        {
+            stream.Write(data.ToArray(), 0, data.Length);
+        }
+
+        public static int Read(this Stream stream, Span<byte> data)
+        {
+            byte[] rawData = new byte[data.Length];
+            int result = stream.Read(rawData, 0, rawData.Length);
+            rawData.CopyTo(data);
+            return result;
         }
 
         /// <summary>
--- a/src/devices/Common/FrameworkCompatibilityExtensions.cs
+++ b/src/devices/Common/FrameworkCompatibilityExtensions.cs
@@ -1,0 +1,85 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+#pragma warning disable CS1591 // Missing documentation (these methods are mostly hidden from the user)
+#if NETFRAMEWORK
+namespace Iot.Device
+{
+    /// <summary>
+    /// .NET Core compatibility helper functions (methods that do not exist in .NET Framework)
+    /// </summary>
+    public static class FrameworkCompatibilityExtensions
+    {
+        public static bool StartsWith(this Span<char> span, string value)
+        {
+            return span.StartsWith(new ReadOnlySpan<char>(value.ToCharArray()));
+        }
+
+        public static bool StartsWith(this Span<char> span, string value, StringComparison comparison)
+        {
+            return span.ToString().StartsWith(value, comparison);
+        }
+
+        public static bool StartsWith(this ReadOnlySpan<char> span, string value, StringComparison comparison)
+        {
+            return span.ToString().StartsWith(value, comparison);
+        }
+
+        public static int CompareTo(this ReadOnlySpan<char> span, string value, StringComparison comparison)
+        {
+            return string.Compare(span.ToString(), value, comparison);
+        }
+
+        public static string GetString(this Encoding encoding, Span<byte> input)
+        {
+            return encoding.GetString(input.ToArray());
+        }
+
+        public static void NextBytes(this Random random, Span<byte> buffer)
+        {
+            for (int i = 0; i < buffer.Length; i++)
+            {
+                buffer[i] = (byte)random.Next();
+            }
+        }
+
+        /// <summary>
+        /// Attempts to add the specified key and value to the dictionary.
+        /// </summary>
+        /// <typeparam name="TKey">Type of key</typeparam>
+        /// <typeparam name="TValue">Type of value</typeparam>
+        /// <param name="dictionary">Dictionary</param>
+        /// <param name="key">Key to add</param>
+        /// <param name="value">Value to add</param>
+        /// <returns>True if the value was added, false otherwise. No exception is thrown</returns>
+        public static bool TryAdd<TKey, TValue>(this Dictionary<TKey, TValue> dictionary, TKey key, TValue value)
+        {
+            if (dictionary.ContainsKey(key))
+            {
+                return false;
+            }
+
+            dictionary.Add(key, value);
+            return true;
+        }
+    }
+}
+
+#pragma warning disable SA1403 // File may only contain a single namespace
+namespace System.Runtime.CompilerServices
+#pragma warning restore SA1403 // File may only contain a single namespace
+{
+    /// <summary>
+    /// Reserved to be used by the compiler for tracking metadata.
+    /// This class should not be used by developers in source code.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class IsExternalInit
+    {
+    }
+}
+#endif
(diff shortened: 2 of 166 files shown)


src/devices/Directory.Build.props
```diff
@@ -3,6 +3,7 @@
   <Import Project="../../eng/Versions.props" />
   <Import Project="../../eng/Analyzers.props" />
   <PropertyGroup>
+    <TargetFrameworksBase>net5.0;netcoreapp2.1;net48;</TargetFrameworksBase>
```
joperezr, line 6: I don't think that adding a net48 configuration is the right thing to do here, instead we should try to add a netstandard2.0 configuration to all projects that support it, and if the build works fine, we can use that instead of the netcoreapp2.1 configuration in case there are no differences between netcoreapp2.1 and netsandard2.0. We already started doing this a while back, but the PR was never completed so perhaps this is a good oportunity to take care of it.

Main advantage of using netstandard2.0 as opposed to net48, is that with it it will be actually making the bindings runnable accross other platforms like Mono, Xamarin, and UWP which will probably be more handy than a net48 configuration.

pgrawehr: I have tried that, but it didn't work. But based on what I have done now, it seems it could actually work. Maybe I'll need your help in getting the packaging stuff right - I can get it to compile, but the way the main Iot.Device.Bindings.dll is built gives me a bit of a headache.

joperezr: Some time ago I merged a PR in Iot.Device.Bindings that will basically get the best target framework for all of the devices for each TFM that Iot.Device.Bindings compiles for, and will try to build  those particular sources for it so in theory this shouldn't be to hard to do now. Work will mostly involve adding netstandard2.0 configs on all devices that support it, and add a netstandard2.0 config to iot.device.bindings. Finally we will just have to add some exclusions into iot.device.bidnings for those devices that won't support netstandard2.0.

src/devices/Directory.Build.props
```diff
@@ -3,6 +3,10 @@
   <Import Project="../../eng/Versions.props" />
   <Import Project="../../eng/Analyzers.props" />
   <PropertyGroup>
+    <!-- Most projects should use this value to build for all possible target frameworks -->
+    <TargetFrameworksBase>net5.0;netcoreapp2.1;netstandard2.0;</TargetFrameworksBase>
+    <!-- If .NET Standard is not supported (and adding support is not useful) use this instead -->
+    <TargetFrameworksReduced>net5.0;netcoreapp2.1</TargetFrameworksReduced>
```
joperezr, line 9: NIT: I would define this one first and then have the one that has netstandard2.0 too be based on this one, that way when we move to target something else the update only goes in one line. Also, I think I would remove these to be more like: DeviceBindingsDefaultTFMs and DeviceBindingsNetCoreTFMs

src/devices/Common/Iot/Device/Graphics/LcdCharacterEncoding.cs
```diff
@@ -102,7 +102,7 @@ public override int GetBytes(char[] chars, int charIndex, int charCount, byte[]
                 }
                 else
                 {
-                    bytes[byteIndex] = _characterMapping.GetValueOrDefault(_unknownLetter);
+                    bytes[byteIndex] = _characterMapping[_unknownLetter];
```
joperezr, line 105: Should we add a check in case this index doesn't exist? that is what GetValueOrDefault used to do

src/devices/Common/MathExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    /// <summary>
+    /// Implementations of some functions missing in older .NET versions
+    /// </summary>
+    public static class MathExtensions
+    {
+        /// <summary>
+        /// Returns val, limited to the range min-max (inclusive)
+        /// </summary>
+        public static double Clamp(double val, double min, double max)
```
joperezr, line 17: It doesn't need to happen now but it would be good to add a few unit tests to these members to make sure they behave as expected.

src/devices/HardwareMonitor/OpenHardwareMonitor.cs
```diff
@@ -657,6 +670,11 @@ public Sensor(ManagementObject instance, string name, string identifier, string?
             /// </summary>
             public SensorType SensorType { get; }
 
+            /// <summary>
+            /// The instance ID. Used to retrieve a new value for the same sensor (repeatedly obtaining the same value from a sensor instance returns the same value)
+            /// </summary>
+            internal int InstanceId { get; }
```
joperezr, line 676: Looks like some of the changes on this binding are not related to just the addition of the configurations. Can we for now only include the changes related to netstandard2.0 configs and then send out a follow up for the HardwareMonitor changes? It just m akes it easier to audit these type of changes in the future.

src/devices/MotorHat/DCMotor3Pwm.cs
```diff
@@ -33,7 +33,7 @@ public override double Speed
             set
             {
                 // Make sure the speed is between -1 and 1
-                _speed = Math.Clamp(value, -1, 1);
+                _speed = value < -1 ? -1 : (value > 1) ? 1 : value;
```
joperezr, line 36: Why not use the Extension you added?

src/devices/Pca9685/Pca9685.cs
```diff
@@ -296,7 +296,17 @@ private void SetOnOffTimeAllChannels(ushort on, ushort off)
         private byte FrequencyToPrescale(double freqHz)
         {
             var desiredPrescale = Math.Round(ClockFrequency / 4096 / freqHz - 1);
-            return (byte)Math.Clamp(desiredPrescale, byte.MinValue, byte.MaxValue);
```
joperezr, line 299: Same here, should we use your extension instead?

joperezr:
Left a few comments and took the liberty to fix the non-Windows build issues, but this is mostly ready to go once the last few comments are addressed. Thanks for taking care of this @pgrawehr

src/devices/Common/Iot/Device/Graphics/LcdCharacterEncoding.cs
pgrawehr: Yes. There is already a check in the factory that typically generates this class, but since the ctor is public, I added a validation there. The "unknown" letter must exist.

src/devices/MotorHat/DCMotor3Pwm.cs
pgrawehr: I sometimes just recoded it, to avoid having to add the extra project reference everyhwere. Either way is fine for me.

src/devices/Common/MathExtensions.cs
pgrawehr: Done

pgrawehr:
Removed the commits that don't belong here and addressed the other findings. 

I have a small problem: I would like to update the DeviceBindingTemplate, too, but that one has not the same Directory.build.props, as it is completely outside of the device tree. Since there are a few similar changes I would like to make to the project files and the template (I want to clean up the test projects as well, since the used versions of xunit and Moq are inconsistent) it would be good if we could just completely exclude that project from build.

src/devices/Directory.Build.props
```diff
@@ -12,6 +14,7 @@
     <LangVersion>9</LangVersion>
     <Nullable>enable</Nullable>
   </PropertyGroup>
+
```
joperezr, line 17:
```suggestion
```
NIT: remove extra line

src/devices/BoardLed/BoardLed.cs
```diff
@@ -119,7 +119,16 @@ private int GetMaxBrightness()
 
         private void SetBrightness(int value)
         {
-            value = Math.Clamp(value, 0, 255);
+            if (value < 0)
```
joperezr, line 122: NIT: Use your extension method instead.

src/devices/Common/FrameworkCompatibilityExtensions.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+#pragma warning disable CS1591 // Missing documentation (these methods are mostly hidden from the user)
+#if NETSTANDARD2_0
```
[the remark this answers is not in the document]
joperezr: I disagree here. Having IfDefs just create maintainability burden (we've had this happen MANY times in dotnet/runtime repo) even when the defines are comming from the SDK, so whenever they are easily avoidable that is preferred. VS has a feature for selecting the configuration you are working with and that will automatically show/hide files that should be excluded. That is the pattern that we have followed with the rest of the repo, so I would rather just exclude the file at least for consistency-sake. If we find that the VS experience is not there for this scenario, we can then either file a request for VS to add support for this, or (in the worst case scenario) move the  whole repo to use this pattern, but I think that what should be the driver here is consistency.

src/devices/SocketCan/CanRaw.cs
```diff
@@ -55,11 +56,10 @@ public void WriteFrame(ReadOnlySpan<byte> data, CanId id)
             {
                 Span<byte> frameData = new Span<byte>(frame.Data, data.Length);
                 data.CopyTo(frameData);
-            }
 
-            ReadOnlySpan<CanFrame> frameSpan = MemoryMarshal.CreateReadOnlySpan(ref frame, 1);
-            ReadOnlySpan<byte> buff = MemoryMarshal.AsBytes(frameSpan);
-            Interop.Write(_handle, buff);
+                byte* buff = (byte*)&frame;
```
joperezr, line 60: Where you able to test this? Is the problem here that Netstandard2.0 doesn't have these MemoryMarshal? Aren't these methods available through the System.Memory package?

src/devices/UFireIse/UFireIse.cs
```diff
@@ -310,6 +310,12 @@ private float ReadFloat(Register register)
         private void WriteRegister(Register register, float data = 0)
         {
             // μFire ISE (Ion Specific Electrode) is 0 as default
+#if NETSTANDARD2_0
+            byte[] bytes = BitConverter.GetBytes(data);
+            ChangeRegister(register);
+            _device.Write(bytes);
+            DelayHelper.DelayMilliseconds(IseCommunicationDelay, allowThreadYield: true);
```
joperezr, line 317: This is no longer throwing the exceptions right?

src/devices/Common/MathExtensions.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// </summary>
+    public static class MathExtensions
+    {
+        /// <summary>
+        /// Returns val, limited to the range min-max (inclusive)
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static double Clamp(double val, double min, double max)
+        {
+#if NETSTANDARD2_0
```
joperezr, line 21: While I'm fine with this for now, an alternative would be to have two implementations of this file that are conditionally compiled, one for the downlevel implementation, and the other one that just forwards the calls to Math to avoid all these ifdefs.

joperezr (approved):
Left some last few comments but this is ready to go once those are answered/addressed.

src/devices/SocketCan/CanRaw.cs
pgrawehr: No, I could not test this, but I'm quite certain that it's equivalent. I did not look for an implementation of these methods because the changed code is simpler than the old. This converted to two different ReadOnlySpan types and finally to a byte*, while you can just directly take the address of the frame.

src/devices/UFireIse/UFireIse.cs
pgrawehr: True, but that exception cannot really happen. The only situation where `TryWriteBytes` returns false is if there's not enough room in the buffer.

src/devices/SocketCan/CanRaw.cs
joperezr: Ok, I checked and they don't seem to be available on System.Memory package for NS2.0. To be honest, it feels a bit weird to change this when we haven't tested it at all even when we have high confidence on the change, but for now let's keep it and if folks using the binding find an issue here they can log the issue and we can investigate then.

pgrawehr: The binding was originally provided by @krwq I think. So maybe he has the hardware available to test it?

src/devices/Common/FrameworkCompatibilityExtensions.cs
```diff
@@ -0,0 +1,86 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+#pragma warning disable CS1591 // Missing documentation (these methods are mostly hidden from the user)
+#if NETSTANDARD2_0
```
joperezr, line 10: Looks like you still haven't removed this Ifdef and instead conditionally compile the file. Same goes for IsExternalInit. I think that is the last remaining feedback I have and this should be good to go.

PR #1538, opened 2021-05-15
Disable SPI CS when it isn't used

Hubble1942:
In applications where the chip-select line is controlled 'manually', the driver interferes by re-/setting the chip-select line as well. This behaviour is disabled by configuring the devices mode with the `SPI_NO_CS`-flag set.


pgrawehr (approved):
Looks good.

PR #1546, opened 2021-05-25
Minor documentation update

pgrawehr:
Update documentation

- Add extra links
- Add Netstandard 2.0 support
- Fix some typos

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1546)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ The repository also contains [Iot.Device.Bindings](https://www.nuget.org/package
 **NOTE**: This repository is still in experimental stage and all APIs are subject to changes.
 
 ## Hardware requirements
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -11,9 +11,14 @@ The repository also contains [Iot.Device.Bindings](https://www.nuget.org/package
 
 **NOTE**: This repository is still in experimental stage and all APIs are subject to changes.
 
+## Hardware requirements
```
Ellerbach, line 14: in Markdown, you are supposed to have one CR before and after each title (exception for the level one when present at first line)

Ellerbach (approved):
Good addition.

README.md
danmoseley: we could consider adding markdown lint (basically port https://github.com/dotnet/runtime/pull/40887). it is possible to suppress its lower value warnings.

pgrawehr: @Ellerbach I'm not so familiar with this syntax yet, but when this is a problem, why does it render correctly?

Ellerbach: Agree @danmoseley . I'm quite a big fan of linter. We do have for C#, we should have for Markdown as well. I'm usually very strict and don't remove much rules. We don't have to be as hardcore as I usually am.
@pgrawehr, like for any language, Markdown has rules, some are minors, like the one I point out as most site like GitHub are extremely tolerant and will always try to render properly something. But it's not the case of all renders. So better have rules more strict so we do increase chances to have the page always rendering nicely

pgrawehr: No opposition from my end there. I thought that markdown was an invention by github itself and so its implementation was the reference by itself. Shouldn't block this PR from merging now, though.

joperezr (approved):

PR #1561, opened 2021-05-28
Add libgpiod rising and falling edge detection

flitjens:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1561)

### DIFF
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -111,6 +111,24 @@ internal partial class Interop
         internal static extern void gpiod_line_release(IntPtr lineHandle);
 
         /// <summary>
+        /// Request rising edge event type notifications on a single line.
+        /// </summary>
+        /// <param name="line">GPIO line handle</param>
+        /// <param name="consumer">Name of the consumer.</param>
+        /// <returns>0 the operation succeeds, -1 on failure.</returns>
+        [DllImport(LibgpiodLibrary, SetLastError = true)]
+        internal static extern int gpiod_line_request_rising_edge_events(SafeLineHandle line, string consumer);
+
+        /// <summary>
+        /// Request falling edge event type notifications on a single line.
+        /// </summary>
+        /// <param name="line">GPIO line handle</param>
+        /// <param name="consumer">Name of the consumer.</param>
+        /// <returns>0 the operation succeeds, -1 on failure.</returns>
+        [DllImport(LibgpiodLibrary, SetLastError = true)]
+        internal static extern int gpiod_line_request_falling_edge_events(SafeLineHandle line, string consumer);
+
+        /// <summary>
         /// Request all event type notifications on a single line.
         /// </summary>
         /// <param name="line">GPIO line handle</param>
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
@@ -89,7 +89,7 @@ namespace System.Device.Gpio.Drivers
         {
             if ((eventTypes & PinEventTypes.Rising) != 0 || (eventTypes & PinEventTypes.Falling) != 0)
             {
-                LibGpiodDriverEventHandler eventHandler = _pinNumberToEventHandler.GetOrAdd(pinNumber, PopulateEventHandler);
+                LibGpiodDriverEventHandler eventHandler = _pinNumberToEventHandler.GetOrAdd(pinNumber, PopulateEventHandler(pinNumber, eventTypes));
 
                 if ((eventTypes & PinEventTypes.Rising) != 0)
                 {
@@ -107,7 +107,7 @@ namespace System.Device.Gpio.Drivers
             }
         }
 
-        private LibGpiodDriverEventHandler PopulateEventHandler(int pinNumber)
+        private LibGpiodDriverEventHandler PopulateEventHandler(int pinNumber, PinEventTypes eventTypes)
         {
             lock (_pinNumberLock)
             {
@@ -120,7 +120,7 @@ namespace System.Device.Gpio.Drivers
                     _pinNumberToSafeLineHandle[pinNumber] = pinHandle;
                 }
 
-                return new LibGpiodDriverEventHandler(pinNumber, pinHandle!);
+                return new LibGpiodDriverEventHandler(pinNumber, pinHandle!, eventTypes);
             }
         }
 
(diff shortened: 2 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1561) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1561) before we can accept your contribution.<br/><br/>:x: flitjens [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1561)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1561) it.</sub>

pgrawehr:
Can you add test cases to `LibGpiodDriverTests.cs` to show the new/improved use cases?

flitjens:
@pgrawehr of course I will write a test if necessary, but in my opinion these tests should be present already since I did not add new functionality (only improved existing functionality).
I also see that there are errors in the Linux Debug and Release pipeline but they seem to be infrastructure issues. Can you tell me what has happened?

flitjens:
Ok, I can reproduce the problem and will solve it.

flitjens:
@pgrawehr I am still working on it. I focus now on the failure of the AddCallbackRemoveCallbackTest and I see the problem. However my solution still not behave as expected. I continue working on it and provide extra tests.

Ellerbach:
[Triage] @flitjens as soon as you'll have more info, please let us know.

flitjens:
> [Triage] @flitjens as soon as you'll have more info, please let us know.

Hi @Ellerbach and @pgrawehr 

Let me first describe the problem in a bit more detail and pinpoint the problem which is currently present in the gpio driver. The problem occurs because LibGpiodDriverEventHandler uses gpiod_line_request_both_edges_events in all situations where the user registers a callback for pin value change. It doesn’t matter if the user registers the callback just for rising or falling, it all results in a detection for both edges.
So what happens when we register a callback for only falling edges in the ads1115 which has a very gpiod_line_event_wait short negative pulse of approximately 8us. In the EventDetectionTask we wait for a line event but since we underwater registered both rising and falling edges we receive both. So at the moment we read the line event value the line is already up because the time between detection and reading the value takes more than 8us. And we registered only a callback for ValueFalling in the LibGpiodDriverEventHandler meaning that the ValueRising PinChangeEventHandler is null so no pin change is reported and that is the reason why the ADS1115 does work with the current implementation. Of course I might be wrong but this is the behavior I experienced.

As far as I understand the problem correctly there are several options:
1.	Use gpiod_line_request_falling_edge_events when a falling event callback is registered.
This is the option I tried first but ran into a problem when I want to add the rising event callback. The problem with gpiod is that once you call gpiod_line_request_falling_edge_events, gpiod_line_request_rising_edge_events or gpiod_line_request_both_edges_events  you can’t switch to another one without canceling the first. I am still working on this scenario to see if there are alternatives.

2.	The best option, in my opinion, is to register up front whether we want to detect rising, falling or both edges and raise an exception if the user tries to register a callback for an event which is not configured. I know this is a breaking API change, and is a pretty big change, so I understand that this is not the option we want right now. However, I think this is the most solid solution.

3.	Changing gpio settings via the memory mapped registers
I think it should be possible to use gpiod_line_request_both_edges_events but underwater change gpio settings via memory mapped registers. What I don’t like about this option is that it is a hacky solution, which is never a good option.

4.	Do nothing
This is always an option.

What do you think and what is the normal procedure for these kind of issues.

Kind regards,
Frank Litjens

pgrawehr:
@flitjens If your observation is correct (which does sound reasonable) then I guess yes, something like 1 or 2 would be required. If I understand you correctly, you proved that when you use the `gpiod_line_request_falling_edge_events` the ADS1115 events are correctly detected, even without additional hardware. 

I think the reason the current solution was used is really because this allows changing the event modes dynamically. Throwing an exception when one attempts to change the event type when one is already registered doesn't sound like a too bad idea, because that's probably a rare situation. Typically the user knows what events he needs when first registering a callback for a pin. 

We'll have to discuss the possible options.

Ellerbach (approved):
Looks good to me. Thanks for finding how to improve this behavior!

Ellerbach:
[Triage] We will rerun the tests and go for solution 2 mentioned above once they'll pass.

And it is not per say a break in the API themselves, they are still the same.  The break is more on the behavior. That said, we don't really see scenarios where you'll have to change your mind on what type you'll register. You still can unregister and reregister, so you still can do it. That can be the case for some specific scenarios, mainly when you need to wait for something at some point and then to release resources, stop doing it. In all cases, you can always register to what you want. And at least, it should then work much better and more precise than before.

Ellerbach:
[Triage] Still having issues with the hardware build and test, still waiting for this to be fixed.

Ellerbach:
> @pgrawehr I am still working on it. I focus now on the failure of the AddCallbackRemoveCallbackTest and I see the problem. However my solution still not behave as expected. I continue working on it and provide extra tests.

As now the hardware pipeline build/test is working, we can see this issue. If you still can work to solve it, it would be perfect. Thanks!

flitjens:
@pgrawehr , @Ellerbach can I assist in some way?

pgrawehr:
@flitjens Yes, sure. It appears that your change has broken a test, see here: https://helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-iot-refs-pull-1561-merge-9e45a825dabc4d0a94/System.Device.Gpio.Tests.dll/console.7f298b4b.log?sv=2019-07-07&se=2021-07-21T15%3A55%3A31Z&sr=c&sp=rl&sig=SZI4%2FjCQQlhkMGaffXU9u66hyvQ17tZady2%2FI%2BRpsVc%3D

You can ignore the failing windows builds for now, but the linux builds should be working again. If you implement your solution from above 

> Use gpiod_line_request_falling_edge_events when a falling event callback is registered.
This is the option I tried first but ran into a problem when I want to add the rising event callback. The problem with gpiod is that once you call gpiod_line_request_falling_edge_events, gpiod_line_request_rising_edge_events or gpiod_line_request_both_edges_events you can’t switch to another one without canceling the first. I am still working on this scenario to see if there are alternatives.

an exception is expected if one changes the edge mode without first deregistering. This is fine (we agreed that this is a rare scenario), but you should add a test that verifies this behavior (which also acts as documentation that an exception in this situation is intentional).

flitjens:
@Ellerbach @pgrawehr 
Hi Laurent and Patrick I understand that you would like to go for option 2, but this requires some additional modifications to the code. What we can do is the following:

1.  The user can specify an edge on which it would like to be notified when the user opens a pin. The default could be to be notified on both edges, by which it is not an API change. But the user is able to specify one which can not be changed without closing the pin.
2. An exception should be thrown when the user registers a callback for an event which is specified when the pin was opened.

What do you think?

pgrawehr:
@flitjens No, actually that was option 1. At the time where `PopulateEventHandler` is called (from `AddCallbackForPinValueChangedEvent`) we know what events we need. So the line

```
LibGpiodDriverEventHandler eventHandler = _pinNumberToEventHandler.GetOrAdd(pinNumber, PopulateEventHandler);
```
could be changed to something like
```
LibGpiodDriverEventHandler eventHandler = _pinNumberToEventHandler.GetOrAdd(pinNumber, p => PopulateEventHandler(p, eventTypes));
```
This should go without any API change, and without the user already giving the required mode when opening the pin.

Ellerbach:
[Triage] @flitjens did you have time to review the proposition @pgrawehr did?

flitjens:
Hi @Ellerbach and @pgrawehr 

Sorry for the delay. So far I did not have time to look at the Patrick's proposal and I will be on holiday for the next 4 weeks. Is it ok with you that I pickup in August?

PR #1567, opened 2021-06-04
Fixes #1539 Linux-arm LIBGPIOD: PinValueChangedEvent consumes too much CPU

### LINKED ISSUE #1539: Linux-arm LIBGPIOD: PinValueChangedEvent  consumes too much CPU
**Describe the bug**

When registering a callback for pin events, then the CPU goes up to about 12% CPU usage just for a single pin:

**Steps to reproduce**

```csharp
 class Program
    {
        {
            var gpioController = new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver(4));

            gpioController.OpenPin(9, PinMode.Input);
            gpioController.RegisterCallbackForPinValueChangedEvent(9, PinEventTypes.Falling | PinEventTypes.Rising, (sender, e) =>
            {
                Console.WriteLine("Pin Event: " + e.PinNumber.ToString() + " - " + e.ChangeType.ToString());
                if (e.ChangeType == PinEventTypes.Rising || e.ChangeType == PinEventTypes.Falling)
                {

                   
                }
            });


            Console.ReadLine();
        }
    }
```

**Expected behavior**

CPU should stay near 0%

**Actual behavior**

With my sample code above CPU is about 12% on a linux-arm embedded board. It gets worser if there are eventhandler on more than one pin...

**Versions used**

.NET 5.0.5 self-contained
 
Version of System.Device.Gpio package 1.4

The CPU usage decreases to about 0-1 %  if I use a longer wait delay here -> instead of 1_000_000 I tried 50_000_000:
image

if increasing the wait time has no drawbacks, then it would be better to increase it by default or at least at a parameter for this.

Alex-111:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1567)

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/LibgpiodDriverEventHandler.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/LibgpiodDriverEventHandler.cs
@@ -51,7 +51,7 @@ namespace System.Device.Gpio.Drivers
                     TimeSpec timeout = new TimeSpec
                     {
                         TvSec = new IntPtr(0),
-                        TvNsec = new IntPtr(1000000)
+                        TvNsec = new IntPtr(50_000_000)
                     };
 
                     WaitEventResult waitResult = Interop.libgpiod.gpiod_line_event_wait(pinHandle, ref timeout);


src/System.Device.Gpio/System/Device/Gpio/LibgpiodDriverEventHandler.cs
```diff
@@ -51,7 +51,7 @@ private Task InitializeEventDetectionTask(CancellationToken token, SafeLineHandl
                     TimeSpec timeout = new TimeSpec
                     {
                         TvSec = new IntPtr(0),
-                        TvNsec = new IntPtr(1000000)
+                        TvNsec = new IntPtr(50_000_000)
```
danmoseley, line 54: Aside, if these aren't pointers and you're compiling with C# 9.0 perhaps it would be cleaner to use `nint` or `nuint`

Alex-111: Please understand I just want to fix a big issue for smaller devices. Your comment is about refactoring the code which was not in my scope. Maybe there are even more places where IntPtr is used....THis optimization should happen seperately...

joperezr:
Extending the time we have between each poll for new events might make us miss many events that people today may already rely on. Can you explain a bit more which type of device you are working with and why is this new value ok in terms of threshhold? perhaps share some profile data that shows the diff on CPU usage. 

I understand you just want to fix the issue for this particular device, but please do also note that changing this, while it may be ok for the devices you are working with and won't miss many events, might be breaking for many other devices that do need the polling to happen much more frequently in order to get all of the events they expect.

src/System.Device.Gpio/System/Device/Gpio/LibgpiodDriverEventHandler.cs
joperezr: TvNsec should be a long, I do agree we should do a pass on the whole repo for cases where the new native types can be used, and we should do that in a separate PR.

Ellerbach: [Triage] @Alex-111, 1 ms maybe to small, 50 ms maybe to high, can you give us a rational and where you picked that number? 
Any comparison with different values? Is this 50 the optimal one?
What is the moment where the CPU usage is really increasing?
Note: for the comments on the type for TimeSpec, we will address that in a different PR, there maybe other adjustments elsewhere.
And thanks again for raising this PR.

Alex-111: My understanding is that the timout time is the time how long the waiting thread blocks. In case of an event the thread should be released anyway regardless the timout time, so even higher values should not be problematic.

The problem of high CPU usage grows when the number of monitored pins increases, so for a typical emebdded linux board I've found out that 50ms is a good value. 
But I also have to mention that I'm not a linux API expert. I'm not sure if increasing the value could have side effects for edge cases.
For me  the change dropped CPU usage to about 0-1%, instead of 12% for one single pin!

Alex-111:
@joperezr 
Why do you think we miss events in case of setting the timeout to a higher value?

My understanding is that the timout time is the time how long the waiting thread blocks. In case of an event, the blocking call should be interuppted regardless the timeout value? Isn't that the case in your environment?

src/System.Device.Gpio/System/Device/Gpio/LibgpiodDriverEventHandler.cs
pgrawehr: Sounds reasonable. The only downside of making the delay too long is that the termination will take extra time. But unless you register-deregister-register an event all the time (which is probably a rare scenario) 50ms shouldn't really hurt.

pgrawehr (approved):

Ellerbach (approved):
[Triage] after the last elements provided, seems like 50 ms is a good timing.

PR #1576, opened 2021-06-24
Add PinValue.IsLow/IsHigh

smdn:
`PinValue` struct already have `Equals()`, `==`/`!=`, and explicit type conversion to `bool`.
But these are little bit redundant and confusing in `if` statements like the codes listed below.

So added new property `IsLow`/`IsHigh` to `PinValue` struct in this PR.

Use cases and comparison:
```cs
// vs. Equals()
PinValue signalA, signalB;

if (signalA.Equals(PinValue.High)) // 🙁This looks redundant.
if (signalA.IsHigh) // 😺It's clear.

if (signalA.Equals(PinValue.High) && signalB.Equals(PinValue.Low)) // 🙁This looks redundant.
if (signalA.IsHigh && signalB.IsLow) // 😺It's clear.

// vs. explicit operator bool
// (in case of positive logic and negative logic value)
PinValue positiveLogicSignal, negativeLogicSignal;

if ((bool)positiveLogicSignal) // 😺This looks fine.
if (positiveLogicSignal.IsHigh) // 😺It's more clear.

if (!(bool)negativeLogicSignal) // 🤔This looks confusing. (If the value is high? or low?)
if (negativeLogicSignal.IsLow)  // 😺It's clear.
```

Similar examples:
- BigInteger.Zero/One and BigInteger.IsZero/IsOne (System.Numerics)
- Point.Empty and Point.IsEmpty (System.Drawing)

(Update)
Purpose of this proposal:
- To increase the way to clarify what the code is truely intended to do.
- To increase the way to write the code friendly to the readers. (especially for beginners and Visual Basic users)
- To provide the above two points as library properties, rather than by defining extension methods by the users individually.

(Update2)
Other naming candidates:
- PinValue.IsZero/IsOne
- PinValue.IsFalse/IsTrue

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1576)

### DIFF
--- a/src/System.Device.Gpio.Tests/PinValueTests.cs
+++ b/src/System.Device.Gpio.Tests/PinValueTests.cs
@@ -15,6 +15,12 @@ namespace System.Device.Gpio.Tests
         {
             Assert.True(PinValue.High.Equals(1));
             Assert.True(PinValue.Low.Equals(0));
+
+            Assert.True(PinValue.High.IsHigh);
+            Assert.False(PinValue.High.IsLow);
+
+            Assert.True(PinValue.Low.IsLow);
+            Assert.False(PinValue.Low.IsHigh);
         }
 
         [Fact]
@@ -37,6 +43,22 @@ namespace System.Device.Gpio.Tests
         {
             Assert.Equal(PinValue.Low, !PinValue.High);
             Assert.Equal(PinValue.High, !PinValue.Low);
+        }
+
+        [Fact]
+        public void VerifyIsLowOrIsHigh()
+        {
+            Assert.True(((PinValue)0).IsLow);
+            Assert.False(((PinValue)0).IsHigh);
+
+            Assert.True(((PinValue)1).IsHigh);
+            Assert.False(((PinValue)1).IsLow);
+
+            Assert.True(((PinValue)int.MaxValue).IsHigh);
+            Assert.False(((PinValue)int.MaxValue).IsLow);
+
+            Assert.True(((PinValue)int.MinValue).IsHigh);
+            Assert.False(((PinValue)int.MinValue).IsLow);
         }
     }
 }
--- a/src/System.Device.Gpio/System/Device/Gpio/PinValue.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/PinValue.cs
@@ -26,6 +26,18 @@ namespace System.Device.Gpio
         public static PinValue Low => new PinValue(0);
 
         /// <summary>
+        /// Indicates whether the value of the current <see cref="PinValue"/> is <see cref="High"/> or not.
+        /// </summary>
+        /// <value><see langword="true"/> when the value is <see cref="High"/>, <see langword="false"/> when the value is <see cref="Low"/>.</value>
+        public bool IsHigh => _value != 0;
+
+        /// <summary>
+        /// Indicates whether the value of the current <see cref="PinValue"/> is <see cref="Low"/> or not.
+        /// </summary>
+        /// <value><see langword="true"/> when the value is <see cref="Low"/>, <see langword="false"/> when the value is <see cref="High"/>.</value>
+        public bool IsLow => _value == 0;
+
+        /// <summary>
         /// Implicit conversion from int. 0 means <see cref="Low"/>, everything else means <see cref="High"/>.
         /// </summary>
         /// <param name="value">Value to set</param>


Ellerbach:
[Triage] You can already do:

```csharp
if (signalA && !signalB)
```

It is at least from our point of view even simpler than your IsHigh or IsLow property proposal. 

If you really feel you need them for a reason we are missing please let us know.

You can for you make a function extension if that's better for you. They'll look like

```csharp
public static class GpioExtension
{
  public static bool IsHigh(this PinValue pin) => pin;
  public static bool IsLow(this PinValue pin) => !pin;
}
```

smdn:
Description updated.

@Ellerbach Sure, we can write like `if (signalA && !signalB)` in this case.

But the point of this PR is not 'easy to write', but 'easy to read', or 'easy to understand'.
I want to express and clarify what the code is truely intended to do.

For example, the code `if (signalA.IsHigh && signalB.IsLow) { DoSomething() }` can be read as "if signal-A is HIGH and signal-B is LOW then do something".
It is friendly to the reader of the code.

It also might be friendly to the Visual Basic users.
ex: `If signalA AndAlso Not signalB Then ...` vs `If signalA.IsHigh AndAlso signalB.IsLow Then ...`

Also I've tried and implemented with 'extension method' approach already. But I have to call them like `if (signalA.IsHigh())`. With extension methods, the extra `()` is needed. This is little bit annoying🙁

So I want this to be supported as a *true* property.
(If C# had supported extension property, I would have used it. But unfortunately, that is still under planning...)

krwq:
@smdn I kinda disagree. I read it as `if there is signal A and no signal B` - I personally don't even like terms `high` and `low` as they refer to historically broken logic since electrons flow from low to high

Ellerbach:
> So I want this to be supported as a _true_ property.
> (If C# had supported extension property, I would have used it. But unfortunately, that is still under planning...)

The best you'll get so far is by your own extension as a function. It's 2 more character while you're already typing 6 or 7 extra ones for the .IsHigh or .IsLow. Auto completion is Visual Studio will drastically reduce this to 2 or 3. 

And even if I really understand your willingness to make it more clear, it seems that there is no real agreement here to add it. And it's is not blocking at all, there is the possibility to have the function extension.

smdn:
And few more question.
Are there any documentations, templates or guidelines for this kind of API suggestion? (not for device bindings, for API expansion like this PR)

Ellerbach:
> Can I keep this PR open? Or should I close this PR?

I, think we can keep it open for 10 days or so to see if any other feedback is provided. If then nothing else is provided, we will then close it. And if after it's closed, there is still a debate and more users want a feature like this one, you can always submit a new one!

smdn:
Thank you! Then I'll leave it open until it's closed.

Ellerbach:
> Are there any documentations, templates or guidelines for this kind of API suggestion? (not for device bindings, for API expansion like this PR)

@krwq I'm sure you have a pointer on some page where that's explained.

krwq:
@smdn for coding guidelines good start is https://github.com/dotnet/runtime/tree/main/docs/coding-guidelines for the most comprehensive coding guidelines our former team member Krzysztof Cwalina has wrote a book called "Framework Design Guidelines" - I can't remember the specific rule but in here the specific guideline is about having multiple members meaning the same thing - that usually confuses users - it's usually better to stay with not ideal name than adding second one.

IMO it's better to close this PR since it's most like won't get merged but we can continue design conversation if you want (might be better to open a separate issue (labeled as question) for that as some folks might ignore notification from closed PRs).

With all good intention about naming GPIO APIs I think in your app you should not really use GPIO directly too much - you should write some kind of helper method giving the GPIO pin a semantic meaning, i.e.:

```csharp
bool TooMuchWaterInTheContainer() => _waterLevelSensorPin == PinValue.High;
```

that usually will make your code much simpler and easier to understand - the syntax for specific low level concept doesn't matter that much.

smdn:
@krwq 
> bool TooMuchWaterInTheContainer() => _waterLevelSensorPin == PinValue.High;

Yes, I also think this kind of approach is a good way to give semantics.
And I wish I could do like below even if it is a helper method. That is my main motivation.
```cs
bool TooMuchWaterInTheContainer() => _waterLevelSensorPin.IsHigh;
```

Anyway, thank you for pointing the documents and explanation!
Now I understand a part of what you concerned.

I think this addition may not be confusing as it looks in my view.
But it is true that this addition will provide same functionality with different name which already exists.
I also agree that it should be avoided in general, especially in API design.

I will resubmit this proposal as an issue and will close this PR after submit, as you suggested.
And I would like to continue to ask for comments from other users at that issue.

Ellerbach:
[Triage] So as per this discussion, waiting to close this PR for you, @smdn, to open an issue. It will be more clean like this. Thanks again for driving this discussion. Note, we'll close the PR anyway in 7 days as previously agreed as well.

smdn:
Closed this PR and moved to issue #1583.
@Ellerbach, @krwq Thank you for your comments and suggestions!

PR #1579, opened 2021-07-01
Adding Ultralight NFC card support

Ellerbach:
This PR is about adding Ultralight NFC card support:
- Native support in all 3 NFC readers PN532, PN5180 and MFRC522
- Full Ultralight feature support:
  - configuration,
  - Fast read, read, write
  - Authentication - except for Ultralight C 
  - Full family support, note: not all cards/I2C card has been tested, some side effects and bugs can be present
  - Counter
- NDEF full support for Ultralight cards
- Various bug fixing in readers
- Documentation
- Added a full sample in all the reader samples

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1579)


src/devices/Mfrc522/Mfrc522.cs
```diff
@@ -902,35 +862,92 @@ public override int Transceive(byte targetNumber, ReadOnlySpan<byte> dataToSend,
...
+            status = CalculateCrc(commandToSend.AsSpan(0, dataToSend.Length), commandToSend.AsSpan(dataToSend.Length));
+            if (status != Status.Ok)
+            {
+                return -1;
+            }
+
+            status = SendAndReceiveData(MfrcCommand.Transceive, commandToSend, receivedBuffer);
+            if (status != Status.Ok)
+            {
+                Debug.WriteLine($"{status}");
```
pgrawehr, line 919: Use the logger for this.

src/devices/Card/Ultralight/UltralightCard.cs
```diff
@@ -0,0 +1,992 @@
...
+
+        /// <summary>
+        /// The block number to authenticate or read or write
+        /// </summary>
+        public byte BlockNumber { get; set; }
+
+        /// <summary>
+        /// The Data which has been read or to write for the specific block
+        /// </summary>
+        public byte[]? Data { get; set; }
```
pgrawehr, line 51: This looks like it should actually be a parameter to all the read/write functions, and clearly not a public writable property.

src/devices/Card/Ultralight/UltralightCard.cs
```diff
@@ -0,0 +1,992 @@
...
+
+        /// <summary>
+        /// The command to execute on the card
+        /// </summary>
+        public UltralightCommand Command { get; set; }
```
pgrawehr, line 56: Probably same for this

src/devices/Card/Ultralight/UltralightCard.cs
```diff
@@ -0,0 +1,992 @@
...
+
+        /// <summary>
+        /// Authentication key
+        /// </summary>
+        public byte[] AuthenticationKey { get; set; } = DefaultPassword;
+
+        /// <summary>
+        /// UUID is the Serial Number, called MAC sometimes
+        /// </summary>
+        public byte[]? SerialNumber { get; set; }
```
pgrawehr, line 71: This is writable?

pgrawehr (commented):
Some comments:
- Visibility of elements (methods, properties, but also whole enums) should be reviewed. Device commands should not need to be public.
- A set of unit tests would be nice. Since the CardTransceiver is an abstract class, it should be mockable.

src/devices/Mfrc522/Mfrc522.cs
Ellerbach: Good catch! A left over!

src/devices/Card/Ultralight/UltralightCard.cs
Ellerbach: It's unfortunately a bit more complicated than that. I used the same mechanism for the Mifare card. Yes, there are ways to make it more transparent but then it won't allow specific scenarios. For example, the way I did it for Mifare allowed user to have a minimum support for Ultralight (at least read and write). If I would have been doing that by hiding this part, it would not have been possible.
The higher level part, with NDEF for example totally hide all this complexity and this low level exposure.

src/devices/Card/Ultralight/UltralightCard.cs
Ellerbach: And the same explanation...

src/devices/Card/Ultralight/UltralightCard.cs
Ellerbach: Yes if you want to create a virtual card.

src/devices/Card/Ultralight/UltralightCard.cs
pgrawehr: If I see it correctly, UltraLightCard is used directly by clients. You use it as high-level interface in your examples. Maybe you should add (in addition to the examples) a bit of documentation describing the overal structure and which high-level interfaces should be used for what. As a newbie, I would be quite confused as to where to start.

Ellerbach: Are the 3 samples not covering enough elements from the 3 readers?
I agree that NFC card are confusing. They really are. Most people are using it to get the ID and NDEF.

src/devices/Card/Mifare/MifareCard.cs
```diff
@@ -1158,8 +1155,9 @@ public bool TryReadNdefMessage(out NdefMessage message)
                     // 62 sectors, 3 blocks, 16 bytes
                     cardSize = 2976;
                     break;
+                /* We don't support those
                 case MifareCardCapacity.Mifare300:
-                case MifareCardCapacity.Unknown:
+                case MifareCardCapacity.Unknown */
```
krwq, line 1160: I'd recommend throwing NotSupportedException for these

src/devices/Card/Ndef/NdefMessage.cs
```diff
@@ -18,6 +18,70 @@ public class NdefMessage
...
+        {
+            int idx = 0;
+            // Check if we have 0x03 so it's a possible, NDEF Entry
+            while (idx < toExtract.Length)
+            {
+                if (toExtract[idx++] == 0x03)
+                {
+                    break;
+                }
+            }
```
krwq, line 36: I'm confused by this. In any buffer with random bytes there is a chance there will be 3 somewhere in there. Is there any reliable way to know at which position the 3 should be?

src/devices/Card/Ndef/NdefMessage.cs
```diff
@@ -18,6 +18,70 @@ public class NdefMessage
...
+        /// <summary>
+        /// Extract an NDEF message from a raw byte array
+        /// </summary>
+        /// <param name="toExtract">The byte array where the message is</param>
+        /// <returns>A byte array containing the message itself</returns>
+        public static byte[]? ExtractMessage(Span<byte> toExtract)
+        {
+            var (idx, size) = GetStartSizeNdef(toExtract);
+            // Finally check that the end terminator TLV is 0xFE
+            var isRealEnd = toExtract[idx + size] == 0xFE;
```
krwq, line 72:
```suggestion
            bool isRealEnd = toExtract[idx + size] == 0xFE;
```

src/devices/Card/Ultralight/AuthenticationConfiguration.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Is user configuration permanently locked against write access, except PWD and PACK
+        /// </summary>
+        public bool IsWrittenLocked { get; set; }
+
+        /// <summary>
+        /// Maximum number of possible try, 0 = disabled, 1 to 7 enabled
+        /// </summary>
+        public byte MaximumNumberOfPossibleTry { get; set; }
```
krwq, line 36:
```suggestion
        public byte MaximumNumberOfPossibleTries { get; set; }
```

src/devices/Card/Ultralight/AuthenticationConfiguration.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Gets or sets the authentication page requirement
+        /// </summary>
+        /// <remarks>If the page is higher than the capacity, it means no authentication required.</remarks>
+        public byte AuthenticationPageRequirement { get; set; }
+
+        /// <summary>
+        /// Is user configuration permanently locked against write access, except PWD and PACK
+        /// </summary>
+        public bool IsWrittenLocked { get; set; }
```
krwq, line 31:
```suggestion
        public bool IsWritingLocked { get; set; }
```

src/devices/Card/Ndef/NdefMessage.cs
Ellerbach: The check is done later with the size of the buffer, and the terminator. I do agree that it's not perfect, that said, in theory, the buffer should start with 0x03 or being part of the very first one. The NDEF specifications do not gives a lot of flexibility for that. So it has to be there or it's not an NDEF buffer (0x03 then length and after length 0xFE)

src/devices/Card/Ultralight/MirrorConfiguration.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    /// <summary>
+    /// The Mirror Configuration
+    /// </summary>
+    public class MirrorConfiguration
+    {
+        /// <summary>
+        /// The size in bytes of the UID Mirror
+        /// </summary>
+        public const byte UidMirrorSize = 14;
```
krwq, line 20: should these be internal?

src/devices/Card/Ultralight/NfcCounterConfiguration.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
```
krwq, line 4: nit: consider removing unnecessary namespaces from all files

src/devices/Card/Ultralight/README.md
```diff
@@ -0,0 +1,159 @@
...
+This class supports Ultralight cards. They are RFID cards responding to ISO 14443 type A. You need a specific card reader like PN532, PN5180 or MFRC522 to read, write those kind of cards.
+
+## Creating a card and reading it
+
+You will find detailed examples for PN532 [here](../../Pn532/samples), for MFRC522 [here](../../Mfrc522/samples) and for PN5180 [here](../../Pn5180/samples).
+
+You'll need first to get the card from an RFID reader. The example below shows how to do it with a MFRC522 and read all the sectors, read the configuration data, and print all the sector data information, read NDEF, write NDEF.
+
+```csharp
+var ultralight = new UltralightCard(mfrc522!, 0);
```
krwq, line 12: nit: indent

krwq: the `!` after `mfrc522` is kinda confusing here IMO

src/devices/Card/Ultralight/README.md
```diff
@@ -0,0 +1,159 @@
...
+    {
+        ultralight.BlockNumber = (byte)block; // Safe cast, can't be more than 255
+        ultralight.Command = UltralightCommand.Read16Bytes;
+        var res = ultralight.RunUltralightCommand();
+        if (res > 0)
+        {
+            Console.Write($"  Block: {ultralight.BlockNumber:X2} - ");
+            for (int i = 0; i < 4; i++)
+            {
+                Console.Write($"{ultralight.Data![i]:X2} ");
```
krwq, line 71: consider making `Data` non-nullable (i.e. return empty array) and fix sample so it doesn't have any nullable related `!`

src/devices/Card/Ultralight/README.md
```diff
@@ -0,0 +1,159 @@
...
+        {
+            res = ultralight.IsFormattedNdef();
+            if (res)
+            {
+                Console.WriteLine("Formating successful");
+            }
+            else
+            {
+                Console.WriteLine("Formated properly");
+            }
```
krwq, line 143: this piece of code is very confusing IMO. can you write better Console.WriteLine messages? When can this return non-formatted ndef? what does it mean it's formatted properly vs successful?

src/devices/Card/Ultralight/README.md
```diff
@@ -0,0 +1,159 @@
...
+    newMessage.Records.Add(new TextRecord("I ❤ .NET IoT", "en", Encoding.UTF8));
+    res = ultralight.WriteNdefMessage(newMessage);
+    if (res)
+    {
+        Console.WriteLine("NDEF data successfully written on the card.");
+    }
+    else
+    {
+        Console.WriteLine("Error writing NDEF data on card");
+    }```
```
krwq, line 157: triple backtick should be on new line

src/devices/Card/Ultralight/UltralightCard.cs
krwq: consider making all `byte[]` properties non-nullable and internally return/set empty buffer on null

src/devices/Card/Ultralight/UltralightCard.cs
```diff
@@ -0,0 +1,995 @@
...
+                        UltralightCardType = UltralightCardType.UltralightNtag203;
+                    }
+                    else
+                    {
+                        NdefCapacity = 144;
+                        UltralightCardType = UltralightCardType.MifareUltralight;
+                    }
+                }
+            }
+
```
krwq, line 218: extra enter

src/devices/Card/Ultralight/UltralightCard.cs
```diff
@@ -0,0 +1,995 @@
...
+                block += 4;
+            }
+
+            var ndef = NdefMessage.ExtractMessage(card);
+
+            try
+            {
+                message = new NdefMessage(ndef);
+            }
+            catch (Exception)
```
krwq, line 693: consider inheriting all expected exception from the common base class and catching only that type here - errors like NullRefernceException or ArgumentException should show here still because they mean library error

src/devices/Card/Ultralight/UltralightCard.cs
```diff
@@ -0,0 +1,995 @@
...
+            if (ret < 0)
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Set the password, the AuthenticationKey is used as the old password
```
krwq, line 887: would be useful to type what's the expected size of the buffer

src/devices/Pn5180/README.md
```diff
@@ -4,7 +4,7 @@ PN5180 is a RFID and NFC reader. It does supports various standards: ISO/IEC 144
 
 ## Documentation
 
-Official documentation can be fond here: https://www.nxp.com/docs/en/data-sheet/PN5180A0XX-C3.pdf
+Official documentation can be fond here: <https://www.nxp.com/docs/en/data-sheet/PN5180A0XX_C3_C4.pdf>
```
krwq, line 7:
```suggestion
Official documentation can be found [here](https://www.nxp.com/docs/en/data-sheet/PN5180A0XX_C3_C4.pdf)
```

krwq (approved):
Overall looks good, I've left couple of suggestions

src/devices/Card/Ultralight/MirrorConfiguration.cs
Ellerbach: No as you need to know what's the size. You can create an NDEF message like:
`My card number is 00000000000000 and it's cool` (I'm using 0 but it can be any character, not even the same one)
It will ben be transformed automatically into:
`My card number is 04ABCDEF123456 and it's cool`
So it's about exposing the size of the Mirror UID, same for the other one

src/devices/Card/Ultralight/README.md
Ellerbach: OK, will do and as well in the MifareCard

src/devices/Card/Ultralight/README.md
Ellerbach: good catch, it's not correct!

src/devices/Card/Ultralight/UltralightCard.cs
Ellerbach: There are really quite a lot which can happen including some null references as explained in the comment. So this is really for convenience and to make sure nothing is not catch. I know it's not fully aligned with the good pattern but I feel it's going to be more confusing if I have a list of 6 or even more possible exceptions.

src/devices/Card/Ultralight/UltralightCard.cs
Ellerbach: it's done in the ProcessAuthentication which is the function called from there. So the buffer size is checked.

PR #1587, opened 2021-07-13
Improve performance of VideoDevice continuous capture

### LINKED ISSUE #1585: VideoDevice.CaptureContinuous low FPS
**Describe the bug**

I was attempting to use the continuous capture on a raspberry pi 4b to stream from the attached camera. When running the app I was only getting 2 FPS from the camera, where as, if I use the inbuilt raspivid tool I get 30 FPS from the same camera. After confirming that I wasn't doing anything wrong, I went digging in the `VideoDevice` source code and found the cause of the slowness. Both the `Capture` and `CaptureContinuous` methods are calling the same `ProcessCaptureData` method, which opens and closes the data stream for every call. In my measurements, the opening and closing of the stream takes around 400ms, which explains why the `CaptureContinuous` is not able to reach a usable frame rate, as most of the time is spent repeatedly opening and closing the stream.

To confirm my suspicion, [I forked the repo](https://github.com/CodedBeard/iot/commit/cf0e4439fe8f9e686c139f76538c3395fc3ea308) and added a new method `CaptureContinuousBytes` that keeps the stream open for the length of the continuous capture. I also added another method `GetFrameDataPooled` for processing the frame data, which uses `ArrayPool<byte>` to reduce the allocations and thus GC pressure. I also avoided using `Stream` as in my use case it was causing uneccessery allocations converting the `byte[]` to `Stream` and then immediately calling ToArray on the stream.

With these changes, while its still not quite as fast as raspivid's framerate, it is at least hitting 20 FPS. I'd be happy to create a PR to have these changes incorrporated into the next version, but thought it best to open an issue for discussion first, as I suspect my current implementation isn't the best.

**Steps to reproduce**

This can be reproduced using the [MJPEG example project](https://github.com/dotnet/iot/tree/main/src/devices/Media/VideoDevice/samples/MjpegStream) in the repo. 

**Expected behavior**

The framerate matches that of the inbuilt raspivid tool

**Actual behavior**

Iot can only manage 2 FPS

**Versions used**

```
 Version:   6.0.100-preview.5.21302.13
 Commit:    d6380bcae7

Runtime Environment:
 OS Name:     debian
 OS Version:  10
 OS Platform: Linux
 RID:         debian.10-arm64
 Base Path:   /home/pi/dotnet/sdk/6.0.100-preview.5.21302.13/

Host (useful for support):
  Version: 6.0.0-preview.5.21301.5
  Commit:  ec3e0b276b

.NET SDKs installed:
  6.0.100-preview.2.21155.3 [/home/pi/dotnet/sdk]
  6.0.100-preview.5.21302.13 [/home/pi/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.0-preview.2.21154.6 [/home/pi/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.0-preview.5.21301.17 [/home/pi/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.0-preview.2.21154.6 [/home/pi/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.0-preview.5.21301.5 [/home/pi/dotnet/shared/Microsoft.NETCore.App]

```
System.Device.Gpio version 1.4
Iot.Device.Bindings version 1.4

CodedBeard:
Fixes #1585 

The intent of this PR is to provide a more performant implementation of the continuous capture methods. This has been tested on a raspberry pi 4b, and was able to provide a 25fps MJPEG stream to multiple clients.

I've not modified any of the existing methods, as I wasn't sure on the backward compatability policy.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1587)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1587) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1587) before we can accept your contribution.<br/><br/>:x: CodedBeard [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1587)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1587) it.</sub>

krwq:
@CodedBeard for Iot.Device.Bindings we're ok with justified breaking changes (for System.Device.Gpio we're more strict) - greatly improving performance definitely counts as that. cc: @Ellerbach

Ellerbach (approved):
Thanks! Looks good to me and thanks for creating the PR.

CodedBeard:
@Ellerbach are you happy for these to remain as additional methods, or would you prefer I update the PR to replace the original methods?

Ellerbach:
[Triage] @CodedBeard feel free to improve and adjust the methods from the original code. This code is still in progress, so really feel free to add and adjust.
So just let us know what you prefer: have it separated like this or adjusting the existing one. We're ver open on any of the option.

CodedBeard:
@Ellerbach my personal feeling, is that it would be better to replace the original implementation with the new one, as I don't feel the original offers much value to the consumer over calling `Capture` in a loop. The other thing which has been bugging me, is I'm not sure that `EnablePooling` is the correct name for the buffer pooling property. Maybe `ImageBufferPoolingEnabled` might be a more appropriate and discoverable name?

I've got more partial implementations in my fork now that I'm happy to PR back once they are in a more usable state. I won't pollute this PR with them as they are for completely different parts, but just want to check if you would like me to open seperate issues for each first? or are you happy for me to just put in a PR with the info I would have included in an issue?

Ellerbach:
@CodedBeard just go for what is best. Change the original version make it better, no problem. Again, this is just a start, not much used, so go for any change that make sense. Keep this PR, adjust anything in there that you want.
We'll review the new changes, make comments if appropriate.
Thanks again for taking the time to improve it!

CodedBeard:
@Ellerbach I think this update will do for now if you are happy with it. Initially I was going to allow the individual caputre to run whilst continuous was already running, but the increase in complexity seemed unwarneted, and so settled on throwing instead for the time being. Ultimately I do still want to allow capturing screenshots while another consumer is using the continuous capture, but I couldn't find an elegant solution so far, and didn't feel it was worth holding back the rest of the changes any longer.

CodedBeard:
Just realised from the failed builds I forgot to update the samples, will get that fixed over the weekend

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
```diff
@@ -70,14 +96,21 @@ public override void Capture(string path)
         /// Capture a picture from the video device.
         /// </summary>
         /// <returns>Picture stream.</returns>
-        public override Stream Capture()
+        public override byte[] Capture()
```
Ellerbach, line 99: Why changing the Stream for a byte[]?

Ellerbach (commented):
All up looks ok, see my comment. Please adjust the samples as well.

src/devices/Media/VideoDevice/Devices/UnixVideoDevice.cs
CodedBeard: If the consumer wants a `Stream`, they can easily make one themselves without incuring additional allocations. Where as, if the consumer wants a `byte[]` like I do, then you have to call ToArray() on the Stream incurring the additional allocation. While that is of course less important for the individual Capture call right now, I didn't want to prevent future changes where it may be more important, like allowing to Capture from multiple connections on a webserver.

Ellerbach (approved):
[Triage] Tank you for the adjustment and the comment on the byte[], we're ok with that.
Please fix the code for the build to pass and we should be good to merge.

CodedBeard:
@Ellerbach sorry for the delay, managed to catch covid :(

The samples are working now, and I added some additional comments.

pgrawehr:
@CodedBeard Is this ready for merge, or do you still have something open?

CodedBeard:
@pgrawehr yep its ready, been really busy at work so not had a chance to follow up since last comment.

PR #1589, opened 2021-07-15
Removed unnecessary init commands

Jimmys20:
Hello, in this PR I removed some unnecessary initialization commands from `LcdInterface.I2c4Bit` class (Display Function, Display Control, Display Mode and Clear Display). The exact same commands are already being sent inside the `Hd44780` class: https://github.com/dotnet/iot/blob/main/src/devices/CharacterLcd/Hd44780.cs#L109. So, right now these initialization commands are sent twice when using `LcdInterface.I2c4Bit`. Also in `LcdInterface.I2c4Bit` these commands are hardcoded, the `Hd44780` class considers other parameters and sends the appropriate commands.

With this change `LcdInterface.I2c4Bit` works more similar to `LcdInterface.Gpio` and only sends the necessary commands:
```
SendCommandAndWait(0x3);
SendCommandAndWait(0x3);
SendCommandAndWait(0x3);
SendCommandAndWait(0x2);
```
https://github.com/dotnet/iot/blob/main/src/devices/CharacterLcd/LcdInterface.Gpio.cs#L129

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1589)


Jimmys20:
Also following override could be removed from `LcdInterface.I2c4Bit` (line 77)
```
public override void SendCommandAndWait(byte command)
{
    // Must not run the init sequence to fast or undefined behavior may occur
    SendCommand(command);
    Thread.Sleep(1);
}
```
as base `LcdInterface` class already has the same implementation for this method (line 101).
```
public virtual void SendCommandAndWait(byte command)
{
    // Must not run the init sequence to fast or undefined behavior may occur
    SendCommand(command);
    Thread.Sleep(1);
}
```

pgrawehr:
Hmmm... This needs testing. I probably added those because I had problems initializing the display sometimes. And maybe doing it twice fixed it, I don't exactly remember. This could also have been caused by some hardware problems, though.

joperezr:
[Triage]: @pgrawehr can you please test to see if these commands where required or if after removing them the device still works?

pgrawehr (approved):
Appears to still work.

pgrawehr:
@joperezr This can be merged according to me.

joperezr (approved):

PR #1598, opened 2021-07-17
Added lcd interface to support SPI

Jimmys20:
Hi, in this pull request I added a new LcdInterface implementation that supports controlling a character lcd using a shift register. To try the sample demo you need to connect an hd44780 type lcd (16x02 used in the sample) to an sn74hc595 shift register the following way:

```
sn74hc595 -> hd44780
       Q0 -> not used
       Q1 -> RS pin
       Q2 -> Enable pin
       Q3 -> D7
       Q4 -> D6
       Q5 -> D5
       Q6 -> D4
       Q7 -> Backlight anode
```

This is the same pinout used on this lcd backpack [product](https://www.adafruit.com/product/292) based on the provided [schematic](https://learn.adafruit.com/i2c-spi-lcd-backpack/downloads). The pins are configurable using the arguments of the `LcdInterface.CreateSpi(int registerSelectPin, int enablePin, int[] dataPins, int backlightPin = -1)` method. I did my tests using a plain sn74hc595 chip a 1602 display on a breadboard.

You can connect the shift register to your microcontroller however you like. The sample uses the `Sn74hc595PinMapping.Minimal` which translates to:

```
Ser = 16    (SR 14   -- data)
SrClk = 20  (SR 11   -- storage register clock)
RClk = 21   (SR 12   -- latch enable to publish storage register)
```

This lcd interface only supports 4bit mode at the moment because we don't have enough output pins on the SN74HC595 shift register to support 8bit mode. For 8bit mode a 16bit shift register is needed or 2 x 8bit daisy chained. Support for that could be added in the future.
Any feedback would be highly appreciated.

### DIFF
--- a/src/devices/CharacterLcd/LcdInterface.Spi.cs
+++ b/src/devices/CharacterLcd/LcdInterface.Spi.cs
@@ -1,0 +1,175 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using Iot.Device.Multiplexing;
+
+namespace Iot.Device.CharacterLcd
+{
+    public abstract partial class LcdInterface : IDisposable
+    {
+        /// <summary>
+        /// This interface allows the control of a character display using SPI protocol.
+        /// The display is controlled using SN74HC595 shift register. This shift register can be controlled
+        /// using a minimum of three pins so this method saves pins on the microcontroller compared to direct GPIO
+        /// connection. This interface adds support for existing products that allow connect to an HD44780 display
+        /// through a shift register e.g. https://www.adafruit.com/product/292
+        /// </summary>
+        private class Spi : LcdInterface
+        {
+            private readonly byte _registerSelectPin;
+            private readonly byte _enablePin;
+            private readonly byte[] _dataPins;
+            private readonly byte _backlightPin;
+
+            private readonly bool _shouldDispose;
+            private Sn74hc595 _shiftRegister;
+            private bool _backlightOn;
+
+            public Spi(int registerSelectPin, int enablePin, int[] dataPins, int backlightPin = -1, Sn74hc595? shiftRegister = null, bool shouldDispose = true)
+            {
+                _registerSelectPin = (byte)(1 << registerSelectPin);
+                _enablePin = (byte)(1 << enablePin);
+
+                // Right now only 4bit mode is supported. 8bit mode would require 16bit shift register or 2x 8bit
+                // because we would need more than 8 output pins to support 8bit mode. This could be implemented in the future.
+                if (dataPins.Length != 4)
+                {
+                    throw new ArgumentException("The length of the array must be 4.", nameof(dataPins));
+                }
+
+                _dataPins = new byte[dataPins.Length];
+
+                for (var i = 0; i < dataPins.Length; i++)
+                {
+                    _dataPins[i] = (byte)(1 << dataPins[i]);
+                }
+
+                if (backlightPin != -1)
+                {
+                    _backlightPin = (byte)(1 << backlightPin);
+                }
+
+                _shouldDispose = shouldDispose || _shiftRegister is null;
+                _shiftRegister = shiftRegister ?? new Sn74hc595(Sn74hc595PinMapping.Minimal);
+
+                _backlightOn = true;
+                Initialize();
+            }
+
+            public override bool EightBitMode => false;
+
+            /// <summary>
+            /// This Display supports enabling/disabling the backlight.
+            /// The text on the display is not affected by disabling the backlight - it is just very hard to read.
+            /// </summary>
+            public override bool BacklightOn
+            {
+                get => _backlightOn;
+                set
+                {
+                    _backlightOn = value;
+                    // Need to send a command to make this happen immediately.
+                    SendCommandAndWait(0);
+                }
+            }
+
+            private byte BacklightFlag
+            {
+                get
+                {
+                    if (BacklightOn)
+                    {
+                        return _backlightPin;
+                    }
+                    else
+                    {
+                        return 0;
+                    }
+                }
+            }
+
+            private void Initialize()
+            {
+                // This sequence (copied from a python example) completely resets the display (if it was
+                // previously erroneously used with 8 bit access, it may not return to normal operation otherwise)
+                SendCommandAndWait(0x3);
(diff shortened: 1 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1598) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1598) before we can accept your contribution.<br/><br/>:x: Jimmys20 [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1598)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1598) it.</sub>

src/devices/CharacterLcd/LcdInterface.Spi.cs
```diff
@@ -0,0 +1,175 @@
...
+            private readonly byte _registerSelectPin;
+            private readonly byte _enablePin;
+            private readonly byte[] _dataPins;
+            private readonly byte _backlightPin;
+
+            private readonly bool _shouldDispose;
+            private Sn74hc595 _shiftRegister;
+            private bool _backlightOn;
+
+            public Spi(int registerSelectPin, int enablePin, int[] dataPins, int backlightPin = -1, Sn74hc595? shiftRegister = null, bool shouldDispose = true)
```
Ellerbach, line 29: please add the /// comments

src/devices/CharacterLcd/LcdInterface.Spi.cs
```diff
@@ -0,0 +1,175 @@
...
+
+                return bits;
+            }
+
+            protected override void Dispose(bool disposing)
+            {
+                if (_shouldDispose)
+                {
+                    _shiftRegister?.Dispose();
+                    _shiftRegister = null!;
```
Ellerbach, line 168: This should be outside of the _shoudDispose test. Setting the _shiftRegister to null even without Dispose, will all the GC to clean the variable even if the original class is not.
I know this is coming from the I2C4bits implementation. There are few like this which should be corrected. Btw, you can correct them as well if possible.

Ellerbach (commented):
Looks good so far! Please add as well the instructions in the README.md so it will be complete.

src/devices/CharacterLcd/LcdInterface.Spi.cs
```diff
@@ -0,0 +1,175 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using Iot.Device.Multiplexing;
+
+namespace Iot.Device.CharacterLcd
+{
+    public abstract partial class LcdInterface : IDisposable
```
krwq, line 9: I find this a bit confusing - if you're using SPI interface why is this not using SpiDevice?

krwq: would it be possible to implement SpiDevice using Sn74hc595? If not consider changing name to `LcdInterface.ShiftRegister` or even explicitly `Sn74hc595`

Jimmys20: Generic `ShiftRegister` binding supports control using `SpiDevice` and even has an example diagram that uses the Sn74hc595: https://github.com/dotnet/iot/tree/main/src/devices/ShiftRegister#using-spi
But for some reason the `Sn74hc595` binding is missing the `SpiDevice` constructor even though it states that is supports SPI control: https://github.com/dotnet/iot/blob/main/src/devices/Sn74hc595/Sn74hc595.cs#L12. I can try to add that, maybe in a separate PR.

Then the user will be able to 'indirectly' control the Lcd using SpiDevice. e.g.:
```
using SpiDevice spiDevice = SpiDevice.Create(new(0, 0));
using Sn74hc595 sr = new(spiDevice);
using LcdInterface lcdInterface = LcdInterface.CreateSpi(registerSelectPin, enablePin, dataPins, backlightPin, sr);
using Lcd1602 lcd = new(lcdInterface);
lcd.Clear();
lcd.Write("Hello World");
```
Still changing the name to LcdInterface.Sn74hc595 might be more appropriate. Do you think following names would be better?
File name -> `LcdInterface.Sn74hc595.cs`
Inner class name -> `private class Sn74hc595LcdInterface : LcdInterface { ... }`
create instance method -> `LcdInterface.CreateSn74hc595(...)`

src/devices/CharacterLcd/LcdInterface.Spi.cs
Jimmys20: done

src/devices/CharacterLcd/LcdInterface.Spi.cs
Jimmys20: Should be fixed now.

src/devices/CharacterLcd/LcdInterface.Spi.cs
krwq: Would this mechanism work with all shift registers? I assume yes, if so then I see two paths forward:
- change this to LcdInterface.ShiftRegister and use base class rather than specific ShiftRegister implementation
- create SpiDevice over ShiftRegister and add SPI interface here (if that makes sense - haven't done the proper review yet, just quick skim)

Jimmys20: I didn't quite understand the second bullet could you please explain?

Jimmys20:
Hi, I made some more changes. I switched to the base `ShiftRegister` class. Also added support for up-to 64bit shift registers (probably don't exist but they can be daisy chained). Renamed to `LcdInterface.ShiftRegister`.

Ellerbach:
[Triage] Thanks a lot for the progress on this. There is now maybe a rename you can do from `CreateShiftRegister` to `CreateFromShiftRegister`.
For the rest, seems good.

Jimmys20:
Thanks for the feedback, I will wait for PR https://github.com/dotnet/iot/pull/1610 to get merged and then I will also update the docs with instructions on how to use this interface.

joperezr:
@Ellerbach anything left to check on this one? seems to me it's ready to go in.

Ellerbach (approved):
Looks good to me

PR #1599, opened 2021-07-20
Sn74hc595 SpiDevice

Jimmys20:
Hello, I made a small change to the `Sn74hc595` binding regarding `SpiDevice` support. Fixed the samples accordingly. Also I noticed that the diagram which demonstrates the required wiring for using the SN74HC595 with SPI had a small mistake https://github.com/dotnet/iot/tree/main/src/devices/ShiftRegister#using-spi

RCLK is connected to SCLK and SRCLK to CE0, it should be the other way around like in the following image.

GPIO-diagram-Raspberry-Pi-4

So I updated the fritzing diagram and the exported png of the diagram

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1599)


src/devices/ShiftRegister/samples/Program.cs
```diff
@@ -12,7 +12,7 @@
 // Uncomment this code to use SPI (and comment the line above)
 // SpiConnectionSettings settings = new(0, 0);
 // using var spiDevice = SpiDevice.Create(settings);
-// var sr = new Sn74hc595(spiDevice, Sn74hc595.PinMapping.Standard);
+// using ShiftRegister sr = new(spiDevice, 8);
```
krwq, line 15: why this change? Isn't example with specific shift register better?

krwq (approved):
LGTM but haven't validated diagram

krwq:
@richlander since you wrote the original binding do you want to take a look?

src/devices/ShiftRegister/samples/Program.cs
Jimmys20: This line was giving error because `src/devices/ShiftRegister/samples/ShiftRegister-driver.csproj` project has no reference to the `src/devices/Sn74hc595/Sn74hc595.csproj` project and also `Sn74hc595.PinMapping.Standard` is no longer a valid way to pass pin mapping because it was refactored.

So I changed it to be similar to this line: https://github.com/dotnet/iot/blob/c8ea79c62cfbde7bad5c24ee598e9402ea3ff579/src/devices/ShiftRegister/samples/Program.cs#L10
I guess ShiftRegister samples should use the generic `ShiftRegister` class and Sn74hc595 samples should use the specific `Sn74hc595` class.

PR #1619, opened 2021-07-29
Add class that supports OS-independent setting of the system clock

pgrawehr:
Adds a small class that allows setting the system clock. Tested on Windows and Linux. Not sure about MacOS yet.

This is helpful if no internet connection is available, but some other real time clock is available (i.e. a DS1307 or an NMEA source).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1619)

### DIFF
--- a/src/devices/Interop/Windows/Kernel32.cs
+++ b/src/devices/Interop/Windows/Kernel32.cs
@@ -1,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal partial class Interop
+{
+    [DllImport("kernel32.dll")]
+    [return: MarshalAs(UnmanagedType.Bool)]
+    public static extern bool SetSystemTime([In] ref SystemTime lpSystemTime);
+
+    [DllImport("kernel32.dll")]
+    [return: MarshalAs(UnmanagedType.Bool)]
+    public static extern bool GetSystemTime([Out] out SystemTime lpSystemTime);
+
+    [Serializable]
+    [StructLayout(LayoutKind.Sequential)]
+    public struct SystemTime
+    {
+        public ushort Year;
+        public ushort Month;
+        public ushort DayOfWeek;
+        public ushort Day;
+        public ushort Hour;
+        public ushort Minute;
+        public ushort Second;
+        public ushort Milliseconds;
+    }
+}
--- a/src/devices/Rtc/RtcBase.cs
+++ b/src/devices/Rtc/RtcBase.cs
@@ -25,29 +25,29 @@ namespace Iot.Device.Rtc
 
         /// <summary>
         /// Set or retrieves the current date/time. This property returns a <see cref="DateTimeOffset"/> and
-        /// is therefore correct regardless of the current time zone (when <see cref="TimeZone"/> is set correctly).
+        /// is therefore correct regardless of the current time zone (when <see cref="LocalTimeZone"/> is set correctly).
         /// </summary>
         public DateTimeOffset DateTime
         {
             get
             {
                 var now = RtcDateTime;
-                return new DateTimeOffset(now.Ticks, TimeZone.GetUtcOffset(now));
+                return new DateTimeOffset(now.Ticks, LocalTimeZone.GetUtcOffset(now));
             }
             set
             {
-                var clockNow = new DateTime((value.UtcDateTime + TimeZone.GetUtcOffset(value)).Ticks, DateTimeKind.Local);
+                var clockNow = new DateTime((value.UtcDateTime + LocalTimeZone.GetUtcOffset(value)).Ticks, DateTimeKind.Local);
                 RtcDateTime = clockNow;
             }
         }
 
         /// <summary>
-        /// Gets or sets the time zone this clock operates in.
-        /// Defaults to the local time zone.
+        /// Gets or sets the time zone this instance will operate in.
+        /// Defaults to the local time zone from the system.
         /// Changing this property will not change the time on the real time clock,
         /// but instead affect the return value of <see cref="DateTime"/>
         /// </summary>
-        public virtual TimeZoneInfo TimeZone
+        public virtual TimeZoneInfo LocalTimeZone
         {
             get => _timeZone;
             set => _timeZone = value;
--- a/src/devices/Rtc/RtcBase.cs
+++ b/src/devices/Rtc/RtcBase.cs
@@ -10,13 +10,55 @@ namespace Iot.Device.Rtc
     /// </summary>
     public abstract class RtcBase : IDisposable
     {
+        private TimeZoneInfo _timeZone;
+
         /// <summary>
-        /// The Device's <see cref="System.DateTime"/>
+        /// The Device's raw <see cref="System.DateTime"/>.
+        /// The caller must be aware of the device's time zone.
+        /// The behavior of the <see cref="System.DateTime.Kind"/> property is implementation-dependent (typically it is ignored)
         /// </summary>
-        public virtual DateTime DateTime
+        public virtual DateTime RtcDateTime
         {
             get => ReadTime();
             set => SetTime(value);
+        }
+
+        /// <summary>
+        /// Set or retrieves the current date/time. This property returns a <see cref="DateTimeOffset"/> and
+        /// is therefore correct regardless of the current time zone (when <see cref="TimeZone"/> is set correctly).
+        /// </summary>
+        public DateTimeOffset DateTime
+        {
+            get
+            {
+                var now = RtcDateTime;
+                return new DateTimeOffset(now.Ticks, TimeZone.GetUtcOffset(now));
+            }
+            set
+            {
+                var clockNow = new DateTime((value.UtcDateTime + TimeZone.GetUtcOffset(value)).Ticks, DateTimeKind.Local);
+                RtcDateTime = clockNow;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the time zone this clock operates in.
+        /// Defaults to the local time zone.
+        /// Changing this property will not change the time on the real time clock,
+        /// but instead affect the return value of <see cref="DateTime"/>
+        /// </summary>
+        public virtual TimeZoneInfo TimeZone
+        {
+            get => _timeZone;
+            set => _timeZone = value;
+        }
+
+        /// <summary>
+        /// Creates an instance of this base class
+        /// </summary>
+        protected RtcBase()
+        {
+            _timeZone = TimeZoneInfo.Local;
         }
 
         /// <summary>
--- a/src/devices/Rtc/SystemClock.cs
+++ b/src/devices/Rtc/SystemClock.cs
@@ -47,6 +47,12 @@ namespace Iot.Device.Rtc
         /// "Change the system clock" must have been granted to the "Users" group (in Security policy management).
         /// On Unix and MacOs, the current user must be root, must be able to sudo without password or the "date" command must have the setUid bit set.
         /// </summary>
+        /// <remarks>
+        /// This method is primarily intended for setting the system clock from an external clock source, such as a DS1307 or a GNSS source when no
+        /// internet connection is available. If an internet connection is available, most operating systems will by default automatically sync the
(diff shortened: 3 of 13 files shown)


Ellerbach:
[Triage] As discussed separately, please use the existing Rtc binding and use the base class to create a SystemRtc device.
Please also add a SetTime(DateTimeOffset time) method to make it easier when setting the time. And add a method to be able to read in which timezone you are (like TimeZoneInfo).

pgrawehr:
@Ellerbach When you have some time, please check whether the unix way of getting/setting the time in SystemClock.cs works on the Mac as well (add case in `GetSystemTimeUtc`/`SetSystemTimeUtc`)

joperezr:
[Triage] Pending on @Ellerbach to test on OSX

Ellerbach:
So seems that it's different on Mac, the sample gives:

```
Unhandled exception. System.IO.IOException: The return value of the date command could not be parsed: 1630664392.N
 (seconds since 01/01/1970)
   at Iot.Device.Rtc.SystemClock.GetDateTimeUtcUnix() in /Users/laurent/Desktop/iot/src/devices/Rtc/SystemClock.cs:line 159
   at Iot.Device.Rtc.SystemClock.GetSystemTimeUtc() in /Users/laurent/Desktop/iot/src/devices/Rtc/SystemClock.cs:line 83
   at <Program>$.<<Main>$>g__TestSystemTime|0_0() in /Users/laurent/Desktop/iot/src/devices/Rtc/samples/Program.cs:line 8
   at <Program>$.<Main>$(String[] args) in /Users/laurent/Desktop/iot/src/devices/Rtc/samples/Program.cs:line 19
```

so seems that on Mac the command returns .N on top of the number. That should be easy to check and remove it.

Ellerbach:
The right command seems to be just this one for MacOs:

```
date -u +%s   
1630664785
```

Now setting up the date seems different as well. The righ forat is just:

```
date MMddhhmmyy
```

MM = month, 2 digits
dd = day, 2 digits
hh = hour, 2 digits
mm = minutes, 2 digits
yy = year, 2 digits

Tested and it works find

pgrawehr:
@Ellerbach Is there no more accurate format for setting the time? I need at least second precision, better yet to milliseconds.

Ellerbach:
> @Ellerbach Is there no more accurate format for setting the time? I need at least second precision, better yet to milliseconds.

I search and didn't find :-( Will give it another try and test if I find any other way.

Ellerbach:
And according to this:, there are commands that can be used to be a bit more precise: https://www.apple.com/server/docs/Command_Line.pdf

But is does require to make sure that the ntp server setup is off.
So, the right order is:

```
 sudo systemsetup -setusingnetworktime off
sudo systemsetup -setdate mm:dd:yy
sudo systemsetup -settime hh:mm:ss
``` 

It seems the best you can do, so seconds only. 
Note that if the first command is not run, you get an error (tried).
As soon as it's on again, of course the date/time changes again.
With the date command on, same behavior, as soon as the device gets back to network, if the automatic setup is on, it will get back to the ntp time.

pgrawehr:
> With the date command on, same behavior, as soon as the device gets back to network, if the automatic setup is on, it will get back to the ntp time.

That's the same on the Raspi, actually. Not sure we should care about that, though, since this function is used particularly when there's no NTP server available. I prefer to document that an active internet connection may interfere with this setting and provide documentation on how to disable that if one really wants to. We shouldn't do any behind-the-scene system setting changes.

pgrawehr:
@Ellerbach Added the MacOs functions, but they seem to fail the unit tests, unfortunately without usable output on the build servers. Can you have a look?

joperezr:
Sorry for the wait, just tried setting the clock with this code on my mac and it didn't work:

```
Unhandled exception. System.UnauthorizedAccessException: Error running date command as local user and as root. Error 1: date: illegal time format
usage: date [-jnRu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ... 
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]

   at Iot.Device.Rtc.SystemClock.SetDateTimeUtcMacOs(DateTime dt) in /Users/joperezr/git/iot/src/devices/Rtc/SystemClock.cs:line 217
   at Iot.Device.Rtc.SystemClock.SetSystemTimeUtc(DateTime dt) in /Users/joperezr/git/iot/src/devices/Rtc/SystemClock.cs:line 73
   at Program.<<Main>$>g__TestSystemTime|0_0() in /Users/joperezr/git/iot/src/devices/Rtc/samples/Program.cs:line 16
   at Program.<Main>$(String[] args) in /Users/joperezr/git/iot/src/devices/Rtc/samples/Program.cs:line 20
```

seems like we just need to format that date string correctly.

pgrawehr:
@joperezr Can you investigate on the correct string to use? I don't have a Mac, so I cannot test.

src/devices/Interop/Windows/Kernel32.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    [DllImport("kernel32.dll")]
+    [return: MarshalAs(UnmanagedType.Bool)]
+    public static extern bool SetSystemTime([In] ref SystemTime lpSystemTime);
+
+    [DllImport("kernel32.dll")]
+    [return: MarshalAs(UnmanagedType.Bool)]
+    public static extern bool GetSystemTime([Out] out SystemTime lpSystemTime);
+
+    [Serializable]
```
krwq, line 17: does this need to be serializable?

src/devices/Rtc/RtcBase.cs
```diff
@@ -10,15 +10,57 @@ namespace Iot.Device.Rtc
...
+        public DateTimeOffset DateTime
+        {
+            get
+            {
+                var now = RtcDateTime;
+                return new DateTimeOffset(now.Ticks, TimeZone.GetUtcOffset(now));
+            }
+            set
+            {
+                var clockNow = new DateTime((value.UtcDateTime + TimeZone.GetUtcOffset(value)).Ticks, DateTimeKind.Local);
```
krwq, line 39: I kinda feel like there should be two time zones:
- RTC time zone (hidden from user, perhaps accepted in the constructor, if not provided it should be the same as local time zone)
- local time zone

when setting this property RTC time zone and local time zone should become the same and therefore RTC time should be directly set to whatever the local time is. If we leave as is I'm sure there will be many edge cases

src/devices/Rtc/RtcBase.cs
```diff
@@ -10,15 +10,57 @@ namespace Iot.Device.Rtc
...
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the time zone this clock operates in.
+        /// Defaults to the local time zone.
+        /// Changing this property will not change the time on the real time clock,
+        /// but instead affect the return value of <see cref="DateTime"/>
+        /// </summary>
+        public virtual TimeZoneInfo TimeZone
```
krwq, line 50: This should be clear in the property name if this is RtcTimeZone or LocalTimeZone

src/devices/Rtc/RtcBase.cs
krwq: i.e. say I'm in UTC time zone and I move the clock to to UTC+1, as long as I provide correct time zone everything should work fine (RTC time zone is UTC, local is UTC+1), the moment I try to explicitly set time I'd expect RTC time zone to be now UTC+1 and internally work in the UTC+1 rather than UTC

src/devices/Rtc/tests/SystemClockTests.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
```
krwq, line 1: copyright header, please check all files

src/devices/Rtc/SystemClock.cs
```diff
@@ -0,0 +1,264 @@
...
+            {
+                var output = RunDateCommandUnix($"{formattedTime}", true, out exitCode);
+                if (exitCode != 0)
+                {
+                    throw new UnauthorizedAccessException($"Error running date command as local user and as root. Error {exitCode}: {output}");
+                }
+            }
+        }
+
+        private static string RunDateCommandUnix(string commandLine, bool asRoot, out int exitCode)
```
krwq, line 222: We should always run as current user and not try to elevate, if doesn't have permissions then it should fail, if possible use linux syscalls rather than starting new process.

krwq: This might have some reference how to do it using syscalls https://stackoverflow.com/questions/28314543/set-system-date-and-time-using-c-in-linux

src/devices/Interop/Windows/Kernel32.cs
pgrawehr: Well, it's a flat struct, these should be made serializable by default.

src/devices/Rtc/SystemClock.cs
pgrawehr: Syscalls would be more complex than that, and they would require root priviledges. 

The above works fine on the Raspi or if the system is set up accordingly. I don't want that I need to run the program with root priviledges just to be able to set the time somewhere within. I've seen other GPS receiver applications do it similarly.

src/devices/Rtc/RtcBase.cs
pgrawehr: I was trying to keep it simple. Having two time zones is complex enough, adding a third (TZ the class returns, TZ the clock is in and UTC) will be mind-blowing for some people. I also don't see a reason why you should set the RTC to something else than your local TZ or utc. 

Additionally, if we want to do that, it would need to be in subclasses, because we can't set the timezone of the system clock, as we expect that after setting it, `DateTime.Now` will return the correct time.

pgrawehr:
@krwq @joperezr 

I have now removed the attempt to use sudo for setting the clock. On the Raspberry Pi, executing `sudo chmod +s /bin/date` once enables everyone to set the clock. On most other linux versions, this should work similarly. 

This permission change needs to be done once, but is then persistent (very similar to the corresponding windows policy)

krwq:
> On the Raspberry Pi, executing `sudo chmod +s /bin/date` once enables everyone to set the clock

Consider adding that to the README :-)

joperezr:
I have now adjusted the format to match what Mac expects. I have validated on my machine that this sets the time correctly when running with sudo. @pgrawehr very sorry for taking so long to push this change.

joperezr (approved):

pgrawehr:
Do not merge yet - there's an unclear build error somewhere.

joperezr:
I think I know what is going on, there is a test project somewhere targeting a runtime that is not being restored, I'm taking a quick look  to the binlog now.

joperezr:
I was able to recover a binlog from the build failure, and what seems to be failing in Linux is that the Rtc unit tests are failing the build when running in Linux. I tried reproing in both Windows and OSX while using the same Linux docker image to run the tests, but I can't get a repro. This must be an issue only when you run Linux docker on Linux.

Ellerbach:
>  This must be an issue only when you run Linux docker on Linux.

So giving the fact that this test will fail while it should succeed in the real life, maybe a good idea not to run it?

pgrawehr:
@joperezr That was the right hint, thanks. The new test project that was added here was still targeting netcoreapp2.1, while all the other projects where updated to 3.1 meanwhile. That appears to have solved it.

pgrawehr:
@Ellerbach This is not a problem of the test itself, but of the test setup. Apparently, a missing runtime version does not always result in the same errors, but this is not a problem of this particular test.

pgrawehr:
@joperezr Regarding OSX: It appears that since we moved to the new build servers, we "lost" the OSX builders. Currently, the builds run only on Windows and Linux. Is there a way to get OSX back?

pgrawehr:
@joperezr Regarding OSX: It appears that since we moved to the new build servers, we "lost" the OSX builders. Currently, the builds run only on Windows and Linux. Is there a way to get OSX back?

joperezr:
> Regarding OSX: It appears that since we moved to the new build servers, we "lost" the OSX builders. Currently, the builds run only on Windows and Linux. Is there a way to get OSX back?

Sure, I'll take care of that in a separate PR.

PR #1620, opened 2021-07-31
Adding doc fx structure generation

Ellerbach:
This PR is about creating a structure with all the Markdown files:

- move all the Markdown files to an includes folder
- move all the media files to a media folder
- adjust all the links
- adjust all the relative link to the repository into absolute links

### DIFF
--- a/tools/GenerateDocFxStructure/CommandLineOptions.cs
+++ b/tools/GenerateDocFxStructure/CommandLineOptions.cs
@@ -1,0 +1,43 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using CommandLine;
+
+namespace GenerateDocFxStructure
+{
+    /// <summary>
+    /// Class for command line options.
+    /// </summary>
+    public class CommandlineOptions
+    {
+        /// <summary>
+        /// Gets or sets the destination folder.
+        /// </summary>
+        [Option('d', "destination", Required = true, HelpText = "Folder containing the destination folder..")]
+        public string? DestinationFolder { get; set; }
+        
+        /// <summary>
+        /// Gets or sets the source folder.
+        /// </summary>
+        [Option('s', "source", Required = true, HelpText = "Folder containing the source folder.")]
+        public string? SourceFolder { get; set; }
+                
+        /// <summary>
+        /// Gets or sets the destination media folder.
+        /// </summary>
+        [Option('m', "media", Required = true, HelpText = "Folder containing the destination media folder.")]
+        public string? MediaFolder { get; set; }
+
+        /// <summary>
+        /// Gets or sets the main repository.
+        /// </summary>
+        [Option('r', "repo", Required = false, HelpText = "The main repository with the branch to target poiting on the folder used as a source. Default value is 'https://github.com/dotnet/iot/tree/main/src/devices'.")]
+        public string Repo { get; set; } = "https://github.com/dotnet/iot/tree/main/src/devices";
+
+        /// <summary>
+        /// Gets or sets a value indicating whether verbose information is shown in the output.
+        /// </summary>
+        [Option('v', "verbose", Required = false, HelpText = "Show verbose messages.")]
+        public bool Verbose { get; set; }
+    }
+}
--- a/tools/GenerateDocFxStructure/Program.cs
+++ b/tools/GenerateDocFxStructure/Program.cs
@@ -1,0 +1,313 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Web;
+using CommandLine;
+using GenerateDocFxStructure.Helpers;
+
+namespace GenerateDocFxStructure
+{
+    class Program
+    {
+        private static readonly string[] _mediaExtentions = new string[] { ".png", ".jpg", ".jpeg", ".mpg", ".mpeg", ".mov", ".mp4", ".bmp", ".gif", ".fzz" };
+        private static CommandlineOptions? _options;
+        private static int _returnvalue;
+        private static MessageHelper? _message;
+
+        private static List<string> allFiles = new List<string>();
+
+        private static int Main(string[] args)
+        {
+            Console.WriteLine("Hello World!");
+            try
+            {
+                Parser.Default.ParseArguments<CommandlineOptions>(args)
+                                   .WithParsed<CommandlineOptions>(RunLogic)
+                                   .WithNotParsed(HandleErrors);
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"ERROR: Parsing arguments threw an exception with message `{ex.Message}`");
+                _returnvalue = 1;
+            }
+
+            Console.WriteLine($"Exit with return code {_returnvalue}");
+
+            return _returnvalue;
+        }
+
+        private static void RunLogic(CommandlineOptions o)
+        {
+            _message = new MessageHelper(o);
+            if (string.IsNullOrEmpty(o.DestinationFolder))
+            {
+                _message.Error($"Destination folder must be specified.");
+                _returnvalue = 1;
+                return;
+            }
+
+            if (string.IsNullOrEmpty(o.SourceFolder))
+            {
+                _message.Error($"Source folder must be specified.");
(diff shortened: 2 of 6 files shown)


tools/GenerateDocFxStructure/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Usage
+
+The tool can be used like this:
+
+`GenerateDocFxStructure -d includesDirectory -m mediaDirectory -s sourceDirectory -r repoUrl`
+
+- `includesDirectory`: the includes destination directory. ex: `c:\iot-page-docfx\includes`
+- `mediaDirectory`: the media destination directory. ex: `c:\iot-page-docfx\media`
+- `sourceDirectory`: the source directory. ex: `c:\iot\src\devices`. It is important to point on the devices directory where all the markdown files will be extracted and analyzed.
```
krwq, line 19: can you also update https://github.com/dotnet/iot/blob/main/Documentation/creating-new-release.md with an example invocation using some easy to adjust folder name?

tools/GenerateDocFxStructure/Program.cs
```diff
@@ -0,0 +1,313 @@
...
+using System.Text.RegularExpressions;
+using System.Web;
+using CommandLine;
+using GenerateDocFxStructure.Helpers;
+
+namespace GenerateDocFxStructure
+{
+    class Program
+    {
+        private static readonly string[] _mediaExtentions = new string[] { ".png", ".jpg", ".jpeg", ".mpg", ".mpeg", ".mov", ".mp4", ".bmp", ".gif", ".fzz" };
```
krwq, line 17: I don't think we should be copying .fzz. Honestly I'd instead show warnings if you see any of:
- .jpeg (you should use .jpg instead)
- .fzz (we should not point to fzz anywhere, it's a source for the diagram picture)

tools/GenerateDocFxStructure/Program.cs
```diff
@@ -0,0 +1,313 @@
...
+        private static CommandlineOptions? _options;
+        private static int _returnvalue;
+        private static MessageHelper? _message;
+
+        private static List<string> allFiles = new List<string>();
+
+        private static int Main(string[] args)
+        {
+            Console.WriteLine("Hello World!");
```
krwq, line 26: 😄

tools/GenerateDocFxStructure/Program.cs
```diff
@@ -0,0 +1,313 @@
...
+            {
+                _message.Error($"ERROR: Documentation folder '{_options.SourceFolder}' doesn't exist.");
+                _returnvalue = 1;
+                return;
+            }
+
+            // We will start to get all the files
+            foreach (string file in Directory.EnumerateFiles(_options.SourceFolder, "*.*", SearchOption.AllDirectories))
+            {
+                allFiles.Add(file.ToLowerInvariant());
```
krwq, line 88: I'd avoid normalizing file names - I'd rather show warning if the name isn't all lower case or uses unexpected characters

tools/GenerateDocFxStructure/Program.cs
```diff
@@ -0,0 +1,313 @@
...
+        /// <summary>
+        /// Get the list of the files in the current directory.
+        /// </summary>
+        /// <param name="folder">The folder to search.</param>
+        private static void ProcessFiles(DirectoryInfo folder)
+        {
+            _message!.Verbose($"Process {folder.FullName} for files.");
+
+            List<FileInfo> files = folder.GetFiles("*.md").OrderBy(f => f.Name).ToList();
+            if (files == null)
```
krwq, line 125: will this not return empty list?

tools/GenerateDocFxStructure/Program.cs
```diff
@@ -0,0 +1,313 @@
...
+                return;
+            }
+
+            foreach (FileInfo fi in files)
+            {
+                _message.Verbose($"Processing {fi.FullName}");
+                string content = File.ReadAllText(fi.FullName);
+
+                // first see if there are links in this file
+                Regex rxContent = new Regex(@"(\[{1}.*\]{1}\({1}.*\){1}?)");
```
krwq, line 137: 1. you can cache this and precompile for some speed (not sure how fast this currently is and if this will actually make any observable difference)
2. Might be worth to comment some example stuff it matches so you don't have to decode it when reading code 😄

krwq: what's the question mark for?

```
(\[{1}.*\]{1}\({1}.*\){1}?)
                         ^
```

krwq: should all `.*` be `.+`? or even `[^()\[\]]` or something along the lines?

tools/GenerateDocFxStructure/Program.cs
```diff
@@ -0,0 +1,313 @@
...
+                        if (relative.Contains("#"))
+                        {
+                            relative = relative.Substring(0, relative.IndexOf("#"));
+                        }
+
+                        // decode possible HTML encoding
+                        relative = HttpUtility.UrlDecode(relative);
+
+                        // check link if not to a URL, in-doc link or e-mail address
+                        if (!relative.StartsWith("http:") &&
```
krwq, line 225: should you warn on unrecognized link type?

tools/GenerateDocFxStructure/Program.cs
```diff
@@ -0,0 +1,313 @@
...
+            var relativeDestinationFilePath = filepath.Substring(_options.SourceFolder!.Length + 1);
+            var newFileDestination = Path.Combine(_options.DestinationFolder!, relativeDestinationFilePath);
+            // Ensure the path exists and is created
+            var dir = Path.GetDirectoryName(newFileDestination);
+            if (!Directory.Exists(dir))
+            {
+                Directory.CreateDirectory(dir!);
+            }
+
+            StreamWriter destinationFile = File.CreateText(newFileDestination);
```
krwq, line 187: StringBuilder + File.WriteAllText might be easier here

krwq: if you stick with StreamWriter use `using` statement and get rid of flush + close

tools/GenerateDocFxStructure/CommandLineOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <summary>
+    /// Class for command line options.
+    /// </summary>
+    public class CommandlineOptions
+    {
+        /// <summary>
+        /// Gets or sets the destination folder.
+        /// </summary>
+        [Option('d', "destination", Required = true, HelpText = "Folder containing the destination folder..")]
+        public string? DestinationFolder { get; set; }
```
krwq, line 17: if this is not expected to be null do not make it nullable, instead use `= null!;` to suppress the warning

krwq (approved):
LGTM, couple of nits, only thing is to update the release instructions with what should you do before release and add some example in there with some paths which are easy to adjust tailored for our repo

tools/GenerateDocFxStructure/Program.cs
Ellerbach: :-) :-) :-) I think I forgot to remove something :-D

tools/GenerateDocFxStructure/Program.cs
Ellerbach: We've been trying multiple combinations on another project See [this](https://github.com/Ellerbach/docfx-companion-tools/tree/main/src/DocLinkChecker) and there's been couple of attempts to find the right way of writing the regex. It seems that after some time, the one in this code has been proven to work in all situations (with absolute files, relative, / or \ or url). 
To answer you question, the last ? is to match the previous element only once.
And, yes the regex can be precompiled or even just as a private readonly variable, that would optimize a bit the time of execuation. On my machine, without optimization, in verbose, it's about 2 seconds for the all repo. I can anyway optimize this a bit.
Current limitation is on image link like: `[image](anotherlink)` which are a bit more complicated to trap. Now, we don't have, if we know we have, it should be then checked within the extracted data to see if any other link (image) is in the extracted material.

tools/GenerateDocFxStructure/Program.cs
Ellerbach: There is a warning later on. So if malformed here, you'll get an error later on

tools/GenerateDocFxStructure/Program.cs
Ellerbach: it does. Will correct.

tools/GenerateDocFxStructure/Program.cs
Ellerbach: example!
image

Issue #1626, opened 2021-08-07
VirtualOutputSegment Write offset


Jimmys20:
Hello, in `Iot.Device.Multiplexing` the class`VirtualOutputSegment` has a method `void Write(byte value)` which states in the xml doc that the **_Least significant bit will be written to index 0._**

For example:
```
VirtualOutputSegment segment = new(8);
segment.Write(0b_1001_0110);

// Result:
// segment index: | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7
//         value: | 0 | 1 | 1 | 0 | 1 | 0 | 0 | 1
```

What I find odd is that when using a segment with bigger length that the written value the result is this:
```
VirtualOutputSegment segment = new(12);
segment.Write(0b_1001_0110);

// Result (offset to the left side of the byte value)
// segment index: | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11
//         value: | 0 | 0 | 0 | 0 | 0 | 1 | 1 | 0 | 1 | 0 |  0 |  1

// Expected result (offset to the right side of the byte value)
// segment index: | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11
//         value: | 0 | 1 | 1 | 0 | 1 | 0 | 0 | 1 | 0 | 0 |  0 |  0
```

To my understanding the least significant bit should always end up to index 0 and any possible offset should end up to the right. To explain with unit tests, I believe that the following two tests should be both be passing.

```
[Fact]
public void SegmentValuesWriteByte()
{
    VirtualOutputSegment segment = new(8);
    segment.Write(0b_1001_0110);

    Assert.True(
        segment[0] == 0 &&
        segment[1] == 1 &&
        segment[2] == 1 &&
        segment[3] == 0 &&
        segment[4] == 1 &&
        segment[5] == 0 &&
        segment[6] == 0 &&
        segment[7] == 1);
}

[Fact]
public void SegmentValuesWriteByteOffset()
{
    // segment length bigger than the written value
    VirtualOutputSegment segment = new(12);
    segment.Write(0b_1001_0110);

    Assert.True(
        segment[0] == 0 &&
        segment[1] == 1 &&
        segment[2] == 1 &&
        segment[3] == 0 &&
        segment[4] == 1 &&
        segment[5] == 0 &&
        segment[6] == 0 &&
        segment[7] == 1);
}
```

[...]

Issue #1544, opened 2021-05-22
OutputSegmentTest fails


Ellerbach:
**Describe the bug**

In Device.IoT.Multiplexing, once of the test is failing:

![image](https://user-images.githubusercontent.com/8145578/119226365-f9340300-bb08-11eb-893f-1d96a0713175.png)

**Steps to reproduce**

Run the test

**Expected behavior**

Test to pass

**Actual behavior**

Test fails

**Versions used**

Latest version from the repo. Latest dotnet version.

PR #1627, opened 2021-08-07
VirtualOutputSegment write bytes in the correct order

Jimmys20:
Hello, in this PR I'm attempting to fix the following issue: https://github.com/dotnet/iot/issues/1544
I assumed that the unit test is correct and that `void Write(ReadOnlySpan<byte> value)` method needed to be fixed.

Also, while looking at this I found another possible bug which I explain here: https://github.com/dotnet/iot/issues/1626. If it's a bug indeed I can try to fix that in this PR as well.

### DIFF
--- a/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
+++ b/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
@@ -81,7 +81,8 @@ namespace Iot.Device.Multiplexing.Utility
 
             for (int i = 0; i < value.Length; i++)
             {
-                WriteByteAsValues(value[i], offset + (i * 8));
+                int index = value.Length - i - 1;
+                WriteByteAsValues(value[index], offset + (i * 8));
             }
         }
 


Jimmys20:
Also this check should be changed to (`if (offset > 0)`)
https://github.com/dotnet/iot/blob/fbc32f96b13c5085882f1f29923c371e20c3438e/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs#L77

when the bytes provided exceed the length of the IOutputSegment, the offset (`(value.Length * 8) - _length`) will be greater than 0

joperezr (approved):
Sorry it took so long for us to review. Thanks for fixing this! I checked and there was actually one unit test that was failing due to this issue and for some reason it wasn't being reported during the build, but your code is fixing that so thanks a lot!

Jimmys20:
Hello @joperezr, there where actually two different issues. I'll try to explain.

Initially I found this open issue https://github.com/dotnet/iot/issues/1544 which is about the failing unit test. This was caused because the bytes where written in reverse order. This is what I tried to fix in this PR. 

While working on this I noticed another possible bug which has to do with the offset which I reported here https://github.com/dotnet/iot/issues/1626 because I was not sure if it was bug or expected behaviour. This was **not** fixed in this PR.

There are no unit tests for the offset scenario which is why I opened that new issue to clarify if the offset is supposed to be at the start or the end of the bytes.

So the other PR that was closed was trying to fix the offset issue.

PR #1628, opened 2021-08-09
Adding FT232H

Ellerbach:
- Adding FT232H
- Adding documentation
- Adjusting FT4222 to reuse part of the existing code with minor braking changes
- Adjusting related samples
- Creating a shared common project for FTDI common elements getting ready to add more of those devices

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1628)


dotMorten:
SPI works great with my SSD1351 display and Adafruit FT232H adapter:

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+            : base(ftDevice.Flags, ftDevice.Type, ftDevice.Id, ftDevice.LocId, ftDevice.SerialNumber, ftDevice.Description)
+        {
+        }
+
+        /// <summary>
+        /// Creates I2C bus related to this device
+        /// </summary>
+        /// <returns>I2cBus instance</returns>
+        /// <remarks>You can create either an I2C, either an SPI device.</remarks>
+        public I2cBus CreateI2cBus() => new Ft232HI2cBus(this);
```
krwq, line 170: should this be abstract on the FtDevice? Same question for GPIO, SPI - we could return null or throw if given device doesn't support the protocol

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+        /// They all have to have different Chip Select. You can use any of the 3 to 15 pin for this function.</remarks>
+        public SpiDevice CreateSpiDevice(SpiConnectionSettings settings) => new Ft232HSpi(settings, this);
+
+        /// <summary>
+        /// Creates GPIO driver related to this device
+        /// </summary>
+        /// <returns>A GPIO Driver</returns>
+        public GpioDriver CreateGpioDriver() => new Ft232HGpio(this);
+
+        internal bool IsI2cMode { get; set; }
```
krwq, line 188: I'd suggest make this public and abstract, non-settable in FtDevice, perhaps rename to `IsI2cAvailable` then `GetI2cBus` could throw if it's not available. Similar for SPI/GPIO

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+            toSend[idx++] = (byte)FtOpcode.SetDataBitsLowByte;
+            // SDA and SCL both output high (open drain)
+            _gpioLowData = (byte)(I2cDataSDAhiSCLhi | (_gpioLowData & 0xF8));
+            _gpioLowDir = (byte)(I2cDirSDAoutSCLout | (_gpioLowDir & 0xF8));
+            toSend[idx++] = _gpioLowData;
+            toSend[idx++] = _gpioLowDir;
+            Write(toSend);
+        }
+
+        private void SetupMpsseMode()
```
krwq, line 271: optional: tiny comment what Mpsse means would be useful

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+        {
+            // Seems that we have to send a wrong command to get the MPSSE moed working
```
krwq, line 273:
```suggestion
            // Seems that we have to send a wrong command to get the MPSSE mode working
```

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+        {
+            if (_connectionSettings.Count == 0)
+            {
+                IsSpiMode = false;
+            }
+        }
+
+        internal void SpiWrite(SpiConnectionSettings settings, ReadOnlySpan<byte> buffer)
+        {
+            if (buffer.Length > 65535)
```
krwq, line 568: do all the pre-checks need to happen on all writes? should settings be part of the initialization instead?

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+            toSend[idx++] = (byte)((buffer.Length - 1) & 0xFF);
+            toSend[idx++] = (byte)((buffer.Length - 1) >> 8);
+            buffer.CopyTo(toSend.Slice(3));
+            Write(toSend);
+            SpiChipSelectEnable((byte)settings.ChipSelectLine, false);
+        }
+
+        internal void SpiRead(SpiConnectionSettings settings, Span<byte> buffer)
+        {
+            if (buffer.Length > 65535)
```
krwq, line 617: same question about checks

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+            toSend[idx++] = (byte)((buffer.Length - 1) & 0xFF);
+            toSend[idx++] = (byte)((buffer.Length - 1) >> 8);
+            Write(toSend);
+            Read(buffer);
+            SpiChipSelectEnable((byte)settings.ChipSelectLine, false);
+        }
+
+        internal void SpiWriteRead(SpiConnectionSettings settings, ReadOnlySpan<byte> bufferWrite, Span<byte> bufferRead)
+        {
+            if ((bufferRead.Length > 65535) || (bufferWrite.Length > 65535))
```
krwq, line 665: ditto

krwq: (disregard comment for buffers - that's correct, only meant settings)

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+            toSend[idx++] = (byte)((bufferWrite.Length - 1) >> 8);
+            bufferWrite.CopyTo(toSend.Slice(3));
+            Write(toSend);
+            Read(bufferRead);
+            SpiChipSelectEnable((byte)settings.ChipSelectLine, false);
+        }
+
+        internal void SpiChipSelectEnable(byte chipSelect, bool enable)
+        {
+            if (chipSelect < 0)
```
krwq, line 714: chipSelect is unsigned

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+            {
+                throw new ArgumentException($"Can't find specific chip select {chipSelect}");
+            }
+
+            var value = conn.ChipSelectLineActiveState;
+            // In case of deselect, we just invert what's needed
+            if (!enable)
+            {
+                value = value == PinValue.High ? PinValue.Low : PinValue.High;
+            }
```
krwq, line 730: I'm confused about this, I'd expect something like this here:

```suggestion
            var value = enabled ? conn.ChipSelectLineActiveState : !conn.ChipSelectLineActiveState;
```

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+                bytesToRead = GetAvailableBytes();
+                if (bytesToRead > 0)
+                {
+                    ftStatus = FtFunction.FT_Read(_ftHandle, in buffer[totalBytesRead], bytesToRead, ref numBytesRead);
+                    if ((ftStatus != FtStatus.Ok) && (bytesToRead != numBytesRead))
+                    {
+                        throw new IOException("Can't read device");
+                    }
+
+                    totalBytesRead += (int)bytesToRead;
```
krwq, line 812: I know they're equal but I'd suggest to use `numBytesRead` here as it's easier to see this is correct without reading all the ifs

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+            int totalBytesRead = 0;
+            uint bytesToRead = 0;
+            uint numBytesRead = 0;
+            FtStatus ftStatus;
+            while ((totalBytesRead < buffer.Length) && (!token.IsCancellationRequested))
+            {
+                bytesToRead = GetAvailableBytes();
+                if (bytesToRead > 0)
+                {
+                    ftStatus = FtFunction.FT_Read(_ftHandle, in buffer[totalBytesRead], bytesToRead, ref numBytesRead);
```
krwq, line 806: should `numBytesRead` be `out`?

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+                return ftStatus == FtStatus.Ok;
+            }
+
+            return true;
+        }
+
+        private uint GetAvailableBytes()
+        {
+            uint availableBytes = 0;
+            var ftStatus = FtFunction.FT_GetQueueStatus(_ftHandle, ref availableBytes);
```
krwq, line 837: the name suggest the queue could mean either read or write queue but you always use it in combination with reading, I'd suggest changing the name of this method and explain that i.e.: write queue will always be empty because we wait on write so this actually always means read queue

src/devices/Ft232H/Ft232HDevice.cs
Ellerbach: I was thinking of this. We can definitely do it for I2CBuc, SpiDevice and GpioDriver

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -0,0 +1,857 @@
...
+                    }
+
+                    totalBytesRead += (int)bytesToRead;
+                }
+            }
+
+            return totalBytesRead;
+        }
+
+        private bool FlushBuffer()
```
krwq, line 819: since this means read all available bytes I'd suggest to call it: `DiscardInput()`

src/devices/Ft232H/Ft232HGpio.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// <inheritdoc/>
+        protected override void OpenPin(int pinNumber)
+        {
+            if ((pinNumber < 0) || (pinNumber >= PinCount))
+            {
+                throw new ArgumentException($"Pin number can only be between 0 and {PinCount - 1}");
+            }
+
+            if (DeviceInformation.IsI2cMode && (
+                (pinNumber >= 0) && (pinNumber <= 2)))
```
krwq, line 47: tiny comment above why 0,1,2 would be enough here :-)

src/devices/Ft232H/Ft232HDevice.cs
Ellerbach: The IsI2cAvailable, IsSpiAvailable, IsGpioAvailable is a good idea. I can add. The IsI2cMode is there to say that I2C is activated. So I can rename this one as IsI2cModeEnabled, can do the same for the 2 others.
As for this board, you can't have both I2c and Spi activated, I had to have this way to check.
What do you say about the properties to expose what's available and not?

src/devices/Ft232H/Ft232HGpio.cs
```diff
@@ -0,0 +1,197 @@
...
+                throw new ArgumentException($"Can't open pin 0, 1 or 2 while I2C mode is on");
+            }
+
+            if (DeviceInformation.IsSpiMode && (
+                (pinNumber >= 0) && (pinNumber <= 2)))
+            {
+                throw new ArgumentException($"Can't open pin 0, 1 or 2 while SPI mode is on");
+            }
+
+            if (DeviceInformation._connectionSettings.Where(m => m.ChipSelectLine == pinNumber).Any())
```
krwq, line 58: `_connectionSettings` - this variable should be called as `ConnectionSettings` because it's either internal or public

src/devices/Ft232H/Ft232HGpio.cs
```diff
@@ -0,0 +1,197 @@
...
+            {
+                throw new ArgumentException($"Pin already open or used as Chip Select");
+            }
+
+            DeviceInformation._PinOpen[pinNumber] = true;
```
krwq, line 63: similar comment about naming of `_PinOpen`

src/devices/Ft232H/Ft232HDevice.cs
Ellerbach: settings is the easy way to get the chipselect and the default state of the chip select pin. You can have as many SpiDevice as you want, just limited by the number of GPIO available. So a maximum of 13. 
Now for the size of the buffer, I have to say I was mazy. I can do a loop and send the sliced buffer as many times as needed. Will consider to add this. This will make less checks

src/devices/Ft232H/Ft232HDevice.cs
Ellerbach: right. Will adjust. Forgot to do it there. I do support SPI Device without chipselect. Great catch!

src/devices/Ft232H/Ft232HDevice.cs
Ellerbach: on the other API it's ref which is used. So I prefer to keep ref so it's homogeneous.

src/devices/Ft232H/Ft232HDevice.cs
Ellerbach: Those are the name of the FTDI API. I can adjust them but for reference to the docs, I would prefer to keep them like they are to make it easier later to debug or implement something new.
And you are correct about what it does, it only checks how many bytes are available to read. I'll adjust the docs of the function.

src/devices/Ft232H/Ft232HGpio.cs
Ellerbach: Good catch. Will definitely add there! And I've explained that in the main readme :-) So I can just copy/paste :-)

src/devices/Ft232H/Ft232HGpio.cs
```diff
@@ -0,0 +1,197 @@
...
+                return ((DeviceInformation._gpioLowDir >> pinNumber) & 0x01) == 0x01 ? PinMode.Output : PinMode.Input;
+            }
+
+            return ((DeviceInformation._gpioHighDir >> (pinNumber - 8)) & 0x01) == 0x01 ? PinMode.Output : PinMode.Input;
+        }
+
+        /// <inheritdoc/>
+        protected override bool IsPinModeSupported(int pinNumber, PinMode mode)
+        {
+            if ((mode == PinMode.InputPullDown) || (mode == PinMode.InputPullUp))
```
krwq, line 121: this is equivalent to `return !((mode == PinMode.InputPullDown) || (mode == PinMode.InputPullUp));`

I'd suggest to do a positive condition though, i.e.: `return mode == PinMode.Input || mode == PinMode.Output;`

src/devices/Card/Ndef/samples/Program.cs
```diff
@@ -411,7 +412,7 @@ Pn5180 Ft4222Pn5180()
         Console.WriteLine();
     }
 
-    var (chip, dll) = FtCommon.GetVersions();
+    var (chip, dll) = Ft4222Common.GetVersions();
```
krwq, line 415: could FtCommon enumerate both Ft4222s and Ft232?

src/devices/Ccs811/samples/Program.cs
```diff
@@ -50,7 +51,7 @@
 else if (platformChoice.KeyChar == '2')
 {
     WriteLine("Creating an instance of a CCS811 using FT4222 drivers.");
-    using (var i2cBus = FtCommon.GetDevices()[0].CreateI2cBus())
+    using (var i2cBus = new Ft4222Device(FtCommon.GetDevices()[0]).CreateI2cBus())
```
krwq, line 54: could FtCommon enumerate both Ft4222 and Ft232? This way we could create a device regardless which device someone is using

src/devices/Ft232H/Ft232HGpio.cs
```diff
@@ -0,0 +1,197 @@
...
+
+            return true;
+        }
+
+        /// <inheritdoc/>
+        protected override PinValue Read(int pinNumber)
+        {
+            if (pinNumber < 8)
+            {
+                var val = DeviceInformation.GetGpioValuesLow();
```
krwq, line 134: could GetGpioValues() combine both Low and High into a single uint16?

src/devices/Ft232H/Ft232HGpio.cs
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        /// <inheritdoc/>
+        protected override void Write(int pinNumber, PinValue value)
+        {
+            if (pinNumber < 8)
+            {
+                if (value == PinValue.High)
+                {
+                    DeviceInformation._gpioLowData |= (byte)(1 << pinNumber);
```
krwq, line 151: would it make sense to have a single uint16 for both low and high and make SetGpioValues() write both? I think that should be fairly cheap

krwq: treat as optional since this is an implementation detail although I'd prefer we didn't touch fields starting with underscore unless in the same class

src/devices/Ft232H/Ft232HI2c.cs
```diff
@@ -0,0 +1,75 @@
...
+        public override I2cConnectionSettings ConnectionSettings => _settings;
+
+        /// <inheritdoc/>
+        public override void Read(Span<byte> buffer)
+        {
+            _i2cBus.Read(_deviceAddress, buffer);
+        }
+
+        /// <inheritdoc/>
+        public override byte ReadByte()
```
krwq, line 35: isn't the default implementation gonna do the same? if so, I'd suggest to remove this. The only time you should override this if you can provide faster implementation than default (i.e. specific interop for reading a single byte)

krwq: same comment for WriteByte

src/devices/Ft232H/Ft232HI2cBus.cs
```diff
@@ -0,0 +1,99 @@
...
+            var ack = DeviceInformation.I2cSendDeviceAddrAndCheckACK((byte)deviceAddress, false);
+            if (!ack)
+            {
+                DeviceInformation.I2cStop();
+                throw new IOException($"Error writing device while setting up address");
+            }
+
+            for (int i = 0; i < buffer.Length; i++)
+            {
+                ack = DeviceInformation.I2cSendByteAndCheckACK(buffer[i]);
```
krwq, line 88: isn't driver providing any way to Write and Read at the same time?

src/devices/Ft232H/Ft232HI2cBus.cs
```diff
@@ -0,0 +1,99 @@
...
+            DeviceInformation.I2cStop();
+        }
+
+        internal void Write(int deviceAddress, ReadOnlySpan<byte> buffer)
+        {
+            DeviceInformation.I2cStart();
+            var ack = DeviceInformation.I2cSendDeviceAddrAndCheckACK((byte)deviceAddress, false);
+            if (!ack)
+            {
+                DeviceInformation.I2cStop();
```
krwq, line 82: consider wrapping the whole code in try..finally and putting I2cStop in the finally block and then get rid of it in all fail paths

src/devices/Ft232H/Ft232HSpi.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        public Ft232HDevice DeviceInformation { get; internal set; }
+
+        internal Ft232HSpi(SpiConnectionSettings settings, Ft232HDevice deviceInformation)
+        {
+            DeviceInformation = deviceInformation;
+            _settings = settings;
+            if ((_settings.ChipSelectLine < 3) || (_settings.ChipSelectLine > Ft232HDevice.PinCountConst))
+            {
+                throw new ArgumentException($"Chip Select line has to be between 3 and {Ft232HDevice.PinCountConst - 1}");
```
krwq, line 31: the error message doesn't match the check above (off by 1 on the upper bound)

src/devices/Ft232H/Ft232HSpi.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <inheritdoc/>
+        public override void Write(ReadOnlySpan<byte> buffer)
+        {
+            DeviceInformation.SpiWrite(_settings, buffer);
+        }
+
+        /// <inheritdoc/>
+        public override void WriteByte(byte value)
+        {
+            DeviceInformation.SpiWrite(_settings, stackalloc byte[1] { value });
```
krwq, line 73: same comment as for I2C for reading/writing single byte

krwq: uhh I just noticed that's actually abstract and we don't provide default implementation which I think we should... Not sure if that would be considered breaking change....

krwq: I just started a conversation if it would be safe to make it virtual - I think many of our internal implementations do it exactly this way

src/devices/Ft232H/README.md
```diff
@@ -0,0 +1,99 @@
+# SPI, GPIO and I2C drivers for FT232H
```
krwq, line 1: consider joining this README with Ft4222 since they're almost the same

src/devices/Ft232H/samples/Program.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Device.Gpio;
+using Iot.Device.Bmxx80;
+using Iot.Device.Bmxx80.FilteringMode;
+using Iot.Device.Common;
+using Iot.Device.Ft232H;
+using Iot.Device.FtCommon;
+using UnitsNet;
+
+Console.WriteLine("Hello FT232H");
+var devices = FtCommon.GetDevices();
```
krwq, line 17: What confuses me is that old (FT4222) FtCommon became Ft4222Common and Ft232 is now FtCommon - should they be just a single FtCommon?

src/devices/Ft4222/Ft4222Common.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Runtime.InteropServices;
+using System.Text;
+using Iot.Device.FtCommon;
+
+namespace Iot.Device.Ft4222
+{
+    /// <summary>
+    /// Common static functions for the FT4222
+    /// </summary>
+    public class Ft4222Common
```
krwq, line 17: honestly I'd just recommend people to always use FtCommon and either remove this or make it internal (depending if it's used anywhere internally)

src/devices/Card/Ndef/samples/Program.cs
Ellerbach: because FT4222 have a specific set of API, the GetVersions is only available with FT4222.

src/devices/FtCommon/FtDevice.cs
```diff
@@ -2,9 +2,11 @@
...
 using System;
+using System.Device.Gpio;
 using System.Device.I2c;
+using System.Device.Spi;
 
-namespace Iot.Device.Ft4222
+namespace Iot.Device.FtCommon
 {
     /// <summary>
     /// FT4222 device information
```
pgrawehr, line 12: This class should now derive from `Board`. Many of the methods already match, I think.

src/devices/FtCommon/FtDevice.cs
```diff
@@ -64,9 +66,23 @@ public FtDevice(FtFlag flags, FtDeviceType type, uint id, uint locId, string ser
         /// Creates I2C bus related to this device
         /// </summary>
         /// <returns>I2cBus instance</returns>
-        public I2cBus CreateI2cBus()
-        {
-            return new Ft4222I2cBus(this);
-        }
+        /// <remarks>You can create either an I2C, either an SPI device.</remarks>
+        public virtual I2cBus CreateI2cBus() => throw new NotImplementedException();
```
pgrawehr, line 70: Make this abstract. Or, if it is reasonable that a derived class won't support it, throw `NotSupportedException` instead. `NotImplementedException` leaves the feeling that something was forgotten.

src/devices/Ft232H/Ft232HDevice.cs
dotMorten: >  we could return null or throw if given device doesn't support the protocol

If it doesn't support it, I'd vote throw

src/devices/FtCommon/FtDevice.cs
Ellerbach: I have checked and was not fully sure but will definitely try.

src/devices/FtCommon/FtDevice.cs
Ellerbach: I won't use abstract as this class can be used in real and is needed, so will use virtual. For the exception, yes, will change. Good point.

Ellerbach:
@mhutch correct and the FT232H code can be adjusted fairly easily to support the 2232 and 4232 versions. I just don't have them at all, so I cannot test. But all up, the hard and core part is done!

Ellerbach (commented):
> One additional finding: You might want to test whether GPIO and I2C/SPI work in parallel, since the Ft4222GpioDriver and the Ft4222I2cDriver both open the device individually.

Thanks. I will check and find some time early this week to go thru all this. I see what I missed when trying to do it. Thanks again.

joperezr:
[Triage] Pending one last look by @pgrawehr and testing the changes by @Ellerbach

Ellerbach:
Thanks @pgrawehr all seems ok. During the series of tests, I found an issue on my side when the GpioController is used alone. So was great to catch it and fix it as well.

Now all is ready to merge!

pgrawehr (approved):

pgrawehr:
@joperezr Could #1654 have broken this build? This one was building before the latest merge.

joperezr:
No, it wasn’t that one. It was the one that introduced a new SDK, since that has stricter rules and new warnings. I’ll take a peek at this today and fix it.

dotMorten:
Nice! Thank you all! This'll make my blogpost a lot shorter now https://xaml.dev/post/Building-a-side-kick-display-for-your-PC-with-NET

joperezr:
Thanks so much for sharing @dotMorten! It would be nice if we add some good blogs talking about real-world end-to-end examples like yours somewhere in our docs. I believe @microhobby also had some cool projects he worked on and used dotnet/iot that are worth including in our docs as well.

dotMorten:
@joperezr Thanks! You're more than welcomed to reference it. The project evolved a bit because I wanted a larger display, hence why I also submitted PR #1642

Ellerbach:
> Nice! Thank you all! This'll make my blogpost a lot shorter now https://xaml.dev/post/Building-a-side-kick-display-for-your-PC-with-NET

Very cool and happy being a part of this :-) I should just make the same one :-) I have few of those little screens :-) That gives me ideas.

PR #1635, opened 2021-08-16
Add possibility to externally extend Arduino library

pgrawehr:
Rewrote the internal interface a bit, so that it's now possible to register external command handlers.

Two of which are provided:
- DHT
- Frequency

### DIFF
--- a/src/devices/Arduino/ArduinoAnalogController.cs
+++ b/src/devices/Arduino/ArduinoAnalogController.cs
@@ -81,7 +81,7 @@ namespace Iot.Device.Arduino
             // This method is called with the logical pin numbering (input pin A0 is 0, A1 is 1, etc)
             // but the SetPinMode method operates on the global numbers
             int fullNumber = ConvertLogicalNumberingSchemeToPinNumber(pinNumber);
-            _board.Firmata.SetPinMode(fullNumber, SupportedMode.AnalogInput);
+            _board.Firmata.SetPinMode(fullNumber, SupportedMode.AnalogInput.Value);
             _board.Firmata.EnableAnalogReporting(pinNumber);
             return new ArduinoAnalogInputPin(_board, this, _supportedPinConfigurations[fullNumber], pinNumber, VoltageReference);
         }
--- a/src/devices/Arduino/ArduinoBoard.cs
+++ b/src/devices/Arduino/ArduinoBoard.cs
@@ -49,6 +49,14 @@ namespace Iot.Device.Arduino
 
         private ILogger _logger;
 
+        private int _lastFrequencyUpdateClock = 0;
+        private int _lastFrequencyUpdateTicks = 0;
+        private int _frequencyReportingPin = -1;
+
+        private Frequency _currentFrequency = Frequency.Zero;
+
+        private object _frequencyLock = new object();
+
         /// <summary>
         /// Creates an instance of an Ardino board connection using the given stream (typically from a serial port)
         /// </summary>
@@ -520,10 +528,92 @@ namespace Iot.Device.Arduino
         }
 
         /// <summary>
+        /// Enable frequency reporting on the given pin.
+        /// </summary>
+        /// <param name="pinNumber">The pin number</param>
+        /// <param name="mode">The mode on which to increase the tick count. Typically rising or falling edge.</param>
+        /// <param name="reportDelay">How often the value should update, in milliseconds. Must be positive and smaller than 2^14</param>
+        public void EnableFrequencyReporting(int pinNumber, FrequencyMode mode, int reportDelay)
+        {
+            if (_frequencyReportingPin != -1)
+            {
+                throw new InvalidOperationException($"Frequency reporting is already enabled for pin {_frequencyReportingPin}");
+            }
+
+            if (reportDelay <= 0 || reportDelay >= Math.Pow(2, 14))
+            {
+                throw new ArgumentOutOfRangeException(nameof(reportDelay), "Value must be between 1 and 2^14");
+            }
+
+            DisableFrequencyReporting(-1);
+
+            var firstQuery = Firmata.EnableFrequencyReporting(pinNumber, mode, reportDelay);
+            _lastFrequencyUpdateClock = firstQuery.TimeStamp;
+            _lastFrequencyUpdateTicks = firstQuery.NewTicks;
+            _frequencyReportingPin = pinNumber;
+
+            // By registering the event last we ensure we have a valid reading already
+            Firmata.OnSysexReply += OnFrequencyReport;
+        }
+
+        /// <summary>
+        /// Returns the last measured frequency. Returns 0 if no frequency measurement is active.
+        /// </summary>
+        /// <returns>The frequency measured during the last interval</returns>
+        public Frequency GetMeasuredFrequency()
+        {
+            if (_frequencyReportingPin == -1)
+            {
+                return Frequency.Zero;
+            }
+
+            lock (_frequencyLock)
+            {
+                return _currentFrequency;
+            }
+        }
+
+        private void OnFrequencyReport(byte[] bytes)
+        {
+            var result = Firmata.DecodeFrequencyReport(bytes);
+            if (!result.Success)
+            {
+                // Wrong message type, this is not typically an error
+                return;
+            }
+
+            lock (_frequencyLock)
+            {
+                double deltaTime = result.TimeStamp - _lastFrequencyUpdateClock;
+                double deltaTicks = result.NewTicks - _lastFrequencyUpdateTicks;
+                if (deltaTime > 0 && deltaTicks > 0) // Otherwise, this just wraps around or no time has passed
+                {
+                    _currentFrequency = Frequency.FromHertz(deltaTicks / (deltaTime / 1000));
+                }
+
+                _lastFrequencyUpdateClock = result.TimeStamp;
+                _lastFrequencyUpdateTicks = result.NewTicks;
+            }
+        }
+
+        /// <summary>
+        /// Disables automatic updating of the frequency counter for the given pin
+        /// </summary>
+        /// <param name="pinNumber">The pin to use</param>
+        public void DisableFrequencyReporting(int pinNumber)
+        {
+            Firmata.DisableFrequencyReporting(pinNumber);
+            Firmata.OnSysexReply -= OnFrequencyReport;
+            _frequencyReportingPin = -1;
+        }
+
+        /// <summary>
         /// Standard dispose pattern
         /// </summary>
         protected override void Dispose(bool disposing)
         {
+            DisableFrequencyReporting(-1);
+            _frequencyReportingPin = -1;
             _isDisposed = true;
             // Do this first, to force any blocking read operations to end
             if (_dataStream != null)
--- a/src/devices/Arduino/ArduinoBoard.cs
+++ b/src/devices/Arduino/ArduinoBoard.cs
@@ -466,6 +466,42 @@ namespace Iot.Device.Arduino
         }
 
         /// <summary>
+        /// Sets the internal pin mode to the given value, if supported.
+        /// </summary>
+        /// <param name="pin">The pin to configure</param>
+        /// <param name="arduinoMode">The mode to set</param>
+        /// <exception cref="TimeoutException">The mode was not updated, either because the command was not understood or
+        /// the mode is unknown by the firmware</exception>
+        /// <remarks>This method is intended for use by <see cref="ExtendedCommandHandler"/> instances. Users should not
+        /// call this method directly. It is the responsibility of the command handler to use the capabilities table to check
+        /// that the mode is actually supported</remarks>
+        public void SetPinMode(int pin, SupportedMode arduinoMode)
+        {
+            Firmata.SetPinMode(pin, arduinoMode.Value);
+        }
+
+        /// <summary>
+        /// Returns the current assignment of the given pin
+        /// </summary>
+        /// <param name="pinNumber">Pin number to query</param>
+        /// <returns>An instance of <see cref="SupportedMode"/> from the list of known modes (or a new instance for an unknown mode)</returns>
+        /// <remarks>Thi is the opposite of <see cref="SetPinMode"/>. See there for usage limitations.</remarks>
+        public SupportedMode GetPinMode(int pinNumber)
+        {
+            byte mode = Firmata.GetPinMode(pinNumber);
+            lock (_knownSupportedModes)
+            {
+                var m = _knownSupportedModes.FirstOrDefault(x => x.Value == mode);
+                if (m == null)
+                {
+                    return new SupportedMode(mode, $"Unknown mode {mode}");
+                }
+
+                return m;
+            }
+        }
+
+        /// <summary>
         /// Creates a GPIO Controller instance for the board. This allows working with digital input/output pins.
         /// </summary>
         /// <returns>An instance of GpioController, using an Arduino-Enabled driver</returns>
--- a/src/devices/Arduino/ArduinoBoard.cs
+++ b/src/devices/Arduino/ArduinoBoard.cs
@@ -112,6 +112,8 @@ namespace Iot.Device.Arduino
         private ArduinoBoard()
         {
             _logger = this.GetCurrentClassLogger();
+            // Add the extended command handlers that are defined in this library
+            _extendedCommandHandlers.Add(new DhtSensor());
         }
 
         /// <summary>
@@ -218,7 +220,6 @@ namespace Iot.Device.Arduino
         public void AddCommandHandler<T>(T newCommandHandler)
             where T : ExtendedCommandHandler
         {
-            Initialize();
             if (newCommandHandler == null)
             {
                 throw new ArgumentNullException(nameof(newCommandHandler));
@@ -240,7 +241,11 @@ namespace Iot.Device.Arduino
             }
 
             _extendedCommandHandlers.Add(newCommandHandler);
-            newCommandHandler.Registered(Firmata);
+            if (_firmata != null)
+            {
+                // Only if already initialized
+                newCommandHandler.Registered(_firmata, this);
+            }
         }
 
         /// <summary>
@@ -259,7 +264,33 @@ namespace Iot.Device.Arduino
                 }
             }
 
-            return null;
+            throw new NotSupportedException($"No handler of type {typeof(T)} found");
+        }
+
+        /// <summary>
+        /// Gets the command handler with the provided type.
+        /// </summary>
+        /// <typeparam name="T">The type to query</typeparam>
+        /// <param name="handler">The out variable</param>
+        /// <returns>True on success, false otherwise</returns>
+        public bool TryGetCommandHandler<T>(
+#if NET5_0_OR_GREATER
+            [NotNullWhen(true)]
+#endif
+            out T handler)
+            where T : ExtendedCommandHandler
+        {
+            foreach (var cmd in _extendedCommandHandlers)
+            {
+                if (cmd.GetType() == typeof(T))
+                {
+                    handler = (T)cmd;
+                    return true;
+                }
+            }
+
+            handler = null!;
+            return false;
         }
 
         /// <summary>
@@ -352,6 +383,12 @@ namespace Iot.Device.Arduino
                 }
 
                 _firmata.EnableDigitalReporting();
+
+                foreach (var e in _extendedCommandHandlers)
+                {
+                    e.Registered(_firmata, this);
+                    e.OnConnected();
+                }
 
                 _initialized = true;
             }
@@ -571,28 +608,6 @@ namespace Iot.Device.Arduino
             Initialize();
 
             Firmata.SetAnalogInputSamplingInterval(timeSpan);
-        }
-
-        /// <summary>
-        /// Special function to read DHT sensor, if supported
-        /// </summary>
-        /// <param name="pinNumber">Pin Number</param>
-        /// <param name="dhtType">Type of DHT Sensor: 11 = DHT11, 22 = DHT22, etc.</param>
-        /// <param name="temperature">Temperature</param>
-        /// <param name="humidity">Relative humidity</param>
-        /// <returns>True on success, false otherwise</returns>
-        public bool TryReadDht(int pinNumber, int dhtType, out Temperature temperature, out RelativeHumidity humidity)
-        {
-            Initialize();
-
-            if (!_supportedPinConfigurations[pinNumber].PinModes.Contains(SupportedMode.Dht))
-            {
-                temperature = default;
-                humidity = default;
-                return false;
-            }
-
-            return Firmata.TryReadDht(pinNumber, dhtType, out temperature, out humidity);
         }
 
         /// <summary>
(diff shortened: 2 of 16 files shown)


src/devices/Arduino/ArduinoAnalogController.cs
```diff
@@ -81,7 +81,7 @@ protected override AnalogInputPin OpenPinCore(int pinNumber)
             // This method is called with the logical pin numbering (input pin A0 is 0, A1 is 1, etc)
             // but the SetPinMode method operates on the global numbers
             int fullNumber = ConvertLogicalNumberingSchemeToPinNumber(pinNumber);
-            _board.Firmata.SetPinMode(fullNumber, SupportedMode.AnalogInput);
+            _board.Firmata.SetPinMode(fullNumber, SupportedMode.AnalogInput.Value);
```
krwq, line 84: I'm confused about this - why is SetPinMode taking a byte rather than PinMode/SupportedMode?

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -32,6 +32,9 @@ namespace Iot.Device.Arduino
     /// </summary>
     public class ArduinoBoard : Board.Board, IDisposable
     {
+        private static readonly List<SupportedMode> _knownSupportedModes = new List<SupportedMode>();
```
krwq, line 35: static variables should start with `s_`

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -49,6 +52,28 @@ public class ArduinoBoard : Board.Board, IDisposable
 
         private ILogger _logger;
 
+        static ArduinoBoard()
+        {
+            // We add all known modes to the list, even though we don't really support them all in the core
+            _knownSupportedModes.Add(SupportedMode.DigitalInput);
```
krwq, line 58: I'd recommend to do the initialization a bit differently:
start with `s_knownSupportedModes` being null, then create a `List<SupportedMode>` local, add all values then assign it to `s_knownSupportedModes`, perhaps even create a private helper method like `GetKnownSupportedValues` and just call it from field initializer. It's very easy to accidentally end up in semi-initialized state when doing this, with local/method you are either null or fully initialized and it's much easier to debug, reason or refactor this code

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -70,17 +95,37 @@ public ArduinoBoard(Stream serialPortStream)
         /// On Linux, possible values include "/dev/ttyAMA0", "/dev/serial0", "/dev/ttyUSB1", etc.</param>
         /// <param name="baudRate">Baudrate to use. It is recommended to use at least 115200 Baud.</param>
         public ArduinoBoard(string portName, int baudRate)
+        : this()
```
krwq, line 98: would it make sense to not have a parameterless constructor at all and only have this constructor and one which takes Stream, then have an `Initialize(Stream)` ctor which is called from the Stream constructor and also in here passing the BaseStream from seriall port? I think this would clean this up a bit

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -49,6 +52,28 @@ public class ArduinoBoard : Board.Board, IDisposable
...
         private ILogger _logger;
 
+        static ArduinoBoard()
+        {
+            // We add all known modes to the list, even though we don't really support them all in the core
+            _knownSupportedModes.Add(SupportedMode.DigitalInput);
+            _knownSupportedModes.Add(SupportedMode.DigitalOutput);
+            _knownSupportedModes.Add(SupportedMode.AnalogInput);
+            _knownSupportedModes.Add(SupportedMode.Pwm);
+            _knownSupportedModes.Add(SupportedMode.Servo);
```
krwq, line 62: `_knownSupportedModes` could be a ReadOnlyList if you used a `List<SupportedMode>` in a local here and then constructed `ReadOnlyList` from that local

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -70,17 +95,37 @@ public ArduinoBoard(Stream serialPortStream)
...
+        /// The list of supported pin modes.
+        /// This list can be extended by adding special modes using <see cref="AddCommandHandler{T}"/>.
+        /// </summary>
+        public static IReadOnlyList<SupportedMode> KnownModes
+        {
+            get
+            {
+                return _knownSupportedModes;
+            }
+        }
```
krwq, line 127:
```suggestion
        public static IReadOnlyList<SupportedMode> KnownModes => _knownSupportedModes;
```

src/devices/Arduino/ArduinoBoard.cs
krwq: ah.. just noticed this is extensible

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -159,51 +204,106 @@ public static List<int> CommonBaudRates()
...
+        public T? GetCommandHandler<T>()
+            where T : ExtendedCommandHandler
+        {
+            foreach (var cmd in _extendedCommandHandlers)
+            {
+                if (cmd.GetType() == typeof(T))
+                {
+                    return (T)cmd;
+                }
+            }
```
krwq, line 258: this won't work with the inheritance, should it? Should you use something like?

```suggestion
                if (cmd is T ret)
                {
                    return ret;
                }
            }
```

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -159,51 +204,106 @@ public static List<int> CommonBaudRates()
...
+
+            throw new NotSupportedException($"No handler of type {typeof(T)} found");
```
krwq, line 260: the error is confusing considering you return `T?`, I'd expect null

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -159,51 +204,106 @@ public static List<int> CommonBaudRates()
...
+        public bool TryGetCommandHandler<T>(
+#if NET5_0_OR_GREATER
+            [NotNullWhen(true)]
+#endif
+            out T handler)
+            where T : ExtendedCommandHandler
+        {
+            foreach (var cmd in _extendedCommandHandlers)
+            {
+                if (cmd.GetType() == typeof(T))
```
krwq, line 278: I think if you returned null from the previous method this method wouldn't be necessary. I'd stick with one way or the other and no error path, look which one is producing cleaner code and leave just that one

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -159,51 +204,106 @@ public static List<int> CommonBaudRates()
...
+        }
+
+        /// <summary>
+        /// Gets the command handler with the provided type.
+        /// </summary>
+        /// <typeparam name="T">The type to query</typeparam>
+        /// <param name="handler">The out variable</param>
+        /// <returns>True on success, false otherwise</returns>
+        public bool TryGetCommandHandler<T>(
+#if NET5_0_OR_GREATER
```
krwq, line 270: we should consider doing something similar the dotnet/runtime does instead (not in this PR) - they add dummy types related to nullability during the build so that it compiles when we're not on .NET 5 as well. This makes it much easier as you don't need to `#if` everywhere

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -159,51 +204,106 @@ public static List<int> CommonBaudRates()
...
+            foreach (var cmd in _extendedCommandHandlers)
+            {
+                if (cmd.GetType() == typeof(T))
+                {
+                    handler = (T)cmd;
+                    return true;
+                }
+            }
+
+            handler = null!;
```
krwq, line 285: why does this need `!`? for `<5.0`?

src/devices/Arduino/ArduinoGpioControllerDriver.cs
```diff
@@ -81,7 +81,7 @@ protected override void SetPinMode(int pinNumber, PinMode mode)
                 throw new NotSupportedException($"Mode {mode} is not supported on Pin {pinNumber}.");
             }
 
-            _arduinoBoard.Firmata.SetPinMode(pinNumber, firmataMode);
+            _arduinoBoard.Firmata.SetPinMode(pinNumber, firmataMode.Value);
```
krwq, line 84: why `.Value`? isn't that a byte?

src/devices/Arduino/DhtSensor.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
```
krwq, line 1: copyright header

src/devices/Arduino/ExtendedCommandHandler.cs
```diff
@@ -0,0 +1,193 @@
+﻿using System;
```
krwq, line 1: also, please check all files

src/devices/Arduino/FirmataCommandSequence.cs
```diff
@@ -11,26 +11,71 @@ namespace Iot.Device.Arduino
     /// A firmata command sequence
     /// Intended to be changed to public visibility later
     /// </summary>
-    internal class FirmataCommandSequence
+    public class FirmataCommandSequence
     {
+        /// <summary>
+        /// Start of sysex command byte. Used as start byte for almost all extended commands.
+        /// </summary>
+        public const byte StartSysex = (byte)FirmataCommand.START_SYSEX;
```
krwq, line 19: I'm missing something here, why is this public, why not just use  `FirmataCommand.START_SYSEX` wherever it's needed?

src/devices/Arduino/FirmataCommandSequence.cs
```diff
@@ -11,26 +11,71 @@ namespace Iot.Device.Arduino
...
         public int Length => _sequence.Count;
 
+        /// <summary>
+        /// Decode an uint from packed 7-bit data.
+        /// This way of encoding uints is only used in extension modules.
+        /// </summary>
+        /// <param name="data">Data. 5 bytes expected</param>
+        /// <param name="fromOffset">Start offset in data</param>
+        /// <returns></returns>
+        public static UInt32 DecodeUInt32(ReadOnlySpan<byte> data, int fromOffset)
```
krwq, line 65: I don't think this should be public, also this doesn't decode UIn32, it reads it, unless this can be casted into something

krwq: is `fromOffset` needed? why not just pass sliced `data`?

src/devices/Arduino/FirmataCommandSequence.cs
```diff
@@ -11,26 +11,71 @@ namespace Iot.Device.Arduino
...
+        {
+            Int32 value = data[fromOffset];
```
krwq, line 67: you should mask that one unused bit even if it should be zero, perhaps Debug.Assert it is actually zero

src/devices/Arduino/FirmataCommandSequence.cs
```diff
@@ -51,6 +96,10 @@ internal bool Validate()
             return true;
         }
 
+        /// <summary>
+        /// Encodes a set of bytes with 7 bits and adds them to the sequence. Each input byte is encoded in 2 bytes.
+        /// </summary>
+        /// <param name="values">Binary data to add</param>
```
krwq, line 102: This should be called WriteBytes, similarly all below - sorry I only commented now and not in the original PR

src/devices/Arduino/FirmataDevice.cs
```diff
@@ -775,7 +788,7 @@ private T PerformRetries<T>(int numberOfRetries, Func<T> operation)
             throw new TimeoutException("Timeout waiting for answer. Aborting. ", lastException);
         }
 
-        internal void SetPinMode(int pin, SupportedMode firmataMode)
+        internal void SetPinMode(int pin, byte firmataMode)
```
krwq, line 791: I think this should take SupportedMode/Firmata mode instead, it's hard to reason about `byte`

src/devices/Arduino/SupportedMode.cs
```diff
@@ -1,95 +1,158 @@
...
+        /// </summary>
+        public string Name
+        {
+            get;
+        }
+
+        /// <summary>
+        /// The <see cref="PinUsage"/> for this internal mode, if applicable
+        /// </summary>
+        public PinUsage PinMode
```
krwq, line 147: I guess the naming here is a bit confusing, why isn't this called `PinUsage` or `Usage`?

src/devices/Arduino/samples/Arduino.Monitor.cs
```diff
@@ -174,7 +181,7 @@ void ChangeMode(object sender, PinValueChangedEventArgs pinValueChangedEventArgs
                         break;
                     case 4:
                         modeName = "Temperature / Humidity";
-                        if (board.TryReadDht(3, 11, out temp, out var humidity))
+                        if (dht.TryReadDht(3, 11, out temp, out var humidity))
```
krwq, line 184: since this is a sample can you write down a comment what 3 and 11 are?

src/devices/Arduino/ArduinoAnalogController.cs
pgrawehr: This is the internal (low-level) value. But maybe I should move that conversion inside the function.

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: I don't think querying base classes is desirable here. You could be querying `ExtendedCommandHandler` itself, which will just get the first element of the list, which is completely useless.

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: Cool, the description says the same 😃

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: It doesn't, but since it also doesn't hurt in <5.0, no #if is required.

src/devices/Arduino/FirmataCommandSequence.cs
pgrawehr: Because FirmataCommand is not public (all the other members of that enum are not used by external components)

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: The serial port is only opened later, so the stream is not available yet. However, I can move the common stuff mostly to the Initialize function.

src/devices/Arduino/FirmataCommandSequence.cs
pgrawehr: This does a decode, it restores an int32 from 5(!) bytes. The firmata protocol requires that the payload of all commands consist of 7 bit data only (the top bit is reserved). Therefore an int32 needs to be sent as 5 bytes (or, as older commands do, even as 8 bytes). I should have added the inverse function, too, though. 

The `fromOffset` was just more convenient for implementation, since the protocol specifies the offsets for each field.

pgrawehr:
@krwq Findings addressed (where I silently resolved the comment, I followed your proposal). Just a few objections noted above.

pgrawehr:
@joperezr Changed the static member to an instance member now and using an ReaderWriterLockSlim for locking. Looks a bit cleaner this way.

pgrawehr:
Added "Do not merge". When this is ready to merge, I will need to sync my other branches against it and make sure the infrastructure introduced here really works for a bunch of use cases.

Ellerbach:
[Triage] @joperezr can you please have a look at the latest changes?

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -32,6 +32,11 @@ namespace Iot.Device.Arduino
     /// </summary>
     public class ArduinoBoard : Board.Board, IDisposable
     {
+        private readonly List<SupportedMode> _knownSupportedModes = new List<SupportedMode>();
+        private readonly List<ExtendedCommandHandler> _extendedCommandHandlers = new List<ExtendedCommandHandler>();
+        private readonly ReaderWriterLockSlim _commandHandlersLock =
+            new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
```
joperezr, line 38: Why do we need to support recursion here? When using this type you want to be very careful if you allow recursion as that can create deadlocks. In general it is better to use a NoRecursion lock unless you absolutely need it.

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -159,51 +184,94 @@ public static List<int> CommonBaudRates()
...
+        /// <param name="newCommandHandler">The new handler</param>
+        public void AddCommandHandler<T>(T newCommandHandler)
+            where T : ExtendedCommandHandler
+        {
+            if (newCommandHandler == null)
+            {
+                throw new ArgumentNullException(nameof(newCommandHandler));
+            }
+
+            _commandHandlersLock.EnterWriteLock();
```
joperezr, line 201: Nit: shouldn't the lock be entered after the if statement? I don't see where you are using _knownSupportedModes after the ifcheck so might as well move the entering of the lock one more line.

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -159,51 +184,94 @@ public static List<int> CommonBaudRates()
...
+                newCommandHandler.Registered(_firmata, this);
+            }
+        }
+
+        /// <summary>
+        /// Gets the command handler with the provided type. An exact type match is performed.
+        /// </summary>
+        /// <typeparam name="T">The type to query</typeparam>
+        /// <returns>The command handler, or null if none was found</returns>
+        public T? GetCommandHandler<T>()
```
joperezr, line 238: NIT: not sure what the use case for this is, but in case you want to support type hirerarchy as well, you can use Type.IsAssignableFrom

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -277,10 +349,47 @@ protected override void Initialize()
...
+            _extendedCommandHandlers.Add(new FrequencySensor());
+        }
+
+        /// <summary>
+        /// Registers the known supported modes. Should only be called once from Initialize.
+        /// </summary>
+        private void RegisterKnownSupportedModes()
+        {
+            // We add all known modes to the list, even though we don't really support them all in the core
+            _knownSupportedModes.Add(SupportedMode.DigitalInput);
```
joperezr, line 374: Shouldn't all of these happen inside a try after enetring a write lock? I know this happens at initialization, but technically I don't think there is anything preventing folks to try to read supported modes in a different thread while doing initialization in another one.

src/devices/Arduino/FirmataCommandSequence.cs
```diff
@@ -11,26 +12,117 @@ namespace Iot.Device.Arduino
...
+            {
+                // Bit 7 of the data must always be 0, or there's either a communication problem or a protocol mismatch
+                if ((data[fromOffset + i] & 0x80) != 0)
+                {
+                    throw new InvalidDataException("An invalid byte was received. The message was probably corrupted");
+                }
+            }
+
+            Int32 value = data[fromOffset];
+            value |= data[fromOffset + 1] << 7;
```
joperezr, line 79: Instead of making the conversion yourself, why not use BitConverter.ToUInt32()?

src/devices/Arduino/FirmataCommandSequence.cs
```diff
@@ -11,26 +12,117 @@ namespace Iot.Device.Arduino
...
+        /// Send an Uint32 as 5 x 7 bits. This form of transmitting integers is only supported by extension modules
+        /// </summary>
+        /// <param name="value">The 32-Bit value to transmit</param>
+        public void SendUInt32(UInt32 value)
+        {
+            byte[] data = new byte[5];
+            data[0] = (byte)(value & 0x7F);
+            data[1] = (byte)((value >> 7) & 0x7F);
+            data[2] = (byte)((value >> 14) & 0x7F);
+            data[3] = (byte)((value >> 21) & 0x7F);
```
joperezr, line 108: Same here, just use BitConverter.TryWriteBytes(UInt32)

src/devices/Arduino/FirmataDevice.cs
```diff
@@ -701,8 +719,9 @@ internal Version QueryFirmwareVersion(out string firmwareName)
                     _firmataStream.WriteByte((byte)FirmataSysexCommand.REPORT_FIRMWARE);
                     _firmataStream.WriteByte((byte)FirmataCommand.END_SYSEX);
                     bool result = _dataReceived.WaitOne(TimeSpan.FromSeconds(FIRMATA_INIT_TIMEOUT_SECONDS));
-                    if (result == false)
+                    if (result == false || _firmwareVersionMajor == 0)
                     {
+                        Thread.Sleep(100);
```
joperezr, line 724: Can we comment on to why we need to sleep here?

joperezr:
Left some small comments but other than those this looks good

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: I generally see it the other way round: When recursion is allowed, calling other private functions withhin the same class, which in turn take the lock as well, causes no problems. Since there are no callbacks or other outgoing calls within a lock, deadlocks cannot happen here.

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: `_extendedCommandHandlers` is also protected by the same lock. The two are updated synchronously.

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: No, see krwq's comment below. This method is designed as a "service query". You ask it for a specific interface (i.e. whether DHT support is available). Asking for base classes doesn't make sense (and wouldn't work, because this only returns one instance at a time, not several).

src/devices/Arduino/FirmataCommandSequence.cs
pgrawehr: Because it's not the same. This converts an uint32 to 5x7 bits. The firmata protocol requires that all message payloads have the topmost bit set to 0, because bytes with the top bit set have special meanings (such as an end-of-message marker).

src/devices/Arduino/FirmataCommandSequence.cs
pgrawehr: See above.

src/devices/Arduino/ArduinoBoard.cs
joperezr: If you check the documentation of the type, it tells you to try to avoid allowin recursion unless necessary. Anyway, my comment is: if you need it, then its fine, if not, better to remove it.

src/devices/Arduino/ArduinoBoard.cs
joperezr: Shouldeach of them have their own lock?

joperezr (approved):
This LGTM, fee free t mergeafter you do your testing.

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: No, they're related and updated together. A new command handler may or may not bring additional modes, but since this is all done within the same method, there's no point in having two locks.

PR #1639, opened 2021-08-19
Width is supposed to be height for second parameter

dotMorten:
Stumbled upon this when porting the code to a different display type. I guess it was never caught because this screen is usually 128x128.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1639)


krwq (approved):
LGTM assuming you have tried that on physical device

joperezr (approved):

joperezr:
Thanks for the fix @dotMorten

Issue #1646, opened 2021-08-25
Add to PN532 abbility to ignore firmware version ack by i2c interface


mcdis:
Some cheap copies, which are based on PN 521, do not respond to the request for the firmware version, but they ask to send it before working.
In arduino, in the source, they just ignore the answer.

Googling for 5 minutes, and you will find a decent amount of queries that people can't read the get firmware version answer.

It worked for me when, instead of exception, I set to ignore the answer.
This only applies to the i2c interface.

joperezr:
Closed by #1651

PR #1651, opened 2021-09-03
PN532: Adjusting constructor to able version check ignore

Ellerbach:
This relate to #1646

- Adding a version check ignore to the constructors
- As it seems devices still need to be asked for the version, kept the request
- Ignore the result if it has been asked in the constructor
- Even if this seems to concern only I2C, implemented for all 3 interfaces

@mcdis please check that this is what you want and that's working for you.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1651)

### DIFF
--- a/src/devices/Pn532/Pn532.cs
+++ b/src/devices/Pn532/Pn532.cs
@@ -123,9 +123,8 @@ namespace Iot.Device.Pn532
             // Set the SAM
             bool ret = SetSecurityAccessModule();
             _logger.LogInformation($"Setting SAM changed: {ret}");
-            // Check the version
-            var ver = IsPn532();
-            if (checkVersion && !ver)
+            // Check the version first in order to initialize the device
+            if (!IsPn532() && checkVersion)
             {
                 throw new Exception("Can't find a PN532");
             }
@@ -158,9 +157,8 @@ namespace Iot.Device.Pn532
             // returns false, some timeout appear. So we will need to apply a second time
             bool ret = SetSecurityAccessModule();
             _logger.LogInformation($"Setting SAM changed: {ret}");
-            // Check the version
-            var ver = IsPn532();
-            if (checkVersion && !ver)
+            // Check the version first in order to initialize the device
+            if (!IsPn532() && checkVersion)
             {
                 throw new Exception("Can't find a PN532");
             }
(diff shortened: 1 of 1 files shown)


src/devices/Pn532/Pn532.cs
```diff
@@ -123,7 +124,8 @@ public Pn532(string portName)
             bool ret = SetSecurityAccessModule();
             _logger.LogInformation($"Setting SAM changed: {ret}");
             // Check the version
-            if (!IsPn532())
+            var ver = IsPn532();
+            if (checkVersion && !ver)
```
joperezr, line 128:
```suggestion
            if (checkVersion && !IsPn532())
```
No need to have the local variable and no need to check the value if checkVersion is false. If you simplify the ifcheck you fix both issues.

src/devices/Pn532/Pn532.cs
```diff
@@ -156,7 +159,8 @@ public Pn532(SpiDevice spiDevice, int pinChipSelect, GpioController? controller
...
```
joperezr, line 163: Same suggestion here

src/devices/Pn532/Pn532.cs
```diff
@@ -172,15 +176,17 @@ public Pn532(SpiDevice spiDevice, int pinChipSelect, GpioController? controller
...
         {
             _logger = this.GetCurrentClassLogger();
             _i2cDevice = i2cDevice ?? throw new ArgumentNullException(nameof(i2cDevice));
             WakeUp();
             bool ret = SetSecurityAccessModule();
             _logger.LogInformation($"Setting SAM changed: {ret}");
             // Check the version
-            if (!IsPn532())
+            var ver = IsPn532();
+            if (checkVersion && !ver)
```
joperezr, line 189: And same here

joperezr (approved):
Left three small comments to fix the if check conditions and save you a potential unnecessary call but other than that this looks good. This can be merged once those three are addressed.

src/devices/Pn532/Pn532.cs
pgrawehr: According to the description in the PR (see above), the request for the version is still required, only the result shall be ignored.

Ellerbach: I confirm from the description and from what I found, you have to call the the check version even if the result should be ignored. Otherwise, seems like the chips doesn't then work properly. It's like initialization, it seems to be required.

joperezr: Oh sorry I missed that part of the description. In that case, my suggestion would be to use my suggestion except reverse the order of the operands so that the check always happens but you won't need the extra variable. So like:

```c#
// Checking version first in order to initialize the device
If (!IsPN532() && checkVersion)
```

Ellerbach: Yes, fair! Thanks for the suggestion. Adjusted.

PR #1653, opened 2021-09-03
Fix for SysFsDriver when callbacks registered events are not fired, after unregister all registered callbacks.

### LINKED ISSUE #1637: After unregister all callbacks, the new registered callbacks events are not fired in Raspberry Pi 4
**Describe the bug**

In GpioController, when you register some callbacks with RegisterCallbackForPinValueChangedEvent() function, the events are fired as expected.
Then, If you unregister all the callbacks registered before with UnregisterCallbackForPinValueChangedEvent() function, your events are not fired now. (This is the expected behaviour).
Now, if you registered new callbacks in GpioController with RegisterCallbackForPinValueChangedEvent() function, the new events are not fired. No matter how many times you unregister and register callbacks, the events will never be fired again.

**Steps to reproduce**

- Create new GpioController
- Open one pin as input / InputPullUp / InputPullDown
- Register some callbacks with RegisterCallbackForPinValueChangedEvent() function.
- Unregister all of them with UnregisterCallbackForPinValueChangedEvent() function.
- Register again some callbacks.

```csharp
class Program
{
    {
        int pin = 12;
        PinMode pinMode = PinMode.InputPullUp;
        using GpioController controller = new();

        for(int i = 0; i < 3; i++)
        {
            Console.WriteLine($"Iteration: {i+1}");
            OpenAndRegisterPin(controller, pin, pinMode);
            Console.WriteLine($"Try pin:{pin} events in 30 seconds.");
            Task.Delay(30000).Wait();
            CloseAndUnregisterPin(controller, pin);
        }
    }

    protected static void OpenAndRegisterPin(GpioController controller, int pin, PinMode pinMode)
    {
        Console.WriteLine($"Opening Pin: {pin}");
        controller.OpenPin(pin, pinMode);

        Console.WriteLine($"Registering pin events.");
        controller.RegisterCallbackForPinValueChangedEvent(pin, PinEventTypes.Falling, OnFalling);
        controller.RegisterCallbackForPinValueChangedEvent(pin, PinEventTypes.Rising, OnRising);
    }

    protected static void CloseAndUnregisterPin(GpioController controller, int pin)
    {
        Console.WriteLine($"Unregistering pin events.");
        controller.UnregisterCallbackForPinValueChangedEvent(pin, OnFalling);
        controller.UnregisterCallbackForPinValueChangedEvent(pin, OnRising);

        Console.WriteLine($"Closing Pin: {pin}");
        controller.ClosePin(pin);
    }

    protected static void OnFalling(object sender, PinValueChangedEventArgs ev) 
    {
        Console.WriteLine($"OnFalling Pin:{ev.PinNumber}");
    }
    protected static void OnRising(object sender, PinValueChangedEventArgs ev) 
    {
        Console.WriteLine($"OnRising Pin:{ev.PinNumber}");
    }
}
```

**Expected behavior**

In all iterations, the console must show OnFallling and/or OnRising messages with the Pin:12.
For example:
```
Iteration: 1
Opening Pin: 12
Registering pin events.
Try pin:12 events in 30 seconds.
OnFalling Pin:12
OnRising Pin:12
Unregistering pin events.
Closing Pin: 12
Iteration: 2
Opening Pin: 12
Registering pin events.
Try pin:12 events in 30 seconds.
OnFalling Pin:12
OnRising Pin:12
Unregistering pin events.
Closing Pin: 12
Iteration: 3
Opening Pin: 12
Registering pin events.
Try pin:12 events in 30 seconds.
OnFalling Pin:12
OnRising Pin:12
Unregistering pin events.
Closing Pin: 12
```

**Actual behavior**

Only in the iteration 1 (i == 0) the event messages are showed.
In the next iterations, this doesn't happened.
For example:
```
Iteration: 1
Opening Pin: 12
Registering pin events.
Try pin:12 events in 30 seconds.
OnFalling Pin:12
OnRising Pin:12
Unregistering pin events.
Closing Pin: 12
Iteration: 2
Opening Pin: 12
Registering pin events.
Try pin:12 events in 30 seconds.
Unregistering pin events.
Closing Pin: 12
Iteration: 3
Opening Pin: 12
Registering pin events.
Try pin:12 events in 30 seconds.
Unregistering pin events.
Closing Pin: 12
```

**Versions used**

Checked In Raspberry Pi 4 Model B with Raspberry Pi OS

```
SDK de .NET (que refleje cualquier global.json):
 Version:   5.0.400
 Commit:    d61950f9bf

```

- `dotnet --info` on the machine where app is being run (not applicable for self-contained apps)
```
SDK de .NET (que refleje cualquier global.json):
 Version:   5.0.400
 Commit:    d61950f9bf
```
Also checked with:
```
SDK de .NET Core (reflejando cualquier global.json):
 Version:   3.1.411
 Commit:    d100cdf718
``` 

- Version of `System.Device.Gpio` package
```
1.5.0
Also checked with:
1.3.0
```

nmunozgarcia:
Fixes #1637

When you registered callbacks in GpioController (when using SysFsDriver), their events are fired as expected.
Then, if you unregistered **all** callbacks, the events are not fired (this is the expected behavior).
Now, If you registered callbacks again (the same or others), their events are not fired. 
This is the bug.

The problem is in SysFsDriver.cs file. The _eventThreadCancellationTokenSource variable is created the first time when SysFsDriver class is also created.
Then, when all callbacks are removed, the _eventThreadCancellationTokenSource variable is canceled.
And, when new callbacks are added, the _eventThreadCancellationTokenSource variable is not recreated.

This fix contains some changes in SysFsDriver.cs file and a test in GpioControllerTestBase.cs file.
Now, when the thread (to listen to the events) is going to start, then the _eventThreadCancellationTokenSource variable is re-created.
The _eventThreadCancellationTokenSource variable is now set as nullable. But, maybe, this fix can be achieved in a better way.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1653)

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
@@ -24,13 +24,13 @@ namespace System.Device.Gpio.Drivers
 
         private static readonly int s_pinOffset = ReadOffset();
 
-        private readonly CancellationTokenSource _eventThreadCancellationTokenSource;
         private readonly List<int> _exportedPins = new List<int>();
         private readonly Dictionary<int, UnixDriverDevicePin> _devicePins = new Dictionary<int, UnixDriverDevicePin>();
         private TimeSpan _statusUpdateSleepTime = TimeSpan.FromMilliseconds(1);
         private int _pollFileDescriptor = -1;
         private Thread? _eventDetectionThread;
         private int _pinsToDetectEventsCount;
+        private CancellationTokenSource? _eventThreadCancellationTokenSource;
 
         private static int ReadOffset()
         {
@@ -69,8 +69,6 @@ namespace System.Device.Gpio.Drivers
             {
                 throw new PlatformNotSupportedException($"{GetType().Name} is only supported on Linux/Unix.");
             }
-
-            _eventThreadCancellationTokenSource = new CancellationTokenSource();
         }
 
         /// <summary>
@@ -518,7 +516,11 @@ namespace System.Device.Gpio.Drivers
                 {
                     try
                     {
-                        _eventThreadCancellationTokenSource.Cancel();
+                        if (_eventThreadCancellationTokenSource != null)
+                        {
+                            _eventThreadCancellationTokenSource.Cancel();
+                            _eventThreadCancellationTokenSource.Dispose();
+                        }
                     }
                     catch (ObjectDisposedException)
                     {
@@ -543,8 +545,11 @@ namespace System.Device.Gpio.Drivers
             {
                 try
                 {
-                    _eventThreadCancellationTokenSource.Cancel();
-                    _eventThreadCancellationTokenSource.Dispose();
+                    if (_eventThreadCancellationTokenSource != null)
+                    {
+                        _eventThreadCancellationTokenSource.Cancel();
+                        _eventThreadCancellationTokenSource.Dispose();
+                    }
                 }
                 catch (ObjectDisposedException)
                 {
@@ -618,13 +623,14 @@ namespace System.Device.Gpio.Drivers
                 {
                     IsBackground = true
                 };
+                _eventThreadCancellationTokenSource = new CancellationTokenSource();
                 _eventDetectionThread.Start();
             }
         }
 
         private void DetectEvents()
         {
-            while (_pinsToDetectEventsCount > 0)
+            while (_pinsToDetectEventsCount > 0 && _eventThreadCancellationTokenSource != null)
             {
                 try
                 {
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
@@ -630,7 +630,12 @@ namespace System.Device.Gpio.Drivers
 
         private void DetectEvents()
         {
-            while (_pinsToDetectEventsCount > 0 && _eventThreadCancellationTokenSource != null)
+            if (_eventThreadCancellationTokenSource == null)
+            {
+                throw new InvalidOperationException("Cannot start to detect events when CancellationTokenSource is null.");
+            }
+
+            while (_pinsToDetectEventsCount > 0)
             {
                 try
                 {
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -371,6 +371,46 @@ namespace System.Device.Gpio.Tests
         }
 
         [Fact]
+        public void AddCallbackRemoveAllCallbackAndAddCallbackAgainTest()
+        {
+            int callback1FiredTimes = 0, callback2FiredTimes = 0;
+            using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+            {
+                controller.OpenPin(InputPin, PinMode.Input);
+                controller.OpenPin(OutputPin, PinMode.Output);
+                controller.Write(OutputPin, PinValue.Low);
+                Thread.Sleep(WaitMilliseconds);
+
+                for (int i = 1; i < 3; i++)
+                {
+                    controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Rising, Callback1Rising);
+                    controller.RegisterCallbackForPinValueChangedEvent(InputPin, PinEventTypes.Falling, Callback2Falling);
+
+                    controller.Write(OutputPin, PinValue.High);
+                    Thread.Sleep(WaitMilliseconds);
+                    controller.Write(OutputPin, PinValue.Low);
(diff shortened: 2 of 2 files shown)


pgrawehr (approved):
Looks good to me. Exactly as expected.

raffaeler:
Well done. Looks good to me as well /cc @joperezr

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
```diff
@@ -518,7 +516,11 @@ private void RemovePinFromPoll(int pinNumber, ref int valueFileDescriptor, ref i
                 {
                     try
                     {
-                        _eventThreadCancellationTokenSource.Cancel();
+                        if (_eventThreadCancellationTokenSource != null)
```
joperezr, line 519: When could this be null? Is this just a check in case there are two different threads removing callbacks? If so, we should probably use a semaphore or lock instead of just a null check here.

joperezr (commented):
Thanks so much for this great contribution @nmunozgarcia. Left a small question but this looks good otherwise.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
nmunozgarcia: I had to add that condition because _eventThreadCancellationTokenSource is nullable after my changes. And I had to make _eventThreadCancellationTokenSource nullable because the "Nullable reference types" feature is enabled.

I had two options:
1. In the SysFsDriver constructor, I could assign a null with an exclamation mark at the end to point that I know what I'm doing when I left _eventThreadCancellationTokenSource no created. I mean this: 
`_eventThreadCancellationTokenSource = null!;`

2. I could make _eventThreadCancellationTokenSource nullable (this is what I did) and add new conditions to avoid compiler errors.
For example: In DetectEvents function, in the first while, I had to add _eventThreadCancellationTokenSource != null;
`while (_pinsToDetectEventsCount > 0 && _eventThreadCancellationTokenSource != null)`
I know _eventThreadCancellationTokenSource is not null because _pinsToDetectEventsCount is greather than 0, but the compiler wouldn't let me compile if I didn't show _eventThreadCancellationTokenSource was not null (adding the 
`_eventThreadCancellationTokenSource != null`)

Maybe the conditions make the project run more slowly?

joperezr: Ok, thanks for pointing out that this was due to nullability concerns. I think I'm okay with the changes you have except for one thing, which is the way you added this condition on the DetectEvents method. I will add a comment with my suggestion there, and explain why I believe we should make t hat change.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
```diff
@@ -618,13 +623,14 @@ private void InitializeEventDetectionThread()
...
                 };
+                _eventThreadCancellationTokenSource = new CancellationTokenSource();
                 _eventDetectionThread.Start();
             }
         }
 
         private void DetectEvents()
         {
-            while (_pinsToDetectEventsCount > 0)
+            while (_pinsToDetectEventsCount > 0 && _eventThreadCancellationTokenSource != null)
```
joperezr, line 633:
```suggestion
            if (_eventThreadCancellationTokenSource == null)
            {
                throw new InvalidOperationException("Cannot start to detect events when CancellationTokenSource is null.");
            }

            while (_pinsToDetectEventsCount > 0)
```

The other two places where you added the null check are just cancelling and disposing the token source (and not setting it to null which I think in this case is correct) and TBH I'm fine with the way you have it where if the condition evaluates to false (meaning the cancellation token source is null) then it is fine to just silently continue, since all those methods are trying to do is to cancel the token and then wait for the thread to exit.

That said, for this use I don't like the fact that we are adding the condition to the while, because a) this should never be true as in theory the CTS should be initialized by `InitializeEventDetectionThread` method before calling this, and b) if that is ever not the case (meaning if the CTS is null on this thread) I much rather consider that as a failed state and we should throw an exception instead of just failing silently, as nobody is setting the CTS to null and nobody should call it while CTS hasn't been initialized.

joperezr: If you agree with my suggestion, I'll go ahead and commit the suggestion and merge the PR as I think apart from that comment this is ready to go.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
nmunozgarcia: In that while condition, I was worry about slow down the event detection thread. I never thought the Nullable reference types feature to let me add a condition before checking if _eventThreadCancellationTokenSource was null and go away from the function with an exception if it's true.

Yeah, I fully agree with your suggestion. And it solves my worry about the efficiency of the loop condition.

PR #1654, opened 2021-09-03
Fix projects with double S.D.G imports

joperezr:
@Ellerbach reported today that VS had yellow warning signs in references in most of the binding projects. 
image

The reason why that was happening is because System.Device.Gpio project reference was added twice to all bindings, one global import, and one individual import per project. This PR is removing all of the individual imports in favor of just keeping the one global import, and I have validated that after doing this, the problem no longer repros:
image

Changes to projects where made using a small program I wrote which uses the MSBuild project model to make the changes, which is why most new-lines are being removed from .csprojs as a side effect.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1654)

### DIFF
--- a/src/devices/AD5328/AD5328.csproj
+++ b/src/devices/AD5328/AD5328.csproj
@@ -1,20 +1,12 @@
 <Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
     <TargetFrameworks>$(DefaultBindingTfms)</TargetFrameworks>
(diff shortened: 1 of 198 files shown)


Ellerbach (approved):
Thanks for fixing it! Looking forward to have it merged!

PR #1662, opened 2021-09-17
Fixes  #1626

### LINKED ISSUE #1626: VirtualOutputSegment Write offset

TheZlodziej:
When using a segment with bigger length than the given value, it writes wrong bytes (as if the offset was being calculated from the left). As mentioned in the #1626 issue:
```
VirtualOutputSegment segment = new(12);
segment.Write(0b_1001_0110);

// Result (offset to the left side of the byte value)
// segment index: | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11
//         value: | 0 | 0 | 0 | 0 | 0 | 1 | 1 | 0 | 1 | 0 |  0 |  1

// Expected result (offset to the right side of the byte value)
// segment index: | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11
//         value: | 0 | 1 | 1 | 0 | 1 | 0 | 0 | 1 | 0 | 0 |  0 |  0
``

### DIFF
--- a/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
+++ b/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
@@ -56,11 +56,8 @@ namespace Iot.Device.Multiplexing.Utility
         /// Does not display output.
         /// </summary>
         public void Write(byte value)
-        {
             // Write to 8 right-most segment values
-            int offset = _length - 8;
-            WriteByteAsValues(value, offset);
-        }
+            => WriteByteAsValues(value, 0);
 
         /// <summary>
         /// Writes discrete underlying bits to a virtual output.


Ellerbach:
Thanks for the submission, quick question: how did you test this?
Other than that, looks good to me. I'll let @richlander check this as he is the original author.

TheZlodziej:
Hi, I tested it with following:
```C#
[Fact]
public void VirtualOutputSegmentWriteByteOffset()
{
    // segment length > written value
    VirtualOutputSegment segment = new(12);
    segment.Write(0b_1001_0110);

    Assert.True(
        segment[0] == 0 &&
        segment[1] == 1 &&
        segment[2] == 1 &&
        segment[3] == 0 &&
        segment[4] == 1 &&
        segment[5] == 0 &&
        segment[6] == 0 &&
        segment[7] == 1);
} // passed

[Fact]
public void VirtualOutputSegmentWriteByte()
{
    VirtualOutputSegment segment = new(8);
    segment.Write(0b_1001_0110);

    Assert.True(
        segment[0] == 0 &&
        segment[1] == 1 &&
        segment[2] == 1 &&
        segment[3] == 0 &&
        segment[4] == 1 &&
        segment[5] == 0 &&
        segment[6] == 0 &&
        segment[7] == 1);
} // passed
 ```

Ellerbach (approved):
Looks good to me. Maybe @richlander can have a look. Thanks again for fixing this.

PR #1668, opened 2021-09-17
Add drivers for Sensirion SHT4x and SCD4x

scalablecory:
This adds new `Sht4x` and `Scd4x` drivers. Tested on Raspberry Pi 4B.

I didn't include it in this PR because I'm lacking the devices to test with, but I also refactored a couple of the other Sensirion drivers to share some code. Will open PR for this later.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1668)


src/devices/Device-Index.md
```diff
@@ -80,7 +80,9 @@
 * [Sense HAT](SenseHat/README.md)
 * [SensorHub - Environmental sensor](SensorHub/README.md)
 * [Servo Motor](ServoMotor/README.md)
+* [SCD4x - CO<sub>2</sub>, Temperature & Humidity Sensor](Scd4x/README.md)
```
joperezr, line 83: No need for you to update this file manually. We have a tool that we run in a ~monthly cadence, which checks all the device bindings we have and updates this list automatically.

src/devices/Scd4x/Scd4x.cs
```diff
@@ -0,0 +1,240 @@
...
+        /// </para>
+        ///
+        /// <para>
+        /// <see cref="StartPeriodicMeasurements"/> must be called first.
+        /// </para>
+        /// </summary>
+        /// <returns>
+        /// The time that the caller must wait before calling <see cref="EndCheckDataReady"/> to retreive the current data ready status.
+        /// </returns>
+        public TimeSpan BeginCheckDataReady()
```
joperezr, line 138: What is the benefit of having this method be public? Same goes for EndCheckDataReady method. Seems to me these could be private as they get called by CheckDataReady method.

src/devices/Scd4x/Scd4x.cs
```diff
@@ -0,0 +1,240 @@
...
+            return Sensirion.ReadUInt16BigEndianAndCRC8(buffer) is ushort response && (response & 0x7FF) != 0;
+        }
+
+        /// <summary>
+        /// <para>
+        /// Reads a periodic CO₂, humidity, and temperature measurement from the sensor.
+        /// </para>
+        ///
+        /// <para>
+        /// <see cref="CheckDataReady"/> should be called first to ensure a measurement is available. Once read, data will no longer be ready.
```
joperezr, line 163: Is there any reason why this method's implementation shouldn't call CheckDataReady? If there is not, you could move that here and then get rid of a lot of public surface area like the begin/end data ready as well as the CheckDataReady, and only leave the ReadPeriodicMeasurment member public

src/devices/Scd4x/Scd4x.cs
```diff
@@ -0,0 +1,240 @@
...
+        /// </para>
+        /// </summary>
+        /// <returns>
+        /// The time that the caller must wait before calling <see cref="EndReadPeriodicMeasurement"/> to retreive the measurement.
+        /// </returns>
+        public TimeSpan BeginReadPeriodicMeasurement()
```
joperezr, line 186: Same coment here, not sure why we have Begin/End members public. If there is a scenario for running this async, then perhaps we should add Task-based (async and await) methods instead.

src/devices/Shared/Sensirion.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// Common code for Sensirion devices.
+    /// </summary>
+    internal static class Sensirion
+    {
+        /// <returns>If the CRC8 matched the data, a <see cref="ushort"/> of data. Otherwise, <see langword="null"/>.</returns>
+        public static ushort? ReadUInt16BigEndianAndCRC8(ReadOnlySpan<byte> bytes)
+        {
+            Debug.Assert(bytes.Length >= 3, $"{nameof(bytes)} must contain at least 3 bytes.");
+
+            _ = bytes[2];
```
joperezr, line 20: mm not sure I understand why you have this line.

src/devices/Sht4x/Sht4x.cs
```diff
@@ -0,0 +1,154 @@
...
+            return EndReadHumidityAndTemperature();
+        }
+
+        /// <summary>
+        /// Begins reading relative humidity and temperature.
+        /// </summary>
+        /// <returns>
+        /// The time that the caller must wait before calling <see cref="EndReadHumidityAndTemperature"/> to retrieve the measurement.
+        /// </returns>
+        public TimeSpan BeginReadHumidityAndTemperature(Sht4xRepeatability repeatability = Sht4xRepeatability.High)
```
joperezr, line 102: Similar comments than the other type, perhaps some of these members should be private, in favor of making the public API surface easier to understand and use.

src/devices/Scd4x/Scd4x.cs
scalablecory: My goal is to allow a single thread to efficiently interleave talking to multiple devices at a time. Basically:

1. Write command to Device A.
2. Write command to Device B.
3. Wait enough time for both commands to complete.
4. Read response from Device A.
5. Read response from Device B.

There exist similar async APIs in some Arduino libraries.

Tasks feel problematic here:

- Platform APIs for I2C are all sync.
    - This means that async would be "fake" and you'd be reliant on the thread pool scaling up.
- Because I2C only supports a single I/O (steps 1, 2, 4, and 5 above) at a time, these platform APIs do their own locks.
    - This means if you have N outstanding I/Os, you'll have N blocked threads. :(

Admittedly this is probably not a huge issue -- I suspect in many cases one will have very few I2C devices. Maybe it's fine to just ignore the async-over-sync problem and hope nobody loads up their I2C bus. We could improve the scenario a little bit by making I2cDevice have async APIs that synchronize around a bus-wide `SemaphoreSlim`, so only one thread is ever blocked at once. The overhead tradeoffs would be interesting.

I didn't feel it was appropriate to make this choice for the caller, so instead chose to just expose the low-level primitives needed to implement it whatever way one might want, without forcing down an async/await path.

src/devices/Shared/Sensirion.cs
scalablecory: Micro-optimization. It forces a single bounds check on the span. Otherwise, the in-order access to `bytes[0]`, `bytes[1]`, `bytes[2]` will result in three bounds checks.

src/devices/Scd4x/Scd4x.cs
scalablecory: I've taken a look at some async code from other drivers and see semi-async drivers like this is a thing. I've updated the code with this.

src/devices/Scd4x/Scd4x.cs
scalablecory: Good point, I've made this change.

PR #1674, opened 2021-09-26
DCMotorWithStartStop fix

### LINKED ISSUE #1675: DCMotorWithStartStop
**Describe the bug**
I wanted to use the DCMotorWithStartStop class by installing Iot.Device.Bindings as a nuget package.
It turned out that the class is not visible even though its code was added in #1398. From what I found out, one line is missing in the previous PR. When I cloned the repo and added ```<Compile Include = "DCMotorWithStartStop.cs" />``` to DCMotor.csproj it looks like the problem is fixed. 

**Versions used**

 OS Name:     Windows
 OS Version:  10.0.19042
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.100-rc.1.21463.6\

Host (useful for support):
  Version: 6.0.0-rc.1.21451.13
  Commit:  d7619cd4b1

.NET SDKs installed:
  3.1.413 [C:\Program Files\dotnet\sdk]
  5.0.100 [C:\Program Files\dotnet\sdk]
  5.0.401 [C:\Program Files\dotnet\sdk]
  6.0.100-rc.1.21463.6 [C:\Program Files\dotnet\sdk]

- Version of `System.Device.Gpio` package 1.5
- Version of `Iot.Device.Bindings` package 1.5

```<TargetFramework>net5.0</TargetFramework>```

### PR DESCRIPTION
**Pull request by wsad4ryba**
Fixes #1675

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1674)

### DIFF
--- a/src/devices/DCMotor/DCMotor.csproj
+++ b/src/devices/DCMotor/DCMotor.csproj
@@ -9,6 +9,7 @@
     <Compile Include="DCMotor3Pin.cs" />
     <Compile Include="DCMotor2PinNoEnable.cs" />
     <Compile Include="DCMotor2PinWithBiDirectionalPin.cs" />
+    <Compile Include="DCMotorWithStartStop.cs" />
     <None Include="README.md" />
   </ItemGroup>
   <ItemGroup>
--- a/src/devices/DCMotor/DCMotor.csproj
+++ b/src/devices/DCMotor/DCMotor.csproj
@@ -5,11 +5,7 @@
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="$(MSBuildThisFileDirectory)../SoftPwm/SoftwarePwm.csproj" />
-    <Compile Include="DCMotor.cs" />
-    <Compile Include="DCMotor3Pin.cs" />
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/devices/DCMotor/DCMotor.csproj [hunk 1]
```diff
@@ -9,6 +9,7 @@
     <Compile Include="DCMotor3Pin.cs" />
     <Compile Include="DCMotor2PinNoEnable.cs" />
     <Compile Include="DCMotor2PinWithBiDirectionalPin.cs" />
+    <Compile Include="DCMotorWithStartStop.cs" />
```
pgrawehr, line 12: Thanks for the fix. However,  wouldn't this better be changed to just a single Include="*.cs"? Seems to be what's used in most places. (If not we should check why the bindings differ in how the source is included)

wsad4ryba: You're right, it's a better way 👍

krwq (approved):

PR #1677, opened 2021-09-30
Porting the BMM150 from .NET nanoFramework

jantielens:
This is a port from of the BMM150 sensor (magnetometer) from the .NET nanoFramework ([link to binding](https://github.com/nanoframework/nanoFramework.IoT.Device/tree/develop/devices/Bmm150))

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1677)


Ellerbach (approved):
Thanks a lot for this migration. Approving as I reviewed the code from the .NET nanoFramework side and all seems ok to work on .NET IoT. Leaving someone else to comment if any. Great to see one more biding coming from .NET nanoFramework!

Ellerbach:
@jantielens, once you'll have fixed the build issues (you can use the local build.bat or build.sh), you can change the state of this PR.

src/devices/Bmm150/Bmm150.cs
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        /// <summary>
+        /// Constructor to use if Bmm150 is behind another element and need a special I2C protocol like
+        /// when used with the MPU9250
+        /// </summary>
+        /// <param name="i2cDevice">The I2C device</param>
+        /// <param name="Bmm150Interface">The specific interface to communicate with the Bmm150</param>
+        /// <param name="shouldDispose">True to dispose the I2C device when class is disposed</param>
+        public Bmm150(I2cDevice i2cDevice, Bmm150I2cBase Bmm150Interface, bool shouldDispose = true)
```
pgrawehr, line 85: I don't think we need the "shouldDispose" here. The `I2cDevice` instance is always bound to a binding and shall not be shared.

src/devices/Bmm150/Bmm150.cs
```diff
@@ -0,0 +1,319 @@
...
+        /// <summary>
+        /// True if there is a data to read
+        /// </summary>
+        public bool HasDataToRead => (ReadByte(Bmp180Register.DATA_READY_STATUS) & 0x01) == 0x01;
+
+        /// <summary>
+        /// Check if the version is the correct one (0x32). This is fixed for this device
+        /// </summary>
+        /// <returns>Returns true if the version match</returns>
+        public bool IsVersionCorrect => ReadByte(Bmp180Register.WIA) == 0x32;
```
pgrawehr, line 205: Unless the exception in the `Initialize` function can be prevented somehow, this never returns false for an outside caller. So maybe make private?

src/devices/Bmm150/Bmm150I2c.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System;
+using System.Device.I2c;
+
+namespace Iot.Device.Bmp180
+{
+    /// <summary>
+    /// Default I2C interface for the Bmm150
+    /// </summary>
+    public class Bmm150I2c : Bmm150I2cBase
```
pgrawehr, line 12: Why is this additional indirection class required? The I2cDevice itself is already intended to be swapped with different implementations

Ellerbach: Let me answer this one: BMI160 from Bosh include this sensor. To access it, similar to the AKS8963 you need a an abstraction like this one. See https://www.bosch-sensortec.com/products/motion-sensors/imus/bmi160/

src/devices/Bmm150/Bmm150.cs
Ellerbach: That is needed if you are using an I2C interface that is built over a regular I2C Device in the case of BMI160 for example.

src/devices/Bmm150/Bmm150I2c.cs
pgrawehr: I know the use case, but why can't the BMI160 directly implement the `I2cDevice` interface (or rather, provide it?). It's actually a kind of I2C hub.

Ellerbach: it can but just for the raw values. For the calibration, you need to access the BMM150 directly thru the BMI160 unfortunately.

pgrawehr: I'm sorry,  but I don't get your point. This interface is just a subset of `I2cDevice`. Nothing device-specific in here.

Ellerbach: To access BMM150 thru BMI160, you need to use BMI160 I2cDevice and its I2C address. So if you dispose it before BMI160, then you'll have issues when disposing BMI160. So it's the reason, the same as for AK8963 that the I2C Device you give should not be disposed before the main device is disposed. I hope it does make more sense.

pgrawehr: That applies to the `shouldDispose` above, but not to the question here: Why are the abstract classes `Bmm150I2cBase` and it's implementation `Bmm150I2c` required? The former is just a subset of `I2cDevice` and the later is just a forwarding adapter.

PR #1680, opened 2021-10-07
Porting Button from .NET nanoFramework

zoekdestep:
This is a port from the Button binding from the .NET nanoFramework (link to [binding](https://github.com/nanoframework/nanoFramework.IoT.Device/tree/develop/devices/Button)).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1680)

### DIFF
--- a/src/devices/Button/GpioButton.cs
+++ b/src/devices/Button/GpioButton.cs
@@ -107,15 +107,14 @@ namespace Iot.Device.Button
 
             if (disposing)
             {
+                _gpioController.UnregisterCallbackForPinValueChangedEvent(_buttonPin, PinStateChanged);
                 if (_shouldDispose)
                 {
-                    _gpioController?.UnregisterCallbackForPinValueChangedEvent(_buttonPin, PinStateChanged);
                     _gpioController?.Dispose();
                     _gpioController = null!;
                 }
                 else
                 {
-                    _gpioController.UnregisterCallbackForPinValueChangedEvent(_buttonPin, PinStateChanged);
                     _gpioController.ClosePin(_buttonPin);
                 }
             }


src/devices/Button/ButtonBase.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        private long _lastPress = DateTime.MinValue.Ticks;
+        private Timer? _holdingTimer;
+
+        /// <summary>
+        /// Delegate for button pressed.
+        /// </summary>
+        /// <param name="sender">The sender object.</param>
+        /// <param name="e">The event data.</param>
+        public delegate void ButtonPressedDelegate(object sender, EventArgs e);
```
pgrawehr, line 33: No need to declare these manually in .Net. Use `Action<T>`

src/devices/Button/ButtonHoldingState.cs
```diff
@@ -0,0 +1,20 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Button
+{
+    /// <summary>
+    /// The different button holding states.
+    /// </summary>
+    public enum ButtonHoldingState
+    {
+        /// <summary>Started</summary>
```
pgrawehr, line 11: Documentation could be improved. The meaning of this enum is not obvious

src/devices/Button/GpioButton.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// <summary>
+        /// Initialization of the button.
+        /// </summary>
+        /// <param name="buttonPin">GPIO pin of the button.</param>
+        /// <param name="pullUp">If the system is pullup (false = pulldown).</param>
+        /// <param name="doublePress">Max ticks between button presses to count as doublepress.</param>
+        /// <param name="holding">Min ms a button is pressed to count as holding.</param>
+        /// <param name="gpio">Gpio Controller.</param>
+        /// <param name="shouldDispose">True to dispose the GpioController.</param>
+        public GpioButton(int buttonPin, bool pullUp, TimeSpan doublePress, TimeSpan holding, GpioController? gpio = null, bool shouldDispose = true)
```
pgrawehr, line 43: I think `pullUp` should be an enum, because we might want to support internal pull-ups/pull-downs as well.

src/devices/Button/GpioButton.cs
```diff
@@ -0,0 +1,126 @@
...
+                    }
+
+                    break;
+            }
+        }
+
+        /// <summary>
+        /// Cleanup.
+        /// </summary>
+        public new void Dispose()
```
pgrawehr, line 92: Why is this required?

src/devices/Button/GpioButton.cs
```diff
@@ -0,0 +1,126 @@
...
+                    _gpioController = null!;
+                }
+                else
+                {
+                    _gpioController.UnregisterCallbackForPinValueChangedEvent(_buttonPin, PinStateChanged);
+                    _gpioController.ClosePin(_buttonPin);
+                }
+
+                base.Dispose(disposing);
+                _disposed = true;
```
pgrawehr, line 122: Put base call outside of if.

src/devices/Button/GpioButton.cs
```diff
@@ -0,0 +1,126 @@
...
+            if (disposing)
+            {
+                if (_shouldDispose)
+                {
+                    _gpioController?.Dispose();
+                    _gpioController = null!;
+                }
+                else
+                {
+                    _gpioController.UnregisterCallbackForPinValueChangedEvent(_buttonPin, PinStateChanged);
```
pgrawehr, line 117: This should be done in either case.

src/devices/Button/GpioButton.cs
zoekdestep: Thank you! Should be fixed. Will resolve for now - if you feel like the fix doesn't answer your comment, feel free to reopen the conversation.

src/devices/Button/GpioButton.cs
pgrawehr: @Ellerbach It probably is, but still for consistency, I would say we should unregister before disposing.

src/devices/Button/ButtonBase.cs
pgrawehr: @zoekdestep Thanks. In such cases, you can just silently resolve a conversation. This is considered equivalent to "I think I have addressed the issue". Less typing ;-)

PR #1683, opened 2021-10-08
Do not cast the result of non-generic Image.Load()

antonfirsov:
Non-generic `Image.Load()` is not guaranteed to return `Image<Rgba32>`, so use the generic `Image.Load<Rgba32>` instead.

~The related [README.MD](https://github.com/dotnet/iot/blob/148fc42738580221e7885eb5c1059e9830213919/src/devices/Pcd8544/README.md) should be fixed too, I will try to do it on the web interface.~ *Edit: Done.* (Don't have the chance to clone the repo, just browsed through ImageSharp use cases and spotted this.)


krwq (approved):
nice find, thanks!

PR #1684, opened 2021-10-08
Offset in VirtualOutputSegment.Write(byte value) shouldn't be calculated (is always 0).

### LINKED ISSUE #1626: VirtualOutputSegment Write offset

TheZlodziej:
Fixes #1626

The #1627 didn't fix the #1626 issue. Since the PR was merged and it had conflict with mine, I'm redoing my changes.

The offset in VirtualOutputSegment.Write(byte value) is calculated with following formula:
```c#
int offset = _length - 8;
```
when in reality it should always be 0.

I tested my code with the same tests as last time:
```c#
[Fact]
public void VirtualOutputSegmentWriteByteOffset()
{
    // segment length > written value
    MyOptSeg segment = new(12);
    segment.Write(0b_1001_0110);

    Assert.True(
        segment[0] == 0 &&
        segment[1] == 1 &&
        segment[2] == 1 &&
        segment[3] == 0 &&
        segment[4] == 1 &&
        segment[5] == 0 &&
        segment[6] == 0 &&
        segment[7] == 1);
} // passed

[Fact]
public void VirtualOutputSegmentWriteByte()
{
    MyOptSeg segment = new(8);
    segment.Write(0b_1001_0110);

    Assert.True(
        segment[0] == 0 &&
        segment[1] == 1 &&
        segment[2] == 1 &&
        segment[3] == 0 &&
        segment[4] == 1 &&
        segment[5] == 0 &&
        segment[6] == 0 &&
        segment[7] == 1);
} // passed
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1684)

### DIFF
--- a/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
+++ b/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
@@ -80,7 +80,6 @@ namespace Iot.Device.Multiplexing.Utility
                 WriteByteAsValues(value[i], offset);
             }
         }
-        }
 
         private void WriteByteAsValues(byte value, int offset)
         {
--- a/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
+++ b/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
@@ -66,9 +66,9 @@ namespace Iot.Device.Multiplexing.Utility
         public void Write(ReadOnlySpan<byte> value)
         {
             // Scenarios
-            // values can be shorter than byteLength
-            // values can be longer than byteLength
-            // values can be same as byteLength
+            // values can be shorter than byteLength e.g. (1 * 8) - 16 = -8 < 0
+            // values can be longer than byteLength e.g. (3 * 8) - 16 = 8 > 0
+            // values can be same as byteLength e.g. (2 * 8) - 16 = 0
             if (value.Length * 8 > _length)
             {
                 throw new ArgumentException($"The bytes provided exceed the length of the {nameof(IOutputSegment)}.");
--- a/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
+++ b/src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
@@ -66,20 +66,20 @@ namespace Iot.Device.Multiplexing.Utility
         public void Write(ReadOnlySpan<byte> value)
         {
             // Scenarios
-            // values can be shorter than byteLength e.g. (1 * 8) - 16 = -8 < 0
-            // values can be longer than byteLength  e.g. (3 * 8) - 16 =  8 > 0
-            // values can be same as byteLength      e.g. (2 * 8) - 16 =      0
-            int offset = (value.Length * 8) - _length;
-            if (offset > 0)
+            // values can be shorter than byteLength
+            // values can be longer than byteLength
+            // values can be same as byteLength
+            if (value.Length * 8 > _length)
             {
-                throw new Exception($"The bytes provided exceed the length of the {nameof(IOutputSegment)}.");
+                throw new ArgumentException($"The bytes provided exceed the length of the {nameof(IOutputSegment)}.");
             }
 
             for (int i = 0; i < value.Length; i++)
             {
-                int index = value.Length - i - 1;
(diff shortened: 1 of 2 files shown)


joperezr:
Sorry for the confusion here, and thanks for re creating your PR. Can you please also include those two unit tests as part of this PR in order to get some regression testing?

src/devices/Common/Iot/Device/Multiplexing/tests/OutputSegmentTests.cs
```diff
@@ -29,6 +29,24 @@ public void SegmentValuesWritePinValues()
...
         }
+        
+        [Fact]
+        public void SegmentValuesWriteByteOffset()
+        {
+            // segment length > written value
+            VirtualOutputSegment segment = new(12);
+            segment.Write(0b_1001_0110);
+
+            Assert.True(
```
krwq, line 40: Can you instead do:

```csharp
Assert.Equal(0, segment[0]);
// ... 
```

also, shouldn't Write write to 8 rightmost bit rather than leftmost? (segment[4]..segment[11])

krwq: per offline conversation right most means least significant which implies index=0

src/devices/Common/Iot/Device/Multiplexing/VirtualOutputSegment.cs
```diff
@@ -70,19 +66,18 @@ public void Write(byte value)
         public void Write(ReadOnlySpan<byte> value)
         {
             // Scenarios
-            // values can be shorter than byteLength e.g. (1 * 8) - 16 = -8 < 0
-            // values can be longer than byteLength  e.g. (3 * 8) - 16 =  8 > 0
-            // values can be same as byteLength      e.g. (2 * 8) - 16 =      0
-            int offset = (value.Length * 8) - _length;
-            if (offset > 0)
+            // values can be shorter than byteLength
+            // values can be longer than byteLength
```
joperezr, line 70: NIT: the comments of the examples where actually nice, can we bring them back?

PR #1695, opened 2021-10-18
#1694 Handle `STATUS_REG` error correctly: distinguishing ext. temp and light sensor

### LINKED ISSUE #1694: SensorHub: Improper handling of STATUS_REG
**Describe the bug**

There's external temperature sensor and light sensor on [Sensor Hub](https://wiki.52pi.com/index.php/DockerPi_Sensor_Hub_Development_Board_SKU:_EP-0106) which status can be read from single registry (`STATUS_REG (0x04)`).

For some reason when board is placed in complete darkness, light sensor reports failure (value `status = 0x08`) - causing check for no error to fail when reading external temperature.

At same time (in complete darkness) it is possible to read external sensor temperature directly from i2c.

**Steps to reproduce**

```csharp
// turn off lights (causing light sensor to fail)
if (sh.TryReadOffBoardTemperature(out var t))
{
    Console.WriteLine($"OffBoard temperature {t}");
}
else
{
    Console.WriteLine("You can measure external temperature only with lights on!");
}
```

**Expected behavior**
Output:
```
OffBoard temperature 23 °C
```

**Actual behavior**
Output:
```
You can measure external temperature only with lights on!
```

In complete darkness, light sensor reports fail trough `STATUS_REG (0x04)`, value `0x08` which causes `TryReadOffBoardTemperature` to not read value from external sensor:
```
            if (status.Equals((byte)RegisterStatusErrors.NO_ERROR))
            {
                temperature = Temperature.FromDegreesCelsius(ReadRegister(Register.TEMP_REG));
                return true;
            }

            return false; // <-- returns here
```

**Versions used**

- using .net 5.0 self-contained build `dotnet publish -c Release -r linux-arm64 -o app/`
- `Iot.Device.Bindings` `1.5.0`
- `System.Device.Gpio` `1.5.0` (transient dependency)

wdolek:
Fixes #1694 

Fixing improper handling of status provided by `STATUS_REG (0x04)` which combines status of two different sensors.

I followed way how current code was implemented but removed `RegisterStatusErrors` as it is bit confusing - there are three different status registers on board (`STATUS_REG`, `ON_BOARD_SENSOR_ERROR`, `BMP280_STATUS`) - only two latter ones are having simply `0` = no error, `1` = error.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1695)

### DIFF
--- a/src/devices/SensorHub/SensorHub.cs
+++ b/src/devices/SensorHub/SensorHub.cs
@@ -19,6 +19,8 @@ namespace Iot.Device.SensorHub
         /// </summary>
         public const int DefaultI2cAddress = 0x17;
 
+        internal const byte NO_ERROR = 0x0;
+
         private readonly I2cDevice _i2cDevice;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


pgrawehr (approved):

aexhg:
Thanks for fixing - I wasn't using the light sensor!

joperezr (approved):

PR #1710, opened 2021-10-31
Add Markdownlint to improve Markdown produced quality

Ellerbach:
- Add Markdownlint to the entire repository
- Cleaned non compliant Markdown file
- Added few exception to skip fully or part of the linter like autogenerated pages
- Adjusting pipeline for the tool to be executed.
  - Note: the markdownlint is executed only on the Windows pipeline as it's the same anyway whatever platform

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1710)


.vsts-ci.yml
```diff
@@ -128,7 +134,7 @@ stages:
             artifactName: BuiltPackageOutputs
             artifactType: container
           condition: eq(variables['_BuildConfig'], 'Release')
-      
+
```
krwq, line 137: is this change needed?

src/devices/UFireIse/README.md
```diff
@@ -29,14 +29,14 @@ To read the ORP (OxidationReductionPotential) value use this example
...
 {
-	if (uFireOrp.TryMeasureOxidationReductionPotential(out ElectricPotential orp))
-	{
-		Console.WriteLine("Eh:" + orp.Millivolts);
-	}
-	else
-	{
-		Console.WriteLine("Not possible to measure pH");
-	}
+ if (uFireOrp.TryMeasureOxidationReductionPotential(out ElectricPotential orp))
```
krwq, line 32: code indentation seemed better before

.vsts-ci.yml
Ellerbach: no, not sure how it happened. I removed the spaces

src/devices/UFireIse/README.md
Ellerbach: fixed

Ellerbach:
@krwq and @joperezr if you can have another look so we can move forward it would be great. Every time we merge Markdown now without this, we may have to adjust them later on ;-)

.markdownlint.json
```diff
@@ -0,0 +1,3 @@
+{
+    "MD013": false
```
joperezr, line 2: NIT: can we add a comment here explaining what does this code mean?

.vsts-ci.yml
```diff
@@ -100,6 +100,12 @@ stages:
         steps:
         - checkout: self
           clean: true
+        # Scan markdown files on style consistency
+        - powershell: |
```
joperezr, line 104: Looks like this step takes about between 30 s and a minute, which for now is fine. If we expect it to take longer, then we should consider disabling the step for official builds and only running on CI.

Ellerbach: That's correct, it can only run for PR. As once in the repo, it should be clean. And there is no need to run it for release.
So any recommendation on the best way to adjust this?

.markdownlint.json
Ellerbach: Json files should not includes comments :-) But I'll add it some documentation. I'll add a page dedicated to help our developer dealing with Markdown linter.
What do you think if I add this one I've created for nanoFramework in there as well (I'll adjust it for us of course)?
https://github.com/nanoframework/nanoframework.github.io/blob/pages-source/content/contributing/markdown-creation.md

joperezr (approved):

PR #1711, opened 2021-11-02
added ht1632

HumJ0218:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1711)


src/devices/Ht1632/Ht1632.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        private GpioController? _controller;
+
+        /// <summary>
+        /// HT1632C 32x8 and 24x16 LED Driver
+        /// </summary>
+        /// <param name="pinMapping">The pin mapping to use by the binding.</param>
+        /// <param name="gpioController">The GPIO Controller used for interrupt handling.</param>
+        /// <param name="shouldDispose">True (the default) if the GPIO controller shall be disposed when disposing this instance.</param>
+        public Ht1632(Ht1632PinMapping pinMapping, GpioController? gpioController = null, bool shouldDispose = true)
```
HumJ0218, line 82: Do I need to add `PinNumberingScheme pinNumberingScheme = PinNumberingScheme.Logical` here?

src/devices/Ht1632/Command.cs
```diff
@@ -0,0 +1,160 @@
...
+    public enum Command : byte
+    {
+        // References middle 8-bits of command codes like:
+        // 100 0000-0000-X
+        //     ~~~~ ~~~~
+
+        /// <summary>
+        /// Turn off both system oscillator and LED duty cycle generator (default)
+        /// </summary>
+        SYS_DIS = 0b_0000_0000,
```
Ellerbach, line 18: For all the public enum and other members, we're using a naming convention that is no C style but C# style. Meaning that this would be without abbreviation when possible, with a capital letter at the beginning of the word, each word after each others without any separation. That will then give for this one: `DisactivateSystemAndLed`

src/devices/Ht1632/Command.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        /// <summary>
+        /// Turn on system oscillator
+        /// </summary>
+        SYS_EN = 0b_0000_0001,
```
Ellerbach, line 23: Here, it can be something like `EnableSystem`

src/devices/Ht1632/Command.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        /// <summary>
+        /// Turn off LED duty cycle generator (default)
+        /// </summary>
+        LED_Off = 0b_0000_0010,
```
Ellerbach, line 28: Here: `DisableLed`, etc

src/devices/Ht1632/Command.cs
Ellerbach: Now the question is: do you need to expose this `Command` public? In case of an internal or private enum, and if those are register like you find in the documentation, it's ok to keep them as they are.

src/devices/Ht1632/Ht1632.cs
Ellerbach: It's ok, you don't need, the controller will get this if needed.

src/devices/Ht1632/Ht1632PinMapping.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            CS = cs;
+            WR = wr;
+            DATA = data;
+        }
+
+        /// <summary>
+        /// Chip select input with pull-high resistor
+        /// </summary>
+        public int CS { get; set; }
```
Ellerbach, line 24: As already pointed out before, please use full names here. `ChipSelect`, `WriteClock`, `Data` or equivalent.

src/devices/Ht1632/Ht1632.cs
```diff
@@ -0,0 +1,189 @@
...
+            { 16, Command.PWM_Duty_16 },
+        };
+
+        private readonly Ht1632PinMapping _pinMapping;
+        private readonly int _cs;
+        private readonly int _wr;
+        private readonly int _data;
+        private readonly bool _shouldDispose;
+
+        private GpioController? _controller;
```
Ellerbach, line 74: Why do you want to have it nullable while you'll set it up in the constructor? You're checking that it's not null and create one if it is. so a private readonly should work as well  here without the nullability.

src/devices/Ht1632/Ht1632.cs
```diff
@@ -0,0 +1,189 @@
...
+        public void Dispose()
+        {
+            if (_shouldDispose)
+            {
+                _controller?.Dispose();
+                _controller = null;
+            }
+            else
+            {
+                _controller?.ClosePin(_cs);
```
Ellerbach, line 143: in theory to have it super clean, you're supposed to close also the pins even if you don't dispose it. So that can be common.
To avoid double disposing, you can use the _isDisposed pattern or equivalent as well.

src/devices/Ht1632/Ht1632.cs
```diff
@@ -0,0 +1,189 @@
...
+        private GpioController? _controller;
+
+        /// <summary>
+        /// HT1632C 32x8 and 24x16 LED Driver
+        /// </summary>
+        /// <param name="pinMapping">The pin mapping to use by the binding.</param>
+        /// <param name="gpioController">The GPIO Controller used for interrupt handling.</param>
+        /// <param name="shouldDispose">True (the default) if the GPIO controller shall be disposed when disposing this instance.</param>
+        public Ht1632(Ht1632PinMapping pinMapping, GpioController? gpioController = null, bool shouldDispose = true)
+        {
```
Ellerbach, line 83: This seems very much like SPI. I guess for performance it will be better. You can use Software SPI as well as this implementation seems very much like this. I may have missed something but what's the key difference here and why would SPI not be appropriate? I see some elements that may be sent less than by a group of 8 but what the doc says about that if we more clocks are sent?

src/devices/Ht1632/README.md
```diff
@@ -0,0 +1,20 @@
+﻿# Holtek HT1632 - 32×8 & 24×16 LED Driver
+
+The devices are a memory mapping LED display controller/driver, which can select a number of ROW and commons. These are 32 ROW & 8 Commons and 24 ROW & 16 Commons. The devices support 16-gradation LEDs for each out line using PWM control with software instructions. A serial interface is conveniently provided for the command mode and data mode. Only three or four lines are required for the interface between the host controller and the devices. The display can be extended by cascading the devices for wider applications.
+
+## Documentation
+
+- [Datasheet](https://www.holtek.com/documents/10179/116711/HT1632D_32D-2v100.pdf)
+
```
Ellerbach, line 8: Can you please add the usage?
Best with a fritzing or equivalent graphic. A picture of the result with the screen can be great (with the bot picture for example).

src/devices/Ht1632/Id.cs
```diff
@@ -0,0 +1,35 @@
...
+    internal enum Id : byte
+    {
+        // References first 3-bits of command codes like:
+        // 100 0000-0000-X
+        // ~~~
+
+        /// <summary>
+        /// READ Mode
+        /// </summary>
+        READ = 0b_110,
```
Ellerbach, line 18: For internals, it's fine to keep read in uppercase if it's like this in the documentation.

src/devices/Ht1632/Ht1632.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// Set COM option (default is N-MOS open drain output and 8 COM)
+        /// </summary>
+        public ComOption ComOption { set => WriteCommand((Command)value); }
+
+        /// <summary>
+        /// Set row PWM duty (1/16 to 16/16, default is 16)
+        /// </summary>
+        public byte PwmDuty { set => WriteCommand(s_pwmCommandMap[value]); }
+
+        private static readonly Dictionary<byte, Command> s_pwmCommandMap = new Dictionary<byte, Command>()
```
Ellerbach, line 48: What is the advantage of having a dictionary compare to a simple array?

Ellerbach (commented):
Few comments, mainly about cosmetic. Otherwise, nice addition!

src/devices/Ht1632/Command.cs
HumJ0218: they are from datasheet P20, just called "Command"

HumJ0218: names comes entirely from the datasheet

src/devices/Ht1632/Ht1632.cs
HumJ0218: I've considered SPI, but it doesn't require a complete 8-bit set of bits. 
The command format is like "XXX-XXXXXXXX-X"
The data format is like "XXX-XXXXXX-XXXX-XXXX-XXXX-..."
not 8-bit bytes
Using today's SPI interface can be difficult, or performance issues

src/devices/Ht1632/README.md
HumJ0218: OK. I will add equivalent graphic and my physical picture later

HumJ0218:
`ClockMode` `ComOption` is for setters. I give them friendly names
 `Command` is for write commands directly. I kept names from datasheet

src/devices/Ht1632/Command.cs
Ellerbach: Issue here is that they are public. We're allowing them to be like that for internal, not for public. In this case, it's public, so it must be adjusted to more developer friendly naming. If command does not need to be public, make it internal for example and it will be good like this

src/devices/Ht1632/Ht1632.cs
Ellerbach: ok, fair enough, let's leave it with GPIO for now.

Ellerbach:
> `ClockMode` `ComOption` is for setters. I give them friendly names `Command` is for write commands directly. I kept names from datasheet

for any public surface, please use friendly names. Command is public so it should have friendly names. If you can make it internal, it's ok. There are still others to fix.

src/devices/Ht1632/README.md
```diff
@@ -0,0 +1 @@
+﻿
```
pgrawehr: This file should not be empty, but include a small documentation on how to use the chip (idealy with a sample wiring diagram), references to the relevant data sheets, etc.

src/devices/Ht1632/samples/Program.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            var index = (x + 16 * y) / 4;
+            var value = (byte)(
+                (image[x + 0, y].R > 127 ? 0b00000001 << 3 : 0) |
+                (image[x + 1, y].R > 127 ? 0b00000001 << 2 : 0) |
+                (image[x + 2, y].R > 127 ? 0b00000001 << 1 : 0) |
+                (image[x + 3, y].R > 127 ? 0b00000001 << 0 : 0));
+            data[index] = value;
+        }
+    }
```
pgrawehr: I suggest to add this function to the device api. Drawing an image should be a common function.

HumJ0218: Got it.
Seems like `SixLabors.ImageSharp` lacks support for monochrome bitmaps :(
Which pixel format should I use here?

HumJ0218: or I just use `Image<TPixel> image` for input

pgrawehr: > or I just use `Image<TPixel> image` for input

Sounds good. Maybe there are conversion functions for TPixel, but I 'm also not familiar with the ImageSharp api yet.

HumJ0218:
This PR has been modified and can be reviewed. @pgrawehr @Ellerbach

src/devices/Ht1632/ClockMode.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Clock modes
+    /// </summary>
+    public enum ClockMode : byte
+    {
+        /// <summary>
+        /// Set slave mode and clock source from external clock, the system clock input from OSC pin and synchronous signal input from SYN pin
+        /// </summary>
+        Slave = Command.Slave,
```
Ellerbach, line 14: Would it be possible to use more inclusive terms? Rather than Slave, Device mays work?
Rather than Master, maybe Main will work?
Also if you remove the RC from the first enum, you may remove from the second one

HumJ0218: About RC, datasheet is sic, because as the primary device there can be both internal oscillator (RC) and external clock (external) modes. Secondary devices can be connected after the RC primary or in the same way after the external clock primary

HumJ0218: It is independent of whether the primary device is in RC mode

src/devices/Ht1632/samples/Ht1632.Sample.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <RootNamespace>Iot.Device.Ht1632.Sample</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="SixLabors.ImageSharp" Version="1.0.4" />
```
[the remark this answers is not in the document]
krwq: you can send second PR to bumpt that to 1.0.4

src/devices/Ht1632/samples/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+using var ht1632 = new Ht1632(new Ht1632PinMapping(cs: 27, wr: 22, data: 17), new GpioController())
+{
+    ComOption = ComOption.NMos16Com,
+    ClockMode = ClockMode.RcPrimary,
+    Enabled = true,
+    PwmDuty = 1,
+    Blink = false,
+    LedOn = true
+};
+{
```
krwq, line 21: I'd suggest to convert these into methods and just call them

src/devices/Ht1632/samples/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+    Console.WriteLine("Show image");
+
+    var image = Image.Load<Rgb24>("./dotnet-bot.bmp");
+    ht1632.ShowImageWith16Com(image);
+
+    Console.ReadLine();
+}
+
+Console.WriteLine("All end");
```
krwq, line 55: :-) please remove

src/devices/Ht1632/samples/Program.cs
krwq: (all of the code with explicit scope)

joperezr:
@HumJ0218 Looks like there are some markdown lint errors in your changes. Can you please take a look?

```
D:/a/1/s/src/devices/Ht1632/samples/README.md:1 MD041/first-line-heading/first-line-h1 First line in a file should be a top-level heading [Context: "## Schematic"]
```

joperezr (approved):

joperezr:
Thanks a lot for the contribution @HumJ0218!

Issue #1719, opened 2021-11-10
The Button implementation does not provide debouncing


raffaeler:
## Feature request
The current https://github.com/dotnet/iot/tree/main/src/devices/Button implementation does not provide any debouncing mechanism.

There is another issue (https://github.com/dotnet/iot/issues/674) requesting debouncing at a lower level. Apparently the Linux driver does not provide any debouncing mechanism.

It would probably better to implement debouncing at least at button level, given that it is very unpractical to use a button without debouncing.

pgrawehr:
I agree, on the button binding this makes sense. I wouldn't put it on the general drivers, because it would add domain-specific knowledge where it doesn't belong and/or suggest hardware support for something that will, in many cases, be simulated in software.

raffaeler:
I first tried to see if it was available in the native driver but I could not find anything like that. Maybe @joperezr can either confirm or not about my research.
In any case, I already submitted a PR for the button changes so that it is immediately available.

PR #1720, opened 2021-11-10
Debouncing implementation and test

raffaeler:
This PR adds support for debouncing to the Button device.
Issue: https://github.com/dotnet/iot/issues/1719
Tested on a Raspberry PI 3B+

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1720)

### DIFF
--- a/src/devices/Button/ButtonBase.cs
+++ b/src/devices/Button/ButtonBase.cs
@@ -20,7 +20,7 @@ namespace Iot.Device.Button
         private long _doublePressTicks;
         private long _holdingMs;
         private TimeSpan _debounceTime;
-        private long _singlePressTicks;
+        private long _debounceStartTicks;
 
         private ButtonHoldingState _holdingState = ButtonHoldingState.Completed;
 
@@ -83,6 +83,11 @@ namespace Iot.Device.Button
         /// <param name="debounceTime">The amount of time during which the transitions are ignored, or zero</param>
         public ButtonBase(TimeSpan doublePress, TimeSpan holding, TimeSpan debounceTime = default(TimeSpan))
         {
+            if (debounceTime.TotalMilliseconds * 3 > doublePress.TotalMilliseconds)
+            {
+                throw new ArgumentException($"The parameter {nameof(doublePress)} should be at least three times {nameof(debounceTime)}");
+            }
+
             _doublePressTicks = doublePress.Ticks;
             _holdingMs = (long)holding.TotalMilliseconds;
             _debounceTime = debounceTime;
@@ -93,7 +98,7 @@ namespace Iot.Device.Button
         /// </summary>
         protected void HandleButtonPressed()
         {
-            if (DateTime.UtcNow.Ticks - _singlePressTicks < _debounceTime.Ticks)
+            if (DateTime.UtcNow.Ticks - _debounceStartTicks < _debounceTime.Ticks)
             {
                 return;
             }
@@ -118,7 +123,7 @@ namespace Iot.Device.Button
                 return;
             }
 
-            _singlePressTicks = DateTime.UtcNow.Ticks;
+            _debounceStartTicks = DateTime.UtcNow.Ticks;
             _holdingTimer?.Dispose();
             _holdingTimer = null;
 
--- a/src/devices/Button/ButtonBase.cs
+++ b/src/devices/Button/ButtonBase.cs
@@ -71,7 +71,7 @@ namespace Iot.Device.Button
         /// Initialization of the button.
         /// </summary>
         public ButtonBase()
-            : this(TimeSpan.FromTicks(DefaultDoublePressTicks), TimeSpan.FromMilliseconds(DefaultHoldingMilliseconds), TimeSpan.Zero)
+            : this(TimeSpan.FromTicks(DefaultDoublePressTicks), TimeSpan.FromMilliseconds(DefaultHoldingMilliseconds))
         {
         }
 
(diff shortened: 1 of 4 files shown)


src/devices/Button/ButtonBase.cs
```diff
@@ -78,17 +80,24 @@ public ButtonBase()
         /// </summary>
         /// <param name="doublePress">Max ticks between button presses to count as doublepress.</param>
         /// <param name="holding">Min ms a button is pressed to count as holding.</param>
-        public ButtonBase(TimeSpan doublePress, TimeSpan holding)
+        /// <param name="debounceTime">The amount of time during which the transitions are ignored, or zero</param>
+        public ButtonBase(TimeSpan doublePress, TimeSpan holding, TimeSpan debounceTime)
```
krwq, line 84: new argument should be optional

raffaeler: Right, I'll fix it now. Anyway the button is not yet in a released nuget package (unless it has just been released this night :-D)

src/devices/Button/ButtonBase.cs
```diff
@@ -19,6 +19,8 @@ public class ButtonBase : IDisposable
 
         private long _doublePressTicks;
         private long _holdingMs;
+        private TimeSpan _debounceTime;
+        private long _singlePressTicks;
```
krwq, line 23: this is misnamed - currently `_doublePressTicks` means amount of time between presses while `_singlePressTicks` means last time we pressed the button. I'd recommend renaming it to something like `_lastPressedTicks` or `_debounceStart` or something along the lines - as is makes reading the code a bit confusing

src/devices/Button/ButtonBase.cs
```diff
@@ -78,17 +80,24 @@ public ButtonBase()
...
             _holdingMs = (long)holding.TotalMilliseconds;
+            _debounceTime = debounceTime;
         }
 
         /// <summary>
         /// Handler for pressing the button.
         /// </summary>
         protected void HandleButtonPressed()
         {
+            if (DateTime.UtcNow.Ticks - _singlePressTicks < _debounceTime.Ticks)
```
krwq, line 96: - Store UtcNow in the beginning of the method and reuse it later
- when using UtcNow anywhere please make it as a sole element on one side of equation (i.e. `UtcNow.Ticks < ...`) - that will make it easier to read

src/devices/Button/ButtonBase.cs
```diff
@@ -78,17 +80,24 @@ public ButtonBase()
         /// </summary>
         /// <param name="doublePress">Max ticks between button presses to count as doublepress.</param>
         /// <param name="holding">Min ms a button is pressed to count as holding.</param>
-        public ButtonBase(TimeSpan doublePress, TimeSpan holding)
+        /// <param name="debounceTime">The amount of time during which the transitions are ignored, or zero</param>
+        public ButtonBase(TimeSpan doublePress, TimeSpan holding, TimeSpan debounceTime = default(TimeSpan))
```
krwq, line 84: if `debounceTime * 2 >= doublePress` then you will never get doublePress event, you should throw if that's the case. I'd recommend to even make it `3*` since timing clicks perfectly would be literally impossible

src/devices/Button/ButtonBase.cs
raffaeler: I'll go with `_debounceStartTicks`

src/devices/Button/ButtonBase.cs
```diff
@@ -104,6 +113,12 @@ protected void HandleButtonPressed()
         /// </summary>
         protected void HandleButtonReleased()
         {
+            if (_debounceTime.Ticks > 0 && !IsPressed)
```
krwq, line 116: can you explain `!IsPressed`? Is that correct? While ignoring you should ignore both pressed and released state

src/devices/Button/ButtonBase.cs
raffaeler: In the pressed handler, there is only once. This is why I didn't store it. Do you really want it locally?
In the released handler, I need it before raising the event while the previous code (double press) comes after the event and some significative time may have passed. It would be risky to reuse the same.

src/devices/Button/ButtonBase.cs
raffaeler: Yes because in the Pressed handler, I ignore the events coming from the driver when debouncing is occurring. This means you also get multiple Released handlers called and I need to "pair" them, otherwise the Pressed event gets called.

raffaeler:
It looks like the failures on the tests do not depend on my code. Could you please confirm @krwq ?

joperezr (approved):

W1zzardTPU:
There's an issue with this implementation, when the button is held down for longer then the debounce timeout. When it is released now, there will be a "pressed" event due to the bounces happening during release, and the desired "released" event is fired, due to the debouncing getting started by "pressed".

I solved this by using a timer for debounce that checks the current button state when finished. Something like below, same for HandleButtonReleased()

```
protected void HandleButtonPressed()
{
    _stateForAfterDebounce = true;  // keep track of the button state, even with bounces, so we can get the current state at the end of the debounce timeout

    if (IsPressed)  // already pressed
        return;
    if (_debounceTimer!=null)   // currently debouncing
        return;

    IsPressed = true;

    if (_debounceTime.TotalMilliseconds > 0)
    {
        _debounceTimer = new Timer((o) =>
        {
            _debounceTimer?.Dispose();
            _debounceTimer = null;

            if (_stateForAfterDebounce == false)  // Button might have been been released while we were in the debounce. Check current state and handle
                HandleButtonReleased();
        }, null, _debounceTime, Timeout.InfiniteTimeSpan);
    }

    ButtonDown?.Invoke(this, new EventArgs());

    if (IsHoldingEnabled)
    {
        _holdingTimer = new Timer(StartHoldingHandler, null, (int)_holdingMs, Timeout.Infinite);
    }
}
```

pgrawehr:
@W1zzardTPU since this PR was merged and closed long ago, could you please open a new issue instead? Otherwise it's likely that this will get lost.

Issue #1430, opened 2021-01-31
Error message functionality for Iot.Device.SocketCan.CanRaw


GKooijmans:
The Linux SocketCan module supports Error Message Frames. These frames offer a mechanism for userspace applications to be notified when the state of the CAN Bus changes. By default these are filtered out. They can be enabled by a setsockopt call.
The current implementation of the Raw CAN socket in Iot.Device.SocketCan.CanRaw contains a method for setting a regular Message Filter of type CAN_RAW_FILTER, but not one for setting an Error Message filter (CAN_RAW_ERR_FILTER).
If it were to be added, it would give an application a lot more control over the underlying interface.

krwq:
@GKooijmans what APIs would you suggest?

GKooijmans:
@krwq Thanks for replying. Perhaps something like:

void FilterErrorMsgs(CanErrors errorfilter);


 public enum CanErrors : uint
    {
        CAN_ERR_TX_TIMEOUT = 0x00000001U,
        CAN_ERR_LOSTARB = 0x00000002U,
        (complete defs at https://github.com/torvalds/linux/blob/master/include/uapi/linux/can/error.h)
     }

GKooijmans:
Also the CanRaw.TryReadFrame method may need some changes. Because an error frame will have CanFrame.IsValid set to false 
And that case the method does not return any data. However an Error Frame does hold data with information regarding the Error that occurred. You could also check for CanId.Error which will be set to true in these frames.

krwq:
@GKooijmans let's wait until https://github.com/dotnet/iot/pull/1277 is merged before we do any work here.

What about removing https://github.com/dotnet/iot/blob/main/src/devices/SocketCan/CanRaw.cs#L92-L98 and returning `frame.IsValid`. We would still need to somehow make sure the [assert below](https://github.com/dotnet/iot/blob/main/src/devices/SocketCan/CanRaw.cs#L101) is not hit.

Are the error frames currently completely filtered out? (don't have a setup closeby and can't remember from the top of my haead) I'd personally simplify this API a bit:

```csharp
public bool SkipErrorFrames { get; set; } // getter and setter would do a bit more, this is just an API shape
```

do we actually care about type of error? If yes then perhaps:

```csharp
public SocketCanErrorFiltering ErrorFiltering { get; set; }

public enum SocketCanErrorFiltering
{
  NoErrorFrames,
  AllErrorFrames,
  // the other two options with user friendly name
}
```

[...]

PR #1725, opened 2021-11-12
Socket can

schauerte:
This change needs re-targeting of the project because it relies on [AddressFamily.ControllerAreaNetwork](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.addressfamily?view=net-5.0) which has been added in .NET Core 3.0

Currently, reading CAN frames in a loop will block the thread until some frame arrives, thus making it hard to gracefully exit the program.

This change enables (opt-in) non-blocking read (and write) of CAN frames by providing `async` methods for socket operations.

It is based on the preparations made for https://github.com/dotnet/iot/pull/1277.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1725)


src/devices/SocketCan/SocketCan.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(DefaultBindingTfms)</TargetFrameworks>
+    <TargetFrameworks>net5.0;netcoreapp3.1</TargetFrameworks>
```
pgrawehr, line 3: Do you really need this change (this removes netstandard support for this binding)? If so, you have to change the dependent tests and/or samples as well.

schauerte: Yes. As discussed in #1277, this would be required.

At first I wanted to get the implementation "right" (for my requirements). I will take a look at the changes proposed by @krwq (see [comment](https://github.com/dotnet/iot/pull/1277#issuecomment-731089781)) next.

But since I had a hard time finding any .NET (Core/5) library for SocketCan and the project here does not use any of the GPIO parts (but is part of [Iot.Device.Bindings](https://www.nuget.org/packages/Iot.Device.Bindings/) "This package provides a set of Device Bindings that use System.Device.Gpio package to communicate with a microcontroller."), I'm not totally convinced that this code should stay here. Does a better place exist?

schauerte:
Can anyone help me to fix the failing checks? I currently don't know what is failing (and why)?

schauerte:
I am closing this pull-request, because I feel, this is not the right place. 
We are going to add more features like handling of error frames (see #1430) and support for CAN-FD and this will probably introduce a lot of breaking changes.

PR #1731, opened 2021-11-18
Make markdown lint only run on PR builds

joperezr:
Seems like our internal build machines don't have npm so trying to run the linter there is failing. Since we don't really care about running it in official builds as we only want the protection to be on PRs, this PR is disabling the step for official builds.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1731)

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -101,10 +101,11 @@ stages:
         - checkout: self
           clean: true
         # Scan markdown files on style consistency
-        - powershell: |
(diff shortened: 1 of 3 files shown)


joperezr:
@Ellerbach Merging in order to unblock clean official build. Will address any feedback in a follow up PR.

PR #1733, opened 2021-11-23
Adding new pipeline and switching authenticode and strong name signing on the repo.

joperezr:
Ever since we created dotnet/iot repository, our amazing community of contributors have helped us make possible the two libraries that we have today. We have also been integrating with the community more and more, by doing things like having open weekly triage with our top contributors, and also by granting the "triage" privilege to these top contributors. We are now ready to start working towards taking that next step and start granting "committer" privileges to our top contributors. We are in the process of writing up an announcement with more details, and will share a link here as well as pin an issue on the repo once that is ready. (cc. @richlander)

In order to do this, some changes need to go on the repo first (all of which are included in this PR):
 - Change public key token (strongname) of the assembly: Currently, we are using Microsoft internal public key token in order to strong name sign the assembly. Given the plan is to start having non-Microsoft committers, we need to change that. As part of these changes, I'm committing the new .snk which will be used going forward. This is a breaking change unfortunately, but that is why we want to take it now that we are about to ship version 2.0 in order to be future proof.
 - Change the authenticode signature of the NuGet package and Assembly: Similar to strongname, since we will now have non-Microsoft comitters we have to use a new certificate as the digital signature. We have worked with the .NET Foundation to get this new cert and will be now be using it to sign both the packages and the binaries we build from now on.
 - Move to a new AzDO tenant: We are currently building the product on the dnceng Azure devops tenant, which is Microsoft-owned. Since we no longer want to require Microsoft infrastructure for our build, we are moving into a new Azure DevOps tenant that is provided by the .NET Foundation and is specific only to us. This of course means that we also need new Azure pipelines for PR builds, CI, and Official builds. Those pipelines and their configuration are also included as part of this PR. This PR is not deleting the old pipeline just yet, as that will have to be decommissioned, disabled and deleted as a subsequent PR.
 - Use snupkgs that are uploaded to NuGet: We used to publish all of our symbols to internal Microsoft symbol servers as part of the build, but this again is something we don't want to depend on any longer as we don't want to depend on Microsoft infrastructure. We are now moving to building .snupkgs as part of the build, and these will be published to NuGet just as the regular packages, which will then later be able to be used from Visual Studio, Visual Studio Code, or other IDEs for source linking and debugging.

This PR is not yet updating all of the documentation just yet as there are some aspects that are not finalized (for example, where will we publish nightly builds and how to consume them) but that will come in a subsequent PR. These changes have been blocking the release of 2.0 package, but once this is merged, we will be ready to go, so thank you so much for all of the people that helped making this happen 😄 

cc: @richlander @danmoseley @ericstj @pgrawehr @raffaeler @krwq @Ellerbach @mmitche 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1733)


Directory.Build.props
```diff
@@ -19,8 +19,20 @@
     <LangVersion>9</LangVersion>
     <!-- Generate snupkg package -->
     <IncludeSymbols>true</IncludeSymbols>
+    <StrongNameKeyId>DotnetIot</StrongNameKeyId>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
+  <Choose>
+    <When Condition="'$(StrongNameKeyId)' == 'DotnetIot'">
```
safern, line 27: Why do we need to condition this? Aren't these settings always used?

azure-pipelines.yml
```diff
@@ -0,0 +1,157 @@
+trigger:
+  batch: true
+  branches:
+    include:
+    - main
+    - release/3.0
```
safern, line 6: What is the plan for branching going forward? Should we include a wildcard so that infra is ready for next releases' branching?

Directory.Build.props
joperezr: we are reusing the properties that arcade uses in their strongname property and targets files, so in theory that key could still represent a different value which would be used at that point. This is mostly what I'm referring to: https://github.com/dotnet/arcade/blob/927f8d4d5036f68a5fc6d042f336bc9458027208/src/Microsoft.DotNet.Arcade.Sdk/tools/StrongName.targets#L24-L76

We could change it and not condition it and turn this into a traditional property group, but doing it like this it just makes it clear why the property StrongNameKeyId MUST be set, since if you don't set it and just set the publickeytoken and publickey, they will get overwritten later.

src/Iot.Device.Bindings/CompatibilitySuppressions.xml
```diff
@@ -733,4 +733,11 @@
     <DiagnosticId>CP1002</DiagnosticId>
     <Target>Windows.Foundation.UniversalApiContract.dll</Target>
   </Suppression>
-</Suppressions>
\ No newline at end of file
+  <!--We are changing public key token being used, so the following are baselined:-->
```
safern, line 736: Nice that the repo has onboarded package validation: 🎉

Directory.Build.props
safern: A ok, makes sense. Maybe worth a comment?

azure-pipelines.yml
joperezr: current release is being shipped from the main branch. Release cadance of the repo makes it so that we have stable releases every month or every other month, depending on the incoming changes. Due to this, having separate branches becomes more of a burden than helping stabilize, so we have currently opted to just continue shipping out of main for now, and for larger things having feature branches until code is mature enough to be merged on main. All that said, I'm fine with deciding to add a wildcard here, I currently just left this as it is on our other pipeline.

joperezr:
FWIW, this is the new pipeline which is already running as part of this PR:

image

safern (approved):

src/Iot.Device.Bindings/Directory.Build.props
```diff
@@ -1,8 +1,8 @@
 <Project>
   <!-- Packaging related properties -->
   <PropertyGroup>
-    <MajorVersion>1</MajorVersion>
-    <MinorVersion>6</MinorVersion>
+    <MajorVersion>2</MajorVersion>
```
safern, line 4: Should `PackageValidationBaselineVersion` use `1.6.0`?

joperezr: Good point, 1.6.0 never shipped but we should probably be using 1.5.0. In any case, I don't want to make those changes as part of this PR but I can send a subsequent for that part. FWIW we did run package validation against 1.5.0 baseline manually and found only expected changes in there, so I'm not too worried. That said, we should definitely change this  in case there is anything that needs to be fixed before 2.0 RTMs

PR #1738, opened 2021-11-29
Cross targeting changes for the upcoming 2.0 release.

joperezr:
This PR will make the retargeting changes we want for our upcoming 2.0 release. As part of these changes, we are:

1. Removing netcoreapp2.1 cross targeting for iot.device.bindings, since that framework is no longer in support. Instead, we are now cross targeting to netcoreapp3.1 on places we used to do so.
2. Switch places where we cross-targeted for net5.0 and use net6.0 instead. This is because net5.0 will soon go out of support as well in favor of net6.0, so we want to be targeting supported versions going forward. This also includes updating the TFM on most of our samples, since they all used to target net5.0.

I'm separately running some validation to ensure we don't have any unexpected breaks, but please let me know if anybody has a concern with any of the above mentioned changes.

cc: @richlander @krwq @Ellerbach @raffaeler @pgrawehr @ericstj 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1738)


samples/force-sensitive-resistor/force-sensitive-resistor.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
pgrawehr, line 5: Suggestion: Create a variable for the default target framework for the samples. Search and replace on 100+ files each time we want to upgrade is a bit ugly.

src/devices/Directory.Build.props
```diff
@@ -5,7 +5,7 @@
   <Import Project="../../eng/Versions.external.props" />
   <PropertyGroup>
     <!-- Most projects should use this value to build for all possible target frameworks -->
-    <DefaultBindingTfms>net5.0;netcoreapp3.1;netstandard2.0;</DefaultBindingTfms>
+    <DefaultBindingTfms>net6.0;netcoreapp3.1;netstandard2.0;</DefaultBindingTfms>
```
pgrawehr, line 8: This will now cause that a .NET 5.0 app will fall back to the netstandard2.0 lib. Not sure we want that.

samples/force-sensitive-resistor/force-sensitive-resistor.csproj
joperezr: Good point, will do.

src/devices/Directory.Build.props
joperezr: Not really, they will fallback to the netcoreapp3.1 asset, and I checked with package validation and there is no loss on API  when they do that as all of the bindings that we have for net5.0 also work with netcoreapp3.1.

PR #1752, opened 2021-12-21
Add Raspberry Pi build HAT

Ellerbach:
This PR is about adding [Raspberry Pi Build HAT](https://www.raspberrypi.com/documentation/accessories/build-hat.html#introducing-the-build-hat):

- Full implementation of the Build HAT functions
- Passive Motor full implementation
- Active Motors full implementation
- All active supported sensors implementation
- Passive sensors full implementation for lights, buttons
- Detailed documentation
- Detailed samples
- Tested on Windows (though USB to Serial dongle) and Linux

Couple of elements for the reviewers:

- The principle from BrickPi3 and some level of compatibility (function names, properties names, events on properties, IMotor interface) has been kept.
- The sensors, especially the distance ones returns limited number of values from 0 to 10 which are supposed to be centimeters, some are in millimeters. I decided, **not** to use UnitsNet for them. BrickPi3 is not using it.
- The motors angles are measures in degrees, same, for compatibility reasons and to keep it simple, I decided **not** to use UnitsNet. All the Lego world is just using the degrees, in all apps, all samples.
- Most sensors has been tested. Some like WeDo has not yet ben tested. As soon as I4ll be able to get some, I'll test them and will adjust if anything needs to be adjusted.

> **Important**: The Build HAT comes **without** a firmware, when powered up, or repowered up, it's only in bootloader mode. Firmware must be uploaded first. So the firmware, available [here](https://github.com/RaspberryPiFoundation/python-build-hat/tree/d2d6ff3f43de8a92ef8c900d1b278f1298a29c0a/buildhat/data) (site under MIT license) has been included in the code.

Enjoy the rest of the year and the review :-)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1752)


src/devices/BuildHat/BuildHat.csproj
```diff
@@ -0,0 +1,49 @@
...
+     <Compile Include="Sensors\UltrasonicDistanceSensor.cs" />
+     <Compile Include="Sensors\ForceSensor.cs" />
+     <Compile Include="Sensors\Sensor.cs" />
+     <Compile Include="Sensors\PassiveLight.cs" />
+     <Compile Include="Sensors\WeDoDistanceSensor.cs" />
+     <Compile Include="Sensors\WeDoTiltSensor.cs" />
+   </ItemGroup>
+
+   <ItemGroup>
+     <None Include="data\firmware.bin">
```
pgrawehr, line 38: I would embed these as resources, otherwise we have to do special tricks to get them deployed correctly during package consumption.

src/devices/BuildHat/Sensors/Sensor.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    /// <summary>
+    /// Interface for a sensor
+    /// </summary>
+    public class Sensor
+    {
+        internal Brick Brick;
+
+        /// <summary>
+        /// Gets true if the motor is connected.
```
pgrawehr, line 17: Description is missleading. Should probably be "if the sensor is connected"

src/devices/BuildHat/Sensors/Sensor.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// To notify a property has changed.
+        /// </summary>
+        public event PropertyChangedEventHandler? PropertyChanged;
+
+        /// <summary>
+        /// To notify a property has changed.
+        /// </summary>
+        public event PropertyChangedEventHandler? PropertyUpdated;
+
+        internal void OnPropertyChanged(string name) => PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(name));
```
pgrawehr, line 61: Add `[CallerMemberName]` attribute.

src/devices/BuildHat/BuildHat.csproj
Ellerbach: yes, I can do that. That's a good catch!

src/devices/BuildHat/Brick.cs
```diff
@@ -0,0 +1,1580 @@
...
+    public class Brick : IDisposable
+    {
+        private const string FirmwareVersion = "Firmware version: ";
+        private const string FirmwareSignature = "Firmware signature: ";
+        private const string GetFirmwareVersion = "version\r";
+        private const string ButtonReleased = ": button released";
+        private const string ButtonPressed = ": button pressed";
+        private const string PowerFaultError = "power fault";
+
+        private SerialPort _port;
```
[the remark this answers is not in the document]
Ellerbach: I was hesitating. If I operate on streams, I'll add the possibility to add as well a constructor to pass a stream.

src/devices/BuildHat/Sensors/Sensor.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <summary>
+        /// To notify a property has changed.
+        /// </summary>
+        public event PropertyChangedEventHandler? PropertyChanged;
+
+        /// <summary>
+        /// To notify a property has changed.
+        /// </summary>
+        public event PropertyChangedEventHandler? PropertyUpdated;
```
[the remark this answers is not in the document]
Ellerbach: Changed means the value has changed. Updated means the value has been updated but it still can be the same.

src/devices/BuildHat/Sensors/Sensor.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// </summary>
+        public bool IsConnected { get; internal set; }
+
+        /// <summary>
+        /// Creates a sensor
+        /// </summary>
+        /// <param name="brick">The brick.</param>
+        /// <param name="port">The port.</param>
+        /// <param name="type">The sensor type.</param>
+        internal Sensor(Brick brick, SensorPort port, SensorType type)
```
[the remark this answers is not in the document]
Ellerbach: what would that bring if it's protected compare to the current implementation?
All the other sensors, including the motors are heritating from it. All the others should be protected as well? I'm happy to adjust if needed of course!

src/devices/BuildHat/Sensors/Sensor.cs
Ellerbach: What will that bring in the current implementation? I may have missed something.

src/devices/BuildHat/Brick.cs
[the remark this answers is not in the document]
Ellerbach: now, I know why I haven't been using streams: ReadlLine()... Almost everything is based of it. And all is text (except firmware update). Using streams would add complexity in this case as I'll have to manually implement all this. I may have missed something here again but seems like the BaseStream is just a Stream, so not text primitives at all.

src/devices/BuildHat/Sensors/Sensor.cs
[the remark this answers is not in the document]
pgrawehr: If you make it protected, it would be possible to externally add additional sensors. Like this, it is not possible to externally inherit from this class. 

Also, if you don't need instances of this class itself, it should probably be abstract.

src/devices/BuildHat/Sensors/Sensor.cs
pgrawehr: If you declare your member as 

```
internal void OnPropertyChanged([CallerMemberName]string name=null)
```

You can call `OnPropertyChanged()` within a property without argument. The compiler will automatically insert the current property name at the call location. It's just a shortcut.

src/devices/BuildHat/Sensors/Sensor.cs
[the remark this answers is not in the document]
Ellerbach: ok, good idea then to have it protected. Someone can build its own version from this.

src/devices/BuildHat/Sensors/Sensor.cs
Ellerbach: I think I'll skip this. reason is: it's not supported in nanoFramework and I want to port it to nano. So if I adjust the shortcut here, I'll anyway will have to go for the long way.

src/devices/BuildHat/Brick.cs
```diff
@@ -0,0 +1,1564 @@
...
+            SelectCombiModesAndRead(port, new int[] { 1, 2, 3 }, false);
+            // Ramp uses first param as initial position, second as target, third is how long, foruth is always 0
+            PortWrite($"port {(byte)port} ; pid {(byte)port} 0 1 s4 0.0027777778 0 5 0 .1 3 ; set ramp {actualPositionDouble.ToString(CultureInfo.InvariantCulture)} {newPosition.ToString(CultureInfo.InvariantCulture)} {duration.ToString(CultureInfo.InvariantCulture)} 0\r");
+            if (blocking)
+            {
+                while (!((motor.Position / 360.0 < newPosition + 2.028) && (motor.Position / 360.0 > newPosition - 2.028)))
+                {
+                    Thread.Sleep(5);
+                }
+            }
```
pgrawehr, line 420: Suggestion: Make these blocking functions async or return a wait handle.

src/devices/BuildHat/Brick.cs
```diff
@@ -0,0 +1,1564 @@
...
+                throw new ArgumentException("Speed can't be 0");
+            }
+
+            speed = speed < -100 ? -100 : speed > 100 ? 100 : speed;
+            ActiveMotor motor = (ActiveMotor)_elements[(byte)port];
+            // We need both the relative and absolute position
+            int actualPosition = motor.Position;
+            int actualAbsolutePosition = motor.AbsolutePosition;
+            double actualPositionDouble = actualPosition / 360.0;
+            int difference = (targetPosition - actualAbsolutePosition + 180) % 360 - 180;
```
pgrawehr, line 326: You should write some unit tests for this function. I'm not sure the values are always in range after the calculations.

src/devices/BuildHat/Brick.cs
```diff
@@ -0,0 +1,1564 @@
...
+        /// Wait to get a sensor connected on a specific port.
+        /// </summary>
+        /// <param name="port">The port.</param>
+        /// <param name="token">A cancellatin token</param>
+        public void WaitForSensorToConnect(SensorPort port, CancellationToken token = default)
+        {
+            // Test sensor, plug a RGB distance one on port C
+            while (GetSensor(port) == null)
+            {
+                token.WaitHandle.WaitOne(50);
```
pgrawehr, line 672: That will never terminate if no sensor is plugged in. Test the cancellation token in the loop.

src/devices/BuildHat/Brick.cs
```diff
@@ -0,0 +1,1564 @@
...
+            }
+
+            if (!token.IsCancellationRequested)
+            {
+                while (!((Sensor)GetSensor(port)).IsConnected)
+                {
+                    token.WaitHandle.WaitOne(50);
```
pgrawehr, line 679: Same here, move the test inside the loop (or within the while condition)

src/devices/BuildHat/Brick.cs
```diff
@@ -0,0 +1,1564 @@
...
+                                        // nothing to report
+                                        default:
+                                            break;
+                                    }
+                                }
+
+                            }
+                            catch (Exception)
+                            {
+                                // Just skip, it can be malformed
```
pgrawehr, line 1192: Maybe add logging?

src/devices/BuildHat/Brick.cs
```diff
@@ -0,0 +1,1564 @@
...
+                line = line.Substring(line.LastIndexOf(':') + 2);
+                data = line.Split(' ');
+                RecommendedPid recoPos = new();
+                recoPos.Pid1 = Convert.ToInt32(data[0], 16);
+                recoPos.Pid2 = Convert.ToInt32(data[1], 16);
+                recoPos.Pid3 = Convert.ToInt32(data[2], 16);
+                recoPos.Pid4 = Convert.ToInt32(data[3], 16);
+                sensor.PositionPid = recoPos;
+            }
+            catch (Exception)
```
pgrawehr, line 1417: Same here. If you have a catchall, you should at least do some logging.

src/devices/BuildHat/Models/PositionWay.cs
```diff
@@ -0,0 +1,20 @@
...
+    public enum PositionWay
+    {
+        /// <summary>Shortest way</summary>
+        Shortest,
+
+        /// <summary>Clockwise way</summary>
+        Clockwise,
+
+        /// <summary>Anto clockwise way</summary>
+        AntoClockwise,
```
pgrawehr, line 18: This should probably be named "AntiClockwise"?

pgrawehr (commented):
Some more findings.

src/devices/BuildHat/Brick.cs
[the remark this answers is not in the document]
pgrawehr: Yes, you would (and should) add a constructor to operate on streams, that's the reason for my suggestion. 

You can wrap a StreamReader/StreamWriter around the Stream, to get the ReadLine()/WriteLine() functions back. I agree, writing that yourself would be ugly.

src/devices/BuildHat/Sensors/Sensor.cs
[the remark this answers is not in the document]
pgrawehr: Good. Then please review the other members for external usage as well, please (e.g. you probably want the events to be public or protected, too)

src/devices/BuildHat/Sensors/Sensor.cs
pgrawehr: Oh, Ok then.

src/devices/BuildHat/Brick.cs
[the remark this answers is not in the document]
Ellerbach: I've been trying with the streams and it's not really working. This may be due to the drivers of my usb to ttl but it's really hugly. I get time outs everywhere, the streams do not properly give the number of available bytes or even if there is the EndOfStream property neither.
This hat can only be used with a serial port, so what is the real core interest to have it with streams? Especially as it does drastically complicate this implementation and requires then quite some try/catch.

src/devices/BuildHat/Brick.cs
Ellerbach: I was thinking of this and made the choice to have it without async. There are 2 reasons, first, the target audience for this hat are beginners and second it's for the port then to nanoFramework. I can add an async one if you think it's needed

src/devices/BuildHat/Brick.cs
Ellerbach: yes, good catch

src/devices/BuildHat/Brick.cs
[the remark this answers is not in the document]
pgrawehr: That's a bit weird, indeed. 

The reason why I prefer Stream is that it is more flexible. It can be forwarded trough another link, written to disk (and replayed) or mocked for unit testing. But if you manage to write unit tests against the current implementation as well, this won't be a blocker for me.

src/devices/BuildHat/Brick.cs
pgrawehr: That makes sense. Fine with me. 

But another thing I just see: You should not read out `motor.Position` twice in the condition, as this could result in undefined behavior, because the value might change between the two calls. Additionally, it seems that `2.028` should be a constant.

src/devices/BuildHat/Brick.cs
pgrawehr: That's not what I meant, actually. We have a logging framework in place. 

In the constructor, do (you might need to add a reference to Iot.Device.Common):
```
_logger = this.GetCurrentClassLogger()
```

Then you can do (without any #if's): 
```
_logger.LogError(ex, $"There was an error parsing the command reply: {ex.Message}");
```

Ellerbach: Fair, I can add the logging. Will do!

src/devices/BuildHat/samples/Program.cs
```diff
@@ -0,0 +1,340 @@
...
+using Brick brick = new("COM3");
+// On a Raspberry PI, you'll use:
+// using Brick brick = new("/dev/serial0");
+var info = brick.BuildHatInformation;
+Console.WriteLine($"version: {info.Version}, firmware date: {info.FirmwareDate}, signature:");
+Console.WriteLine($"{BitConverter.ToString(info.Signature)}");
+Console.WriteLine($"Vin = {brick.InputVoltage.Volts} V");
+Console.WriteLine("Select what you want to test:");
+Console.WriteLine(" 1. Display elements detials");
+Console.WriteLine(" 2. Display connection/disconnecions");
```
danmoseley, line 25:
```suggestion
Console.WriteLine(" 1. Display element details");
Console.WriteLine(" 2. Display connection/disconnection");
```
?

src/devices/BuildHat/Brick.cs
Ellerbach: And it was useful to ad. Found out one mistake. Thanks!

src/devices/BuildHat/Brick.cs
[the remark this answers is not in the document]
Ellerbach: For the unit test, I added an internal constructor and used reflection to call internal methods. There is not much to test fully anyway. Mainly one function which makes the math on how to get from one potision to another

pgrawehr (approved):

PR #1755, opened 2021-12-23
New transport layer for HardwareAccess

pgrawehr:
The existing WMI transport showed to be unreliable in practical environments. For unknown reasons, the sensors where often just not there. I have added a new HTTP-based API to OpenHardwareMonitor, which should be much more reliable. 

This was anyway necessary during upgrade to .NET 5.0, because the WMI registration components are not available any more.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1755)


src/devices/Common/Iot/Device/Common/JsonParser.cs
```diff
@@ -0,0 +1,507 @@
...
+    //      "{\"Value\":10}".FromJson<object>().GetType() == typeof(Dictionary<string,object>)
+    // - No JIT Emit support to support AOT compilation on iOS
+    // - Attempts are made to NOT throw an exception if the JSON is corrupted or invalid: returns null instead.
+    // - Only public fields and property setters on classes/structs will be written to
+    //
+    // Limitations:
+    // - No JIT Emit support to parse structures quickly
+    // - Limited to parsing <2GB JSON files (due to int.MaxValue)
+    // - Parsing of abstract classes or interfaces is NOT supported and will throw an exception.
+    public class JsonParser
```
krwq, line 33: could this use System.Text.Json?

src/devices/HardwareMonitor/JsonDefs.cs
```diff
@@ -0,0 +1,139 @@
...
+        public HardwareJson()
+        {
+            NodeId = String.Empty;
+            Name = String.Empty;
+            Sensors = Array.Empty<SensorJson>();
+            Parent = string.Empty;
+            HardwareType = string.Empty;
+        }
+
+        public string NodeId
```
Ellerbach, line 52: to simplify, you can have all in 1 line ;-)

src/devices/Common/Iot/Device/Common/JsonParser.cs
Ellerbach, line 33: you may give a try to https://www.nuget.org/packages/System.Text.Json, that should work and work with .NET Standard too

src/devices/HardwareMonitor/OpenHardwareMonitorHttp.cs
```diff
@@ -0,0 +1,280 @@
...
+        public void Dispose()
+        {
+            _version = string.Empty;
+            _httpClient.Dispose();
+        }
+
+        /// <summary>
+        /// Returns a non-rooted path combined from the input paths that uses forward slashes (for web apis)
+        /// </summary>
+        private string UnixPathJoin(string a, string b)
```
raffaeler, line 197: Take a look at Uri methods to create the uri. Maybe TryCreate method.

src/devices/HardwareMonitor/OpenHardwareMonitorHttp.cs
```diff
@@ -0,0 +1,280 @@
...
+            ret = ret.Replace("//", "/");
+            if (ret.StartsWith("/"))
+            {
+                ret = ret.Substring(1);
+            }
+
+            return ret;
+        }
+
+        private bool UpdateSensor(SensorHttp sensor, out double value)
```
raffaeler, line 226: What about calling it TryUpdateSensor?

src/devices/HardwareMonitor/OpenHardwareMonitorHttp.cs
pgrawehr: Doesn't work out-of-the-box, but with some trickery.

src/devices/HardwareMonitor/OpenHardwareMonitorWmi.cs
```diff
@@ -0,0 +1,250 @@
...
+namespace Iot.Device.HardwareMonitor
+{
+    /// <summary>
+    /// This class connects to a running instance of OpenHardwareMonitor and reads out all available values using WMI.
+    /// This works only if OpenHardwareMonitor (https://openhardwaremonitor.org/) is currently running.
+    /// While the tool needs to be run with elevated permissions, the application using this binding does not.
+    /// The connection using WMI has proven to be unreliable (the WMI classes are sometimes just not visible), therefore
+    /// OHM versions greater > 0.10.0 use a rest api instead.
+    /// </summary>
+    internal sealed class OpenHardwareMonitorWmi : IDisposable, IOpenHardwareMonitorInternal
```
krwq, line 32: considering this is legacy what's the benefit of supporting this?

src/devices/HardwareMonitor/SensorJson.cs
```diff
@@ -0,0 +1,62 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
krwq, line 2: consider moving all `*Json.cs` files to a separate folder like: `model/` or `JsonModel/`, moving them to a separate namespace and removing prefixes

krwq (approved):
generally looks good, I assume this is moderately tested - I left couple of optional comments

src/devices/HardwareMonitor/OpenHardwareMonitorWmi.cs
pgrawehr: It would work with the officially released version of OHM. The new transport requires a new build, which should be available shortly.

PR #1759, opened 2022-01-07
Make all internal types in Common internal OR public

pgrawehr:
This cleans up a few dependency issues between the bindings and the common project. The classes marked internal (including the Interop classes) are now public when building CommonHelpers.csproj but still internal for the final assembly. This removes the need to link to individual source files outside their binding folders, which causes build problems when they're used from multiple places.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1759)


src/devices/Common/Interop/Unix/Libasound/Interop.libasound.cs
```diff
@@ -0,0 +1,152 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+#if BUILDING_IOT_DEVICE_BINDINGS
+internal
+#else
+public
+#endif
+    partial class Interop
```
krwq, line 17: I think this will be more readable:

```suggestion
#if BUILDING_IOT_DEVICE_BINDINGS
internal partial class Interop
#else
public partial class Interop
#endif
```

krwq: same comment for other places

src/devices/Common/Interop/Unix/Libc/Interop.libc.cs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+[Flags]
+
+#if BUILDING_IOT_DEVICE_BINDINGS
+internal
+#else
+public
+#endif
+    enum MemoryMappedFlags
```
krwq, line 68: perhaps we could make all of these subclasses nested so that we don't have as many ifdefs

src/devices/Common/Iot/Device/Common/NumberHelper.cs
```diff
@@ -8,7 +8,12 @@ namespace Iot.Device.Common
     /// <summary>
     /// Helpers for number.
     /// </summary>
-    internal static class NumberHelper
+#if BUILDING_IOT_DEVICE_BINDINGS
+    internal
+#else
+    public
+#endif
+    static class NumberHelper
```
krwq, line 16: another option for ifdefs is do following:
define just `partial class` without any visibility

create a single file which defines all classes and call it like something like: Visibilty.Public.cs, and put just:

```csharp
public partial static class NumberHelper {}
//... other classes
```

and another one with all internal and change csproj to include one or the other

pgrawehr: This is a good idea. I'll consider it.

src/devices/Common/Interop/Unix/Libc/Interop.libc.cs
pgrawehr: Done

src/devices/Common/Iot/Device/Common/NumberHelper.cs
pgrawehr: That worked quite well. Since the default when no visibility is given is "internal", I only had to provide a "Visibility.Public" file.

src/devices/Common/ClassVisibility.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
```
krwq, line 1: copyright header

src/devices/Common/ClassVisibility.cs
```diff
@@ -0,0 +1,43 @@
...
+
+// These instructions make the classes public for building the individual projects.
+// The default visibility is "internal" so they stay hidden if this section is removed for the final library build.
+#if !BUILDING_IOT_DEVICE_BINDINGS
+
+#pragma warning disable SA1403 // File may only contain a single namespace
+
+public partial class Interop
+{
+}
```
krwq, line 15: consider putting these in a single line:

```suggestion
public partial class Interop { }
```

src/devices/Common/Common.csproj
```diff
@@ -1,13 +1,15 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>$(DefaultBindingTfms)</TargetFrameworks>
     <EnableDefaultItems>false</EnableDefaultItems>
     <RootNamespace>
     </RootNamespace>
     <Nullable>enable</Nullable>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <NoWarn>$(NoWarn);CS8321;SA1205;</NoWarn>
```
krwq, line 9: can you write comment on what individual warnings mean? Also consider doing SA1205 per file rather than globally

src/devices/Common/Interop/Unix/Libasound/Interop.libasound.struct.cs
```diff
@@ -0,0 +1,97 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Disable these StyleCop rules for this file, as we are using native names here.
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+#pragma warning disable SA1307 // Field should begin with upper-case letter
+#pragma warning disable CS1591
```
krwq, line 7: can you add a comment what CS1591 is?

src/devices/Common/Interop/Windows/WinUser.cs
```diff
@@ -5,7 +5,9 @@
 using System.Runtime.InteropServices;
 
 #pragma warning disable SA1300
-internal partial class Interop
+#pragma warning disable CS1591
```
krwq, line 8: please add a comment on what warnings are

src/devices/Media/VideoDevice/Devices/Interop.videodev2.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Disable these StyleCop rules for this file, as we are using native names here.
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+#pragma warning disable SA1307 // Field should begin with upper-case letter
+
+internal partial class InteropVideodev2
```
krwq, line 8: consider following class structure:

```csharp
partial class Interop
{
  public class Videodev2 { /* ... */ }
}
```

src/devices/Media/VideoDevice/Devices/Interop.videodev2.struct.cs
```diff
@@ -0,0 +1,412 @@
...
+
+// Disable these StyleCop rules for this file, as we are using native names here.
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+#pragma warning disable SA1307 // Field should begin with upper-case letter
+
+using System;
+using System.Runtime.InteropServices;
+using Iot.Device.Media;
+
+internal partial class InteropVideodev2
```
krwq, line 12: consider moving this under Interop class

src/devices/Pn532/ErrorCode.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-namespace IoT.Device.Pn532
+namespace Iot.Device.Pn532
```
krwq, line 4: good catch

krwq (approved):
LGTM after addressing comments

src/devices/Common/ClassVisibility.cs
pgrawehr: This is not allowed by the current StyleCop rules. I could disable it here, but I'd rather keep it like this.

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -12,6 +12,9 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
     <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage)</TargetsForTfmSpecificContentInPackage>
+    <!--Partial elements should declare access modifier: Disabled for this project to simplify handling in common project. See also there.
+    If we disable this here, the warning is still emmited for the individual projects.-->
+    <NoWarn>$(NoWarn);SA1205</NoWarn>
```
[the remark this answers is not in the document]
pgrawehr: I would need to disable it in many files in the common project, which would somewhat jeopardize the nice solution with not specifying a visibility for the internal classes. 

As the documentation says: Because all files are also compiled using a separate project, the warning does come up for classes not in the common project.

src/devices/Media/VideoDevice/Devices/Interop.videodev2.cs
pgrawehr: That does not work here, because the class `Interop` is public while building the binding and resides in a different assembly. Partial classes cannot be merged from two assemblies.

src/devices/Directory.Build.props
```diff
@@ -13,12 +13,18 @@
...
     <SystemDeviceModelProjectName>System.Device.Model</SystemDeviceModelProjectName>
     <SystemDeviceModelPath>$(MSBuildThisFileDirectory)$(SystemDeviceModelProjectName)/$(SystemDeviceModelProjectName).csproj</SystemDeviceModelPath>
+    <CommonProjectName>Common</CommonProjectName>
+    <CommonProjectPath>$(MSBuildThisFileDirectory)$(CommonProjectName)/$(CommonProjectName).csproj</CommonProjectPath>
     <LangVersion>9</LangVersion>
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(MSBuildProjectName)' != '$(CommonProjectName)' AND '$(MSBuildProjectName)' != '$(SystemDeviceModelProjectName)'">
+  <!-- These projects are included into every binding. They must not depend on each other, or we get a circular dependency -->
```
joperezr, line 23: NIT: Indenting on this comment.

src/devices/Directory.Build.props
```diff
@@ -13,12 +13,18 @@
...
+    <ProjectReference Include="$(SystemDeviceModelPath)" />
```
joperezr, line 24: Usually when we we use infrastructure to automatically Opt-in all of the child projects into, it is good practice to provide an Opt-out switch, for example, a property like `Condition="'$(OmitCommonProjectReference)' != 'true'"` or something like that. If you do that, then instead of having to use the Condition you have above in the ItemGroup, you can just set those properties onto the Common and System.Device.Model projects themselves which would remove the ProjectReference from there. You don't have to make that change now, and I'm fine to keeping this as is, just wanted to share that other point of view.

src/devices/Arduino/Arduino.sln
```diff
@@ -15,7 +15,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Mcp3xxx", "..\Mcp3xxx\Mcp3x
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CharacterLcd", "..\CharacterLcd\CharacterLcd.csproj", "{D47A6627-4041-4CEA-8903-A6C67C6993CF}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CommonHelpers", "..\Common\CommonHelpers.csproj", "{CD593083-8E94-4B60-86BF-DF3FB7873893}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Common", "..\Common\Common.csproj", "{CD593083-8E94-4B60-86BF-DF3FB7873893}"
```
joperezr, line 18: How come I only see the .sln change on some of the binding projects you changed? For example, I don't see the change in the BuildHat.sln

src/devices/Common/ClassVisibility.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+// These instructions make the classes public for building the individual projects.
+// The default visibility is "internal" so they stay hidden if this section is removed for the final library build.
+#if !BUILDING_IOT_DEVICE_BINDINGS
+
+#pragma warning disable SA1403 // File may only contain a single namespace
+
+public partial class Interop
```
joperezr, line 16: Honestly I don't love the fact of separating access modifiers from their implementation using partials. Given you would only have one instace per file, why not instead just add the ifdef on the class implementation itself to select if the class is either public or itnernal? I think having it separate like this will be harder to maintain and prone to errors.

joperezr: FWIW, we are already using that approach with at least three helper classes in common, I think we should use the same with the rest.

src/devices/Media/VideoDevice/Devices/Interop.videodev2.cs
joperezr: Perhaps then consider moving this file over to the Common project? I do feel like having a central place for all of the Interop code makes sense.

src/devices/Pn532/ErrorCode.cs
joperezr: Wow!

joperezr:
Thanks so much for all of this great cleanup @pgrawehr! I left some minor comments, and the only major one is on how to execute the visibility of the common classes, but other than that this looks great!

src/devices/Arduino/Arduino.sln
pgrawehr: Good catch. The compiler doesn't check the sln files, so probably I only updated the ones I had to open. I'll fix this.

src/devices/Common/ClassVisibility.cs
pgrawehr: This was @krwq's suggestion, and I think it's nice. It reduces the boilerplate on many files, and I think it reduces the possibility of errors. One would clearly not by accident change this file, but when one changes one of the other files (and i.e. adds an access specifier there) a build error would result. We could make this a bit more waterproof by anyway adding the explicit internal case.

src/devices/Media/VideoDevice/Devices/Interop.videodev2.cs
pgrawehr: The problem here is that these classes depend on public(!) enums in the Media project. I could move them, too, but then _they_ would be a bit on the odd side.

krwq (approved):

krwq:
@joperezr I think for this specific case we should vote on the design - I personally prefer cleaner code (less `#if`s).

Please vote:
- 🚀 for less `#if`s
- 🎉 for explicit visibility (with `#if`s)

krwq:
Either way I agree with offline comment - if we decide to keep Visibility file we should be always explicit to make sure there is compile error on accidental change (also create Visibiltity.Interal.cs)

src/devices/Media/VideoDevice/Devices/Interop.videodev2.cs
pgrawehr: No, that's what I'm saying: Those enums are public. (I did not test whether they are rightfully so, though)

Issue #1644, opened 2021-08-20
Display abstraction interface


dotMorten:
I've noticed all the various displays have different methods exposed to work with them. I'd like to suggest an interface (or abstract base class) is provided for all of them, so we can write common code, and quickly swap out the display used in our code, without having to change anything but the display creation.
I've found that I'd often change displays to something better, and having rewrite quite a lot of code because of it, and it would also make it easier for people to adapt existing code and apps for the display they have.

For instance it's pretty typical that there's a method for `SendBitmap(Bitmap)`, `ClearScreen()` and `FillRect(Color,Rectangle`)`, and it would be nice to be able to write display-agnostic code against all of these. It would also allow us to build a common text and line drawing API for all of these.
The interface would also need a few getters to get display info like width/height, bit depth, etc, information about whether there's on/off control (and if so methods for doing that).

**Describe the ideal solution**

Suggestion for interface
```cs
public interface IDisplay
{
     void SendBitmap(Bitmap bmp);
     void FillRect(Color color, Rectangle rect);
     void ClearScreen();
     Task ResetScreenAsync();
     void SetDisplayOn();
     void SetDisplayOff();     
     bool IsDisplayOnOffSupported() { get; } 
     ushort Width { get; }
     ushort Height { get; }
     byte BitsPerPixel { get; }
     bool IsColorDisplay { get; }
}
```

pgrawehr:
I fully agree with the requirement. Additionally, there should be an implementation for `ICharacterLcd` for graphic displays, so they easily support printing text. This can probably be done in a generic way, by providing a class that implements `ICharacterLcd` and outputs a bitmap. Oh, but note that we should not be using `System.Drawing.Bitmap` any more, but the ImageSharp equivalent instead (see #1403). 

And then, it would be great if we had an OS-independent way of taking a screen shot of the main display, so that one could do things like mirroring (part of) the main screen to a SPI display. Looking for the linux way of doing that didn't really bring up many solutions for me yet. [This is independent, but would make these displays a lot more useful]

raffaeler:
I am also favorable to the proposal but I have a couple of observations.

The name of the interface is ambiguous. Displays based on the Hitachi interface can't do most of the thing specified in the interface. I would rather split text and graphics display methods. They could eventually be implemented both on most graphical devices.

Also, given the amount of devices that can display something, it would be better providing a method to return its capabilties. This could avoid the mess of catching the `NotImplementedException`.

richlander:
[...]

Issue #1403, opened 2021-01-07
We should replace dependency on System.Drawing with ImageSharp


krwq:
System.Drawing is officially deprecated, we should use ImageSharp instead. See: https://docs.microsoft.com/en-us/dotnet/api/system.drawing?view=dotnet-plat-ext-5.0#remarks

Ellerbach:
[Triage] We need to list all the bindings that uses System.Drawing and identify groups which need to be adjusted together. We also need to make sure that the person who'll do some of the work has the proper hardware to test it.

Ellerbach:
Here is the list of references to System.Drawing in the repo:
- samples\led-matrix-weather
- src\devices\Apa102 -> seems to only use Color
- src\devices\Bh1745 -> seems to only use Color
- src\devices\Common\Iot\Device\Graphics -> seems to only use Color
- src\devices\GoPiGo3 -> seems to only use Color
- src\devices\Media\VideoDevice -> uses Color and PixelFormat, Size and maybe few others. This will be more complicated to migrate
- src\devices\PiJuice -> seems to only use Color
- src\devices\RGBLedMatrix -> Color, Rectangle and most likely few others. This may require quite some efforts
- src\devices\SenseHat -> seems to only use Color
- src\devices\Ssd1351 -> uses Color, Bitmap, Point. This will require quite some efforts
- src\devices\Tcs3472x -> seems to only use Color
- src\devices\Ws28xx -> seems to only use Color

Those only using Color are quite straight forward to migrate. They require always the same pattern to transform the color space to the right one. A limited hardware can be needed to test but the pattern will always be the same. So that can limit the errors when the hardware is not available.

The problematic ones will be VideoDevice, RGBLedMAtrix and Ssd1351. They will for sure require the hardware to test and make sure all is working perfectly.

MaxPrimeAERY:
VideoDevice doesn't work for me properly. I get an error in this part of code
 video = device.CaptureContinuous();
 Bitmap myBitmap = new Bitmap(video);

System.ArgumentException: Parameter is not valid. 
   at System.Drawing.Image.InitializeFromStream(Stream stream)
   at System.Drawing.Bitmap..ctor(Stream stream)

Ellerbach:
@MaxPrimeAERY can you please create a separate issue for this problem? This is not related to this specific issue.

JimBobSquarePants:
[...]

PR #1761, opened 2022-01-12
Another WS2812B driver (with updated APA102)

HumJ0218:
I wrote another WS2812B driver. It is used in a similar way to APA102.
I added a method for correcting visual chromatic aberration.
This driver has been working steadily for more than 3 days on a 150-LEDs-strip.
If you find it appropriate, I will continue to add the remaining project files.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1761)


src/devices/Common/IIntegratedLed.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Drawing;
+
+namespace Iot.Device
+{
+    public abstract class IntegratedLed : IDisposable
```
krwq, line 9: Can we rename it to something like `LedStrip`? (P.S. the filename also has `I` prefix and it should match the class name instead)

src/devices/Common/IIntegratedLed.cs
```diff
@@ -0,0 +1,36 @@
...
+        public double Gamma { get; set; } = 2.2;
+
+        /// <summary>
+        /// Colors of LEDs
+        /// </summary>
+        public Span<Color> Pixels => _pixels;
+
+        public abstract void Flush();
+
+        protected Color[] _pixels;
```
krwq, line 23: Protected members should follow the same naming conventions as public: `Pixels`, Honestly I'd recommend to just make this public rather than artificially wrapping the other member with Span for no real benefit (use Span when it can save allocations, if it's already allocated then there is no really point in doing so unless you plan to hide some implementation detail perhaps)

src/devices/Common/IIntegratedLed.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        protected byte FixGamma(byte v)
```
krwq, line 25: should this be internal?

src/devices/Common/IIntegratedLed.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            return (byte)(Math.Pow(v/255.0, Gamma)*255);
```
krwq, line 27: nit: spaces around operators

```suggestion
            return (byte)(Math.Pow(v / 255.0, Gamma) * 255);
```

src/devices/Ws2812b/Ws2812b.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the WS2812B device.
+        /// </summary>
+        /// <param name="spiDevice">The SPI device used for communication.</param>
+        /// <param name="length">Number of LEDs</param>
+        public Ws2812b(SpiDevice spiDevice, int length)
+        {
+            _spiDevice = spiDevice ?? throw new ArgumentNullException(nameof(spiDevice));
+            _pixels = new Color[length];
```
krwq, line 38: the constructor of the abstract class should take the length instead and allocate this buffer. If you're gonna make array public make sure to only make getter available publicly (i.e. `public Color[] Pixels { get; private set; }`)

src/devices/Ws2812b/Ws2812b.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// Update color data to LEDs
+        /// </summary>
+        public override void Flush()
+        {
+            for (int i = 0; i<_pixels.Length; i++)
+            {
+                Color pixel = _pixels[i];
+                GetSequence(FixGamma(pixel.G)).CopyTo(_buffer.AsSpan(ResetDelay+i*PixelSize+ColorSize*0));
```
krwq, line 50: nit: spaces around operators, here and everywhere

src/devices/Ws2812b/Ws2812b.cs
```diff
@@ -0,0 +1,78 @@
...
+        public const int SpiClockFrequency = 3000000;
+
+        private SpiDevice _spiDevice;
+        private readonly byte[] _buffer;
+
+        private const byte Bit0 = 0b0100; // 333ns low(dummy), 333ns high(T0H), 667ns low(T0L)
+        private const byte Bit1 = 0b1100; // 667ns high(T1H), 667ns low(T1L)
+        private const byte ResetDelay = 105; // 280us low
+
+        private const byte ColorSize = 4;
```
krwq, line 27: I think `ColorBitSize` is more appropriate name, also please move it above consts defining bits so it's clearer they're related

src/devices/Ws2812b/Ws2812b.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// Update color data to LEDs
+        /// </summary>
+        public override void Flush()
+        {
+            for (int i = 0; i<_pixels.Length; i++)
+            {
+                Color pixel = _pixels[i];
+                GetSequence(FixGamma(pixel.G)).CopyTo(_buffer.AsSpan(ResetDelay+i*PixelSize+ColorSize*0));
+                GetSequence(FixGamma(pixel.R)).CopyTo(_buffer.AsSpan(ResetDelay+i*PixelSize+ColorSize*1));
+                GetSequence(FixGamma(pixel.B)).CopyTo(_buffer.AsSpan(ResetDelay+i*PixelSize+ColorSize*2));
```
krwq, line 52: This is inappropriate use of Span and doesn't help prevent allocations.

Instead consider following:
- edit `GetSequence` to have following signature: `private void WriteSequence(byte b, Span<byte> destination)`
- here, pass in the Span where you want sequence to be written

krwq (approved):
LGTM logicwise, couple of nits and span uses should be fixed.

krwq:
(also fix build errors)

src/devices/Ws2812b/Ws2812b.cs
Ellerbach: Tip: do 'ctrl + k + d' with Visual Studio and all the formatting will be magically applied :-)

HumJ0218: > Tip: do 'ctrl + k + d' with Visual Studio and all the formatting will be magically applied :-)

It didn't work, and I usually use Ctrl+E+F , it should be same.
I suspect it's a bug, it's been like that since a certain update.

HumJ0218: > Tip: do 'ctrl + k + d' with Visual Studio and all the formatting will be magically applied :-)

I'm now using vs 17.0.4

joperezr:
[Triage] @HumJ0218 can you please address the comments that @krwq added so that we can go ahead and merge this PR?

Ellerbach (approved):
Looks good  to me once all comments addressed.

HumJ0218:
@joperezr can you go ahead this PR? I have a SK6812-RGBW depends on this and ready to commit

krwq:
@HumJ0218 sorry for late review, just a question: how is this Ws2812b different from https://github.com/dotnet/iot/blob/main/src/devices/Ws28xx/Ws2812B.cs ?

krwq:
codewise LGTM but still would like to understand first why there is a need for two bindings with the same name

pgrawehr:
Please fix the merge conflict as well. `CommonHelpers.csproj` has been renamed to just `Common.csproj`

HumJ0218:
@krwq I remember the previous ws2812b didn't perform well (on devices except Raspberry Pi 3B), and I felt the need to unify the operation of this type of LED-strip

joperezr:
[Triage] @HumJ0218 We are still a bit confused here. We believe that there shouldn't be two WS2812B device bindings. Think about it this way, if I'm a developer with a WS2812B and I want to use this repo's bindings for my project, and I see two classes for my device, which one should I choose? This confusion will be hard to explain, so we believe that the right way to approach this would be to have a single WS2812B class, which could potentially use two different underlying implementations by using different constructors, but in the end we should only have one type.

Does that make sense? Would it be possible for you to make these requested changes? Is there a reason why you still think that the two types should be separate?

Ellerbach:
@HumJ0218 I had a deeper look. Also I recently did some work on the .NET nanoFramework side. See here https://github.com/nanoframework/nanoFramework.IoT.Device/tree/develop/devices/Ws28xx.Esp32 and here https://github.com/nanoframework/nanoFramework.IoT.Device/tree/develop/devices/Ws28xx.
And for what you want to acheive, the best is to create a BitmapImage class. This is where you can add your specific encoding. There are quite few already, the names are not the best (we can revisit that). So far what I've seen is the following:

* 3 bit encoding RGB or GRB (Neo3 in this repo)
* 8 bit encoding RGB or GRB (WS2808 in this repo)
* What you are bringing adding the gama component if I can read properly

Our original idea was to hide somehow this complexity with driver names and classes. Unfortunately, the same "name" can use different implementations of the underlying encoding.

So, rather than trying to continue in this direction, we may rethink the approach and have a better naming for each of the encoding and better documentation. And rethink the classes for the constructors with a default bitmap image to keep some compatibility.

Let me know if you need help to move your implementation into a BitmapImage. Also if you are willing to give a try for the renaming and exposing them as public (will be needed for the constructor), that would be great. Same for the documentation :-). Now, if you're willing to do only the first step, that's fine, I can take over the next ones.

HumJ0218:
@Ellerbach This driver is also flawed in its specific performance, so I‘m no longer considering whether to add a new ws2812b driver. But I still want to discuss about color, pixels, and images.
Because of functions for graphics drawing is more mature, I prefer to use `System.Drawing`. I can draw lines, ellipses, rectangles, and write text on image as much as I want, and transform images at will. The only downside is that it is no longer available in environments such as Linux
`Neo3` and `SixLabors.ImageSharp`, their support for the above features is incomplete.
I think it's important to strengthen graphics-related features first, or to be careful about choosing a more powerful graphics library.

pgrawehr:
@HumJ0218 Yes, we are working on this, see the discussion on #1403, #1644 and others. We expect that in the future, the bindings will take some kind of abstract image type and just do the conversion to the required format for the hardware. The entire drawing part will be external (either in a new library or with e.g. ImageSharp). We haven't made any final decision yet.

krwq:
[Triage]: Doubled Ws2812b implementation should be merged or removed from this PR and done through BitmapImage abstract class (i.e. https://github.com/dotnet/iot/blob/main/src/devices/Ws28xx/BitmapImageNeo3.cs) - if there is any improvements you're doing here please merge it with existing BitmapImage for this binding. Apa102 should be implemented also with BitmapImage, it should be relatively simple change

joperezr:
[Triage]: @HumJ0218 let us know if you are still interested in applying the suggestions from @krwq above.

PR #1762, opened 2022-01-13
Added MPU6050 support

serega404:
Based on the MPU6886 codebase

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1762)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1762) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1762) before we can accept your contribution.<br/><br/>:x: serega404 [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1762)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1762) it.</sub>

Ellerbach:
Thanks for the submission. The MPU6050 is quite similar to the MPU9250 and MPU6500. There is a lot in common starting with the names of the registers, the values, etc.
It would be great if you can either align all the naming with the existing MPU9250 and MPU6500, or even better, merge them both so you can have a common base class.
Looking at both it seems that the default address is the same, that the access is the same, the accelerator range is the same, etc, etc.
It seems there is really a lot you can reuse.
Your naming implementation is different than for the 2 others. Please adjust all this and have a deep look at this before we can make more code review.

Ellerbach (changes requested):
As I wrote in the comment, please reuse and merge this code with the existing MPU6500 and PLU9250. There a lot which is the exact same. Thanks!

serega404:
I'm thinking in this case to create a universal code for the MPU6000 series by removing mpu6500 and replacing it with mpu6000, which will support both boards

Ellerbach:
> I'm thinking in this case to create a universal code for the MPU6000 series by removing mpu6500 and replacing it with mpu6000, which will support both boards

That will work. Please then adjust what's in the MPU9250 directory, please do not change the existing names there to avoid breaking changes (if one or 2 are really needed let's see). We may then think of renaming the directory to reflect the family change.
Thanks again!

serega404:
I just discovered that there is an MPU6500 class inside the MPU9250 folder, which is very strange. I think maybe this class will be able to work with MPU6050. When I get home, I'll check it out.

Ellerbach:
> I just discovered that there is an MPU6500 class inside the MPU9250 folder, which is very strange. I think maybe this class will be able to work with MPU6050. When I get home, I'll check it out.

the MPU9250 = MPU6500 + AK8963
So reason why it's in it. Now, if there is more of the MPU family that can be added, we can rename the folder and be more descriptive as well on what's supported.

src/devices/Mpu9250/Mpu6050.cs
```diff
@@ -0,0 +1,775 @@
...
+        private GyroscopeBandwidth _gyroscopeBandwidth;
+        internal bool _wakeOnMotion;
+
+        /// <summary>
+        /// Initialize the MPU6050
+        /// </summary>
+        /// <param name="i2cDevice">The I2C device</param>
+        public Mpu6050(I2cDevice i2cDevice)
+        {
+            _i2cDevice = i2cDevice;
```
Ellerbach, line 52: you should check for null and throw in that case

Ellerbach (changes requested):
Looking at the Mpu6050 class file, I feel like 80% is common with the current Mpu6500. I would make more sense to have a base call with all what is common, and then have only the differences in each class.
This will allow to avoid copy/paste of code and have a very large part which is the same.

serega404:
I've already thought about it, I'm afraid I might break something in the mpu6500 part, since the mpu6050 will become the base class for the mpu6500, because the mpu6500 is the mpu6050 + second accelerometer and calibration capability. Since I don't have an mpu6500, I won't be able to test the code for this device.

Ellerbach:
> I've already thought about it, I'm afraid I might break something in the mpu6500 part, since the mpu6050 will become the base class for the mpu6500, because the mpu6500 is the mpu6050 + second accelerometer and calibration capability. Since I don't have an mpu6500, I won't be able to test the code for this device.

I have an MPU9250. So I can test the changes. Right now, with all the same code we won't merge it tike this. Really don't fear extracting what is common and having the MPU6050 based on the MPU6500 or having a third one that contains what is common.

Ellerbach:
@serega404 did you have time to move forward? Do you need help?

serega404:
There hasn't been time yet, but I'll come back to that later.

Ellerbach:
@Sergio0694 , I have been removing duplicates. I'm trying to push it on your branch but you have raised the PR on you main branch, so makes it a bit complicated.

Ellerbach:
So a PR is waiting for you in your repository, merge it and we should see it properly here.

serega404:
Unfortunately, I stopped working on this project and switched to others aimed at preserving important Internet resources, because I am afraid of the emergence of the "Iron Curtain". I will try to finish the edits by the end of this week.

Ellerbach:
@joperezr do you have an idea why the build is failing for Windows only?

pgrawehr:
@Ellerbach The build is failing due to a breaking change with respect to the latest official release. These checks run in the windows build only. You need to create an exclusion file (if the changes are acceptable).

Ellerbach:
> The build is failing due to a breaking change with respect to the latest official release

Ho! That does make sense now! Because of the folder name and namespaces! 

> You need to create an exclusion file (if the changes are acceptable).

Some does. So I guess @serega404 we can resume the discussion on those namespace and folder :-)

joperezr:
Sorry for missing the question @Ellerbach, yes as @pgrawehr points out package validation checks for breaking changes only run on windows, since they only run when we call pack. To repro this problem locally, you can always just call the pack task directly on Iot.Device.Bindings project like: `dotnet pack Iot.Device.Bindings.csproj`. In order to generate the suppressions so that packaging won't fail after the breaking change is discussed and accepted, you can call: `dotnet pack Iot.Device.Bindings.csproj /p:GenerateCompatibilitySuppressionFile=true` which should automatically update the suppressions file and the build will now pass.

joperezr:
Of course, as @pgrawehr pointed out, it goes without saying that any changes into the suppressions file should be very well understood and discussed since it would be silencing a breaking change.

src/devices/Mpu6xxx9xxx/README.md
```diff
@@ -1,6 +1,6 @@
-# MPU6500/MPU9250 - Gyroscope, Accelerometer, Temperature and Magnetometer (MPU9250 only)
+# MPU6xxx/MPU9250 - Gyroscope, Accelerometer, Temperature and Magnetometer (MPU9250 only)
```
Ellerbach, line 1:
```suggestion
# MPU6050/MPU6500/MPU9250 - Gyroscope, Accelerometer, Temperature and Magnetometer (MPU9250 only)
```

Ellerbach (approved):
Thanks, looks good to me and seems all ok for the various MPU. Yes, the folder name seems better like this, will increase the visibility.

Ellerbach:
@joperezr seems that I can't rerun the test (tried on another PR). Do you mind rerunning it for this one? The Windows build is not related to the PR but still, would be nicer with all green :-)

PR #1763, opened 2022-01-15
Added LP55231 Support

ibebbs:
Adds device binding for [Texas Instrument's Lp55231](https://www.ti.com/product/LP55231) with sample showing it's use with [Sparkfun's Lp55231 breakout](https://learn.sparkfun.com/tutorials/lp55231-breakout-board-hookup-guide/all).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1763)


ibebbs:
As far as understand it, the unit test failures causing the "Some checks were not successul" have nothing to do with the changes in this PR.

Could someone let me know if I need to amend this PR to resolve the issue?

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,168 @@
...
+            Span<byte> bytes = stackalloc byte[2];
+            bytes[0] = (byte)register;
+            bytes[1] = data;
+            _device.Write(bytes);
+        }
+
+        private void SetIntensity(byte index, byte value)
+        {
+            byte register = (byte)(Register.REG_D1_PWM + index);
+            WriteRegister(register, value);
```
Ellerbach, line 95: as you seems to have all the registers, (Register)((byte)Register.REG_D1_PWM + index) should make the trick with just the register function

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// Setting this value will stop eny programs currently running.
+        /// </remarks>
+        public bool Enabled
+        {
+            get
+            {
+                var register = ReadRegister(Register.REG_CNTRL1);
+                return (register & (byte)Control1RegisterFlags.Enabled) > 0;
+            }
+            set
```
Ellerbach, line 136: you're missing an enter here. I'm surprised the build didn't raised it :-)

src/devices/Lp55231/MiscFlags.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// <summary>
+    /// Flags related to the REG_MISC register
+    /// </summary>
+    [Flags]
+    public enum MiscFlags : byte
+    {
+        /// <summary>
+        /// Use internal clock
+        /// </summary>
+        ClockSourceSelection = 0b00000001,
```
Ellerbach, line 14: tip: it's better to make it 0b0000_0001
so using the underscore to separate in 2 the binary representation. The eye have hard time to catch more than groups of 3 or 4 elements. And if something is missing on a 8 elements in a raw, it definitely won't on a 2 time 4, you're increasing the changes :-)

src/devices/Lp55231/MiscFlags.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Use external clock
+        /// </summary>
+        ExternalClockDetection = 0b00000010,
+
+        /// <summary>
+        /// Enable PWM cycle power save
+        /// </summary>
+        PWMCyclePowersaveEnable = 0b00000100,
```
Ellerbach, line 24: Pwm rather than PWM, we're using is like this everywhere. So it's for consistency.

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// Default I2C address of Motor Hat
+        /// </summary>
+        public const int I2cAddressBase = 0x32;
+
+        /// <summary>
+        /// Gets the channel of the red element of the specific instance
+        /// </summary>
+        /// <param name="instance">The RGB instance (0-2)</param>
+        /// <returns>The channel number</returns>
+        public static byte RedChannel(byte instance) => (byte)(6 + instance);
```
Ellerbach, line 26: how do you use those 3 functions? Can you add something in the samples and the readme? I can't get my head around how to use those because I most likely don't have the sensor. But an explanation in the readme and a sample would definitely help!

Ellerbach (commented):
Thanks for the submission, few comments and suggestions.

src/devices/Lp55231/Control1RegisterFlags.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
```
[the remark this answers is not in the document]
ibebbs: Done

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,168 @@
...
+{
+    /// <summary>
+    /// Raspberry Pi Motor Hat based on PCA9685 PWM controller
+    /// </summary>
+    public class Lp55231 : IDisposable
+    {
+        /// <summary>
+        /// Default I2C address of Motor Hat
+        /// </summary>
+        public const int I2cAddressBase = 0x32;
```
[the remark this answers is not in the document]
ibebbs: Done

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Lp55231"/> class with the specified I2C settings.
+        /// </summary>
+        /// <param name="settings">The I2C settings of the Lp55231.</param>
+        /// <remarks>
+        /// The default i2c address is 0x60, but the HAT can be configured in hardware to any address from 0x60 to 0x7f.
+        /// The PWM hardware used by this HAT is a PCA9685. It has a total possible frequency range of 24 to 1526 Hz.
+        /// Setting the frequency above or below this range will cause PWM hardware to be set at its maximum or minimum setting.
+        /// </remarks>
+        public Lp55231(I2cConnectionSettings? settings = default)
```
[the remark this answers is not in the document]
ibebbs: Added a new constructor to allow passing either I2cSettings or an I2cDevice. If a device is passed in, it won't be disposed when dispose is called.

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,168 @@
...
+                new RgbLed(2, SetIntensity)
+            };
+        }
+
+        private byte ReadRegister(Register register)
+        {
+            return ReadRegister((byte)register);
+        }
+
+        private byte ReadRegister(byte register)
```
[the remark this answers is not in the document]
ibebbs: Done

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,168 @@
...
+            _device.WriteByte((byte)register);
+            return _device.ReadByte();
+        }
+
+        private void WriteRegister(Register register, byte data)
+        {
+            WriteRegister((byte)register, data);
+        }
+
+        private void WriteRegister(byte register, byte data)
```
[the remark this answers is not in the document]
ibebbs: Done

src/devices/Lp55231/Lp55231.cs
ibebbs: Done

src/devices/Lp55231/MiscFlags.cs
ibebbs: Done

src/devices/Lp55231/MiscFlags.cs
ibebbs: Done

src/devices/Lp55231/Lp55231.cs
ibebbs: I've made these internal. They're basically wrappers around the calculation for the appropriate channel number and only used in the `RGBLed` class. The reason they look so odd is that the red channels are kept separate from the blue and green channels as they latter are driven by the IC's internal charge pump whereas the former is driver directly from Vdd (because blue and green LEDS require higher current). This effectively makes the channel layout GBGBGBRRR.

src/devices/Lp55231/Lp55231.cs
ibebbs: Where is the missing enter? Layout looks correct to me??

ibebbs:
I believe I had addressed all the feedback from the review.

src/devices/Lp55231/Lp55231.cs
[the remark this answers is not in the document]
Ellerbach: The pattern is not to create the I2cDevice so the constructor with I2cSettings should be removed. And when passing the I2cDevice, the I2cDevice should be disposed.
Check some of the other bindings using I2C, that will give you a good idea of what's needed.

src/devices/Lp55231/Lp55231.cs
Ellerbach: between } and set. There should be a blank new line after a }

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// </summary>
+        /// <param name="device">The I2C device to communicate with the Lp55231.</param>
+        /// <remarks>
+        /// The default i2c address is 0x60, but the HAT can be configured in hardware to any address from 0x60 to 0x7f.
+        /// The PWM hardware used by this HAT is a PCA9685. It has a total possible frequency range of 24 to 1526 Hz.
+        /// Setting the frequency above or below this range will cause PWM hardware to be set at its maximum or minimum setting.
+        /// </remarks>
+        public Lp55231(I2cDevice device)
+        {
+            _device = device;
```
Ellerbach, line 57: you should check for null and throw an exception _device = device ?? throw NullArgumentException(nameof(device));

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,174 @@
...
+                new RgbLed(2, SetIntensity)
+            };
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Lp55231"/> class with the specified I2C settings.
+        /// </summary>
+        /// <param name="settings">The I2C settings of the Lp55231.</param>
+        /// <remarks>
+        /// The default i2c address is 0x60, but the HAT can be configured in hardware to any address from 0x60 to 0x7f.
```
Ellerbach, line 72: As you mention HAT, is this sensor part of another hardware? Also you mention PCA9685 so not sure it's the right description :-)

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,174 @@
...
+        private readonly I2cDevice _device;
+        private readonly IReadOnlyList<RgbLed> _rgbLeds;
+        private readonly bool _disposeDevice;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Lp55231"/> class with the specified I2C settings.
+        /// </summary>
+        /// <param name="device">The I2C device to communicate with the Lp55231.</param>
+        /// <remarks>
+        /// The default i2c address is 0x60, but the HAT can be configured in hardware to any address from 0x60 to 0x7f.
```
Ellerbach, line 51: See the other comment, not sure about the comment all up

src/devices/Lp55231/Lp55231.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Lp55231"/> class with the specified I2C settings.
+        /// </summary>
+        /// <param name="settings">The I2C settings of the Lp55231.</param>
+        /// <remarks>
+        /// The default i2c address is 0x60, but the HAT can be configured in hardware to any address from 0x60 to 0x7f.
+        /// The PWM hardware used by this HAT is a PCA9685. It has a total possible frequency range of 24 to 1526 Hz.
+        /// Setting the frequency above or below this range will cause PWM hardware to be set at its maximum or minimum setting.
+        /// </remarks>
+        public Lp55231(I2cConnectionSettings? settings = default)
```
Ellerbach, line 76: you should not have this constructor. Pattern is not to create an I2C device in constructors. The bus ID can vary, the I2cDevice can be attached to another type of hardware. There are so many cases so that's the reason we are not doing it.

src/devices/Lp55231/README.md
```diff
@@ -0,0 +1,41 @@
+﻿# Lp55231
```
Ellerbach, line 1: Maybe something like this:
```suggestion
﻿# Lp55231 - Nine-Channel RGB, White-LED Driver
```

src/devices/Lp55231/RgbLed.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            if (setIntensity == null)
+            {
+                throw new ArgumentNullException(nameof(setIntensity));
+            }
+
+            _instance = instance;
+            _setIntensity = setIntensity;
+        }
+
```
Ellerbach, line 39: Maybe we can add a Color property directly?

src/devices/Lp55231/Lp55231.cs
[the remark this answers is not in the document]
ibebbs: Done

src/devices/Lp55231/Lp55231.cs
ibebbs: Done

src/devices/Lp55231/RgbLed.cs
ibebbs: I initially considered using Color but decided against it as:
1) It meant taking a dependency on another assembly (System.Drawing); and
2) It didn't allow color components to be changed independently.

Despite this, given your suggestion, I have refactored to use Color. I will add the ability to change individual channel intensities at a later date.

src/devices/Lp55231/README.md
ibebbs: Done

src/devices/Lp55231/Lp55231.cs
ibebbs: Removed.

src/devices/Lp55231/Lp55231.cs
ibebbs: Amended/Removed

src/devices/Lp55231/Lp55231.cs
ibebbs: Amended/Removed

src/devices/Lp55231/Lp55231.cs
ibebbs: Done

ibebbs:
Completed further changes addressing PR feedback

Ellerbach (commented):
I guess you will hate me but as System.Drawing is deprecated, we decided to move to ImageSharp. For the Color, won't change much, really almost nothing. Sorry, I forgot to mention that. And you'll have a bit more work to adjust. And thanks again for making it with colors!
And you still can keep the individual components, it's not shocking to me.
Rest looks great to me!

ibebbs:
No worries, I've replaced System.Drawing.Common with SixLabors.ImageSharp.

joperezr (approved):

joperezr:
Thanks so much for your contribution @ibebbs!

PR #1785, opened 2022-01-28
"Arduino C# Compiler"

pgrawehr:
Draft PR, so that I get a regular CI run

This "compiler" (actually a metadata converter and code analyzer) can take a compiled Iot program written using the Arduino binding and flash it as a standalone application to an ESP32 or similar 32-bit microcontroller. 

Test and debug your application on the PC using the Arduino binding, and when everything works as expected, just flash it to the ESP32 and run it without the PC. 

There are still a lot of open ends, but several more-or-less complex programs already work (see the WeatherStation example). Further documentation (both on usage as well as on internals) is still to be written.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1785)


src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -137,7 +137,7 @@ internal RaspberryPi3Driver(RaspberryPi3LinuxDriver linuxDriver)
             {
                 RaspberryBoardInfo.Model.RaspberryPi3B or
                 RaspberryBoardInfo.Model.RaspberryPi3APlus or
-                RaspberryBoardInfo.Model.RaspberryPi3BPlus or
+                RaspberryBoardInfo.Model.RaspberryPi3BPlus or
```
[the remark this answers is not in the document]
pgrawehr: It appears to be caused by an incorrect line ending somewhere. I undid the change, because it definitely doesn't belong here.

pgrawehr:
Please make sure to merge #1841 first (may require a conflict resolution afterwards)

pgrawehr:
@joperezr Anything open here? I would be happy if this could be merged before I leave for my holidays.

joperezr:
Haven't had the chance to take another look. I will take a look at it before Thursday.

joperezr:
Let's revert all of the changes that are going into the azure-pipelines.yml as I would prefer not to block CI due to a problem with this since this is a tool, and the script is pulling an external repo so that seems very fragile. Apart from the fact that our current CI is running in around 10-15 mins, and your two new legs are averaging 60 mins so we definitely don't want to quadruple the build time for PRs for something that won't really add any protection to the shipping code.

 Also, seems like part of your changes are doing additional things which we don't want, for example you are changing the vmImage for some of the pools, and you are passing in /macCpuCount to the build command causing a regression in build performance.

eng/ArduinoCsCI.cmd
```diff
@@ -0,0 +1,66 @@
+echo on
```
joperezr, line 1: Per my previous comment, let's just remove this file for now. We can later decide to add a separate pipeline that we can conditionally invoke in some PRs that where it is relevant.

src/devices/Board/I2cBusExtensions.cs
```diff
@@ -44,5 +44,112 @@ public static List<int> PerformBusScan(this I2cBus bus, int lowest = 0x3, int hi
...
+        /// <returns>A list of bus addresses that are in use, an empty list if no device was found</returns>
+        /// <remarks>This method should never throw an exception. Bus scanning may interfere with normal device operation,
+        /// so this should not be done while devices are being used.</remarks>
+        public static (List<int> FoundDevices, int LowestAddress, int HighestAddress) PerformBusScan(this I2cBus bus, IProgress<float>? progress, int lowestAddress = 0x3, int highestAddress = 0x77)
+        {
+            List<int> addresses = new();
+            int numberOfDevicesToScan = highestAddress - lowestAddress;
+            float currentPercentage = 0;
+            float stepPerDevice = 100.0f / numberOfDevicesToScan;
+            if (lowestAddress < 0)
```
joperezr, line 64: NIT: You should probably move all of this argument validation to the top of the method body since you don't want to do any initialization in a failure case.

src/devices/Board/I2cBusExtensions.cs
```diff
@@ -44,5 +44,112 @@ public static List<int> PerformBusScan(this I2cBus bus, int lowest = 0x3, int hi
...
+            if (numberOfDevicesToScan <= 0)
+            {
+                return (addresses, 0, 0);
+            }
+
+            for (int addr = lowestAddress; addr <= highestAddress; addr++)
+            {
+                try
+                {
+                    using (var device = bus.CreateDevice(addr))
```
joperezr, line 83: NIT: I know I haven't commented on this in the past, but we usually try to avoid using `var` keyword unless the type can be inferred by the assignment like for example: `var x = new Device();` so please try to switch all of these other uses where you are not calling the constructor directly and instead use the concrete type. There is a roslyn analyzer that flags this and it does provide a fix-all solution, so perhaps we should go and run this code-fixer all over our source code in a subsequent PR.

tools/ArduinoCsCompiler/ArduinoCsCompiler.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Library</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <Configurations>Debug;Release</Configurations>
+    <Copyright>The .NET Foundation</Copyright>
+    <Company>The .NET Foundation</Company>
+	<EnableDefaultItems>false</EnableDefaultItems>
```
joperezr, line 10: NIT: you have some tabs instead of spaces here.

tools/ArduinoCsCompiler/ArduinoCsCompiler.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
joperezr, line 1: I think for now we should remove this project (as well as all  of the projects underneath this like the samples, Frontend, and tests) from building when build.cmd is invoked using some Exclusions in build.proj file in the root of the repo. 99.9% of the people building the repo won't really care about building them, and this is currently being built since we have:

https://github.com/dotnet/iot/blob/6be64b7477d86aca49503498e862d7be9962489d/build.proj#L38

Alternatively we can disable BuildTools target entirely on the default build.cmd and only enable it when people care about it, so perhaps in your Readme you can add a line saying that they should call build like `build.cmd /p:ToolsBuild=true`

eng/ArduinoCsCI.cmd
pgrawehr: This is the build script, not the yml file. We might adapt or change this, but deleting won't help.

tools/ArduinoCsCompiler/ArduinoCsCompiler.csproj
pgrawehr: I basically agree, but is it worth the extra complexity? The tools take like 2 seconds of the overall build time.

pgrawehr:
> Let's revert all of the changes that are going into the azure-pipelines.yml as I would prefer not to block CI due to a problem with this since this is a tool, and the script is pulling an external repo so that seems very fragile. 

Actually, it also tests the existing Arduino binding against (simulated) hardware. What about pinning the revision on the external repo, so that changes there would only affect this when explicitly updated?

> Apart from the fact that our current CI is running in around 10-15 mins, and your two new legs are averaging 60 mins so we definitely don't want to quadruple the build time for PRs for something that won't really add any protection to the shipping code.

Yes, that's indeed bad. I'll (try to) add a variable to normally not run the tool tests. 

> 
> Also, seems like part of your changes are doing additional things which we don't want, for example you are changing the vmImage for some of the pools, and you are passing in /macCpuCount to the build command causing a regression in build performance.

Sure, removed those changes again, didn't intend to leave that in.

I'll look into the "nits" later. Those should be easy.

azure-pipelines.yml
```diff
@@ -89,6 +90,28 @@ stages:
       artifact: config
       condition: eq(variables['BuildConfiguration'], 'Release')
 
+  - job: Windows_ArduinoIntegration
```
joperezr, line 93: Let's not add this to the main build yet. We should either make a new pipeline for this that can be invoked manually for a PR (probably ideal) or have some sort of logic that detects for changes on the Arduino or compiler bindings and would only run this job if any changes are detected. This latter option is less ideal, mainly cause you can't really invoke it if you want to run it even when changes are not detected, and it is also less maintainable since could be prone to breaking if files get moved around.

joperezr (approved):
Left one comment about removing this from the main pipeline, but other than that, this looks good to go.

azure-pipelines.yml
pgrawehr: I actually don't see much of a problem the way it is now. This is easily identifyable as it is a separate build, and when it breaks on a PR it can quickly be checked whether the breakage is due to a bad change in the Arduino binding or whether it is due to some problem completely unrelated to the PR. This is much like we do with the hardware tests now. The number of PRs we have is not so big that this extra check cost really a lot of time right now.

pgrawehr:
Merged as-is. Instead of endless discussions on possible build issues I think it's easier to address those if they show up. It would be easy to disable the Arduino build if it kept breaking the build unjustified.

PR #1786, opened 2022-01-30
packages required by .NET Standard only

tibel:
Those package references are needed by .NET Standard 2.0 only.
.NET 6.0 includes the needed types already.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1786)

### DIFF
--- a/src/System.Device.Gpio/System.Device.Gpio.csproj
+++ b/src/System.Device.Gpio/System.Device.Gpio.csproj
@@ -8,22 +8,23 @@
     <IsPackable>true</IsPackable>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
+
   <ItemGroup>
     <Compile Remove="winmd\**" />
     <EmbeddedResource Remove="winmd\**" />
     <None Remove="winmd\**" />
     <None Include="buildTransitive\net5.0\System.Device.Gpio.targets" Pack="true" PackagePath="\buildTransitive\net5.0" />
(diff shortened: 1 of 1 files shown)


pgrawehr (approved):
I guess CI wouldn't run if the observation wasn't correct. Thanks!

joperezr (changes requested):
Even though the changes are correct, technically speaking removing a dependency from a framework is considered a breaking change, and given this is in System.Device.Gpio package, I think we should be more hesitant to take these kinds of breaks. For example, I would suggest delaying this for our next major bump release, as doing it now would be a breaking change and violate SemVer

joperezr:
I really appreciate the contribution @tibel, and again, the changes you are making here are the correct ones, I just don't think we should do them yet. My suggestion is to open an issue and mark it as 3.0 milestone, and close this PR for now, and re-open once we are close to the 3.0 release.

joperezr:
After doing some additional checks and with the help of @ericstj, we found that actually this wouldn't be a breaking change since all of the package dependencies that are being removed from "net6.0" here already have a matching assembly identity that ships inbox and tooling will prefer over the NuGet packages, which means that when targeting "net6.0" these extra nuget dependencies would have never even made it to the compiler anyway. Because of that, I'm fine with going ahead and merging this now, as this is no longer considered a breaking change.

joperezr (approved):

joperezr:
Thanks for the fix @tibel, and sorry for the initial confusion here.

PR #1789, opened 2022-02-05
NMEA Support binding

pgrawehr:
This adds NMEA 0183 parsing and sending features.

- Parses a large number of NMEA message types
- Can send out all messages as well
- Message router to forward messages between different devices or interfaces
- Supports forwarding unknown messages
- TCP and UDP servers for NMEA
- Autopilot support
- Fully tested on a real boat. Unit test covers all parsing and string building
- Calculator for compass deviations
- Basic geodetic functions for WGS84 positions

Original concept by @krwq 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1789)


src/devices/Nmea0183/Route.cs
```diff
@@ -0,0 +1,252 @@
...
+using System.Linq;
+using System.Text;
+using Iot.Device.Common;
+
+namespace Iot.Device.Nmea0183
+{
+    /// <summary>
+    /// High-level representation of a route (a list of waypoints)
+    /// </summary>
+    public sealed class Route
```
Ellerbach, line 15: Any good reason to seal it?

src/devices/Nmea0183/RoutePoint.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// <param name="indexInRoute">The index of this point</param>
+        /// <param name="totalPointsInRoute">The total number of points on this route</param>
+        /// <param name="waypointName">The name of this waypoint</param>
+        /// <param name="position">The position of the waypoint</param>
+        /// <param name="bearingToNextWaypoint">The direction to the next waypoint (optional, will be calculated when the route is constructed)</param>
+        /// <param name="distanceToNextWaypoint">The distance to the next waypoint (optional, will be calculated when the route is constructed)</param>
+        /// <remarks>
+        /// Route and point names should use ASCII characters only. Some devices may understand the extended ASCII code page (values > 127),
+        /// but this is rare. Most devices will silently truncate any names longer than 10 chars.
+        ///
```
Ellerbach, line 31: extra line here

src/devices/Nmea0183/Sentences/EngineRevolutions.cs
```diff
@@ -0,0 +1,203 @@
...
+using System.Globalization;
+using System.Text;
+using UnitsNet;
+
+namespace Iot.Device.Nmea0183.Sentences
+{
+    /// <summary>
+    /// Source for the revolution data
+    /// </summary>
+    public enum RotationSource
```
Ellerbach, line 15: You should move it to a new file for clarity.

src/devices/Nmea0183/Sentences/HeadingAndDeclination.cs
```diff
@@ -0,0 +1,191 @@
...
+        public override string ToNmeaParameterList()
+        {
+            if (Valid)
+            {
+                StringBuilder b = new StringBuilder();
+                b.AppendFormat(CultureInfo.InvariantCulture, "{0:F1},", HeadingTrue.Normalize(true).Degrees);
+                if (Deviation.HasValue)
+                {
+                    b.AppendFormat(CultureInfo.InvariantCulture, "{0:F1},{1},", Math.Abs(Deviation.Value.Degrees),
+                        Deviation.Value.Degrees >= 0 ? "E" : "W");
```
Ellerbach, line 157: You are using quite often "E" and "W" and "N" and "S" all up. Does it makes sense to create a global const or char enum? You even have the `CardinalDirection` as internal byte.

src/devices/Nmea0183/Sentences/HeadingAndDeclination.cs
```diff
@@ -0,0 +1,191 @@
...
+
+            return string.Empty;
+        }
+
+        /// <inheritdoc />
+        public override string ToReadableContent()
+        {
+            if (Valid && Declination.HasValue)
+            {
+                return $"Magnetic Heading: {HeadingTrue.Degrees:F1}°, Declination: {Declination.Value.Degrees:F1}°";
```
Ellerbach, line 185: the ° should be encoded as unicode like we do in the samples.

src/devices/Nmea0183/Sentences/HeadingTrue.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            return string.Empty;
+        }
+
+        /// <inheritdoc />
+        public override string ToReadableContent()
+        {
+            if (Valid)
+            {
+                return $"True Heading: {Angle.Normalize(true).Degrees:F1}°";
```
Ellerbach, line 103: same here for the degree, should be unicode encoded. I guess it should appear elsewhere as well

src/devices/Nmea0183/Sentences/RecommendedMinimumNavigationInformation.cs
```diff
@@ -0,0 +1,332 @@
...
+
+            // ddddmm.mm
+            double? degreesMinutes = integerDegrees * 100 + minutes;
+            return (degreesMinutes, direction);
+        }
+
+        /// <summary>
+        /// Navigation status enum
+        /// </summary>
+        public enum NavigationStatus : byte
```
Ellerbach, line 319: You may want to move this one in a new file as well

src/devices/Nmea0183/Sentences/TransducerDataSet.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// - B (bar)
+        /// - P (Pascal, Percentage)
+        /// - D (degrees)
+        /// - G (Acceleration, g-Force)
+        /// - C (degrees Celsius)
+        /// - M (Mass, Liters)
+        /// - V (Voltage)
+        /// - A (Ampere)
+        /// </summary>
+        public string Unit { get; set; }
```
Ellerbach, line 51: Would it make sense to have an enum rather than a 1 length string with those know values?
Similar question for the rest. It doesn't prevent to work if there is another unknown, you'll still get the value anyway

src/devices/Nmea0183/Sentences/WindSpeedAndAngle.cs
```diff
@@ -0,0 +1,210 @@
...
+        public WindSpeedAndAngle(Angle angle, Speed speed, bool relative)
+            : base(OwnTalkerId, Id, DateTimeOffset.UtcNow)
+        {
+            Speed = speed;
+            Relative = relative;
+            Angle = angle;
+            Valid = true;
+            if (Speed.Unit == UnitsNet.Units.SpeedUnit.Knot)
+            {
+                SpeedUnit = "N";
```
Ellerbach, line 38: I would really use an enum for all those 1 string units. Will make things a bit simpler to use and to read as well as a human

Ellerbach (approved):
That's a very very very complete parser! Very impressive work and I know you're using it in the real life. So just few comments but all up, looks all good to me.

src/devices/Neo-M8/NeoM8.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public sealed class NeoM8 : IDisposable
+    {
+        private SerialPort _sp;
+
+        /// <summary>
+        /// Constructs object representing Neo-M8
+        /// </summary>
+        /// <param name="serialPortName">Name of the serial port to use for communication</param>
+        public NeoM8(string serialPortName)
```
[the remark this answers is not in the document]
pgrawehr: The Neo-M8 is a concrete device, NMEA is the protocol. When @krwq started with the design of this code, he started with this concrete (very simple) receiver. Let's leave it to him whether we want to keep that separate.

src/devices/Nmea0183/Sentences/TransducerDataSet.cs
pgrawehr: The 1-character string is used in the message protocol itself, and since the values are not even unique (e.g. P can be pascals as well as percentage), I'd rather leave the interpretation to the client.

src/devices/Nmea0183/Sentences/HeadingAndDeclination.cs
pgrawehr: I've not done this for a long time. Our code files are all encoded in UTF8-BOM, so this should give no problems at all (otherwise, we would be in big trouble with files such as https://github.com/dotnet/iot/blob/main/src/devices/Common/Iot/Device/Graphics/LcdCharacterEncodingFactory.cs)

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -0,0 +1,334 @@
...
+
+            return true;
+        }
+
+        /// <summary>
+        /// Returns true if the two positions are (almost) equal. This ignores the altitude.
+        /// </summary>
+        /// <param name="position">Position to compare with</param>
+        /// <returns>True if the two positions are closer than about 1cm</returns>
+        public bool EqualPosition(GeographicPosition position)
```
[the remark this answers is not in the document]
pgrawehr: Implementing the equality operators, that does make sense. But I can't think of a valid use case where you would add, subtract or multiply positions. Distances between positions are implemented, but on a different class. I'll consider adding those functions via extension methods.

src/devices/Nmea0183/Route.cs
pgrawehr: It's not really designed for inheritance. A list of something should normally be aggregated, not inherited from.

pgrawehr:
Ugly. The macOS hang is due to https://github.com/dotnet/runtime/issues/64551

pgrawehr:
> wow! Do we have a sense of how much larger now Iot.Device.Bindings will be? (I'm wondering if we should create another nuget package just for this)

Iot.Device.Bindings seems to grow from 1365kb to 1520kb (release build), so by about 155kb.

krwq:
@pgrawehr were any pieces here auto-generated? if so, I'd recommend to generate it as part of the build as well

src/devices/Nmea0183/Raw8BitEncoding.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Iot.Device.Nmea0183
+{
+    /// <summary>
+    /// An encoding that converts 8-Bit data to unicode, by just interpreting the input values as the lower part of a
+    /// char. This works for all ASCII characters as well as if the input and output devices use the same code page.
+    /// </summary>
+    public class Raw8BitEncoding : Encoding
```
krwq, line 14: wouldn't unicode encoding just work here?

src/devices/Nmea0183/Identification.Extra.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            return Equals((Identification)obj);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                var hashCode = (shipNameField != null ? shipNameField.GetHashCode() : 0);
```
krwq, line 52: please use https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=net-6.0

src/devices/Nmea0183/Identification.Extra.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(this, obj))
+            {
+                return true;
+            }
+
+            if (obj.GetType() != GetType())
```
krwq, line 39: any particular reason to block derived classes? I'd keep this to minimum: null check and refer to typed Equals

azure-pipelines.yml
```diff
@@ -47,7 +47,7 @@ stages:
       displayName: 'Execute Markdownlint'
       condition: eq(variables['build.reason'], 'PullRequest')
 
-    - script: build.cmd -ci
+    - script: build.cmd -ci /maxCpuCount:1
```
krwq, line 50: can you elaborate on this? cc: @joperezr

src/devices/Common/Iot/Device/Common/AngleExtensions.cs
```diff
@@ -0,0 +1,160 @@
...
+                    r = -(Math.Abs(r) % (Math.PI * 2));
+                    if (r < -Math.PI)
+                    {
+                        r += Math.PI * 2;
+                    }
+                }
+            }
+
+            // Return in same unit as original input
+            return Angle.FromRadians(r).ToUnit(self.Unit);
```
krwq, line 65: I'm confused, what's `ToUnit` actually doing? is it just internal representation or something else?

azure-pipelines.yml
pgrawehr: This is for testing only. It allows to see the output of all test runs in CI. I'll remove it when everything is ready.

src/devices/Common/Iot/Device/Common/AngleExtensions.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        /// <summary>
+        /// Calculate the difference between two angles. Useful to compute the angle error between a desired and an actual track.
+        /// </summary>
+        /// <param name="currentTrack">First angle, actual direction</param>
+        /// <param name="destinationTrack">Second angle, desired direction</param>
+        /// <returns>The difference</returns>
+        public static Angle Difference(Angle currentTrack, Angle destinationTrack)
+        {
+            currentTrack = currentTrack.Normalize(true);
```
krwq, line 76: why do you need to normalize before subtraction, shouldn't one normalize after the operation be enough?

src/devices/Common/Iot/Device/Common/AngleExtensions.cs
pgrawehr: The UnitsNet types contain the value in the original unit plus that unit. The calculation within this method needs to use degrees, because we don't know the equivalents for 360° resp 180° for arbitrary units. So at the top of the method, the angle is converted to degrees, and here it is converted back to an Angle instance with the same unit as was initialy used.

src/devices/Common/Iot/Device/Common/AngleExtensions.cs
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                }
+            }
+
+            // Return in same unit as original input
+            return Angle.FromRadians(r).ToUnit(self.Unit);
+        }
+
+        /// <summary>
+        /// Calculate the difference between two angles. Useful to compute the angle error between a desired and an actual track.
```
krwq, line 69: you should mention which direction you're calculating the difference

src/devices/Common/Iot/Device/Common/AngleExtensions.cs
krwq: this feels like it does exactly the same as TrueToMagnetic except more complicated, can you explain the difference if there is any?

src/devices/Common/Iot/Device/Common/AngleExtensions.cs
```diff
@@ -0,0 +1,160 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using UnitsNet;
+
+namespace Iot.Device.Common
```
krwq, line 11: I'll suggest to add this in a separate PR with more explanation

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -0,0 +1,371 @@
...
+        /// </summary>
+        /// <param name="a">First instance to compare</param>
+        /// <param name="b">Second instance to compare</param>
+        /// <returns>True on inequality, false otherwise</returns>
+        public static bool operator !=(GeographicPosition? a, GeographicPosition? b)
+        {
+            return !(a == b);
+        }
+
+        private static string GetEastOrWest(double sign)
```
krwq, line 172: should this be called angle? also might be worth adding a comment about expected range

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -0,0 +1,371 @@
...
+            string strEastOrWest = GetEastOrWest(normalizedVal);
+
+            args[0] = deg;
+            args[1] = DegreesSymbol;
+            args[2] = min;
+            args[3] = MinutesSymbol;
+            args[4] = sec.ToString("00.00");
+            args[5] = SecondsSymbol;
+            args[6] = strEastOrWest;
+            string strLonRet = string.Format(CultureInfo.InvariantCulture, "{0}{1} {2:00}{3} {4}{5}{6}", args);
```
krwq, line 205: can you just use string interpolation here without allocating array?

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -0,0 +1,371 @@
...
+
+            return true;
+        }
+
+        /// <summary>
+        /// Returns true if the two positions are (almost) equal. This ignores the altitude.
+        /// </summary>
+        /// <param name="position">Position to compare with</param>
+        /// <returns>True if the two positions are closer than about 1cm</returns>
+        public bool EqualPosition(GeographicPosition position)
```
krwq, line 272: you should take `almost` delta as a parameter here

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -0,0 +1,371 @@
...
+        {
+            if (position == null)
+            {
+                return false;
+            }
```
krwq, line 277: I'd just throw ArgumentNullException here

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -0,0 +1,371 @@
...
+            if ((Math.Abs((position.Longitude - Longitude)) < ComparisonEpsilon) && (Math.Abs(position.Latitude - Latitude) < ComparisonEpsilon))
+            {
+                ret = true;
+            }
+            else
+            {
+                ret = false;
+            }
+
+            return ret;
```
krwq, line 289:
```suggestion
            return (Math.Abs((position.Longitude - Longitude)) < ComparisonEpsilon) && (Math.Abs(position.Latitude - Latitude) < ComparisonEpsilon);
```

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -0,0 +1,371 @@
...
+        }
+
+        /// <summary>
+        /// Equality comparer. Compares the two positions for equality within about 1cm.
```
krwq, line 293: In this case I think this is fine to pick default value... could this call into the other method though?

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -0,0 +1,371 @@
...
+            {
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Returns a string representation of this position.
+        /// </summary>
+        /// <returns>A string representation in degrees, minutes and seconds for both latitude and longitude</returns>
+        // TODO: Add different formatting options and add parsing feature
```
krwq, line 346: remove this, we'll add it when we need to, alternatively open the issue - still - TODOs should be out of the code as it can potentially stay there forever

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -0,0 +1,371 @@
...
+        {
+            if (Double.IsNaN(Latitude) || Double.IsNaN(Longitude))
+            {
+                return "NaN";
+            }
+
+            if (Double.IsInfinity(Latitude) || Double.IsInfinity(Longitude))
+            {
+                return "Infinity";
+            }
```
krwq, line 357: is this technically possible? should we be instead not allowing to create such objects?

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -0,0 +1,371 @@
...
+
+            var strLatRet = GetLatitudeString(Latitude);
+            var strLonRet = GetLongitudeString(Longitude);
+
+            return string.Concat(strLatRet, " / ", strLonRet, " Ellipsoidal Height: ", EllipsoidalHeight.ToString("F0"));
```
krwq, line 362: string interpolation?

src/devices/Common/Iot/Device/Common/GreatCircle.cs
```diff
@@ -0,0 +1,356 @@
...
+using UnitsNet;
+
+namespace Iot.Device.Common
+{
+    /// <summary>
+    /// This class provides some basic geodetic calculations on the WGS84 ellipsoid.
+    /// This returns exact distances between two points, considering the non-spherical shape of the earth.
+    /// The results are accurate for distances greater than about 1cm.
+    /// </summary>
+    public static class GreatCircle
```
krwq, line 17: I'm not sure I like the name; I like the original name `GeoidCalculations` though - I'd recommend to put the other file in different namespace and re-use that name

src/devices/Common/Iot/Device/Common/GreatCircle.cs
```diff
@@ -0,0 +1,356 @@
...
+
+        /// <summary>
+        /// Calculate the coordinate one will be when traveling for the given distance in the given direction
+        /// </summary>
+        /// <param name="start">Starting point</param>
+        /// <param name="direction">Initial direction</param>
+        /// <param name="distance">Distance to travel</param>
+        /// <returns>The new position</returns>
+        /// <exception cref="ArgumentNullException">The start position is null</exception>
+        public static GeographicPosition CalcCoords(GeographicPosition start, Angle direction, Length distance)
```
krwq, line 187: These names need some work - can we do this in separate PR?

src/devices/Common/Iot/Device/Common/PositionExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            return new GeographicPosition(position.Latitude, NormalizeAngleTo180Degrees(position.Longitude), position.EllipsoidalHeight);
+        }
+
+        /// <summary>
+        /// Normalizes the angle to +/- 180°
+        /// </summary>
+        public static double NormalizeAngleTo180Degrees(double angleDegree)
+        {
+            angleDegree %= 360;
```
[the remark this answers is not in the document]
krwq: (also this is not really an extension)

src/devices/Neo-M8/Neo-M8.csproj
```diff
@@ -0,0 +1,19 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <None Include="README.md" />
+    <ProjectReference Include="../Nmea0183/Nmea0183.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.IO.Ports" Version="4.6.0-preview8.19405.3" />
```
krwq, line 16: use variable for version - we should already have it defined somewhere

pgrawehr:
> @pgrawehr were any pieces here auto-generated? if so, I'd recommend to generate it as part of the build as well

No, there's no autogenerated code here. Everything done by plain typing ⌨️

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
pgrawehr: No, because this one compares all three values (lat, long and height), while `EqualPosition` explicitly only compares lat and long

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
pgrawehr: Been there, done that 😆 I'll remove it and write a ticket.

src/devices/Nmea0183/Identification.Extra.cs
pgrawehr: Good catch, this code gets auto-generated by Resharper, and I never really thought of what this means.

src/devices/Common/Iot/Device/Common/AngleExtensions.cs
pgrawehr: Improved documentation

src/devices/Common/Iot/Device/Common/GreatCircle.cs
pgrawehr: The problem is, that that would not be unique. We might also have distance calculations based on loxodromes instead of the great circle. But let's discuss that together with improved namings separatelly.

src/devices/Common/Iot/Device/Common/AngleExtensions.cs
pgrawehr: You where right, this first line is superfluous. The two methods are not exactly equal, one returns a direction angle (0..360) the other a delta (-180..+180) and the use case is different.

src/devices/Nmea0183/RoutePoint.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            return ReferenceEquals(this, obj) || (obj is RoutePoint other && Equals(other));
+        }
+
+        /// <summary>
+        /// Standard hash function
+        /// </summary>
+        public override int GetHashCode()
+        {
+            return WaypointName.GetHashCode() ^ Position.GetHashCode();
```
krwq, line 137: please combine hash codes as mentioned in the other examples, here and everywhere

src/devices/Nmea0183/SentenceCache.cs
```diff
@@ -0,0 +1,503 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Clean up everything
+        /// </summary>
+        public void Dispose()
+        {
+            _source.OnNewSequence -= OnNewSequence;
+            _sentences.Clear();
```
krwq, line 491: should you null out fields so they can be collected by gc?

krwq (approved):
I have skimmed through most of the files and not done the super thorough review, please fix the hash code calculation (use hash code combiner rather than xor or custom math). What I looked at overall looks good. Given this has been tested on the physical boat I have relatively high confidence in this version being a fine first iteration. Thanks for putting this up!

pgrawehr:
>  please fix the hash code calculation (use hash code combiner rather than xor or custom math). 

@krwq I have not found any use of `HashCode.Combine` in the code base, and it's use would require some workarounds, since it is not supported on .netstandard2.0.

krwq:
@pgrawehr it should work starting netstandard2.1. Simplest workaround is tuples: `(123, "asdf").GetHashCode()` and I'm ok with using that.

The reason I ask you to change this is that it may turn into a security issue one day (we've had cases similar to that in the past)

pgrawehr:
> @pgrawehr it should work starting netstandard2.1. Simplest workaround is tuples: `(123, "asdf").GetHashCode()` and I'm ok with using that - it should be using correct way of combining hash codes underneath.
> 
> The reason I ask you to change this is that it may turn into a security issue one day (we've had cases similar to that in the past)

You might want to explain that to me. I'm agreeing to the answer given [here](https://security.stackexchange.com/questions/127807/any-vulnerability-in-net-implementation-for-c-object-gethashcode): `GetHashCode` is not intended for security considerations. I've recently been quite confused that some implementations (e.g `string.GetHashCode`) are now written such that the hashcode of the same string changes when the application restarts.

joperezr:
For the Hashcode.Combine availability issue, simply use https://www.nuget.org/packages/Microsoft.Bcl.HashCode/

raffaeler:
Congrats @pgrawehr for the great work!

PR #1709, opened 2021-10-30
Added Support for SK6812-Strips

Rdlprmpf:
Moved existing sample to subfolder, added new sample for SK6812, updated the Readme.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1709)

PR #1800, opened 2022-02-20
Added support for WS2815B

Rdlprmpf:
- Added support for WS2815B (same as WS2812, but with different color order (RGB instead of GRB)
- Replaced the sample with a more complete one as discussed [in my last pull request](https://github.com/dotnet/iot/pull/1709)


HumJ0218:
and another LED SK6812-RGBW, have a bonus WHITE channel

( I commited a PR #1761 a new WS2812B and when it passed I plan to commit a return-to-zero-code LED device includes all of them like WS2812(SK6812), SK6812-RGBW. It seems like WS2815 is also on the road :P

src/devices/Ws28xx/BitmapImageNeo3.cs
```diff
@@ -19,20 +19,35 @@ internal class BitmapImageNeo3 : BitmapImage
...
+        public BitmapImageNeo3(int width, int height, ColorMode colorMode)
+            : this(width, height)
+        {
+            ColorMode = colorMode;
+        }
+
         public BitmapImageNeo3(int width, int height)
             : base(new byte[width * height * BytesPerPixel + ResetDelayInBytes], width, height, width * BytesPerPixel)
         {
         }
```
krwq, line 31:
```suggestion
        public BitmapImageNeo3(int width, int height, ColorMode colorMode = ColorMode.GRB)
            : base(new byte[width * height * BytesPerPixel + ResetDelayInBytes], width, height, width * BytesPerPixel)
        {
            ColorMode = colorMode;
        }
```

src/devices/Ws28xx/BitmapImageNeo3.cs
```diff
@@ -19,20 +19,35 @@ internal class BitmapImageNeo3 : BitmapImage
...
         public override void SetPixel(int x, int y, Color c)
         {
             var offset = y * Stride + x * BytesPerPixel;
-            Data[offset++] = _lookup[c.G * BytesPerComponent + 0];
-            Data[offset++] = _lookup[c.G * BytesPerComponent + 1];
-            Data[offset++] = _lookup[c.G * BytesPerComponent + 2];
-            Data[offset++] = _lookup[c.R * BytesPerComponent + 0];
-            Data[offset++] = _lookup[c.R * BytesPerComponent + 1];
-            Data[offset++] = _lookup[c.R * BytesPerComponent + 2];
+            Data[offset++] = _lookup[ColorMode == ColorMode.GRB ? c.G : c.R * BytesPerComponent + 0];
```
krwq, line 45: would it make sense to  create just one if statement per pixel or loop?

src/devices/Ws28xx/ColorMode.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
```
krwq, line 1: copyright header missing

src/devices/Ws28xx/BitmapImageNeo3.cs
```diff
@@ -19,20 +19,35 @@ internal class BitmapImageNeo3 : BitmapImage
         // data means we have to add 30 bytes of zero padding.
         private const int ResetDelayInBytes = 30;
 
+        public BitmapImageNeo3(int width, int height, ColorMode colorMode)
```
krwq, line 22: Honestly I think you should just derive from `BitmapImageNeo3` and replace SetPixel with the new version

krwq: and remove ColorMode as it feels like an implementation detail of the device in here

src/devices/Ws28xx/samples/LEDStripSample/Animations.cs
```diff
@@ -0,0 +1,237 @@
+﻿using System.Drawing;
```
krwq, line 1: copyright

src/devices/Ws28xx/samples/LEDStripSample/Animations.cs
```diff
@@ -0,0 +1,237 @@
...
+            return SupportsSeparateWhite ? Color.FromArgb(0, source.R, source.G, source.B) : source;
+        }
+
+        /// <summary>
+        /// Animation similar to "Knight Rider".
+        /// </summary>
+        /// <param name="token">The token.</param>
+        public async void KnightRider(CancellationToken token)
+        {
+            // https://youtu.be/oNyXYPhnUIs
```
krwq, line 65: I wouldn't put a link as it may go out of date one date

src/devices/Ws28xx/samples/LEDStripSample/MenuId.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
```
krwq, line 1: add copyright header in all files where it's missing

src/devices/Ws28xx/BitmapImageNeo3.cs
Rdlprmpf: OK

src/devices/Ws28xx/BitmapImageNeo3.cs
Rdlprmpf: OK

src/devices/Ws28xx/BitmapImageNeo3.cs
```diff
@@ -38,7 +36,12 @@ public override void SetPixel(int x, int y, Color c)
             Data[offset++] = _lookup[c.B * BytesPerComponent + 2];
         }
 
-        private static readonly byte[] _lookup = new byte[256 * BytesPerComponent];
+        protected static readonly byte[] _lookup = new byte[256 * BytesPerComponent];
+
+        protected static int BytesPerComponent => 3;
+
+        protected static int BytesPerPixel => BytesPerComponent * 3;
```
krwq, line 43: I'm kinda curious about these two being changed from const to static properties, why not just proetcted const?

src/devices/Ws28xx/samples/LEDStripSample/Program.cs
```diff
@@ -0,0 +1,404 @@
...
+}
+
+if (ledCount == 0)
+{
+    return;
+}
+
+while (true)
+{
+    Console.Write($"{Environment.NewLine}1: WS2812b{Environment.NewLine}2: WS2815b{Environment.NewLine}3: SK68012{Environment.NewLine}0: Exit{Environment.NewLine}Type of Strip: ");
```
krwq, line 43: I think this will be way more readable if you just do separate Console.WriteLine for each of these

src/devices/Ws28xx/README.md
```diff
@@ -2,11 +2,12 @@
 
 This binding allows you to update the RGB LEDs on Ws28xx / SK6812 and based strips and matrices.
 
-To see how to use the binding in code, see the [sample](samples/Ws28xx_Samples/Program.cs).
+To see how to use the binding in code, see the [sample](samples/LEDStripSample/Program.cs).
 
 ## Documentation
 
 * WS2812B: [Datasheet](https://cdn-shop.adafruit.com/datasheets/WS2812B.pdf)
+* WS2815B: [Datasheet](http://www.world-semi.com/DownLoadFile/138)
```
krwq, line 10: Please also add very tiny sample here how to use or initialize this class, even a single comment after line 49 will suffice

src/devices/Ws28xx/BitmapImageNeo3.cs
Rdlprmpf: I'm always having bad feelings about exposing fields. Even if they're protected.

See also [this discussion on StackOverflow](https://stackoverflow.com/a/3182664). I think these are still valid points.

src/devices/Ws28xx/samples/LEDStripSample/Program.cs
Rdlprmpf: OK

src/devices/Ws28xx/README.md
Rdlprmpf: OK.

src/devices/Ws28xx/BitmapImageNeo3.cs
krwq: We're allowed to make breaking changes later, we assume are derived classed are used only by device implementation which should live in this repo - it's not meant to be a generic purpose abstraction

Rdlprmpf: While I do not agree to "ignore" good practice just because we're allowed to make breaking changes, I see the point that this class is internal anyway, so it won't really be a problem to change these back to fields. Will do that later today :-)

Any opinion about renaming the class from BitmapImage to something else? (See my comment [above](https://github.com/dotnet/iot/pull/1800#issuecomment-1052434106)).

krwq: btw const is not exactly a field, it gets better compiler optimizations

krwq:
Thanks @Rdlprmpf!

PR #1801, opened 2022-02-20
retry when epoll_wait fails with EINTR

### LINKED ISSUE #971: UnhandledException in WasEventDetected method
On my test system the method _WasEventDetected_ from the _SysFsDriver.Linux.cs_ throws an Exception because the _Interop.epoll_wait_ method fails. While I haven't found the reason yet, how am I supposed to catch this exception? Since this code is executed in a new internal Thread, it will jump to my unhandled exception handler and kills my application.

tibel:
Fixes #971

Implement retry handling for `epoll_wait` in `SysFsDriver.WasEventDetected` similar to `LibGpiodDriverEventHandler` implementation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1801)

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
@@ -467,7 +467,7 @@ namespace System.Device.Gpio.Drivers
                         continue;
                     }
 
-                    throw new IOException("Error while waiting for pin interrupts.");
+                    throw new IOException($"Error while waiting for pin interrupts. (ErrorCode={errorCode})");
                 }
 
                 if (waitResult > 0)
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
@@ -6,6 +6,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Threading;
 
 namespace System.Device.Gpio.Drivers
@@ -15,6 +16,8 @@ namespace System.Device.Gpio.Drivers
     /// </summary>
     public class SysFsDriver : UnixDriver
     {
+        private const int ERROR_CODE_EINTR = 4; // Interrupted system call
+
         private const string GpioBasePath = "/sys/class/gpio";
         private const string GpioChip = "gpiochip";
         private const string GpioLabel = "/label";
@@ -457,6 +460,13 @@ namespace System.Device.Gpio.Drivers
                 int waitResult = Interop.epoll_wait(pollFileDescriptor, out epoll_event events, 1, PollingTimeout);
                 if (waitResult == -1)
                 {
+                    var errorCode = Marshal.GetLastWin32Error();
+                    if (errorCode == ERROR_CODE_EINTR)
+                    {
+                        // ignore Interrupted system call error and retry
+                        continue;
+                    }
+
                     throw new IOException("Error while waiting for pin interrupts.");
                 }
 
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
@@ -445,7 +445,15 @@ namespace System.Device.Gpio.Drivers
             }
 
             // Ignore first time because it will always return the current state.
-            Interop.epoll_wait(pollFileDescriptor, out _, 1, 0);
+            while (Interop.epoll_wait(pollFileDescriptor, out _, 1, 0) == -1)
+            {
+                var errorCode = Marshal.GetLastWin32Error();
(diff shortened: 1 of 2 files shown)


src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
```diff
@@ -457,6 +460,13 @@ private unsafe bool WasEventDetected(int pollFileDescriptor, int valueFileDescri
...
                 if (waitResult == -1)
                 {
+                    var errorCode = Marshal.GetLastWin32Error();
+                    if (errorCode == ERROR_CODE_EINTR)
+                    {
+                        // ignore Interrupted system call error and retry
+                        continue;
+                    }
+
                     throw new IOException("Error while waiting for pin interrupts.");
```
pgrawehr, line 470: Please include the error code in the message.

pgrawehr (approved):
LGTM

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
tibel: @pgrawehr Do you mean the comment or exception message?

I didn't want to change the exception message in this pull request.
What about using `ExceptionHelper` for the exception (also on other places in `SysFsDriver`)?
Personally I prefer to handle that in a separate PR to keep this one simple.

tibel:
Any chance this gets merged and released soon?

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
```diff
@@ -457,6 +460,13 @@ private unsafe bool WasEventDetected(int pollFileDescriptor, int valueFileDescri
                 int waitResult = Interop.epoll_wait(pollFileDescriptor, out epoll_event events, 1, PollingTimeout);
                 if (waitResult == -1)
                 {
+                    var errorCode = Marshal.GetLastWin32Error();
```
krwq, line 463: we also have `epoll_wait` on line 448 in the same file, would be good to also retry there

krwq (approved):
LGTM after fixing the other instance

krwq:
@tibel we will bring the subject of creating minor release during our next triage meeting. My rough guess we will want at minimum https://github.com/dotnet/iot/pull/1789 merged and will likely wait for this one to be merged as well. Both PRs are really close to be merged.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
tibel: I was not sure how to handle that one as the result is ignored.
```c#
// Ignore first time because it will always return the current state.
Interop.epoll_wait(pollFileDescriptor, out _, 1, 0);
```

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
krwq, line 470: might be a good idea to add error code in this message as well for easier debugging

tibel: My idea is to change it to use `ExceptionHelper` in a follow-up PR.

krwq: I see @pgrawehr already asked for that. I'd prefer we changed it in this PR, it's not much different concern. I'd just add append something like ` (ErrorCode={ERROR_CODE_EINTR})` and change string to interpolated. We can create another issue to have more uniform way of throwing but it's better to improve it now as we're touching this code

krwq: if you're planning to send a follow up then that's fine too but just want to make sure we actually do that - usually it's simpler to just do it in a single PR, also we can do the temporary fix now and follow up better fix in case you forget to send follow-up 😄 (sorry but happens to everyone for various reasons)

tibel: Added error code to message as requested.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
krwq: add variable and while loop, just retry on EINTR but ignore errors as we're not sure how to handle them in there

krwq (approved):

joperezr (approved):

joperezr:
Thanks for the fix @tibel we will discuss tomorrow about our next minor release during our triage meeting and will let you know when that will be so you can expect news about that tomorrow.

src/System.Device.Gpio/System/Device/Gpio/Drivers/SysFsDriver.cs
krwq: honestly I'm not sure so good to have just single thing fixed in here.

PR #1808, opened 2022-02-25
Added support for new Raspberry Pi 4 board revision b03115

### LINKED ISSUE #1798: New Revision b03115 of Raspberry Pi not identified correctly
In priciple this is the same issue as in #1270 that was fixed with #1376 again.
I have a new revision that is not correctly detected as Rasperry Pi 4 and hence the same problem again.

**Hint:** I've posted the information below already as a comment at #1376, but @pgrawehr [asked ](https://github.com/dotnet/iot/pull/1376#issuecomment-1044515742)me to open a new issue here.

**_Details:_**
When looking at _/proc/cpuinfo_ I get the following output (cut down to the relevant part, important parts highlighted as **bold text**):
> Hardware        : BCM2711
> Revision        : **b03115**
> Serial          : 10000000aeb9cbea
> Model           : Raspberry Pi 4 Model B Rev **1.5**

This matches with the data that calling the _pinout_ tool retrieves (again cut down):
> Revision           : **b03115**
> SoC                : BCM2711
> RAM                : 2GB

The "funny" part here is, that even the [offical documentation](https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#new-style-revision-codes-in-use) does not list this board revision (but it states, that "_this list is not exhaustive_"...)

A quick fix for this issue could be to extend [RaspberryBoardInfo.cs, line 166](https://github.com/dotnet/iot/blob/main/src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardInfo.cs#L166)
`0x3111 or 0x3112 or 0x3114 or 0x3115 => Model.RaspberryPi4,`

But I'd like to question, if the approach to hardcode those regularly updated/supplemented numbers really is a good way? Shouldn't there better be an alternative (read: extensible) way, that let's one influence the board detection mechanism? Maybe some kind of delegate or by providing a kind of option pattern?

As @pgrawehr  mentioned, the current way is the only reliable way to detect the board type, I want to clarify the concern I've written above: 
I do not question the mechanism in general. I rather question the _implementation_. Resolving the board's revision number into the correct board type is basically a dictionary lookup. Currently this is a hard coded switch expression and I doubt this is a good way ragrding the "moving target" of revisions that are beeing released from time to time.
Maybe there is no better way to do it at the moment, then I'm fine with it - but I wanted to set you thinking about it... In my eyes there should at least be some kind of extensibility mechanism for that.

Mr-Bart-Simpson:
Fixes #1798 

The newly discovered Raspberry Pi 4 board revision b03115 is now supported by the board detection logic.

There is still an ongoing discussion at #1798 on how such new revisions should be handled in the future or how the logic should be changed / opened for extensibility.
However, this (trivial) patch fixes the issue for the moment (i.e. makes such new boards useable) until the aforementioned discussion leads to a greater change.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1808)

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardInfo.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/RaspberryBoardInfo.cs
@@ -163,7 +163,7 @@ namespace System.Device.Gpio
             0x20D3 => Model.RaspberryPi3BPlus,
             0x20E0 => Model.RaspberryPi3APlus,
             0x20A0 or 0x2100 => Model.RaspberryPiComputeModule3,
-            0x3111 or 0x3112 or 0x3114 => Model.RaspberryPi4,
+            0x3111 or 0x3112 or 0x3114 or 0x3115 => Model.RaspberryPi4,
             0x3140 => Model.RaspberryPiComputeModule4,
             0x3130 => Model.RaspberryPi400,
             _ => Model.Unknown,


pgrawehr (approved):

krwq (approved):

krwq:
Thanks @Mr-Bart-Simpson! I'll suggest to create a new issue for the long-term solution and if you have any suggestions or finding that would be super welcome in that discussion

PR #1811, opened 2022-02-27
Update UnitsNet to V4.123

pgrawehr:
This is not yet the major version change, but still an update shouldn't hurt.

### DIFF
--- a/eng/Versions.external.props
+++ b/eng/Versions.external.props
@@ -1,7 +1,7 @@
 <Project>
   <!-- These references to third-party libraries are included in all projects except System.Device.Gpio -->
   <ItemGroup Condition="'$(MSBuildProjectName)' != 'System.Device.Gpio'">
-    <PackageReference Include="UnitsNet" Version="4.77.0" />
+    <PackageReference Include="UnitsNet" Version="4.123.0" />
     <PackageReference Include="SixLabors.ImageSharp" Version="1.0.2" />
   </ItemGroup>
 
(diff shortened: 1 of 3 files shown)


NuGet.config
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <clear />
```
joperezr, line 4: We shouldn't do this. We want to make sure that only internal feeds are used for package restore. Instead, we should push the UnitesNet package into our internal feeds.

joperezr:
While I'm not opposed to this change, is there a specific bug fix or feature we are interested in from the new version? Updating a dependency adds risk (however minimal) and it also makes it so that it could conflict for consuming projects which also reference a different version of the dependency themselves, so in general when taking updates, I think we should at least consider this before we blindly update to each new version. That said, if there is a bug fix or feature we want from this new version, then I'm all for it.

NuGet.config
pgrawehr: The new build servers don't complain, so....

I can't upload there, so I need to do this to get the build to run. But I agree, it's not really nice.

pgrawehr:
> That said, if there is a bug fix or feature we want from this new version, then I'm all for it.

Initially, I thought so. But the change I was expecting is not there, and I need to clarify why first.

joperezr:
[Triage] Closing this one for now and will wait for the new version with the fix we are waiting for.

Issue #1832, opened 2022-03-10
Introduce a Serial Port class


raffaeler:
The initial discussion started [here](https://github.com/dotnet/iot/issues/1797).

## Background and motivation
The `System.IO.Ports.SerialPort` class has been widely adopted in the .NET world to support serial port communication. Since it was first added in the .NET Framework, its implementation is quite old and does not support many of the new language and runtime improvements.

As opposed to many other classes, it is very difficult validating the behavior of the serial port class because it strictly depends on the serial port driver and the connected hardware. For this reason it is preferable not to touch the existing implementation which could cause serious breaking changes. Instead, it is better providing a brand new implementation exposing a modern API and flexible API.

Anyway, the lower layer code (interacting with the serial hardware) is precious and already working cross-platform, therefore it is desirable to reuse it as much as possible.

This is not yet an API proposal that will follow right after collecting all the basic requirements that will shape the new implementation.

## Requirements
This is a list of basic requirements that will guide shaping the final API. In the triage team we expect to see this list growing and discussions will follow to include or not them in the final list.

### Modernization
- Use tasks in place of the Begin/End pattern (used in `BeginRead`, `EndRead`, `BeginWrite`, `EndWrite`)
- Use `Span<byte>` in place of `byte[]` in the API parameters
- Possibly exposing a parameterless constructor to evaluate the opportunity to use Dependency Injection
- Being minimal and let the user easily add an "application layer" to support various communication patterns. Examples:
  - Pure text communication (considering UTF8, UTF16 and endianess variants). Evaluate whether exposing text oriented reads using extension methods.
  - Fixed size (binary) packet oriented communication (w/o header)
  - Dynamic size (binary) packet oriented communication (where the header contains the packet size)

### Features
* Expose the installed COM port list by name including the low level name (e.g. COM10) and also the PnP name (e.g. "Silicon Labs Dual CP210x...")
* Thread-affinity is important to communicate over the serial port. In certain scenarios, the asynchronous communication is interleaved with operations that needs to be fulfilled in a single thread. The current implementation has an `EventLoop` class that could be wrapped to this purpose. It would be even more interesting to reuse an implementation for a cross-platform message-pump (if available) in the runtime (@stephentoub suggestions?)
* Provide the ability to bitbang the pins of the serial port like CTS/RTS/DSR/DTR/DCD/RING
* Provide `DisposeAsync`
[...]

PR #1834, opened 2022-03-17
Feature serial port

raffaeler:
This is the initial work to create a new SerialPort class as for https://github.com/dotnet/iot/issues/1832.
The initial phase is the following:
- Project structure
- Basic definitions (enums, classes)
- Messages definition
- Interop (Platform Invoke)

More to come.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1834)


src/System.Device.Ports/System.Device.Ports.SerialPort/SerialData.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// In the Windows implementation, this is EV_RXCHAR.
+        /// </summary>
+        Chars = 0x01,
+
+        /// <summary>
+        /// The end of file character was received and placed in the input buffer.
+        /// In the Windows implementation, this is EV_RXFLAG.
+        /// </summary>
+        Eof = 0x02
+    }
```
Ellerbach, line 22: Something to think about is a nothing of WatchChar we have introduced in nanoFramework: https://github.com/nanoframework/System.IO.Ports/blob/develop/System.IO.Ports/SerialData.cs.

See usage here: https://github.com/nanoframework/System.IO.Ports#case-of-watchchar

src/System.Device.Ports/System.Device.Ports.SerialPort/Handshake.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// The RTS line is set by the transmitter. The receiver
+        /// should assert the CTS to let the trasmitter begin.
+        /// </summary>
+        RequestToSend,
+
+        /// <summary>
+        /// Both the Request-to-Send (RTS) hardware control and
+        /// the XON/XOFF software controls are used
+        /// </summary>
+        RequestToSendXOnXOff
```
pgrawehr, line 40: This enum should be reviewed. `RequestToSendXOnXOff` is probably never used (I can't think of a reason way you would use hw flow control and sw flow control at the same time), but instead `RtsCts` handshake (bidirectional full hardware flow control) should be added.

src/System.Device.Ports/System.Device.Ports.SerialPort/Parity.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace System.Device.Ports.SerialPort
+{
+    /// <summary>
+    /// Specifies the parity bit for a SerialPort object.
+    /// </summary>
+    public enum Parity
+    {
+        /// <summary>
+        /// No parity check occurs.
```
pgrawehr, line 12: This description is wrong. Right would be: No parity bit is sent or expected.

src/System.Device.Ports/System.Device.Ports.SerialPort/Handshake.cs
raffaeler: Totally agree, I also have never seen that mix

src/System.Device.Ports/System.Device.Ports.SerialPort/Parity.cs
raffaeler: Good catch, as I said many descriptions from the official docs are inaccurate.

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialData.cs
raffaeler: I see, this is substantially the same principle used in Win32 for watching a specific character.
Why did you rename the `Eof=0x02` to `WatchChar=0x02` in `SerialData` give they have the same meaning?
If I understand well this enum can be left as-is for compat reasons. Then we can add the `WatchChar` property to specify what is the character that will trigger the event.
Did I miss something?

src/System.Device.Ports/System.Device.Ports.SerialPort/System.Device.Ports.SerialPort.csproj
```diff
@@ -0,0 +1,57 @@
...
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(NETCoreSdkRuntimeIdentifier.StartsWith('win'))">$(TargetFrameworks);net6.0-windows10.0.17763.0</TargetFrameworks>
+    <LangVersion>9</LangVersion>
+    <Nullable>enable</Nullable>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <IsPackable>true</IsPackable>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+  </PropertyGroup>
+
+  <ItemGroup>
```
joperezr, line 13: In order to keep it simple, let's for now start without all of this ItemGroup definitions, shouldn't be needed at least for now, so should be fine to remove all of this.

src/System.Device.Ports/System.Device.Ports.SerialPort/System.Device.Ports.SerialPort.csproj
```diff
@@ -0,0 +1,57 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(NETCoreSdkRuntimeIdentifier.StartsWith('win'))">$(TargetFrameworks);net6.0-windows10.0.17763.0</TargetFrameworks>
```
joperezr, line 5: Let's also for now remove the windows-specific build and try to for now just work on a RID-less configuration. We can later add a windows vs linux specific build later when we need to.

src/System.Device.Ports/System.Device.Ports.SerialPort/System.Device.Ports.SerialPort.csproj
```diff
@@ -0,0 +1,57 @@
...
+    <LangVersion>9</LangVersion>
```
joperezr, line 6: NIT: Do you actually need this LangVersion or did you just copy this? If you don't need any features that you are not getting by default then let's just remove this for now.

src/System.Device.Ports/System.Device.Ports.SerialPort/System.Device.Ports.SerialPort.csproj
```diff
@@ -0,0 +1,57 @@
...
+    <None Remove="winmd\**" />
+    <Compile Remove="Interop.Unix.DefaultPathBufferSize.cs" />
+    <None Include="buildTransitive\net5.0\System.Device.Ports.targets" Pack="true" PackagePath="\buildTransitive\net5.0" />
+    <None Include="Interop\Unix\Interop.Unix.DefaultPathBufferSize.cs" />
+    <PackageReference Include="Microsoft.DotNet.GenAPI" Version="$(MicrosoftDotNetGenApiPackageVersion)">
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
joperezr, line 25: Do you need these dependencies? I suppose System.Memory you will for sure to get Span, but not sure about S.T.T.E

src/System.Device.Ports/System.Device.Ports.SerialPort/System.Device.Ports.SerialPort.csproj
```diff
@@ -0,0 +1,57 @@
...
+    <PackageReference Include="System.Memory" Version="$(SystemMemoryPackageVersion)" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="$(SystemThreadingTasksExtensionsPackageVersion)" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' != 'net6.0'">
```
joperezr, line 30: Let's also clean the following two ItemGroups and only add them when we need to.

src/System.Device.Ports/System.Device.Ports.SerialPort/System.Device.Ports.SerialPort.csproj
```diff
@@ -0,0 +1,57 @@
...
+    <!--Remove the Linux-specific code-->
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <!--Remove Windows Code-->
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Update="Resources\Strings.Designer.cs">
```
joperezr, line 39: Why do we need a Strings.Designer? I would imagine you should just need the resx file bellow (without metadata) for all of our strings. My advise here is to do what a traditional project would do, so perhaps even re-add the strings.resx from Visual Studio which should add the right modifications into this .csproj.

src/System.Device.Ports/System.Device.Ports.SerialPort/buildTransitive/net5.0/System.Device.Ports.targets
```diff
@@ -0,0 +1,10 @@
+<Project>
```
joperezr, line 1: Let's remove this file for now and only re-add it when we need to.

src/System.Device.Ports/System.Device.Ports.SerialPort/ISR.cs
```diff
@@ -0,0 +1,139 @@
...
+
+using System.Device.Ports.SerialPort.Resources;
+using System.Resources;
+
+namespace System.Device.Ports.SerialPort
+{
+    /// <summary>
+    /// Error Messages
+    /// </summary>
+    internal class ISR
```
joperezr, line 12: We shouldn't need to create our own handling of SR for the resources. We should instead just use the traditional resource manager APIs from the code as any normal library.

src/System.Device.Ports/System.Device.Ports.SerialPort/SafeSerialDeviceHandle.Unix.cs
```diff
@@ -0,0 +1,53 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Device.Ports.SerialPort.Resources;
+using System.Diagnostics;
+using System.IO;
+using System.Net.Sockets;
+using System.Runtime.InteropServices;
+
+using Microsoft.Win32.SafeHandles;
```
joperezr, line 11: NIT: extra space.

src/System.Device.Ports/System.Device.Ports.SerialPort/InternalResources.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.ComponentModel;
+using System.Device.Ports.SerialPort.Resources;
+using System.IO;
+
+using Marshal = System.Runtime.InteropServices.Marshal;
```
joperezr, line 8: NIT: why do we need to use aliases?

src/System.Device.Ports/System.Device.Ports.SerialPort/Interop/Unix/Interop.Unix.Libraries.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Device.Ports.SerialPort
```
joperezr, line 4: As discussed during our call last week, I think it is better to just remove all of the PInvoke calls for now since I think it is best to start from the top (API design) to the bottom since I believe we will end up finding that our implementation will look very different to the one in System.IO.Ports given our API shape will also be different.

src/System.Device.Ports/System.Device.Ports.SerialPort/Resources/Strings.Designer.cs
```diff
@@ -0,0 +1,459 @@
...
+    ///   A strongly-typed resource class, for looking up localized strings, etc.
+    /// </summary>
+    // This class was auto-generated by the StronglyTypedResourceBuilder
+    // class via a tool like ResGen or Visual Studio.
+    // To add or remove a member, edit your .ResX file then rerun ResGen
+    // with the /str option, or rebuild your VS project.
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "17.0.0.0")]
+    [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
+    [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
+    internal class Strings {
```
joperezr, line 25: As mentioned earlier, we shouldn't need to do this. We should just be using ResourceManager APIs like any other application would.

src/System.Device.Ports/System.Device.Ports.SerialPort.Tests/System.Device.Ports.SerialPort.Tests.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <LangVersion>9</LangVersion>
+    <IsPackable>false</IsPackable>
+    <!-- Keep quiet about duplicate package references from props files -->
+    <NoWarn>$(NoWarn);NETSDK1023</NoWarn>
```
joperezr, line 8: Is this actually needed?

joperezr:
Thanks for starting to work on this @raffaeler and sorry for the delayed review. I added some comments but feel free to only address what you think you should for this first iteration. I'm fine if you decide to address some of those changes in a follow up PR.

src/System.Device.Ports/System.Device.Ports.SerialPort/System.Device.Ports.SerialPort.csproj
raffaeler: I just added a resx to the project. It probably used the VS template from Winform or similar.
I now copied the file, removed it from VS and re-added. Let's see if that's better.

src/System.Device.Ports/System.Device.Ports.SerialPort/Handshake.cs
raffaeler: We will have to futher discuss what handshakes we want to support because I am reading an old book on the serial port and there are plenty variations.

src/System.Device.Ports/System.Device.Ports.SerialPort/InternalResources.cs
raffaeler: I just got the code from the original SerialPort class, no particular reason.

joperezr (approved):
This looks good to me as a starting point. We should probably consider naming the library System.Device.Ports instead of System.Device.Ports.SerialPort, but no need to block the PR for that as we can continue to make progress and discuss naming and structure later in API Review.

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialData.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace System.Device.Ports.SerialPort
+{
+    /// <summary>
+    /// Specifies the type of character that was received on the serial port of the SerialPort object.
+    /// </summary>
+    public enum SerialData
+    {
+        /// <summary>
+        /// A character was received and placed in the input buffer.
+        /// In the Windows implementation, this is EV_RXCHAR.
```
krwq, line 13: I'd leave platform implementation details out of here

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialPort.Settings.cs
```diff
@@ -0,0 +1,728 @@
...
+        {
+            get => _breakState;
+            set
+            {
+                /*
+                if (value == _breakState)
+                {
+                    return;
+                }
+                */
```
krwq, line 208: can you write a comment why this is commented out?

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialPort.Settings.cs
```diff
@@ -0,0 +1,728 @@
...
+        protected internal abstract void SetBreakState(bool breakState);
+
+        /// <summary>
+        /// Gets the number of bytes of data in the receive buffer.
+        /// </summary>
+        public int BytesToRead
+        {
+            get
+            {
+                if (!_isOpen)
```
krwq, line 228: should we have similar check everywhere?

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialPort.Settings.cs
```diff
@@ -0,0 +1,728 @@
...
+        /// <summary>
+        /// Gets or sets a value that enables the Data Terminal Ready (DTR) signal during serial communication.
+        /// </summary>
+        /// <param name="value">true to enable Data Terminal Ready (DTR); otherwise, false. The default is false.</param>
+        protected internal abstract void SetDtrEnable(bool value);
+
+        /// <summary>
+        /// Gets or sets the byte encoding for pre- and post-transmission conversion of text.
+        /// </summary>
+        public Encoding Encoding
```
krwq, line 383: I'd prefer we didn't depend on encoding (i.e. remove all string related helper methods)

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialData.cs
raffaeler: Agree. I just ported the class in the project to resolve other issues in the first place.

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialPort.Settings.cs
raffaeler: I am going to add this comment:
```
In the old implementation SetBreakState is set every time using platform specific code for this reason, this code is commented but I left it here because the old behavior is still to be confirmed.
```

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialPort.Settings.cs
raffaeler: I have to see as soon as something starts working.
I am trying to port the same behavior of the `SerialPort` even if the code is different.

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialPort.Settings.cs
raffaeler: As I wrote in my comment after the last push, I am willing to remove any read/write method from the SerialPort class.
Instead, I plan to add a sort of "plugin" (based on an interface) that allows to customize the typology of reads/writes.
For example Laurent asked to have separate streams for the read and write channels.

PR #1841, opened 2022-03-31
Improve performance for large SPI buffer transfers

pgrawehr:
- Support waiting for SPI write commands. When using fast data rates, a buffer overrun
will otherwise occur
- Write SPI data in 7-bit encoding format (requires firmata update)
- Make flow control parameter controllable from outside
- Various stability fixes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1841)

### DIFF
--- a/src/devices/Arduino/ArduinoBoard.cs
+++ b/src/devices/Arduino/ArduinoBoard.cs
@@ -63,9 +63,26 @@ namespace Iot.Device.Arduino
         /// The device is initialized when the first command is sent. The constructor always succeeds.
         /// </remarks>
         /// <param name="serialPortStream">A stream to an Arduino/Firmata device</param>
+        /// <param name="usesHardwareFlowControl">True to indicate that the stream supports hardware flow control (can be a serial port
+        /// with RTS/CTS handshake or a network stream where the protocol already supports flow control)</param>
+        public ArduinoBoard(Stream serialPortStream, bool usesHardwareFlowControl)
+        {
+            _dataStream = serialPortStream ?? throw new ArgumentNullException(nameof(serialPortStream));
+            StreamUsesHardwareFlowControl = usesHardwareFlowControl;
+            _logger = this.GetCurrentClassLogger();
+        }
+
+        /// <summary>
+        /// Creates an instance of an Ardino board connection using the given stream (typically from a serial port)
+        /// </summary>
+        /// <remarks>
+        /// The device is initialized when the first command is sent. The constructor always succeeds.
+        /// </remarks>
+        /// <param name="serialPortStream">A stream to an Arduino/Firmata device</param>
         public ArduinoBoard(Stream serialPortStream)
         {
             _dataStream = serialPortStream ?? throw new ArgumentNullException(nameof(serialPortStream));
+            StreamUsesHardwareFlowControl = false;
             _logger = this.GetCurrentClassLogger();
         }
 
@@ -80,6 +97,7 @@ namespace Iot.Device.Arduino
         {
             _dataStream = null;
             _serialPort = new SerialPort(portName, baudRate);
+            StreamUsesHardwareFlowControl = false; // Would need to configure the serial port externally for this to work
             _logger = this.GetCurrentClassLogger();
         }
 
@@ -96,7 +114,6 @@ namespace Iot.Device.Arduino
         public bool StreamUsesHardwareFlowControl
         {
             get;
-            set;
         }
 
         /// <summary>
@@ -191,7 +208,6 @@ namespace Iot.Device.Arduino
 #endif
             out ArduinoBoard? board)
         {
-            board = null;
             try
             {
                 Stream networkStream;
@@ -214,8 +230,7 @@ namespace Iot.Device.Arduino
                     networkStream = new NetworkStream(socket, true);
                 }
 
-                board = new ArduinoBoard(networkStream);
-                board.StreamUsesHardwareFlowControl = true;
+                board = new ArduinoBoard(networkStream, true);
                 if (!(board.FirmataVersion > new Version(1, 0)))
                 {
                     // Actually not expecting to get here (but the above will throw a SocketException if the remote end is not there)
@@ -226,6 +241,7 @@ namespace Iot.Device.Arduino
             }
             catch (SocketException)
             {
+                board = null;
                 return false;
             }
         }
--- a/src/devices/Arduino/ArduinoBoard.cs
+++ b/src/devices/Arduino/ArduinoBoard.cs
@@ -80,10 +80,8 @@ namespace Iot.Device.Arduino
         /// </remarks>
         /// <param name="serialPortStream">A stream to an Arduino/Firmata device</param>
         public ArduinoBoard(Stream serialPortStream)
+        : this(serialPortStream, false)
         {
-            _dataStream = serialPortStream ?? throw new ArgumentNullException(nameof(serialPortStream));
-            StreamUsesHardwareFlowControl = false;
-            _logger = this.GetCurrentClassLogger();
         }
 
         /// <summary>
@@ -797,6 +795,12 @@ namespace Iot.Device.Arduino
                 {
                     // Ignore
                 }
+            }
+
+            if (_firmata != null)
+            {
+                // Can end the next possible moment (otherwise might just throw a bunch of warnings before actually terminating anyway)
+                _firmata.InputThreadShouldExit = true;
             }
 
             _isDisposed = true;
--- a/src/devices/Arduino/ArduinoBoard.cs
+++ b/src/devices/Arduino/ArduinoBoard.cs
@@ -89,6 +89,17 @@ namespace Iot.Device.Arduino
         protected ILogger Logger => _logger;
 
         /// <summary>
+        /// Set this to true if the underlying stream uses some kind of hardware or low-level flow control (RTS/CTS for
+        /// a serial port, or a TCP socket). Setting this to true may improve performance on bulk transfers (such as
+        /// large SPI blocks) but can result in buffer overflows if flow control is not working. Default: false
+        /// </summary>
+        public bool StreamUsesHardwareFlowControl
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
         /// The list of supported pin modes.
         /// This list can be extended by adding special modes using <see cref="AddCommandHandler{T}"/>.
         /// </summary>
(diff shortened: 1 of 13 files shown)


src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -88,6 +88,17 @@ public ArduinoBoard(string portName, int baudRate)
         /// </summary>
         protected ILogger Logger => _logger;
 
+        /// <summary>
+        /// Set this to true if the underlying stream uses some kind of hardware or low-level flow control (RTS/CTS for
+        /// a serial port, or a TCP socket). Setting this to true may improve performance on bulk transfers (such as
+        /// large SPI blocks) but can result in buffer overflows if flow control is not working. Default: false
+        /// </summary>
+        public bool StreamUsesHardwareFlowControl
```
joperezr, line 96: Should this be readonly and set in the constructor though? Or do we want to allow it being mutable?

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -159,17 +170,52 @@ public IReadOnlyList<SupportedMode> KnownModes
...
+        /// <param name="useAutoReconnect">True to use an auto-reconnecting stream. Helpful when using an unreliable connection.</param>
+        /// <param name="board">Returns the board if successful</param>
+        /// <returns>True on success, false otherwise</returns>
+        public static bool TryConnectToNetworkedBoard(IPAddress boardAddress, int port, bool useAutoReconnect,
+#if NET5_0_OR_GREATER
+            [NotNullWhen(true)]
 #endif
             out ArduinoBoard? board)
         {
             board = null;
```
joperezr, line 165: Nit: it is usually better to set the out parameter right next to the return statement so that assignment is clearer when reading the code.

src/devices/Arduino/ArduinoGpioControllerDriver.cs
```diff
@@ -14,26 +14,28 @@ namespace Iot.Device.Arduino
...
         private readonly IReadOnlyCollection<SupportedPinConfiguration> _supportedPinConfigurations;
         private readonly Dictionary<int, CallbackContainer> _callbackContainers;
         private readonly ConcurrentDictionary<int, PinMode> _pinModes;
         private readonly object _callbackContainersLock;
         private readonly AutoResetEvent _waitForEventResetEvent;
         private readonly ILogger _logger;
+        private readonly ConcurrentDictionary<int, PinValue?> _outputPinValues;
 
-        internal ArduinoGpioControllerDriver(ArduinoBoard arduinoBoard, IReadOnlyCollection<SupportedPinConfiguration> supportedPinConfigurations)
+        internal ArduinoGpioControllerDriver(FirmataDevice device, IReadOnlyCollection<SupportedPinConfiguration> supportedPinConfigurations)
```
joperezr, line 26: If we do end up moving to the new SDK, we could have these parameters use the new `!!` operator for parameter null checks

src/devices/Arduino/BlockingUnorderedCollection.cs
```diff
@@ -0,0 +1,108 @@
...
+{
+    /// <summary>
+    /// Represents a collection that removes objects based on a certain pattern
+    /// </summary>
+    internal class BlockingUnorderedCollection<T>
+    {
+        private object _lock = new object();
+        private List<T?> _container = new List<T?>();
+
+        public BlockingUnorderedCollection()
```
joperezr, line 23: Nit: no need to explicitly declare a parameterless constructor

src/devices/Arduino/BlockingUnorderedCollection.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Iot.Device.Arduino
+{
+    /// <summary>
+    /// Represents a collection that removes objects based on a certain pattern
+    /// </summary>
+    internal class BlockingUnorderedCollection<T>
```
joperezr, line 18: Nit: Blocking -> Concurrent?

src/devices/Arduino/ReconnectingNetworkStream.cs
```diff
@@ -0,0 +1,192 @@
...
+        public ReconnectingNetworkStream(Func<Stream> connect)
+        : this(null, connect)
+        {
+        }
+
+        public override bool CanRead
+        {
+            get
+            {
+                return SafeExecute(x => x.CanRead, false);
```
joperezr, line 35: Nit: all of these can be rewritten to:
`public override bool Foo => SafeExecute(...)`

src/devices/Arduino/ReconnectingNetworkStream.cs
```diff
@@ -0,0 +1,192 @@
...
+                lock (_lock)
+                {
+                    try
+                    {
+                        _streamImplementation = _reconnect();
+                    }
+                    catch (IOException)
+                    {
+                        // Ignore
+                    }
```
joperezr, line 113: Can we add mire verbose comments why it is ok to ignore?

joperezr (commented):
Did a quick pass and added some minor comments, will take another look today.

src/devices/Arduino/ArduinoGpioControllerDriver.cs
pgrawehr: Sounds like a nice feature. This here is internal, so it should never be used incorrectly.

src/devices/Arduino/BlockingUnorderedCollection.cs
pgrawehr: The runtime already has `BlockingCollection<T>` and `ConcurrentBag<T>`, but not the combination I need here. I'll rename to `BlockingConcurrentBag<T>` as this is what this represents.

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: Correct, this should be defined in the ctor. Changing it at runtime will likely cause problems.

joperezr:
>         /// <param name="serialPortStream">A stream to an Arduino/Firmata device</param>

NIT: this constructor can now be rewritten to instead just call the new constructor and passes false for false for `usesHardwareFlowControl`

---
Refers to: src/devices/Arduino/ArduinoBoard.cs:81 in 01b6991. [](commit_id = 01b69913153d3b2b84b74f6b12e7690c14f56347, deletion_comment = False)

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -691,6 +772,16 @@ public void SetAnalogPinSamplingInterval(TimeSpan timeSpan)
             Firmata.SetAnalogInputSamplingInterval(timeSpan);
         }
 
+        /// <summary>
+        /// Performs a software reset of the Arduino firmware
+        /// </summary>
+        public void SoftwareReset()
+        {
+            Initialize();
+            Firmata.SendSoftwareReset();
```
joperezr, line 781: Should all these APIs be checking if firmata is disposed?

src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -754,5 +845,43 @@ internal void DisableSpi()
...
             }
         }
+
+        /// <summary>
+        /// Pings the device, to get an estimate about the round-trip time.
+        /// With some Wifi setups, the round trip time may be significantly higher than desired.
+        /// </summary>
+        /// <param name="number">The number of pings to send</param>
+        /// <returns>The list of reply times. Contains a negative value for lost packets</returns>
+        public List<TimeSpan> Ping(int number)
```
joperezr, line 855: I wonder if this method should be async and take a cancellation Token, but that whould probably require _firmata.QueryFrimwareVersion to also be async.

src/devices/Arduino/ArduinoGpioControllerDriver.cs
```diff
@@ -14,26 +14,28 @@ namespace Iot.Device.Arduino
...
     {
-        private readonly ArduinoBoard _arduinoBoard;
+        private readonly FirmataDevice _device;
         private readonly IReadOnlyCollection<SupportedPinConfiguration> _supportedPinConfigurations;
         private readonly Dictionary<int, CallbackContainer> _callbackContainers;
         private readonly ConcurrentDictionary<int, PinMode> _pinModes;
         private readonly object _callbackContainersLock;
         private readonly AutoResetEvent _waitForEventResetEvent;
         private readonly ILogger _logger;
+        private readonly ConcurrentDictionary<int, PinValue?> _outputPinValues;
```
joperezr, line 24: I'm not following why do you need this ConcurrentDictionary. not sure why you want to track the values, but this will have a significant perf hit because ConcurrentDictionary uses locks which of course slow perf overall.

pgrawehr: This is used to check whether a given pin is already set to a certain value. Accessing this dictionary should be significantly faster than what it uses to send a command.

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: The call to `Initialize()` performs that test.

src/devices/Arduino/ArduinoBoard.cs
pgrawehr: That's a bit difficult, because this command is not really designed for async calls, as it uses no sequence counting. So if you where to run multiple async pings in parallel the result would be undefined. This is mostly for testing the speed of a connection, you should not use it e.g. for keep-alive scenarios.

joperezr (approved):
Minor things, but otherwise looks good. Feel free to merge once you have addressed or replied ro my comments.

src/devices/Arduino/BlockingConcurrentBag.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Iot.Device.Arduino
+{
+    /// <summary>
+    /// Represents a collection that removes objects based on a certain pattern
+    /// </summary>
+    internal class BlockingConcurrentBag<T>
```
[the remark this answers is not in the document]
pgrawehr: I did not find a built-in type that fits, therefore I had to write my own. If ever we want this to be public, we would clearly need a better specification and an (at least informal) API review.

src/devices/Arduino/ExtendedCommandHandler.cs
```diff
@@ -181,6 +195,30 @@ protected virtual void OnSysexData(ReplyType type, byte[] data)
...
+        /// This method is called to check whether the reply is a valid ACK/NOACK for the given command sequence.
+        /// Can be used to avoid accepting something as command reply that is completely unrelated (such as an asynchronous callback).
+        /// In different words, this should return false if the given reply is not something that is an answer to a synchronous command.
+        /// </summary>
+        /// <param name="sequence">The sequence that was sent</param>
+        /// <param name="reply">The reply</param>
+        /// <returns>True if this reply matches the sequence. True is the default, for backwards compatibility</returns>
+        protected virtual bool IsMatchingAck(FirmataCommandSequence sequence, byte[] reply)
+        {
+            return true;
```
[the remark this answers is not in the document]
pgrawehr: I'll maybe learn it sometime. I'm just not a fan of that syntax.

src/devices/Arduino/ExtendedCommandHandler.cs
```diff
@@ -205,6 +243,16 @@ protected virtual void Dispose(bool disposing)
...
         }
 
+        /// <summary>
+        /// Called by the infrastructure when the parser reports an error or information message.
+        /// The default implementation does nothing.
+        /// </summary>
+        /// <param name="message">The message text</param>
+        /// <param name="exception">The exception observed (may be null)</param>
+        protected internal virtual void OnErrorMessage(string message, Exception? exception)
+        {
```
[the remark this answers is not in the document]
pgrawehr: This class is abstract. This method by default shall do nothing, but can be overwritten as needed.

src/devices/Arduino/tests/BlockingConcurrentBagTest.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Iot.Device.Arduino;
+using Xunit;
+
+namespace Iot.Device.Arduino.Tests
+{
+    public class BlockingConcurrentBagTest
+    {
```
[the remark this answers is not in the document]
pgrawehr: Possibly. It tests the use cases that are currently relevant.

PR #1844, opened 2022-04-07
Fix ca2208 errors

Jimmys20:
This PR fixes all [CA2208](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2208) violations and enables a rule so that they won't get violated again in the future.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1844)


pgrawehr:
Thanks for the submission. I can't seem to find where you enabled that rule, though. Can you make sure the rule is enabled so that any future builds would fail if it gets violated again?

src/devices/SoftwareSpi/SoftwareSpi.cs
```diff
@@ -145,7 +145,7 @@ public override void TransferFullDuplex(ReadOnlySpan<byte> dataToWrite, Span<byt
         {
             if (dataToWrite.Length != dataToRead.Length)
             {
-                throw new ArgumentException(nameof(dataToWrite));
+                throw new ArgumentException(null, nameof(dataToWrite));
```
pgrawehr, line 148: I'm not sure passing null here is a good idea. Wouldn't this possibly lead to a NullReferenceException when handling the error?

Jimmys20: This is what visual studio code fix does by default. I could replace null with actual message describing what caused the exception. 

ca2208-codefix_null_msg

pgrawehr: That would be the ideal case, for sure.

Jimmys20: Replaced null parameters with error messages.

Jimmys20:
I added the rule in editorconfig with severity level error.

.editorconfig
```diff
@@ -135,6 +135,9 @@ csharp_space_between_method_declaration_parameter_list_parentheses = false
 csharp_space_between_parentheses = false
 csharp_space_between_square_brackets = false
 
+# CA2208: Instantiate argument exceptions correctly
+dotnet_diagnostic.CA2208.severity = error
```
pgrawehr, line 139: Sorry for not being precise enough. This only affects VS, but not the build on the command line. Add the warning to the file `src/StyleCop.Analyzers.ruleset` instead. Add a line with `<Rule Id="CA2208" Action="Error" />`.

Jimmys20: No worries, should be ok now.

pgrawehr (approved):

krwq:
@Jimmys20 there seem to be some merge conflict here but otherwise it can be merged

PR #1860, opened 2022-05-12
[SerialPort] Draft implementation (Windows and portions of Linux code)

raffaeler:
Windows specific implementation. More will follow, please do not merge yet.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1860)


src/System.Device.Ports/System.Device.Ports.SerialPort/SerialPort.Settings.cs
```diff
@@ -29,6 +29,28 @@ public partial class SerialPort
...
+
+        /// <summary>
+        /// The default value for XOFF
+        /// </summary>
+        public const byte DefaultXOFFChar = (byte)19;
+
+        /// <summary>
+        /// The default value for EOF
+        /// </summary>
+        public const byte EOFChar = (byte)26;
```
krwq, line 52: consider `enum SerialPortCharacters : byte`

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialPort.Settings.cs
```diff
@@ -65,12 +86,16 @@ public partial class SerialPort
         /// The name of the serial port whose default value is platform dependent
         /// and set to a proper default name in the derived platform-specific classes
         /// </summary>
-        protected string _portName = String.Empty;
+        protected string _portName = string.Empty;
```
krwq, line 89: protected members should follow public API rules:
- `PortName`
- property rather than field

pgrawehr: and could be public IMHO

src/System.Device.Ports/System.Device.Ports.SerialPort/xlf/Strings.de.xlf
```diff
@@ -2,11 +2,36 @@
...
+      <trans-unit id="ArgumentOutOfRange_Bounds_Lower_Upper">
+        <source>Argument must be between {0} and {1}</source>
+        <target state="new">Argument must be between {0} and {1}</target>
+        <note />
+      </trans-unit>
       <trans-unit id="ArgumentOutOfRange_Enum">
         <source>Enum value was out of legal range</source>
         <target state="new">Enum value was out of legal range</target>
         <note />
       </trans-unit>
```
pgrawehr, line 29: Did you incorrectly copy the sources of these files? There are no translations in here, neither in the new strings (which would be expected) nor in the existing ones. 

Maybe cross-check at the end. If we don't have any translations, we shouldn't generate the satellite assemblies either.

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialPort.Settings.cs
raffaeler: There is already a public property named `PortName`.
The reason for the field being protected is because the default name for the Windows and Linux implementation is different.
In this case I can use the public setter of the property (I am doing it). But there are other fields that need to be accessed directly from the platform-specific implementation to bypass the validation (see `_rtsEnable`).

src/System.Device.Ports/System.Device.Ports.SerialPort/SerialPort.Settings.cs
raffaeler: Done

joperezr:
@raffaeler still working on this one? Is it ready to go?

raffaeler:
Hi @joperezr , I am still working on it.
I already have a new commit to push, but I am not sure if I will have issues with rebasing after the changes to the TFMs

joperezr:
[Triage]: This is ready for last review and merge. Looks like the tests are not running in CI (since they are passing on Linux which is not yet implemented). I'll take a look and make sure they run and pass, and then merge this PR.

PR #1861, opened 2022-05-15
Fix an access violation when leaking LibGpiodDriver

### LINKED ISSUE #1849: Segfault in finalization with LibGpiodDriver
**Describe the bug**

Leaking a GpioController that uses LibGpiodDriver and has a pin open can segfault due to the pin being released after the chip has been destroyed. This appears to be a race condition and does not occur always.

GpioController creates a SafeChipHandle for the chip and a SafeLineHandle for the pin/line. Destroying the chip invalidates the line, since libgpiod frees the line's memory when destroying the chip. I suspect that the reason for the crash is that during finalization, the runtime does not know that the chip should be destroyed last (or that the line should not be freed if the chip has already been freed).

**Steps to reproduce**

The code below reproduces the segfault within a few tries.

```csharp
GpioController controller1 = new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver(4));
controller1.OpenPin(10, PinMode.Output);
GpioController controller2 = new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver(4));
controller2.OpenPin(11, PinMode.Output);
GpioController controller3 = new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver(4));
controller3.OpenPin(12, PinMode.Output);
GpioController controller4 = new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver(4));
controller4.OpenPin(13, PinMode.Output);
GpioController controller5 = new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver(4));
controller5.OpenPin(14, PinMode.Output);

for (int i = 0; i < 10; i++)
{
    GC.Collect();
    GpioController controller6 = new GpioController(PinNumberingScheme.Logical, new LibGpiodDriver(4));
    controller6.OpenPin(15, PinMode.Output);
    controller6.ClosePin(15);
    controller6.Dispose();
    GC.Collect();
    Thread.Sleep(20);
}
```

**Expected behavior**

The GPIO chip and line/pin are destroyed/released correctly. No segfault.

**Actual behavior**

Segfault.

See the attached log, generated with GDB. Unfortunately I do not have debug symbols for dotnet itself, but the stack trace for libgpiod is visible.

To summarize, what happens is:

1. The pins 10, 11, 12, 13, and 14 are opened (`gpiod_chip_get_line`)
2. The leaked chip `controller2` is destroyed, which also frees the line struct at `0xaaaaaae0dbb0`.
3. The chip `controller6` is destroyed. This is just noise.
4. The line struct at `0xaaaaaae0dbb0` from `controller2` is released again, causing a crash because the memory has already been reused and overwritten.

[crash.txt](https://github.com/dotnet/iot/files/8487735/crash.txt)

**Versions used**

`dotnet --info` on the build machine (self-contained build for linux-arm64)

```plain
 Version:   5.0.407
 Commit:    1a1785612e

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  20.04
 OS Platform: Linux
 RID:         ubuntu.20.04-x64
 Base Path:   /home/user/dotnet/sdk/5.0.407/

Host (useful for support):
  Version: 5.0.16
  Commit:  1016676966

.NET SDKs installed:
  5.0.407 [/home/user/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 5.0.16 [/home/user/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 5.0.16 [/home/user/dotnet/shared/Microsoft.NETCore.App]

```

`System.Device.Gpio` version 2.1.0

pgrawehr:
Fixes #1849 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1861)


src/System.Device.Gpio.Tests/LibGpiodDriverTests.cs
```diff
@@ -71,5 +71,37 @@ static void PinChanged(object sender, PinValueChangedEventArgs args)
                 gc.UnregisterCallbackForPinValueChangedEvent(InputPin, PinChanged);
             }
         }
+
+        /// <summary>
+        /// Ensure leaking instances of the driver doesn't cause a segfault
+        /// See #1849 for a description of this test case
```
krwq, line 77:
```suggestion
        /// Regression test for https://github.com/dotnet/iot/issues/1849
```

src/System.Device.Gpio/Interop/Unix/SafeLineHandle.cs
```diff
@@ -19,8 +19,9 @@ public SafeLineHandle()
 
         protected override bool ReleaseHandle()
         {
-            // Contrary to intuition, this does not invalidate the handle (see comment on declaration)
-            Interop.libgpiod.gpiod_line_release(handle);
+            // Do not actually free anything, the handle is not ours (but rather a memory pointer into an
+            // array owned by the chip)
+            handle = IntPtr.Zero;
```
krwq, line 24: I'm curious here. The doc for `gpiod_line_release` claims something else than this comment. Not implementing ReleaseHandle kills the purpose of using SafeHandle to me. IMO:
- We should keep ReleaseHandle as it was
- Remove ReleaseLock and replace usage with Dispose
- SetPinMode IMO seems to me misimplemented as it calls to ReleaseLock and IMO it should re-acquire the line after that (assuming it's not possible to change direction without first doing so) and replace entry in the dictionary

am I missing something here?

src/System.Device.Gpio.Tests/LibGpiodDriverTests.cs
```diff
@@ -71,5 +71,37 @@ static void PinChanged(object sender, PinValueChangedEventArgs args)
...
+        /// </summary>
+        [Fact]
+        public void LeakingDriverDoesNotCrash()
```
krwq, line 80: also looking at implementation seems SetPinMode is calling ReleaseLock. Can you also add a test where we change mode several times and see if anything breaks?

src/System.Device.Gpio/Interop/Unix/SafeLineHandle.cs
krwq: FYI AFAIK SafeHandle's dispose makes sure we don't call ReleaseHandle multiple times and zeroes out the handle

src/System.Device.Gpio.Tests/LibGpiodDriverTests.cs
pgrawehr: That test exists and was added when the current implementation was done. Without that call to ReleaseLock, SetPinMode fails when done repeatedly.

src/System.Device.Gpio/Interop/Unix/SafeLineHandle.cs
pgrawehr: This whole thing is quite delicate, as we see now. I'm trying to change as little as possible, since everything else works fine. Closing and reopening the line may have other side effects (such as glitches on the line and slower performance), so I don't think we should do that if it can be avoided. 

Unfortunately, the "fix" now broke a bunch of other tests, possibly due to leaked references.

pgrawehr: Now implemented differently - SaveLineHandle no longer derives from SafeHandle, because it must be released when the chip (in our case the LibGpiodController instance) is released. The whole problem exists because if the chip is released (by the finalizer) before an open line is, the crash occurs as that automatically  frees the memory for all lines, invalidating the handle.

joperezr:
@pgrawehr what is the state of this?

pgrawehr:
Well, we decided that we want a "real" fix for the problem, but so far nobody had a better idea. The test code reproduces the problem and the suggested changes fix that, so I think it would be better than nothing.

PR #1866, opened 2022-05-21
Add encoder support to Seesaw device

FloGs:
Enables suppport for encoder functions in Seesaw device.

Adds the following methods to Seesaw class:
- GetEncoderPosition / SetEncoderPosition (Module Encoder / Function EncoderPosition)
- GetEncoderDelta (Module Encoder / Function EncoderDelta)
- EnableEncoderInterrupt (Module Encoder / EncoderIntenset)
- DisableEncoderInterrupt (Module Encoder / EncoderIntenclr)

Adds sample project to get (rotary) encoder positions.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1866)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1866) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1866) before we can accept your contribution.<br/><br/>:x: FloGs [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1866)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1866) it.</sub>

src/devices/Seesaw/samples/Seesaw.Sample.Encoder.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Device.I2c;
+using Iot.Device.Seesaw;
+
+using Seesaw seesawDevice = new Seesaw(I2cDevice.Create(new I2cConnectionSettings(1, 0x36)));
+
+seesawDevice.EnableEncoderInterrupt();
```
pgrawehr, line 10: I suppose the interrupt requires a GPIO pin? You are not using any GPIO pins in this example. Can you add an example to use the interrupt?

FloGs: For the implementation of the encoder functions I followed the documentation of Adafruit's Circuit Python. The python method  [enable_encoder_interrupt](https://docs.circuitpython.org/projects/seesaw/en/latest/api.html#adafruit_seesaw.seesaw.Seesaw.enable_encoder_interrupt) does not use any interrupt either.

I extended the seesaw class with the encoder functionality with the goal to talk to the [Adafruit I2C QT Rotary Encoder](https://learn.adafruit.com/adafruit-i2c-qt-rotary-encoder). It uses seesaw and returns changes of the rotary encoder. The interrupt to be enabled on the board is board specific and therefore implemented directly by the board (i.e. by the Adafruit I2C QT Rotary Encoder Board). With the enable_encoder_interrupt function I can enable it without knowing exactly which GPIO is used.

pgrawehr: @FloGs Sorry for being unclear. The `EnableEncoderInterrupt()` enables the interrupt _output_ of the device. That pin is fixed on the breakout (labeled "INT") and therefore doesn't need an argument. 

To use it though, you need a GpioController on the host (e.g. your Raspberry Pi) and an input pin configured for interrupt triggering there. So instead of polling for changes on the encoder (which is what your example currently does) you wait until the interrupt is raised and only then read the new value. That saves a lot of CPU time.

FloGs: @pgrawehr Thank you for the explanation! It seems that I had not fully understood what the EnableEncoderInterrupt is used for. I was so focused on Seesaw and its functions and therefore just implemented the functions that are also used in the Python library. I connected the board via STEMMA QT cable, and therefore overlooked the INT pin.

After your explanation I now understand what the method is used for. Of course it makes much more sense that way :sweat_smile: 

I will adapt my test setup accordingly and then modify the sample project to use GpioController.

FloGs: @pgrawehr The example project now uses interrupts both for the detection of the rotary encoder position and for the detection of a push switch change.

pgrawehr (approved):
Looks good, thanks!

src/devices/Seesaw/SeesawEncoder.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace Iot.Device.Seesaw
+{
+    public partial class Seesaw : IDisposable
+    {
+        /// <summary>
+        /// Read the current position of the encoder
+        /// </summary>
+        /// <param name="encoder">Which encoder to use, defaults to 0</param>
+        /// <returns>The encoder position as a 32 bit signed integer</returns>
+        /// <exception cref="InvalidOperationException"></exception>
```
Ellerbach, line 16: Would be nice to put a description here. Also, for all comments, best to finish with a dot.

src/devices/Seesaw/SeesawEncoder.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            return BinaryPrimitives.ReadInt32BigEndian(Read(SeesawModule.Encoder, SeesawFunction.EncoderPosition + encoder, 4, 8000));
+        }
+
+        /// <summary>
+        /// Set the current position of the encoder
+        /// </summary>
+        /// <param name="position">Encoder position</param>
+        /// <param name="encoder">Which encoder to use, defaults to 0</param>
+        /// <exception cref="InvalidOperationException"></exception>
```
Ellerbach, line 32: Same here, please add the description

src/devices/Seesaw/SeesawEncoder.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            Write(SeesawModule.Encoder, SeesawFunction.EncoderPosition + encoder, buffer);
+        }
+
+        /// <summary>
+        /// The change in encoder position since it was last read
+        /// </summary>
+        /// <param name="encoder">Which encoder to use, defaults to 0</param>
+        /// <returns>The encoder change as a 32 bit signed integer</returns>
+        /// <exception cref="InvalidOperationException"></exception>
```
Ellerbach, line 51: And here and everywhere else :-)

Ellerbach (approved):
Just missing few comments, otherwise, looks good to me! Thanks for the addition!

FloGs:
@Ellerbach Thanks for pointing this out. I have added the missing exception comments and finished all comments with a dot.

Issue #1869, opened 2022-06-02
RFID reader sample can't read card


fadyanwar:
When running the sample found at https://github.com/dotnet/iot/blob/2cc943168d4b20339a4c3046a5c8a952b273afcb/src/devices/Mfrc522/samples/Program.cs

using the below shown command, no result was shown when card was presented to RFID reader

dotnet run --project iot/src/devices/Mfrc522/samples/Mfrc522.Sample.csproj

.NET SDK (reflecting any global.json):
 Version:   6.0.300
 Commit:    8473146e7d

Runtime Environment:
 OS Name:     debian
 OS Version:  11
 OS Platform: Linux
 RID:         debian.11-arm64
 Base Path:   /home/pi/.dotnet/sdk/6.0.300/

Host (useful for support):
  Version: 6.0.5
  Commit:  70ae3df4a6

.NET SDKs installed:
  6.0.300 [/home/pi/.dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.5 [/home/pi/.dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.5 [/home/pi/.dotnet/shared/Microsoft.NETCore.App]

To install additional .NET runtimes or SDKs:
  https://aka.ms/dotnet-download

joperezr:
Thanks for logging the issue @fadyanwar. Can you please share details on which specific card reader are you trying to use? cc: @Ellerbach who wrote most of this code and who has tested it.

joperezr:
Also did you try debugging it by any chance? If so, did you see if it was able to successfully connect to the device? From what i'm reading in that example there are several ways to connect to the reader, I'm curious as to which one you picked and to see if you have tried the others.

fadyanwar:
Hello @joperezr The card reader is RF522, the card itself is a ISO14443A and I'm connecting over SPI. I did test using a python script using the same configuration and it's working fine. I've tried to take out segments of code to figure out which line is not working, I think it's the loop part found here https://github.com/dotnet/iot/blob/2cc943168d4b20339a4c3046a5c8a952b273afcb/src/devices/Mfrc522/samples/Program.cs#L113

Find below full output

Hello MFRC522!
Do you want to use the MFRC522 on a device like a Raspberry Pi or through a FT4222?
  1. Raspberry Pi or equivalent
  2. FT4222
1
How do you want to connect MFRC5222?
  1. SPI
  2. I2C
[...]

PR #1871, opened 2022-06-04
Fixing an inverted timeout condition

fadyanwar:
Contributes to: https://github.com/dotnet/iot/issues/1869

Card reading timeout should happen when dtTimeout < DateTime.Now not the other way around, otherwise ListenToCardIso14443TypeA(out Data106kbpsTypeA card, TimeSpan timeout) will always return false. i.e. since dtTimeout is in the future it will be always greater than DateTime.Now

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1871)

### DIFF
--- a/src/devices/Mfrc522/Mfrc522.cs
+++ b/src/devices/Mfrc522/Mfrc522.cs
@@ -254,7 +254,7 @@ namespace Iot.Device.Mfrc522
                     break;
                 }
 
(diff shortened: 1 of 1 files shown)


Ellerbach (approved):
Thanks.

PR #1881, opened 2022-07-10
Update file-scoped namespaces and unused usings

shaggygi:
Cleaned up a few areas in main GPIO and test projects.
- Updated to C# 10.
- Updated to file-scoped namespaces.
- Removed unused usings.

**NOTE**: This might appear to be a large update, but mainly because of the namespace format changes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1881)

### DIFF
--- a/src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerSoftwareTests.cs
@@ -1,11 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
-using System.Collections.Generic;
-using System.Text;
 using System.Threading;
-using Iot.Device.Mcp25xxx.Register.MessageReceive;
 using Moq;
 using Xunit;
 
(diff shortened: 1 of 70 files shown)


joperezr:
Good to hear from you @shaggygi, thanks for the PR, I will take a look at it very soon but at least looks great with a quick initial skim. Thanks for modernizing the repo. I will prioritize this PR due to obvious reasons, as it spans many files and we don't want to cause merge conflicts.

Question: did you use a specific analyzer (or analyzers) to make these changes?

joperezr (approved):
Thanks @shaggygi

shaggygi:
@joperezr good to hear from you, as well.  I had some downtime and was looking over the repo.  I used the unused using analyzer, but manually did the file-scoped namespaces.  I can't find where I read it, but there is supposed to be one in the works.  I hope so as that would be a huge time savings.  😄

joperezr:
I mainly asked because it would be good to do this repo-wide. Even if there is no analyzer for this, looks like it should be easy enough to write one.

shaggygi:
Yeah, but not sure about projects that are not set to 10, preview, or latest.  Not sure if it would pick those up or not.

shaggygi:
Well, look what was released in VS17P3 today.  I'll give it a try on the binding projects this weekend.

image

PR #1883, opened 2022-07-12
Updated sample projects to DefaultSampleTfms.

shaggygi:
Updated random sample projects to DefaultSampleTfms.

There were a couple `netstandard2.0` but left as wasn't sure the intent on usage.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1883)

### DIFF
--- a/src/devices/Amg88xx/tests/Amg88xx.Tests.csproj
+++ b/src/devices/Amg88xx/tests/Amg88xx.Tests.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>$(DefaultTestsTfms)</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
     <IsPackable>false</IsPackable>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
   </PropertyGroup>
--- a/src/devices/Amg88xx/tests/Amg88xx.Tests.csproj
+++ b/src/devices/Amg88xx/tests/Amg88xx.Tests.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>$(DefaultSampleTfms)</TargetFramework>
     <IsPackable>false</IsPackable>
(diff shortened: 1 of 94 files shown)


raffaeler (approved):
Looks good to me

src/devices/Amg88xx/tests/Amg88xx.Tests.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
```
joperezr, line 3: Can we undo the change for tests projects? For samples I'm fine with the update, it makes sense, but for tests we are effectively losing coverage of older nca 3.1 TFMs which we do want to have tests for. In fact, we are specifically restoring the 3.1 runtime so that we can run these tests and ensure they work in 3.1:

https://github.com/dotnet/iot/blob/4eb8d2b8595b86d113ee6f96990e344fdea9e6df/global.json#L6

shaggygi: > Can we undo the change for tests projects?

Sure, but could we add a `DefaultTestsTfms` property in the `Directory.Build.props` file?

```xml
<DefaultTestsTfms>netcoreapp3.1</DefaultTestsTfms>
```

So the Amg88xx.Tests.csproj mentioned above would look like...

```xml
  <PropertyGroup>
    <TargetFramework>$(DefaultTestsTfms)</TargetFramework>
```

Output
```console
Amg88xx.Tests -> C:\iot\src\devices\Amg88xx\tests\bin\Debug\netcoreapp3.1\Amg88xx.Tests.dll
```

joperezr: Sure, that sounds good to me, just for test projects that are not testing netcoreapp3.1 we should leave those targeting what they currently have. Essentially the idea would be for this to be a non-impacting change, just a form of centralization of properties

shaggygi: There are not many test projects and only a couple that were not netcoreapp3.1.  I moved test back to original and removed the `DefaultTestsTfms`.  Having build errors now.  Looking into.

joperezr (approved):

PR #1890, opened 2022-07-15
Fix a possible GC hole

### LINKED ISSUE #1886: `Iot.Device.Common.ValueArray<T>` is not safe and contains GC holes
**Describe the bug**
I want to preface this by stating that I'm not a consumer of these APIs, I just happened to come across this while perusing [source.dot.net](https://source.dot.net/).

The [`ValueArray<T>`](https://github.com/dotnet/iot/blob/1e0f186df8ee1c9f8fca44446bb6426f972d4ee3/src/devices/Common/Iot/Device/Common/ValueArray.cs) type introduced in https://github.com/dotnet/iot/pull/1701 stores a sequence of 8 elements, and allows consumers to retrieve a `Span<T>` wrapping them. However, the method for which it does this is not safe and causes the GC to lose track of the references, meaning that the object they belong to may be eligible for garbage collection while the `Span<T>` is in use.

For reference, this is the current implementation of `ValueArray<T>.AsSpan`:
```cs
public unsafe Span<T> AsSpan()
{
    return new Span<T>(Unsafe.AsPointer(ref _e0), _count);
}
```
And this would be an implementation that retains GC tracking (compatible with .NET Core/Standard 2.1 and newer):
```cs
public Span<T> AsSpan()
{
    return MemoryMarshal.CreateSpan(ref _e0, _count);
}
```
Unfortunately, `MemoryMarshal.CreateSpan` is not available on .NET Standard 2.0. A close equivalent to this API used to exist that would have made it possible to create `Span<T>` instances for `ValueArray<T>`s stored within classes safely (see https://github.com/dotnet/runtime/issues/27690 and https://github.com/dotnet/runtime/issues/24562).

As the `ValueArray<T>` type is stored within `class`es, changing it to a `ref struct` is also not an option, meaning that there is no safe way to implement `AsSpan` on `ValueArray<T>` under .NET Standard 2.0.

pgrawehr:
Fixes #1886 (partially) 

Will need another update when the
new [RefEscapes] attribute is available, to make this
finally really safe.

Note: This removes the public AsSpan() method for
netstandard2.0, because it cannot be made safe at all there.

CC: @DaZombieKiller, @raffaeler, @joperezr 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1890)


src/devices/Common/Iot/Device/Common/ValueArray.cs
```diff
@@ -91,22 +91,53 @@ public int Count
...
+        /// Returns a span pointing to this instance
+        /// </summary>
+        /// <returns>A span representing this instance</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+#if !BUILDING_IOT_DEVICE_BINDINGS
+        public
+#else
+        internal
+#endif
+        unsafe Span<T> AsSpan()
```
raffaeler, line 121: I would personally leave out `unsafe` at all, even for the netstandard2.0 case, by using `Memory<T>`.

pgrawehr: That would make the API very different in use for netstandard2.0 and require `#ifs` in all places where it's used, which is ugly.

raffaeler: This is why I suggested in the triage to entirely replace the `Span` with `Memory` for all the cases.
Consumers can just take the `Memory<T>` (which own the memory) from which they can obtain the Span as they always did.
IMO safety should come first, otherwise you leave to the consumers the responsibility to avoid the bug, which is not necessarily easy to find.

src/devices/Common/tests/ValueArrayTests.cs
```diff
@@ -40,6 +40,7 @@ public void AssigningDefaultClearsArray()
             Assert.Equal(0, _myArray[2]);
         }
 
+#if NET5_0_OR_GREATER
```
krwq, line 43: honestly I think we should just not build tests against lower versions

krwq: (nothing to change in this PR)

src/devices/Common/Iot/Device/Common/ValueArray.cs
```diff
@@ -91,22 +91,53 @@ public int Count
         /// <inheritdoc />
         public bool IsReadOnly => false;
 
+#if NET5_0_OR_GREATER
```
krwq, line 94: we should skim through usages in our repo to check if we do it correctly

DaZombieKiller:
Since the `Unsafe.AsPointer` approach is still used at all on .NET Standard 2.0, the GC holes are still present on that target. Even internal uses of the API are unsafe, as unfortunately using `Unsafe.AsPointer` to create a `Span<T>` to managed memory is always a GC hole (unless it is pinned). The `Span<T>` is also still indirectly exposed publicly as it gets passed to public APIs such as `I2cDevice.Read`, which may be complicated to solve.

There are a few alternatives however. On .NET Standard 2.0 you can change `ValueArray<T>` to just store a `T[]`, which would allow you to safely return a `Span<T>`, at the cost of no longer really being a "value array" (it would also change the semantics of copying a `ValueArray<T>`, so perhaps it is not really a desirable change after all).

Another option is to go through and change any APIs relying on `AsSpan` to access items manually. For the indexer, this would actually be more efficient across the board _in addition_ to being safe on NS2.0, as it won't need to create a new span on every access:
```cs
public T this[int index]
{
    get
    {
        if ((uint)index >= (uint)_count)
            ThrowHelper.ThrowIndexOutOfRangeException();

        return Unsafe.Add(ref _e0, index);
    }

    set
    {
        if ((uint)index >= (uint)_count)
            ThrowHelper.ThrowIndexOutOfRangeException();

        Unsafe.Add(ref _e0, index) = value;
    }
}
```

src/devices/Common/Iot/Device/Common/ValueArray.cs
john-h-k: `Memory<T>` is much slower (removes almost all benefit of the `AsSpan` method as memory -> span is quite costly) and is also more unsafe than `Span<T>` (it can suffer from cross-thread tearing and other issues)

DaZombieKiller:
> The `Span<T>` is also still indirectly exposed publicly as it gets passed to public APIs such as `I2cDevice.Read`, which may be complicated to solve.

Thinking about this problem further, you may be able to solve it by exposing `GetPinnableReference` on `ValueArray<T>` like so:
```cs
[EditorBrowsable(EditorBrowsableState.Never)]
public unsafe ref T GetPinnableReference()
{
    fixed (byte* p = &Unsafe.As<T, byte>(ref _e0))
    {
        return ref Unsafe.AsRef<T>(p);
    }
}
```
This will allow you to pin a `ValueArray<T>` with a `fixed` block and get a pointer to its first element. From there you can create a `Span<T>` that is valid within that block. Any usages such as this on NS2.0:
```cs
_i2cDevice.Read(_readBuff.AsSpan());
```
Would now become:
```cs
fixed (byte* pReadBuff = _readBuff)
{
    _i2cDevice.Read(_readBuff.AsSpan());
}
```

src/devices/Common/tests/ValueArrayTests.cs
pgrawehr: There are some test assemblies that explicitly build against .NET4.8, just to test the netstandard2.0 behaviors, where they are different.

src/devices/Common/Iot/Device/Common/ValueArray.cs
```diff
@@ -91,22 +91,53 @@ public int Count
...
         /// <summary>
         /// Returns a span pointing to this instance
         /// </summary>
         /// <returns>A span representing this instance</returns>
+        /// <remarks>
+        /// Due to a shortcoming of the Garbage collector, it is the caller's responsibility
```
Sergio0694, line 100: Nit: not really a shortcoming of the GC, rather the language (no escape semantics yet).

src/devices/Common/Iot/Device/Common/ValueArray.cs
```diff
@@ -277,11 +290,43 @@ public void RemoveAt(int index)
         {
             get
             {
-                return AsSpan()[index];
+                if ((uint)index >= (uint)_count)
+                {
+                    throw new IndexOutOfRangeException();
```
DaZombieKiller, line 265: nit: this can use a throw helper to improve the generated code a bit, see [this page](https://docs.microsoft.com/en-us/windows/communitytoolkit/developer-tools/throwhelper#technical-details) for details.

src/devices/Common/Iot/Device/Common/ValueArray.cs
```diff
@@ -211,15 +226,15 @@ public void Insert(int index, T item)
             }
 
             int i = _count - 1;
-            var self = AsSpanFull();
+
+            _count++; // Increase count first, so we can access [i + 1]
```
DaZombieKiller, line 230: You can do `Unsafe.Add(ref _e0, i)` rather than incrementing the count, which will work and prevent potentially exposing the temporary length change to callers in multi-threaded scenarios.

src/devices/Dhtxx/Devices/Dht11.cs
```diff
@@ -24,9 +25,9 @@ public Dht11(int pin, PinNumberingScheme pinNumberingScheme = PinNumberingScheme
         {
         }
 
-        internal override RelativeHumidity GetHumidity(Span<byte> readBuff) => RelativeHumidity.FromPercent(readBuff[0] + readBuff[1] * 0.1);
+        internal override RelativeHumidity GetHumidity(IReadOnlyList<byte> readBuff) => RelativeHumidity.FromPercent(readBuff[0] + readBuff[1] * 0.1);
 
-        internal override Temperature GetTemperature(Span<byte> readBuff)
+        internal override Temperature GetTemperature(IReadOnlyList<byte> readBuff)
```
DaZombieKiller, line 30: Changing these APIs to take `IReadOnlyList<T>` rather than `Span<T>` will cause a boxing allocation every time `ValueArray<T>` is passed in.

As long as it's within a `fixed` block and the `Span<T>` is not used after the block ends, it should be safe to pass in a `Span<T>` created via `Unsafe.AsPointer(ref _e0)`, as the GC will not move or reclaim the pinned memory during that period.

In the absence of `AsSpan`, that can be achieved like so:
```cs
fixed (byte* p = _readBuff)
{
    device.Read(new Span<byte>(p, _readBuff.Count));
}
```

DaZombieKiller:
@pgrawehr
> Sorry this takes a few rounds.

No worries, I'm willing to provide any assistance I can until this reaches a point where the API is safe :)

> For writing the array, a copy (typically from a stackalloc array) is required, but since we're talking about arrays of 5 or 6 bytes, a stack-only copy of that is cheap.

Yup, that sounds perfectly reasonable to me. If it matters, it might be worth testing the performance difference between that copy and a `fixed` block, but I'm not sure if it would matter much.

> What I'm not sure: Is the `ValueArray` boxed when it's converted to one of it's interfaces? If so, converting it to a `IReadOnlyList` or `IEnumerable` would create a new heap object each time, kind of defeating the purpose of all of this...

That is correct. I have provided a solution to that in the review I just posted. The allocation of an `IEnumerator` can actually be avoided too, by making the `struct` enumerator type `public`, as is done on existing collection types: [`List<T>.Enumerator`](https://github.com/dotnet/runtime/blob/c670547c235566ef860490242bb630671baad8b5/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L1102).

src/devices/Common/Iot/Device/Common/ValueArray.cs
pgrawehr: Ok, but I think that should be done in a separate commit over (more or less) the whole repo.

src/devices/Common/Iot/Device/Common/ValueArray.cs
pgrawehr: Good idea, but the class is not thread safe in either case (the other way round, it would probably be possible to insert more than the maximum number of elements in a parallel loop)

DaZombieKiller: That is a fair point. I must have skimmed and not realized that this was code in an `Add`/`Insert` method to begin with.

src/devices/Dhtxx/Devices/Dht11.cs
pgrawehr: Changed the argument type to directly take `ValueArray<byte>`, so there's no boxing conversion required. The size of this is actually the same as that of a `Span<byte>` so the amount of stack data to copy doesn't change.

pgrawehr:
I think it's now safe and fast (for the current use cases). I didn't consider that copying a `ValueArray<byte>` (for parameter passing) is equivalent to copying a `Span<byte>` in terms of performance and memory size.

DaZombieKiller:
Looks good to me!

pgrawehr:
[Triage] The team decided that it's better to dump the type altogether, as its use is limited and risky. The only use in the DHT binding can easily be replaced and performance is not an issue there, as the sensor can anyway only be read at around 3 second intervals.

src/devices/Dhtxx/Devices/Dht11.cs
```diff
@@ -2,7 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Collections.Generic;
```
joperezr, line 5: I suppose we don't need these added namespaces everywhere anymore?

src/devices/Dhtxx/DhtBase.cs
```diff
@@ -20,12 +21,13 @@ namespace Iot.Device.DHTxx
...
     public abstract class DhtBase : IDisposable
     {
+        private const int BufferSize = 5;
         private readonly CommunicationProtocol _protocol;
 
         /// <summary>
         /// Read buffer
         /// </summary>
-        private ValueArray<byte> _readBuff = new ValueArray<byte>(5);
+        private byte[] _readBuff = new byte[BufferSize];
```
joperezr, line 30: Why do you need to initialize it here? Looks to me that everywhere it is used it gets initialized again anyway.

joperezr (approved):
Left a couple of comments but looks good otherwise.

src/devices/Dhtxx/DhtBase.cs
DaZombieKiller: If you set `MinTimeBetweenReads` to a value that will trigger the early-out in `ReadData` before you read from the sensor for the first time, you would get a `null` result handed back unless the buffer is initialized here.

pgrawehr: Yea, the main reason for the initialization here is just to avoid having to make the field nullable.

Issue #1671, opened 2021-09-22
Discussion around the introduction of a Pin class


raffaeler:
Discussion started [here](https://github.com/dotnet/iot/pull/1665#discussion_r714130485) on this code.
```cs
// set up for LED and pin
using GpioController led = new();
led.OpenPin(pin, PinMode.Output);
```

@krwq initially suggested that:
> Feels like Pin class could be useful 

I commented about the requirements that this would trigger to be able to share the pin with other controllers/pin:
- sharing the private _openPins dictionary (making it static?) 
- revising the events logic 

@joperezr said that:
> The original design of the API was very much considering that we wouldn't have a Pin class, so now that we have gone through that route, a lot would have to change if we were to introduce it now

One additional comment from me. What are the implications in using multiple `GpioController` instances? This could give the answer to the issues behind the abstraction needed to implement `Pin`.

/cc @Ellerbach @pgrawehr

pgrawehr:
On the technical perspective: It is possible to create multiple GpioController instances, and as long as these don't use the same pins, there won't be any problems.* But if you use the same pin from different instances, the behavior is more or less undefined. It is highly recommended to use the `Board` class as root of your device instances, as it handles these cases nicely and throws exceptions if you try to reuse pins. 

From an API perspective, adding a method `Pin OpenPin(int pinNumber)` to GpioController would be easy, but as already stated would also require updating many bindings to use this new object type instead of just numbers. While it may look nicer, I doubt it helps that much, since you actually rarely access pins directly. Most of the time, you'd use a binding that does this for you. 

Maybe, to simplify some introductory scenarios, we should introduce a `Led` binding and a `Button` binding? 

. * Note that this applies only to "native" Gpio Pins, i.e. on the Raspi, but not on external devices, such as an Arduino or an FT4222.

Ellerbach:
In .NET nanoFramework, we do have a GpioPin. See https://github.com/nanoframework/System.Device.Gpio/blob/develop/System.Device.Gpio/GpioPin.cs

GpioController is the only one which can open or close pins, ownership is to it.

GpioPin is present for 2 reasons:

* Performances. It's much faster to write or read, change the state, toggle without having to browse a collection all times.
* Compatibility with the previous API which includes a GpioPin (Windows.Devices namespaces).

Now, all the Iot.Device bindings, like for .NET IoT **only** takes a GpioContoller or a null one in any binding constructor. Reason is for the ownership.

Ellerbach:
[...]

PR #1895, opened 2022-07-23
Adding GpioPin

Ellerbach:
Adding GpioPin. Following design discussions here: https://github.com/dotnet/iot/issues/1671

This PR now implements what's been discussed during the API proposal call and is summarized here: https://github.com/dotnet/iot/issues/1920

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1895)


src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -87,7 +87,7 @@ protected virtual int GetLogicalPinNumber(int pinNumber)
     /// The driver attempts to open the pin without changing its mode or value.
     /// </summary>
     /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
-    public void OpenPin(int pinNumber)
+    public GpioPin OpenPin(int pinNumber)
```
pgrawehr, line 90: Unfortunately, I think this is a binary breaking change (because the caller must insert an extra `pop` if he doesn't need the return value).

raffaeler: I second that. Either we kee this change for v3.0 or we can add a different method

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -67,63 +67,59 @@ public GpioController(PinNumberingScheme numberingScheme)
...
+
+    /// <summary>
+    /// Gets the logical pin number in the controller's numbering scheme.
+    /// </summary>
+    /// <param name="pinNumber">The pin number</param>
+    /// <returns>The logical pin number in the controller's numbering scheme.</returns>
+    protected virtual int GetLogicalPinNumber(int pinNumber)
+    {
+        return (NumberingScheme == PinNumberingScheme.Logical) ? pinNumber : _driver.ConvertPinNumberToLogicalNumberingScheme(pinNumber);
+    }
```
raffaeler, line 83: Is this a sort of out-of sync or what? I can't see the difference

joperezr:
I was expecting this to be a draft PR, or a PR against a feature branch. What is the intention of this @Ellerbach? This would have to go through APIReview first before it can get merged.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
joperezr: Agreed, this can't be done. Even for 3.0, it would be really hard to accept this kind of breaking change. We would need a new API for this.

Ellerbach: OK, I'll create an OpenPinEx function then. Thanks for the feedbacks

Ellerbach:
> This would have to go through APIReview first before it can get merged

That's fine. I can mark the PR as Draft then. If you prefer, I can just push the branch directly to the repo so it will be its own feature branch.

krwq:
@joperezr can we do small API review? This API is not part of the framework and we didn't go through API review for other smaller things

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
krwq: I'd be ok with making this tiny breaking change for 3.0, especially given this is source compatible

src/System.Device.Gpio/System/Device/Gpio/GpioPin.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// <value>
+        /// The pin number of the GPIO pin.
+        /// </value>
+        public int PinNumber
+        {
+            get
+            {
+                return _pinNumber;
+            }
+        }
```
krwq, line 33:
```suggestion
        public int PinNumber => _pinNumber;
```

Ellerbach:
If the breaking change is acceptable for 3.0 and well documented, it's fine to wait for that version to be out. In the mean time, it does allow time to align on the design.

joperezr:
I don't think this is a small thing to be honest. This was one of the biggest design decisions made on the original design, and I do think that getting this reviewed by more people would help as most of us are already a bit biased.

src/System.Device.Gpio/System/Device/Gpio/GpioPin.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        /// <summary>
+        /// Toggles the output of the general purpose I/O (GPIO) pin if the pin is configured as an output.
+        /// </summary>
+        public void Toggle() => _controller.Write(_pinNumber, !_controller.Read(_pinNumber));
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            _controller.ClosePin(_pinNumber);
```
krwq, line 112: I'd recommend nulling `_controller` as well in the Dispose

src/System.Device.Gpio/System/Device/Gpio/GpioPin.cs
```diff
@@ -0,0 +1,116 @@
...
+namespace System.Device.Gpio
+{
+    /// <summary>
+    /// Represents a general-purpose I/O (GPIO) pin.
+    /// </summary>
+    public sealed class Gpio​Pin : IDisposable
+    {
+        private readonly int _pinNumber;
+        private readonly GpioController _controller;
+        private bool _disposedValue;
```
krwq, line 13: if you null `controller` you can just null check rather than having this field

src/System.Device.Gpio/System/Device/Gpio/GpioPin.cs
```diff
@@ -0,0 +1,116 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Device.Gpio
+{
+    /// <summary>
+    /// Represents a general-purpose I/O (GPIO) pin.
+    /// </summary>
+    public sealed class Gpio​Pin : IDisposable
```
krwq, line 9: I'm curious about this. If we make this abstract there we could potentially create better implementation for i.e. RaspberryPi3Driver which can directly write to memory rather than going through controller. Leaving as is might be ok but we should consider being able to use this in LedMatrix code which could benefit from that design change

Ellerbach: Why not, we can definitely make it abstract for other usage. Let me think of this. The key question then for the abstract is passing or not the controller and having a default controller based implementation or not

pgrawehr: It would be slightly more expensive, since it would require another virtual call.

krwq: if most derived GpioPin is used directly JIT will devirtualize that and make it equally fast. The only time you'd pay for virtual cost is if you used GpioPin directly. For most usages it won't matter since hardware/OS will limit us but for where it matters it will be faster to do it this way (I think)

joperezr: As part of this proposal we should run some benchmarks comparing old approach vs new.

krwq:
Some notes (+ my after-thoughts) from the offline conversation between me and @Ellerbach today:
- abstract class/interface has benefits of being able to remove all validation code and do some pre-calculations and really save time
- interface can be implemented with struct which has extra benefit of being able to efficiently store list of pins if used directly
- class implementing interface/abstract class when used directly rather than base type it will not have virtual cost (although such usage might be much harder)
- two interesting opposite use cases: LED Matrix (stretch of the perf), LCD character screen (many pins which might need to come from different controllers)
- non-abstract/no-interface can also remove extra validation if i.e. using non-validating driver overload and since we pre-validate that's safe
- pin management makes things much safer so non-validating pin might not always be a good idea
- if we cannot make specialized abstract class work fast enough with i.e. LED Matrix then why bother making it abstract if it doesn't solve largest use case for that?
- LED matrix is currently using registers directly, perhaps improving that interface to work with other controllers than just RPi3 might be a better idea for perf scenarios than Pin class - possibly register should always return a pointer and be part of GpioController/GpioDriver - this way pref-interested parties can use that, for drivers which don't support that it would either throw or return null sentinel
- in general we should do some experimentation here and we're in agreement that low-level concepts are really hard to get right in high level APIs and solving both perf and safety at the same time might not be possible so we should find a good sweet spot between or shift to one side

src/System.Device.Gpio/System/Device/Gpio/GpioPin.cs
Ellerbach: `_controller` is not disposed here, just the pin closed.

src/System.Device.Gpio/System/Device/Gpio/GpioPin.cs
Ellerbach: the controller is not nulled as not disposed at all.

Ellerbach:
Few elements to take into consideration:
* The changes are mainly about removing the `sealed` so it will allow to create other GpioPin based on a driver for example if needed
* From the conversation mentioned up, having an abstract class does not need to be necessary as we will use the controller in almost all cases

src/System.Device.Gpio/System/Device/Gpio/GpioPin.cs
```diff
@@ -0,0 +1,110 @@
...
+                }
+
+                _controller.UnregisterCallbackForPinValueChangedEvent(_pinNumber, value);
+            }
+        }
+
+        /// <summary>
+        /// Toggles the output of the general purpose I/O (GPIO) pin if the pin is configured as an output.
+        /// </summary>
+        public void Toggle() => _controller.Write(_pinNumber, !_controller.Read(_pinNumber));
```
joperezr, line 101: This is probably not a good idea, could cause race conditions. If someone wants to toggle a pin, we should let them call Write and have them deal with keeping track of state.

Ellerbach: Toggling is something that most drivers do low level. In the case of most of our drivers, it's not the case. But if you take FT4222, it can be done. It's also the case in .NET nanoFramework. 
You can always get back the state with Read. And as a remonder, there should not be race conditions as nothing is thread safe in low level hardware. If you want to keep track and do operations in parallel, you should take care of that. So here, there is no risk.

src/System.Device.Gpio/System/Device/Gpio/GpioPin.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        /// <inheritdoc/>
+        public void Dispose()
```
joperezr, line 104: As discussed offline, I don't think this type should be disposable

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -96,6 +96,7 @@ public void OpenPin(int pinNumber)
 
         OpenPinCore(pinNumber);
         _openPins.TryAdd(pinNumber, null);
+        return new GpioPin(pinNumber, this);
```
joperezr, line 99: As discussed, it might be valuable to have the driver initialize the pin instead of the controller, in case a subclassed driver wants to return a subclassed Gpio Pin. The way this is today, a subclass of GpioPin can never be created, since OpenPin is not virtual. We should either:

- Have GpioPin be sealed and leave the rest as is.
- Keep GpioPin unsealed but make it abstract, and also change OpenPinCore's return value to be GpioPin. Also changed GpioDriver's OpenPin methods to return GpioPin.

Ellerbach: I directly pass the driver now and removed the disposal.
I did not touch the driver at all, no need for it to return a GpioPin, the controller keep the control of opening and closing the pins.
I've updated the API design to add the use case of having extender and how pain it is in the current situation vs with a list of GpioPin.

joperezr: Would there ever be a reason for GpioPin to be extended by someone? I kind of thought that there could be cases where we would want to let people extend GpioPin, and if that is the case, then you will have to delegate the task of opening the pins to the driver, or else it will never be possible to create them. For example, imagine I create a Ft422GpioDriver and Ft422GpioPin classes, how can I ever get a Ft422GpioPin object back? This wouldn't be possible even with creating a Ft422GpioController, because OpenPin method that handles opening the pins is not virtual.

If we don't expect people to try to do the scenario I described above, then let's make GpioPin sealed to make our intention clear.

Ellerbach: Something like this should just work:

```csharp
namepace System.Device.Gpio
{
  public class Ft422GpioPin: GpioPin
  {
     public Ft422GpioPin(int pinNumber, GpioDriver driver): base(pinNumber, driver)
     {
        driver.OpenPin(pinNumber); 
     }
  }
}

// Then usage:
var pin = new Ft422GpioPin(2, new Ft422GpioDriver());
```

joperezr: What is the role of the controller in that case? I think what you are proposing is confusing in two reasons:
- GpioPins now get created in two different ways, depending on if you want a normal pin (in which case you create them via a controller) or a special pin (where you do it via ctor passing in the driver)
- ownership is even more complicated now, who disposes the driver? In our case is the controller, but presumably in your case it is the pin?

My suggestion is to have the same consistent usage no matter if you are a custom pin or not. That can be accomplished simply by also changing the driver's OpenPin to return GPioPin, and that way everything is always created via de controller

Ellerbach: > My suggestion is to have the same consistent usage no matter if you are a custom pin or not. That can be accomplished simply by also changing the driver's OpenPin to return GPioPin, and that way everything is always created via de controller

I guess you mean by the driver then rather than the controller ;-) That's possible as well.

Note that the example was just to answer your question on how it will be done. To extend a bit, that can as well add a way to close the pin:

```csharp
namepace System.Device.Gpio
{
  public class Ft422GpioPin: GpioPin, IDisposable
  {
     private bool _shouldDispose;
     public Ft422GpioPin(int pinNumber, GpioDriver driver, bool shouldDispose = false): base(pinNumber, driver)
     {
        driver.OpenPin(pinNumber); 
        _should Dispose = shouldDispose;
     }

    public void ClosePin()
    {
      driver.ClosePin(PinNumber);
    }

    public void Dispose()
    {
        if (_shouldDispose)
        {
            driver.Dispose();
        }
    }
  }
}

// Then usage:
var driver = new Ft422GpioDriver();
var pin1 = new Ft422GpioPin(1, driver);
var pin2 = new Ft422GpioPin(2, driver);
// Do whatever 
// Then clean
pin1.Dispose();
pin2.Dispose();
driver.Dispose();
```

Additional thoughts? :-)

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
```diff
@@ -82,6 +82,12 @@ protected internal virtual void SetPinMode(int pinNumber, PinMode mode, PinValue
     /// <returns>The value of the pin.</returns>
     protected internal abstract PinValue Read(int pinNumber);
 
+    /// <summary>
+    /// Toggle the current value of a pin.
+    /// </summary>
+    /// <param name="pinNumber">The pin number in the driver's logical numbering scheme.</param>
+    protected internal abstract void Toggle(int pinNumber);
```
pgrawehr, line 89: That's a real breaking change (binary as well as source). I think this should have a default implementation instead.

pgrawehr (commented):
Looks good overall. What's not so nice IMHO is that the code to keep the current state of the pin for the Toggle operation is duplicated all over the place. 

Please also add a test case for this functionality. 

I'll have to take a look whether the now valid duplicate open causes some problems in the pin management for the board. I don't expect so, as this should have triggered a test then.

src/System.Device.Gpio.Tests/GpioPinTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            _mockedGpioDriver.CallBase = true;
+        }
+
+        public void Dispose()
+        {
+            _mockedGpioDriver.VerifyAll();
+        }
+
+        [Fact]
+        public void TestOpnPin()
```
raffaeler, line 29: Typo: `TestOpenPin` in place of `TestOpnPin`

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
raffaeler: Why not `PinValue Toggle(int pinNumber)`?
I don't know if the hw implementation provides the value, but in the other cases it is certainly available.
If the hardware driver does not provide the value, we can decide to either return null or read the value and return it. 
I prefer this signature to avoid future breaking changes in case the hw driver will provide the value in the next releases.

raffaeler (commented):
Great work @Ellerbach!

Ellerbach:
> Why not PinValue Toggle(int pinNumber)?

Hardware does not provide. I know that for all current implementation we keep track, so could be provided but in real hardware, it's not.

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
Ellerbach: > That's a real breaking change (binary as well as source). I think this should have a default implementation instead.

Some hardware provide this feature, so does not. We agreeded that the braking change is ok as it then surfaced on the GpioController following the API call. And the best way to handle the change is to move it to the driver because it's the **only** place tracking really the state of the pins.

raffaeler (approved):
At this point I guess we can go on with this :)

raffaeler:
We were discussing in the triage to shape the api as `PinValue? Toggle(int pinNumber)`.
My point was the following:
- In the software implementation returning the value is free because we track the values
- In the hardware implementation providing the value, we return it
- In the hardware implementation **not** providing the value, we can return null (telling that is not available)

@joperezr Do you have an opinion on this?

src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
krwq, line 89: can this be virtual instead of abstract? (by default Write(!Read())) then concerns about breaking change will be adressed and we don't need to repeat ourselves since most of the drivers don't really natively support it?

src/System.Device.Gpio/System/Device/Gpio/Drivers/HummingBoardDriver.cs
```diff
@@ -82,6 +82,9 @@ protected internal override PinMode GetPinMode(int pinNumber)
     /// <inheritdoc/>
     protected internal override PinValue Read(int pinNumber) => _internalDriver.Read(pinNumber);
 
+    /// <inheritdoc/>
+    protected internal override void Toggle(int pinNumber) => _internalDriver.Read(pinNumber);
```
krwq, line 86: Is this correct? I'd prefer we didn't have to override it

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
```diff
@@ -21,6 +22,7 @@ public class LibGpiodDriver : UnixDriver
     private readonly ConcurrentDictionary<int, SafeLineHandle> _pinNumberToSafeLineHandle;
     private readonly ConcurrentDictionary<int, LibGpiodDriverEventHandler> _pinNumberToEventHandler;
     private readonly int _pinCount;
+    private readonly ConcurrentDictionary<int, PinValue> _pinValue;
```
krwq, line 25: I'm looking at this and wondering - perhaps GpioDriver could have virtual `CreateGpioPin` with default implementation - then in here you'd return `LibGpiodPin` and move `CurrentPinValue`, `EventHandler`, `SafeLineHandle` to `LibGpiodPin` - this way you could have just one `ConcurrentDictionary` here and using pin directly would have advantage of not touching ConcurrentDictionary

src/System.Device.Gpio/System/Device/Gpio/Drivers/HummingBoardDriver.cs
krwq: with virtual GpioDriver.Toggle you can just remove this

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -87,15 +89,17 @@ protected virtual int GetLogicalPinNumber(int pinNumber)
...
     {
         if (IsPinOpen(pinNumber))
         {
-            throw new InvalidOperationException($"Pin {pinNumber} is already open.");
+            return _gpioPins[pinNumber];
         }
 
         OpenPinCore(pinNumber);
         _openPins.TryAdd(pinNumber, null);
+        _gpioPins[pinNumber] = new GpioPin(pinNumber, _driver);
```
krwq, line 101: line 100 seems redundant

src/System.Device.Gpio/System/Device/Gpio/Drivers/LibGpiodDriver.cs
krwq: we talked offline and we can do this as a separate change most likely without any API changes

src/System.Device.Gpio/System/Device/Gpio/Drivers/Windows10Driver.cs
```diff
@@ -14,6 +14,7 @@ public class Windows10Driver : GpioDriver
 {
     private static readonly WinGpio.GpioController s_winGpioController = WinGpio.GpioController.GetDefault();
     private readonly Dictionary<int, Windows10DriverPin> _openPins = new Dictionary<int, Windows10DriverPin>();
+    private readonly Dictionary<int, PinValue> _pinValues = new Dictionary<int, PinValue>();
```
krwq, line 17: why does LibGpiodDriver need ConcurrentDictionary while this one is ok with regular Dictionary?

krwq: (it's an existing issue anyway so don't worry about this in this PR)

src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -519,12 +532,14 @@ private void ThrowBadPin(string argument)
         protected override void OpenPin(int pinNumber)
         {
             // No-op
+            _pinValues.TryAdd(pinNumber, PinValue.Low);
```
krwq, line 535: nit: comment above is no longer true

src/devices/Pcx857x/tests/Pcx857xTest.cs
```diff
@@ -161,6 +161,8 @@ protected override PinValue Read(int pinNumber)
                 return PinValue.Low;
             }
 
+            protected override void Toggle(int pinNumber) => Write(pinNumber, !Read(pinNumber));
```
krwq, line 164: can be removed given current virtual implementation

src/devices/Seesaw/SeesawGpioDriver.cs
```diff
@@ -165,9 +169,13 @@ protected override PinValue Read(int pinNumber)
                 throw new InvalidOperationException("Can not read a value from a pin that is not open.");
             }
 
-            return _seesawDevice.ReadGpioDigital((byte)pinNumber) ? PinValue.High : PinValue.Low;
+            _pinValues[pinNumber] = _seesawDevice.ReadGpioDigital((byte)pinNumber) ? PinValue.High : PinValue.Low;
+            return _pinValues[pinNumber];
         }
 
+        /// <inheritdoc/>
+        protected override void Toggle(int pinNumber) => Write(pinNumber, !_pinValues[pinNumber]);
```
krwq, line 177: I'm not sure dictionary actually makes it faster for those low level drivers - reading/writing pin is supposedely fastest operation you can make if you think about it. It might be currently faster this way because of OS but in principal that shouldn't be the case

tools/ArduinoCsCompiler/Hal/ArduinoNativeGpioDriver.cs
```diff
@@ -65,6 +65,8 @@ protected override PinValue Read(int pinNumber)
             return _hardwareLevelAccess.ReadPin(pinNumber);
         }
 
+        protected override void Toggle(int pinNumber) => Write(pinNumber, !Read(pinNumber));
```
krwq, line 68: nit: can be removed

krwq (approved):
LGTM with couple of nits. feel free to ignore or fix separately

PR #1908, opened 2022-08-04
Added TCA9548A device binding

asheesh1996:
Added support for TCA9548A multiplexer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1908)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1908) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1908) before we can accept your contribution.<br/><br/>:x: asheesh1996 [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1908)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1908) it.</sub>

src/devices/Tca9548a/Tca9548A.cs
```diff
@@ -0,0 +1,229 @@
...
+        /// <summary>
+        /// Channel 7 Byte (2^7 = 128)
+        /// </summary>
+        Channel7 = 0x80
+    }
+
+    /// <summary>
+    /// Mux States
+    /// </summary>
+    public enum MUXState : byte
```
pgrawehr, line 61: Rename this to `MuxState`. Type names in public Apis should use strong PascalCasing.

src/devices/Tca9548a/Tca9548A.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            return channels;
+        }
+
+        /// <summary>
+        /// Returns all the connected device on selected channel
+        /// </summary>
+        /// <param name="channel">Selected Channel on Multiplexer</param>
+        /// <returns></returns>
+        public IEnumerable<int> ScanChannelsForDeviceAddress(Channels channel)
```
pgrawehr, line 181: You should be able to replace this with a call to `I2cBusExtensions.PerformBusScan` (in the `Iot.Device.Boards` namespace). That does the same already.

Thinking about it... Since that method takes an `I2cBus` as argument, I think it would make sense to add a method `public I2cBus CreateBusFromChannel(Channels channel)` with something like `return new I2cBus(<ArbitraryNumberToIdentifyChannel>)`

src/devices/Tca9548a/Tca9548A.cs
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            if (_shouldDispose)
+            {
+                _i2CDevice?.Dispose();
+                _i2CDevice = null!;
+            }
```
pgrawehr, line 224: set `_i2CDevice` to null regardless of the value of `_shouldDispose` (using this class again should throw an exception afterwards).

src/devices/Tca9548a/README.md
```diff
@@ -0,0 +1,49 @@
...
+using Tca9548A tca9548a = new Tca9548A(i2cDevice);
+
+foreach (Channels channel in Tca9548A.DeviceChannels)
+{
+    tca9548a.SelectChannel(Channels.Channel0);
+    if (tca9548a.TryGetSelectedChannel(out Channels selectedChannel))
+    {
+        Console.WriteLine(selectedChannel);
+    }
+}x`
```
pgrawehr, line 31: Is that x a typo?

src/devices/Tca9548a/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+![TCA9548A](./TCA9548A.jpg)
+
+TCA9548A [datasheet](https://www.ti.com/lit/ds/symlink/tca9548a.pdf)
+
+You will find this device as ["Adafruit TCA9548A 1-to-8 I2C Multiplexer Breakout"](https://learn.adafruit.com/adafruit-tca9548a-1-to-8-i2c-multiplexer-breakout)
+
+## Usage
+
+Create a ```Tca9548A``` class and pass the I2C device. The default I2C address is provided in the class.
```
pgrawehr, line 17: Your example is a bit pointless and doesn't really show an useful use case. Can you extend the sample (and the diagram) to include for instance to BMEXXX sensors with the same address?

src/devices/Tca9548a/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Sample wiring
+
+**Important** to understand:
+
+If you are getting lower data rate from sensor connected via Multiplexer on Raspberry Pi, you need to adjust the bus speed.
+
+This example uses the I2C bus 1 with GPIO 2 and 3.
+
+![Wiring sample](TCA9548A-fritzing-schema.png)
```
pgrawehr, line 49: Please also include the .fzz file in your PR.

pgrawehr (commented):
Thanks for the addition!

I have some minor comments and a suggestion.

src/devices/Tca9548a/Tca9548A.cs
asheesh1996: I have added an extension function in I2cBusExtensions class to get an I2cBus object from I2cDevice object

src/devices/Tca9548a/README.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+### Sample wiring
+
+**Important** to understand:
+If you are getting lower data rate from sensor connected via Multiplexer on Raspberry Pi, you need to adjust the bus speed.
+
+This example uses the I2C bus 1 with GPIO 2 and 3 on raspberry pi and gathers data from to BNO055 sensors and 4 BMP180 sensors over TCA9548A Mux
+
+![Wiring sample](TCA9548A_Connections.png)
```
pgrawehr, line 96: Thanks, that is much more helpful now. Do we still need the `TCA9548A-fritzing-schema.png` file?

src/devices/Board/I2cBusExtensions.cs
```diff
@@ -44,5 +44,16 @@ public static List<int> PerformBusScan(this I2cBus bus, int lowest = 0x3, int hi
...
+        /// <summary>
+        /// Create I2cBus Object type from I2cDevice
+        /// </summary>
+        /// <param name="i2cDevice">The I2cDevice for which I2cBus object needs to be created</param>
+        /// <returns></returns>
+        public static I2cBus CreateBusFromI2CDevice(this I2cDevice i2cDevice)
+        {
+            return I2cBus.Create(i2cDevice.ConnectionSettings.BusId);
+        }
+
```
pgrawehr, line 57: Unfortunately, that is not going to work, because it's not generic. 

The I2cDevice you have in hand here does not necessarily point to a device on the primary I2c Controller, it could be pointing to a device attached to an USB-To-I2c controller such as the FT4222. In this case, you would get something wrong or an exception here, as the system doesn't necessarily have a default controller at all. 

I think, to work around this, you would need to have your binding operate on the bus, not an individual device. Which logically doesn't seem to be that wrong to me.

pgrawehr (changes requested):
That was a good idea, but unfortunately, it fails in some use cases we have to consider. 

Please ping me if you want me to make a suggestion to fix it. Of course, you can try yourself first.

src/devices/Tca9548a/README.md
asheesh1996: Will remove it.

src/devices/Board/I2cBusExtensions.cs
asheesh1996: So Should I create an I2CBus for the device and from that create I2CDevices for the multiplexer?

asheesh1996: I would like to know what approach will be better for this?

pgrawehr: I need to think about it for a bit and will make a suggestion tonight. But yes, it will probably be something like that.

pgrawehr: What about this:

Change the constructor of `Tca9548A` to:
```csharp
 public Tca9548A(I2cBus masterBus, int multiplexerAddress, bool shouldDisposeBus = true)
{
    _masterBus = masterBus ?? throw new ArgumentNullException(nameof("masterBus"));
    _multiplexerDevice = masterBus.CreateDevice(multiplexerAddress);
}
```

So you specify that this multiplexer operates on the given bus and has the given address. 

You would instantiate it like so:
```csharp
var bus = I2cBus.Create(1);
var tca = new Tca9458A(bus, Tca9548A.DefaultI2cAddress);
// ...
```
or actually better yet (as more generic:)

```csharp
var board = new RaspberryPiBoard();
var bus = board.CreateOrGetI2cBus(1);
var tca = new Tca9458A(bus, Tca9548A.DefaultI2cAddress);
// ...
```

and then, in your `Tca9458A` class, you add a method like:

```csharp
public I2cBus CreateBusForChannel(int channel)
{
     return new Tca9548Bus(this, channel);
}
```
with Tca9548Bus being an internal class that would automatically remember that it was in fact one of the slave busses of the TCA.

Similarly, you had

```csharp
public I2cDevice CreateDevice(I2cConnectionSettings settings)
{
    // Verify settings.Channel is a valid sub-channel ID.
     return new Tca9458Bus(this, settings.Channel).CreateDevice(settings.DeviceAddress);
}
```
(this method is actually just a shortcut on what the bus would offer as well). The actual instance of I2cDevice returned would also be an internal implementation remembering that it was a part of the TCA. 

This all looks a bit complicated at first, but I think it will help keep things universally usable, even with special top-level controllers. And it has the nice effect that you could forward the obtained I2cDevice instance to any binding without having to care that it's attached to a multiplexer, as the channel switching would happen behind the scenes. 

I'm not currently suggesting to remove the existing `SelectChannel` method, as there might be use cases where you want to still do that manually (e.g. if you want to talk to multiple devices in parallel). 

What do you think?

asheesh1996: Hi, 
I was also thinking the best way would be to pass an I2cBus object type in the constructor. 

But I am confused about creating different I2Cbus for all the channels because all the devices connected to the TCA9548A mux will be on the same bus on which TCA9548A multiplexer is.

pgrawehr: Yes, the trick would be to create artificial bus numbers. So `CreateBusForChannel(1)` would return bus number 0x101 (or maybe `own_device_I'd << 8 + channel` so it would even work with multiple tca's).

asheesh1996: Hi, I have updated the object creation method for  TCA, but I was still not able to implement the artificial bus creation for different channel

src/devices/Tca9548a/Tca9548A.cs
```diff
@@ -0,0 +1,156 @@
...
+using System.Device.Model;
+using System.IO;
+using Iot.Device.Board;
+
+namespace Iot.Device.Tca9548a
+{
+    /// <summary>
+    /// Tca9548A - 8-Channel I2C Switch/Multiplexer
+    /// </summary>
+    [Interface("BNO055 - 8-Channel I2C Switch/Multiplexer")]
```
Ellerbach, line 16: please adjust the interface name here ;-)
```suggestion
    [Interface("Tca9548A - 8-Channel I2C Switch/Multiplexer")]
```

src/devices/Tca9548a/Tca9548A.cs
```diff
@@ -0,0 +1,156 @@
...
+        private I2cDevice _i2CDevice;
+
+        /// <summary>
+        /// Creates a Multiplexer Instance
+        /// </summary>
+        /// <param name="i2CBus">The I2C Bus on which Mux is</param>
+        /// <param name="address"> Address of the Mux on I2C bus</param>
+        /// <param name="shouldDispose">true to dispose the I2C device at dispose</param>
+        /// <exception cref="ArgumentNullException">Exception thrown if I2C device is null</exception>
+        public Tca9548A(I2cBus i2CBus, byte address = DefaultI2cAddress, bool shouldDispose = true)
```
Ellerbach, line 42: why not passing directly the I2cDevice instead of the I2cBus?

Ellerbach (commented):
Thanks, nice addition, few comments

src/devices/Tca9548a/samples/Program.cs
```diff
@@ -0,0 +1,111 @@
...
+                    }
+                    else if (device == Bmp180.DefaultI2cAddress)
+                    {
+                        using I2cDevice bmpDevice = bus.CreateDevice(Bmp180.DefaultI2cAddress);
+                        using Bmp180 i2cBmp280 = new(bmpDevice);
+                        // set samplings
+                        i2cBmp280.SetSampling(Sampling.Standard);
+                        // read values
+                        Temperature tempValue = i2cBmp280.ReadTemperature();
+                        Console.WriteLine($"Temperature: {tempValue.DegreesCelsius:0.#}\u00B0C");
```
krwq, line 79: Consider instead to create I2cDevice implementation which works more or less as following:

```csharp
Tca9548A i2cMux = new Tca9548A(i2cBus);
I2cBus channel1Bus = i2cMux.GetBus(1);
using Bmp280 bmp280 = new (channel1Bus.CreateDevice(Bmp180.DefaultI2cAddress));

// when bmp280.Whatever() is called then virtual I2cDevice switches to the right channel automatically and Reads/Writes to that i2cDevice
```

krwq (commented):
Really nice device. I think we can further improve usability though. See my comment above

src/devices/Tca9548a/Tca9548A.cs
krwq: Another design alternative:

```csharp
public Tca9548A : IList<I2cBus>
{
  public Tca9548A(I2cBus parentBus) {...}
  public T this[int i] => new Tc29548Bus(_parentBus, channel); // even better store it before hand in the private list and do not create it every time, this is just a concept idea
  // ...
}
```

```csharp
public Tca9548ABus : I2cBus // entire class can be internal if it doesn't have anything special over I2cBus
{
  internal Tca9548ABus(Tca9548A parent, int channel) {...}
  public I2cDevice CreateDevice(int address) => new Tca9548AChannel(...);
}
```

```csharp
public class Tca9548AChannel : I2cDevice
{
    internal Tca9548AChannel(Tca9548ABus parent, int channel) { ...}
    
    {
        SelectChannel(...);
        // do the writing
    }
   // etc...
}   
```

pgrawehr: @krwq  That's actually what I suggested already in my post. That is my preferable solution.

pgrawehr: @Ellerbach Because it needs both: The bus (from the main controller) and the device (of the TCA itself). That's because one can create client devices from the bus, but not determine the bus from a device. The suggested implementation in `I2cBusExtensions` (see individual commits) does not work.

asheesh1996: Hi @krwq, I am still not able to understand how the Tca9548AChannel object will select channel on the TCA9548A device in this implementation which is on different i2c address when we have to read or write. 
```

public class Tca9548AChannel : I2cDevice
{
    internal Tca9548AChannel(Tca9548ABus parent, int channel) { ...}
    
    {
        SelectChannel(...);
        // do the writing
    }   
}   
```

pgrawehr: Maybe the names are not optimal yet (maybe `Tca9548AChannel` should better be named `Tca9548ADevice`). 

The idea is that if you do `channel1Bus.CreateDevice(Bmp180.DefaultI2cAddress)` (with `channel1Bus` being an instance of a `Tca9548ABus`), then you get an instance of a `Tca9548AChannel`, which is a subclass of `I2cDevice`.  When the client binding now uses this instance (of which he doesn't know the exact type) the channel selection happens behind the scene, because of this implementation:

```csharp
    {
        SelectChannel(_channel);
        _realBus.Write(buffer);
    }   
``` 

This basically first selects the channel (which is a member of this class) and then forwards the call to the underlying I2C controller for the actual write. So this secretly inserts a channel selection command before every write command. (Of course, it would be useful to cache the currently selected channel as to avoid setting the same value over and over again, but that's a detail).

asheesh1996: > But to select a channel via any device created by TCA bus, it needs a reference to the original TCA device at address say 0x70 , so do I reference the main TCA class in all the devices created

krwq: @asheesh1996 the implementation detail doesn't matter. The idea is that user only sees I2cDevice and if it writes to that all magic happens behind the scenes

src/devices/Tca9548a/Tca9548ADevice.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Iot.Device.Tca9548a
+{
+    /// <summary>
+    /// I2C Device on TCA9548A channel
+    /// </summary>
+    public class Tca9548ADevice : I2cDevice
```
krwq, line 16: This is exactly the idea of the I2cDevice, note that unless this offers more than I2cDevice I'd recommend to keep this internal and just return `I2cDevice` from the parent.

src/devices/Tca9548a/Tca9548A.cs
```diff
@@ -46,6 +46,17 @@ public Tca9548A(I2cBus i2CBus, byte address = DefaultI2cAddress, bool shouldDisp
             _shouldDispose = shouldDispose;
         }
 
+        /// <summary>
+        /// Creates an I2c Device for selected sensor on Multiplexer
+        /// </summary>
+        /// <param name="tcaChannel">TCA Channel on which Device Exists</param>
+        /// <param name="address">I2c address of the device</param>
+        /// <returns></returns>
+        public I2cDevice CreateDeviceOnChannel(Channels tcaChannel, int address)
```
krwq, line 55: I'll recommend to rename this for consistency: `CreateI2cDevice`

src/devices/Tca9548a/Tca9548A.cs
```diff
@@ -0,0 +1,167 @@
...
+                selectedChannel = Channels.Channel0;
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Returns all the connected devices on all channels of Multiplexer
+        /// </summary>
+        /// <returns> Dictionary having Keys as Channels and Values as IEnumerable integer type</returns>
+        public Dictionary<Channels, IEnumerable<int>> ScanAllChannelsForDeviceAddress()
```
krwq, line 118: Once you add CreateI2cBus I believe this won't be needed since you will get this for free, is my understanding correct?

src/devices/Tca9548a/Tca9548A.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            return channels;
+        }
+
+        /// <summary>
+        /// Returns all the connected device on selected channel
+        /// </summary>
+        /// <param name="channel">Selected Channel on Multiplexer</param>
+        /// <returns></returns>
+        public IEnumerable<int> ScanChannelsForDeviceAddress(Channels channel)
```
krwq, line 134: I'd recommend to replace channel enum with simply `int NumberOfChannels { get; }` and `int` on the arguments. Or alternatively implement `IList<I2cBus>` and then just override `Count` rather than adding `NumberOfChannels`

krwq (commented):
We're heading in the good direction - this is pretty close now.

- I2cDevice implementation looks good, I'd recommend to make this class internal though (unless it offers anything extra on top of I2cDevice)
- Tca9548A - CreateI2cDevice implementation looks good (arguably int might be easier to use rather than `Channel`)
- Tca9548A - I'll recomend to also create CreateI2cBus method - I2cBus would only call to Tca9548A internal methods but nothing special other than that
- Tca9548A - you might want to make everything but CreateI2cBus and CreateI2cDevice and basic functionality internal
- you might also want to add slight improvement to not switch channel if it's already on the right one to avoid unnecesarilly writing to parent I2cDevice

Note that once this implementation provides I2cBus/I2cDevice we will be able to combine muxes together - that's the biggest benefit of doing this work.

pgrawehr:
One question: Are there similar devices available that would use the same binding (Eg is there a TCA9548b)? 

Because if so, the namespace should be more generic. (I'm not asking to implement those devices now, it's just about considering future extensions)

pgrawehr:
If I read https://www.ti.com/interface/i2c/switches-and-multiplexers/products.html?keyMatch=TCA9548 correctly, there are other, very similar devices with 2 or 4 channels. The naming is `[P|T]CA954x`. So at least the last digit of the namespace should be changed to `x`. Not sure about what to do with the lead character, as it can be P or T.

krwq:
@pgrawehr pick one, we should put all possible names in the README.md in the title so it's easier to find

pgrawehr:
I have updated the project name and the namespace to TCA954x, since there are several similar chips available. No functional changes otherwise, therefore to actually support e.g. TCA9544A, additional changes (and probably an additional derived type) will be necessary.

pgrawehr (approved):

pgrawehr:
@asheesh1996 I cleaned up the documentation and the API slightly, but hopefully I have not changed the behavior (I have no way of testing). There were several methods that exposed the same functionality twice (e.g. the `SwitchTCA9548AState` method was removed, because the same can be achieved by calling `SelectChannel` with an empty set). Unless you have some objections, I think this is ready to merge now.

src/devices/Tca954x/README.md
```diff
@@ -0,0 +1,115 @@
...
+You will find this device as ["Adafruit TCA9548A 1-to-8 I2C Multiplexer Breakout"](https://learn.adafruit.com/adafruit-tca9548a-1-to-8-i2c-multiplexer-breakout)
+
+A list of [different variants](https://www.ti.com/interface/i2c/switches-and-multiplexers/products.html?keyMatch=TCA9548) is available on the Texas Instruments Website.
+
+## Usage
+
+Create a ```Tca9548A``` class and pass the I2C device. The default I2C address is provided in the class.
+
+```csharp
+            Console.WriteLine("Hello TCA9548A!");
```
Ellerbach, line 24: detail, but you have everything on the left properly inented?

src/devices/Tca954x/Tca9548A.cs
```diff
@@ -0,0 +1,196 @@
...
+        /// <summary>
+        /// Array of all possible Multiplexer Channels
+        /// </summary>
+        private static readonly MultiplexerChannel[] DeviceChannels = new MultiplexerChannel[]
+        {
+            MultiplexerChannel.Channel0, MultiplexerChannel.Channel1, MultiplexerChannel.Channel2, MultiplexerChannel.Channel3, MultiplexerChannel.Channel4,
+            MultiplexerChannel.Channel5, MultiplexerChannel.Channel6, MultiplexerChannel.Channel7
+        };
+
+        internal I2cDevice _i2CDevice;
```
Ellerbach, line 38: for internal, isn't it better to use a property?

src/devices/Tca954x/Tca9548A.cs
```diff
@@ -0,0 +1,196 @@
...
+
+        /// <summary>
+        /// Creates a Multiplexer Instance
+        /// </summary>
+        /// <param name="i2cDevice">The I2C Device of the Mux itself</param>
+        /// <param name="shouldDispose">true to dispose the I2C device at dispose</param>
+        /// <exception cref="ArgumentNullException">Exception thrown if I2C device is null</exception>
+        public Tca9548A(I2cDevice i2cDevice, bool shouldDispose = true)
```
Ellerbach, line 46: Outside of this PR: @pgrawehr and @krwq we should discuss the I2cDevice disposal in bindings. Originally, we said we won't dispose them. because they have not being created in the binding. Now, there are quite some binding disposing them. 
We have something similar for GpioController with a shouldDispose pattern. So maybe to be clear as well, like here but with a _shouldDispoeI2c, we may bring a bit of clarity.

Ellerbach (approved):
Few remarks, otherwise, all good for me.

src/devices/Tca954x/Tca9548A.cs
pgrawehr: Yea, good point. I think, most of the time this is just confusing for the user.

src/devices/Tca954x/Tca9548A.cs
pgrawehr: Changed this to private. Can think of making it a property when we introduce derived classes or siblings.

src/devices/Tca954x/README.md
pgrawehr: Yea, indentation seems right.

src/devices/Tca954x/Tca9548A.cs
pgrawehr: Oh, good you found this one - looking at this, just found a major bug in the channel implementation.

asheesh1996:
No issues, thanks for refactoring the code base :)

pgrawehr:
@asheesh1996 Thanks again for this contribution!

PR #1915, opened 2022-08-31
Add BiColorBargraph using Ht16k33

richlander:
This PR adds a binding for the [Adafruit 24-bar bicolor bargraph](https://www.adafruit.com/product/1721). It uses the same HT16K33 chip as is used in the [`Display`](https://github.com/dotnet/iot/tree/main/src/devices/Display) folder.

giphy

I've also written a binding for [8x8 Matrices](https://www.adafruit.com/product/2042), which I'll add in another PR. This and bicolor bargraph are two additional bindings implemented at https://github.com/adafruit/Adafruit_CircuitPython_HT16K33 for this same chip.

I created a separate folder for HT16K33 in this PR. I had previously looked for it and not found it due to the `Display` folder naming, which I consider poorly chosen. We could rename the `Display` if that's decided to be a good thing.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1915)

### DIFF
--- a/src/devices/Display/BarColor.cs
+++ b/src/devices/Display/BarColor.cs
@@ -11,21 +11,21 @@ namespace Iot.Device.Display
         /// <summary>
         /// Disable LED.
         /// </summary>
-        OFF = 0,
+        Off = 0,
 
         /// <summary>
         /// Enable red LED.
         /// </summary>
-        RED = 1,
+        Red = 1,
 
         /// <summary>
         /// Enable green LED.
         /// </summary>
-        GREEN = 2,
+        Green = 2,
 
         /// <summary>
         /// Enable both green and red LEDs, producing a yellow color.
         /// </summary>
-        YELLOW = 3
+        Yellow = 3
     }
 }
--- a/src/devices/Display/BarColor.cs
+++ b/src/devices/Display/BarColor.cs
@@ -1,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// Describes LED colors in a Bargraph.
+    /// </summary>
+    public enum BarColor
+    {
+        /// <summary>
+        /// Disable LED.
+        /// </summary>
+        OFF = 0,
+
+        /// <summary>
+        /// Enable red LED.
+        /// </summary>
+        RED = 1,
+
+        /// <summary>
+        /// Enable green LED.
+        /// </summary>
+        GREEN = 2,
+
+        /// <summary>
+        /// Enable both green and red LEDs, producing a yellow color.
+        /// </summary>
+        YELLOW = 3
+    }
+}
--- a/src/devices/Display/BiColorBarGraph.cs
+++ b/src/devices/Display/BiColorBarGraph.cs
@@ -67,9 +67,16 @@ namespace Iot.Device.Display
             {
                 Flush();
             }
-    }
+        }
 
         /*
+            Task: Update data for the bargraph
+
+            The following diagram shows the intended orientation of the bargraph.
+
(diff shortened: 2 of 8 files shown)


src/devices/Display/BarColor.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// Describes LED colors in a Bargraph.
+    /// </summary>
+    public enum BarColor
+    {
+        /// <summary>
+        /// Disable LED.
+        /// </summary>
+        OFF = 0,
```
Ellerbach, line 14: With our naming convention, those public enums should be like `Off`, `Red`, `Green`, `Yellow`.

src/devices/Display/BiColorBarGraph.cs
```diff
@@ -0,0 +1,148 @@
...
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// Represents a 24-Segment bargraph that can display multiple colors.
+    /// </summary>
+    // Product: https://www.adafruit.com/product/1721
+    public class BiColorBarGraph : Ht16k33
+    {
+        private byte[] _displayBuffer = new byte[7];
```
Ellerbach, line 15: `private readonly` maybe for both?

src/devices/Display/BiColorBarGraph.cs
```diff
@@ -0,0 +1,148 @@
...
+            foreach (byte b in data)
+            {
+                _displayBuffer[startAddress++] = b;
+            }
+
+            if (BufferingEnabled)
+            {
+                Flush();
+            }
+    }
```
Ellerbach, line 70: indentation of this bracket seems missing some spaces

src/devices/Display/BiColorBarGraph.cs
```diff
@@ -0,0 +1,148 @@
...
+            int unit = index / 12;
+            int segment = index / 4;
+            int third = segment % 3;
+            int bit = index % 4 + (index / 12) * 4;
+            int mask = 1 << bit;
+            int bufferIndex = (third * 2) + 1;
+            byte red = _displayBuffer[bufferIndex];
+            byte green = _displayBuffer[bufferIndex + 1];
+
+            if (value is BarColor.OFF)
```
Ellerbach, line 128: why not a switch rather than the series of if?

Ellerbach (commented):
Few comments but all up, looks good!

src/devices/Display/BiColorBarGraph.cs
richlander: Both of those are being assigned to in the code. That means they cannot be `readonly`, right?

gfoidl: Use `_displayBuffer.AsSpan().Clear();` on L47 then it can be `readonly`, as nothing -- except the ctor -- assigns a new reference to the fields.

Writing to an array element of a readonly array-field doesn't violate "readonly" if that was kind of the question.

richlander: Got it. Repeated what you proposed on the other PR. Thanks for showing me!

src/devices/Display/BiColorBarGraph.cs
```diff
@@ -0,0 +1,189 @@
...
+                    _displayBuffer[5] = fill;
+                    break;
+                case BarColor.Green:
+                    _displayBuffer[2] = fill;
+                    _displayBuffer[4] = fill;
+                    _displayBuffer[6] = fill;
+                    break;
+                case BarColor.Yellow:
+                    Span<byte> displayBuffer = _displayBuffer;
+                    displayBuffer.Fill(0xFF);
```
richlander, line 64:
```suggestion
                    displayBuffer.Fill(fill);
```

Ellerbach (approved):
Looks good to me :-) Thanks for the addition :-)

PR #1924, opened 2022-09-13
Speed up the build by only performing one restore, and parallelizing build

joperezr:
These changes are cutting the build time in my local machine by at least 50%, but much more. Reason is we were performing restores several times for the same project, as well as building the same project multiple times. These changes make it so that every project is only built and restored once.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1924)


joperezr:
As an example, in CI the individual builds used to take 10-14 mins in average to build the whole repo, and CI in this PR is building the whole repo in less than 5 mins.

PR #1926, opened 2022-09-17
Add LED matrix bindings for IS31FL3731 driver

richlander:
Used by:

* [Adafruit 16x9 Charlieplexed PWM LED Matrix Driver - IS31FL3731](https://www.adafruit.com/product/2946)
* [Adafruit 15x7 CharliePlex LED Matrix Display FeatherWings](https://www.adafruit.com/product/2965)
* [Adafruit 16x8 CharliePlex LED Matrix Bonnets](https://www.adafruit.com/product/4127)
* [Pimoroni 17x7 Scroll pHAT HD](https://www.adafruit.com/product/3473)
* [Pimoroni 28x3 (r,g,b) Led Shim](https://www.adafruit.com/product/3831)
* [Pimoroni Keybow 2040 with 4x4 matrix of RGB LEDs](https://shop.pimoroni.com/products/keybow-2040)
* [Pimoroni 5x5 RGB Matrix Breakout](https://shop.pimoroni.com/products/5x5-rgb-matrix-breakout)

Port of https://github.com/adafruit/Adafruit_CircuitPython_IS31FL3731.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1926)


src/devices/Is31fl3731/Is31Fl3731.cs
```diff
@@ -0,0 +1,278 @@
...
+        : this(i2cDevice)
+        {
+            Width = width;
+            Height = height;
+        }
+
+        /// <summary>
+        /// Default I2C address for device.
+        /// </summary>
+        public static readonly int DefaultI2cAddress;
```
Ellerbach, line 75: ok, now I understand why no the const in the other classes :-)

src/devices/Is31fl3731/Is31Fl3731.cs
```diff
@@ -0,0 +1,278 @@
...
+        /// <summary>
+        /// Disable all LEDs.
+        /// </summary>
+        /// <param name="rate">Set the showdown mode. `true` sets device into shutdown mode. `false` sets device into normal operation.</param>
+        public void EnableBlinking(int rate)
+        {
+            int value = 0;
+            if (rate > 0)
+            {
+                rate /= 270;
```
Ellerbach, line 159: maybe add a comment on why those values or where they are coming from?

src/devices/Is31fl3731/Is31Fl3731.cs
```diff
@@ -0,0 +1,278 @@
...
+            _i2cDevice.Write(new byte[] { COMMAND_REGISTER, register });
+            byte[] data = new byte[value.Length + 1];
+            data[0] = address;
+            value.CopyTo(data.AsSpan(1));
+            _i2cDevice.Write(data);
+        }
+
+        private byte Read(int address)
+        {
+            byte[] buffer = new byte[1];
```
Ellerbach, line 219: same here for stackallock and Span<byte>

src/devices/Is31fl3731/Is31Fl3731.cs
```diff
@@ -0,0 +1,278 @@
...
+            _i2cDevice.Write(new byte[] { (byte)address });
+            return _i2cDevice.ReadByte();
+        }
+
+        private void Write(byte register, byte address, byte value)
+        {
+            _i2cDevice.Write(new byte[] { COMMAND_REGISTER, register });
```
Ellerbach, line 226: And here and everywhere else :-)

src/devices/Is31fl3731/Is31Fl3731.cs
```diff
@@ -0,0 +1,278 @@
...
+            // Reset device
+            Reset();
+
+            // set display mode (bits d4:d3)
+            // 00 = picture mode
+            // 01 = auto frame
+            // 1x audio frame play mode
+            Write(FUNCTION_REGISTER, CONFIGURATION_REGISTER, 0);
+            // set data page
+            _i2cDevice.Write(new byte[] { COMMAND_REGISTER, 0 });
```
Ellerbach, line 101: Span<byte> with stackallock

src/devices/Is31fl3731/LedShimRgb28x1.cs
```diff
@@ -0,0 +1,78 @@
...
+        public override int GetLedAddress(int x, int y) => (x, y) switch
+        {
+            // y == 0
+            (< 7, 0) => 118 - x,
+            (< 15, 0) => 141 - x,
+            (< 21, 0) => 106 + x,
+            (21, 0) => 15,
+            (_, 0) => x - 14,
+            // y == 1
+            (<2, 1) => 69 - x,
```
Ellerbach, line 40: you're missing some spaces in the following lines

src/devices/Is31fl3731/LedShimRgb28x1.cs
```diff
@@ -0,0 +1,78 @@
...
+        };
+
+        /// <summary>
+        /// Write RGB value.
+        /// </summary>
+        /// <param name="x">Horizontal pixel position.</param>
+        /// <param name="r">Red brightness value 0->255.</param>
+        /// <param name="g">Green brightness value 0->255.</param>
+        /// <param name="b">Blue brightness value 0->255.</param>
+        public void WritePixelRgb(int x, int r, int g, int b)
```
Ellerbach, line 71: Same feedback as for the other one: usage of Color

Ellerbach (commented):
Couple of comments. Also, we should have a discussion regarding simple things like Color. With changes in the graphic libs, we need to come with something simple we can use anywhere for consistency

src/devices/Is31fl3731/Bonnet16x8.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <param name="i2cDevice">The <see cref="System.Device.I2c.I2cDevice"/> to create with.</param>
+        public Bonnet16x8(I2cDevice i2cDevice)
+        : base(i2cDevice, 16, 8)
+        {
+        }
+
+        /// <summary>
+        /// Default I2C address for device.
+        /// </summary>
+        public static new readonly int DefaultI2cAddress = 0x74;
```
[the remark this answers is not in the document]
richlander: Because this code is hiding an inherited member, which requires `new`.

```csharp
public static const int DefaultI2cAddress = 0x74;
```

Produces the following error:

> Bonnet16x8.DefaultI2cAddress' hides inherited member 'Is31fl3731.DefaultI2cAddress'. Use the new keyword if hiding was intended. [Is31Fl3731]csharp(CS0108)

src/devices/Is31fl3731/Is31Fl3731.cs
```diff
@@ -0,0 +1,278 @@
...
+    // Datasheet: https://cdn-learn.adafruit.com/assets/assets/000/030/994/original/31FL3731.pdf
+    // Product: https://www.adafruit.com/product/2946
+    // Product: https://www.adafruit.com/product/2974
+    public abstract class Is31fl3731 : IDisposable
+    {
+        // Register addresses
+        // Command register
+        // Points pages one to nine
+        // table 2 in datasheet
+        private const byte COMMAND_REGISTER = 0xFD;
```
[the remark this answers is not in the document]
richlander: Not sure about that. It's gonna make the code look pretty ugly without target typed enums.

Let's take a look.

This line:

```charp
Write(FUNCTION_REGISTER, DISPLAY_REGISTER, (byte)value);
```

Will turn into:

```charp
Write(Registers.FUNCTION_REGISTER, Registers.DISPLAY_REGISTER, (byte)value);
```

Right? I think that makes the C# look worse as compared to the Python code I read. Seems like this is trading cleanup at the top of the file for worse readability within the core logic.

src/devices/Is31fl3731/Is31Fl3731.cs
```diff
@@ -0,0 +1,278 @@
...
+
+        private I2cDevice _i2cDevice;
+
+        /// <summary>
+        /// Initialize IS31FL3731 device
+        /// </summary>
+        /// <param name="i2cDevice">The <see cref="System.Device.I2c.I2cDevice"/> to create with.</param>
+        public Is31fl3731(I2cDevice i2cDevice)
+        {
+            _i2cDevice = i2cDevice;
```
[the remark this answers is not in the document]
richlander: Done.

src/devices/Is31fl3731/Is31Fl3731.cs
richlander: Right.

src/devices/Is31fl3731/Is31Fl3731.cs
```diff
@@ -0,0 +1,278 @@
...
+            // 0 = shutdown mode
+            // 1 = normal operation
+            int mode = shutdown ? 0 : 1;
+            Write(FUNCTION_REGISTER, SHUTDOWN, (byte)mode);
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            _i2cDevice?.Dispose();
```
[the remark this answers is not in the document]
richlander: Got it. I removed `: IDisposable` from the class definition, too.

src/devices/Is31fl3731/Is31Fl3731.cs
```diff
@@ -0,0 +1,278 @@
...
+            get => ReadLedPwm(x, y);
+            set => WriteLedPwm(x, y, value);
+        }
+
+        /// <summary>
+        /// Set value for LED.
+        /// </summary>
+        public void WritePixel(int x, int y, int brightness, bool enable, bool blink)
+        {
+            WriteLedBlink(x, y, blink);
```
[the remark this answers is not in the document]
richlander: Good question. It's basically impossible for this code to know what values are in/out of range since there are multiple addressing schemes that are used higher-up. Certainly, the code could check for values are obviously out-of-range, like 1000, but I don't think that would have much value since values that are in-range for one matrix are out-of-range for another. This check should be done a layer up.

src/devices/Is31fl3731/Is31Fl3731.cs
richlander: Done.

src/devices/Is31fl3731/Is31Fl3731.cs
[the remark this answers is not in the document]
richlander: But then code like this is bad, right?

```csharp
Breakout11x7 matrix = new(I2cDevice.Create(new I2cConnectionSettings(busId: 1, Breakout11x7.DefaultI2cAddress)));
```

It's entirely possible that I'm missing knowledge of disposability.

src/devices/Is31fl3731/LedShimRgb28x1.cs
richlander: I'm not following. Can you elaborate?

src/devices/Is31fl3731/LedShimRgb28x1.cs
richlander: Fixed.

src/devices/Is31fl3731/BreakoutRgb5x5.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// Write RGB value.
+        /// </summary>
+        /// <param name="x">Horizontal pixel position.</param>
+        /// <param name="y">Vertical pixel position.</param>
+        /// <param name="r">Red brightness value 0->255.</param>
+        /// <param name="g">Green brightness value 0->255.</param>
+        /// <param name="b">Blue brightness value 0->255.</param>
+        public void WritePixelRgb(int x, int y, int r, int g, int b)
```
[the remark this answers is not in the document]
richlander: Fixed.

richlander:
I need to re-test all the devices and then I'll switch this PR to "Ready for review".

src/devices/Is31fl3731/Is31fl3731.cs
```diff
@@ -0,0 +1,275 @@
...
+    // Datasheet: https://cdn-learn.adafruit.com/assets/assets/000/030/994/original/31FL3731.pdf
+    // Product: https://www.adafruit.com/product/2946
+    // Product: https://www.adafruit.com/product/2974
+    public abstract class Is31fl3731
+    {
+        // Register addresses
+        // Command register
+        // Points pages one to nine
+        // table 2 in datasheet
+        private const byte COMMAND_REGISTER = 0xFD;
```
Ellerbach, line 22: rather than private const, a nice enum is always preferable especially in that case where you have quite a lot of them.

src/devices/Is31fl3731/Is31Fl3731.cs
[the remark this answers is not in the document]
Ellerbach: I really prefer to use enums because they avoid problems. You can group them. And in this case have 2 different enums, one for the register and one for the command.
Short term const seems super nice. Then when it comes to maintenance, they are not good at all. Enums forces you to use the right one. Here on the Write function, I can use the wrong one very easilly.

So most likely your write will turn out like with 2 different enums. And when you have possibility to mix elements, it's much easier for maintenance and readability with enums:

```csharp
Write(Function.FUNCTION_REGISTER, Register.DISPLAY_REGISTER, (byte)value);
```

src/devices/Is31fl3731/Is31fl3731.cs
richlander: I took your advice.

I first tried enums. That looked pretty rough to me since I had to add a cast at every callsite: https://github.com/dotnet/iot/pull/1926/commits/f3c7b6334af0b3547c035b0c82ab2f7fe5d4cd44

I then switched to static classes. That resolved the issue: https://github.com/dotnet/iot/pull/1926/commits/ecd652f87f44bd0de2671b8420d41fdda75a6e8f

I left the API signatures as `byte`. The reason is that there are multiple registers and their values overlap. Also, one of the registers takes values 0-7 for 8 pages of data. Forcing the use of an enum or static class to iterate from `0..7` doesn't make much sense. 

Seem better?

richlander:
This link seems to consistently fail.

image

src/devices/Is31fl3731/BreakoutRgb5x5.cs
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// Represents 5x5 RGB matrix, driven by an IS31FL3731 LED chip.
+    /// </summary>
+    // Product: https://shop.pimoroni.com/products/5x5-rgb-matrix-breakout
+    public class BreakoutRgb5x5 : Is31fl3731
+    {
+        private byte[][] _addresses = new byte[][]
```
Ellerbach, line 18: maybe even static readonly? So less chances to "damage" it and as it's anyway like a constant, a bit less memory footprint when it's used?

src/devices/Is31fl3731/Registers.cs
```diff
@@ -0,0 +1,69 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// Register addresses for the Command Register.
+    /// </summary>
+    // Enables pages one to nine
+    // Table 2 in datasheet
+    public static class CommandRegister
```
Ellerbach, line 11: why public cstatic lasses and not enums? you don't seem to override them or anything like this.

src/devices/Is31fl3731/samples/Program.Matrix.cs
```diff
@@ -0,0 +1,205 @@
...
+using System.Device.I2c;
+using System.Linq;
+using System.Threading;
+using Iot.Device.Display;
+
+// For 16x9 matrix
+// https://www.adafruit.com/product/2974
+using I2cDevice i2cDevice = I2cDevice.Create(new I2cConnectionSettings(busId: 1, Backpack16x9.DefaultI2cAddress));
+Backpack16x9 matrix = new(i2cDevice);
+// For 16x8 matrix Charlieplex bonet
```
Ellerbach, line 15: maybe itroduce an empty line between each block of matrix? you can as well use 4 slash or dashes or anything to make it more redable

Ellerbach (approved):
Very minor comments, all look very nice :-) mpatient to buy few of those and find time to use them :-)

src/devices/Is31fl3731/samples/MatrixRgb.sample.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
pgrawehr, line 1: Please put all sample projects into separate folder (samples\sample1, samples\sample2, etc.). Having multiple .csproj files in the same folder causes problems on macos due to a compiler bug.

src/devices/Is31fl3731/samples/LedShim28x1.sample.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net60</TargetFrameworks>
```
pgrawehr, line 4: TFM must be `net6.0`

src/devices/Is31fl3731/samples/MatrixRgb.sample.csproj
```diff
@@ -0,0 +1,11 @@
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net60</TargetFrameworks>
```
pgrawehr, line 4: TFM must be `net6.0`.

pgrawehr (changes requested):
Few comments that (I think) explain the build failures.

src/devices/Is31fl3731/samples/MatrixRgb.sample.csproj
richlander: I believe that there are multiple bindings that have that problem. Is this our new policy?

src/devices/Is31fl3731/samples/Program.Matrix.cs
richlander: Good call. Done.

pgrawehr (approved):

pgrawehr:
This needs a conflict resolution.

PR #1930, opened 2022-09-22
Add binding for IS31FL3730 LED driver

richlander:
Used by:

- https://shop.pimoroni.com/products/microdot-phat
- https://shop.pimoroni.com/products/led-dot-matrix-breakout

Port of https://github.com/pimoroni/microdot-phat.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1930)


src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,447 @@
...
+            byte[] buffer = _buffers[matrix];
+
+            if (value > 0)
+            {
+                buffer.AsSpan().Fill(255);
+            }
+            else
+            {
+                buffer.AsSpan().Fill(0);
+            }
```
gfoidl, line 168:
```suggestion
            buffer.AsSpan().Fill(value > 0 ? 255 : 0);
```
More a matter of taste, but IL-size is also smaller (which I don't believe is a criteria here).

Or
```suggestion
            if (value > 0)
            {
                buffer.AsSpan().Fill(255);
            }
            else
            {
                buffer.AsSpan().Clear();
            }
```
`Clear` is (in general) better than `Fill(0)` (better optimization in the implementation).

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,447 @@
...
+            Write(CONFIGURATION_REGISTER, (byte)_configurationValue);
+        }
+
+        private void Write(byte address, byte[] value)
+        {
+            byte[] data = new byte[value.Length + 1];
+            data[0] = address;
+            value.CopyTo(data.AsSpan(1));
+            _i2cDevice.Write(data);
+        }
```
gfoidl, line 292:
```suggestion
        private void Write(byte address, ReadOnlySpan<byte> value)
        {
            Debug.Assert(value.Length < 256);
            
            Span<byte> data = stackalloc byte[value.Length + ];
            data[0] = address;
            value.CopyTo(data.Slice(1));
            _i2cDevice.Write(data);
        }
```
Avoids the temp. allocation.

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,447 @@
...
+
+        private void Write(byte address, byte value)
+        {
+            _i2cDevice.Write(new byte[] { address, value });
```
gfoidl, line 296:
```suggestion
            _i2cDevice.Write(stackalloc byte[] { address, value });
```

gfoidl (commented):
I see the draft, but hope you're OK with early feedback.

src/devices/Is31fl3730/BreakoutPair5x7.cs
```diff
@@ -0,0 +1,96 @@
...
+        private Is31fl3730 _is31fl3730;
+        private I2cDevice? _i2cDevice;
+
+        /// <summary>
+        /// Initialize IS31FL3730 device
+        /// </summary>
+        /// <param name="i2cDevice">The <see cref="System.Device.I2c.I2cDevice"/> to create with.</param>
+        public BreakoutPair5x7(I2cDevice? i2cDevice = null)
+        {
+            _i2cDevice = i2cDevice is not null ? i2cDevice : I2cDevice.Create(new(1, Is31fl3730.DefaultI2cAddress));
```
Ellerbach, line 33: We're not using this pattern and always leave the creation of the I2cDevice to the caller. Reason is because bus may be different, ownership as well and finally adaptors like FT4222 or Arduino. So don't pass the null as default and raise and exception if null.

src/devices/Is31fl3730/BreakoutPair5x7.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// Fill All LEDs.
+        /// </summary>
+        public void Fill(int value) => _is31fl3730.FillAll(value);
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            _is31fl3730?.Dispose();
+            _is31fl3730 = null!;
+            _i2cDevice?.Dispose();
```
Ellerbach, line 92: with the pattern we are using, as we don't create I2cDevice in the bindings, we don't dispose them. It's to the caller to take care of this.

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,381 @@
...
+    // Datasheet: https://cdn-shop.adafruit.com/product-files/3017/31FL3730.pdf
+    // Product: https://shop.pimoroni.com/products/microdot-phat
+    // Product: https://shop.pimoroni.com/products/led-dot-matrix-breakout
+    // Related repo: https://github.com/pimoroni/microdot-phat
+    // Based on: https://github.com/pimoroni/microdot-phat/blob/master/library/microdotphat/matrix.py
+    public class Is31fl3730 : IDisposable
+    {
+        // Function register
+        // table 2 in datasheet
+        private const byte CONFIGURATION_REGISTER = 0x0;
```
Ellerbach, line 24: rather than using constants, use a Register.cs file with enums in there being internal. Like this, you will be able to keep the register names like they are (all uppercase). And it's much better than constants.

src/devices/Is31fl3730/MatrixMode.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    /// <summary>
+    /// Represents an IS31FL3731 LED Matrix driver
+    /// </summary>
+    public enum MatrixMode
+    {
+        /// <summary>
+        /// Represents a 8x8 LED matrix.
+        /// </summary>
+        Size8x8,
```
Ellerbach, line 21: As you are using explicit check of MatrixMode > 0 earlier, make the first element explicit to 0

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,381 @@
...
+            the first 5 bits. If you are writing to matrix 2 then writing one row at a
+            time won't work. Your content will be 90d flipped and the content (if you write
+            the same 5-bit encoded bitmap) will be 2 pixels short. Instead, you need to
+            translate row-based content to columns, similar to the `else` clause below.
+
+            Separately, you can disable one matrix or the other. That's largely a separate
+            concern. It has nothing to do with byte structure.
+            */
+
+            int logicalRow = matrix is 0 ? y : x;
```
Ellerbach, line 249: in theory, you should check that matrix, x and y are in the proper range because here it can break if not in the range.
Same for couple of other functions below

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,381 @@
...
+            int r = m[row];
+            return r & mask;
+        }
+
+        /// <summary>
+        /// Update decimal point for matrix.
+        /// </summary>
+        public void UpdateDecimalPoint(int matrix, int value)
+        {
+            int matrixOneMask = 128;
```
Ellerbach, line 279: const for those 2 would be better as you are using them like this

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,381 @@
...
+        {
+            _i2cDevice?.Dispose();
+            _i2cDevice = null!;
+        }
+
+        private void Write(byte address, byte[] value)
+        {
+            Span<byte> data = stackalloc byte[value.Length + 1];
+            data[0] = address;
+            value.CopyTo(data[1..]);
```
Ellerbach, line 349: I guess you have to adjust the imports to add the new range (I love them btw :-))

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        /// <summary>
+        /// Write LED for matrix.
+        /// </summary>
+        /// <param name="matrix">The matrix to use.</param>
+        /// <param name="x">The x dimension for the LED.</param>
+        /// <param name="y">The y dimension for the LED.</param>
+        /// <param name="value">The value to write.</param>
+        public void WriteLed(int matrix, int x, int y, int value)
```
Ellerbach, line 154: Something to consider as you only have 2 matrix: use an enum rather than an int?

src/devices/Is31fl3730/MicroDotphat30x7.cs
```diff
@@ -0,0 +1,96 @@
...
+        private Is31fl3730 _is31fl3730;
+        private I2cDevice? _i2cDevice;
+
+        /// <summary>
+        /// Initialize IS31FL3730 device
+        /// </summary>
+        /// <param name="i2cDevice">The <see cref="System.Device.I2c.I2cDevice"/> to create with.</param>
+        public BreakoutPair5x7(I2cDevice? i2cDevice = null)
+        {
+            _i2cDevice = i2cDevice is not null ? i2cDevice : I2cDevice.Create(new(1, Is31fl3730.DefaultI2cAddress));
```
Ellerbach, line 33: same remark as for the other one, we don't create I2cDevice from the bindings.

src/devices/Is31fl3730/MicroDotphat30x7.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// Fill All LEDs.
+        /// </summary>
+        public void Fill(int value) => _is31fl3730.FillAll(value);
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            _is31fl3730?.Dispose();
+            _is31fl3730 = null!;
+            _i2cDevice?.Dispose();
```
Ellerbach, line 92: And similar as my comment for the other one, no need to dispose as we don't create it in the binding

Ellerbach (commented):
Couple of comments for this very nice addition :-)

src/devices/Is31fl3730/Is31fl3730.cs
krwq: just FYI we prefer cleaner code to microoptimizations in this code-base except where it really matters. Allocations - yes, microoptimization not quite. Here I2C transmission itself will likely take more than 99% of the time and this won't make any observable difference. I'd rather write this code as:

```csharp
buffer.AsSpan().Fill(value > 0 ? 255 : 0);
```

src/devices/Is31fl3730/Is31fl3730.cs
richlander: Is there guidance on how to make this work for .NET Standard 2.0? I didn't find that.

richlander: I went with this: https://github.com/dotnet/iot/pull/1930/commits/d82f7ff6fe4e67c4401537151595da18840315a6

src/devices/Is31fl3730/MicroDotphat30x7.cs
richlander: I followed that guidance, however, when I came to supporting the Micro Dot pHAT, I found that there was a need to specify three I2C addresses. That's a bit of a pain. What do you think?

Here are two constructors: https://github.com/dotnet/iot/pull/1930/commits/0d1e95fbea3852ca4b49b1ab99f131dc7d2c8b18#diff-c1aa6bb26153d700cceaa57064762e4c6dce51672088c47a5f883ded6f76c12eR28-R47

I'm happy to do whatever you think best. I wanted to demonstrate the UX of the two approaches.

richlander: I resolved this.

pgrawehr:
is this very different from #1926? Should the two be completely independent?

richlander:
Yes. They are very different. The two drivers do not work similarly. If you look at the datasheets, you'll see that there very little similarity w/no opportunity for code sharing. Still, good question.

src/devices/Is31fl3730/MatrixValues.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// Matrix Values.
+    /// </summary>
+    public static class MatrixValues
```
pgrawehr, line 9: Does this need to be public?

src/devices/Is31fl3730/Registers.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// Register addresses for the Function Register.
+    /// </summary>
+    public static class FunctionRegister
```
pgrawehr, line 9: There's probably no need to make this public.

src/devices/Is31fl3730/Registers.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <summary>
+        /// Address for Reset Register.
+        /// </summary>
+        public static byte Reset = 0xFF;
+    }
+
+    /// <summary>
+    /// Register addresses for the Configuration Register.
+    /// </summary>
+    public static class ConfigurationRegister
```
pgrawehr, line 50: and this, too.

pgrawehr (changes requested):
Looks mostly good. Some general comments:
- Documentation of some classes could be improved (was apparently just copy-pasted in some places)
- Review the visibility of enums and constants. Ideally, bindings should not expose register numbers and other low-level constants.

src/devices/Display/IMatrix.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        int Height { get; }
+
+        /// <summary>
+        /// Indexer for matrix.
+        /// </summary>
+        int this[int x, int y] { get; set; }
+
+        /// <summary>
+        /// Fill LEDs (0 is dark; 255 is all lit).
```
Ellerbach, line 29: if it goes from 0 to 255, why not a byte?

src/devices/Is31fl3730/Current.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// 35 mA
+        /// </summary>
+        CmA35 = 0b1110,
+
+        /// <summary>
+        /// 40 mA
+        /// </summary>
+        CmA40 = 0b0,
```
Ellerbach, line 36: even if those are just 0, maybe better to align with 4 0 like the rest?

src/devices/Is31fl3730/Current.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// 45 mA
+        /// </summary>
+        CMA45 = 0b0001,
```
Ellerbach, line 41:
```suggestion
        CmA45 = 0b0001,
```

src/devices/Is31fl3730/DotMatrix.cs
```diff
@@ -0,0 +1,98 @@
...
+    // Product: https://shop.pimoroni.com/products/led-dot-matrix-breakout
+    // Related repo: https://github.com/pimoroni/microdot-phat
+    public class DotMatrix : IMatrix
+    {
+        private readonly DotMatrix5x7[] _matrices;
+
+        /// <summary>
+        /// Initialize virtual Dot Matrix.
+        /// </summary>
+        /// <param name="matrices">The  to create with.</param>
```
Ellerbach, line 22: I think something is missing in the comment

src/devices/Is31fl3730/DotMatrix.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        /// <inheritdoc/>
+        public int this[int x, int y]
+        {
+            get
+            {
+                var (matrixIndex, index) = GetMatrixIndex(x, y);
+                return _matrices[matrixIndex][index, y];
+            }
+            set
```
Ellerbach, line 53: I think you are missing an empty line before the set

src/devices/Is31fl3730/DotMatrix10x7.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Collections.Generic;
+using System.Device.I2c;
+using System.Threading;
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// Represents an IS31FL3731 LED Matrix driver
+    /// </summary>
+    // Datasheet: https://cdn-shop.adafruit.com/product-files/3017/31FL3730.pdf
```
Ellerbach, line 15: those comments goes into a /// <remark> comment

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,423 @@
...
+using System.Collections.Generic;
+using System.Device.I2c;
+using static System.Linq.Enumerable;
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// Represents an IS31FL3731 LED Matrix driver
+    /// </summary>
+    // Datasheet: https://cdn-shop.adafruit.com/product-files/3017/31FL3730.pdf
```
Ellerbach, line 14: those comments goes into a /// <remark> comment

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,423 @@
...
+
+        /// <summary>
+        /// Brightness of LED matrix (override default value (128; max brightness); set before calling Initialize method).
+        /// </summary>
+        public int Brightness { get; set; }
+
+        /// <summary>
+        /// Full current setting for each row output of LED matrix (override default value (40 mA)); set before calling Initialize method).
+        /// </summary>
+        public Current Current { get; set; }
```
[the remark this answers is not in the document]
Ellerbach: +1, the interest of having such properties is for them to be dynamic, so you can adjust them on the fly without having to call the initialize function

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,423 @@
...
+        {
+            new byte[8],
+            new byte[8]
+        };
+        private readonly bool[] _enabled = new bool[2];
+        private I2cDevice _i2cDevice;
+        private int _configurationValue = 0;
+
+        /// <summary>
+        /// Initialize IS31FL3730 device
```
Ellerbach, line 32: I give the comment here but it's better to finish all comments with a dot. Some are finished properly like this, some are not. The linter isn't configured to force it but I have to say, I have a preference for sentences :-) So it's optional but still consistency is a good way to go.

src/devices/Is31fl3730/Registers.cs
Ellerbach: and each enum should be in its own file

Ellerbach (commented):
Few more comments.

src/devices/Is31fl3730/Current.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Iot.Device.Multiplexing.Utility;
+
+namespace Iot.Device.Display
+{
+    /// <summary>
+    /// Represents an IS31FL3731 LED Matrix driver
+    /// </summary>
+    public enum Current
```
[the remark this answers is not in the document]
richlander: Fixed. I copied the text from the datasheet.

src/devices/Is31fl3730/MatrixValues.cs
richlander: Good call. Fixed.

src/devices/Is31fl3730/Registers.cs
richlander: Good call. Fixed.

src/devices/Is31fl3730/Registers.cs
richlander: Fixed.

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,423 @@
...
+        }
+
+        /// <summary>
+        /// Write LED for matrix.
+        /// </summary>
+        /// <param name="matrix">The matrix to use.</param>
+        /// <param name="x">The x dimension for the LED.</param>
+        /// <param name="y">The y dimension for the LED.</param>
+        /// <param name="value">The value to write.</param>
+        public void Write(int matrix, int x, int y, int value)
```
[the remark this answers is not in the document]
richlander: Fixed.

src/devices/Display/IMatrix.cs
richlander: I switched to `PinValue` instead.

src/devices/Is31fl3730/Is31fl3730.cs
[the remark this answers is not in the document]
richlander: Fixed. Good call.

src/devices/Is31fl3730/DotMatrix10x7.cs
richlander: I typically leave datasheet comments for the code. That strikes me as the best model. Thoughts?

richlander:
I'm having some namespace problems, as you can see with https://github.com/dotnet/iot/pull/1930/commits/09217241fd74149b69a283df98ec3df6ab675202. Any advice on that would be appreciated.

src/devices/Is31fl3730/DotMatrix10x7.cs
Ellerbach: Fair enough! You're right, it does not necessarily brings a lot of value. We can assume the person has anyway access to the datasheet of the sensor so they can cable it properly. And also, we've seen that those URL can change. So good for the code.

src/devices/Is31fl3730/FunctionRegister.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Display.Internal
+{
+    /// <summary>
+    /// Register addresses for the Function Register.
+    /// </summary>
+    internal static class Is31fl3730_FunctionRegister
```
Ellerbach, line 9: You should not use underscore in the name of classes or enums.
also the file name should be the same as the class/enum.

richlander: Right. Do you have a suggestion on better naming? The issue is this: https://github.com/dotnet/iot/blob/2580229760fd983b7ab6f0280ee2746cd4edca9c/src/devices/Is31fl3731/Registers.cs#L52

richlander: I can go back and break that file into one per type.

Ellerbach: You can just remove the underscore and that would work. In all cases, rename as well the file so it's more coherent.

Ellerbach (approved):
Thanks, looks great to me now!

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,383 @@
...
+        public DotMatrix5x7 this[int index] => GetMatrix(index);
+
+        /// <summary>
+        /// Update bright of LED matrix for each of 128 items.
+        /// </summary>
+        // From datasheet:
+        // Set high bit for the 128th item
+        // Set high bit low for the remaining 127 items
+        // And then set lower bits as appropriate, to represent 0-127
+        public void UpdateBrightness(byte value) => Write(Is31fl3730FunctionRegister.Pwm, value);
```
pgrawehr, line 83: According to the documentation, not all values are valid (only 0-128). If that's correct, this method should contain some input validation.

src/devices/Is31fl3730/Is31fl3730.cs
```diff
@@ -0,0 +1,383 @@
...
+
+        /// <summary>
+        /// Update current setting for each row output of LED matrix (default value is 40 mA).
+        /// </summary>
+        public void UpdateCurrent(Current value) => Write(Is31fl3730FunctionRegister.LightingEffect, (byte)value);
```
pgrawehr, line 88: While we're not doing that everywhere correctly, it would be good to check for valid values here as well. The user could pass any value that is not part of the enum.

pgrawehr (approved):
LGTM. You might wish to take a look at input validation, though.

src/devices/Is31fl3730/Is31fl3730.cs
richlander: Interesting. That's a good point. Fair.

PR #1939, opened 2022-10-01
NMEA support improvements

pgrawehr:
- Add AIS parser, including AIS target manager
- Add messages for engine surveillance
- Improve performance of Geoid calculations
- Add NMEA Udp server
- Add NMEA TCP client
- Fix bugs
- Improve simulation

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1939)


THIRD-PARTY-NOTICES.TXT
```diff
@@ -69,3 +69,10 @@ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
 CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+License notice for AIS parser (in NMEA-0183 binding)
```
joperezr, line 74: We should check with @richlander, but I believe we have to also paste the whole license here as well.

src/devices/Nmea0183/samples/NmeaSimulator/NmeaSimulator.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>$(DefaultSampleTfms)</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
joperezr, line 5: Default sample TFMs is already net6.0, so I don't believe this is needed:

[https://github.com/dotnet/iot/blob/8be21fcd2c691bad349878592fce9fc726970358/src/devices/Directory.Build.props#L9](https://github.com/dotnet/iot/blob/8be21fcd2c691bad349878592fce9fc726970358/src/devices/Directory.Build.props#L9)

src/devices/Nmea0183/tests/Nmea0183.Tests.csproj
```diff
@@ -11,5 +11,8 @@
   <ItemGroup>
     <ProjectReference Include="..\Nmea0183.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Shouldly" Version="4.0.3" />
```
joperezr, line 15: Also consider moving the version of the package to versions.props.

src/devices/Nmea0183/tests/Ais/AddressedSafetyRelatedMessageTests.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Iot.Device.Nmea0183.Ais;
+using Iot.Device.Nmea0183.AisSentences;
+using Shouldly;
```
joperezr, line 6: As you know we usually just use Asserts for our tests. Did you mainly use the dependency for the better messages that you get or is there something in specific that Assert doesn't support? I only ask because it would be good to be somewhat consistent on the repo.

pgrawehr: I added it because the original repo I took these classes from used it and I didn't want to rewrite all the tests. I didn't know this library before, but I think it's very nice. 

I do agree that this generates a bit of a mess between Assert and ShouldBe-syntax, but is this really an issue? I don't think so, since both variants are very easy to understand (and we should encourage users for more test coverage, anyway...!)

src/devices/Nmea0183/samples/NmeaSimulator/NmeaSimulator.csproj
pgrawehr: True. I guess this change was done before we changed the sample TFM

src/devices/Nmea0183/tests/Nmea0183.Tests.csproj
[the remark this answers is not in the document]
pgrawehr: I think it's helpful to have ProjectReferences and PackageReferences separated into groups, for easier readability.

pgrawehr: Will do when we find that this library is helpful elsewhere.

src/devices/Common/tests/GeographicPositionTest.cs
```diff
@@ -39,10 +40,31 @@ public void EqualityGeographicPosition()
         [Fact]
         public void DefaultFormatting()
         {
-            Assert.Equal("10° 00' 00.00\"N / 32° 30' 00.00\"E Ellipsoidal Height 12345m", new GeographicPosition(10.0, 32.5, 12345).ToString());
-            Assert.Equal("111° 06' 08.42\"S / 32° 59' 15.36\"W Ellipsoidal Height 200m", new GeographicPosition(-111.10234, -32.9876, 200).ToString());
-            Assert.Equal("11° 00' 00.00\"S / 33° 00' 00.00\"E Ellipsoidal Height 200m", new GeographicPosition(-10.99999999999999999, 32.999999999999999, 200).ToString());
+            Assert.Equal("10° 00' 00.00\"N 032° 30' 00.00\"E Ellipsoidal Height 12345m", new GeographicPosition(10.0, 32.5, 12345).ToString());
```
krwq, line 43: 0 in front of the numbers usually suggests octal system

src/devices/Common/tests/GeographicPositionTest.cs
```diff
@@ -39,10 +40,31 @@ public void EqualityGeographicPosition()
...
+            Assert.Equal("11° 06' 08.42\"S 132° 59' 15.36\"W Ellipsoidal Height 200m", new GeographicPosition(-11.10234, -132.9876, 200).ToString());
+            Assert.Equal("11° 00' 00.00\"S 033° 00' 00.00\"E Ellipsoidal Height 200m", new GeographicPosition(-10.99999999999999999, 32.999999999999999, 200).ToString());
         }
 
+        [Theory]
+        [InlineData("10.000° 23.500°", 10.0, 23.5, 12345, "D3 D3")]
+        [InlineData("10.000°N 23.500°E", 10.0, 23.5, 12345, "D3N D3E")]
+        [InlineData("-20.000°S -123.500°W", -20.0, -123.5, 12345, "D3N D3E")] // Like this probably unexpected (with the sign), but correct for decimal output
+        [InlineData("20.000°S 123.500°W", -20.0, -123.5, 12345, "U3N U3E")] // Therefore we have this
```
krwq, line 52: why is negative and positive giving same result?

krwq: or perhaps it's not, can you explain difference between U3N and D3N?

src/devices/Nmea0183/Ais/AisParser.cs
```diff
@@ -0,0 +1,296 @@
...
+            }
+
+            var sentenceParts = sentenceWithoutChecksum.Split(',');
+            var packetHeader = sentenceParts[0];
+            if (!ValidPacketHeader(packetHeader))
+            {
+                throw new FormatException($"Unrecognised message: packet header {packetHeader}: {sentence}");
+            }
+
+            // var radioChannelCode = sentenceParts[4];
```
krwq, line 108: we should remove commented out code

src/devices/Nmea0183/Ais/AisSafetyState.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// The ship is moving away from us or it is very far away
+        /// </summary>
+        Safe,
+
+        /// <summary>
+        /// There's an imminent risk of collision with this ship
+        /// </summary>
+        Dangerous,
```
krwq, line 25: Should this be InDanger? Danger suggests it's a threat rather than being in danger

src/devices/Nmea0183/Ais/AisTarget.cs
```diff
@@ -0,0 +1,190 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
```
krwq, line 3: If Ais folder is something you copied from somewhere else please add a comment on top of where was it copied from and give credit to authors (can't remember if this was Ais or something else you mentioned was copied)

src/devices/Nmea0183/Ais/AisTarget.cs
```diff
@@ -0,0 +1,190 @@
...
+        public string FormatMmsi()
+        {
+            string m = Mmsi.ToString(CultureInfo.InvariantCulture);
+
+            // Normally, an MMSI should start with at most two zeros, but some targets report otherwise illegal 4-digit MMSI numbers.
+            while (m.Length < 9)
+            {
+                m = "0" + m;
+            }
+
```
krwq, line 112: https://learn.microsoft.com/en-us/dotnet/api/system.string.padleft?view=netcore-3.1#system-string-padleft(system-int32-system-char)

src/devices/Nmea0183/Ais/MmsiType.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// </summary>
+        Mob,
+
+        /// <summary>
+        /// Similar to the above. Sound an alarm!
+        /// </summary>
+        Epirb,
+
+        /// <summary>
+        /// Defined, but probably not used.
```
krwq, line 67: Honestly - I'd refrain from adding `Rarely used` and similar in the summary - remark might be a better place for that.

If ShouldTriggerAlarm would be useful utility consider adding it somewhere instead and add generic remark in the enum  rather than explaining that individually

src/devices/Nmea0183/Ais/PayloadDecoder.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Original code taken from https://github.com/yellowfeather/AisParser, under MIT License
```
krwq, line 4: would be good to create an issue in that repo and check if them if they could provide something suitable we could directly re-use or some shared component so that both repos don't maintain it

src/devices/Common/tests/GeographicPositionTest.cs
pgrawehr: Not here. It is customary to always give all digits when writing or transmitting coordinates, as this prevents errors. Coordinates often need to be transmitted over poor radio channels.

src/devices/Common/tests/GeographicPositionTest.cs
pgrawehr: "D" is latitude/longitude with sign, "U" is without sign. The sign is typically not used when N/S or E/W is printed. So "D3N" is, while technically possible, not really useful, one would use U3N instead (or just D3 without N)

src/devices/Nmea0183/Ais/AisSafetyState.cs
pgrawehr: That's what's meant: That other ship is dagerous to us. A "dangerous AIS target" is one that I have on my screen and is dangerous to me.

src/devices/Nmea0183/Ais/AisTarget.cs
pgrawehr: I did this, but this particular file is entirely from me.

src/devices/Nmea0183/Ais/PayloadDecoder.cs
pgrawehr: I'll open an issue there, but given the number of bugs I fixed (and the additions I needed), the code is more or less a rewrite, with the exception of the message definitions. The other project seems unmaintened, so I strongly doubt there will be the required support.

src/devices/Nmea0183/Ais/MmsiType.cs
pgrawehr: Ok, moved that to the remark section above. There is an automatic alarm triggering mechanism available on the `AisManager` class. That one triggers (among other things) when any of these are in range.

joperezr:
[Triage]: Hey @krwq can you PTAL?

src/devices/Nmea0183/AisManager.cs
```diff
@@ -0,0 +1,927 @@
...
+        /// Gets the list of all active targets of the given type
+        /// </summary>
+        /// <typeparam name="T">A type of target, must be a derivative of <see cref="AisTarget"/>.</typeparam>
+        /// <returns>An enumeration of all targets of that type</returns>
+        public IEnumerable<T> GetSpecificTargets<T>()
+        where T : AisTarget
+        {
+            lock (_lock)
+            {
+                return _targets.Values.Where(x => x is T).Cast<T>();
```
raffaeler, line 332: You can directly use `.OfType<T>()` and remove both `.Where` and `.Cast`

raffaeler (approved):
I don't know anything about GPS, hopefully the tests cover enough to ensure the algorithm functionalities.
My comments are purely cosmetic.
Overall, great addition!

src/devices/Nmea0183/Ais/PayloadDecoder.cs
pgrawehr: No reaction so far.

src/devices/Nmea0183/Ais/Payload.cs
```diff
@@ -0,0 +1,325 @@
...
+            var result = Convert.ToInt32(bitValue.Substring(1), 2);
+            if (bitValue.StartsWith("1"))
+            {
+                result = (int)(result - Math.Pow(2, bitValue.Length - 1));
+            }
+
+            return result;
+        }
+
+        public void WriteInt(int var, int length)
```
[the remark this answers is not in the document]
pgrawehr: I'll look over it. Note that in this case, `var` is the name of the variable, not the type...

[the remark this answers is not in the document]
raffaeler: > I'll look over it. Note that in this case, `var` is the name of the variable, not the type...

Yes, I understood that. It is just a aesthetic issue, nothing more.

THIRD-PARTY-NOTICES.TXT
richlander: Yes.

src/devices/Nmea0183/AisManager.cs
pgrawehr: Nice! I didn't know this one.

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -132,6 +132,37 @@ public static void GetDegreesMinutesSeconds(double angle, int secDigits, out dou
             }
         }
 
+        /// <summary>
+        /// Returns the given angle as degrees, minutes and seconds
```
krwq, line 136:
```suggestion
        /// Returns the given angle as degree and minutes
```

not seeing seconds anywhere in the out params

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -132,6 +132,37 @@ public static void GetDegreesMinutesSeconds(double angle, int secDigits, out dou
...
 
+        /// <summary>
+        /// Returns the given angle as degree and minutes
+        /// </summary>
+        /// <param name="angle">Input angle, in degrees</param>
+        /// <param name="minDigits">Number of digits for the minutes</param>
+        /// <param name="normalizedVal">Normalized angle value (to -180 to 180)</param>
+        /// <param name="degrees">Full degrees</param>
+        /// <param name="minutes">Minutes including requested number of digits</param>
+        public static void GetDegreesMinutes(double angle, int minDigits, out double normalizedVal, out double degrees, out double minutes)
```
krwq, line 143: should degrees be `int`?

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
```diff
@@ -132,6 +132,37 @@ public static void GetDegreesMinutesSeconds(double angle, int secDigits, out dou
...
+        {
+            angle = GeographicPositionExtensions.NormalizeAngleTo180Degrees(angle);
+            normalizedVal = angle;
+            angle = Math.Abs(angle);
```
krwq, line 147: what happens with sign?

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
pgrawehr: Good catch. Double doesn't make sense there.

src/devices/Common/Iot/Device/Common/GeographicPosition.cs
pgrawehr: It's discarded here. If you need it, you get it back in the `normalizedVal` parameter. A position is normally printed without a sign but instead with N/S or E/W preprended. So instead of `-30° 20' 23.2"` one writes `30° 20' 23.2" S` (and certainly not  `-30° 20' 23.2" S` or `30° -20' 23.2"`).

src/devices/Nmea0183/Ais/AisParser.cs
```diff
@@ -0,0 +1,297 @@
...
+        {
+            var data = sentence.Substring(1);
+
+            var checksum = 0;
+            foreach (var ch in data)
+            {
+                checksum ^= (byte)ch;
+            }
+
+            return Convert.ToInt32(checksum.ToString("X"), 16);
```
krwq, line 289: I think I'm missing something on this line, are we converting to base 16 string and then convert it back to int?

This CalculateSum seems a bit bad from my perspective, i.e. "aa" will produce checksum 0. "bb" as well.... The answer seems to be always 1 byte, it seems it's missing some shift somewhere

pgrawehr: I removed this implementation, because it was a duplicate. The algorithm however was correct. This checksum is defined as the XOR of all the characters in the (ASCII) string. Not a very solid check, indeed.

src/devices/Nmea0183/NmeaTcpClient.cs
```diff
@@ -0,0 +1,165 @@
...
+using Iot.Device.Common;
+using Iot.Device.Nmea0183.Sentences;
+using Microsoft.Extensions.Logging;
+
+namespace Iot.Device.Nmea0183
+{
+    /// <summary>
+    /// A TCP Server bidirectional sink and source. Provides NMEA sentences to each connected client.
+    /// </summary>
+    public class NmeaTcpClient : NmeaSinkAndSource
```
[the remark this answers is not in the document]
pgrawehr: The default implementation of the parser works with streams only. These abstractions are the most common transport options and very useful if connecting several sources together. These abstractions really make it plug-and-play to connect different sources together. Note also that we already have the corresponding server class, so having the symmetric client is just appropriate.

src/System.Device.Gpio/CompatibilitySuppressions.xml
```diff
@@ -1,103 +1,103 @@
...
+  <Suppression>
+    <DiagnosticId>CP0008</DiagnosticId>
+    <Target>T:System.Device.Spi.DataFlow</Target>
     <Left>lib/netstandard2.0/System.Device.Gpio.dll</Left>
     <Right>lib/net6.0-windows10.0.17763/System.Device.Gpio.dll</Right>
   </Suppression>
   <Suppression>
     <DiagnosticId>CP0002</DiagnosticId>
-    <Target>M:System.Device.Gpio.PinEventTypes.#ctor</Target>
+    <Target>M:System.Device.Spi.DataFlow.#ctor</Target>
```
krwq, line 15: can you revert these? I don't think you've made any changes to System.Device.Gpio, this shouldn't be needed

src/devices/Nmea0183/Ais/AisParser.cs
```diff
@@ -0,0 +1,284 @@
...
+                        _fragments.Remove(messageId);
+                        return _payloadDecoder.Decode(encodedPayload, numFillBits);
+                    }
+                }
+
+                return null; // More parts expected
+            }
+        }
+
+        public int ExtractChecksum(string sentence, int checksumIndex)
```
krwq, line 273: nit: checksumIndex seems to be misnamed given we're adding `+`, probably should be called `dataLength` or something

src/devices/Nmea0183/NmeaTcpClient.cs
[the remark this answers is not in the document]
krwq: We generally should avoid adding utilities like this anywhere. I'd rather go with generic purpose utilities which solve this problem than adding helpers in every library...

src/devices/Nmea0183/SentenceCache.cs
```diff
@@ -92,35 +149,54 @@ public bool TryGetCurrentPosition(out GeographicPosition? position, out Angle tr
...
-            var gll = (PositionFastUpdate?)GetLastSentence(PositionFastUpdate.Id);
-            var gga = (GlobalPositioningSystemFixData?)GetLastSentence(GlobalPositioningSystemFixData.Id);
-            var rmc = (RecommendedMinimumNavigationInformation?)GetLastSentence(RecommendedMinimumNavigationInformation.Id);
-            var vtg = (TrackMadeGood?)GetLastSentence(TrackMadeGood.Id);
-            var hdt = (HeadingTrue?)GetLastSentence(HeadingTrue.Id);
+            var gll = (PositionFastUpdate?)GetLastSentence(source, PositionFastUpdate.Id);
+            var gga = (GlobalPositioningSystemFixData?)GetLastSentence(source, GlobalPositioningSystemFixData.Id);
+            var rmc = (RecommendedMinimumNavigationInformation?)GetLastSentence(source, RecommendedMinimumNavigationInformation.Id);
+            var vtg = (TrackMadeGood?)GetLastSentence(source, TrackMadeGood.Id);
+            var hdt = (HeadingTrue?)GetLastSentence(source, HeadingTrue.Id);
```
krwq, line 173: nit: given this code is complex enough I'd generally not recommend using abbreviations (feel free to ignore in this PR)

src/devices/Nmea0183/SentenceCache.cs
```diff
@@ -195,6 +273,33 @@ public bool TryGetCurrentPosition(out GeographicPosition? position, bool extrapo
...
+        /// <param name="id">Id to query</param>
+        /// <returns>The last sentence of that type and source, null if not found</returns>
+        public NmeaSentence? GetLastSentence(string? source, SentenceId id)
+        {
+            if (source == null)
+            {
+                return GetLastSentence(id);
+            }
+
+            lock (_lock)
```
krwq, line 289: can this cache grow indefinitely or is there an upper limit?

src/devices/Nmea0183/SentenceCache.cs
```diff
@@ -92,35 +149,54 @@ public bool TryGetCurrentPosition(out GeographicPosition? position, out Angle tr
...
+        /// <param name="source">Only look at this source (otherwise, if multiple sources provide a position, any is used)</param>
         /// <param name="extrapolate">True to extrapolate the current position using speed and track</param>
         /// <param name="track">Track (course over ground)</param>
         /// <param name="sog">Speed over ground</param>
         /// <param name="heading">Vessel Heading</param>
+        /// <param name="messageTime">Time of the position report that was used</param>
+        /// <param name="now">The current time (when working with data in the past, this may be the a time within that data set)</param>
         /// <returns>True if a valid position is returned</returns>
-        public bool TryGetCurrentPosition(out GeographicPosition? position, bool extrapolate, out Angle track, out Speed sog, out Angle? heading)
+        public bool TryGetCurrentPosition(
```
krwq, line 160: I might be missing something but why is this function in the SentenceCache? I'd expect this code to only do caching without any extra logic

src/devices/Nmea0183/Sentences/EngineData.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using UnitsNet;
+
+namespace Iot.Device.Nmea0183.Sentences
+{
+    /// <summary>
+    /// A helper DTO to transfer engine data in one blob
+    /// </summary>
+    [Serializable]
```
krwq, line 16: why do we need the attribute?

src/devices/Nmea0183/Sentences/EngineData.cs
```diff
@@ -0,0 +1,98 @@
...
+                detail.OperatingTime, detail.Temperature);
+        }
+
+        /// <summary>
+        /// The NMEA2000 bus timestamp (not really relevant, I think)
+        /// </summary>
+        public int MessageTimeStamp
+        {
+            get;
+        }
```
krwq, line 56: nit: I'd suggest more succint form in this file:

```suggestion
        public int MessageTimeStamp { get; }
```

src/devices/Nmea0183/Sentences/NmeaSentence.cs
```diff
@@ -67,6 +67,7 @@ protected NmeaSentence(TalkerId talker, SentenceId id, DateTimeOffset time)
         public TalkerId TalkerId
         {
             get;
+            set;
```
krwq, line 70: why settable? could this be i.e. `init` or `required` property?

src/devices/Nmea0183/Sentences/NmeaSentence.cs
```diff
@@ -86,7 +87,7 @@ public DateTimeOffset DateTime
             {
                 return _dateTime;
             }
-            protected set
+            set
```
krwq, line 90: same question

src/devices/Nmea0183/Sentences/ProprietaryMessage.cs
```diff
@@ -0,0 +1,110 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using SixLabors.ImageSharp;
```
krwq, line 10: ImageSharp here?

src/devices/Nmea0183/Sentences/ProprietaryMessage.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            if (length == 4 && inverseEndianess)
+            {
+                part = part.Substring(2, 2) + part.Substring(0, 2);
+            }
+
+            if (length == 8 && inverseEndianess)
+            {
+                part = part.Substring(6, 2) + part.Substring(4, 2) + part.Substring(2, 2) + part.Substring(0, 2);
+            }
```
krwq, line 98: could you instead TryParse first as UInt32 then if needed:

```csharp
public static UInt32 ReverseBytes(UInt32 value)
{
     return (value & 0x000000FFU) << 24 | (value & 0x0000FF00U) << 8 |
         (value & 0x00FF0000U) >> 8 | (value & 0xFF000000U) >> 24;
}
```

and cast to int in the end?

src/devices/Nmea0183/tests/Ais/AidToNavigationReportMessageTests.cs
```diff
@@ -0,0 +1,85 @@
...
+using Xunit;
+
+namespace Iot.Device.Nmea0183.Tests.Ais
+{
+    public class AidToNavigationReportMessageTests : MessageTestBase
+    {
+        [Fact]
+        public void Should_parse_message()
+        {
+            const string sentence = "!AIVDM,1,1,,B,ENk`sR9`92ah97PR9h0W1T@1@@@=MTpS<7GFP00003vP000,2*4B";
```
krwq, line 16: If tests are copied from anywhere please makes sure to add a comment on top

src/devices/Nmea0183/tests/Ais/AisManagerTest.cs
```diff
@@ -0,0 +1,415 @@
...
+            _manager.TrackEstimationParameters.MaximumPositionAge.ShouldBeEquivalentTo(TimeSpan.FromSeconds(20));
+            _manager.DimensionToStern.ShouldBeEquivalentTo(Length.Zero);
+            _manager.OwnMmsi.ShouldBeEquivalentTo(269110660u);
+            _manager.OwnShipName.ShouldNotBeEmpty();
+        }
+
+        [Fact]
+        public void FeedWithRealDataAndCheckGeneralAttributes()
+        {
+            using NmeaLogDataReader reader = new NmeaLogDataReader("Reader", "../../../Nmea-2021-08-25-16-25.txt");
```
krwq, line 47: We should probably copy this file to output directory, we should not try to guess the paths with `..`

src/devices/Nmea0183/tests/Ais/AisManagerTest.cs
```diff
@@ -0,0 +1,415 @@
...
+
+            Assert.True(relativePos1.ClosestPointOfApproach < relativePos1.Distance);
+            Assert.Equal(1790, relativePos1.TimeToClosestPointOfApproach(latestPacketDate)!.Value.TotalSeconds, 0);
+        }
+
+        [Fact]
+        public void CheckSafetyPermanently()
+        {
+            // This does a safety check all the time. Very expensive...
+            using NmeaLogDataReader reader = new NmeaLogDataReader("Reader", "../../../Nmea-2021-08-25-16-25.txt");
```
krwq, line 163: might be worth creating utility `GetDataLogger` so you can change this easier everywhere if needed

src/devices/Nmea0183/tests/NetworkPortScanner.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Net;
+using System.Text;
+
+namespace Iot.Device.Nmea0183.Tests
+{
+    /// <summary>
+    /// This class helps finding a free network port for a test.
+    /// Could be moved to the real library, but with what name?
+    /// </summary>
+    internal class NetworkPortScanner
```
krwq, line 19: I'd really prefer we did not have to use TCP or UDP anywhere in the product and not have to test it this way

krwq: (these kind of stuff tends to be flaky and networking code is always a bit fragile and prone to errors)

src/devices/Nmea0183/tests/GeodeticFunctionTests.cs
```diff
@@ -293,8 +293,8 @@ public void RouteCalculationInverse()
         {
             GeographicPosition start = new GeographicPosition(0, 1, 0);
             GeographicPosition end = new GeographicPosition(0, 2, 0);
-            var points = GreatCircle.CalculateRoute(start, end, 1000);
-            Assert.Equal(112, points.Count);
+            var points = GreatCircle.CalculateRoute(start, end, Length.FromMeters(1000));
+            Assert.Equal(113, points.Count);
```
krwq, line 297: why did this test need adjustment? was it incorrect before?

krwq (approved):
Sorry this took so long but it was really large PR 😄 

Generally speaking looks good to me, there is only one blocking item left before we merge:
- ensure to add comment with attribution to original code if needed also in tests (if not needed then ignore this)

As a follow up I'd recommend to clean up this code and remove as much code as possible without losing much functionality:
- remove networking utilities - while they're useful it's better to focus on core abstraction and provide something more general purpose separately
- try to cleanup the code, generalize repeating code etc

src/devices/Nmea0183/SentenceCache.cs
pgrawehr: Yes, this caches the last messages of each type only, and since the list of possible message types is limited (maybe 30), this cannot really grow big.

src/devices/Nmea0183/Sentences/EngineData.cs
pgrawehr: Because DTOs should generally have this attribute.

src/devices/Nmea0183/tests/GeodeticFunctionTests.cs
pgrawehr: IIRC, some change caused the first or last point to be included in the path as well, which is correct. So this change reflects a fix in the implementation.

src/devices/Nmea0183/tests/Ais/AidToNavigationReportMessageTests.cs
pgrawehr: Good catch, forgot to add the comment to the tests. Done now.

src/devices/Nmea0183/Sentences/NmeaSentence.cs
pgrawehr: Changed

src/devices/Nmea0183/Sentences/NmeaSentence.cs
pgrawehr: This one needs to be settable, because there are a few cases where the time of a message is adjusted after its creation (particularly in messages that do not transport the time)

src/devices/Nmea0183/Sentences/ProprietaryMessage.cs
pgrawehr: Isn't much shorter, but probably better to understand. Done

src/devices/Nmea0183/tests/NetworkPortScanner.cs
pgrawehr: I know, but this is the best possibility to really do an integration test on several parts at once. And it _did_ help me find compatibility bugs on different platforms (e.g. macOS has several weird bugs in the UDP code).

src/devices/Nmea0183/tests/Ais/AisManagerTest.cs
pgrawehr: True, I was just to lazy for that...

Now embedded the test data as resources, to avoid this problem.

src/System.Device.Gpio/CompatibilitySuppressions.xml
pgrawehr: Sure, I must have missed this somewhere. This file keeps getting changed with the other one for no apparent reason (this PR doesn't even touch System.Device.Gpio)

src/devices/Nmea0183/Ais/AisParser.cs
pgrawehr: No, that's right here, it's the index of the checksum in the input string, the +1 is for the checksum delimiter.

src/devices/Nmea0183/SentenceCache.cs
pgrawehr: You are right, I separated these methods to a separate class. To much logic in the same class is poor design.

pgrawehr:
> Generally speaking looks good to me, there is only one blocking item left before we merge:
> 
>     * ensure to add comment with attribution to original code if needed also in tests (if not needed then ignore this)

Done.

> 
> 
> As a follow up I'd recommend to clean up this code and remove as much code as possible without losing much functionality:
> 
>     * remove networking utilities - while they're useful it's better to focus on core abstraction and provide something more general purpose separately

I kind of agree. It would be good if we could separate such utilities for general use. We'll have to see how to best achieve that.

> 
>     * try to cleanup the code, generalize repeating code etc

That could be more difficult. There's indeed a lot of boilerplate code here, but since the messages don't follow a general scheme but instead are individually defined, I don't see much potential there. I'm somewhat trying to achieve that with the high-level interfaces, so that (for instance) one can call `TryGetCurrentPosition` and the system automatically combines the data from several input messages.

PR #1941, opened 2022-10-02
Add a few new WeatherHelper functions

pgrawehr:
This PR adds some new functions to the `WeatherHelper` class.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1941)

### DIFF
--- a/src/devices/Common/Iot/Device/Common/WeatherHelper.cs
+++ b/src/devices/Common/Iot/Device/Common/WeatherHelper.cs
@@ -331,5 +331,84 @@ namespace Iot.Device.Common
         }
 
         #endregion
+
+        /// <summary>
+        /// Simplified air density (not taking humidity into account)
+        /// </summary>
+        /// <param name="airPressure">Measured air pressure</param>
+        /// <param name="temperature">Measured temperature</param>
+        /// <returns>Approximate standard air density</returns>
+        public static Density CalculateAirDensity(Pressure airPressure, Temperature temperature)
+        {
+            double gasConstant = 287.058; // J / (kg * K), for dry air
+            var result = airPressure.Pascals / (gasConstant * temperature.Kelvins);
+            return Density.FromKilogramsPerCubicMeter(result);
+        }
+
+        /// <summary>
+        /// Calculates the air density
+        /// </summary>
+        /// <param name="airPressure">Measured air pressure</param>
+        /// <param name="temperature">Measured temperature</param>
+        /// <param name="humidity">Measured relative humidity</param>
+        /// <returns>Approximate standard air density at sea level</returns>
+        /// <remarks>From https://de.wikipedia.org/wiki/Luftdichte </remarks>
+        public static Density CalculateAirDensity(Pressure airPressure, Temperature temperature, RelativeHumidity humidity)
+        {
+            double rs = 287.058;
+            double rd = 461.523;
+            var pd = CalculateSaturatedVaporPressureOverWater(temperature);
+            // It's still called "constant" even though it's not constant
+            double gasConstant = rs / (1 - (humidity.Percent / 100) * (pd.Pascals / airPressure.Pascals) * (1 - (rs / rd)));
+            var result = airPressure.Pascals / (gasConstant * temperature.Kelvins);
+            return Density.FromKilogramsPerCubicMeter(result);
+        }
+
+        /// <summary>
+        /// Calculates the wind chill temperature - this is the perceived temperature in (heavy) winds at cold temperatures.
+        /// This is only useful at temperatures below about 20°C, above use <see cref="CalculateHeatIndex"/> instead.
+        /// Not suitable for wind speeds &lt; 5 km/h.
+        /// </summary>
+        /// <param name="temperature">The measured air temperature</param>
+        /// <param name="windSpeed">The wind speed (measured at 10m above ground)</param>
+        /// <returns>The perceived temperature. Note that this is not a real temperature, and the skin will never really reach
+        /// this temperature. This is more an indication on how fast the skin will reach the air temperature. If the skin
+        /// reaches a temperature of about -5°C, frostbite might occur.</returns>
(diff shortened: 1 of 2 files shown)


src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -331,5 +331,84 @@ public static Length CalculateAltitude(Pressure pressure, Pressure seaLevelPress
...
+
+        /// <summary>
+        /// Simplified air density (not taking humidity into account)
+        /// </summary>
+        /// <param name="airPressure">Measured air pressure</param>
+        /// <param name="temperature">Measured temperature</param>
+        /// <returns>Approximate standard air density</returns>
+        public static Density CalculateAirDensity(Pressure airPressure, Temperature temperature)
+        {
+            double gasConstant = 287.058; // J / (kg * K), for dry air
```
Ellerbach, line 343: I assume that for all the constants used here, the same source for them apply than all the existing calculation? Otherwise, would be great to add any additional source.

Ellerbach (approved):
Looks good to me!

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
pgrawehr: Yes, this is basically the same as below, when humidity is not known.

krwq: can we extract heard-coded repeating numbers to constants in this file?

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -331,5 +331,84 @@ public static Length CalculateAltitude(Pressure pressure, Pressure seaLevelPress
...
+        {
+            if (windSpeed < Speed.Zero)
+            {
+                throw new ArgumentOutOfRangeException(nameof(windSpeed), "The wind speed cannot be negative");
+            }
+
+            double va = temperature.DegreesCelsius;
+            if (windSpeed < Speed.FromKilometersPerHour(1))
+            {
+                // otherwise, the result is complete crap, because the second and third terms of the equation are 0
```
krwq, line 389: you might want to re-word 😄

src/devices/Common/tests/WeatherTests.cs
```diff
@@ -185,5 +185,53 @@ public void GetRelativeHumidityFromActualAirTemperature(double inTemp, double in
...
+        public void CalculateAirDensity(double inPressure, double inTemp, double inHumidity, double expected)
+        {
+            var result = WeatherHelper.CalculateAirDensity(Pressure.FromMillibars(inPressure),
+                Temperature.FromDegreesCelsius(inTemp), RelativeHumidity.FromPercent(inHumidity));
+
+            Assert.Equal(expected, result.KilogramsPerCubicMeter, 4);
+        }
+
+        [Theory]
+        [InlineData(0, 0, 1.75)] // This is to be expected, see WP article
```
krwq, line 214: please link article directly here

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
pgrawehr: @krwq We can, but I think this would only be useful for numbers that occur more than once. Otherwise it doesn't improve readability IMHO.

krwq: I can see the same constant used here and method below

krwq: ah, I see you changed that, thanks!

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -12,6 +12,21 @@ namespace Iot.Device.Common
...
+
+        /// <summary>
+        /// Gas constant of vapor, J / (kg * K)
+        /// </summary>
+        public const double SpecificGasConstantOfVapor = 461.523;
+
+        /// <summary>
+        /// Default atmospheric temperature gradient = 0.0065K/m (or 0.65K per 100m)
+        /// </summary>
+        public const double DefaultTemperatureGradient = 0.0065;
```
krwq, line 28: I'd recommend to keep them private/internal, if we make it public let's at least expose UnitsNet

pgrawehr: There's likely no Unitsnet equivalent for this strange unit.

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
```diff
@@ -331,5 +346,84 @@ public static Length CalculateAltitude(Pressure pressure, Pressure seaLevelPress
...
+            if (windSpeed < Speed.Zero)
+            {
+                throw new ArgumentOutOfRangeException(nameof(windSpeed), "The wind speed cannot be negative");
+            }
+
+            double va = temperature.DegreesCelsius;
+            if (windSpeed < Speed.FromKilometersPerHour(1))
+            {
+                // otherwise, the result is unusable, because the second and third terms of the equation are 0, resulting in a constant offset from the input temperature
+                windSpeed = Speed.FromKilometersPerHour(1);
```
krwq, line 405: would it make sense to just return temperature here instead?

krwq (approved):

src/devices/Common/Iot/Device/Common/WeatherHelper.cs
pgrawehr: No, because if the wind speed is low, the returned temperature is above the input temperature. This is expected and explained in the documentation. Only if the wind speed is zero or almost zero, the formula gets completely off. So this condition basically limits the output temperature to be at most about 2 degrees over the input temperature.

Issue #1940, opened 2022-10-02
Master build is broken after Version updates


pgrawehr:
**Describe the bug**

Running `build -pack` on current master results in these errors:

```
C:\projects\iot\.dotnet\sdk\7.0.100-rc.1.22431.12\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(203,
6): error : Für dieses Projekt kann kein Paket erstellt werden, weil die Paketerstellung deaktiviert wurde. Fügen Sie d
er Projektdatei <IsPackable>true</IsPackable> hinzu, um die Erstellung eines Pakets aus diesem Projekt zu ermöglichen.
[C:\projects\iot\src\devices\Media\VideoDevice\samples\MjpegStream\MjpegStream.csproj]
C:\projects\iot\.dotnet\sdk\7.0.100-rc.1.22431.12\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(221,
5): error NU5104: Eine stabile Version eines Pakets darf keine Abhängigkeit mit einem Vorabrelease aufweisen. Ändern Si
e entweder die Versionsspezifikation der abhängigen Komponente "System.Device.Gpio [2.3.0-dev, )", oder aktualisieren S
ie das Versionsfeld in der NUSPEC-Datei. [C:\projects\iot\tools\ArduinoCsCompiler\samples\WeatherStation\WeatherStation
.csproj]
C:\projects\iot\.dotnet\sdk\7.0.100-rc.1.22431.12\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(221,
5): error NU5104: Eine stabile Version eines Pakets darf keine Abhängigkeit mit einem Vorabrelease aufweisen. Ändern Si
e entweder die Versionsspezifikation der abhängigen Komponente "System.Device.Gpio [2.3.0-dev, )", oder aktualisieren S
ie das Versionsfeld in der NUSPEC-Datei. [C:\projects\iot\tools\ArduinoCsCompiler\ArduinoCsCompiler.csproj]
C:\projects\iot\.dotnet\sdk\7.0.100-rc.1.22431.12\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(221,
5): error NU5104: Eine stabile Version eines Pakets darf keine Abhängigkeit mit einem Vorabrelease aufweisen. Ändern Si
[...]

PR #1942, opened 2022-10-03
Fix build issue when trying to run pack

joperezr:
Fixes https://github.com/dotnet/iot/issues/1940

cc: @pgrawehr 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1942)

### DIFF
--- a/Build.cmd
+++ b/Build.cmd
@@ -1,3 +1,9 @@
 @echo off
-powershell -ExecutionPolicy ByPass -NoProfile -command "& """%~dp0eng\common\Build.ps1""" -restore -build %*"
+setlocal
+
+for %%x in (%*) do (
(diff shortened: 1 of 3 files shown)


pgrawehr (approved):
Thanks, that indeed fixes it. 

It's a bit ugly that our build is not really controlled by the default arguments to build.cmd. Many of the arguments specified in `build /?` don't work as expected (e.g. -test is ignored, because the tests always run), and other important ones are properties and therefore not documented.

joperezr:
Yup, that is a valid concern. `build.cmd` infrastructure mainly comes from Arcade, so that is what gets printed via help. Feel free to log an issue about having the script work as intended by arcade, and I'd be happy to fix it.

PR #1956, opened 2022-10-12
Add HX711 I2C weight sensor module

krwq:
Note: this device is not actually HX711. It's an I2C module based on HX711. It comes by different names and doesn't have official chip name and is called with different names:
- HX711 I2C weight sensor module
- DFRobot KIT0176

I used naming used by https://github.com/DFRobot/DFRobot_HX711_I2C

https://wiki.dfrobot.com/HX711_Weight_Sensor_Kit_SKU_KIT0176

The datasheet for this module is non-existing and code I was porting is not very well written and final result is a combination of C/Python and reverse engineering and I think it ended up much cleaner than original code


krwq:
it's not HX711, it's I2C module on top of HX711 - HX711 is just backend

src/devices/Hx711I2c/Hx711I2c.cs
```diff
@@ -0,0 +1,349 @@
...
+            CalibrationScale = GetAutomaticCalibrationScale();
+        }
+
+        /// <summary>
+        /// Gets I2C address depending on A0 and A1 pin settings.
+        /// </summary>
+        /// <param name="a0">Value of A0 pin.</param>
+        /// <param name="a1">Value of A1 pin.</param>
+        /// <returns>Address of the device.</returns>
+        public static int GetI2cAddress(PinValue a0, PinValue a1)
```
Ellerbach, line 100: If there are 2 addresses, makes sens to have the second as a constant as well

src/devices/Hx711I2c/Hx711I2c.cs
```diff
@@ -0,0 +1,349 @@
...
+using System.Threading;
+using UnitsNet;
+
+namespace Iot.Device.Hx711
+{
+    /// <summary>
+    /// DFRobot KIT0176: I2C 1kg Weight Sensor Kit
+    /// </summary>
+    [Interface("DFRobot KIT0176: I2C 1kg Weight Sensor Kit")]
+    public class Hx711I2c : IDisposable
```
Ellerbach, line 19: So really looks like the same as the one I4ve pointed out. Just your is SPI, the other one is I2C. So would be great to reuse the naming, align some, etc.

src/devices/Hx711I2c/Hx711I2c.cs
krwq: there are 4, also they are physically labelled A0/A1 on this specific module

src/devices/Hx711I2c/Hx711I2c.cs
krwq: it's already mostly similar:
Offset - same
Tare - same
nano: Read; me: GetWeight
then there is couple of things they support and I don't and reverse

they have SampleAveraging I do support but do not expose directly

krwq: also HX711 is GPIO and they seem to just use SPI for implementation. The module I have is only using HX711 as backend

src/devices/Hx711I2c/Hx711I2c.cs
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        /// <summary>
+        /// Gets I2C address depending on A0 and A1 pin settings.
+        /// </summary>
+        /// <param name="a0">Value of A0 pin.</param>
+        /// <param name="a1">Value of A1 pin.</param>
+        /// <returns>Address of the device.</returns>
+        public static int GetI2cAddress(PinValue a0, PinValue a1)
+            => DefaultI2cAddress + ((bool)a0 ? 1 : 0) + ((bool)a1 ? 2 : 0);
```
Ellerbach, line 101: you should as well present the second I2C address as a constant

src/devices/Hx711I2c/Hx711I2c.cs
```diff
@@ -0,0 +1,349 @@
...
+        public bool IgnoreResetButton { get; set; }
+
+        /// <summary>
+        /// Sets the weight (in grams) used for automatic calibration.
+        /// Value is only relevant when CAL button has been pressed.
+        /// </summary>
+        [Property]
+        public ushort AutomaticCalibrationWeight
+        {
+            // It does not seem to be possible to read from this register
```
Ellerbach, line 67: even if not possible, you can store the data you write?
Same for the other ones?

src/devices/Hx711I2c/Hx711I2c.cs
```diff
@@ -0,0 +1,349 @@
...
+                if (watchdog < 0)
+                {
+                    // this never happened during testing
+                    // but given original code could fail reading
+                    // that could cause infinite loop if it somehow happened so
+                    // as a prevention measure better to throw exception
+                    throw new InvalidOperationException("Getting weight failed. Getting bogus result.");
+                }
+
+                sum += (int)weightGrams;
```
Ellerbach, line 257: this is a bit dangerous because of potential overflow. Don't you want to use the float and make the math average every time? few more operations but no risk of overload. A bit of loss in precision but if you're sampling many times, it's because the sensor is not very precise.

src/devices/Hx711I2c/Hx711I2c.cs
```diff
@@ -0,0 +1,349 @@
...
+
+            return Mass.FromGrams((rawReading - Offset) / CalibrationScale);
+        }
+
+        /// <summary>
+        /// Gets average raw reading.
+        /// </summary>
+        /// <param name="samples">Number of times sampling occurs</param>
+        /// <returns>Raw reading value</returns>
+        public float GetRawReading(int samples = 20)
```
Ellerbach, line 236: if you have grams, then why not using UnitsNet?

Ellerbach (approved):
Couple of comments but all up, looks good.

src/devices/Hx711I2c/Hx711I2c.cs
```diff
@@ -0,0 +1,349 @@
...
+            ReadRegister(Hx711I2cRegister.REG_DATA_GET_CALIBRATION, calibrationBytes);
+            Reverse4ByteArray(calibrationBytes);
+            return BitConverter.ToSingle(calibrationBytes, 0);
+        }
+
+        /// <summary>
+        /// Same as Tare but will also blink LED next to RST button.
+        /// </summary>
+        [Command]
+        public void HardwareTare()
```
joperezr, line 149: NIT: Should we just have one Tare button that takes in a bool for blinking the LED?

joperezr (approved):
Left a few minor comments, but looks good otherwise.

pgrawehr:
Check the Readme file for markdown errors.

src/devices/Hx711I2c/Hx711I2c.cs
krwq: I think unless I add all 4 of them the second will be confusing while it may be obvious for me that second is `A0=1, A1=0` some other person might thing of second as `A0=1, A1=1` or `A0=0, A1=1`

src/devices/Hx711I2c/Hx711I2c.cs
krwq: This is not grams, it's more like ADC reading. It will be gram-like after you divide that by CalibrationScale. Gram-like because unless you pick starting point (offset/tare/zero) the weight is kinda meaningless.

It's still useful to expose it so that Offset can be set/stored appropriately if needed. If I use grams here it will be very tempting to use this value where it's not supposed to be

src/devices/Hx711I2c/Hx711I2c.cs
```diff
@@ -0,0 +1,349 @@
...
+        public void Tare()
+        {
+            Offset = GetRawReading();
+        }
+
+        /// <summary>
+        /// Re-initializes the sensor and sets arbitrarly chosen calibration values.
+        /// </summary>
+        [Command]
+        public void ResetSensor()
```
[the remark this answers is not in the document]
krwq: It's not requirement. Actually python code never calls it while C code always does. I think it resets internal registers but it's not mentioned in the spec when it should be used. I added it in case someone encounters any issues

src/devices/Hx711I2c/Hx711I2c.cs
```diff
@@ -0,0 +1,349 @@
...
+        /// <summary>
+        /// Gets data from sensor with scale which was determined with automatic calibration made when CAL button was pressed.
+        /// This value might differ from actual calibration scale used if no measurements have been made after CAL button was pressed.
+        /// </summary>
+        /// <returns>Calibration scale</returns>
+        public float GetAutomaticCalibrationScale()
+        {
+            // TODO: in netstandard2.0 we cannot use ReadOnlySpan overload of BitConverter
+            // TODO: once we drop support we should use Span overload
+            // TODO: once we're at 6.0 or higher we should use BinaryPrimitives.ReadSingleBigEndian
```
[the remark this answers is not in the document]
krwq: honestly it won't make much difference in terms of perf - this sensor reads slow enough that reducing allocation won't help. Possibly async would make more sense but that would make porting to nano harder

src/devices/Hx711I2c/Hx711I2c.cs
krwq: I've changed sum to 64-bit and it should not be possible anymore (SampleAveraging is 32-bit and sample is 24-bit which gives at most 56-bit so 64 should be enough) - having said that if someone uses uint.MaxValue as SampleAveraging this will take days to complete (I won't question what user wants though).

src/devices/Hx711I2c/Hx711I2c.cs
krwq: I actually don't initially write anything for these two but I suppose I could initialize them in the ctor

src/devices/Hx711I2c/samples/Program.cs
```diff
@@ -0,0 +1,33 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Device.I2c;
+using System.Threading;
+using Iot.Device.Ft4222;
+using Iot.Device.Hx711;
+
+List<Ft4222Device> ft4222s = Ft4222Device.GetFt4222();
```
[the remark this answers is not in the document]
krwq: Honestly, we should eventually add article to every single I2C/SPI device how to use it. I can add a comment how you'd do that without FT4222 - generally I prefer to use samples I tested and I don't have any RPis handy at the moment

src/devices/Hx711I2c/Hx711I2c.cs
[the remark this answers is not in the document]
krwq: I'll remove it from the sample. If we see people start seeing issues without it we can adjust sample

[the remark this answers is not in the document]
krwq: Ok, I did more experimentation - it seems this needs to be called once after powering up. I'll move this to ctor

src/devices/Hx711I2c/Hx711I2c.cs
krwq: I added also SampleAveraging to match nano closer (also I found I needed that to get faster but less accurate sampling)

src/devices/Hx711I2c/Hx711I2c.cs
```diff
@@ -0,0 +1,394 @@
...
+using System;
+using System.Buffers.Binary;
+using System.Device.Gpio;
+using System.Device.I2c;
+using System.Device.Model;
+using System.Diagnostics;
+using System.Threading;
+using UnitsNet;
+
+namespace Iot.Device.Hx711
```
raffaeler, line 13: What about setting the namespace declaration on a single line? (Terminating it with ';')

PR #1959, opened 2022-10-18
mcp25xx add read and write methods

Andrew25455:
Test only on mcp2515 
For mcp25xx add methods for setting, work with buffers, read and write

### DIFF
--- a/src/devices/Mcp25xxx/Bitrates.cs
+++ b/src/devices/Mcp25xxx/Bitrates.cs
@@ -6,7 +6,7 @@ namespace Iot.Device.Mcp25xxx
     /// <summary>
     /// Bitrates for CAN bus
     /// </summary>
-    public enum Bitrates
+    public enum Bitrates : byte
     {
         /// <summary>
         /// Speed 8MHz 1000kBPS CFG1
--- a/src/devices/Mcp25xxx/Bitrates.cs
+++ b/src/devices/Mcp25xxx/Bitrates.cs
@@ -1,0 +1,596 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Mcp25xxx
+{
+    /// <summary>
+    /// Bitrates for CAN bus
+    /// </summary>
+    public enum Bitrates
+    {
+        /// <summary>
+        /// Speed 8MHz 1000kBPS CFG1
+        /// </summary>
+        MCP_8MHz_1000kBPS_CFG1 = 0x00,
+
+        /// <summary>
+        /// Speed 8MHz 1000kBPS CFG2
+        /// </summary>
+        MCP_8MHz_1000kBPS_CFG2 = 0x80,
+
+        /// <summary>
+        /// Speed 8MHz 1000kBPS CFG3
+        /// </summary>
+        MCP_8MHz_1000kBPS_CFG3 = 0x80,
+
+        /// <summary>
+        /// Speed 8MHz 500kBPS CFG1
+        /// </summary>
+        MCP_8MHz_500kBPS_CFG1 = 0x00,
+
+        /// <summary>
+        /// Speed 8MHz 500kBPS CFG2
+        /// </summary>
+        MCP_8MHz_500kBPS_CFG2 = 0x90,
+
+        /// <summary>
+        /// Speed 8MHz 500kBPS CFG3
+        /// </summary>
+        MCP_8MHz_500kBPS_CFG3 = 0x82,
+
+        /// <summary>
+        /// Speed 8MHz 250kBPS CFG1
+        /// </summary>
+        MCP_8MHz_250kBPS_CFG1 = 0x00,
+
+        /// <summary>
+        /// Speed 8MHz 250kBPS CFG2
+        /// </summary>
+        MCP_8MHz_250kBPS_CFG2 = 0xB1,
+
+        /// <summary>
+        /// Speed 8MHz 250kBPS CFG3
+        /// </summary>
+        MCP_8MHz_250kBPS_CFG3 = 0x85,
+
+        /// <summary>
+        /// Speed 8MHz 200kBPS CFG1
+        /// </summary>
+        MCP_8MHz_200kBPS_CFG1 = 0x00,
+
+        /// <summary>
+        /// Speed 8MHz 200kBPS CFG2
+        /// </summary>
+        MCP_8MHz_200kBPS_CFG2 = 0xB4,
+
+        /// <summary>
+        /// Speed 8MHz 200kBPS CFG3
+        /// </summary>
+        MCP_8MHz_200kBPS_CFG3 = 0x86,
+
+        /// <summary>
+        /// Speed 8MHz 125kBPS CFG1
+        /// </summary>
+        MCP_8MHz_125kBPS_CFG1 = 0x01,
+
+        /// <summary>
+        /// Speed 8MHz 125kBPS CFG2
+        /// </summary>
+        MCP_8MHz_125kBPS_CFG2 = 0xB1,
+
+        /// <summary>
+        /// Speed 8MHz 125kBPS CFG3
+        /// </summary>
+        MCP_8MHz_125kBPS_CFG3 = 0x85,
+
+        /// <summary>
+        /// Speed 8MHz 100kBPS CFG1
+        /// </summary>
+        MCP_8MHz_100kBPS_CFG1 = 0x01,
+
+        /// <summary>
+        /// Speed 8MHz 100kBPS CFG2
+        /// </summary>
+        MCP_8MHz_100kBPS_CFG2 = 0xB4,
+
+        /// <summary>
+        /// Speed 8MHz 100kBPS CFG3
+        /// </summary>
+        MCP_8MHz_100kBPS_CFG3 = 0x86,
+
+        /// <summary>
+        /// Speed 8MHz 80kBPS CFG1
+        /// </summary>
+        MCP_8MHz_80kBPS_CFG1 = 0x01,
+
+        /// <summary>
+        /// Speed 8MHz 80kBPS CFG2
+        /// </summary>
+        MCP_8MHz_80kBPS_CFG2 = 0xBF,
+
+        /// <summary>
+        /// Speed 8MHz 80kBPS CFG3
+        /// </summary>
+        MCP_8MHz_80kBPS_CFG3 = 0x87,
+
+        /// <summary>
+        /// Speed 8MHz 50kBPS CFG1
+        /// </summary>
+        MCP_8MHz_50kBPS_CFG1 = 0x03,
+
+        /// <summary>
+        /// Speed 8MHz 50kBPS CFG2
+        /// </summary>
+        MCP_8MHz_50kBPS_CFG2 = 0xB4,
+
+        /// <summary>
+        /// Speed 8MHz 50kBPS CFG3
+        /// </summary>
+        MCP_8MHz_50kBPS_CFG3 = 0x86,
+
+        /// <summary>
+        /// Speed 8MHz 40kBPS CFG1
+        /// </summary>
+        MCP_8MHz_40kBPS_CFG1 = 0x03,
+
+        /// <summary>
+        /// Speed 8MHz 40kBPS CFG2
+        /// </summary>
+        MCP_8MHz_40kBPS_CFG2 = 0xBF,
+
+        /// <summary>
+        /// Speed 8MHz 40kBPS CFG3
+        /// </summary>
+        MCP_8MHz_40kBPS_CFG3 = 0x87,
+
+        /// <summary>
+        /// Speed 8MHz 33k3BPS CFG1
+        /// </summary>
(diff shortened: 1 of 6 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1959) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1959) before we can accept your contribution.<br/><br/>:x: Andrew25455 [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=1959)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=1959) it.</sub>

pgrawehr:
@Andrew25455 You can ignore failures of the build step "Run Helix tests". These are hardware tests that are unfortunately a bit flaky. Unless you change something in the System.Device.Gpio assembly, it's very unlikely you break those.

Andrew25455:
Thank you Patrick! I wanted green pipeline for publish nuget for tests, but I will test with dll's.

Andrew25455:
Tested. It works correctly with board - https://www.waveshare.com/wiki/2-CH_CAN_HAT

src/devices/Mcp25xxx/Mcp25xxx.cs
```diff
@@ -241,6 +248,74 @@ public void Reset()
...
+                    operationMode));
+        }
+
+        /// <summary>
+        /// Read messages from all buffers
+        /// </summary>
+        /// <param name="byteCount">Number of bytes to read.  This must be one or more to read.</param>
+        /// <returns>Array of messages</returns>
+        /// <exception cref="ArgumentOutOfRangeException"></exception>
+        public byte[][] ReadMessages(int byteCount)
```
pgrawehr, line 295: `byte[][]` as return type here is a bit ugly to use for the caller. Can we use `List<byte[]>` instead (as it is a list of messages, not a matrix)? Or maybe even something more elegant?

Also, I'm confused about why we need to specify `byteCount` here. As far as I know, the length of each CAN-Bus message is encoded in the data stream. How would an user know how many bytes to read before looking at the messages?

pgrawehr (commented):
Looks good overall, thank you so far!

Just a few comments and questions. How did you test this, by the way? What did you connect to the CAN Bus interface?

src/devices/Mcp25xxx/Bitrates.cs
```diff
@@ -0,0 +1,596 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Mcp25xxx
+{
+    /// <summary>
+    /// Bitrates for CAN bus
+    /// </summary>
+    public enum Bitrates : byte
```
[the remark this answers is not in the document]
pgrawehr: ... and since it's not a flag field, it should probably be called `Bitrate` instead.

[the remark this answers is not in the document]
Ellerbach: And also as it's a public enum, the naming convention should apply. So `MCP_8MHz_1000kBPS_CFG1` should become `Mcp8MHz1000kBpsConfiguration1`

src/devices/Mcp25xxx/Mcp25xxx.cs
```diff
@@ -241,6 +248,74 @@ public void Reset()
...
+        {
+            WriteByte(Address.Cnf1, (byte)cnf1);
+            WriteByte(Address.Cnf2, (byte)cnf2);
+            WriteByte(Address.Cnf3, (byte)cnf3);
+        }
+
+        /// <summary>
+        /// Set mode
+        /// </summary>
+        public void SetMode(OperationMode operationMode)
```
Ellerbach, line 278: you are missing comment for the parameters

src/devices/Mcp25xxx/TxBuffer.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Mcp25xxx
+{
+    /// <summary>
+    /// TX Buffers
```
Ellerbach, line 7: can you try to be a bit more explicit on what TX means? I guess it's transmission. So `Transmission buffers` may help in understanding the elements

Andrew25455:
Thank you for your review! I fixed comments. Refactor Bitrates enum to method with data in Dictionary.

I was tested this library with two Rpi with two CAN hat ([photo](https://drive.google.com/file/d/1qazUTakWuwv-p4HWKY722uCBNYA-5og7/view?usp=sharing)) . One of this use your library and another - can-utils.  In another test i connected to Rpi with hat our custom board with with C++ firmware, and it worked too.

pgrawehr:
One question still: How does one know how many bytes to read? Is the transmission just like a stream or is it packet-based?

Andrew25455:
CAN bus is packet-based protocol. In mcp2515  received message will be in buffer. But buffer will contain cropped message only with id, data length code and data. Max size of this cropped message is id = 32bit, data length = 4bit and data can be 8*8bit  => 12,5byte. I added TransmitBufferMaxSize constant 14byte, because it is max chip SRAM per buffer.

pgrawehr:
> CAN bus is packet-based protocol. In mcp2515 received message will be in buffer. But buffer will contain cropped message only with id, data length code and data. Max size of this cropped message is id = 32bit, data length = 4bit and data can be 8*8bit, total - 12,5byte. I added TransmitBufferMaxSize constant 14byte, because it is max chip SRAM per buffer.

Ok, that makes sense. What reason would one then have to call `ReadMessages()` with an argument of less than `TransmitBufferMaxSize`? The reason I'm asking is just because we should, if possible, avoid to offer features that could cause unnecessary confusion for the users.

Andrew25455:
Sorry for the delay, i thought a lot about your message. 
As result, i change methods contract from bytes to CanMessage object. 
In this object i also add common can Id and extended Id, witch i have forgotten

src/devices/Mcp25xxx/Mcp25xxx.cs
```diff
@@ -292,11 +293,21 @@ public void SetMode(OperationMode operationMode)
...
-        /// Read messages from all buffers
+        /// Read arrived messages
         /// </summary>
-        /// <param name="byteCount">Number of bytes to read. This must be one or more to read.</param>
         /// <returns>List of messages received</returns>
-        public List<byte[]> ReadMessages(int byteCount = TransmitBufferMaxSize)
+        public CanMessage[] ReadMessages()
+        {
+            var buffersData = ReadAllBuffers(TransmitBufferMaxSize);
+            return buffersData.Select(CanMessage.CreateFromBytes).ToArray();
```
pgrawehr, line 302: The only problem I see here is that this will now throw an `ArgumentException` when (for whatever reason) the input data is garbage. This is confusing and should probably be wrapped into some other error.

pgrawehr:
> Sorry for the delay, i thought a lot about your message. As result, i change methods contract from bytes to CanMessage object. In this object i also add common can Id and extended Id, witch i have forgotten

No hurry, this is going to get very good. I like the new approach, but you should put an eye on possible error cases. Now an invalid message may cause `ReadMessage()` to throw an `ArgumentException`, which is confusing, as this is clearly not an user error. Maybe you add some unit tests for these error cases?

Andrew25455:
I think it was bad idea - return CanMessage object from read method, because if converting of first message throw exception then second message will be lost. I see only one solution - return raw byte array to user. 
Patrick, do you have a solution, maybe?

pgrawehr:
> I think it was bad idea - return CanMessage object from read method, because if converting of first message throw exception then second message will be lost. I see only one solution - return raw byte array to user. Patrick, do you have a solution, maybe?

Honestly, I don't like it very much that the method currently returns a List of elements which were possibly received from different interfaces. Most of the time, the two interfaces will probably connect to very different networks, and the user wants to know on which interface a message was received. So if you split that (add a "interfaceNo" parameter to the ReadMessage function), an error would throw exactly where it should be.

Ellerbach:
I would go for something in line with Patrick's suggestion. In that case the exception should be transformed into an error message. So you would return a list of a class that may look like this (just as an idea):
```csharp
List<CanMessageList> ReadMessages()

CanMessageList 
{
    public CanMessage CanMessage { get; set; } // Can be null when there is an error
    public ReceivedMessageType ReceivedMessageType {get; set; }
    public CanError CanError { get; set; }
}

enum CanError
{
    NoError,
    SomeError,
    Etc;
}
```

Andrew25455:
Thank you for your ideas,  
Patrick, in your solution our users need to know about buffers and their quantity. I think we need to provide more high-level interface for our users.
Laurent, honestly, I don't like this pattern as much as the pattern to throw an exception :)

When I tried to add your solutions in code, I found more elegant solution, as I think. I separated CanMessage to two models. One of this is model for sending message and another - for received message.  First model can throw exception when model is initialising , another - can throw exception when user call methods GetId or GetData. 
What about this solution?

Andrew25455:
@pgrawehr Could you review this?

src/devices/Mcp25xxx/McpBitrate.cs
```diff
@@ -0,0 +1,232 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Iot.Device.Mcp25xxx
+{
+    /// <summary>
+    /// Standard Frequency and Speed for CAN bus
+    /// </summary>
+    public enum CanBusFrequencyAndSpeed
```
Ellerbach, line 12: please have the enem and the class in different files to increase readability

Ellerbach (approved):
Looks good to me now. Thanks!

pgrawehr:
@Andrew25455 Is it intentional that now still ReadMessages() throws an exception when illegal data is received? If so, it must at least be documented. 

Additionally, I would like to see a ReadMessage(bufferNo) method that returns only the data of one buffer. Your method doesn't allow finding out on which port the message was received. You could also add the receiving interface as a property to ReceivedCanMessage.

Andrew25455:
Patrick, I can not find exceptions, with will throw when illegal data is received.  I extracted all checks and exceptions to GetId() and GetData() methods, for user call after receive message.

I think that `ReadMessage(bufferNo) ` method will be repeat `ReadRxBuffer(RxBufferAddressPointer addressPointer, int byteCount = 1)` method and my goal was hide technical details like buffer. Because of this i don't want to add type of buffer property to ReceivedCanMessage.

pgrawehr:
@Andrew25455 I had a first look into this. I did update the test a bit, as you where never really testing reads. Your simulation device tested the readBuffer as output, but it should be an input (for the class under test). 

I can confirm I didn't get it to crash on wrong returns. 

For the other problem, I'm still checking options. Looking at the main class, we have the following methods:

```csharp
public ReceivedCanMessage[] ReadMessages();
public List<byte[]> ReadAllBuffers(int byteCount);
public byte[] ReadRxBuffer(RxBufferAddressPointer addressPointer, int byteCount = 1);
public byte Read(Address address);
```

The first three all do basically the same. For the user, this is confusing, there should normally only be one way to do things, unless there's a reason. And the last one should probably be private anyway. 

I'll try to provide a suggestion later.

Andrew25455:
> I see your point, but I think the receiver channel is not a technical detail (since the channels could be connected to two entirely different networks). I will take a look at it.
The chip mcp2515 use this two buffers for one CAN bus line. And message can be in the first or second buffer, this decision is made by the chip.  In documentation: 
```Any message detected on the CAN bus is checked for errors and then matched against the user-defined filters to see if it should be moved into one of the two receive buffers.```

Because of it I think, that it is a technical detail

Andrew25455:
Patrick, thanks for this tests, i missed it.

> The first three all do basically the same. For the user, this is confusing, there should normally only be one way to do things, unless there's a reason. And the last one should probably be private anyway.

Last method (Read(Address address)) can not be private, because it is reading register, not messages from buffer. 
You are right, ReadAllBuffers method need to be private, because it is duplicate with ReadMessages.
About ReadRxBuffer and buffers we are discussing in the message above

Andrew25455:
Patrick, thanks for this test, i missed it.
I add ReceiveBuffer property to ReceivedCanMessage, because you are right it is important. Masks and filters can be added to this buffers and this property will help to separate messages in this situation. 
I also merged ReadAllBuffers and ReadMessages methods, for this realisation

Last method (Read(Address address)) can not be private, because it is reading register, not messages from buffer.

pgrawehr (approved):

pgrawehr:
@Andrew25455 Thanks, I think this is looking good now.

Andrew25455:
Patrick, Laurent, thank you, for your help, guys!  
 
 I need to do something with this PR?

pgrawehr:
@Andrew25455 no, not at the moment. I'll bring it up in tomorrow's team call, and if nobody has any reservations or wants to take another look, it will be merged.

Issue #1962, opened 2022-10-20
Provide a simple way to obtain the native GPIO driver used by the controller


raffaeler:
This is related to this issue: https://github.com/dotnet/iot/issues/1467

@joperezr [suggested to use reflection](https://github.com/dotnet/iot/issues/1467#issuecomment-791040524) to obtain the driver type name

I am proposing to add a new member that returns the driver name so that it is easier to make diagnostic.
Over the time we have seen many issues that could be solved simply by changing the driver. Adding a property with the driver `System.Type` is simple and could greatly help diagnostics.

I am wondering whether we could also expose the native driver name as well (`libgpiod` / `sysfs`).

There is an additional thing to consider. Certain peripherals can be created without specifying the controller (null as parameter). The controller is then created under the hood but it is not accessible from the code. In this case there would not be the possibility to obtain the driver name.

raffaeler:
We discussed this during the triage and agreed we have two distinct problems here.

1. Provide a new `GpioController` method returning a new object that contains **at least** the information the drivers in used for this instance. This proposal is in line with the initial proposal of this thread.
More specificatlly for 'drivers' I mean at least:
   * managed driver
   * board model
   * native OS driver

The alternative proposal I made during the triage was to provide the information above with a separate API so that we could know the result of the detection even before new-ing the GpioController. Anyway, getting the info from the `GpioController` is also needed because users can manually pick a different driver and we may want to know from the `GpioController` what is the driver in use. This alternate proposal is dropped.

2. @pgrawehr raised this second problem which is the current inability to customize the board detection logic and its extensibility.
Anyway, making changes to the current API would cause a huge breaking change to all the bindings. Most of the time the binding is called with a `null` in place of the `GpioController` which is then created internally (and thus, without the chance to customize the driver).

In order to solve this problem and avoid breaking changes, I proposed to adopt a solution similar to how the assembly loading hook works today in the runtime.
We could add a static delegate inside the GpioController class that, if defined from the user, would hook the driver creation process. This would permit to customize the driver without changing the bindings.

Also, the static delegate should received some information about the hardware the GpioController is intended for. In fact, it is possible to have **more than one** "boards" on the same physical machine. Different boards could require different drivers therefore the handler must know which hw is the driver intended for. Patrick proposed to pass the hw tree.

@joperezr asked to create a prototype in order to evaluate the possible implications for the aforementioned APIs.

@pgrawehr said he will work on prototyping the point 1. 

[...]

PR #1965, opened 2022-10-23
First draft for new ComponentInformation support

pgrawehr:
Concept for first part of #1962 

All important classes (`GpioDriver`, `GpiController`, `I2cDriver`, `SpiDriver` etc.) now contain a method `QueryComponentInformation` that return an instance of `ComponentInformation`. This class represents a tree of the active drivers and/or bindings. Further extension will allow to provide possible drivers as well.

[Further extension: It is intended that later `ComponentInformation` can be provided to a factory that constructs a device instance from such an information tree. One problem there is however, how one would know what is available. Providing a class that knows all possible drivers is quite ugly, and even instantiating a driver without initializing it (e.g. just to find out it's capabilities or _whether_ it would work) is currently not possible without relying on exceptions. Separating the constructor from the initialization could improve this. ]

## Example
This code:

```csharp
using Board b = Board.Create();
using ic2 = b.CreateOrGetI2cBus(1);
Console.WriteLine("Hardware detected: ");
Console.WriteLine(b.QueryComponentInformation(true));
```
Generates this output on a Raspberry Pi 4:
```text
Hardware detected: 
{Iot.Device.Board.RaspberryPiBoard} Raspberry Pi
 └{Iot.Device.Board.ManagedGpioController} Managed GPIO Controller
  └{System.Device.Gpio.Drivers.RaspberryPi3Driver} Generic Raspberry Pi Wrapper driver
   └{System.Device.Gpio.Drivers.RaspberryPi3LinuxDriver} Raspberry Pi 4 Model B Rev 1.2 linux driver with 28 pins and an interrupt driver
    └{System.Device.Gpio.Drivers.LibGpiodDriver} Gpio Driver
 └{Iot.Device.Board.I2cBusManager} I2C Bus Manager, Bus number 1

```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1965)


raffaeler:
With reagards to the output, it would be certainly useful to have both the tree (as an object graph), but also a string to uniquely identify the 'Component'.
In a first scenario I may disergard entirely which board I am using, but I do need `libgpiod` under the hood.
In an opposite scenario, I really need the RPi4 and I have to make sure the correct board was selected.

Finally I would split the driver and the board information in two different levels of the tree, so that comparing the information would be easier.

pgrawehr:
@raffaeler Looks like we have a similar understanding. Look at the code, I think it exactly has what you're asking for. Each component is currently identified by it's type name.

src/System.Device.Gpio/System/Device/ComponentInformation.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+        }
+
+        public ComponentInformation(object instance, string name, string version)
+            : this(instance, name, version, ComponentState.Unknown)
+        {
+        }
+
+        public ComponentInformation(object instance, string name, string version, ComponentState componentState)
+        {
```
raffaeler, line 27: What about using `Version` type instead of a string? Does it fit with this scenario?
My point is that would be much easier to compare

src/System.Device.Gpio/System/Device/ComponentInformation.cs
```diff
@@ -0,0 +1,95 @@
...
+
+#pragma warning disable CS1591
+namespace System.Device
+{
+    public record ComponentInformation
+    {
+        private readonly List<ComponentInformation> _subComponents;
+
+        public ComponentInformation(object instance, string name)
+            : this(instance, name, string.Empty)
```
raffaeler, line 17: This is a record, why not using the primary ctor and optional parameters with a default value?

src/System.Device.Gpio/System/Device/ComponentInformation.cs
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+
+        public override string ToString()
+        {
+            StringBuilder b = new StringBuilder($"{{{Type}}} {Name}{Environment.NewLine}");
+            SubComponentToString(b, 1);
+
+            return b.ToString();
+        }
```
raffaeler, line 93: I would prefer the ToString being without prefixes and suffixes (crlf)

raffaeler (commented):
Sorry for not reading the code first, but I didn't have time.
Love the general structure. I added a couple of comments even if it is still very early for a deep review.
I am thinking whether or not it could be possible to go in much detail about the "capabilities" so that it is easier to take decisions based on that instead of a raw string with the board model.

src/System.Device.Gpio/System/Device/ComponentInformation.cs
pgrawehr: I considered this, unfortunately the `Version` struct does not support modern semantic Versioning with Version tags.

src/System.Device.Gpio/System/Device/ComponentInformation.cs
pgrawehr: I prefer explicit overloads, but that's just syntax and can be adjustedat the end.

src/System.Device.Gpio/System/Device/IQueryComponentInformation.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Interface to support querying component information from a class
+    /// </summary>
+    public interface IQueryComponentInformation
+    {
+        /// <summary>
+        /// Query information about a component and it's children.
+        /// </summary>
+        /// <param name="onlyActive">True to return only active components, false to also list possible alternatives or inactive drivers</param>
```
joperezr, line 18: not sure I love `onlyActive`, should it instead be something like `listAllComponents`?

pgrawehr: As discussed: It will be better to have a separate method for that.

src/System.Device.Gpio/System/Device/ComponentInformation.cs
```diff
@@ -0,0 +1,95 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+#pragma warning disable CS1591
```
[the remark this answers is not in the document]
pgrawehr: That's just "member not documented". I regularly disable this while designing an API.

src/System.Device.Gpio/System/Device/ComponentInformation.cs
```diff
@@ -0,0 +1,95 @@
...
+namespace System.Device
+{
+    public record ComponentInformation
```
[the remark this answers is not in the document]
pgrawehr: It should be a reference type, since it anyway contains references to other instances. The reason it is not readonly has to do with creating it in derived classes. I  need to look at that again.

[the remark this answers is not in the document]
pgrawehr: Done, the record is now read-only. Fortunately, records can still be easily "updated" by using a copy and the `with` keyword.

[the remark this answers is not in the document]
raffaeler: > records can still be easily "updated" by using a copy and the with keyword
Oh well, you get a copy of the object with one or more mutated fields. The original object did not change at all.

Ellerbach:
I like the idea. Now, in terms of implementation, that will require to add the IQueryComponentInformation and the implementation in all Boards and associated drivers, correct?

raffaeler:
> I like the idea. Now, in terms of implementation, that will require to add the IQueryComponentInformation and the implementation in all Boards and associated drivers, correct?

Given that the abstract driver class implements the interface, the answer is yes.
But there is more. If I understood correctly, in your upcoming FTH binding, you should also have the opportunity to provide the IQueryComponentInformation there.

pgrawehr:
> > I like the idea. Now, in terms of implementation, that will require to add the IQueryComponentInformation and the implementation in all Boards and associated drivers, correct?
> 
> Given that the abstract driver class implements the interface, the answer is yes. But there is more. If I understood correctly, in your upcoming FTH binding, you should also have the opportunity to provide the IQueryComponentInformation there.

I'm actually not sure yet we really need the interface itself (just yet). We will have the method in several abstract base classes, but since we don't need something like a "list of all objects which can be queried" the interface itself is probably not needed. But this won't change the overall concept.

raffaeler:
@pgrawehr Not sure to understand what exactly is your current design.
Could you please describe it? (maybe the issue is a better place)

pgrawehr:
I'm trying out some refactoring, to be able to check the design for the next stage. This will be separated from this PR.

pgrawehr:
So, I think the concept should hold now. This is now waiting for PR #2019 to avoid conflicts. 
Also, I haven't really done much testing yet.

raffaeler:
@pgrawehr I'll wait for some descriptive document to better undertand the code intent. Thank you

src/System.Device.Gpio/System/Device/ComponentInformation.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        private readonly List<ComponentInformation> _subComponents;
+
+        public ComponentInformation(object instance, string name)
+            : this(instance, name, ComponentState.Active)
+        {
+        }
+
+        public ComponentInformation(object instance, string name, ComponentState componentState)
```
krwq, line 24: can we skip ComponentState for this initial review?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -282,4 +282,12 @@ protected override void Dispose(bool disposing)
         _internalDriver = null!;
         base.Dispose(disposing);
     }
+
+    /// <inheritdoc />
+    public override ComponentInformation QueryComponentInformation()
+    {
+        var ret = new ComponentInformation(this, "Generic Raspberry Pi Wrapper driver", ComponentState.Active);
```
joperezr, line 289: Not sure if the `Wrapper driver` part of the description could be better.

src/System.Device.Gpio/System/Device/ComponentInformation.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            get;
+            init;
+        }
+
+        public IReadOnlyList<ComponentInformation> SubComponents => _subComponents.AsReadOnly();
+
+        public void AddSubComponent(ComponentInformation subComponent)
+        {
+            if (_subComponents == null)
```
pgrawehr, line 72: If the level of the subcomponents is greater than X, abort (circular dependencies). Or check in printing!

src/System.Device.Gpio/System/Device/ComponentInformation.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            // Don't add the same element twice
+            if (_subComponents.All(x => x.Instance != subComponent.Instance))
+            {
+                _subComponents.Add(subComponent);
+            }
+        }
+
+        private static string PrintComponentDescription(ComponentInformation component)
```
krwq, line 84: protect against recursion

raffaeler (commented):
Please, consider checking for circular references

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -74,6 +77,17 @@ public GpioController(PinNumberingScheme numberingScheme)
     /// </summary>
     public virtual int PinCount => _driver.PinCount;
 
+    /// <summary>
+    /// Returns the collection of open pins
+    /// </summary>
+    protected IEnumerable<GpioPin> OpenPins
```
joperezr, line 83: Can we make this internal instead? Or private protected? I don't want this API to be public or protected.

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
```diff
@@ -100,4 +100,10 @@ private static bool IsBeagleBoneKernel()
             return false;
         }
     }
+
+    /// <summary>
+    /// Query information about a component and it's children.
+    /// </summary>
+    /// <returns>A tree of <see cref="ComponentInformation"/> instances.</returns>
+    public abstract ComponentInformation QueryComponentInformation();
```
krwq, line 108: consider

```csharp
public virtual ComponentInformation QueryComponentInformation() => new ComponentInformation(this, this.GetType().FullName);
```

krwq: even directly on the interface

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -767,6 +772,36 @@ protected override void Dispose(bool disposing)
...
+
+        ComponentInformation ci = new ComponentInformation(this, sb.ToString(), ComponentState.Active);
+
+        if (_interruptDriver != null)
+        {
+            ci.AddSubComponent(_interruptDriver.QueryComponentInformation());
+        }
+
+        return ci;
+    }
```
krwq, line 803: Possibly
`string ToString() { return $"{{ {string.Join(Subcomponents.Select(...))} }}"; }`

`string ToString() { return $"{{ {_interruptDriver}, ... }}"; }`

src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs
pgrawehr: I removed the interface as we tried to keep the public API as small as possible.

tools/ArduinoCsCompiler/tests/GarbageCollectorTests.cs
```diff
@@ -22,6 +22,7 @@ public GarbageCollectorTests(FirmataTestFixture fixture)
             : base(fixture)
         {
             Compiler.ClearAllData(true, false);
+            CompilerSettings.CreateKernelForFlashing = false;
```
krwq, line 25: seems like unrelated change

tools/ArduinoCsCompiler/tests/FirmataIlExecutorTests.cs
```diff
@@ -315,7 +315,7 @@ public void LargeStructTest(string methodName, Int32 argument1, Int32 argument2,
         [InlineData(nameof(TestMethods.CastClassTest), 0, 0, 1)]
         public void CastTest(string methodName, Int32 argument1, Int32 argument2, Int32 expected)
         {
-            LoadCodeMethod(methodName, argument1, argument2, expected, new CompilerSettings() { CreateKernelForFlashing = true, UseFlashForKernel = true });
+            LoadCodeMethod(methodName, argument1, argument2, expected, new CompilerSettings() { CreateKernelForFlashing = false, UseFlashForKernel = false });
```
krwq, line 318: are those firmata changes related to this PR?

src/System.Device.Gpio/System/Device/I2c/I2cDevice.cs
```diff
@@ -1,13 +1,21 @@
...
 /// <summary>
 /// The communications channel to a device on an I2C bus.
 /// </summary>
 public abstract partial class I2cDevice : IDisposable
 {
+    /// <summary>
+    /// This event is raised when <see cref="QueryComponentInformation"/> of the base class is called. This is can be used for bindings
+    /// to intercept that call on their own I2C device, so they can add additional information
+    /// </summary>
+    public event Action<ComponentInformation>? OnQueryComponentInformation;
```
krwq, line 17: This should be just callback rather than event especially given 1:1 relationship with device. Why not `Func<ComponentInformation>` instead?

krwq: ok my question in the end got answered by the code below. First comment is still valid though

krwq: btw I like this approach much more than inheritance initially proposed but I'll suggest to do only printing in the initial PR and in the following one we can focus on the approach

src/devices/Pca9685/Pca9685PwmChannel.cs
```diff
@@ -64,5 +65,11 @@ protected override void Dispose(bool disposing)
             _parent?.SetChannelAsDestroyed(_channel);
             _parent = null!;
         }
+
+        /// <inheritdoc />
+        public override ComponentInformation QueryComponentInformation()
+        {
+            return new ComponentInformation(this, "Pca9685 PWM Channel");
```
krwq, line 72: I'll suggest to move this to base class and just use type name given spaces do not add virtually any new information

src/devices/SoftPwm/SoftwarePwmChannel.cs
```diff
@@ -176,5 +177,14 @@ protected override void Dispose(bool disposing)
 
             base.Dispose(disposing);
         }
+
+        /// <inheritdoc />
+        public override ComponentInformation QueryComponentInformation()
+        {
+            var self = new ComponentInformation(this, "Software Pwm driver");
+            self.AddSubComponent(_controller.QueryComponentInformation());
+            self.Properties["Pin"] = _pin.ToString(CultureInfo.InvariantCulture);
```
krwq, line 186: I suggest we should change pin to always produce invariant culture information on regular ToString. I don't know how other programmers feel about translating diagnostic messages but I'm personally always annoyed by wacky translation done in my native tongue. It might make sense for other things but for diagnostics it just doesn't feel right

tools/ArduinoCsCompiler/tests/FirmataIlExecutorTests.cs
pgrawehr: Yes, kind of. The changes added a few extra types to the code set required for some tests, exceeding the allocated flash space. This change reduces the memory usage of the tests so they pass again.

tools/ArduinoCsCompiler/tests/GarbageCollectorTests.cs
pgrawehr: See below.

src/System.Device.Gpio/System/Device/I2c/I2cDevice.cs
pgrawehr: So you mean to remove this for now?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
pgrawehr: Neither am I...

src/System.Device.Gpio/System/Device/ComponentInformation.cs
pgrawehr: done

src/System.Device.Gpio/System/Device/ComponentInformation.cs
```diff
@@ -0,0 +1,146 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
krwq, line 1: [Triage] Consider using dynamic (expando object)

pgrawehr:
I had a look at `ExpandoObject` and `DynamicObject` but neither of them are really helpful here. They allow a lot of dynamic type construction, but have several drawbacks, too:
- For netstandard, a reference to "Microsoft.CSharp" is necessary, which doesn't seem very lightweight
- `QueryComponentInformation` would need to return `dynamic`, which is an untyped thing, leaving the caller confused about what he can do with the object. We don't want runtime exceptions trying to get debug information.
- One can define new properties on `ExpandoObject`, but not methods. It doesn't seem possible to override ToString() as well, which we need. `ExpandoObject` is sealed, so deriving from it is also not possible. `DynamicObject` allows deriving, but still requires users to use dynamic all over the place. 
- In the next release, we want to be able to construct a driver tree from an user-defined `ComponentInformation` tree. We can't expect people to construct correct instances of ExpandoObject. This would also be a problem for us, as eventually all bindings would create ComponentInformation instances, and we want them to be (at least somewhat) comparable and consistent.

raffaeler:
> * For netstandard, a reference to "Microsoft.CSharp" is necessary, which doesn't seem very lightweight

This means just .NET Framework and Mono. I won't be too worried for those.

> * `QueryComponentInformation` would need to return `dynamic`, which is an untyped thing, leaving the caller confused about what he can do with the object. We don't want runtime exceptions trying to get debug information.

That's by design, if we want to get rid of a type.

> * One can define new properties on `ExpandoObject`, but not methods. It doesn't seem possible to override ToString() as well, which we need. `ExpandoObject` is sealed, so deriving from it is also not possible. `DynamicObject` allows deriving, but still requires users to use dynamic all over the place.

You are right, but this can be resolved with some extension methods.

> * In the next release, we want to be able to construct a driver tree from an user-defined `ComponentInformation` tree. We can't expect people to construct correct instances of ExpandoObject. This would also be a problem for us, as eventually all bindings would create ComponentInformation instances, and we want them to be (at least somewhat) comparable and consistent.

Let's discuss this. I understand @krwq being worried to introduce a new type which is basically everywhere. 
The alternative solution we are discussing here is to minimize the impact by using dynamic + ExpandoObject and then a class defining the extension methods needed to make the dev experience better.
I am not completely against `ComponentInformation`, but the evaluation should be a fair comparison between the two solutions.

raffaeler:
BTW, even if I prefer the extension methods approach, this code using Expand includes the support to methods:
```
dynamic GetData()
{
    dynamic d = new ExpandoObject();
    d.Name = "abc";
    d.Count = 10;
    d.ToString = (Func<string>)(() =>
    {
        return $"Name:{d.Name} Count:{d.Count}";
    });
    return d;
}

dynamic x = GetData();
Console.WriteLine(x.ToString());
// output: 
// Name:abc Count:10
```

src/System.Device.Gpio/System/Device/ComponentInformation.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace System.Device
+{
+    /// <summary>
+    /// A class to provide informational data about system components and active drivers and bindings.
+    /// An instance of this class is typically obtained by calling QueryComponentInformation on any supported object.
+    /// The structure represents a tree of connected devices, e.g. a controller and its associated drivers.
```
krwq, line 17: [Triage] Add note in all new APIs in System.Device.Gpio that API is experimental or for debugging purposes and that may change in the future

joperezr:
@krwq looks like @pgrawehr has addressed your comments and this is ready to go. Can you take one final look and merge this one?

raffaeler (approved):

src/devices/Ft232H/Ft232HI2cBus.cs
```diff
@@ -108,5 +111,17 @@ internal void Write(int deviceAddress, ReadOnlySpan<byte> buffer)
 
             DeviceInformation.I2cStop();
         }
+
+        public override ComponentInformation QueryComponentInformation()
+        {
+            var self = new ComponentInformation(this, "Ftx232HI2c I2C Bus driver");
+            self.Properties["BusNo"] = "0";
+            foreach (var device in _usedAddresses)
```
Ellerbach, line 119: Can you add all the properties of the `DeviceInformation `?

src/devices/Ft232H/Ft232HDevice.cs
```diff
@@ -149,5 +150,11 @@ public Ft232HDevice(FtDevice ftDevice)
             : base(ftDevice.Flags, ftDevice.Type, ftDevice.Id, ftDevice.LocId, ftDevice.SerialNumber, ftDevice.Description)
         {
         }
+
+        /// <inheritdoc />
+        public override ComponentInformation QueryComponentInformation()
+        {
+            return base.QueryComponentInformation() with { Description = "Ftx232 Device" };
```
Ellerbach, line 157: Can you add all the properties of the `DeviceInformation `?

src/devices/Ft4222/Ft4222I2cBus.cs
```diff
@@ -138,5 +142,19 @@ protected override void Dispose(bool disposing)
...
             _ftHandle = null!;
         }
+
+        public override ComponentInformation QueryComponentInformation()
+        {
+            var self = new ComponentInformation(this, "FT4222 I2C Bus driver");
+            self.Properties["BusNo"] = "0";
+            self.Properties["Description"] = DeviceInformation.Description;
+            self.Properties["SerialNumber"] = DeviceInformation.SerialNumber;
+            foreach (var device in _usedAddresses)
```
Ellerbach, line 152: Can you add all the properties of the `DeviceInformation `?

Ellerbach (approved):
Looks good, if you can add the rest of the DeviceInformation property on the FT devices, it would be perfect!

PR #1966, opened 2022-10-25
Regenerate device listing

krwq:
Quickly regenerate device listing so it's up to date.

I've also removed HT16k33 folder since it's now under Display

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1966)


tools/device-listing/Program.cs
```diff
@@ -52,6 +52,7 @@
     "power",
     "voc",
     "thermocouple",
+    "weight",
```
Ellerbach, line 55: while you are there, don't you want to make the little changes used here to move to a json file? :-) https://github.com/nanoframework/Samples/tree/main/device-listing

If not, no worry, that can be done later.

src/devices/Device-Index.md
```diff
@@ -32,6 +32,7 @@
 * [Charlieplex Segment binding](Charlieplex/README.md)
 * [Cpu Temperature](CpuTemperature/README.md)
 * [DC Motor Controller](DCMotor/README.md)
+* [DFRobot KIT0176: I2C 1kg Weight Sensor Kit - HX711 (Gravity: I2CWeight Sensor)](Hx711I2c/README.md)
```
Ellerbach, line 35: while regenerating this file, what about going to the title of the README and adjusting them to add more about the device names? Like this one: https://github.com/dotnet/iot/blob/main/src/devices/Bmxx80/README.md
Would be great to list of the devices.

src/devices/README.md
```diff
@@ -31,8 +31,9 @@ Our vision: the majority of .NET bindings are written completely in .NET languag
 * [ADXL345 - Accelerometer](Adxl345/README.md)
```
Ellerbach, line 31: maybe reorder the file as well by alphabetic order?

tools/device-listing/Program.cs
krwq: I'd prefer to do that later, unfortunately don't have much time this week as I'm preparing for .NET Conf and I'd like the new devices to be available on device listing since they're used in the presentation :-)

src/devices/Device-Index.md
krwq: I've created https://github.com/dotnet/iot/issues/1967 to track all comments you've put here but unfortunately I'd prefer this one was scoped to simple change. I was thinking about doing this before I opened PR but realized I'm a bit short on time so decided to just do simple change

Ellerbach (approved):
Looks good with the issue to track what needs to be improved

PR #1998, opened 2022-12-24
Adding AM2320 (migration from .NET nanoFramework)

Ellerbach:
This PR is about adding AM2320

- migration from .NET nanoFramework (replacing SpanByte by Span<byte), new by stackalloc)
- adjusting the sample

Origin: https://github.com/nanoframework/nanoFramework.IoT.Device/tree/develop/devices/Am2320

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1998)

### DIFF
--- a/src/devices/Am2320/Am2320.cs
+++ b/src/devices/Am2320/Am2320.cs
@@ -39,83 +39,99 @@ namespace Iot.Device.Am2320
         /// <summary>
         /// Gets the last read temperature.
         /// </summary>
-        /// <remarks>
-        /// If last read was not successful, it returns <code>default(Temperature).</code>
-        /// </remarks>
-        [Telemetry]
-        public Temperature Temperature
+        /// <param name="temperature">[Out] The current temperature on success.</param>
+        /// <returns>True on success, false if reading failed.</returns>
+        [Telemetry("Temperature")]
+        public bool TryReadTemperature(
+#if NET5_0_OR_GREATER
+        [NotNullWhen(true)]
+#endif
+                out Temperature temperature)
         {
-            get
+            temperature = default;
+            if (IsOutDated())
             {
-                if (IsOutDated())
-                {
-                    ReadData();
-                }
-
-                return IsLastReadSuccessful ? GetTemperature() : default(Temperature);
+                ReadData();
             }
+
+            if (IsLastReadSuccessful)
+            {
+                temperature = GetTemperature();
+                return true;
+            }
+
+            return false;
         }
 
         /// <summary>
         /// Gets the last read of relative humidity in percentage.
         /// </summary>
-        /// <remarks>
-        /// If last read was not successful, it returns <code>default(RelativeHumidity).</code>
-        /// </remarks>
-        [Telemetry]
-        public RelativeHumidity Humidity
+        /// <param name="humidity">[Out] The current relative humidity on success.</param>
+        /// <returns>True on success, false if reading failed.</returns>
+        [Telemetry("Humidity")]
+        public bool TryReadHumidity(
+#if NET5_0_OR_GREATER
+            [NotNullWhen(true)]
+#endif
+                    out RelativeHumidity humidity)
         {
-            get
+            humidity = default;
+            if (IsOutDated())
             {
-                if (IsOutDated())
-                {
-                    ReadData();
-                }
-
-                return IsLastReadSuccessful ? GetHumidity() : default(RelativeHumidity);
+                ReadData();
             }
+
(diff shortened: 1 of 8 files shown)


src/devices/Am2320/Am2320.cs
```diff
@@ -0,0 +1,214 @@
...
+        public bool IsLastReadSuccessful { get; internal set; }
+
+        /// <summary>
+        /// Gets the last read temperature.
+        /// </summary>
+        /// <remarks>
+        /// If last read was not successful, it returns <code>default(Temperature).</code>
+        /// </remarks>
+        [Telemetry]
+        public Temperature Temperature
```
pgrawehr, line 46: Preferably change these properties to TryGetters. The operation is potentially expensive, has a side effect and may fail. We've had a bunch of problems with this `if (IsLastReadSuccessful())` approach for the DHT sensor family, which is conceptually very similar.

src/devices/Am2320/Am2320.cs
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        /// <summary>
+        /// Gets the last read of relative humidity in percentage.
+        /// </summary>
+        /// <remarks>
+        /// If last read was not successful, it returns <code>default(RelativeHumidity).</code>
+        /// </remarks>
+        [Telemetry]
+        public RelativeHumidity Humidity
```
pgrawehr, line 66: And here

src/devices/Am2320/Am2320.cs
```diff
@@ -0,0 +1,214 @@
...
+
+                return IsLastReadSuccessful ? GetHumidity() : default(RelativeHumidity);
+            }
+        }
+
+        /// <summary>
+        /// Gets the device information.
+        /// </summary>
+        [Property]
+        public DeviceInformation DeviceInformation
```
pgrawehr, line 83: and here

src/devices/Am2320/Am2320.cs
```diff
@@ -0,0 +1,214 @@
...
+            get
+            {
+                Span<byte> buff = stackalloc byte[11];
+
+                // 32 bit device ID
+                // Forces the sensor to wake up and wait 10 ms according to documentation
+                _i2c.WriteByte(0x00);
+                Thread.Sleep(10);
+
+                // Sending functin code 0x03, start regster 0x00 and 4 registers
```
pgrawehr, line 94: Comment apears to be wrong (copy-pasted from `ReadData()` method below?)

src/devices/Am2320/Am2320.cs
Ellerbach: That code was done before we adjusted all those. So adjusting for the Try pattern. I'll adjust that back to nanoFramework and will use the Obsolete as well on nanoFramework.

src/devices/Am2320/Am2320.cs
```diff
@@ -39,83 +39,99 @@ public class Am2320 : IDisposable
...
-        /// <remarks>
-        /// If last read was not successful, it returns <code>default(Temperature).</code>
-        /// </remarks>
-        [Telemetry]
-        public Temperature Temperature
+        /// <param name="temperature">[Out] The current temperature on success.</param>
+        /// <returns>True on success, false if reading failed.</returns>
+        [Telemetry("Temperature")]
+        public bool TryReadTemperature(
+#if NET5_0_OR_GREATER
```
pgrawehr, line 46: Must be `#if !NET5_0_OR_GREATER` ...

pgrawehr, line 46: This is not really necessary here, since `Temperature` is a value type.

src/devices/Am2320/samples/Program.cs
```diff
@@ -22,8 +23,8 @@
 
 while (true)
 {
-    var temp = am2330.Temperature;
-    var hum = am2330.Humidity;
+    am2330.TryReadTemperature(out Temperature temp);
+    am2330.TryReadHumidity(out RelativeHumidity hum);
     if (am2330.IsLastReadSuccessful)
```
pgrawehr, line 28: I suggest to show how to do it right: `if (am2330.TryReadTemperature(out Temperature temp) && ...`

pgrawehr (approved):

src/devices/Am2320/samples/Program.cs
Ellerbach: Great catch! You're fully right!

PR #2000, opened 2022-12-26
Adjust for ConfigurableFirmata V3

pgrawehr:
Removed reference to a deleted file
Removed outdated sample file (use the one in the Firmata repo instead, should be officially published shortly)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2000)


joperezr (approved):
These are the types of breaks that I was afraid of and that we should prevent by not having this leg run on CI. We have also been pinged by the infrastructure crew on an effort to try to reduce build times when possible, so I think the best course of action is to disable this leg from running as part of CI. We can keep the script around and running locally, or we can also have a separate pipeline for running it, and that way we can be able to run it on demand on a PR when making changes around the area.

joperezr:
Merging to unblock other PRs

PR #2005, opened 2023-01-08
Reverse byte order of ATQA calculated by ListenToIsoCard14443TypeA

### LINKED ISSUE #2004: Mfrc522.ListenToCardIso14443TypeA returns incorrect ATQA
```Iot.Device.Mfrc522.ListenToCardIso14443TypeA``` invokes ```PiccRequestA``` to perform a REQA, which returns the answer (ATQA) as a 2-byte array. ```ListenToCard14443TypeA``` converts this into a ushort. However, it treats the two bytes as a big-endian integer, whereas ISO 14443-3 specifies that the bits are sent from LSB to MSB. The MFRC522 reverses the bits within each byte but doesn't reverse the byte order. Therefore, the conversion should treat these two bytes as a little-endian integer.

Repro: listen for a Mifare Classic 1K card on an MFRC522. The expected ATQA is 0x0004 but the value 0x0400 is reported. As an example, ```Iot.Device.Card.Mifare.MifareCard.SetCapacity``` is unable to compute the card capacity when using an MFRC522, because the expected and reported ATQA values do not match.

Suggested fix: compute the ushort using

```csharp
            card.Atqa = BinaryPrimitives.ReadUInt16LittleEndian(atqa);
```

jdbruner:
Fixes #2004 

The two bytes of the ATQA need to be treated as a little-endian integer.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2005)

### DIFF
--- a/src/devices/Mfrc522/Mfrc522.cs
+++ b/src/devices/Mfrc522/Mfrc522.cs
@@ -264,7 +264,7 @@ namespace Iot.Device.Mfrc522
             }
             while (true);
 
-            card.Atqa = BinaryPrimitives.ReadUInt16BigEndian(atqa);
+            card.Atqa = BinaryPrimitives.ReadUInt16LittleEndian(atqa);
             var status = Select(out byte[]? nfcId, out byte sak);
             if (status != Status.Ok)
             {


Ellerbach (approved):
Looks good to me. With which card have you tested it?

jdbruner:
I have a couple of different Mifare Classic 1K cards (alas, they have no printed labels and I don't recall the brand). According to the NXP datasheet (MF1S50YYX_V1), depending upon the specific part, the expected ATQA values are one of 0x0044, 0x0004, 0x0042, and 0x0002. When I read these cards with an NFC app on my Android phone it tells me their ATQA is 0x0004.

Without this change the ATQA as a ushort was 0x0400. After applying the change, the reported ushort is 0x0004.

--John
________________________________

PR #2010, opened 2023-01-14
Update UnitsNet to V5.0

pgrawehr:
Updates UnitsNet to V5.0.0-rc010, fixing all the compile breakages due to breaking changes. 

@joperezr Can we upload this package to the internal server? It seems the build is failing with an unpatched nuget.config.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2010)

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <clear />
     <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
     <!--  Begin: Package sources from dotnet-runtime -->
     <!--  End: Package sources from dotnet-runtime -->
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
+    <clear />
     <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
     <!--  Begin: Package sources from dotnet-runtime -->
     <!--  End: Package sources from dotnet-runtime -->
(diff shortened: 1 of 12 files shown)


NuGet.config
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <clear />
```
joperezr, line 4: We can't make this change, it's a security policy in all of our repos so making it will break the build.

joperezr:
> Can we upload this package to the internal server?

We can, but should we instead wait for it to GA instead of uploading an RC?

NuGet.config
pgrawehr: I know, but like this at least the CI runs correctly.

joperezr: Right, CI will pass, but official builds will break. This is allowing sources like nuget.org to be used, which our official build infra checks disallow

pgrawehr:
@joperezr V5.0.0 final is out and now included in the PR. Can you please upload it to the internal feed so that the build passes? Thank you.

joperezr:
You should now be able to consume 5.1.0 here

eng/Versions.external.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <!-- These references to third-party libraries are included in all projects except System.Device.Gpio and the build wrapper project -->
   <ItemGroup Condition="'$(MSBuildProjectName)' != 'System.Device.Gpio' And '$(MSBuildProjectName)' != 'build'">
-    <PackageReference Include="UnitsNet" Version="4.77.0" />
+    <PackageReference Include="UnitsNet" Version="5.0.0" />
```
joperezr, line 4:
```suggestion
    <PackageReference Include="UnitsNet" Version="5.1.0" />
```

Issue #2012, opened 2023-01-15
Button transition is not detected correctly when an external resistor is used


raffaeler:
If an external pull up resistor is wired to the pin button and the user creates the button with `PinMode.Input` (without internal resistors), the transition in `PinStateChanged` is not detected correctly.
See: https://github.com/dotnet/iot/blob/main/src/devices/Button/GpioButton.cs#L78

**Steps to reproduce**
1. Use an external pull-up resistor
2. Configure the button with `PinMode.Input`

The gpio button always need either a pull-up or pull-down. It then can be internal (provided by the board) and configured by `PinMode` or external.

In order to fix this problem, we may use **one of** the following strategies:
1. add a boolean `IsExternalResistor` (defaulting to `false`) parameter specifying whether the the `PinMode` is either `InputPullUp` or `InputPullDown`. This is a binary breaking change, but does not break the compilation.
2. completely remove the the `PinMode` parameter and add two boolean values:
  - `IsPullUp` (`false` means pull-down)
  - `IsExternalResistor` (`false` means configured through `PinMode`)
  This requires the user to make changes in her code.
3. Add a new constructor overload (eventuallty deprecating the previous ones). This does not break any code (not even binary) but still exhibit the bug when using the old constructors.

With regards to option 2, instead of the two booleans we may introduce an `enum` with the following fields:
- PullUpInternal
- PullUpExternal
- PullDownInternal
- PullDownExternal

raffaeler:
Fixed with #2013

PR #2013, opened 2023-01-15
Button with external resistor bug as for #2012

raffaeler:
This PR implements solution 1 and 3 proposed in https://github.com/dotnet/iot/issues/2012

I ask the reviewers whether they prefer:
- option 1 (adding a single parameter to the existing constructor)
- option 3 (adding two ctor overloads)

In case of option 3, please tell me whether you prefer to:
- mark the old ctors as obsolete
- replace the two boolean values with an enum

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2013)


raffaeler:
After a rebase, this PR got all the doc changes that has already been merged in main. Don't know why.
The only changes of this PR are in the folder `src/devices/Button` (including its `README.md`)

src/devices/Button/GpioButton.cs
```diff
@@ -13,50 +13,63 @@ namespace Iot.Device.Button
...
-            if (_pinMode == PinMode.Input | _pinMode == PinMode.InputPullDown | _pinMode == PinMode.InputPullUp)
+            _eventPinMode = pinMode;
+            _gpioPinMode = isExternalResistor && (pinMode == PinMode.InputPullUp || pinMode == PinMode.InputPullDown)
+                ? _gpioPinMode = PinMode.Input
+                : _gpioPinMode = pinMode;
+
+            if (_gpioPinMode == PinMode.Input | _gpioPinMode == PinMode.InputPullDown | _gpioPinMode == PinMode.InputPullUp)
             {
-                _gpioController.OpenPin(_buttonPin, _pinMode);
+                _gpioController.OpenPin(_buttonPin, _gpioPinMode);
```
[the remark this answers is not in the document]
raffaeler: Correct. Still not sure whether keep or drop the old two ctors. We may discuss this, I don't have a strong opinion about the various options.

raffaeler:
Since this PR introduce a source and binary breaking change (the old two ctors were replaced with different parameters), I tried to run the command `build -pack /p:GenerateCompatibilitySuppressionFile=true` as discussed in the Triage.

Anyway the command failed with the following output:
```
H:\Samples\_forks\iot\src\devices\Arduino\samples\Monitor\Arduino.Monitor.cs(90,59): error CS1503: Argument 2: cannot convert from 'System.Device.Gpio.GpioController' to 'bool' [H:\Samples\_forks\iot
\src\devices\Arduino\samples\Monitor\Arduino.Monitor.csproj::TargetFramework=net6.0]
H:\Samples\_forks\iot\src\devices\Arduino\samples\Monitor\Arduino.Monitor.cs(90,82): error CS1503: Argument 4: cannot convert from 'System.Device.Gpio.PinMode' to 'System.Device.Gpio.GpioController?'
 [H:\Samples\_forks\iot\src\devices\Arduino\samples\Monitor\Arduino.Monitor.csproj::TargetFramework=net6.0]
H:\Samples\_forks\iot\src\devices\Arduino\samples\Monitor\Arduino.Monitor.cs(90,97): error CS1503: Argument 5: cannot convert from 'System.TimeSpan' to 'bool' [H:\Samples\_forks\iot\src\devices\Ardui
no\samples\Monitor\Arduino.Monitor.csproj::TargetFramework=net6.0]
    0 Warning(s)
    3 Error(s)

Time Elapsed 00:03:42.99
Build failed with exit code 1. Check errors above.
```

Apart the compatibility suppression xml file not being generated, the rest of the PR is good and ready for review/approval.

raffaeler:
@pgrawehr I had to change the ctor call in [Arduino.Monitor.cs](https://github.com/dotnet/iot/pull/2013/files#diff-381bf9b58ab356ba3acf4c946343245237d0d85de7409545b305b88ee0b637d7).
Since previously the PinMode was `Input`, I don't know if it was pull-up or pull-down.
Please take a look at the current values and tell me whether I should change them (I don't know how the Arduino board is configured).

raffaeler:
@joperezr I see helix failing with this error. Is there anything I can do?

```
    System.Device.Gpio.Tests.SysFsDriverTests.AddCallbackFallingEdgeNotDetectedTest [FAIL]
      Assert.False() Failure
      Expected: False
      Actual:   True
      Stack Trace:
        /_/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs(246,0): at System.Device.Gpio.Tests.GpioControllerTestBase.AddCallbackFallingEdgeNotDetectedTest()
```

joperezr:
> I see helix failing with this error. Is there anything I can do?

Those are the hardware tests. Given their nature and what they test, they tend to be flaky sometimes. Usually we just re-run to make sure it was just a fluke. In this case we can be pretty sure about it since your PR isn't really touching S.D.G and the other Linux Debug build (which also runs the tests) passed.

raffaeler:
> Those are the hardware tests. Given their nature and what they test, they tend to be flaky sometimes. Usually we just re-run to make sure it was just a fluke. In this case we can be pretty sure about it since your PR isn't really touching S.D.G and the other Linux Debug build (which also runs the tests) passed.

Thanks, but recently I got merge permissions, but not the azp ones :-)

pgrawehr:
> @pgrawehr I had to change the ctor call in [Arduino.Monitor.cs](https://github.com/dotnet/iot/pull/2013/files#diff-381bf9b58ab356ba3acf4c946343245237d0d85de7409545b305b88ee0b637d7). Since previously the PinMode was `Input`, I don't know if it was pull-up or pull-down. Please take a look at the current values and tell me whether I should change them (I don't know how the Arduino board is configured).

The button is pulled down with an external resistor in the example. See src\devices\Arduino\ArduinoSample_Monitor.png for the schematic of the example.

raffaeler:
> The button is pulled down with an external resistor in the example. See src\devices\Arduino\ArduinoSample_Monitor.png for the schematic of the example.

@pgrawehr 
I just commited and pushed the change with `isPullUp=false` and `HasExternalResistor=true` on the Arduino.Monitor.cs.

raffaeler:
@pgrawehr do you have any idea why the `Linux Build Build_Release` failed twice in a row?
If you think that this is a random failure, please azp this PR again.
Thank you

pgrawehr:
> @pgrawehr do you have any idea why the `Linux Build Build_Release` failed twice in a row?
> If you think that this is a random failure, please azp this PR again.
> Thank you

Yea, that particular test is flaky. I'll check a bunch of the latest builds and disable it if it's really always the same.

src/devices/Button/GpioButton.cs
```diff
@@ -13,56 +13,61 @@ namespace Iot.Device.Button
...
         /// <param name="gpio">Gpio Controller.</param>
         /// <param name="shouldDispose">True to dispose the GpioController.</param>
         /// <param name="debounceTime">The amount of time during which the transitions are ignored, or zero</param>
-        public GpioButton(int buttonPin, TimeSpan doublePress, TimeSpan holding, GpioController? gpio = null, bool shouldDispose = true, PinMode pinMode = PinMode.InputPullUp, TimeSpan debounceTime = default(TimeSpan))
+        public GpioButton(int buttonPin, TimeSpan doublePress, TimeSpan holding, bool isPullUp = true, bool hasExternalResistor = false,
+            GpioController? gpio = null, bool shouldDispose = true, TimeSpan debounceTime = default)
             : base(doublePress, holding, debounceTime)
         {
             _gpioController = gpio ?? new GpioController();
             _shouldDispose = shouldDispose;
```
pgrawehr, line 60: Make sure _shouldDispose is true when gpio was null (our own GpioController was created)

src/devices/Button/GpioButton.cs
```diff
@@ -13,56 +13,61 @@ namespace Iot.Device.Button
...
-            else
-            {
-                throw new ArgumentException("GPIO pin can only be set to input, not to output.");
-            }
+            _eventPinMode = isPullUp ? PinMode.InputPullUp : PinMode.InputPullDown;
+            _gpioPinMode = hasExternalResistor
+                ? _gpioPinMode = PinMode.Input
+                : _gpioPinMode = _eventPinMode;
+
+            _gpioController.OpenPin(_buttonPin, _gpioPinMode);
```
pgrawehr, line 69: This may need some error handling here, since OpenPin can fail (invalid pin number, unsupported pin mode, etc.). In this case, the gpioController needs to be disposed immediately (if _shouldDispose is true), since we're in a constructor. The caller has no way of using a finaly arount the exception.

src/devices/Button/GpioButton.cs
```diff
@@ -66,8 +72,19 @@ public class GpioButton : ButtonBase
...
+            {
+                _gpioController.OpenPin(_buttonPin, _gpioPinMode);
+                _gpioController.RegisterCallbackForPinValueChangedEvent(
+                    _buttonPin,
+                    PinEventTypes.Falling | PinEventTypes.Rising,
+                    PinStateChanged);
+            }
+            catch (Exception)
+            {
+                _gpioController.Dispose();
```
pgrawehr, line 85: Almost perfect. Add `if (_shouldDispose)` here. (For some reason, I cannot post suggestions.)

pgrawehr (approved):

src/devices/Button/GpioButton.cs
```diff
@@ -13,55 +13,81 @@ namespace Iot.Device.Button
...
         {
             _gpioController = gpio ?? new GpioController();
-            _shouldDispose = shouldDispose;
+            _shouldDispose = gpio == null ? true : shouldDispose;
             _buttonPin = buttonPin;
-            _pinMode = pinMode;
-            _debounceTime = debounceTime;
+            HasExternalResistor = hasExternalResistor;
+
+            _eventPinMode = isPullUp ? PinMode.InputPullUp : PinMode.InputPullDown;
```
Ellerbach, line 70: This is dangerous! Because some board do not support PullUp/PullDown. It's the reason why originally you have to pass the pull type.
If you want to do something like this, you need first to check if the pin mode is supported, if not, then choose one that is a default one.
See here for the good way to do it: https://github.com/dotnet/iot/blob/main/src/devices/Dhtxx/DhtBase.cs#L192

src/devices/Button/GpioButton.cs
```diff
@@ -75,7 +101,7 @@ private void PinStateChanged(object sender, PinValueChangedEventArgs pinValueCha
             switch (pinValueChangedEventArgs.ChangeType)
             {
                 case PinEventTypes.Falling:
-                    if (_pinMode == PinMode.InputPullUp)
+                    if (_eventPinMode == PinMode.InputPullUp)
```
Ellerbach, line 104: same here, you can't really do it like this is pull up is not supported

Ellerbach (commented):
PullUp/PullDown are not supported everywhere. So you can't create them just like this. You have to test and then keep the chosen mode to compare later on.

src/devices/Button/GpioButton.cs
raffaeler: @Ellerbach I am aware that not every pin can be internally pulled up or down.
I didn't change the logic of the original author, I just added support for the external pullup/down resistor.

I was not aware of `IsPinModeSupported` in `GpioController`. Good to know, Now I can throw an exception if it's not.

src/devices/Button/GpioButton.cs
raffaeler: The `_eventPinMode` is always good because it is used to keep track if it is either pulled-up internally or externally.
The validation only goes when I open the pin. But if the resistor is external, the raising/falling logic is basically the same.

src/devices/Button/GpioButton.cs
raffaeler: I revised two cases:
- if the button can be configured as input (certain gpio can only be output)
- if the button can be configured with the configuration deriving from `isPullUp` and `HasExternalResistor`

src/devices/Button/GpioButton.cs
raffaeler: To make this more clear: `_eventPinMode` is **not** used in `OpenPin`. 
I just 'reused" the enumeration to keep track whether the button transition is high to low or vice-versa.

src/devices/Button/GpioButton.cs
pgrawehr: I think this is fine. If the mode is not supported, it will throw, which is expected, because it's an user error.

Ellerbach (approved):
Thanks a lot for the tweaks, looks good to me now.

raffaeler:
> Thanks a lot for the tweaks, looks good to me now.

Could you please you merge it?

PR #2019, opened 2023-01-20
Improve FT232H, add FT2232H and FT4232H

Ellerbach:
This PR improves the initial work on FT232H.
It adds as well the full support for FT2232H and FT4232H.
Improvement of the documentation.
Addition for new sample.

Both 3 devices has been tested successfully with I2C using a TSL2561 sensor. GPIO pins tested including for the specific Bit Band mode on FT4232H.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2019)

### DIFF
--- a/src/devices/Ft232H/Ft2232HDevice.cs
+++ b/src/devices/Ft232H/Ft2232HDevice.cs
@@ -17,9 +17,9 @@ namespace Iot.Device.Ft2232H
     public class Ft2232HDevice : Ftx232HDevice
     {
         /// <summary>
-        /// Gets all the FT232H connected
+        /// Gets all the FT2232H connected
         /// </summary>
-        /// <returns>A list of FT232H</returns>
+        /// <returns>A list of FT2232H</returns>
         public static List<Ft2232HDevice> GetFt2232H()
         {
             List<Ft2232HDevice> ft2232s = new List<Ft2232HDevice>();
--- a/src/devices/Ft232H/Ft2232HDevice.cs
+++ b/src/devices/Ft232H/Ft2232HDevice.cs
@@ -3,37 +3,33 @@
 
 using System;
 using System.Collections.Generic;
-using System.Device.Gpio;
-using System.Device.I2c;
-using System.Device.Spi;
-using System.IO;
-using System.Runtime.InteropServices;
-using System.Threading;
-using Iot.Device.Board;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
 using Iot.Device.Common;
 using Iot.Device.FtCommon;
 
-namespace Iot.Device.Ft232H
+namespace Iot.Device.Ft2232H
 {
     /// <summary>
     /// FT232H Device
     /// </summary>
-    public class Ft232HDevice : Ftx232HDevice, IDisposable
+    public class Ft2232HDevice : Ftx232HDevice
     {
         /// <summary>
         /// Gets all the FT232H connected
         /// </summary>
         /// <returns>A list of FT232H</returns>
-        public static List<Ft232HDevice> GetFt232H()
+        public static List<Ft2232HDevice> GetFt2232H()
         {
-            List<Ft232HDevice> ft232s = new List<Ft232HDevice>();
-            var devices = FtCommon.FtCommon.GetDevices(new FtDeviceType[] { FtDeviceType.Ft232H });
+            List<Ft2232HDevice> ft2232s = new List<Ft2232HDevice>();
+            var devices = FtCommon.FtCommon.GetDevices(new FtDeviceType[] { FtDeviceType.Ft2232H });
             foreach (var device in devices)
             {
-                ft232s.Add(new Ft232HDevice(device));
+                ft2232s.Add(new Ft2232HDevice(device));
             }
 
-            return ft232s;
+            return ft2232s;
         }
 
         /// <summary>
@@ -50,6 +46,7 @@ namespace Iot.Device.Ft232H
             switch (pin)
             {
                 case "ADBUS0":
+                case "BDBUS0":
                 case "D0":
                 case "TCK":
                 case "SK":
@@ -57,6 +54,7 @@ namespace Iot.Device.Ft232H
                 case "SDL":
                     return 0;
                 case "ADBUS1":
+                case "BDBUS1":
                 case "D1":
                 case "DO":
                 case "TDI":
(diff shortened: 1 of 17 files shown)


src/devices/Ft232H/Ftx232HDevice.cs
```diff
@@ -0,0 +1,1074 @@
...
+        internal bool[] PinOpen = new bool[PinCountConst];
+        internal PinMode[] GpioDirections = new PinMode[PinCountConst];
+        internal bool UsesMpseeForGpio = true;
+
+        internal List<SpiConnectionSettings> ConnectionSettings = new List<SpiConnectionSettings>();
+
+        /// <summary>
+        /// Gets all the FT232H connected
+        /// </summary>
+        /// <returns>A list of FT232H</returns>
```
pgrawehr, line 66: Comment is wrong (this apparently returns any type of FTxxxx device)

src/devices/Ft232H/Ftx232HDevice.cs
```diff
@@ -0,0 +1,1074 @@
...
+            List<Ftx232HDevice> ft232s = new List<Ftx232HDevice>();
+            var devices = FtCommon.GetDevices(new FtDeviceType[] { FtDeviceType.Ft232H, FtDeviceType.Ft2232H, FtDeviceType.Ft4232H, FtDeviceType.Ft2232 });
+            foreach (var device in devices)
+            {
+                var dev = device.Type switch
+                {
+                    FtDeviceType.Ft2232 or FtDeviceType.Ft2232H => new Ft2232HDevice(device),
+                    FtDeviceType.Ft232H => new Ft232HDevice(device),
+                    FtDeviceType.Ft4232H => new Ft4232HDevice(device),
+                    _ => new Ftx232HDevice(device),
```
pgrawehr, line 78: Is this fallback useful? Are you assuming that any unknown device will just work with the fallback?

src/devices/Ft232H/Ft2232HDevice.cs
```diff
@@ -0,0 +1,165 @@
...
+    public class Ft2232HDevice : Ftx232HDevice
+    {
+        /// <summary>
+        /// Gets all the FT232H connected
+        /// </summary>
+        /// <returns>A list of FT232H</returns>
+        public static List<Ft2232HDevice> GetFt2232H()
+        {
+            List<Ft2232HDevice> ft2232s = new List<Ft2232HDevice>();
+            var devices = FtCommon.FtCommon.GetDevices(new FtDeviceType[] { FtDeviceType.Ft2232H });
```
pgrawehr, line 26: The method you're calling here (`FtCommon.GetDevices(FtDeviceType[])`) is kind of weird, because you could  provide the same type multiple times. Maybe use a lambda instead of the list argument? Or check for duplicates?

pgrawehr (commented):
Looks good to me, some minor findings.

Maybe a namespace adjustment would be appropriate? It's hard to find the FT4232 in a folder called FT232 (there's also a FT4222...)

src/devices/Ft232H/Ftx232HDevice.cs
Ellerbach: yes, see the sample where I'm using the Ftx232HDevice rather than a named one.

Ellerbach: ho, got your question, you're right, it should not be needed but I leave it like this otherwise the switch complains that there is no default

src/devices/Ft232H/Ft2232HDevice.cs
Ellerbach: the GetDevices method actually returns only the types you are interested in. Idea like in your previous comment is being able to select just 1 type. Which I'm doing here. But in the previous one, you can select multiple types.
Those are common functions for FT4222 for example as well

Ellerbach:
> Maybe a namespace adjustment would be appropriate? It's hard to find the FT4232 in a folder called FT232 (there's also a FT4222...)

This is what I've tried to do:
Ftx232HDevice => FtCommon
Ft2232HDevice => Ft2232H
Ft232HDevice => Ft232H (same as before)
Ft4232HDevice => Ft4232H
Now, I've kept all the Ft232HGpio, Spi, I2c, I2cBus in the Ft232H namespace for compatibility (same as before)
All the above *HDevice uses those ones for the specific Gpio, Spi, I2c and I2cBus operations.

I know the names are not that easy to read without breaking what was present, so if you have a better suggestion, happy to apply it.

src/devices/Ft232H/Ftx232HDevice.cs
Ellerbach: And I checked all of them, I forgot to adjust quite some. Thanks!

src/devices/Ft232H/Ft2232HDevice.cs
pgrawehr: Sorry if I was unclear. I meant that if you call `FtCommon.FtCommon.GetDevices(new FtDeviceType[] { FtDeviceType.Ft2232H , FtDeviceType.Ft2232H});` you suddenly get two devices. It's silly to do that, but I think it should be prevented.

pgrawehr:
> > Maybe a namespace adjustment would be appropriate? It's hard to find the FT4232 in a folder called FT232 (there's also a FT4222...)
> 
> This is what I've tried to do: Ftx232HDevice => FtCommon Ft2232HDevice => Ft2232H Ft232HDevice => Ft232H (same as before) Ft4232HDevice => Ft4232H Now, I've kept all the Ft232HGpio, Spi, I2c, I2cBus in the Ft232H namespace for compatibility (same as before) All the above *HDevice uses those ones for the specific Gpio, Spi, I2c and I2cBus operations.
> 
> I know the names are not that easy to read without breaking what was present, so if you have a better suggestion, happy to apply it.

The names of the classes are fine. I was just thinking about changing the namespace. I am aware that this would be a breaking change. We've done it in other places as well, but I leave it to you to decide.

pgrawehr (approved):

src/devices/Ft232H/Ft2232HDevice.cs
Ellerbach: > Sorry if I was unclear. I meant that if you call `FtCommon.FtCommon.GetDevices(new FtDeviceType[] { FtDeviceType.Ft2232H , FtDeviceType.Ft2232H});` you suddenly get two devices. It's silly to do that, but I think it should be prevented.

OK, I see your point. But that's an internal function. Still I can adjust this to make sure the list is deduplicated: https://github.com/dotnet/iot/blob/a0db92f8c766771434095662ce2a0c0e47bec24c/src/devices/FtCommon/FtCommon.cs#L72

Ellerbach: So I made it unique. Thanks for point out

Ellerbach:
@krwq do you want to have a quick look at it or as Patrick's already reviewed it, I can go and merge it?

PR #2022, opened 2023-01-21
UltralightCard fixes - byte order of ATQA, number of blocks in EV1

jdbruner:
The ATQA returned by NFC anticollision is converted by a ushort by the Mfrc522, Pn532, and Pn5180 transceivers. In the case of the Mfrc522, the byte order was reversed - which was corrected by #2005. However, perhaps because the ATQA ushorts were inconsistently calculated by the different transceivers, it is also inconsistently used.  Iot.Device.Card.Mifare.SetCapacity expects the ATQA of a Mifare classic 1K card to be 0x0004 and the ATQA of a Mifare classic 4K card to be 0x0002 (which are the values defined by NXP for those parts). However, Iot.Device.Card.Ultralight.UltralightCard.IsUltralightCard expects the ATQA of an Ultralight card to be 0x4400, when it should be 0x0044 according to the NXP spec.

In addition, the property UltralightCard.NumberBlocks returns the number of blocks based upon the type of the card. The values returned for UltralightEV1 cards are incorrect. This causes the sample code to attempt to read past the end of the card prevents the card configuration from being decoded.

This change aligns the expected ATQA and the actual number of blocks with the spec.

Tested with an Mfrc522 and a Pn532 using Mifare Classic and Ultralight cards. The cards are now correctly identified.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2022)

### DIFF
--- a/src/devices/Card/Ultralight/UltralightCard.cs
+++ b/src/devices/Card/Ultralight/UltralightCard.cs
@@ -76,7 +76,7 @@ namespace Iot.Device.Card.Ultralight
         /// <param name="ATQA">The ATQA</param>
         /// <param name="SAK">The SAK</param>
         /// <returns>True if this is an Ultralight</returns>
-        public static bool IsUltralightCard(ushort ATQA, byte SAK) => (ATQA == 0x4400) && (SAK == 0);
+        public static bool IsUltralightCard(ushort ATQA, byte SAK) => (ATQA == 0x0044) && (SAK == 0);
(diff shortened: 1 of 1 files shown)


Ellerbach (approved):
Looks good to me. Thanks for fixing those nit in all the cards.

Issue #1581, opened 2021-07-01
OpenPinDefaultsModeToLastMode test gives different results on different hardware


pgrawehr:
The test `OpenPinDefaultsModeToLastMode` behaves differently on our build servers than when testing on local Raspberry Pi3 and 4. The changes in PR #1520 only changed the expectation on this test, which made it fail on the build servers but pass locally. 

The current theory is that a recent update of the sysfs library causes this difference (the servers have not been patched in quite a while). 

The test is currently disabled for SysFs.

krwq:
marking as bug since it's unclear if it's a product or test issue

Issue #629, opened 2019-07-25
AddCallbackRemoveAllCallbackTest is failing intermitently in test runs


joperezr:
**Ensure to re-enable or remove the test as part of this issue**

Because of the way that we currently have our SysFs Gpio Driver implemented, this test may run into a race condition intermittently and cause it to fail. Because of that, I'm going to disable it for now when executing SysFs driver tests or RaspberryPi driver tests as they use the same logic underneath. This issue is to track the work to make eventing logic work better for those drivers, and to enable the test once that is done.

krwq:
It seems to be flaky also on Windows so I'm disabling this comepletely

Ellerbach:
[Triage] This is still valid. SysFs driver is used only in few boards not supporting anything else. What should be added is a warning if libgpio is not installed maybe?

krwq:
[Triage] This seems to have been fixed at one point

PR #2023, opened 2023-01-22
Disable a flaky test for SysFs

pgrawehr:
[This report](https://dev.azure.com/dotnet/IoT/_test/analytics?definitionId=179&contextType=build) indicates that this particular test is responsible for most of the helix test failures.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2023)

### DIFF
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -230,9 +230,16 @@ public abstract class GpioControllerTestBase
     [Fact]
     public void AddCallbackFallingEdgeNotDetectedTest()
     {
+        var driver = GetTestDriver();
+        if (driver is SysFsDriver)
+        {
+            // This test is unreliable (flaky) with SysFs.
+            return;
+        }
+
         bool wasCalled = false;
         AutoResetEvent ev = new AutoResetEvent(false);
-        using (GpioController controller = new GpioController(GetTestNumberingScheme(), GetTestDriver()))
+        using (GpioController controller = new GpioController(GetTestNumberingScheme(), driver))
         {
             controller.OpenPin(InputPin, PinMode.Input);
             controller.OpenPin(OutputPin, PinMode.Output);
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -234,7 +234,7 @@ public abstract class GpioControllerTestBase
         if (driver is SysFsDriver)
         {
             // This test is unreliable (flaky) with SysFs.
-            // See #629 and possibly #1581
+            // See https://github.com/dotnet/iot/issues/629 and possibly https://github.com/dotnet/iot/issues/1581
             return;
         }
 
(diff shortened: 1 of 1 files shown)


src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -230,9 +230,16 @@ void Callback(object sender, PinValueChangedEventArgs pinValueChangedEventArgs)
     [Fact]
     public void AddCallbackFallingEdgeNotDetectedTest()
     {
+        var driver = GetTestDriver();
+        if (driver is SysFsDriver)
+        {
+            // This test is unreliable (flaky) with SysFs.
```
krwq, line 236:
```suggestion
            // [ActiveIssue("IssueLinkHere")] This test is unreliable (flaky) with SysFs.
```

krwq (approved):
Thanks!

joperezr:
Have we found evidence of runs where ONLY this test fails? What I've seen in the past, is that sometimes a board would end up in a bad state of pinmode on the input pin and it essentially fails around 10 or so tests. If that's the case, then I think that this PR isn't really getting us much value, and it is dropping coverage. My suggestion is wait for a PR to fail the tests and validate that this test is the only one failing, sound good?

pgrawehr:
> Have we found evidence of runs where ONLY this test fails? What I've seen in the past, is that sometimes a board would end up in a bad state of pinmode on the input pin and it essentially fails around 10 or so tests. If that's the case, then I think that this PR isn't really getting us much value, and it is dropping coverage. My suggestion is wait for a PR to fail the tests and validate that this test is the only one failing, sound good?

As I already have noted:[ This statistic](https://dev.azure.com/dotnet/IoT/_test/analytics?definitionId=179&contextType=build) shows that this is the only test that failed in all of the recent builds. I've manually checked 2 or 3 recent helix failures, and it was always this test (and only this test).

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
pgrawehr: Is this used elsewhere? I cannot find such an attribute.

krwq: we use it in dotnet/runtime I'd guess it works here as well (I'm guessing one of the thing arcade might have pulled in) - feel free to ignore this comment though, it just makes it easier to find pieces of code related to particular issue

joperezr:
I see, sorry @pgrawehr I missed that. One last ask, can we please log an issue to track the fix of the test, and add a comment where you are disabling it to point to that issue? Basically the same thing that @krwq suggested.

pgrawehr:
Ticket #629 exists already.

The issue could also be related to #1581

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -230,9 +230,17 @@ void Callback(object sender, PinValueChangedEventArgs pinValueChangedEventArgs)
...
+            // See #629 and possibly #1581
```
krwq, line 237:
```suggestion
            // See https://github.com/dotnet/iot/issues/629 and possibly https://github.com/dotnet/iot/issues/1581
```

PR #2027, opened 2023-02-04
MFRC522 NFC transceiver support for Write16Bytes

jdbruner:
Mifare's Write16Bytes command, which is used by Mifare Classic and Mifare Ultralight (where it is called "compatibility write") requires two writes to the card. The first one transfers the block number, and the second contains the data to be written. The Mfrc522 transceiver was not doing this, so writes to a Mifare Classic card always failed. Changed the Transceive function to treat Write16Bytes the same as counter increment, decrement, and restore (which also require two writes to the card).

In addition to renaming TwoStepsIncDecRestore to TwoStepsWrite16IncDecRestore, there were two issues in this method. First, the calculation of the CRC for the second transfer was incorrect. Second, it appears that the Mfrc522 actually returns a single byte response from the second transfer (which looks like the ACK/NAK), but dataFromCard.Length is 0, so SendAndReceiveData will return an error even though the write succeeded. Actual errors are already detected separately, so the simplest change is to pass NULL rather than dataFromCard. This makes dataFromCard unused, so it is removed from the method interface.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2027)

### DIFF
--- a/src/devices/Mfrc522/Mfrc522.cs
+++ b/src/devices/Mfrc522/Mfrc522.cs
@@ -867,6 +867,8 @@ namespace Iot.Device.Mfrc522
             // Use built in functions for authentication in case of classic Mifare cards
             if ((dataToSend[0] == (byte)MifareCardCommand.AuthenticationA) || (dataToSend[0] == (byte)MifareCardCommand.AuthenticationB))
             {
+                // UltralightCommand.GetVersion has the same command code as MifareCardCommand.AuthenticationA
+                // GetVersion returns data; AuthenticationA does not
                 if ((dataFromCard == null) || (dataFromCard.Length == 0))
                 {
                     status = SendAndReceiveData(MfrcCommand.MifareAuthenticate, dataToSend.ToArray(), null);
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=2027) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=2027) before we can accept your contribution.<br/><br/>:x: jdbruner [sign now](https://cla.dotnetfoundation.org/dotnet/iot?pullRequest=2027)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/iot?pullRequest=2027) it.</sub>

jdbruner:
I should also have noted that, in addition to renaming TwoStepsIncDecRestore to TwoStepsWrite16IncDecRestore, there were two issues in this method. First, the calculation of the CRC for the second transfer was incorrect. Second, it appears that the Mfrc522 actually returns a single byte response from the second transfer (which looks like the ACK/NAK), but dataFromCard.Length is 0, so SendAndReceiveData will return an error. Errors are already detected separately, so the simplest change is to pass NULL rather than dataFromCard. This makes dataFromCard unused, so it is removed from the method interface.

Ellerbach (approved):
Thanks for the fix, looks good to me.

PR #2028, opened 2023-02-04
multiple fixes for 2K and 4K cards

jdbruner:
MifareCard.cs notes that it only supports 1K cards. There is some code for 2K and 4K cards, but there are some shortcomings (particularly for 4K cards). This change introduces a number of fixes to resolve those shortcomings.

# Background
Mifare Classic comes in two sizes: 1K and 4K. However, when Mifare Plus cards are operated at security level 1, they look like Mifare Classic cards. Mifare Plus comes in 2K and 4K sizes.

Memory is organized into sectors, which comprise data blocks and a sector trailer block. All blocks are 16 bytes.
- 1K cards have 16 sectors with 4 blocks per sector
- 2K cards have 32 sectors with 4 blocks per sector
- 4K cards have 32 sectors with 4 blocks per sector and 8 more sectors with 16 blocks per sector

The sector trailer is the last block in the sector. It contains the two authentication keys (A and B), and access bytes. The access bytes encode eight different configurations for data blocks and the sector trailer. There are four sets of 3-bit access conditions. For sectors less than 32, each set of access conditions applies to one block within the sector. For sectors 32 and above, the first three sets of access conditions apply to 5-block groups (0-4, 5-9, 10-14), and the final set of access conditions applies to the sector trailer.

# Changes

The pre-existing code assumed that there were always 4 blocks per sector and the sector trailer was always the fourth one. To handle larger sectors, this change introduces the concept of a "block group" within a sector, which represents the blocks that are associated with the access conditions. There are four block groups per sector. For sectors less than 32 there is one block per group. For sectors above 32, there are five data blocks per group; the last group is the sector trailer.

Convenience methods to convert between block numbers and block groups were added, and they replace previously hardcoded calculations (such as ```blockNumber % 4```) that are only correct for the first 32 blocks.

The ```FormatNdef``` method formats the entire card for tags according to the NDEF specification. In general, the blocks on a Mifare Classic card may be assigned to multiple applications. The assignment is indicated by the Mifare Application Directory, which resides in sector 0 in all cards and also in sector 16 for cards larger than 1K. The previous code wasn't writing sector 16 correctly, and it wasn't using the correct checksum for the Mifare Application Directory in sector 16.

In addition, there are some miscellaneous fixes.

```EncodeSectorAndClockTailer``` was an unfortunate misspelling. It is renamed to ```EncodeSectorAndBlockTailer```, but for backward compatibility there is now a method with the old name that simply calls the new one.

Renamed the name of the first parameter of SetCapacity from ATAQ to ATQA.

Changed ```AuthenticateBlockKeyA``` to treat ```KeyA``` in the same way that ```AuthenticateBlockKeyB``` treats ```KeyB```.

# Verification
Verified that both 1K and 4K cards are now correctly formatted and that read/write access to the second half of 4K cards (with large sectors) work correctly. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2028)

### DIFF
--- a/src/devices/Card/Mifare/MifareCard.cs
+++ b/src/devices/Card/Mifare/MifareCard.cs
@@ -524,6 +524,7 @@ namespace Iot.Device.Card.Mifare
         /// <param name="accessTypes">An array of 3 AccessType determining access of each block</param>
         /// <returns>The 3 bytes encoding the rights</returns>
         /// This is a synonym of EncodeSectorAndBlockTailer (for backward compatibility)
+        [Obsolete("deprecated, use EncodeSectorAndBlockTailer instead")]
         public (byte B6, byte B7, byte B8) EncodeSectorAndClockTailer(AccessSector accessSector, AccessType[] accessTypes) =>
             EncodeSectorAndBlockTailer(accessSector, accessTypes);
 
@@ -913,8 +914,6 @@ namespace Iot.Device.Card.Mifare
         /// <param name="message">The NDEF Message to write</param>
         /// <param name="writeKeyA">True to write with Key A</param>
         /// <returns>True if success</returns>
-        /// TODO: check the MAD to determine which sectors can be used, instead of assuming
-        /// that all of them are available
         public bool WriteNdefMessage(NdefMessage message, bool writeKeyA = true)
         {
             if ((KeyB is not object or { Length: not 6 }) && (!writeKeyA))
--- a/src/devices/Card/Mifare/MifareCard.cs
+++ b/src/devices/Card/Mifare/MifareCard.cs
@@ -11,16 +11,21 @@ namespace Iot.Device.Card.Mifare
 {
     /// <summary>
     /// A Mifare card class
-    /// So far only supports the classical 1K cards
+    /// Supports Mifare Classic 1K and 4K
+    /// Also supports Mifare Plus 2K and 4K operating in SL1
     /// </summary>
     public class MifareCard
     {
+        private const byte BytesPerBlock = 16;
+        private const byte BlocksPerSmallSector = 4;
+        private const byte BlocksPerLargeSector = 16;
+        private const byte NumberOfSmallSectors = 32;
         private static readonly byte[] Mifare1KBlock1 = new byte[] { 0x14, 0x01, 0x03, 0xE1, 0x03, 0xE1, 0x03, 0xE1, 0x03, 0xE1, 0x03, 0xE1, 0x03, 0xE1, 0x03, 0xE1 };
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/devices/Card/Mifare/MifareCard.cs [hunk 1]
```diff
@@ -888,6 +913,8 @@ public bool FormatNdef(ReadOnlySpan<byte> keyB = default)
         /// <param name="message">The NDEF Message to write</param>
         /// <param name="writeKeyA">True to write with Key A</param>
         /// <returns>True if success</returns>
+        /// TODO: check the MAD to determine which sectors can be used, instead of assuming
```
Ellerbach, line 916: is the comment for the intellisense comment, meaning the developer needs to check it or it for the code and that needs to be done in a later fix/improvement?

src/devices/Card/Mifare/MifareCard.cs
```diff
@@ -487,7 +523,17 @@ public AccessType BlockAccess(byte blockNumber, byte[] sectorData)
         /// <param name="accessSector">The access desired</param>
         /// <param name="accessTypes">An array of 3 AccessType determining access of each block</param>
         /// <returns>The 3 bytes encoding the rights</returns>
-        public (byte B6, byte B7, byte B8) EncodeSectorAndClockTailer(AccessSector accessSector, AccessType[] accessTypes)
+        /// This is a synonym of EncodeSectorAndBlockTailer (for backward compatibility)
+        public (byte B6, byte B7, byte B8) EncodeSectorAndClockTailer(AccessSector accessSector, AccessType[] accessTypes) =>
```
Ellerbach, line 527: Can you mark this one as obsolete? So we can remove it in a further release?

Ellerbach (commented):
Thanks for the improvements and fixes. Just few comments and then we'll be good to go.

src/devices/Card/Mifare/MifareCard.cs
jdbruner: Yes, now marked as obsolete.

src/devices/Card/Mifare/MifareCard.cs
jdbruner: It is a note to the developer. I've removed it and will open a bug for the additional change.

Ellerbach (approved):
Thanks a lot for those fixes, looks all good to me

Ellerbach:
@krwq maybe you want to have a quick look so we can merge this one shortly?

krwq (approved):

krwq:
Thanks @jdbruner!

PR #2032, opened 2023-02-07
Board: Improve I²C device management to avoid null reference exceptions

### LINKED ISSUE #2031: Scanning the I²C bus leaves all devices open
## Description

`I2cBus.PerformBusScan()` opens all devices but never explicitly closes them. This results in a null reference exception when this method is called twice, or if the user tries to open a device found by the scan.

## Reproducing

```cs
// Show all connected FTDI devices
foreach (FtDevice device in FtCommon.GetDevices())
    Console.WriteLine($"{device.Type} (ID={device.Id})");

// I have one device connected (a FT232H)
using Ft232HDevice f232h = new(FtCommon.GetDevices().First());
I2cBus bus = f232h.CreateOrGetI2cBus(0);

// First scan succeeds
foreach (int address in bus.PerformBusScan())
    Console.WriteLine($"Found address {address:X2} ({address})");

// Second scan blows up
foreach (int address in bus.PerformBusScan())
    Console.WriteLine($"Found address {address:X2} ({address})");
```

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Iot.Device.Bindings
  StackTrace:
   at Iot.Device.Ft232H.Ft232HI2c.Read(Span`1 buffer)
   at Iot.Device.Ft232H.Ft232HI2c.ReadByte()
   at Iot.Device.Board.I2cBusExtensions.PerformBusScan(I2cBus bus, Int32 lowest, Int32 highest)
   at FtdiAdc.Core.Class1.Go() in C:\Users\scott\Documents\GitHub\Csharp-Data-Visualization\projects\iot\FtdiAdc\FtdiAdc.Core\Class1.cs:line 34
```

## Workaround

Manually releasing devices after the scan fixes the problem

```cs
foreach (int address in bus.PerformBusScan())
    Console.WriteLine($"Found address {address:X2} ({address})");

// release all created devices
for (int i = 0; i < 255; i++)
    bus.RemoveDevice(i);

foreach (int address in bus.PerformBusScan())
    Console.WriteLine($"Found address {address:X2} ({address})");
```

## Version

Tested on `net6.0-windows10.0.17763.0` with `Iot.Device.Bindings` (NuGet package) version 2.2.0

## Proposed fix

Add a `finally` block to the try/catch in the I2C Extensions class which manually removes all devices after creating them. I'll open a PR in a few minutes to resolve this issue.

Thanks to @Ellerbach for providing feedback in the Discord 🚀 

https://github.com/dotnet/iot/blob/a0db92f8c766771434095662ce2a0c0e47bec24c/src/devices/Board/I2cBusExtensions.cs#L26-L46

swharden:
Remove devices after creating them while scanning the I2C bus. 

Thanks to @Ellerbach for providing feedback in the Discord 🚀

Resolves #2031

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2032)


Ellerbach (commented):
Looks good to me. thanks. I let @pgrawehr to have a look.

raffaeler (approved):
Looks good!

Ellerbach (approved):

src/devices/Board/I2cBusExtensions.cs
```diff
@@ -91,6 +91,10 @@ public static (List<int> FoundDevices, int LowestAddress, int HighestAddress) Pe
                 {
                     // Ignore and continue
                 }
+                finally
+                {
+                    bus.RemoveDevice(addr);
+                }
```
pgrawehr, line 97: What hardware are you running on? At least `UnixI2cDevice.Dispose()` already calls `RemoveDevice`,  so this shouldn't be necessary.

krwq: agreed, we should be removing device from the bus in the I2cDevice.Dispose - I'm guessing FT232H might be missing that (assuming that's what you're using)

krwq: Seems that's already there
https://github.com/dotnet/iot/blob/main/src/devices/Ft232H/Ft232HI2c.cs#L50

might be worth debugging... `using` should be taking care of the RemoveDevice since it calls Dispose

krwq: I think the problem is here:
https://github.com/dotnet/iot/blob/main/src/devices/Ft232H/Ft232HI2cBus.cs#L36-L41

we're adding to a list even when error occurs

krwq (changes requested):
see @pgrawehr's comment

src/devices/Board/I2cBusExtensions.cs
krwq: (might be worth to review other similar implementations and see what they do)

swharden:
Thanks for your input everyone! I agree that the destructor _should_ be cleaning things up, and it remains unclear to me why it is not. For reference here is the test code I am using to evaluate this issue. The code change in this PR causes this failing test to pass.

I'll take a closer look to figure out exactly what is going wrong, and whether a more targeted code change could similarly resolve this issue.

```cs
// Licensed to the .NET Foundation under one or more agreements.

using System.Device.I2c;
using System.Linq;
using Iot.Device.Board;
using Iot.Device.Ft232H;
using Iot.Device.FtCommon;
using Xunit;

namespace System.Device.Gpio.Tests;

[Trait("feature", "ft232h")]
public class Ft232hTests
{
    [Fact]
    public void DevicesReleaseAfterScanning()
    {
        // I have one device connected (a FT232H)
        using Ft232HDevice f232h = new(FtCommon.GetDevices().First());
        I2cBus bus = f232h.CreateOrGetI2cBus(0);

        // First scan succeeds
        foreach (int address in bus.PerformBusScan())
        {
            Console.WriteLine($"Found address {address:X2} ({address})");
        }

        // Second scan throws a null reference exception
        foreach (int address in bus.PerformBusScan())
        {
            Console.WriteLine($"Found address {address:X2} ({address})");
        }
    }
}
```

swharden:
I haven't figured everything out yet (it's taking me a long time to debug this) but I have a better understanding of the core issue now. 

Here's a failing test:

```cs
[Fact]
public void PerformBusScan2()
{
    using Ft232HDevice f232h = new(FtCommon.FtCommon.GetDevices()[0]);
    I2cBus bus = f232h.CreateOrGetI2cBus(0);

    // open a device then dispose of it
    I2cDevice device1 = bus.CreateDevice(0x48);
    device1.ReadByte();
    device1.Dispose(); // this sets _i2cBus to null ⚠️

    // open the same device a second time
    I2cDevice device2 = bus.CreateDevice(0x48); // this succeeds
    device2.ReadByte(); // this throws a null reference exception ☠️
}
```

Disposing the I2C _device_ sets the _bus_ to `null` ... then it fails later when it tries to read from that null bus

https://github.com/dotnet/iot/blob/301f79574d664ab8af766e79bdc5d74b86ddb473/src/devices/Ft232H/Ft232HI2c.cs#L48-L54

swharden:
@krwq, thanks for your useful comments above. @Ellerbach I'm aware you're familiar with FT232H so I'll pose this to both of you. I think I arrived at a fix, but this has potential for unexpected behavior so I really value your input. 

Currently the FT232H's I2C _device_ destructor sets the _bus_ to `null` which is what causes these problems later (see my last comment).
https://github.com/dotnet/iot/blob/301f79574d664ab8af766e79bdc5d74b86ddb473/src/devices/Ft232H/Ft232HI2c.cs#L48-L54

This code fixes the issue, but it's not clear to me what the intent was of setting those variables to `null` inside the destructor in the first place. I'll update my PR to use this solution, but I want to check to make sure I'm not missing some intended functionality here. Thanks for your input!
```cs
{
    try
    {
        _i2cBus.RemoveDevice(_deviceAddress);
    }
    catch (ArgumentException)
    {
        // dont worry if the device has already been removed
    }

    base.Dispose(disposing);
}
```

src/devices/Board/I2cBusExtensions.cs
swharden: discussion continued in https://github.com/dotnet/iot/pull/2032#issuecomment-1423432449

src/devices/Ft232H/Ft232HI2c.cs
```diff
@@ -47,9 +47,15 @@ public override void WriteRead(ReadOnlySpan<byte> writeBuffer, Span<byte> readBu
         /// <inheritdoc/>
         protected override void Dispose(bool disposing)
         {
-            _i2cBus?.RemoveDevice(_deviceAddress);
-            _i2cBus = null!;
-            _settings = null!;
```
krwq, line 52: The reason this is set to null is so that GC can collect this. How could this code be causing issue? Is it possible the problem is actually related to this: https://github.com/dotnet/iot/pull/2032#discussion_r1100084654 ?

Ellerbach: OK, so the I2cBus should not then be set to null but settings still should.
Seems that there is some edge problem. not setting to null won't change much. If the I2cBus is not reused later, it will just be GC as well.

So then fix should more looks like:

```csharp
{
    try
    {
        _i2cBus?.RemoveDevice(_deviceAddress);
    }
    catch (ArgumentException)
    {
        // dont worry if the device has already been removed
    }

    _settings = null!;
    base.Dispose(disposing);
}
```

Ellerbach: @swharden thanks a lot for the deeper investigation.

jdbruner: Shouldn't there be a check for if(disposing) so that this doesn't attempt to use managed objects that may already have been GC'd during finalization?

swharden:
> sharing some code ... Possibly some common base class or something

I like this idea, but it seems out of scope for this PR and may cause merge issues with #2019 which appears to be actively worked on. Do you recommend I attempt these changes in this PR, or would it be a better idea to open an issue for it and work on it in a new PR after #2032 and #2019 are successfully merged?

> 3 failing and 8 successful checks

I see the CI release build has failed multiple times with errors that appear unrelated to the content of this PR. Can someone clarify whether this indicates there's a problem with my code modification, or if it's just an issue with the CI system? Thanks!

src/devices/Ft232H/Ft232HI2c.cs
swharden: I made the change suggested by @Ellerbach in https://github.com/dotnet/iot/pull/2032#discussion_r1101147901 and am pushing updates to this PR now

pgrawehr:
@swharden  #2019 is merged now, so you should be able to rebase and resolve the conflicts.

src/devices/Ft232H/Ft232HI2c.cs
```diff
@@ -47,8 +47,15 @@ public override void WriteRead(ReadOnlySpan<byte> writeBuffer, Span<byte> readBu
         /// <inheritdoc/>
         protected override void Dispose(bool disposing)
         {
-            _i2cBus?.RemoveDevice(_deviceAddress);
-            _i2cBus = null!;
+            try
+            {
+                _i2cBus.RemoveDevice(_deviceAddress);
+            }
+            catch (ArgumentException)
```
krwq, line 54: why is RemoveDevice throwing though?

jdbruner: It will throw if Dispose is called a second time, which is allowed. Dispose should be idempotent. The previous code won't call RemoveDevice in a second (or subsequent) invocation of Dispose, because _i2cBus will be null.

I confess I do not understand why setting _i2cBus to null in the Dispose method of one instance will cause that field to be null in a new instance.

krwq: for Device use this pattern: https://github.com/dotnet/iot/blob/main/src/System.Device.Gpio/System/Device/I2c/Devices/UnixI2cDevice.cs#L36-L53 but assume `_shouldDisposeBus` is always false

swharden:
I mimicked the behavior found in the Unix I2C [bus](https://github.com/dotnet/iot/blob/main/src/System.Device.Gpio/System/Device/I2c/UnixI2cBus.cs) and [device](https://github.com/dotnet/iot/blob/main/src/System.Device.Gpio/System/Device/I2c/Devices/UnixI2cDevice.cs) classes and the problem persists. Modifying the destructor to not set the bus to null solves this issue, but it's still unclear why.

At this point **I'm curious if this same bug is currently impacting Unix I2C systems too**. Can somebody easily test whether this is the case? Here's some sample code which may blow up on a Unix system, but I'm not sure:

```cs
I2cBus bus = myBoard.CreateOrGetI2cBus(0);

I2cDevice device1 = bus.CreateDevice(address);
device1.ReadByte();
device1.Dispose();

I2cDevice device2 = bus.CreateDevice(address);
device2.ReadByte(); // throws NullReferenceException
```

As an aside, what do you think about renaming `Ft232HI2c` to `Ft232HI2cDevice` to be consistent with other classes like `UnixI2cDevice`? I keep getting confused whether that class refers to the bus or the device since the name is ambiguous. To minimize impact of this breaking change I can create an empty class with the original name and mark it `obsolete` with a clear message indicating what it has been renamed to.

swharden:
... okay I made more progress and I now believe the bug is in `Board/I2cBusManager.cs`

If you create the same device twice, it doesn't actually instantiate a new I2C device class. Instead, it returns the old class stored in the `_devices` dictionary. This is a problem if our device was previously instantiated, disposed of externally, and something important internally was set to `null`. 

https://github.com/dotnet/iot/blob/3ff5e600f9af1acd9a4adb34f245c90b8eb9cecd/src/devices/Board/I2cBusManager.cs#L69-L79

There are a few different options for how to proceed, and since this now affects `I2cBusManager.cs` and I'm not very familiar with this library I'd really like your feedback on how best to proceed. One option is to gut all the device caching features and force `CreateDevice()` to instantiate a new device class every time it is called (consistent with its name). A less dramatic solution would be to stop setting the bus to null when devices are disposed of, but that doesn't feel great because it seems odd to dispose of a class you intend to re-use later.

The good news is that this is finally figured out! What a rabbit hole 🙃 


My gut reaction is to modify `Board.cs` to stop caching busses like this and force instantiating a new device class whenever ``

pgrawehr:
@swharden Thanks for tracking this down. I'll take a look.

pgrawehr:
@swharden I think you are right, the bus manager is just duplicating the logic of the actual bus implementation (`Ft232HI2cBus` in your case). And since it does not forward itself to the bus instance it manages, it doesn't get notified with the `RemoveDevice` calls. If I remember correctly, all bus instances keep a list of attached devices, so this isn't necessary on the bus manager. 

You can simulate part of the problem with these tests in `BoardTests.cs`: (Note that this uses a simulator implementation of I2cBus, which does not fully behave like the real thing - update it if needed)

```csharp
       [Fact]
        public void CreateDeviceTwice()
        {
            using Board board = CreateBoard();
            I2cBus bus = board.CreateOrGetI2cBus(0);

            // open a device then dispose of it
            I2cDevice device1 = bus.CreateDevice(0x55);
            device1.ReadByte();
            I2cDevice device2 = bus.CreateDevice(0x55);
            device2.ReadByte();
        }

        [Fact]
        public void CreateAndDisposeDeviceTwice()
        {
            using Board board = CreateBoard();
            I2cBus bus = board.CreateOrGetI2cBus(0);

            // open a device then dispose of it
            I2cDevice device1 = bus.CreateDevice(0x55);
            device1.ReadByte();
            device1.Dispose(); // this sets _i2cBus to null ⚠️

            // open the same device a second time
            I2cDevice device2 = bus.CreateDevice(0x55); // this succeeds
            device2.ReadByte(); // this throws a null reference exception ☠
            Assert.False(ReferenceEquals(device1, device2));
        }
```

The tests pass for me after I removed all references to the dictionary from `I2cBusManager`.

Additionally, it might have reduced the confusion if `Ft232HI2c.Write/Read` would throw an `ObjectDisposedException` instead of an NRE if `_i2cBus` is null (and similarly in other places). 

Feel free to rename `Ft232HI2c` to `Ft232HI2cDevice`. This is clearly a naming error, and in such cases we're not very peculiar about breaking changes. Question to @Ellerbach : Any reason this class is public, by the way? The ctor is internal. 

The tests that are in the `System.Device.Gpio.Tests` assembly are run on actual hardware (Raspberry Pis) with the unix driver during CI. So if you add some test there, you would see whether it works on linux.

swharden:
## Tests run on actual hardware

> The tests that are in the `System.Device.Gpio.Tests` assembly are run on actual hardware (Raspberry Pis) with the Unix driver during CI. So if you add some test there, you would see whether it works on Linux.

Wow, that's fantastic! Thanks for pointing me in this direction. In addition to just being a RasPI, it looks like there is expected to be a [BME280 I2C pressure sensor](https://www.mouser.com/datasheet/2/783/BST-BME280-DS002-1509607.pdf) at address `0x77` (`119`). I'll add some tests here to demonstrate this issue and clarify when it is resolved.

https://github.com/dotnet/iot/blob/58c8b14329181743728c36ba26e4384c4fa244bd/src/System.Device.Gpio.Tests/ProtocolTests.cs#L40-L48

## Solution: don't store devices in the bus manager

> all bus instances keep a list of attached devices, so this isn't necessary on the bus manager.

> The tests pass for me after I removed all references to the dictionary from I2cBusManager.

I'll fix this issue by removing the dictionary in the bus manager.

## Checking for null

> Additionally, it might have reduced the confusion if Ft232HI2c.Write/Read would throw an `ObjectDisposedException` instead of an NRE if _i2cBus is null (and similarly in other places).

This is the line that throws the null reference exception, but in theory every function in this class that interacts with the bus could throw for the same reason.

https://github.com/dotnet/iot/blob/58c8b14329181743728c36ba26e4384c4fa244bd/src/devices/Ft232H/Ft232HI2c.cs#L29-L32

Performing a null check on every read or write seems a bit costly. I agree `ObjectDisposedException` would be more informative than a `NullReferenceException`, but `_i2cBus` isn't nullable so I think the best way to proceed is to leave it how it is and assume it will never be `null`.

If you disagree let me know and I'll make it nullable and add null checks in all the functions of this class that reference the bus. I guess I2C is a pretty slow protocol anyway, so in context adding a bunch of null checks is probably a negligible hit to overall performance.

## Renaming

> Feel free to rename `Ft232HI2c` to `Ft232HI2cDevice`. This is clearly a naming error, and in such cases we're not very peculiar about breaking changes

Sounds good! Will do.

swharden:
@pgrawehr I believe I made all the recommended changes and this is ready to merge now. Let me know what you think!

TLDR:
* I added this failing test https://github.com/dotnet/iot/pull/2032/commits/6cef0f20cd68e24380dbf4e8b1400a1f35412d98
* I got it to pass by ensuring `I2cBusManager.CreateDevice()` always instantiates and returns a new device https://github.com/dotnet/iot/pull/2032/commits/4c679a01d0a76ab2905a60f09d129b7dc3a3d9bd
* Note that `I2cBusManager` still uses a dictionary to store created devices so `I2cBusManager.RemoveDevice()` can locate and dispose of them if users call that function instead of disposing of the device directly.

src/devices/Ft232H/Ft232HI2c.cs
```diff
@@ -7,51 +7,29 @@
 namespace Iot.Device.Ft232H
 {
     /// <summary>
-    /// I2C Device for FT232H
+    /// This class has been replaced by <see cref="Ft232HI2cDevice"/>
```
Ellerbach, line 10: If you don't use this one anymore, then let's go and remove it!

Ellerbach (commented):
I'm ok to rename the FT232HI2c yo FT232HI2cDevice. So don't keep the old one, I think it's confusing. And then please also adjust the sample and the read me

swharden:
> I'm ok to rename the FT232HI2c yo FT232HI2cDevice. So don't keep the old one

Done!

> And then please also adjust the sample and the read me

Supporting the previous point, I think both are okay as they are because they use `var`

https://github.com/dotnet/iot/blob/58c8b14329181743728c36ba26e4384c4fa244bd/src/devices/Ft232H/samples/Program.cs#L121

https://github.com/dotnet/iot/blob/58c8b14329181743728c36ba26e4384c4fa244bd/src/devices/Ft232H/README.md?plain=1#L76

> Use _devices[deviceAddress] = newDevice, since you want to always update. With this change, the oldest entry will persist, but you probably want to keep the newest.

Done!

---

I think this PR is ready to merge now, but let me know if you have any additional suggestions. Thanks for all your input along the way! 🚀

pgrawehr (approved):

Ellerbach:
@krwq is all ok for you now here? 
@swharden thanks for the persistence and making all the changes!

pgrawehr:
We need to update the compatibilitysuppressions still.

src/devices/Board/I2cBusManager.cs
```diff
@@ -68,13 +68,8 @@ public I2cBusManager(Board board, int bus, int[]? pins, I2cBus busInstance)
...
         public override I2cDevice CreateDevice(int deviceAddress)
         {
-            if (_devices.TryGetValue(deviceAddress, out I2cDevice? device))
-            {
-                return device;
-            }
-
             I2cDevice newDevice = _busInstance.CreateDevice(deviceAddress);
-            _devices.Add(deviceAddress, newDevice);
+            _devices[deviceAddress] = newDevice;
```
krwq, line 72: at this point we should consider making `_devices` either `HashSet<I2cDevice>` or `Dictionary<int, HashSet<I2cDevice>>` - perhaps even change design so that `RemoveDevice(int)` is replaced with `RemoveDevice(I2cDevice)`. Alternatively we should throw on CreateDevice if it was created again (we should also check if it was already disposed though). My personal favor is we should change the design.

@pgrawehr thoughts?

krwq (approved):
Let's make iterative changes from now on. My comment is still valid though

src/devices/Board/I2cBusManager.cs
pgrawehr: That would probably be better, but I'm not sure it's worth the effort. Would need to make the changes and then check whether they show any benefit.

krwq: let's do it separately if needed in that case - current state is still better than it was - that's why I approved in the first place

pgrawehr:
@swharden Thanks again for figuring this out.

PR #2044, opened 2023-02-18
A set of minor fixes

pgrawehr:
This PR includes a set of minor fixes for problems found while using some bindings.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2044)

### DIFF
--- a/src/devices/Arduino/FirmataCommandSequence.cs
+++ b/src/devices/Arduino/FirmataCommandSequence.cs
@@ -33,11 +33,24 @@ namespace Iot.Device.Arduino
         /// Create a new command sequence
         /// </summary>
         /// <param name="command">The first byte of the command</param>
-        internal FirmataCommandSequence(FirmataCommand command = FirmataCommand.START_SYSEX)
+        internal FirmataCommandSequence(FirmataCommand command)
         {
             _sequence = new List<byte>()
             {
                 (byte)command
+            };
+        }
+
+        internal FirmataCommandSequence(FirmataCommand command, int pin)
+        {
+            if (pin > 15)
+            {
+                throw new ArgumentOutOfRangeException(nameof(pin), "Shorthand commands can only be used with pin numbers <= 15");
+            }
+
+            _sequence = new List<byte>()
+            {
+                (byte)(((byte)command) | pin)
             };
         }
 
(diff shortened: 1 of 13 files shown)


src/devices/Arduino/FirmataCommandSequence.cs
```diff
@@ -33,14 +33,27 @@ public class FirmataCommandSequence : IEquatable<FirmataCommandSequence>
...
         {
             _sequence = new List<byte>()
             {
                 (byte)command
             };
         }
 
+        internal FirmataCommandSequence(FirmataCommand command, int pin)
+        {
+            if (pin > 15)
```
Ellerbach, line 46: Are all the boards with the same limit? If I'm using an ESP32 I can get more. So did I miss here?

src/devices/Arduino/FirmataDevice.cs
```diff
@@ -1074,23 +1077,53 @@ public void WriteReadI2cData(int slaveAddress,  ReadOnlySpan<byte> writeData, Sp
...
-            pwmCommandSequence.WriteByte((byte)pin);
             // The arduino expects values between 0 and 255 for PWM channels.
-            // The frequency cannot be set.
+            // The frequency cannot currently be set using the protocol
             int pwmMaxValue = _supportedPinConfigurations[pin].PwmResolutionBits; // This is 8 for most arduino boards
             pwmMaxValue = (1 << pwmMaxValue) - 1;
             int value = (int)Math.Max(0, Math.Min(dutyCycle * pwmMaxValue, pwmMaxValue));
+
+            // At most 14 bits used?
+            if (((pwmMaxValue & 0x3FFF) == pwmMaxValue) && (pin <= 15))
```
Ellerbach, line 1113: same here, why the 15 pins limits?

Ellerbach (approved):
Seems good to me, just few question on why some pin limitations. Maybe a small comment is enough if there is this limit. May help for maintenance.

src/devices/Arduino/FirmataCommandSequence.cs
pgrawehr: This is a limit of the command, not of the protocol or the board. There are some commands that have a one-byte shorthand form, but this form is limited to 15 pins, because the pin number is encoded in a nibble only. All such commands have a longer form as well, see the destinction below.

src/devices/Arduino/FirmataDevice.cs
pgrawehr: See other comment: If the pin number is <=15, a shorthand form of the command can be used.

PR #2045, opened 2023-02-21
add protocol awareness to NFC card transceivers

jdbruner:
NFC is a complex ecosystem with multiple protocols at different layers of the stack. NFC chipsets support varying combinations of these, including special handling for some protocol-specific commands. As an example, ISO/IEC 14443 is one set of RF standards, for which 14443-3 defines initialization and anticollision (of two types, A and B), and 14443-4 defines a transmission protocol atop 14443-3. However, some protocols, notably Mifare Classic and Mifare Ultralight, use a proprietary command set atop 14443-3. Mifare Classic uses a proprietary authentication scheme, so NFC chipsets implement this authentication protocol. (Alas, security by obscurity was not successful at preventing Mifare Classic authentication from being broken, particularly because it uses a relatively short 64-bit key.)

Some Mifare Classic commands require multiple RF exchanges - these include Write16Bytes, and counter operations. The PN532 provides direct support for these commands, but this must be handled in software for other transceivers.

The PN532 also provides direct support for other protocols, such as chaining and retransmission for 14443-4.

In .NET IoT, all card transceivers derive from the ```CardTransceiver``` abstract base class and override a specific implementation of the abstract ```Transceive``` method. This creates a common interface between the different NFC card types (```MifareCard```, ```Ultralight```, ```CreditCard```) and the NFC transceiver
```C#
public abstract int Transceive(byte targetNumber, ReadOnlySpan<byte> dataToSend, Span<byte> dataFromCard);
```

However, the diversity of protocols and transceiver support creates challenges to implement this interface. The various implementations must deduce the type of data exchange that is being performed and handle the special cases (such as Mifare authentication and Mifare writes) appropriately. However, this is difficult to do in general, because there are multiple commands, and the commands used by different protocols overlap. For example, Mifare Classic's ```AuthenticationA``` command uses the same command byte as Ultralight's ```GetVersion``` command. For transceivers that support other protocols as well (such as Innovision Jewel or FeliCa) there is an even greater risk that commands will overlap.

Therefore, this CR redefines the transceiver interface with a *breaking change*. A new enum, ```Iot.Device.Card.NfcProtocol``` defines protocols that may be understood specially by NFC transceivers. A new abstract property ```SupportedProtocols``` is added to ```CardTransceiver``` and overridden by each transceiver to indicate the protocols it supports, including:
- ISO/IEC 14443-3
- ISO/IEC 14443-4
- Mifare (Classic)
- Innovision Jewel
- JIS X 6319-4 (FeliCa)
- ISO/IEC 15693 (ICODE SLIX)

The ```Transceive``` method has an additional parameter that specifies the protocol that the caller intends to use for this transaction:
```C#
public abstract int Transceive(byte targetNumber, ReadOnlySpan<byte> dataToSend, Span<byte> dataFromCard, NfcProtocol protocol);
```

Each transceiver handles the protocols according to its specific implementation:
- ```Mfrc522``` uses this as an indication that it should handle a request as Mifare Classic (including authentiation and two-stage write operations) or 14443-3 (such as for Ultralight). [In theory the Mfrc522 could also be used for 14443-4, but its small FIFO is a significant obstacle.]
- ```Pn532``` uses this to determine whether it should invoke its special handlers for various protocols (Mifare, Jewel, FeliCa, 14443-4) or simply pass through the request (14443-3, notably for Ultralight)
- ```Pn5180``` uses this to recognize when it should perform special handling for Mifare Classic

```MifareCard``` always requests Mifare protocol. ```Ultralight``` requests 14443-3 except for the "compatibility write" command, which has the same command code as Mifare Classic's write command, and requires a two-step operation. ```CreditCard``` requests 14443-4.

In addition, two additional abstract properties are defined in ```CardTransceiver```:
```C#
public abstract uint MaximumWriteSize { get; }
public abstract uint MaximumWriteSize { get; }
```

so that callers can determine whether their requests are appropriately sized.

This also includes some miscellaneous fixes:
- added support for Mifare 2-step writes to PN5180
- changed the timeout for Type A transceive in the PN5180
- fixed a PN532 issue with commands that returned no data (e.g., Ultralight write was actually successful, but it was reported as a failure)

Tested on a Raspberry Pi 4B with MFRC522, PN532, and PN5180 using Mifare Classic and Ultralight cards. This change didn't impact the code path in PN5180 for CreditCard, but I was not able to verify that there are no regressions here.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2045)

### DIFF
--- a/src/devices/Card/NfcProtocol.cs
+++ b/src/devices/Card/NfcProtocol.cs
@@ -1,0 +1,58 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Iot.Device.Card
+{
+    /// <summary>
+    /// NFC protocol
+    /// These include standards as well as proprietary command sets, for which transceivers
+    /// may have special support. For example, Mifare is conveyed across ISO/IEC 14443-3 (Type A),
+    /// and transceivers have built-in support for Mifare authentication commands.
+    /// </summary>
+    [Flags]
+    public enum NfcProtocol
+    {
+        /// <summary>
+        /// Unknown or unspecified
+        /// </summary>
+        Unknown = 0,
+
+        /// <summary>
+        /// ISO/IEC 14443-3 (Type A or B)
+        /// </summary>
+        Iso14443_3 = (1 << 0),
+
+        /// <summary>
+        /// ISO/IEC 14443-4 (Type A or B)
+        /// </summary>
+        Iso14443_4 = (1 << 1),
+
+        /// <summary>
+        /// Mifare Classic
+        /// (proprietary commands on top of ISO/IEC 14443-3 Type A)
+        /// </summary>
+        Mifare = (1 << 2),
+
+        /// <summary>
+        /// Innovision Jewel
+        /// </summary>
+        Jewel = (1 << 3),
+
+        /// <summary>
+        /// JIS X 6319-4 (compatible with FeliCa)
+        /// </summary>
+        JisX6319_4 = (1 << 4),
+
+        /// <summary>
+        /// JIS X 6319-4 (compatible with FeliCa)
+        /// </summary>
+        FeliCa = JisX6319_4,
+
+        /// <summary>
+        /// ISO/IEC 15693
+        /// </summary>
+        Iso15693 = (1 << 5)
+    }
+}
--- a/src/devices/Card/NfcProtocol.cs
+++ b/src/devices/Card/NfcProtocol.cs
@@ -31,7 +31,7 @@ namespace Iot.Device.Card
 
         /// <summary>
         /// Mifare Classic
-        /// (proprietary commands on top of ISO/IEC 14443-3 Type A)
+        /// Proprietary commands on top of ISO/IEC 14443-3 Type A
         /// </summary>
         Mifare = (1 << 2),
 
@@ -41,14 +41,9 @@ namespace Iot.Device.Card
         Jewel = (1 << 3),
 
         /// <summary>
-        /// JIS X 6319-4 (compatible with FeliCa)
+        /// JIS X 6319-4. Compatible with FeliCa.
         /// </summary>
         JisX6319_4 = (1 << 4),
-
-        /// <summary>
-        /// JIS X 6319-4 (compatible with FeliCa)
-        /// </summary>
-        FeliCa = JisX6319_4,
 
         /// <summary>
         /// ISO/IEC 15693
--- a/src/devices/Mfrc522/Mfrc522.cs
+++ b/src/devices/Mfrc522/Mfrc522.cs
@@ -34,6 +34,15 @@ namespace Iot.Device.Mfrc522
         /// </summary>
         public const SpiMode DefaultSpiMode = SpiMode.Mode0;
 
+        /// <inheritdoc/>
+        public override uint MaximumReadSize => 62;     // 64-byte FIFO, 2 byte CRC
+
+        /// <inheritdoc/>
+        public override uint MaximumWriteSize => 62;    // 64-byte FIFO, 2 byte CRC
+
+        /// <inheritdoc/>
+        public override NfcProtocol SupportedProtocols => NfcProtocol.Iso14443_3 | NfcProtocol.Mifare;
+
         private readonly int _pinReset;
         private readonly ILogger _logger;
         private readonly SerialPort? _serialPort;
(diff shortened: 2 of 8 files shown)


src/devices/Card/NfcProtocol.cs
```diff
@@ -0,0 +1,58 @@
...
+        Iso14443_3 = (1 << 0),
+
+        /// <summary>
+        /// ISO/IEC 14443-4 (Type A or B)
+        /// </summary>
+        Iso14443_4 = (1 << 1),
+
+        /// <summary>
+        /// Mifare Classic
+        /// (proprietary commands on top of ISO/IEC 14443-3 Type A)
```
Ellerbach, line 34:
```suggestion
        /// Proprietary commands on top of ISO/IEC 14443-3 Type A
```

src/devices/Card/NfcProtocol.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        Mifare = (1 << 2),
+
+        /// <summary>
+        /// Innovision Jewel
+        /// </summary>
+        Jewel = (1 << 3),
+
+        /// <summary>
+        /// JIS X 6319-4 (compatible with FeliCa)
```
Ellerbach, line 44:
```suggestion
        /// JIS X 6319-4. Compatible with FeliCa.
```

src/devices/Card/NfcProtocol.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        JisX6319_4 = (1 << 4),
+
+        /// <summary>
+        /// JIS X 6319-4 (compatible with FeliCa)
```
Ellerbach, line 49:
```suggestion
        /// JIS X 6319-4. Compatible with FeliCa.
```

src/devices/Card/NfcProtocol.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        FeliCa = JisX6319_4,
```
Ellerbach, line 51: why defining this one equal to the previous one? Isn't 1 enough for both?

src/devices/Mfrc522/Mfrc522.cs
```diff
@@ -34,6 +34,15 @@ public class MfRc522 : CardTransceiver, IDisposable
         /// </summary>
         public const SpiMode DefaultSpiMode = SpiMode.Mode0;
 
+        /// <inheritdoc/>
+        public override uint MaximumReadSize => 62;     // 64-byte FIFO, 2 byte CRC
```
Ellerbach, line 38: maybe make those 3 new properties static?
They can then be used easily to understand if a protocol is supported. They are always the same regardless of the instance

src/devices/Pn5180/Pn5180.cs
```diff
@@ -54,6 +54,17 @@ public class Pn5180 : CardTransceiver, IDisposable
         /// </summary>
         public const SpiMode DefaultSpiMode = System.Device.Spi.SpiMode.Mode0;
 
+        /// <inheritdoc/>
+        public override uint MaximumReadSize => 508;
```
Ellerbach, line 58: Same comment here for the static properties (will apply to all the readers)

Ellerbach (commented):
Thank you very much for the proposal and the PR. We've been discussing this on Discord and I agree this is needed.
I'm fine all up with the implementation. It will make things much easier when we'll have to deal with even more protocols.
In the mean time, few comments.

src/devices/Card/NfcProtocol.cs
jdbruner: Sure; I'll change this (and the same pattern in the other cases also).

src/devices/Card/NfcProtocol.cs
jdbruner: In retrospect, I don't have a strong reason, so I'll remove the duplicate.

src/devices/Mfrc522/Mfrc522.cs
jdbruner: An interface can have static virtual properties, but I don't think an abstract class can. They could be defined as static properties in each specific Transceiver class, but the compiler won't be able to enforce consistency - possibly leading to unintentional errors such as misspellings.

One reason I made them instance properties is that the card types are constructed with a reference to the CardTransceiver. Then, for instance, UltralightCommand can check whether the response to a ReadFast is too big by comparing it against _rfid.MaximumReadSize

OK to keep them as-is?

jdbruner: Actually, I think SupportedProtocols makes more sense as a const (or static property), so I made that change. However, I think it is preferable to keep the maximum read/write size as a virtual property. OK?

jdbruner: I thought that ```static abstract``` was only for interfaces - and ```CardTransceiver``` is an abstract class, not an interface. As I recall, I tried it anyway, and I was rewarded with errors from the compiler. I Did I miss something?

minbaby:
@Ellerbach @jdbruner after merge this mr, [piplines](https://github.com/dotnet/iot/runs/11929450467)  failed. some one can resolve it.

PR #2050, opened 2023-03-15
Update for March servicing

richlander:
Happy Pi day!

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2050)

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,15 +1,15 @@
 {
   "tools": {
-    "dotnet": "7.0.200",
+    "dotnet": "7.0.202",
     "runtimes": {
       "dotnet/x64": [
         "3.1.32",
-        "6.0.14"
+        "6.0.15"
       ]
     }
   },
   "sdk": {
-    "version": "7.0.200"
+    "version": "7.0.202"
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "7.0.0-beta.23117.3",


joperezr:
Seems like the Button.Tests unit tests (cc: @raffaeler ) are failing on this PR, but not sure if they are just flaky or something here is affecting it which I doubt. I'll re-run to double-check.

raffaeler:
@joperezr I opened all the logs but I only see:
- `Platform 'esp32:esp32' not found: platform not installed`
- `##[error]Artifact BuildPackages already exists for build 89568.`

Could you please give me the link to the failed test?

joperezr:
Yeah, this is because if the unit test project that fails doesn't have the UI thread it won't log the errors because the runner has had a bug for a long time where it uses `Console.WriteLine` as opposed to `Log.WriteLine` which means that in a multi-threaded build, you would likely miss the errors getting printed. Seems like this is failing consistently on Windows, we should try to pull the changes on this branch locally and run those tests and we should be able to repro

raffaeler:
Got it, but where did you see the button tests failing?
Can you tell me at least which one was failing?

joperezr:
Merging since the Arduino leg has now been disabled and therefore won't re-run.

PR #2058, opened 2023-03-20
Adding RPI dtoverlay to check in configuration file

Ellerbach:
Adding RPI dtoverlay to check in configuration file. This is complementary from the existing elements that will return the default pins and configurations.
Those extra functions will actually check that the /boot/config.txt file contains the proper activation and will return the actual pin setup based o this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2058)


joperezr:
cc @pgrawehr 

This is one of the changes that might benefit from running the Arduino pipeline on demand. We should probably wait before merging this (or adding those changes here as well) in order to ensure that these changes aren't breaking the Arduino Board and compiler build.

pgrawehr:
Basically a nice idea, but I have some reservations regarding stability. Text comparisons aren't really stable for things like this, and we wouldn't even know whether the configuration file is the one that was used during boot. 

What problem are you trying to solve?

src/devices/Board/RaspberryPiBoard.cs
```diff
@@ -520,6 +615,215 @@ public override PinUsage DetermineCurrentPinUsage(int pinNumber)
             return PinUsage.Unknown;
         }
 
+        /// <summary>
+        /// Checks if the I2C overlay is activated in the configuraztion file.
+        /// </summary>
+        /// <returns>True if it is.</returns>
+        public bool IsI2cOverlayActivate()
```
krwq, line 622: nit: IsI2cOverlayActive or just IsI2cActived

src/devices/Board/samples/Board.Sample.csproj
```diff
@@ -4,11 +4,13 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>$(DefaultSampleTfms)</TargetFramework>
     <LangVersion>preview</LangVersion>
+    <EnableDefaultItems>false</EnableDefaultItems>
```
krwq, line 7: any reason for explicit?

src/devices/Board/tests/RpiBoardDtOverlayTests.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        public RpiBoardDtOverlayTests()
+        {
+            board = new RaspberryPiBoard();
+        }
+
+        [Fact]
+        public void CheckI2cConfig()
+        {
+            board.ConfigurationFile = Path.Combine("ConfigFiles", "config.I2c.txt");
```
krwq, line 29: do we also want to add actual tests for System.Device.Gpio.Tests? You could do similar test but use default path (make sure to add `[Trait(...)]` attribute similar to what there already is for PWM/SPI/I2C tests

krwq (approved):
Overall looks good, it would be good to use these in System.Device.Gpio.Tests - that would simplify diagnostics of CI failures

Ellerbach:
> Overall looks good, it would be good to use these in System.Device.Gpio.Tests - that would simplify diagnostics of CI failures

I think I will have to write specific tests to see what's activated because it's a bit of a chicken and egg then. If System.Device.Gpio.Tests depends on Board which depends back on System.Device.Gpio, not sure how that can work ;-)

So I was more think of a simple test just outputting the results of the I2C activation, SPI, PWM. So we know if it's there or not. And if there, the result for the pins setup as well. Makes sense?

src/devices/Board/samples/Board.Sample.csproj
Ellerbach: Yes, because there are now 2 sample projects. But I can definitely merge both. Thoughts?

src/devices/Board/RaspberryPiBoard.cs
Ellerbach: I'm open for any, I'll adjust then SPI and PWM. Any preference?

src/devices/Board/tests/RpiBoardDtOverlayTests.cs
Ellerbach: I added a specific test that output the configuration. It will always pass, it's not intended to fail, just outputting the configuration. I guess it's a good compromise

src/devices/Board/samples/Board.Sample.csproj
Ellerbach: merged both samples finally, it's easier

src/devices/Board/RaspberryPiBoard.cs
Ellerbach: I simplified, makes more sense

src/devices/Board/tests/RpiBoardDtOverlayTests.cs
Ellerbach: We can once merged find the best way to use it in the other tests

krwq (approved):

pgrawehr:
> > What problem are you trying to solve?
> 
> Main problem we're trying to solve is find a way to know in our test if a specific function is activated on the hardware. We know that one of the RPI do not have PWM but we don't know which one. And with a bit of a hack, still in the tests, we can add the potential missing string and reboot the board so next time the function will be activated. And if you think about it, it may help so we don't really need anyone to physically connect to the machines, adjust the config file.
> 
> The other element is for people to know if a specific function is activated or not. And we may even build like what we want to do for the test the string they have to add in the config file.

Yea, but wouldn't we just fail anyway, because the corresponding `/dev/` node (e.g. `/dev/pwm-0`) is not there? Maybe we should (additionally?) make sure that we get a nice looking error message in such a case. 

> 
> > Text comparisons aren't really stable
> 
> In general, I agree, in this case, if you don't have those exact strings, it won't get activated.

What about spaces and other white space in the strings? Wouldn't the settings still be valid? Or parameters in different order?

> 
> > we wouldn't even know whether the configuration file is the one that was used during boot
> 
> On RPI, it's always the same by default. And for people who'll really have to use this, the chances they'll change the file is very very low! Now, as you've seen, you can as well specify a path to any file, so if you've changed it and still want to check, then you can. That's what I am using for the tests.

Yes, it's physically the same file. But it might have been changed without a reboot. So the current state of `/boot/config.txt` might not match the actual device configuration. 
> 
> We can as well add serial and few others btw.

Ellerbach:
> Yea, but wouldn't we just fail anyway, because the corresponding `/dev/` node (e.g. `/dev/pwm-0`) is not there?

We would. It's just that you can know if things are setup properly or not.

> What about spaces and other white space in the strings? Wouldn't the settings still be valid? Or parameters in different order?

Spaces removed from beginning and trailing. They are not allowed in the middle of the string. The setup already take into consideration those. There maybe edge cases and I need to run few physical checks for some of those. The good news is that it's very strict all up.

src/devices/Board/tests/RpiBoardDtOverlayTests.cs
```diff
@@ -30,78 +33,78 @@ public void DisplayActivations()
...
             {
                 // This is not really a test but a way to see the configuration of the Raspberry PR used to run all the tests
-                Console.WriteLine("This is not really a test but a way to see the configuration of the Raspberry PR used to run all the tests.");
-                var isI2c = board.IsI2cActivated();
-                Console.WriteLine($"Is I2C overlay actvated? {isI2c}");
+                _output.WriteLine("This is not really a test but a way to see the configuration of the Raspberry PR used to run all the tests.");
+                var isI2c = _board.IsI2cActivated();
+                _output.WriteLine($"Is I2C overlay actvated? {isI2c}");
 
                 for (int busid = 0; busid < 2; busid++)
```
pgrawehr: Are all of the PIs in our CI PI 3's? Just asking, because RPI4 has up to 6 I2C busses (and Uarts, and SPIs)

src/devices/Board/RaspberryPiBoard.cs
```diff
@@ -19,10 +24,70 @@ namespace Iot.Device.Board
...
+            "dtoverlay=i2c1,pins_2_3",
+            "dtoverlay=i2c1,pins_44_45",
+            "dtoverlay=i2c3,pins_2_3",
+            "dtoverlay=i2c3,pins_4_5",
+            "dtoverlay=i2c4,pins_6_7",
+            "dtoverlay=i2c4,pins_8_9",
+            "dtoverlay=i2c5,pins_10_11",
+            "dtoverlay=i2c5,pins_12_13",
+            "dtoverlay=i2c6,pins_0_1",
+            "dtoverlay=i2c6,pins_22_23",
```
pgrawehr, line 48: If I read the documentation correctly, the parameter is optional. So it should be sufficient to test for `dtoverlay=i2c0`, `dtoverlay=i2c1` etc.

src/devices/Board/RaspberryPiBoard.cs
```diff
@@ -520,6 +615,215 @@ public override PinUsage DetermineCurrentPinUsage(int pinNumber)
...
+                return new int[0];
+            }
+
+            // Checks if there is an overlay because it does override the default configuration
+            var dtoverlay = _activateI2c.Where(m => m.StartsWith($"dtoverlay=i2c{busId}")).FirstOrDefault();
+            if (!string.IsNullOrEmpty(dtoverlay))
+            {
+                // we have an overlay, extract the pins and check them
+                // dtoverlay=i2c1,pins_2_3
+                var pins = dtoverlay.Substring(dtoverlay.IndexOf('_') + 1).Split('_');
```
pgrawehr, line 660: In the default case (when no pins are given), this will fail. So if `dtoverlay.IndexOf('_')` returns -1, the else case must be used (call `GetDefaultPinAssignmentForI2c()`)

src/devices/Board/tests/ConfigFiles/config.I2c.txt
```diff
@@ -0,0 +1,2 @@
+dtparam=i2c=on
+dtoverlay=i2c3,pins_2_3
```
pgrawehr, line 2: Add a test with just `dtoverlay=i2c0`

src/devices/Board/tests/RpiBoardDtOverlayTests.cs
```diff
@@ -0,0 +1,201 @@
...
+{
+    /// <summary>
+    /// Tests for the dtoverlay file for Raspberry Pi.
+    /// </summary>
+    public class RpiBoardDtOverlayTests
+    {
+        private readonly RaspberryPiBoard _board;
+        private readonly ITestOutputHelper _output;
+
+        public RpiBoardDtOverlayTests(ITestOutputHelper output)
```
pgrawehr, line 24: It seems you're running this test on any hardware. Isn't this supposed to fail?

pgrawehr (changes requested):

src/devices/Board/tests/RpiBoardDtOverlayTests.cs
krwq: AFAIK yes they are all Pi3s

src/devices/Board/RaspberryPiBoard.cs
Ellerbach: Let me test all this. I'll adjust like for SPI if not mandatory.

src/devices/Board/tests/RpiBoardDtOverlayTests.cs
Ellerbach: The added test displaying the configuration will never fail. And the rest of the tests are using predefined config files.

pgrawehr (approved):

joperezr:
Thanks @Ellerbach

PR #2061, opened 2023-03-25
change Mfrc522.ReselectTarget to use WUPA instead of REQA

jdbruner:
```Mfrc522.ReselectTarget``` is used to reselect a card, notably after an error occurs. It does a ```HALT``` and then uses ```IsCardPresent``` to reconnect with the card, which eventually uses ```REQA```. However, according to ISO 14443-3, if the card successfully executes a ```HALT``` then it will not respond to a ```REQA```; instead, the PCD is required to use ```WUPA``` (wakeup). As a result, if ```ReselectTarget``` is called when the card is not in an error state, the card is non-responsive until it is taken out of range of the transceiver so that a completely new session can begin.

This PR adds an optional ```reselect``` parameter (default ```false```) to ```IsCardPresent``` and also adds an optional ```commandCode``` parameter to the private function ```PiccRequestA``` to specify which command to send to the card. ```ReselectTarget``` uses this to request reselection with ```WUPA```.

With this change, it is now possible to halt and wake up cards. Calling ```ReselectTarget``` (typically using something like ```MifareCard.ReselectCard```) on a card that is not in error state no longer puts the card in a non-responsive state.

```Pn532.ReselectTarget``` already uses ```WUPA```., and ```Pn5180.ReselectTarget``` is not implemented for Type A.

### DIFF
--- a/src/devices/Mfrc522/Mfrc522.cs
+++ b/src/devices/Mfrc522/Mfrc522.cs
@@ -296,8 +296,9 @@ namespace Iot.Device.Mfrc522
         /// Check if a new card is present.
         /// </summary>
         /// <param name="atqa">ATQA buffer must be 2 bytes length and will contain the ATQA answer if there is a card.</param>
+        /// <param name="reselect">true if this is reselecting an existing card (e.g., after a halt)</param>
         /// <returns>true if there is a card, else false.</returns>
-        public bool IsCardPresent(byte[] atqa)
+        public bool IsCardPresent(byte[] atqa, bool reselect = false)
         {
             if (atqa is not object or { Length: not 2 })
             {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/devices/Mfrc522/Mfrc522.cs [hunk 1]
```diff
@@ -296,8 +296,9 @@ public bool ListenToCardIso14443TypeA(out Data106kbpsTypeA card, TimeSpan timeou
         /// Check if a new card is present.
         /// </summary>
         /// <param name="atqa">ATQA buffer must be 2 bytes length and will contain the ATQA answer if there is a card.</param>
+        /// <param name="reselect">true if this is reselecting an existing card (e.g., after a halt)</param>
         /// <returns>true if there is a card, else false.</returns>
-        public bool IsCardPresent(byte[] atqa)
+        public bool IsCardPresent(byte[] atqa, bool reselect = false)
```
joperezr, line 301: NIT: In order to not be binary breaking, any reason why we shouldn't just add new overloads as opposed to using default parameters?

joperezr:
> What do I need to do in order to resolve the API compatibility errors 

You just rerun the pack command locally and pass in the property `/p:ApiCompatGenerateSuppressionFile=true` so that the errors get baselined. I ran that locally so I'll just take the liberty to push those changes here. That said, I did ask the question above to consider just adding a new overload instead so as to not be binary breaking. @jdbruner if you end up deciding to do that, please just revert the change that I'm about to push as it won't be needed any longer.

jdbruner:
Thanks. I thought about creating an overload, but from a documentation perspective I think it is easier to understand the optional argument than the overloaded alternatives.

PR #2067, opened 2023-04-11
add support for Mifare application directory (resolves #2030)

### LINKED ISSUE #2030: MifareCard NDEF read/write don't work on multi-application cards
In the ```MifareCard``` class, ```WriteNdefMessage``` and ```TryReadNdefMessage``` have two issues:
1. They assume that the entire card is available for the NDEF message. A multi-application card could use some portion of the memory for other purposes. They should consult the Mifare Application Directory in sector 0 (and, for 2K and 4K cards, sector 16) to find the range of contiguous sectors that are assigned to the NDEF tag.
2. They do not account for the larger sector size in the second half of 4K cards. (The first 32 sectors have 4 blocks per sector; the last 8 sectors of a 4K card have 16 blocks per sector.)

jdbruner:
Add support for the Mifare Application Directory, as described in [AN10787](https://www.nxp.com/docs/en/application-note/AN10787.pdf).

The application directory occupies sector 0 (blocks 1 and 2) and - for 2K and 4K cards - sector 16 (blocks 64, 65, and 66). The directory is actually a sector map that defines the assignment of sectors to applications. Applications are identified by a 16-bit value (encoded in little-endian order on two bytes).

Previously, ```MifareCard.FormatNdef``` created the application directory, assigning all sectors to NDEF content. The NDEF content could be read and written with ```TryReadNdefMessage``` and ```WriteNdefMessage```. However, as noted in #2030, these methods assume that the entire card has been formatted for NDEF, so if the directory was formatted elsewhere and specifies something else, they do not work correctly. In addition, they do not handle the last eight sectors of a 4K card correctly (because those sectors are a different size than the first 32 sectors).

This contribution adds a new class ```MifareDirectory``` to manage the Mifare application directory for a ```MifareCard```. The new class ```MifareDirectoryEntry``` describes a directory entry (i.e., the assignment of sectors to an application). The new class ```MifareApplicationIdentifier``` represents an application identifier.

```MifareDirectory.CreateEmpty``` creates an empty ```MifareDirectory``` object for a specified ```MifareCard```.

```MifareDirectory.LoadFromCard``` loads the directory from a specified ```MifareCard```.

```GetApplications``` returns an enumeration of ```MifareDirectoryEntry``` objects representing the directory contents. ```TryGetApplication``` returns the ```MifareDirectoryEntry``` of a specified application, if it exists. Sectors are allocated to applications with ```Allocate``` and freed with ```Free```. After the directory has been modified, it must be written back to the card with ```StoreToCard```.

```FormatNdef```, ```IsFormattedNdef```, ```WriteNdefMessage```, and ```TryReadNdefMessage``` have been reimplemented to use the directory. An overload of ```FormatNdef``` can specify that only a portion of the card is to be allocated to NDEF content. The remainder is free and can be assigned to other applications. For example, to format 14 sectors for NDEF and assign one sector to an issuer-specific application ID, the following could be used:

```C#
        // MifareCard mifareCard was created previously
        var res = mifareCard.FormatNdef(14);
        if (res)
        {
            var cardIssuerId = new MifareApplicationIdentifier(0xFF01);
            var directory = MifareDirectory.LoadFromCard(mifareCard);
            var entry = directory?.Allocate(cardIssuerId, 1);
            res = entry is object;
            if (res)
            {
                directory!.CardPublisherSector = entry!.FirstSector;
                res = directory.StoreToCard(mifareCard.KeyB);
            }
        }
```

This was also added to the Pn532 sample code. In addition, some miscellaneous changes were made to both the Pn532 and Mfrc522 sample code (including a workaround for #1869).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2067)

### DIFF
--- a/src/devices/Card/Mifare/MifareApplicationIdentifier.cs
+++ b/src/devices/Card/Mifare/MifareApplicationIdentifier.cs
@@ -1,0 +1,157 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Buffers.Binary;
+
+namespace Iot.Device.Card.Mifare
+{
+    /// <summary>
+    /// Mifare application identifier as defined in https://www.nxp.com/docs/en/application-note/AN10787.pdf
+    /// This identifies an application within a Mifare application directory.
+    /// It is an unsigned 16-bit quantity (2 bytes, little-endian order)
+    /// The high-order 8 bits identify the function cluster, and
+    /// the low-order 8 bits define an application code within that cluster
+    /// </summary>
+    public struct MifareApplicationIdentifier : IEquatable<MifareApplicationIdentifier>
+    {
+        /// <summary>
+        /// Construct a MifareApplicationIdentifier from an ushort
+        /// </summary>
+        /// <param name="appId">application identifier as a ushort</param>
+        public MifareApplicationIdentifier(ushort appId) => _appId = appId;
+
+        /// <summary>
+        /// Construct a MifareApplicationIdentifier from a sequence of bytes
+        /// </summary>
+        /// <param name="bytes">two bytes representing the application identifier</param>
+        /// <exception cref="ArgumentException">the input is not 2 bytes long</exception>
+        public MifareApplicationIdentifier(ReadOnlySpan<byte> bytes) =>
+            _appId = BinaryPrimitives.ReadUInt16LittleEndian(bytes);
+
+        /// <summary>
+        /// Copy the MifareApplicationIdentifier to a sequence of bytes
+        /// </summary>
+        /// <param name="bytes">bytes that will receive the value</param>
+        /// <exception cref="ArgumentException">the output span is not 2 bytes long</exception>
+        public void CopyTo(Span<byte> bytes) => BinaryPrimitives.WriteUInt16LittleEndian(bytes, _appId);
+
+        /// <summary>
+        /// The function cluster for this application identifier (high-order 8 bits)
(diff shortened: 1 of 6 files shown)


src/devices/Card/Mifare/MifareApplicationIdentifier.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// <summary>
+        /// Identifies that a sector does not exist
+        /// This is used for entries in the directory that are beyond the end of the card,
+        /// for example, sectors 32 through 39 of a 2K card
+        /// </summary>
+        public static MifareApplicationIdentifier AdminSectorNotApplicable = new(0x0005);
+
+        #endregion
+
+        private readonly ushort _appId;
```
Ellerbach, line 155: we usually place all the private variable first

src/devices/Card/Mifare/MifareDirectory.cs
```diff
@@ -0,0 +1,412 @@
...
+
+                appIdBytes.CopyTo(_data.AsSpan(index));
+                index += 2;
+            }
+
+            RecalculateCrc();
+            return new MifareDirectoryEntry(appId, entry.FirstSector, numberOfSectors);
+        }
+
+        private readonly MifareCard _card;
```
Ellerbach, line 409: same, can you please move the private fields declaration on the top?

Ellerbach (approved):
Great improvement and much cleaner than originaly. Thanks for the improvement. Minor comments.

PR #2069, opened 2023-04-12
Remove image sharp

### LINKED ISSUE #1403: We should replace dependency on System.Drawing with ImageSharp

### LINKED ISSUE #1767: Library requirements to replace System.Drawing
This discussion was opened [here](https://github.com/dotnet/iot/issues/1403#issuecomment-1016397355). I now opened this thread to collect the pros/cons for any proposal coming from the community members.

As stated by @krwq, the main things to consider for the libraries being proposed here are the following:
- license
- support level / responsiveness on github (i.e. in case of blocking issue or security issue how long will we wait for the update; how long if we contribute fix ourselves)
- small fonts support (i.e. will it look good with 32x32 LED matrix or some other tiny display)
- transition difficulty
- anything relevant to dotnet/iot scenarios you can think of

I also stressed about the importance of these other following points:
- ability to easily create "drivers" to support any of the existing and future displays (typically interfaced via SPI/I2C interfaces)
- the client code should be ideally the same when using the standard graphics card / GPU (HDMI output) or a simpler SPI/I2C device. This would allow to move the (drawing) code from one device to another with a zero or low effort.
- possibly supporting acceleration where available (RPi has a GPU)
These last requirements are not mandatory, but I believe they could be a big plus.

Any discussion beyond the requisited can done "in-person" on the discord channel (displays area): https://discord.gg/GyEfttux (my handle is raf)

## Things that need to be changed

Here is the list of references to System.Drawing in the repo:
- samples\led-matrix-weather
- src\devices\Apa102 -> seems to only use Color
- src\devices\Bh1745 -> seems to only use Color
- src\devices\Common\Iot\Device\Graphics -> seems to only use Color
- src\devices\GoPiGo3 -> seems to only use Color
- src\devices\Media\VideoDevice -> uses Color and PixelFormat, Size and maybe few others. This will be more complicated to migrate
- src\devices\PiJuice -> seems to only use Color
- src\devices\RGBLedMatrix -> Color, Rectangle and most likely few others. This may require quite some efforts
- src\devices\SenseHat -> seems to only use Color
- src\devices\Ssd1351 -> uses Color, Bitmap, Point. This will require quite some efforts
- src\devices\Tcs3472x -> seems to only use Color
- src\devices\Ws28xx -> seems to only use Color

Those only using Color are quite straight forward to migrate. They require always the same pattern to transform the color space to the right one. A limited hardware can be needed to test but the pattern will always be the same. So that can limit the errors when the hardware is not available.

The problematic ones will be VideoDevice, RGBLedMAtrix and Ssd1351. They will for sure require the hardware to test and make sure all is working perfectly.

/cc @A-J-Bauer @krwq

pgrawehr:
Fixes #1767 (and probably a few others)

- Removes all references to ImageSharp from Iot.Device.Bindings.dll
- Replace standard types (Point, Color, Size, Rectangle) with the types from `System.Drawing.Primitives.dll` (these are supported on all platforms and even part of the core runtime)
- Provide a new `BitmapImage` class, which is an abstract image type
- Provide a `ImageFactoryRegistry` static class to create instances of images
- Provide a `SkiaSharpConnector` as (one possible) adapter that provides the actual implementation.

I suggest to create a new package for the SkiaSharpConnector (maybe better name it SkiaSharpAdapter?), because other adapters are possible (e.g. plain old System.Drawing.Imaging would do if only used on Windows) and because SkiaSharp is a quite heavy dependency (~100MB with the native dependencies)

Rough testing done, but a quick test of all the heavier affected bindings would be good:
- VideoDevice (@raffaeler ?)
- RgbLedMatrix (@richlander ?)
- Ws28xx (@krwq ?)
- Ssd1351 (TBD)
- BuildHat (@Ellerbach ?)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2069)

### DIFF
--- a/src/Iot.Device.Bindings.SkiaSharpAdapter/Directory.Build.props
+++ b/src/Iot.Device.Bindings.SkiaSharpAdapter/Directory.Build.props
@@ -1,0 +1,13 @@
+<Project>
+  <!-- Packaging related properties -->
+  <PropertyGroup>
+    <MajorVersion>2</MajorVersion>
+    <MinorVersion>3</MinorVersion>
+    <Description>This package contains the BitmapImage adapter using SkiaSharp. This library is an optional package to provide imaging functionality for Iot.Device.Bidnings.dll</Description>
+    <PackageTags>.NET Core IoT Device Bindings SkiaSharp Adapter</PackageTags>
+    <EnablePackageValidation>false</EnablePackageValidation>
+    <PackageValidationBaselineVersion>2.2.0</PackageValidationBaselineVersion>
+  </PropertyGroup>
+
+  <Import Project="../../Directory.Build.props" />
+</Project>
--- a/src/Iot.Device.Bindings.SkiaSharpAdapter/Iot.Device.Bindings.SkiaSharpAdapter.csproj
+++ b/src/Iot.Device.Bindings.SkiaSharpAdapter/Iot.Device.Bindings.SkiaSharpAdapter.csproj
@@ -1,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <LangVersion>10</LangVersion>
+    <Nullable>enable</Nullable>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <EnableDefaultItems>true</EnableDefaultItems>
+    <IsPackable>true</IsPackable>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
+    <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage)</TargetsForTfmSpecificContentInPackage>
+  </PropertyGroup>
+  
+  <ItemGroup>
+    <Compile Include="../devices/SkiaSharpAdapter/*.cs" />
+  </ItemGroup>
+  
+  <ItemGroup>
+    <ProjectReference Include="..\Iot.Device.Bindings\Iot.Device.Bindings.csproj" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractionsPackageVersion)" />
+    <PackageReference Include="SkiaSharp" Version="2.88.3" />
+    <!-- The NativeAssets packages for Windows and MacOS are included in the above by default -->
+    <PackageReference Include="SkiaSharp.NativeAssets.Linux" Version="2.88.3" />
+  </ItemGroup>
+
+
+</Project>
--- a/src/Iot.Device.Bindings/CompatibilitySuppressions.xml
+++ b/src/Iot.Device.Bindings/CompatibilitySuppressions.xml
@@ -3,6 +3,20 @@
 <Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
   <Suppression>
     <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:BuildHat.Models.BuildHatInformation</Target>
+    <Left>lib/net6.0/Iot.Device.Bindings.dll</Left>
+    <Right>lib/net6.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:BuildHat.Models.LedMode</Target>
+    <Left>lib/net6.0/Iot.Device.Bindings.dll</Left>
+    <Right>lib/net6.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
     <Target>T:Iot.Device.Common.ValueArray`1</Target>
     <Left>lib/net6.0/Iot.Device.Bindings.dll</Left>
     <Right>lib/net6.0/Iot.Device.Bindings.dll</Right>
@@ -18,6 +32,13 @@
   <Suppression>
     <DiagnosticId>CP0001</DiagnosticId>
     <Target>T:Iot.Device.Ft232H.Ft232HI2c</Target>
+    <Left>lib/net6.0/Iot.Device.Bindings.dll</Left>
+    <Right>lib/net6.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:Iot.Device.Media.PixelFormat</Target>
     <Left>lib/net6.0/Iot.Device.Bindings.dll</Left>
     <Right>lib/net6.0/Iot.Device.Bindings.dll</Right>
     <IsBaselineSuppression>true</IsBaselineSuppression>
@@ -150,6 +171,20 @@
   </Suppression>
   <Suppression>
     <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:BuildHat.Models.BuildHatInformation</Target>
+    <Left>lib/netcoreapp3.1/Iot.Device.Bindings.dll</Left>
+    <Right>lib/netstandard2.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:BuildHat.Models.LedMode</Target>
+    <Left>lib/netcoreapp3.1/Iot.Device.Bindings.dll</Left>
+    <Right>lib/netstandard2.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
     <Target>T:Iot.Device.Display.BarColor</Target>
     <Left>lib/netcoreapp3.1/Iot.Device.Bindings.dll</Left>
     <Right>lib/netstandard2.0/Iot.Device.Bindings.dll</Right>
@@ -159,6 +194,27 @@
     <DiagnosticId>CP0001</DiagnosticId>
     <Target>T:Iot.Device.Ft232H.Ft232HI2c</Target>
     <Left>lib/netcoreapp3.1/Iot.Device.Bindings.dll</Left>
+    <Right>lib/netstandard2.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:Iot.Device.Media.PixelFormat</Target>
+    <Left>lib/netcoreapp3.1/Iot.Device.Bindings.dll</Left>
+    <Right>lib/netstandard2.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:BuildHat.Models.BuildHatInformation</Target>
+    <Left>lib/netstandard2.0/Iot.Device.Bindings.dll</Left>
+    <Right>lib/netstandard2.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:BuildHat.Models.LedMode</Target>
+    <Left>lib/netstandard2.0/Iot.Device.Bindings.dll</Left>
     <Right>lib/netstandard2.0/Iot.Device.Bindings.dll</Right>
     <IsBaselineSuppression>true</IsBaselineSuppression>
   </Suppression>
@@ -179,6 +235,13 @@
   <Suppression>
     <DiagnosticId>CP0001</DiagnosticId>
     <Target>T:Iot.Device.Ft232H.Ft232HI2c</Target>
+    <Left>lib/netstandard2.0/Iot.Device.Bindings.dll</Left>
+    <Right>lib/netstandard2.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:Iot.Device.Media.PixelFormat</Target>
     <Left>lib/netstandard2.0/Iot.Device.Bindings.dll</Left>
     <Right>lib/netstandard2.0/Iot.Device.Bindings.dll</Right>
     <IsBaselineSuppression>true</IsBaselineSuppression>
@@ -276,6 +339,34 @@
   </Suppression>
   <Suppression>
     <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:Iot.Device.CharacterLcd.Hd44780.#ctor(SixLabors.ImageSharp.Size,Iot.Device.CharacterLcd.LcdInterface)</Target>
+    <Left>lib/net6.0/Iot.Device.Bindings.dll</Left>
+    <Right>lib/net6.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:Iot.Device.CharacterLcd.LcdRgb.#ctor(SixLabors.ImageSharp.Size,Iot.Device.CharacterLcd.LcdInterface,System.Device.I2c.I2cDevice)</Target>
+    <Left>lib/net6.0/Iot.Device.Bindings.dll</Left>
+    <Right>lib/net6.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:Iot.Device.CharacterLcd.LcdRgb.#ctor(SixLabors.ImageSharp.Size,System.Device.I2c.I2cDevice,System.Device.I2c.I2cDevice)</Target>
+    <Left>lib/net6.0/Iot.Device.Bindings.dll</Left>
+    <Right>lib/net6.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:Iot.Device.CharacterLcd.LcdRgb.SetBacklightColor(SixLabors.ImageSharp.Color)</Target>
+    <Left>lib/net6.0/Iot.Device.Bindings.dll</Left>
+    <Right>lib/net6.0/Iot.Device.Bindings.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
     <Target>M:Iot.Device.Common.GeographicPosition.GetDegreesMinutesSeconds(System.Double,System.Int32,System.Double@,System.Double@,System.Double@,System.Double@)</Target>
     <Left>lib/net6.0/Iot.Device.Bindings.dll</Left>
     <Right>lib/net6.0/Iot.Device.Bindings.dll</Right>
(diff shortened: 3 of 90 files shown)


joperezr:
[Traige]: FYI @raffaeler and @krwq, we are going to do a design review next week to look into this one in more detail. @pgrawehr will go over the problem and design, so we can try to get this in for 3.0.

pgrawehr:
[Triage] 

With Jose's words, this PR tries to solve these problems:
- We need to get rid of ImageSharp (because it changed license conditions)
- We try to avoid just replacing all ImageSharp types we currently expose to whatever new API we use, to avoid further breaking changes in case we need to change APIs again
- Instead, in the APIs we expose, we only use system types (e.g. from System.Drawing.Primitives) and some self-made abstraction types (e.g. `BitmapImage`)
- We provide an implementation for this abstraction (using SkiaSharp), 
- Optionally, we might provide an implementation using other implementations, like the old one from System.Drawing.Common, for Windows only. 
- The abstraction layer sits on top of `Iot.Device.Bindings`, so that it doesn't need to be referenced when not needed.

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
```diff
@@ -22,6 +22,7 @@
     <!-- Excluding samples and test projects when getting source files -->
     <_ExcludeProjectReferences Include="$(DeviceRoot)**/samples/**/*.csproj" />
     <_ExcludeProjectReferences Include="$(DeviceRoot)**/tests/**/*.csproj" />
+    <_ExcludeProjectReferences Include="$(DeviceRoot)SkiaSharpConnector/*.csproj" />
```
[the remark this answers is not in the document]
joperezr: Oh I see you created an extra package for that. Can you explain what is the reasoning for separating that piece?

src/devices/Arduino/ArduinoAnalogController.cs
```diff
@@ -35,60 +35,58 @@ public override int PinCount
             get;
         }
 
-        public override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber)
+        /// <inheritdoc />
```
joperezr, line 38: I'm failing to understand how all of the changes in the Arduino device are related to this PR. Are those just there because when working on the sample you made some adjustments? Does it make sense to separate these changes into a separate PR?

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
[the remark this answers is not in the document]
pgrawehr: That's by design. The intent is that the connector is in an extra nuget, so that it can be exchanged for another implementation without touching Iot.device.bindings or its API. That means that we can provide different connectors without affecting the API (or the user could write their own).

src/devices/Arduino/ArduinoAnalogController.cs
pgrawehr: Yes, most of these changes are improvements for the example. I'm considering to split that into a separate PR, but I'm waiting until the design has been roughly finalized, to avoid diverging branches.

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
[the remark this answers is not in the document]
joperezr: I'm not following why keeping it as part of IoT.Device.Bindings would **not allow for an external implementation without touching Iot.device.bindings**. The part that seems wrong to me, is that we are adding an abstract class with no default implementation on the main library, which seems wrong, as ideally we would want IoT.Device.bindings to not require additional packages on top for some bindings to work. Model of having an included implementation on the library, plus the ability to use a different one via a NuGet package seems to satisfy both goals.

src/devices/Arduino/ArduinoAnalogController.cs
joperezr: Got it, yeah, we'd probably want those in a separate PR. If anything, it'll make reviewing this PR easier 😃

src/Iot.Device.Bindings/Iot.Device.Bindings.csproj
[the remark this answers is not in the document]
pgrawehr: I want to avoid that the iot.device.bindings again references large dependencies. SkiaSharp's native implementation file is ~10MB per platform, and one might not want to carry this dependency around if one doesn't need it. There are many bindings that don't need any imaging features, so if you only use those bindings, you would not need to have a large dependency lying around for nothing. 

So either we have some bindings that require an additional package to work, or we have many bindings that pull a dependency they don't need. I prefer the first solution.

pgrawehr:
This is now mostly ready for review. A few points still:
- The new package needs to be generated. Haven't found how to do that yet.
- The example code and the actual graphics library changes should be separated to separate PRs or at least commits.

pgrawehr:
I have now rebased the PR and separated the changes into logical units. Even though they're not fully independent, it should be easier to review now. 

I suggest to use Rebase or real merge instead of squash for this PR, to keep the stuff separated in history.

src/Iot.Device.Bindings/CompatibilitySuppressions.xml
```diff
@@ -1,6 +1,20 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <!-- https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/diagnostic-ids -->
 <Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:BuildHat.Models.BuildHatInformation</Target>
```
joperezr, line 6: It's really hard to go through all of these, but I'm assuming they are pretty much all around removing APIs that exposed ImageSharp types, right?

src/Iot.Device.Bindings.SkiaSharpAdapter/Directory.Build.props
```diff
@@ -0,0 +1,13 @@
+<Project>
+  <!-- Packaging related properties -->
+  <PropertyGroup>
+    <MajorVersion>2</MajorVersion>
+    <MinorVersion>3</MinorVersion>
+    <Description>This package contains the BitmapImage adapter using SkiaSharp. This library is an optional package to provide imaging functionality for Iot.Device.Bidnings.dll</Description>
+    <PackageTags>.NET Core IoT Device Bindings SkiaSharp Adapter</PackageTags>
+    <EnablePackageValidation>false</EnablePackageValidation>
+    <PackageValidationBaselineVersion>2.2.0</PackageValidationBaselineVersion>
```
joperezr, line 9: I'm fine if you want to keep this as is since PackageValidation is disabled, but it might be better to just introduce this property when we actually enable PackageValidation.

src/Iot.Device.Bindings.SkiaSharpAdapter/Iot.Device.Bindings.SkiaSharpAdapter.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <LangVersion>10</LangVersion>
+    <Nullable>enable</Nullable>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <EnableDefaultItems>true</EnableDefaultItems>
+    <IsPackable>true</IsPackable>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
joperezr, line 10: Do we need to define this? I thought we had it in one of the props files.

src/Iot.Device.Bindings.SkiaSharpAdapter/Iot.Device.Bindings.SkiaSharpAdapter.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
```
joperezr, line 11: Why do we need the ref assembly?

src/devices/Arduino/samples/ApiChecker/TestCases.cs
```diff
@@ -57,48 +59,20 @@ public static void Run(ArduinoBoard board)
...
-            Console.ReadKey();
-            pin.DisableAnalogValueChangedEvent();
-            pin.Dispose();
-            analogController.Dispose();
-        }
-
-        private static int GetAnalogPin1(ArduinoBoard board)
-        {
-            int analogPin = 15;
+            int analogPin;
```
joperezr, line 64: Seems like the PR still has non-ImageSharp-related changes. I understand why RgbLedTest is changing in this project, but might be easier to review if these other things are moved to a separate PR.

pgrawehr: It's hard to fully separate the stuff, because the changes depend on each other. Neither of the changes will build without the other. But I separated it in groups if you look at the individual commits.

src/Iot.Device.Bindings/CompatibilitySuppressions.xml
pgrawehr: There are mainly three kinds of changes: 
- Apis that were changed from ImageSharp to System.Drawing (e.g. `ImageSharp.Color` to `System.Drawing.Color`)
- Apis that were using ImageSharp bitmap types now take `Iot.Device.Common.BitmapImage`
- Individual types that were in incorrect namespaces

For the user, the changes typically only require adapting their using definitions.

pgrawehr:
Isolated the relevant changes now. The Ili9341 binding is temporarily broken now, though. The fix will be provided in a separate PR.

src/Iot.Device.Bindings.SkiaSharpAdapter/Directory.Build.props
```diff
@@ -0,0 +1,13 @@
+<Project>
+  <!-- Packaging related properties -->
+  <PropertyGroup>
+    <MajorVersion>2</MajorVersion>
```
krwq, line 4: do we actually care about those versions? Presumably adapter is not getting shipped?

krwq: wait a sec.. looking at code later seems like this will be a new package, do we actually need it? could this be constructed similarly to other common projects for Iot.Device.Bindings? I might have missed the triage meeting where you discussed this

src/devices/Arduino/samples/ApiChecker/RgbLedTest.cs
```diff
@@ -57,8 +55,7 @@ public void DoTest()
             float blueCorrection = 0.95f;
             while (!Console.KeyAvailable)
             {
-                var hsv = new Hsv(angle, 1.0f, 1.0f);
-                var rgb = converter.ToRgb(hsv);
+                Color rgb = Color.Blue;
```
krwq, line 58: isn't this sample just displaying blue rather than color changing over time?

src/devices/BuildHat/Models/LedMode.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-namespace BuildHat.Models
+namespace Iot.Device.BuildHat.Models
```
krwq, line 4: does this need sample update?

src/devices/CharacterLcd/samples/LcdConsole.Samples.cs
```diff
@@ -163,5 +163,91 @@ public static void WriteTest(ICharacterLcd lcd)
             console.Write("Test finished");
             console.Dispose();
         }
+
+        /// <summary>
+        /// A test method
+        /// </summary>
+        /// <param name="lcd">The lcd driver</param>
+        public static void ValueTest(ICharacterLcd lcd)
```
krwq, line 171: where are you calling this new method?

src/devices/Common/Iot/Device/Graphics/BitmapImage.cs
```diff
@@ -49,12 +56,82 @@ protected BitmapImage(byte[] data, int width, int height, int stride)
...
+        /// <param name="height">Height of the image, in pixels</param>
+        /// <param name="pixelFormat">Desired pixel format</param>
+        /// <returns>A new bitmap with the provided size</returns>
+        public static BitmapImage CreateBitmap(int width, int height, PixelFormat pixelFormat)
+        {
+            VerifyFactoryAvailable();
+            return s_currentFactory!.CreateBitmap(width, height, pixelFormat);
+        }
+
+        private static void VerifyFactoryAvailable()
```
krwq, line 103: if you put `[MemberNotNull(nameof(s_currentFactory))]` on this method then you won't need `s_currentFactory!` everywhere

src/devices/Common/Iot/Device/Graphics/BitmapImage.cs
```diff
@@ -70,5 +147,83 @@ public virtual void Clear(Color color = default)
...
+        /// Return the data pointer as a raw span of bytes
+        /// </summary>
+        /// <returns>A span of bytes</returns>
+        public abstract Span<byte> AsByteSpan();
+
+        /// <summary>
+        /// Return the data pointer as a raw span of integers. For 32bit images, this equals to one pixel per entry.
+        /// </summary>
+        /// <returns>A span of integers</returns>
+        public virtual Span<int> AsIntSpan()
```
krwq, line 181: does this need to be virtual? could it be just regular method? do we actually need it? Would be better if callsite did conversion

src/devices/Common/Iot/Device/Graphics/BitmapImage.cs
```diff
@@ -70,5 +147,83 @@ public virtual void Clear(Color color = default)
...
+
+        /// <summary>
+        /// Saves this bitmap to a file
+        /// </summary>
+        /// <param name="filename">The filename to save it to</param>
+        /// <param name="fileType">The filetype to use.</param>
+        /// <remarks>
+        /// Generally, the method is not checking that the filename extension matches the file type provided.
+        /// </remarks>
+        public virtual void SaveToFile(string filename, ImageFileType fileType)
```
krwq, line 194: same question for this.. does this need to be virtual? can callsite do this instead?

src/devices/Common/Iot/Device/Graphics/BitmapImage.cs
```diff
@@ -70,5 +147,83 @@ public virtual void Clear(Color color = default)
...
+                SaveToStream(fs, fileType);
+            }
+        }
+
+        /// <summary>
+        /// Save the image to a stream
+        /// </summary>
+        /// <param name="stream">The stream to save the data to</param>
+        /// <param name="format">The image format</param>
+        public abstract void SaveToStream(Stream stream, ImageFileType format);
```
krwq, line 207: Consider `WriteToStream` or just `WriteTo` since Stream is an argument it's obvious we're writing to stream

src/devices/Common/Iot/Device/Graphics/Converters.cs
```diff
@@ -0,0 +1,114 @@
...
+        public static unsafe BitmapImage ToBitmapImage(Bitmap bmp)
+        {
+            if (bmp == null)
+            {
+                throw new ArgumentNullException(nameof(bmp));
+            }
+
+            if (Environment.OSVersion.Platform != PlatformID.Win32NT)
+            {
+                throw new PlatformNotSupportedException("This operation is only supported on Windows");
```
krwq, line 40: wasn't system.drawing available on Linux as well (IIRC it needed libgdi+)

src/devices/Common/Iot/Device/Graphics/Converters.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>
+        /// Adjusts the target position and size so that a given image can be copied to a target image without scaling and without further cropping.
+        /// This ensures the destination rectangle, starting at the given point lies within the image.
+        /// </summary>
+        /// <param name="image">The input image size</param>
+        /// <param name="leftTop">[in, out] The top left corner of the input image to show. If at the bottom or right edge of the destination, this will be
+        /// reset so that the right edge of the input image is at the right edge of the destination</param>
+        /// <param name="destination">The destination rectangle. If this is larger than the input image, the size will be cropped</param>
+        public static void AdjustImageDestination(BitmapImage image, ref Point leftTop, ref Rectangle destination)
```
krwq, line 74: since we're always allocating destination and point why aren't they `out`?

src/devices/Common/Iot/Device/Graphics/IGraphics.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Iot.Device.Graphics
+{
+    /// <summary>
+    /// This declarative interface provides the basis for drawing functions using image specific APIs.
+    /// This interface is empty. The implementation is provided via extension methods.
+    /// </summary>
+    public interface IGraphics : IDisposable
```
krwq, line 16: why not just regular class in that case? what's the value of this being interface and having extension methods elsewhere?

src/devices/Common/Iot/Device/Graphics/ImageFileType.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// Image file type list
+    /// </summary>
+    public enum ImageFileType
+    {
+        /// <summary>
+        /// Undefined
+        /// </summary>
+        Undefined,
```
krwq, line 14: IMO we should stick with either `Undefined` or `Unspecified` everywhere

src/devices/Ili9341/Ili9341Imaging.cs
```diff
@@ -6,6 +6,7 @@
 using System.Drawing.Imaging;
 using System.Runtime.InteropServices;
 
+#pragma warning disable CA1416 // Temporarily, will be removed after binding is updated
```
krwq, line 9: please remove the comment and replace with link to new issue (unless it's already in the other PR)

krwq: same comment everywhere where applies

src/devices/Is31fl3731/samples/MatrixRgb/Program.cs
```diff
@@ -80,7 +80,7 @@ void TestPixels(byte r, byte g, byte b)
         foreach (int x in Enumerable.Range(0, 5))
         {
             matrix.Fill(0);  // Clear display
-            Color color = Color.FromRgb(r, g, b);
+            Color color = Color.FromArgb(r, g, b);
```
krwq, line 83: this is confusing since argb refers to 4 values but we're passing 3, why aren't we using FromRgb instead?

krwq: Nvm, looked at the wrong Color class documentation

src/devices/Lp55231/Lp55231.cs
```diff
@@ -56,9 +55,9 @@ public Lp55231(I2cDevice i2cDevice)
 
             _leds = new[]
             {
-                Color.FromRgba(0, 0, 0, byte.MaxValue),
-                Color.FromRgba(0, 0, 0, byte.MaxValue),
-                Color.FromRgba(0, 0, 0, byte.MaxValue)
+                Color.FromArgb(255, 0, 0, 0),
```
krwq, line 58: I think we can just not pass the first value now

src/devices/Media/VideoDevice/samples/MjpegStream/Controllers/Image.cs
```diff
@@ -153,14 +155,18 @@ private async void WriteModifiedBufferBody(object sender, NewImageBufferReadyEve
...
-                g.DrawString(DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss"), new Font("Tahoma", 20), Brushes.White, new PointF(0, 0));
-                using var ms = new MemoryStream();
-                myBitmap.Save(ms, ImageFormat.Jpeg);
-                await HttpContext.Response.BodyWriter.WriteAsync(CreateHeader(e.Length));
-                await HttpContext.Response.BodyWriter.WriteAsync(ms.ToArray());
-                await HttpContext.Response.BodyWriter.WriteAsync(CreateFooter());
+                var myBitmap = BitmapImage.CreateFromStream(stream);
+                var g = myBitmap.GetDrawingApi();
+                g.DrawText(DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss"), "Tahoma", 20, Color.White, new Point(0, 0));
+                using (var ms = new MemoryStream())
```
krwq, line 161: any particular reason for not using non-nesting `using`?

src/devices/Media/VideoDevice/samples/MjpegStream/MjpegStream.csproj
```diff
@@ -7,6 +7,7 @@
   
   <ItemGroup>
     <ProjectReference Include="../../../Media.csproj" />
+    <ProjectReference Include="..\..\..\..\SkiaSharpAdapter\SkiaSharpAdapter.csproj" />
```
krwq, line 10: \/\/\/\/\/\/\/\/

krwq: (fixing line above would be more consistent though)

src/devices/RGBLedMatrix/samples/RGBLedMatrix.sample.csproj
```diff
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <ProjectReference Include="../RGBLedMatrix.csproj" />
+    <ProjectReference Include="..\..\SkiaSharpAdapter\SkiaSharpAdapter.csproj" />
```
krwq, line 11: we should use backslashes on line above

src/devices/SkiaSharpAdapter/SkiaSharpImageFactory.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <inheritdoc />
+        public BitmapImage CreateBitmap(int width, int height, PixelFormat pixelFormat)
+        {
+            return new SkiaSharpBitmap(width, height, pixelFormat);
+        }
+
+        /// <inheritdoc />
+        public BitmapImage CreateFromStream(Stream file)
+        {
+            using var image = SKImage.FromEncodedData(file);
```
krwq, line 28: once this gets disposed presumably bitmap won't?

src/devices/Ws28xx/RawPixelContainer.cs
```diff
@@ -0,0 +1,83 @@
...
+
+using System;
+using System.Drawing;
+
+namespace Iot.Device.Ws28xx
+{
+    /// <summary>
+    /// An abstract class that acts as a data container for device-dependent pixel arrays
+    /// </summary>
+    public abstract class RawPixelContainer
```
krwq, line 12: could this potentially use your bitmap image with correct image format?

src/Iot.Device.Bindings.SkiaSharpAdapter/Directory.Build.props
```diff
@@ -0,0 +1,13 @@
+<Project>
```
krwq, line 1: overall do we need to update samples to mention that package should be used?

src/Iot.Device.Bindings.SkiaSharpAdapter/Directory.Build.props
pgrawehr: Yes, we want this as a separate package (to reduce dependencies for those who do not need any bitmap functionality or want to use another imaging library). 

I'm not sure how the version for the release is actually computed in our setup though (question for @joperezr )

src/devices/Arduino/samples/ApiChecker/RgbLedTest.cs
pgrawehr: Yes, this will be fixed with the next PR.

src/devices/BuildHat/Models/LedMode.cs
pgrawehr: At the moment, it appears to build fine.

src/devices/Ili9341/Ili9341Imaging.cs
pgrawehr: That's part of the other PR.

src/Iot.Device.Bindings.SkiaSharpAdapter/Directory.Build.props
pgrawehr: It's done (or will be done) for all projects requiring the imaging library.

src/Iot.Device.Bindings.SkiaSharpAdapter/Iot.Device.Bindings.SkiaSharpAdapter.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <LangVersion>10</LangVersion>
```
joperezr, line 5: Can we use Latest instead? That would make sure that this is not **pinning down** in the future.

src/devices/Ili9341/samples/Program.cs
```diff
@@ -10,7 +10,7 @@
 using System.Threading.Tasks;
 using Iot.Device.Ft4222;
 using Iot.Device.Ili9341;
-
+#pragma warning disable CA1416 // Temporarily, will be removed after binding is updated
```
joperezr, line 13: Can you either explain what is this for or add a link to an issue to track us actually doing this work?

Ellerbach (approved):
Amazing work! Ship it :-)

src/devices/RGBLedMatrix/RgbLedMatrix.cs
```diff
@@ -386,7 +385,7 @@ public void StartRendering()
         /// <param name="y">Upper left y coordinate on the display</param>
         /// <param name="bitmap">System.Drawing.Bitmap object to draw</param>
         /// <param name="backBuffer">true if want use back buffer, false otherwise</param>
-        public unsafe void DrawBitmap(int x, int y, Bitmap bitmap, bool backBuffer = false)
+        public unsafe void DrawBitmap(int x, int y, BitmapImage bitmap, bool backBuffer = false)
```
raffaeler, line 388: This is probably the most delicate method as it is unsafe.
I guess you tested this enough :)

src/devices/RGBLedMatrix/RgbLedMatrix.cs
```diff
@@ -442,7 +430,7 @@ public unsafe void DrawBitmap(int x, int y, Bitmap bitmap, bool backBuffer = fal
         /// <param name="repGreen">replacement color for the green color</param>
         /// <param name="repBlue">replacement color for the blue color</param>
         /// <param name="backBuffer">true if want use back buffer, false otherwise</param>
-        public unsafe void DrawBitmap(int x, int y, Bitmap bitmap, byte red, byte green, byte blue, byte repRed,
+        public unsafe void DrawBitmap(int x, int y, BitmapImage bitmap, byte red, byte green, byte blue, byte repRed,
```
raffaeler, line 433: Same for this other unsafe method

raffaeler (approved):

krwq (approved):
Please address as you see fit since there is nothing particularly blocking. It would be good to cleanup the library from helper methods.

src/devices/Ili9341/samples/Program.cs
pgrawehr: This will be fixed with #2084, which shall be reviewed and merged just after.

src/devices/RGBLedMatrix/RgbLedMatrix.cs
pgrawehr: Actually, I don't have this hardware to test it, but the change is actually quite minor. And I now removed the `unsafe`, since it's no longer required (because we're using Span<byte>)

src/devices/SkiaSharpAdapter/SkiaSharpImageFactory.cs
pgrawehr: Jup, it wouldn't work otherwise...

src/devices/RGBLedMatrix/RgbLedMatrix.cs
raffaeler: Do you mean you removed the `unsafe` keyword from your next PR?

src/Iot.Device.Bindings.SkiaSharpAdapter/Directory.Build.props
pgrawehr: I think we should match, which is the case now (IIRC). But we should (in a separate PR) maybe improve the way the version number is calculated or defined, so we're sure we get the same number everywhere. That is burried somewhere in the arcade repo I guess, and I don't really understand that stuff.

src/devices/Common/Iot/Device/Graphics/Converters.cs
pgrawehr: We are using their old values (see first lines)

src/devices/Common/Iot/Device/Graphics/IGraphics.cs
pgrawehr: Doesn't really matter whether it's an interface or an empty class type. In either case, it's intended to be a handle only. The reason I'm using extension methods is because it is up to the implementation to provide the set of drawing functions. With this approach, it is possible to add new drawing functions without changing the interface.

src/devices/RGBLedMatrix/RgbLedMatrix.cs
pgrawehr: I comited it already here

PR #2082, opened 2023-05-31
Changed rock pi 4 board pin numbering to match documentation.

Sleepwalk42:
This corrects the board pin numbers used by the RockPi4bPlusDriver to match the numbering in the documentation (see [this product brief](https://dl.radxa.com/rockpi4/docs/hw/rockpi4/radxa_rock4bp_product_brief_Revision_1.1.pdf) for the official documentation. 

The original implementation used pin numbers that were 1 less than what the board document specified. If you wanted GPIO code that would work on both the raspberry pi and the rock 4 (their form factor and GPIO header are almost identical), you'd need to apply a -1 to the pin numbers if you were running on a rock 4 board.

This will break existing code using this driver with board pin numbering. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2082)

### DIFF
--- a/src/devices/Gpio/Drivers/RockPi4BPlusDriver.cs
+++ b/src/devices/Gpio/Drivers/RockPi4BPlusDriver.cs
@@ -13,26 +13,43 @@ namespace Iot.Device.Gpio.Drivers
     /// </remarks>
     public class RockPi4bPlusDriver : Rk3399Driver
     {
-        private static readonly int[] _pinNumberConverter = new int[]
-        {
-            -1, -1, MapPinNumber(2, 'A', 7), -1, MapPinNumber(2, 'B', 0), -1, MapPinNumber(2, 'B', 3),
-            MapPinNumber(4, 'C', 4), -1, MapPinNumber(4, 'C', 3), MapPinNumber(4, 'C', 2), MapPinNumber(4, 'A', 3),
-            MapPinNumber(4, 'C', 6), -1, MapPinNumber(4, 'C', 5), MapPinNumber(4, 'D', 2), -1, MapPinNumber(4, 'D', 4),
-            MapPinNumber(1, 'B', 0), -1, MapPinNumber(1, 'A', 7), MapPinNumber(4, 'D', 5), MapPinNumber(1, 'B', 1),
-            MapPinNumber(1, 'B', 2), -1, -1, MapPinNumber(2, 'A', 0), MapPinNumber(2, 'A', 1), MapPinNumber(2, 'B', 2),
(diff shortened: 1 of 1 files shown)


src/devices/Gpio/Drivers/RockPi4BPlusDriver.cs
```diff
@@ -13,26 +13,43 @@ namespace Iot.Device.Gpio.Drivers
...
         /// <inheritdoc/>
         protected override int PinCount => 40;
 
         /// <inheritdoc/>
         protected override int ConvertPinNumberToLogicalNumberingScheme(int pinNumber)
         {
-            int num = _pinNumberConverter[pinNumber];
-
-            return num != -1 ? num : throw new ArgumentException($"Board (header) pin {pinNumber} is not a GPIO pin on the {GetType().Name} device.", nameof(pinNumber));
+            return pinNumber switch
```
Ellerbach, line 22: Thanks for this improvement bu you still miss -1 which is a valid option. It means "no pin". So you should definitely add it.

src/devices/Gpio/Drivers/RockPi4BPlusDriver.cs
```diff
@@ -13,26 +13,43 @@ namespace Iot.Device.Gpio.Drivers
...
+                35 => MapPinNumber(4, 'A', 5),
+                36 => MapPinNumber(4, 'A', 4),
+                37 => MapPinNumber(4, 'D', 6),
+                38 => MapPinNumber(4, 'A', 6),
+                40 => MapPinNumber(4, 'A', 7),
+                _ => throw new ArgumentException($"Board (header) pin {pinNumber} is not a GPIO pin on the {GetType().Name} device.", nameof(pinNumber))
+            };
         }
     }
 }
```
raffaeler, line 55: FWIW I double checked the 40 pin numbering assignment with the specs and they match.

src/devices/Gpio/Drivers/RockPi4BPlusDriver.cs
Sleepwalk42: Really? That seems inconsistent with the rest of the code in this library. I can't find anywhere else where we treat -1 as a valid pin in a board numbering. Look at the same function in RaspberryPi3LinuxDriver for example.

Ellerbach: You are correct, I got catch by the previous implementation.

Ellerbach (approved):

krwq (approved):

krwq:
Thanks @Sleepwalk42!

PR #2083, opened 2023-05-31
Fix TryDecodeData212_424Kbps check bug in Pn532 driver

GEEKiDoS:
it's a very obvious issue

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2083)

### DIFF
--- a/src/devices/Pn532/Pn532.cs
+++ b/src/devices/Pn532/Pn532.cs
@@ -629,6 +629,9 @@ namespace Iot.Device.Pn532
         {
             try
             {
+                // toDecode[1] is POL_RES, which specifies the packet length.
+                // 20 means the packet contains a system code.
+                // See https://nxp.com/docs/en/user-guide/141520.pdf page 116 for more details.
                 if ((toDecode[1] != 18) && (toDecode[1] != 20))
                 {
                     return null;
--- a/src/devices/Pn532/Pn532.cs
+++ b/src/devices/Pn532/Pn532.cs
@@ -629,7 +629,7 @@ namespace Iot.Device.Pn532
         {
             try
             {
-                if ((toDecode[1] != 18) || (toDecode[1] != 20))
+                if ((toDecode[1] != 18) && (toDecode[1] != 20))
                 {
                     return null;
                 }


GEEKiDoS:
@dotnet-policy-service agree

GEEKiDoS:
> @GEEKiDoS please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

src/devices/Pn532/Pn532.cs
```diff
@@ -629,7 +629,7 @@ private bool SetParameters(ParametersFlags parametersFlags)
         {
             try
             {
-                if ((toDecode[1] != 18) || (toDecode[1] != 20))
+                if ((toDecode[1] != 18) && (toDecode[1] != 20))
```
pgrawehr, line 632: If I'm getting this right, the method never did anything, right? The old condition is always true.

pgrawehr (approved):

src/devices/Pn532/Pn532.cs
GEEKiDoS: > If I'm getting this right, the method never did anything, right? The old condition is always true.

yes

GEEKiDoS:
> Could you please give some context? Is there a related issue to read from the reasons for this change? (I am not familiar with the source code of this device)

```c#
private static readonly uint8_t[] FELICA_POLL_CMD = { 0x00, 0xFF, 0xFF, 0x01, 0x00 };

//  in method
var felicaData = _pn532.ListPassiveTarget(MaxTarget.One, TargetBaudRate.B212kbps, FELICA_POLL_CMD);
if (felicaData != null)
{
    var decrypted  = TryDecodeData212_424Kbps(felicaData.AsSpan()[1..]);
    // this is always null
    if (decrypted is null)
        return 0;
	// ...
}
```

raffaeler:
Got it, butI have no clue what `18` and `20` means in the pn532 context (not easy to understand even with the PDF opened).
Maybe @Ellerbach can provide a hint.

GEEKiDoS:
> Got it, butI have no clue what `18` and `20` means in the pn532 context (not easy to understand even with the PDF opened). Maybe @Ellerbach can provide a hint.

I guess `18` means `TargetBaudRate.B212kbps` type card and `20` for `TargetBaudRate.B424kbps` one.

raffaeler:
I can't find anything related in the specs.
It looks like a "config command" but the numbers do not match.
I leave here the link to the specs: https://www.nxp.com/docs/en/nxp/data-sheets/PN532_C1.pdf

GEEKiDoS:
> I can't find anything related in the specs. It looks like a "config command" but the numbers do not match. I leave here the link to the specs: https://www.nxp.com/docs/en/nxp/data-sheets/PN532_C1.pdf

Ah I got it, see this: https://www.nxp.com/docs/en/user-guide/141520.pdf

In page 116, it says that byte was the length of 212/424 Kbps card polling response, so 20 means response contains system code and 18 does not.

Ellerbach (approved):
Looks good, thanks for the catch

raffaeler:
> Ah I got it, see this: https://www.nxp.com/docs/en/user-guide/141520.pdf
> 
> In page 116, it says that byte was the length of 212/424 Kbps card polling response, so 20 means response contains system code and 18 does not.

Great, thank you.
May I ask you to just add a comment specifying that this byte is called `POL_RES` and specify the packet length (maybe referring to the spec chapter)?

raffaeler (approved):

raffaeler:
Thank you for the comment in the code,

PR #2084, opened 2023-05-31
Example and Binding updates for new image handling

pgrawehr:
Update examples related to image handling. This PR depends on #2069.

Includes new and updated bindings:
- Ili9342 QVGA display (moved namespace to Ili934x, because very similar to Ili9341)
- Ili9341 QVGA display driver updated
- Axp192 Power controller, including device-dependent abstraction for M5Tough
- Fixed analog reporting in Arduino binding for never boards

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2084)


src/devices/Ili934x/samples/images/OpenMenu.png
krwq, line 1: What's the source of these pictures?

src/devices/Common/Iot/Device/Graphics/MouseClickSimulatorUInput.uinput.cs
```diff
@@ -0,0 +1,210 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
krwq, line 1: this feels like it should be in separate PR and is a bit out of scope for this repo

src/devices/Common/Interop/Unix/Interop.uinput.cs
```diff
@@ -0,0 +1,201 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
krwq, line 1: is this generated? why do we need all these?

src/devices/Common/Interop/Unix/Interop.X11.cs
```diff
@@ -0,0 +1,429 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
```
krwq, line 4: same question, why do we need X11 interop added as part of image library refactoring?

src/devices/Axp192/M5ToughPowerControl.cs
```diff
@@ -0,0 +1,228 @@
...
+            }
+        }
+
+        private void SendCpuToSleep()
+        {
+            // Sends the board to sleep mode
+            TimeSpan sleepDelay = TimeSpan.FromSeconds(10);
+            const int wakeupPin = 39; // The screen interrupt is connected to this pin
+
+            if (!(_board is ArduinoBoard ab))
```
krwq, line 183: why does Axp192 need to depend on arduino board? this complicates porting to things like nano or using this library standalone

pgrawehr: It's not the AXP that depends on the board, but the M5Though device. This is a device containing an AXP, an ESP32, a display and several other components, and for this library, Arduino is currently the only way to talk to it. In here, we need to send some special commands to the CPU.

pgrawehr:
Closing, as now integrated in various other PRs.

Issue #1070, opened 2020-05-05
Include the major version when referencing *.so libraries


Tragetaschen:
**Describe the bug**

Linux distributions typically split libraries like libgpiod into multiple packages.

One specific split is between `libgpiod` and `libgpiod-dev`, where the former contains the necessary binaries and symlinks to run programs while the latter is meant for development, where you (traditionally) invoke a linker to produce a binary. The latter typically uses the unversioned symlink to simplify the build and the binary referencing the library then includes the versioned reference. In this split, the full version (`libgpiod.so.2.1.3`) and a symlink with the major version (`libgpiod.so.2`) are in the main package, the unversioned symlink (`libgpiod.so`) is in the `-dev` package, which doesn't get installed per default.

For example for Ubuntu:
- https://packages.ubuntu.com/bionic/amd64/libgpiod1/filelist
- https://packages.ubuntu.com/bionic/amd64/libgpiod-dev/filelist

Embedded distributions like Yocto per default do not install the `-dev` package into the image when you declare a runtime dependency on a library. Including the `-dev` package in the image can have unwanted side-effects of pulling in other development dependencies (for example I ended up with a Python interpreter in my image for a different -dev package in the past)

My own Yocto image doesn't include any `-dev` packages, so in order to use the `LibGpiodDriver` in my code, I have to "translate" the `DllImport` string by
```csharp
static MyType()
    => NativeLibrary.SetDllImportResolver(typeof(GpioDriver).Assembly, libgpiodFixup);

private static IntPtr libgpiodFixup(string libraryName, Assembly assembly, DllImportSearchPath? searchPath)
{
    if (libraryName == "libgpiod")
        if (NativeLibrary.TryLoad("libgpiod.so.2", assembly, searchPath, out var result))
            return result;
    return IntPtr.Zero;
}
```
However, this "trick" only works exactly once per assembly (only one resolver per assembly), and the `NativeLibrary` API is meant to be used within the library requiring a specialized resolver. If System.Device.Gpio would decide to start using that functionality in the future, my own fixup would be broken.

Since the C# interop code typically relies on a specific major ABI version when declaring methods from an external library like libgpiod, that major version should be included in the `DllImport` instead of the unversioned .so.

**Steps to reproduce**

Run code using `LibGpiodDriver` on a Linux image without the libgpiod-dev package.

**Expected behavior**
Run without exception.

**Actual behavior**
`DllNotFoundException`

**Versions used**

Add following information:

[...]

Issue #1992, opened 2022-12-05
SysFsDriver keeps gpio line in-use when application crashes


HakanL:
**Describe the bug**

When using the SysFsDriver and the application crashes then the Gpio line is left in-use and can't be used again with the libgpiod driver.

**Steps to reproduce**

1. Run gpioinfo 0 to verify that line 17 is unused.
2. Download and run this demo app: https://github.com/HakanL/IotGpioDriverError which crashes (throws an exception).
3. Run gpioinfo 0 again to see that line 17 says now that it's in-use by sysfs.

If I try to run the demo app again then that works, but if I switch to use libgpiod instead then I'll get error code 16 (resource in use / EBUSY).

**Expected behavior**

Any gpio resource allocated by the SysFsDriver should be released, even if the application crashes.

**Actual behavior**

Used gpio lines are kept in-use.

**Versions used**
- dotnet 7.0
- System.Device.Gpio: 2.2.0
- Iot.Device.Bindings: 2.2.0

Tracking external issue: https://github.com/dotnet/runtime/issues/79155

pgrawehr:
Thanks for this observation. I don't know whether there's an easy fix for this problem, since it feels awkward having to use the sysfs driver to properly initialize libgpiod after such a crash.

raffaeler:
@HakanL Could you please post the stack trace of the fault when running on libgpiod? I don't have a RPi here now.

HakanL:
> Thanks for this observation. I don't know whether there's an easy fix for this problem, since it feels awkward having to use the sysfs driver to properly initialize libgpiod after such a crash.

Understood, it may not be possible to release the resource in the scenario (in the SysFsDriver). This is the first time I've seen an application on Linux keeping resources after it's killed, which is the reason why I wasn't expecting this behavior (and waited a long time to reboot). I also tested to run the application again with the SysFsDriver, and exiting it properly (after it had crashed before), to see if it would release the line. The application runs, but it doesn't release the line. It is possible of course that it's an issue with the SysFs "system", but that is outside my expertise.

raffaeler:
> Sure, here's the stack trace (but IMHO the issue is in the SysFsDriver, the Gpio line is marked as in-use even when the process is completely gone.

Yes, I got this, but I wanted to understand where exactly the problem happens.

I guess that the SysFsDriver should subscribe some process-termination hook to close the native resource, but I am not familiar with that code.

krwq:
[...]

PR #2090, opened 2023-06-22
Bind to explicit major version of libgpiod

joperezr:
Fixes https://github.com/dotnet/iot/issues/1070

cc: @rafal-mz @Ellerbach @krwq @pgrawehr 

Things that I need help with for testing this:

- [ ] Ensure that libgpiod driver still works as expected when picking it manually
- [ ] Ensure that libgpiod driver is now the default in boards that haven't installed the `libgpiod-dev` package
- [ ] If possible, Ensure that this works in several boards, not just a Raspberry Pi.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2090)

### DIFF
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -15,7 +15,7 @@ internal partial class Interop
 {
     internal partial class libgpiod
     {
-        // Binding to the 2.x mayor version of libgpiod which comes installed on most distros today.
+        // Binding to the 2.x major version of libgpiod which comes installed on most distros today.
         private const string LibgpiodLibrary = "libgpiod.so.2";
         internal static IntPtr InvalidHandleValue = new IntPtr(-1);
 
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -15,7 +15,8 @@ internal partial class Interop
 {
     internal partial class libgpiod
     {
-        private const string LibgpiodLibrary = "libgpiod";
+        // Binding to the 2.x mayor version of libgpiod which comes installed on most distros today.
+        private const string LibgpiodLibrary = "libgpiod.so.2";
         internal static IntPtr InvalidHandleValue = new IntPtr(-1);
 
         /// <summary>


src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -15,7 +15,8 @@ internal partial class Interop
 {
     internal partial class libgpiod
     {
-        private const string LibgpiodLibrary = "libgpiod";
+        // Binding to the 2.x mayor version of libgpiod which comes installed on most distros today.
```
jdbruner, line 18: typo - "major" :-)

joperezr: Great find 😆, I guess my spanish-is-my-first-language side was displayed here 😝

joperezr:
```suggestion
        // Binding to the 2.x major version of libgpiod which comes installed on most distros today.
```

Ellerbach (approved):
Looks good to me

pgrawehr:
I tested the behavior (with the Board sample project). 

- Initially, I had the packages `libgpiod-dev` and `libgpiod2` installed on my Pi4. As expected, the test showed that the libgpiod driver was used for interrupt handling.
- Then I uninstalled both of these packages, and it no longer worked (sysfs was used)
- Then I re-added libgpiod2 and it worked again.

I have not tested whether libgipod2 is installed by default on a recent Raspberry Pi OS image.

joperezr:
Thanks for testing @pgrawehr! Yeah I think the behavior you are seeing is expected, and indeed it is my believe that libgpiod2 comes installed by default in the RPI OS which is the main value that the PR is bringing. If someone would be able to validate that, it would be great and I would consider this ready to go in.

raffaeler:
Test successful on RPi4
- both packages installed: uses libgpiod driver
- libgpiod2 packages only: uses libgpiod driver
- no packages: uses sysfs driver
Note: after running the app with sysfs and re-installing the two packages, the app crashes because of [this](https://github.com/dotnet/iot/issues/1992)

raffaeler:
Test failed on RPi3 with Raspbian version 'stretch'.
### compiled with .NET 6
The package `libgpiod2` is **not** available on 'stretch' therefore it always use sysfs.
On 'stretch' it is only available `libgpiod1` (even when installing `libgpiod-dev`)
Anyway the app did not crash.

### compiled with .NET 7
Same app/binaries of the RPi4 test, compiled as self-contained:
```
Failed to load /home/pi/test/IotButton1/libcoreclr.so, error: /lib/arm-linux-gnueabihf/libm.so.6: version `GLIBC_2.27' not found (required by /home/pi/test/IotButton1/libcoreclr.so)
Segmentation fault
```
This error is new to me

raffaeler:
I have no other hardware boards to test on.

raffaeler (approved):

krwq:
[Triage] We should figure out if we can make it still work with libgpiod 1

raffaeler:
The available Raspbian versions are the following (Bullseye is the  first available at 64 bit)

Version | Codename | Debian release date | RPI OS release date
-- | -- | -- | --
9 | Stretch | June 2017 | August 2017
10 | Buster | July 2019 | July 2019
11 | Bullseye | August 2021 | November 2021

Source: https://raspberrytips.com/raspberry-pi-os-versions/

This is the summary of my tests:
Version | Codename | libgpiod version | Test result
-- | -- | -- | --
9 | Stretch | 1 | failure
10 | Buster | 2 | success
11 | Bullseye | 2  | success

Given that the string "libgpiod.2" is a const, we have few options:
1. bind to an arbitrary name and create a symbolic link (ln utility) on the first run
2. parameterize the library name and dynamic load the library instead of pinvoke
3. support only v1 and v2 duplicating the class (with some compile trick to avoid duplicated code
4. drop support for Stretch

@joperezr I believe that before or later option 3 and 4 will fail again in case of a v3. What do you think?

krwq:
[Triage] This will be merged post 3.0

joperezr:
Sorry, catching up with this thread.

I don't think there is anything wrong with us dropping stretch for newer versions of the packages. People using stretch, can always use packages from 2.x era, but for 3.x stuff we will require higher OS, and I think that is fine as long as it is well documented. Once libgpiod 3.x is available, we can then use bump to use that in our next major version release. I don't think this PR should be merged post 3.0 release, it should either be included in it, or just wait till 4.0 as it is a breaking change. We know it is breaking, and we always have, which is why we wanted to use the major version bump to take this break as we believe the overall benefits would be more than the setbacks.

raffaeler:
> People using stretch, can always use packages from 2.x era

I tried to install libgpiod2 on stretch and the installation (apt i) failed.
I guess the same will happen at the time a libgpiod3 will come out, but this is just a guess.

joperezr:
Right, what I meant here is that people using stretch can still use our System.Device.Gpio and Iot.Device.Bindings 2.x packages, which does support stretch. What I'm saying is that now that we are shipping a new major version, we can decide to drop support for stretch with the libgpiod driver (meaning potentially losing some people there and would just endup using sysfs there) in order to have much more people using libgpiod driver in newer distros (which we know this driver behaves much better than sysfs).

joperezr:
As long as we communicate this correctly, in our release notes and package notes we can say which package is recommended depending on your base Linux distro.

Tragetaschen:
> Given that the string "libgpiod.2" is a const, we have few options:
> 
> 1. bind to an arbitrary name and create a symbolic link (ln utility) on the first run

This can also be done using [NativeLibrary.SetDllImport](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary.setdllimportresolver?view=net-7.0#system-runtime-interopservices-nativelibrary-setdllimportresolver(system-reflection-assembly-system-runtime-interopservices-dllimportresolver)). That is, chose a well known string constant and then in the resolver try to load `libgpiod.so.2` and `libgpiod.so.1` in order to be compatiple with both. This is what I did "from the outside" to load the major-versioned libgpiod on a system without the `-dev` dependency installed.

From the inadvertent tests in #1070, there doesn't seem to be an incompatibility between the used exports between 1 and 2 as both work fine.

raffaeler:
@joperezr I would definitely prefer to support both and not dropping the old OS just because of this.
The proposed solution from @Tragetaschen (which I also used in the past) works great in .NET (core) but [not in netstandard](https://apisof.net/catalog/1a7eace8-7ea8-e6d6-a5fe-a9c90239af11).

joperezr:
My only concerns with that approach would be perf impact, which could be mitigated by making sure we somehow cache the result so we only have to find the right version once, as opposed to having to do it for every single PInvoke. If we are able to do that, then I'm fine with closing this PR and then doing this post 3.0 release (as we wouldn't be doing a breaking change anymore so no need to tie this to a major version bump)

Tragetaschen:
@joperezr The work of loading a native library is O(1) with the constant being 1 ;-)
The libgpiod.so.* is loaded exactly once and so the resolver registered with NativeLibrary is asked exactly once to intercept that load attempt.

krwq:
We'd presumably call SetDllImportResolver in one of the static ctors?

raffaeler:
> We'd presumably call SetDllImportResolver in one of the static ctors?

@krwq We can definitely use `NativeLibrary` for .NET, but we have to do something for netstandard as well.

Tragetaschen:
> We'd presumably call SetDllImportResolver in one of the static ctors?
Or use `ModuleInitializerAttribute`

joperezr:
Thanks for the suggestion @Tragetaschen I think that is a good idea for us to try. In favor of that, I'm closing this PR so that we instead can go with the NativeLibrary load approach. @Tragetaschen given you are familiar to how all of that works, would you be willing to contribute a PR to achieve this 😃?

Tragetaschen:
Under different circumstances, I would gladly go ahead, but I changed jobs a couple of years back and don't have any embedded platform to run and test code on.

joperezr:
Understood, appreciate the guidance and I hope you'd be okay if we tag you on the PR once we have it ready in case you have a chance to do a quick review. 

FWIW, looks like runtime is attempting to do something super similar due to the exact same problem but with a different native library, and they have a PR open to fix this https://github.com/dotnet/runtime/pull/88851.

PR #2091, opened 2023-06-26
[SHT4x] Read serial number of sensor module

gjnijenhuis:
The current module has no implementation for reading the serial number of the SHT4x device. 
This PR implements the reading of the serial number of the SHT4x device.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2091)

### DIFF
--- a/src/devices/Sht4x/Sht4x.cs
+++ b/src/devices/Sht4x/Sht4x.cs
@@ -81,6 +81,25 @@ namespace Iot.Device.Sht4x
         }
 
         /// <summary>
+        /// Reads the serial number of the device
+        /// </summary>
+        /// <returns>
+        /// A 32 bit integer with the serial number.
+        /// If a CRC check failed for a reading, it will be <see langword="null"/>.
+        /// </returns>
+        public int? ReadSerialNumber()
+        {
+            Span<byte> buffer = stackalloc byte[6];
+            _device.WriteByte((byte)0x89);
+            _device.Read(buffer);
+
+            ushort? msb = Sensirion.ReadUInt16BigEndianAndCRC8(buffer);
+            ushort? lsb = Sensirion.ReadUInt16BigEndianAndCRC8(buffer.Slice(3, 3));
+
+            return lsb + (msb << 16);
+        }
+
+        /// <summary>
         /// Reads relative humidity and temperature.
         /// </summary>
         /// <returns>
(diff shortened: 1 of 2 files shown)


gjnijenhuis:
@dotnet-policy-service agree

pgrawehr (approved):

raffaeler (approved):

src/devices/Sht4x/Sht4x.cs
```diff
@@ -80,6 +80,25 @@ public void Reset()
...
+        public int? ReadSerialNumber()
+        {
+            Span<byte> buffer = stackalloc byte[6];
+            _device.WriteByte((byte)0x89);
+            _device.Read(buffer);
+
+            ushort? msb = Sensirion.ReadUInt16BigEndianAndCRC8(buffer);
+            ushort? lsb = Sensirion.ReadUInt16BigEndianAndCRC8(buffer.Slice(3, 3));
+
+            return lsb + (msb << 16);
```
Ellerbach, line 99: why not using version instead of an int?
You definitely get 2 elements:

image

gjnijenhuis: In to the example C code provided by Sensirion the serial number is also output as a 32 bit integer. When the manufacturer of the chip interprets the serial number in this way I think it would be best to be consistent and also use a 32 bit integer here.

https://github.com/Sensirion/embedded-sht/blob/master/sht4x/sht4x.c#L103

gjnijenhuis: Agreed, in this code the serial is just stored in a 32 bit integer, at the pointer which was passed through at function call. There is no example code that uses it.

The Adafruit library has it also implemented in a similar way and has an example on how the serial is used:

Code:
https://github.com/adafruit/Adafruit_SHT4X/blob/master/Adafruit_SHT4x.cpp#L83

Example:
https://github.com/adafruit/Adafruit_SHT4X/blob/master/examples/SHT4test/SHT4test.ino#L26

```
Serial.println("Found SHT4x sensor");
Serial.print("Serial number 0x");
Serial.println(sht4.readSerial(), HEX);
```

Ellerbach: ok, still, there is no check. So I4m fine with this int.

PR #2092, opened 2023-07-02
New Ili9342 device, moved namespace accordingly

pgrawehr:
Ili9431 was moved to Ili943x

Minor update to example to support new API. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2092)


src/devices/Ili934x/tests/DummySpiDriver.cs
```diff
@@ -0,0 +1,50 @@
...
+        private List<byte> _outputData;
+        public DummySpiDriver()
+        {
+            _outputData = new List<byte>();
+            ConnectionSettings = new SpiConnectionSettings(0, 1);
+        }
+
+        public override SpiConnectionSettings ConnectionSettings { get; }
+
+        public List<byte> Data
```
krwq, line 28:
```suggestion
        public List<byte> Data { get; } = new List<byte>();
```

src/devices/Ili934x/Ili934x.sln
```diff
@@ -0,0 +1,143 @@
...
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Axp192", "..\Axp192\Axp192.csproj", "{079BECB2-C2F8-4DC0-B9A2-111EA143A3C5}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Nmea0183", "..\Nmea0183\Nmea0183.csproj", "{5277BF07-40B8-4E65-A777-20ECB2A420CF}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SkiaSharpAdapter", "..\SkiaSharpAdapter\SkiaSharpAdapter.csproj", "{DD919A7C-32DD-4D0B-B7B4-61285E188672}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tests", "tests", "{22FB5850-0D20-4703-80BA-5C902119F31E}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Ili934x.Tests", "tests\Ili934x.Tests.csproj", "{212B3113-99B9-4A53-8210-8C3512379A5F}"
```
krwq, line 24: are all those needed in the ILI project?

src/devices/Ili934x/samples/Ili934x.Samples.csproj
```diff
@@ -0,0 +1,17 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>$(DefaultSampleTfms)</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+  	<ProjectReference Include="../Ili934x.csproj" />
+    <ProjectReference Include="../../Ft4222/Ft4222.csproj" />
+    <ProjectReference Include="../../Arduino/Arduino.csproj" />
```
krwq, line 11: nit: use backslashes everywhere

src/devices/Ili934x/Rgb565.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            get
+            {
+                return _value;
+            }
+            set
+            {
+                _value = value;
+            }
+        }
```
krwq, line 134:
```suggestion
        public ushort PackedValue { get; set; }
```

and remove extra field

src/devices/Ili934x/Rgb565.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            get
+            {
+                int gbyte = ((Swap(_value) & 0x7E0) >> 3);
+                if (gbyte == 0)
+                {
+                    return 0;
+                }
+
+                return gbyte | 0x7;
```
krwq, line 58: should this be 0x3?

src/devices/Ili934x/Rgb565.cs
```diff
@@ -0,0 +1,199 @@
...
+            retval |= (UInt16)(color.G >> 2);
+            // shift right to make room for the red or blue Value
+            retval <<= 5;
+            // combine with the 6 MSB if the red or blue value
+            retval |= (UInt16)(color.B >> 3);
+
+            return new Rgb565(Swap(retval));
+        }
+
+        private static ushort Swap(ushort val) => (ushort)(val >> 8 | val << 8);
```
krwq, line 122: nit: brackets would be useful here

src/devices/Ili934x/Chsc6440.cs
```diff
@@ -0,0 +1,410 @@
...
+
+namespace Iot.Device.Ili934x
+{
+    /// <summary>
+    /// Binding for Chipsemi CHSC6540 capacitive touch screen controller
+    /// Used for instance on the M5Tough in conjunction with an ILI9342 display controller.
+    /// Note: The M5Core2, while being very similar to the M5Tough otherwise, has a FT6336U instead.
+    /// The two chips appear to be similar, but the documentation is poor.
+    /// </summary>
+    public class Chsc6440 : IDisposable
```
krwq, line 22: can the touch screen be added separately?

src/devices/Ili934x/Chsc6440.cs
```diff
@@ -0,0 +1,410 @@
...
+        /// This event is fired when the user "clicks" a position
+        /// Call <see cref="EnableEvents"/> to use event handling
+        /// </summary>
+        public event Action<object, Point>? Touched;
+
+        /// <summary>
+        /// This event is fired repeatedly when the user drags over the screen
+        /// Call <see cref="EnableEvents"/> to use event handling.
+        /// </summary>
+        public event Action<object, DragEventArgs>? Dragging;
```
krwq, line 64: Add some prefix like "On" suggesting this is an event. Dragging suggest the type is `bool`

krwq: (same for other events here)

src/devices/Ili934x/Chsc6440.cs
```diff
@@ -0,0 +1,410 @@
...
+                byte pts = 0;
+                int p0f = 0;
+                Span<byte> register = stackalloc byte[1]
+                {
+                    0x02
+                };
+
+                if (IsPressed())
+                {
+                    Span<byte> data = stackalloc byte[11];
```
krwq, line 206: (non-blocking) could some structure be used here instead? the bit twiddling below looks a bit confusing

src/devices/Ili934x/Chsc6440.cs
```diff
@@ -0,0 +1,410 @@
...
+        public virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                _updateThreadActive = false;
+                var temp = _updateThread;
+                if (temp != null)
+                {
+                    temp.Join();
+                }
```
krwq, line 377:
```suggestion
                _updateThread?.Join();
                _updateThread = null;
```

src/devices/Ili934x/Chsc6440.cs
```diff
@@ -0,0 +1,410 @@
...
+                        _gpioController.Dispose();
+                    }
+                }
+
+                _gpioController = null;
+
+                if (_i2c != null)
+                {
+                    _i2c.Dispose();
+                }
```
krwq, line 396: _i2c?.Dispose();

src/devices/Ili934x/Chsc6440.cs
```diff
@@ -0,0 +1,410 @@
...
+
+                _i2c = null!;
+                _updateEvent.Dispose();
```
krwq, line 399:
```suggestion
                _updateEvent?.Dispose();
                _updateEvent = null;
```

src/devices/Ili934x/Ili9341.cs
```diff
@@ -81,52 +91,63 @@ public Ili9341(SpiDevice spiDevice, int dataCommandPin, int resetPin, int backli
...
-        ///         RRRRRGGG GGGBBBBB
-        ///         43210543 21043210
-        ///
-        /// For ColorSequence.BGR
-        ///         BBBBBGGG GGGRRRRR
-        ///         43210543 21043210
-        /// </returns>
-        private (byte Low, byte High) Color565(Color color)
+        /// <remarks>This is of type int, because all image sizes use int, even though this can never be negative</remarks>
+        public virtual int ScreenWidth
```
krwq, line 112: this could be just `public virtual int ScreenWidth => 240;`

src/devices/Ili934x/Ili9341.cs
```diff
@@ -137,42 +158,59 @@ public Ili9341(SpiDevice spiDevice, int dataCommandPin, int resetPin, int backli
...
-                displayBytes[i * 2 + 1] = colourBytes[1];
+                for (int i = x; i < x + w; i++)
+                {
+                    _screenBuffer[i + j * ScreenWidth] = c;
+                }
             }
 
-            // specify a location for the rows and columns on the display where the data is to be written
-            SetWindow(x, y, x + w - 1, y + h - 1);
+            if (doRefresh)
```
krwq, line 179: I'd suggest to create AutoRefresh property or similar and either do it everywhere or nowhere

src/devices/Ili934x/Ili9341Imaging.cs
```diff
@@ -0,0 +1,167 @@
...
+                int bottomRow = ScreenHeight;
+                int w = ScreenWidth;
+                for (int y = 0; y < ScreenHeight; y++)
+                {
+                    for (int x = 0; x < w; x++)
+                    {
+                        if (!Rgb565.AlmostEqual(_screenBuffer[x + y * w], _previousBuffer[x + y * w], 2))
+                        {
+                            topRow = y;
+                            goto reverse;
```
krwq, line 106: consider not using goto in this method

src/devices/Ili934x/Ili934x.sln
pgrawehr: Not yet, but I hope we get it back

src/devices/Ili934x/Rgb565.cs
pgrawehr: In structs, I prefer to use explicit fields, to clearly see the size of the struct and also the order of the fields, since this is important for many use cases (here as well).

src/devices/Ili934x/Chsc6440.cs
pgrawehr: Sure, but I really wouldn't prolong this even further.

src/devices/Ili934x/Chsc6440.cs
pgrawehr: No, see e.g. https://learn.microsoft.com/en-us/dotnet/api/system.windows.window?view=windowsdesktop-7.0 It's a verb in past tense (`Touched`, `Clicked`, etc. ) for the events, and a method such as `OnTouched`, `OnClicked` for the event _handlers_.

src/devices/Ili934x/Chsc6440.cs
pgrawehr: I agree, but this is just the internal decoding of a binary I2C reply, as we do it in many other bindings as well. This unpacks the bits into structures.

src/devices/Ili934x/Ili9341.cs
pgrawehr: I think that would make the use more complex. ClearScreen is really the only command that should be used with an immediate effect. In all other cases, you need more than one drawing operation before you can publish the new screen content. I'm going to remove the doRefresh parameter from the public FillRectangle method.

src/devices/Ili934x/Ili9341Imaging.cs
pgrawehr: I know this is uncommon, but to jump out of nested loops it's really the clearest thing to do IMHO.

src/devices/Ili934x/Rgb565.cs
pgrawehr: Good catch!

PR #2095, opened 2023-07-04
Fixes #2094 - Ft232H I2C Create Device always returns 'Device already open'

### LINKED ISSUE #2094: Ft232H I2C Create Device always returns 'Device already open'
I'm unable to create an I2C device using the Ft232H device. It always returns 'Device already open'.
Even running the sample code returns this issue.

Looking through the code I found this line of code:

https://github.com/dotnet/iot/blob/30f0875fbc34c4928f894b6fb104337f72083faa/src/devices/Ft232H/Ft232HI2cBus.cs#L40-L43

This checks if the device is already in the used addresses list, and if not in the list it will throw the error. 
This is the wrong way around. The error should be thrown when the device is in the list, not when it is not in the list.
Inverting the check makes the samples run fine again.

gjnijenhuis:
Fixes #2094
Inverted the 'Device Already Open' check so it will work as designed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2095)

EDIT (@krwq): changed syntax so that issue gets auto closed on merge

### DIFF
--- a/src/devices/Ft232H/Ft232HI2cBus.cs
+++ b/src/devices/Ft232H/Ft232HI2cBus.cs
@@ -37,7 +37,7 @@ namespace Iot.Device.Ft232H
         /// <inheritdoc/>
         public override I2cDevice CreateDevice(int deviceAddress)
         {
-            if (!_usedAddresses.ContainsKey(deviceAddress))
+            if (_usedAddresses.ContainsKey(deviceAddress))
             {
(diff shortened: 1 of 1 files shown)


krwq:
Thank you @gjnijenhuis!

PR #2117, opened 2023-08-05
Binding for some common Gui operations

pgrawehr:
Add two virtual devices:
- Take screenshots
- Simulate a mouse

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2117)


Ellerbach:
I have a question: what is different from the excellent https://github.com/gui-cs/Terminal.Gui?
I've been using Terminal.Gui and it does all this, works on all platforms and OS.

pgrawehr:
> I have a question: what is different from the excellent [gui-cs/Terminal.Gui](https://github.com/gui-cs/Terminal.Gui)? I've been using Terminal.Gui and it does all this, works on all platforms and OS.

I didn't know that library, but I don't see it doing any of this (or I couldn't find it). 

My binding is for 
- a) creating a virtual mouse, not to read the mouse. So the purpose is to be able to create an Iot App that takes some sensors (e.g. a joystick, or an accelerometer) and use it to control the desktop OS. 
- b) take a graphical screenshot of the desktop OS that can be forwarded to a Iot screen.

Together, these two allow running a desktop OS on something like an Ili934x (on SPI) and a touchscreen or to create custom input devices for a kiosk application. And with "desktop OS" I really mean the full GUI (of either Windows or Unix)

Ellerbach:
OK, so it's quite complementary and both can be used at the same time. I'm a big fan of Terminal.Gui. Really works nicely. I'm using it a lot in personal projects. So you can get a gui in a command line with support of mouse as well. So integration of both can be nice.

src/devices/Gui/Interop/Interop.uinput.cs
```diff
@@ -0,0 +1,205 @@
...
+        internal const int ABS_TOOL_WIDTH = 0x1c;
+        internal const int ABS_VOLUME = 0x20;
+        internal const int ABS_MISC = 0x28;
+
+        internal const int SYN_REPORT = 0;
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct input_id
+        {
+            public UInt16 bustype;
```
Ellerbach, line 121: similar as a previous question, any reason in this file to use the UInt64, UInt16, etc rather than the ulong, ushort, etc versions?

src/devices/Gui/ScreenCapture.Windows.cs
```diff
@@ -0,0 +1,65 @@
...
+            try
+            {
+                using (Bitmap bitmap = new Bitmap(area.Width, area.Height, System.Drawing.Imaging.PixelFormat.Format32bppArgb))
+                {
+                    using (System.Drawing.Graphics g = System.Drawing.Graphics.FromImage(bitmap))
+                    {
+                        g.CopyFromScreen(new System.Drawing.Point(area.Left, area.Top), System.Drawing.Point.Empty, new System.Drawing.Size(area.Width, area.Height));
+                    }
+
+                    // For some reason, we need to swap R and B here. Strange...
```
Ellerbach, line 31: in most linux system, the colors are stored in BBGGRR and not RRGGBB like in Windows ;-)

Ellerbach (approved):
Few comments but all up, it looks really good and interesting scenario, definitely something that can bring a lot of value!

src/devices/Gui/Interop/Interop.X11.cs
```diff
@@ -0,0 +1,452 @@
...
+namespace Iot.Device.Gui
+{
+    internal partial class InteropGui
+    {
+        internal const string X11 = "libX11.so";
+
+        // Definitions for this file come from X.h and Xlib.h from the Raspberry Pi (32 Bit Raspberry Pi OS)
+        // Some structures may have different alignment on 64 bit systems, so that it may be required to replace
+        // some field types with nint/nuint
+        internal static UInt32 AllPlanes = 0xFFFF_FFFF;
```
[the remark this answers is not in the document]
pgrawehr: For values used in interop calls, I prefer to be explicit about the size. I know that in c#, uint is _defined_ to be 32 bits, but this is not the case in the C world.

src/devices/Gui/Interop/Interop.X11.cs
```diff
@@ -0,0 +1,452 @@
...
+        private static extern unsafe IntPtr XOpenDisplay(byte* displayName);
+
+        /// <summary>
+        /// Opens a named display
+        /// </summary>
+        /// <param name="displayName">Display name, or null for the default display</param>
+        /// <returns>A display handle</returns>
+        internal static unsafe IntPtr XOpenDisplay(string displayName)
+        {
+            var bytes = Encoding.ASCII.GetBytes(displayName);
```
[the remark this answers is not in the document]
pgrawehr: Yes, this can be opened several times.

src/devices/Gui/ScreenCapture.Windows.cs
pgrawehr: Ah, thanks for the explanation. Except that this is the Windows implementation that messes things up, it seems...

src/devices/Gui/IPointingDevice.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// Move the cursor by the given amount
+        /// </summary>
+        /// <param name="x">Move by this amount in x</param>
+        /// <param name="y">Move by this amount in y</param>
+        void MoveBy(int x, int y);
+
+        /// <summary>
+        /// Click once with the given button
+        /// </summary>
+        /// <param name="button">Button to press</param>
```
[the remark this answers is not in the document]
pgrawehr: Actually, the linux implementation currently only supports one button at a time. But I'll fix that, so that the flag value indeed is useful.

src/devices/Gui/ScreenCapture.Windows.cs
Ellerbach: Depending on the tools used, they may also store internally the "reverse" way.

PR #2120, opened 2023-08-17
Bind to explicit major version of libgpiod - Try 2

### LINKED ISSUE #1070: Include the major version when referencing *.so libraries

joperezr:
Fixes #1070 

This is essentially trying to use version 2.0 of libgpiod to remove the need of installing the -dev package, and adds some logic in the case that version 2.0 is not found so that it can instead look for version 1.6. This was inspired by @Tragetaschen [comment on the first attempt PR](https://github.com/dotnet/iot/pull/2090#issuecomment-1632659911), as well as following precedence to a [very similar change that was made in dotnet/runtime](https://github.com/dotnet/runtime/pull/88851) solving a very similar situation.

cc: @pgrawehr @raffaeler @Ellerbach 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2120)

### DIFF
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
@@ -8,15 +8,48 @@
 using System;
 using System.Device.Gpio;
 using System.Runtime.InteropServices;
+#if NET6_0_OR_GREATER
+using System.Runtime.Loader;
+using System.Reflection;
+#endif
 // Since this is only used on Linux, and in C on Linux sizeof(long) == sizeof(void*) this is a valid alias.
 using NativeLong = System.IntPtr;
 
 internal partial class Interop
 {
-    internal partial class libgpiod
+    internal static partial class Libgpiod
     {
+#if NET6_0_OR_GREATER
+        private const string LibgpiodLibrary = "libgpiod.so.2";
+#else
         private const string LibgpiodLibrary = "libgpiod";
-        internal static IntPtr InvalidHandleValue = new IntPtr(-1);
+#endif
+        internal static IntPtr InvalidHandleValue;
+
+        static Libgpiod()
+        {
+            InvalidHandleValue = new IntPtr(-1);
+#if NET6_0_OR_GREATER
+            Assembly currentAssembly = typeof(Libgpiod).Assembly;
+
+            AssemblyLoadContext.GetLoadContext(currentAssembly)!.ResolvingUnmanagedDll += (assembly, libgpiodName) =>
+            {
+                if (assembly != currentAssembly || libgpiodName != LibgpiodLibrary)
+                {
+                    return IntPtr.Zero;
+                }
+
+                // If loading the 2.x libgpiod failed, we may be running in a Linux distribution that has the 1.x
+                // version installed, so we try to load that instead.
+                if (NativeLibrary.TryLoad("libgpiod.so.1", out IntPtr handle))
+                {
+                    return handle;
+                }
+
+                return IntPtr.Zero;
+            };
+#endif
+        }
 
         /// <summary>
         /// Release all resources allocated for the gpiochip iterator and close the most recently opened gpiochip(if any).
--- a/src/System.Device.Gpio/Interop/Unix/SafeChipHandle.cs
+++ b/src/System.Device.Gpio/Interop/Unix/SafeChipHandle.cs
@@ -17,9 +17,9 @@ internal class SafeChipHandle : SafeHandle
 
     protected override bool ReleaseHandle()
     {
-        Interop.libgpiod.gpiod_chip_close(handle);
+        Interop.Libgpiod.gpiod_chip_close(handle);
         return true;
     }
 
-    public override bool IsInvalid => handle == IntPtr.Zero || handle == Interop.libgpiod.InvalidHandleValue;
+    public override bool IsInvalid => handle == IntPtr.Zero || handle == Interop.Libgpiod.InvalidHandleValue;
 }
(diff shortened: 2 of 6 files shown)


raffaeler (approved):
Looks good!
Good to see the uppercase refactoring as well.

src/System.Device.Gpio/Interop/Unix/libgpiod/Interop.libgpiod.cs
```diff
@@ -8,15 +8,48 @@
...
+#else
         private const string LibgpiodLibrary = "libgpiod";
-        internal static IntPtr InvalidHandleValue = new IntPtr(-1);
+#endif
+        internal static IntPtr InvalidHandleValue;
+
+        static Libgpiod()
+        {
+            InvalidHandleValue = new IntPtr(-1);
+#if NET6_0_OR_GREATER
```
Ellerbach, line 32: just to double check; with any other version, we will have the existing behavior, right? In that case, looks all good to me!

raffaeler: The following line is what always existed in current versions:
```
        private const string LibgpiodLibrary = "libgpiod";
```

This line of code will now used only for `netstandard` (which we basically use only for Mono on Linux).

pgrawehr (approved):

krwq:
nice job figuring this out!

PR #2121, opened 2023-08-23
Fixes the button holding test randomly failing on Mac

raffaeler:
This PR aim to fix the holding button test which randomly fails on Mac.
Since the holding state is measured with a `System.Threading.Timer`, it probably failed because the Timer thread was not scheduled before the assert made on the main (test) thread.

This change does the following:
- Parameterize the TestButton to use shorter debounce and holding times
- Measure the actual time elapsed when the holding event happens (if ever)

### DIFF
--- a/src/devices/Button/tests/ButtonTests.cs
+++ b/src/devices/Button/tests/ButtonTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading;
+using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


Ellerbach (approved):
Looks good! Thanks for fixing it!

raffaeler:
There may be other tests having the same issue.
But IMHO is better to see how this goes and then I'll proceed with the others.

PR #2125, opened 2023-08-29
Added CM4 model check to IsPi4

Daniel-Foley:
Fixes https://github.com/dotnet/iot/issues/2126
Compute Module 4 needs to be handled the same way as a pi4. Right now, due to the device model of a CM4 being "Raspberry Pi Compute Module 4" it is being treated as a pi3(only in code that uses the IsPi4 bool). This causes a problem with the pin mapping and prevents the CM4 from having it's input/output and pullup/pulldown changed on OpenPin().

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2125)


Ellerbach (approved):
Looks good to me!

raffaeler (approved):

PR #2124, opened 2023-08-27
Update RaspberryBoardInfo.cs for pi400 Rev 1.1

stevetms:
pi400 Rev 1.1 Revision is 0x3131

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2124)

PR #2127, opened 2023-08-30
Fixes the button holding test randomly failing on Mac (part 2)

raffaeler:
This is the continuation of #2121 

The main point is that tests failed on mac just because the button holding uses a `System.Threading.Timer` and this secondary timer is served very rarely on Mac.
The latest fail was instead due to the *timeout too short*. This is a timeout to only avoid the test can hang forever.

These changes consists in:
- Making the timeout very long because the already fixed test failed because of a timeout (https://github.com/dotnet/iot/pull/2124#issuecomment-1699463438)
- Fixing the only other test that asserts Holding as `true`
- No need to fix the tests asserting Holding as `false`
- Shortening the times used by the tests to avoid wasting time during the CI process

> Please note there is no open issue for this problem

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2127)

### DIFF
--- a/src/devices/Button/tests/ButtonTests.cs
+++ b/src/devices/Button/tests/ButtonTests.cs
@@ -50,7 +50,7 @@ namespace Iot.Device.Button.Tests
         }
 
         [Fact]
-        public async void If_Button_Is_Held_Holding_Event_Fires()
+        public async Task If_Button_Is_Held_Holding_Event_Fires()
         {
             bool pressed = false;
             bool holding = false;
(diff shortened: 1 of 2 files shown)


Ellerbach:
The question now is why do the tests fails on Linux now? We had this issue on Mac debug but never elsewhere.

raffaeler:
> The question now is why do the tests fails on Linux now? We had this issue on Mac debug but never elsewhere.

The failures that you are seeing now are due to a change I made to see if they could solve the issue on the Mac.
The problem is related to the delta between the end of the timer and the holding event.
I am not sure why the holding event was designed to have a started and completed state. The started happens when the timer expires and the complete when the button is released.
On the mac the complete never happens so I wanted to see if at least I could see the started event. With this change, the tests on Windows passes, but not on Linux.

joperezr:
Looks like the test hasn't been failing for a while so closing this for now. If we start seeing Mac builds failing again then we can revisit this.

Issue #2138, opened 2023-09-28
suggest make this method protected


HumJ0218:
https://github.com/dotnet/iot/blob/c80f9988b8ad95e96a096543caa15fbb3c58f0a2/src/devices/Is31fl3731/Is31fl3731.cs#L190

in order to facilitate the subclasses in implementing a one-time update of all pixel states.

krwq:
Are you planning to add new device binding? If so feel free to change that in that PR if it's needed.

PR #2144, opened 2023-10-07
add WriteAllPixels to Is31fl3731

HumJ0218:
in order to facilitate the subclasses in implementing a one-time update of all pixel states. #2138

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2144)

### DIFF
--- a/src/devices/Is31fl3731/Is31fl3731.cs
+++ b/src/devices/Is31fl3731/Is31fl3731.cs
@@ -235,8 +235,8 @@ namespace Iot.Device.Display
 
         private void WriteLedPwm(ReadOnlySpan<byte> value)
         {
-            int num = FrameRegister.Pwm + GetLedAddress(0, 0);
-            Write(0, (byte)num, value);
+            int address = FrameRegister.Pwm + GetLedAddress(0, 0);
+            Write(0, (byte)address, value);
         }
 
         private void WriteLed(int x, int y, bool enable)
--- a/src/devices/Is31fl3731/Is31fl3731.cs
+++ b/src/devices/Is31fl3731/Is31fl3731.cs
@@ -100,6 +100,28 @@ namespace Iot.Device.Display
         }
 
         /// <summary>
+        /// Set value for all LEDs.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/devices/Is31fl3731/Is31fl3731.cs
```diff
@@ -99,6 +99,28 @@ public void WritePixel(int x, int y, int brightness, bool enable, bool blink)
             WriteLedPwm(x, y, brightness);
         }
 
+        /// <summary>
+        /// Set value for all LEDs.
+        /// </summary>
+        public void WriteAllPixels(int[,] brightness)
```
pgrawehr, line 105: Any reason you use `int` here? If this was `byte[,]`, you could even convert it to a span without an actual copy.

HumJ0218: Consistent with `int this[int x, int y]`
Should this and similar members be changed to byte?

HumJ0218: In addition, the pixel arrangement of each specific LED board is different, and it is necessary to correspond the pixel coordinates to the actual data address. So the xy coordinate in the original code does not use byte, but int, probably for this reason

HumJ0218: I'm also wondering why the previous code used int instead of byte as the brightness value. This change may involve code changes for multiple other classes, and a separate PR may be required to make this change.

As for direct copying, I don't think it's feasible. Because the pixel layout of the actual LED board may not exactly correspond to the data address, a GetAddress method is required to obtain the actual address, resulting in no direct copy

pgrawehr: Ok, that's a different issue then. Let's leave it as-is for now.

pgrawehr (approved):

Ellerbach (approved):
Looks good to me. Thanks for the addition.

PR #2145, opened 2023-10-08
Update Board info for Pi 5

pjgpetecodes:
The Board info for the new Raspberry Pi 5 is missing.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2145)

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
@@ -144,7 +144,7 @@ public class RaspberryPi3Driver : GpioDriver
             RaspberryBoardInfo.Model.RaspberryPi5 or
             RaspberryBoardInfo.Model.RaspberryPi400 => new RaspberryPi3LinuxDriver(),
             RaspberryBoardInfo.Model.RaspberryPiComputeModule4 or
-            RaspberryBoardInfo.Model.RaspberryPiComputeModule3 => new RaspberryPiCm3Driver(),                        
+            RaspberryBoardInfo.Model.RaspberryPiComputeModule3 => new RaspberryPiCm3Driver(),
             _ => null,
         };
     }
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
@@ -141,9 +141,10 @@ public class RaspberryPi3Driver : GpioDriver
             RaspberryBoardInfo.Model.RaspberryPiZeroW or
             RaspberryBoardInfo.Model.RaspberryPiZero2W or
             RaspberryBoardInfo.Model.RaspberryPi4 or
+            RaspberryBoardInfo.Model.RaspberryPi5 or
             RaspberryBoardInfo.Model.RaspberryPi400 => new RaspberryPi3LinuxDriver(),
             RaspberryBoardInfo.Model.RaspberryPiComputeModule4 or
-            RaspberryBoardInfo.Model.RaspberryPiComputeModule3 => new RaspberryPiCm3Driver(),
+            RaspberryBoardInfo.Model.RaspberryPiComputeModule3 => new RaspberryPiCm3Driver(),                        
             _ => null,
         };
     }
--- a/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
@@ -8,7 +8,7 @@ using System.Threading.Tasks;
 namespace System.Device.Gpio.Drivers;
 
 /// <summary>
-/// A GPIO driver for the Raspberry Pi 3 or 4, running Raspbian (or, with some limitations, ubuntu)
+/// A GPIO driver for the Raspberry Pi 3 or 4, running Raspbian or Raspberry Pi OS (or, with some limitations, ubuntu)
 /// </summary>
 public class RaspberryPi3Driver : GpioDriver
 {
@@ -141,7 +141,6 @@ public class RaspberryPi3Driver : GpioDriver
             RaspberryBoardInfo.Model.RaspberryPiZeroW or
             RaspberryBoardInfo.Model.RaspberryPiZero2W or
             RaspberryBoardInfo.Model.RaspberryPi4 or
-            RaspberryBoardInfo.Model.RaspberryPi5 or
             RaspberryBoardInfo.Model.RaspberryPi400 => new RaspberryPi3LinuxDriver(),
             RaspberryBoardInfo.Model.RaspberryPiComputeModule4 or
             RaspberryBoardInfo.Model.RaspberryPiComputeModule3 => new RaspberryPiCm3Driver(),
--- a/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
@@ -471,11 +471,11 @@ public class GpioController : IDisposable
     private static GpioDriver GetBestDriverForBoardOnLinux()
     {
 
-        boardInfo = RaspberryBoardInfo.LoadBoardInfo();
+        var boardInfo = RaspberryBoardInfo.LoadBoardInfo();
 
-        boardInfo.BoardModel switch
+        switch (boardInfo.BoardModel)
         {
-            RaspberryBoardInfo.Model.RaspberryPi3B or
+            case RaspberryBoardInfo.Model.RaspberryPi3B or
             RaspberryBoardInfo.Model.RaspberryPi3APlus or
             RaspberryBoardInfo.Model.RaspberryPi3BPlus or
             RaspberryBoardInfo.Model.RaspberryPiZeroW or
@@ -489,32 +489,22 @@ public class GpioController : IDisposable
 
                 if (internalDriver is object)
                 {
-                    console.WriteLine($"Using internal RaspberryPi3LinuxDriver");
                     return new RaspberryPi3Driver(internalDriver);
                 }
 
-                console.WriteLine($"Using UnixDriver");
                 return UnixDriver.Create();
-                break;
-
+                
             case RaspberryBoardInfo.Model.RaspberryPi5:
 
-                //
                 // For now, for Raspberry Pi 5, we'll use the LibGpiodDriver.
-                //
                 // We need to create a new driver for the Raspberry Pi 5, 
                 // because the Raspberry Pi 5 uses an entirely different GPIO controller (RP1)
-                //
-                console.WriteLine($"Using LibGpiodDriver");
                 return new LibGpiodDriver(4);
-                break;
-            
-            case else:
+                
+            default:
 
-                console.WriteLine($"Using UnixDriver");
                 return UnixDriver.Create();
-                break;
-
+                
         };       
     }
 
--- a/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
@@ -25,6 +25,7 @@ public class GpioController : IDisposable
     private const string BaseBoardProductRegistryValue = @"SYSTEM\HardwareConfig\Current\BaseBoardProduct";
     private const string RaspberryPi2Product = "Raspberry Pi 2";
     private const string RaspberryPi3Product = "Raspberry Pi 3";
+    private const string RaspberryPi5Product = "Raspberry Pi 5";
 
     private const string HummingBoardProduct = "HummingBoard-Edge";
     private const string HummingBoardHardware = @"Freescale i.MX6 Quad/DualLite (Device Tree)";
@@ -504,6 +505,11 @@ public class GpioController : IDisposable
             baseBoardProduct == RaspberryPi2Product || baseBoardProduct.StartsWith($"{RaspberryPi2Product} "))
         {
             return new RaspberryPi3Driver();
+        }
+
+        if (baseBoardProduct == RaspberryPi5Product || baseBoardProduct.StartsWith($"{RaspberryPi5Product} "))
+        {
+            return new LibGpiodDriver();
         }
 
         if (baseBoardProduct == HummingBoardProduct || baseBoardProduct.StartsWith($"{HummingBoardProduct} "))
--- a/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
@@ -474,15 +474,15 @@ public class GpioController : IDisposable
 
         switch (boardInfo.BoardModel)
         {
-            case RaspberryBoardInfo.Model.RaspberryPi3B
-                or RaspberryBoardInfo.Model.RaspberryPi3APlus
-                or RaspberryBoardInfo.Model.RaspberryPi3BPlus
-                or RaspberryBoardInfo.Model.RaspberryPiZeroW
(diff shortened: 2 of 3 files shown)


src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -141,9 +141,10 @@ RaspberryBoardInfo.Model.RaspberryPi3BPlus or
             RaspberryBoardInfo.Model.RaspberryPiZeroW or
             RaspberryBoardInfo.Model.RaspberryPiZero2W or
             RaspberryBoardInfo.Model.RaspberryPi4 or
+            RaspberryBoardInfo.Model.RaspberryPi5 or
             RaspberryBoardInfo.Model.RaspberryPi400 => new RaspberryPi3LinuxDriver(),
```
pgrawehr, line 145: The GPIO Hardware of the RPi5 seems to be quite different from previous versions. Is the RaspberryPi3LinuxDriver really working correctly?

pjgpetecodes: Fair point... I'll see if I can build it all locally and try it on my Pi5 and see what's what. :-)

raffaeler: I read the pre-release datasheet for the new custom IC called `RP1` which is new to the RPi5.
It governs all the I/O, serial ports, I2C, SPI, PWM, video and USB communications. The HW architecture is very different from the previous boards even if they apparently maintained the external pin compatibility (They also use MUXes for the I/O, meaning that you can remap the ping assignments).

The main point about that IC is to offload the communications from the main processor to support low-power communication while the rest of the board is in sleep mode. It also supports 1.8V levels but they have to be carefully used because, once set any voltage higher than 1.8V can damage the board.

Said that, the libgpiod could be enough to decouple the new functionalities but since there are more features, it will probably deserve a new separate `RasperryPi5LinuxDriver` to support them.

I don't have a new RPi5, but any test you could do is more than welcome :)

pjgpetecodes: Is there a recommended workflow for development / debugging for GPIO? Are we able to debug directly on the Pi? @pgrawehr @raffaeler ...

raffaeler: > Is there a recommended workflow for development / debugging for GPIO? Are we able to debug directly on the Pi? 

The point here is that there is no direct access for the C# code to the GPIO, therefore debugging the GPIO access does not help much.  The 'dotnet/iot` library provides two "drivers": `sysfs` (GPIOs mapped on files) and `libgpiod`. They are just the C# code accessing the file system or the `libgpiod` api.

- `sysfs` is the simpler (and slower) way. If you can successfully access the GPIOs using the default tools, it should work from `dotnet/iot` as well. Even if this works, it is not the best solution as file mapped GPIOs have a number of issues.
- `libgpiod` provides access to the GPIOs via APIs. I understood that there is no current version for `libgpiod` for the RPi5. In this case there is nothing you can do from dotnet. A soon as the Raspberry Foundation should provide it, we will have to see if the APIs are exactly the same or not.

Alternate solutions are:
- fork `libgpiod` and provide support for the `RPi5`, you should download and implement the `RP1` specifications inside the `libgpiod` source code. This step is definitely not trivial.
- implement the `RP1` specs directly in C# bypassing libgpiod. It is doable but, again, not trivial.
AFAIK we never did this for any other board.

P.S. As I mentioned in the previous message the RP1 specs have a number of new features that may not fit in the current `libgpiod` API. This is why I believe we need something different on the C# wrapper side as well.

pjgpetecodes: Gotcha... Not trivial is definitely a takeaway here, especially given that I've not been involved in the code at all here.

Having a quick play...

## `sysfs`

 I can see that `sysfs` looks like it's able to access the GPIO... 

image

However, it doesn't actually manage to change the physical state of pins... Of course, this is deprecated in favour of `libgpiod`...

## `pinctrl`

`pinctrl` seems to work fine;

image

With the pin happily setting and clearing.

## `gpiod`

Installing gpiod and running gpiodetect I get;

image

gpioinfo for pinctrl-rp1 gives;

image

And I'm able to toggle GPIO15 with gpioset;

image

## Python `GPIOZero`

As a side note... Python GPIOZero is functioning correctly.

(I think there's a blob post waiting to happen here!)

raffaeler: Great work!
From your tests, it looks like `libgpiod` is already working, which is great.
I also suspect that `sysfs` writes did not work for some other reasons. Once the GPIO is mapped, read or write should no make any difference, unless you don't have the appropriate permissions. Try as `root`.

At this point you can make the test you initially planned. Try creating the `GpioController` explicitly passing the `LibGpiodDriver` instance and see if you can read/write the GPIOs.

pjgpetecodes: Cool beans and thanks!

This side of things is all quite new to me to be fair, not played with GPIO at this level before., very interesting indeed.

I'll have a play and see what I can do!

I'm guessing I'll have to clone, switch to my working branch, and build this on the pi, then manually add the system.device.gpio dlls to my project to begin with...?

pgrawehr: @pjgpetecodes There are two options to work with the library:
##Variant 1
- Check out the repo on the PC, build
- Create / update an example
- Publish it for arm64 and copy to the pi
- Run on the Pi, use SSH for remote debugging, if needed

##Variant 2
- Checkout on the PI
- Build on the Pi
- Execute the examples / Integration tests

I'm normally going with 1, unless I'm running the GPIO testbench. Then I check out on the Pi and run `dotnet test System.Device.Gpio.sln`. (For the GPIO tests to work, you need to have pins 5 and 6 connected with a resistor)

raffaeler: > I'm guessing I'll have to clone, switch to my working branch, and build this on the pi, then manually add the system.device.gpio dlls to my project to begin with...?

You can just cross-compiling everything on Windows and then copy to the RPi5. Your test code could be any basic binding. I suggest you to directly reference the dotnet/iot libraries so that you recompile the libraries as well. 

This is the typical way to cross-compile using self-contained to avoid any dependency and being able to sync the sources as suggested by Patrick.
- Release vs Debug: pick Debug if you want to avoid optimizations and sync the sources
- linx-arm vs linux-arm64. This depends on the OS on the RPi

```
dotnet deploy -c Release -r linux-arm --self-contained=true
``` 
Please note that the pdbs will point to sources that you just compiled.

In any case, if you just want to make a test, you don't need to debug. You can just deploy the release version as self-contained and see if if works (you can write on the console some message to have detailed info).

HTH

pjgpetecodes: Arghhh... My comments above were stuck in pending... Apologies @raffaeler and @pgrawehr ... Thanks for the help mind you!

pgrawehr: Hey, that's good news, at least we have one driver that works for the Pi5. Can you (just for reference) do the same test with using the RaspberryPi3Driver and the SysFsDriver?

PinNumberingScheme.Board is only supported with the RaspberryPi3Driver, because it needs to know the mapping from the logical pins to the physical pins. Because most boards and drivers don't support that either, we've mostly given up support for that anyway and stick to the logical numbering.

pjgpetecodes: So... 

With the original code;

- `RaspberryPi3Driver` driver gives `Not a supported Raspberry Pi type: ` (Fairly obviously)`
- `SysFSDriver` just works!

With Raspberry Pi 5 added in the `boardInfo.BoardModel` switch in the `CreateInternalRaspberryPi3LinuxDriver` function, creating the `RaspberryPi3Driver` and passing it into the GpioController gives `Error 13 Initializing the Gpio driver`

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3Driver.cs
```diff
@@ -141,6 +141,7 @@ RaspberryBoardInfo.Model.RaspberryPi3BPlus or
             RaspberryBoardInfo.Model.RaspberryPiZeroW or
             RaspberryBoardInfo.Model.RaspberryPiZero2W or
             RaspberryBoardInfo.Model.RaspberryPi4 or
+            RaspberryBoardInfo.Model.RaspberryPi5 or
```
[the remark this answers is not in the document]
pjgpetecodes: Thanks... Done... I've also added the Pi5 Board reference to the GpioController class along with a switch choosing the LibGpiodDriver when on windows.

However, I guess this will only be used if running on Windows 10 or the now (pretty much) defunct Windows 10 IoT?

Do let me know if you think this should be removed though?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -506,6 +507,11 @@ private static GpioDriver GetBestDriverForBoardOnWindows()
             return new RaspberryPi3Driver();
         }
 
+        if (baseBoardProduct == RaspberryPi5Product || baseBoardProduct.StartsWith($"{RaspberryPi5Product} "))
+        {
+            return new LibGpiodDriver();
```
pgrawehr, line 512: Shouldn't this be in the Linux method? AFAIK there's no supported windows version for rpi at this time and libgpiod is not for windows.

pgrawehr: For now, let's keep this simple. As you've already noticed above, Windows isn't really supported any more (the defunct Windows 10 Iot is only supported on the Raspberry Pi 3, and for Windows 11, there are no officially supported versions for any Pi at all at this time). 

So for now, move this code to `GetBestDriverForLinux()` so at least we get a valid driver there. If I understand correctly, you need to do `new LibGpiodDriver(4)` though. After this, I would say this PR is ready to get merged.

After that, investigation is needed onto the next steps. We probably need a fully new `RaspberryPi5Driver` then, that handles the GPIO device for the new RP1 chip. Don't forget to take a look at the `RaspberryPiBoard` class, which contains the logic to program the mux for the different GPIO alternate functions. This logic will likely also change.

pjgpetecodes: Ok done... 

(Ignoring the brain farts of commits - We can squash)... It works...



Of course, if you use a Pin Numbering scheme of Board instead of Logical it throws a Platform Not Supported error (I like the board numbering scheme personally).

pjgpetecodes: I tried PWM and a Servo out of interest and it doesn't seem to work interestingly... I wonder is this a limitation of the LibGpioDriver?

raffaeler: > I tried PWM and a Servo out of interest and it doesn't seem to work interestingly... I wonder is this a limitation of the LibGpioDriver?

What kind of issue is exhibiting? Any specific error?

pjgpetecodes: No errors... Just no PWM out of the pin oddly...

raffaeler: Quite strange. Could you please check the other pins with a scope to see if the issue is only a mapping problem?

pjgpetecodes: Yup... I'll also try python to double check it's not some other fault!

pjgpetecodes: Yup... I'll also try python to double check it's not some other fault!

pgrawehr: I'm assuming you have added the corresponding lines to `boot/config.txt` (I'm expecting that to still be the same).

pjgpetecodes: I've enabled pwm in the Raspberry Pi config tool... (Same thing I think)

pjgpetecodes: With Python and `gpiozero` I get jittery servo movement from BCM18... But there's movement...

raffaeler: > With Python and `gpiozero` I get jittery servo movement from BCM18... But there's movement...

Could you re-try with C# after running the Python code?

pjgpetecodes: Nothing sadly... gpiozero, while jittery, works... I sure it's bit bashed PWM in gpiozero by default given that its jitterry and gives this warning;

`/usr/lib/python3/dist-packages/gpiozero/output_devices.py:1509: PWMSoftwareFallback: To reduce servo jitter, use the pigpio pin factory.See https://gpiozero.readthedocs.io/en/stable/api_output.html#servo for more info
  warnings.warn(PWMSoftwareFallback(`

No movement at all for .NET... Which will be trying to use Hardware PWM.

With that in mind... When I try software PWM from the bindings it works... But not ideal...

```
using SoftwarePwmChannel pwmChannel1 = new SoftwarePwmChannel(18, 50, 0.5, true);
            using ServoMotor servoMotor1 = new ServoMotor(
                pwmChannel1,
                180,
                900,
                2100);

```

raffaeler: There is probably some new initialization stuff needed to make it work on the RPi5 but it's hard to say.

pjgpetecodes: I'll have a dig around and a play.... See what I can turn up...

pjgpetecodes: BTW... I tried adding the following to the config.txt file and rebooting;

```
dtoverlay=pwm-2chan,pin=18,func=2,pin2=13,func2=4
```

But, no dice.

Worth trying eh.

pgrawehr (approved):

pjgpetecodes:
Looks like this is stuck for some reason... Not sure why...

raffaeler (approved):

raffaeler:
> Looks like this is stuck for some reason... Not sure why...

No worries, we just have to review the changes this week during the triage.

pjgpetecodes:
> > Looks like this is stuck for some reason... Not sure why...
> 
> No worries, we just have to review the changes this week during the triage.

Ahhh... Gotcha... Makes sense... Thanks.

I thought I'd done something wrong!

Interestingly it looks like something failed with the build though?

pgrawehr:
@pjgpetecodes No, it's all fine with the build now. Because you're a first-time contributor, the build doesn't start automatically, though. Therefore I had to force it to run.

raffaeler:
> Interestingly it looks like something failed with the build though?

Yes, a couple of builds failed because there are random (known) issues happening in the pipelines that makes some tests fail. The issues on the pipeline will be hopefully fixed soon.
Anyway, I forced re-running the failed platforms and now they are all green.

pjgpetecodes:
Wahoo... Awesome... Thanks!

In the meanwhile... I'm doing some learning on how the current driver works and how to create a new one for the Pi 5!

PR #2146, opened 2023-10-12
Allow creation of duplicate device addresses for the Tca9548A.

### LINKED ISSUE #2141: creating multiple devices at the same address with the tca9548a
**Describe the bug**
Using the tca9548a multiplexer class, I am unable to create two or more devices on different channels at the same address at the same time.  If I attempt to do so without first removing the previously added device, I receive an argument exception. 

This implies that in order for me to be able to use the tca9548a, I must create and remove the device every time i need to use it before any other code creates another device at the same address on a different channel.  Is that the intent of the tca9548a - if so, then this might not be a bug and might be intended behavior.

My understanding is that the purpose of the tca9548a multiplexer is to all us to have devices on different channels at the same address so that we can avoid addressing collisions.  AFAIK, we should be able to switch between channels to interact with the different devices.  Example, I should be able to have an airflow sensor at address 0x61 on channel0 and an mcp4725 dac on channel1 at address 0x61.  Unless I am misinterpreting the intended usage of the tca9548a class, I would not expect that I would be required to remove one device from channel0 before adding the other device to channel1.

I would expect that I could simply create deviceA at address 0x61 on channel0 then create deviceB at address 0x61 on channel1 without exception.  I would expect that before I use deviceA, I switch to channel0 and before I use deviceB I switch to channel1.  It is my understanding this is how the tca9548a is intended to work.

**Steps to reproduce**
Using a C# console app, targeting linux-arm, and rpi4b, standard raspberry pi os, with the I2cBus enabled and Bus 1 configured:
- Create a new I2cBus for BusId 1
- Create a new I2cDevice for the multiplexer at address 0x70
- Create a New Tca9548A passing in the bus and mux device
- Create an I2cDevice at address 0x61 on channel0
- Create an I2CDevice at address 0x61 on channel1
- Expect to receive an System.ArgumentExcetpin: Device with address 0x97 is already open from System.Device.I2c.UnixI2cBus

```csharp
var muxAddress = 0x70;
var dacAddress = 0x61;
var airflowSensorAddress = 0x61;
var i2cBusId = 1;

var bus = I2cBus.Create(i2cBusId);

var muxDevice = I2cDevice.Create(new I2cConnectionSettings(1, muxAddress));
var multiplexer = new Tca9548A(muxDevice, bus, true);
multiplexer[0].CreateDevice(airflowSensorAddress);
multiplexer[1].CreateDevice(dacAddress);
```

**Expected behavior**

I expect a device to be created on channel0 at address 0x61 and another device on channel1 at address 0x61 and that I can call `multiplexer.SelectChannel(MultiplexerChannel.Channel0);` or 
`multiplexer.SelectChannel(MultiplexerChannel.Channel1`); in order to work with deviceA or deviceB without having to remove deviceA or deviceB after each usage

**Actual behavior**

An argument exception is thrown from UnixI2cBus
```
Unhandled exception. System.ArgumentException: Device with address 0x97 is already open. (Parameter 'deviceAddress')
   at System.Device.I2c.UnixI2cBus.<>c.<CreateDevice>b__12_1(Int32 addr, I2cDevice dev)
   at System.Collections.Concurrent.ConcurrentDictionary`2.AddOrUpdate(TKey key, Func`2 addValueFactory, Func`3 updateValueFactory)
   at System.Device.I2c.UnixI2cBus.CreateDevice(Int32 deviceAddress)
   at Iot.Device.Tca954x.Tca9548AChannelBus.CreateDevice(Int32 deviceAddress)
   at Program.<Main>$(String[] args) in C:\source\bvc\iot-samples\dac-sample\Program.cs:line 25
Aborted

```

**Versions used**
```
.NET SDK:
 Version:   7.0.101
 Commit:    bb24aafa11

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.101\

Host:
  Version:      7.0.1
  Architecture: x64
  Commit:       97203d38ba

.NET SDKs installed:
  7.0.101 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.12 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.16 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.1 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 6.0.12 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.16 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 7.0.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```
- target machine is linux-arm rpi4b running raspberry os with a self contained, single file app
- Version of `System.Device.Gpio` 3.0.0
- Version of `Iot.Device.Bindings` 3.0.0
-->

pgrawehr:
Fixes #2141 

This should fix the use case where multiple devices with the same address are used with a TCA9548A. That's one of the main use cases for this device, after all. I have no such device, so the code is only tested using unit testing. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2146)

### DIFF
--- a/src/System.Device.Gpio/System/Device/I2c/I2cConnectionSettings.cs
+++ b/src/System.Device.Gpio/System/Device/I2c/I2cConnectionSettings.cs
@@ -6,7 +6,7 @@ namespace System.Device.I2c;
 /// <summary>
 /// The connection settings of a device on an I2C bus.
 /// </summary>
-public sealed class I2cConnectionSettings
+public sealed class I2cConnectionSettings : IEquatable<I2cConnectionSettings>
 {
     private I2cConnectionSettings()
     {
@@ -38,4 +38,46 @@ public sealed class I2cConnectionSettings
     /// The bus address of the I2C device.
     /// </summary>
     public int DeviceAddress { get; }
+
+    /// <summary>
+    /// Equality comparator
+    /// </summary>
+    /// <param name="other">The other instance</param>
+    /// <returns>True or false</returns>
+    public bool Equals(I2cConnectionSettings? other)
+    {
+        if (ReferenceEquals(null, other))
+        {
+            return false;
+        }
+
+        if (ReferenceEquals(this, other))
+        {
+            return true;
+        }
+
+        return BusId == other.BusId && DeviceAddress == other.DeviceAddress;
+    }
+
+    /// <summary>
+    /// Basic equality overload
+    /// </summary>
+    /// <param name="obj">The other instance</param>
+    /// <returns>True on value equality, false otherwise</returns>
+    public override bool Equals(object? obj)
+    {
+        return ReferenceEquals(this, obj) || (obj is I2cConnectionSettings other && Equals(other));
+    }
+
+    /// <summary>
+    /// Standard hash function
+    /// </summary>
+    /// <returns>A hash code</returns>
+    public override int GetHashCode()
+    {
+        unchecked
+        {
+            return (BusId * 397) ^ DeviceAddress;
+        }
+    }
 }
--- a/src/devices/Tca954x/Tca9548A.cs
+++ b/src/devices/Tca954x/Tca9548A.cs
@@ -81,6 +81,20 @@ namespace Iot.Device.Tca954x
         /// </remarks>
         public void SelectChannel(MultiplexerChannel multiplexChannels)
         {
+            if (_activeChannels.HasValue)
+            {
+                MultiplexerChannel selectedChannel = _activeChannels.Value;
+                if (selectedChannel == multiplexChannels)
+                {
+                    return;
+                }
+                else
+                {
+                    _i2CDevice.WriteByte(Convert.ToByte(multiplexChannels));
+                    _activeChannels = multiplexChannels;
+                }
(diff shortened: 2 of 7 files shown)


boondoggler74 (approved):
looks good to me and works as expected when integration testing locally

Issue #2147, opened 2023-10-13
GPIO driver should add batch write and read function


HumJ0218:
Although `GpioController` provides `Write(ReadOnlySpan<PinValuePair> pinValuePairs)` for batch writes, the actual implementation is still written one by one.

This may prevent some GPIO controllers from operating multiple IO ports at the same time, resulting in the actual single write of the corresponding ports still changing one by one, resulting in inefficient communication and hardware performance that does not match expectations.

Is it possible to add true bulk read-write methods to GpioController and GpioDriver to improve communication efficiency where possible?

pgrawehr:
As some drivers already support that, I guess this should be quite easy to do. Just move `protected virtual void Write(ReadOnlySpan<PinValuePair> pinValuePairs)` to `GpioDriver` and call that directly from the GpioController. Then it can be overwritten.

PR #2148, opened 2023-10-15
Add batch read-write methods for GPIO

HumJ0218:
for #2147 

The original `public void Write(ReadOnlySpan<PinValuePair> pinValuePairs)` and `public void Read(Span<PinValuePair> pinValuePairs)` in `GpioController` were moved to `GpioDriver`, and were made overrideable to support a more efficient batch read-write method implementation in specific GPIO driver classes.

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
@@ -132,6 +132,31 @@ public abstract class GpioDriver : IDisposable
     protected internal abstract void RemoveCallbackForPinValueChangedEvent(int pinNumber, PinChangeEventHandler callback);
 
     /// <summary>
+    /// Write the given pins with the given values.
+    /// </summary>
+    /// <param name="pinValuePairs">The pin/value pairs to write.</param>
+    protected internal virtual void Write(ReadOnlySpan<PinValuePair> pinValuePairs)
+    {
+        for (int i = 0; i < pinValuePairs.Length; i++)
+        {
+            Write(pinValuePairs[i].PinNumber, pinValuePairs[i].PinValue);
+        }
+    }
+
+    /// <summary>
+    /// Read the given pins with the given pin numbers.
+    /// </summary>
+    /// <param name="pinValuePairs">The pin/value pairs to read.</param>
+    protected internal virtual void Read(Span<PinValuePair> pinValuePairs)
+    {
+        for (int i = 0; i < pinValuePairs.Length; i++)
+        {
+            int pin = pinValuePairs[i].PinNumber;
+            pinValuePairs[i] = new PinValuePair(pin, Read(pin));
+        }
+    }
+
+    /// <summary>
     /// Disposes this instance, closing all open pins
     /// </summary>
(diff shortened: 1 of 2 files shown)


src/System.Device.Gpio/System/Device/Gpio/GpioDriver.cs
```diff
@@ -131,6 +131,31 @@ protected internal virtual ValueTask<WaitForEventResult> WaitForEventAsync(int p
     /// <param name="callback">Delegate that defines the structure for callbacks when a pin value changed event occurs.</param>
     protected internal abstract void RemoveCallbackForPinValueChangedEvent(int pinNumber, PinChangeEventHandler callback);
 
+    /// <summary>
+    /// Write the given pins with the given values.
+    /// </summary>
+    /// <param name="pinValuePairs">The pin/value pairs to write.</param>
+    public virtual void Write(ReadOnlySpan<PinValuePair> pinValuePairs)
```
pgrawehr, line 138: GpioDriver methods should be `protected internal`, as we don't want anybody to call these except trough the GPIO controller.

pgrawehr (changes requested):
Looks good from an implementation point of view, but this has broken the test suite. Could get a bit nasty to fix, because `Span<T>` cannot be part of a method that is mocked.

pgrawehr (approved):
Looks good to me. We'll discuss it in the next team meeting.

Ellerbach (commented):
Looks good. Once merge, I can definitely get some perf improvements for the FT family as well.

pgrawehr:
Helix tests are still failing. 

The baseline comparison tests can be ignored for now, that's easy to clarify and fix.

HumJ0218:
I'm a bit puzzled about this compilation error.

Why would the "internal protected" of the base class be considered as "public" in the compilation process, causing the compilation error CP0019?

pgrawehr:
@HumJ0218  Can you show exactly what change causes this? Also, are you sure you meant C**P**0019? I cannot find this error code. What exactly is reported?

HumJ0218:
> @HumJ0218 Can you show exactly what change causes this? Also, are you sure you meant C**P**0019? I cannot find this error code. What exactly is reported?

https://dev.azure.com/dotnet/IoT/_build/results?buildId=99483&view=logs&j=fa59fe4e-195c-56fa-189b-58fd241f10dd&t=afa1ce5e-d2bc-56f3-ad78-7883f25a3e3a&l=874

pgrawehr:
Oh, that. That's just our compatibility check. It verifies that the build doesn't break backwards compatibility against the previous release. Since changing a method from public to protected is a breaking change, it complains. In this case, I think we're fine with accepting that breaking change and you can update the configuration accordingly. Run the following commands to automatically update the suppressions for this warning: 

```
build -pack /p:GenerateCompatibilitySuppressionFile=true
```
It should update the file `src\Iot.Device.Bindings\CompatibilitySuppressions.xml`.

Ellerbach:
[Triage] @HumJ0218 is the use case still valid? If yes, can you please address the conflict and adjust the different elements? If no, we will close the PR which can anyway always been reopened.

PR #2158, opened 2023-11-05
Add support for a few new messages, fix warnings from AisManager

### LINKED ISSUE #2143: AIS Manager is not properly calculating safety information
The AisManager is not properly taking the own position into account to calculate the CPA and TCPA values (closest point of approach and time to closest point of approach). Also, it uses a wrong time for extrapolating the vessel positions.

This has the impact that warnings about close encounters are not properly generated and distances between vessels are calculated incorrectly, often giving unreasonably large CPA values. 

A fix is being worked on.

pgrawehr:
Fixes #2143 

- Added support for a few new NMEA messages (namely the SeaSmart messages for Engine Detail, Engine Fast update and Fluid Level). These messages directly wrap NMEA2000 messages, but are understood by several converters. 
- Fixed AisManager proximity warnings. CPA and TCPA values of nearby vessels should now be correct
- Fixed calculation of magnetic deviation table - do not remove data due to unclear constraints

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2158)


src/devices/Nmea0183/Sentences/FluidData.cs
```diff
@@ -0,0 +1,35 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using UnitsNet;
+
+#pragma warning disable CS1591
```
Ellerbach, line 11: Why deactivating the intelissence comments?

Ellerbach (approved):
Great work again on all those navigation elements.

src/devices/Nmea0183/Sentences/FluidData.cs
pgrawehr: Because I'm lazy... No, I sometimes add this line when I create a class so that I don't have to write the documentation before the final layout of the class is set. Of course, I should have removed this...

PR #2168, opened 2023-11-17
Vcnl4040

fehdem:
Added new device binding for VCNL4040 incl. samples, tests and documentation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2168)


src/devices/Vcnl4040/README.md
```diff
@@ -0,0 +1,111 @@
+# VCNL4040
```
Ellerbach, line 1:
```suggestion
# VCNL4040 - Proximity and Ambient Light Sensor

```

src/devices/Vcnl4040/README.md
```diff
@@ -0,0 +1,111 @@
...
+The VCNL4040 is a proximity and ambient light sensor device designed for various applications that require accurate measurement of proximity and light levels.
+
+## Key features of the sensor
```
Ellerbach, line 4: Markdown 101: all titles needs to be separated with a new line before and after (except the first one as no need before). You'll have to adjust almost all the page to follow the proper style.

src/devices/Vcnl4040/README.md
```diff
@@ -0,0 +1,111 @@
...
+* [Datasheet](https://www.vishay.com/docs/84274/vcnl4040.pdf)
+* [Application note](https://www.vishay.com/docs/84307/designingvcnl4040.pdf)
+
+## Additional information and hints
+For a comprehensive description of the VCNL4040, please refer to the datasheet and, in particular, the application note. The following contains only some additional information about features and behaviors that may not be immediately apparent.
+
+Consider also the documentation of the API classes for further information.
+* Information on the configuration of the ambient light sensor can also be found in the documentation of the ```AmbientLightSensor``` and ```AmbientLightInterruptConfiguration``` classes.
+* Information on the configuration of the proximity sensor can also be found in the documentation of the ```ProximitySensor```, ```EmitterConfiguration```, ```ReceiverConfiguration``` and ```ProximityInterruptConfiguration``` classes.
+
```
Ellerbach, line 22: Markdown 101: only 1 new line before paragraphs, no need to have 2.

```suggestion
```

src/devices/Vcnl4040/README.md
```diff
@@ -0,0 +1,111 @@
...
+   - Changing the integration time lead to adjustments in resolution and range.
+
+* **Threshold Configuration:**
+   - Configuring threshold levels involves considering the current resolution for accurate threshold values.
+   - Thresholds are set in counts, which are influenced by the current resolution setting.
+
+The illuminance measurement is configured by selecting an appropriate integration time. A shorter integration time increases the measurement range but reduces resolution. Moreover, the integration time determines the minimum cycle time for consecutive measurements. The table below illustrates the relationship between integration time, resolution, and measurement range.
+
+|Integration time [ms]|Resolution [lux]|Detection range [lux]|
+|-|-|-|
```
Ellerbach, line 37: Markdown 101: you need at least 3 separators for a table.

```suggestion
|---|---|---|
```

src/devices/Vcnl4040/README.md
```diff
@@ -0,0 +1,111 @@
...
+|80|0.1|6335.5|
+|160|0.05|3276.7|
+|320|0.025|1638.3|
+|640|0.0125|819.1|
+
+#### Interrupts
```
Ellerbach, line 43: overall, can you add couple of lines of code to illustrate how to properly use each of those features?

src/devices/Vcnl4040/README.md
```diff
@@ -0,0 +1,111 @@
...
+#### Board
+![Board](./samples/test_setup_board.png)
+
+## Build and publish samples
+```
+dotnet publish --output ./output --self-contained --runtime linux-arm --framework net6.0 -p:PublishSingleFile=true
+```
+
+Notes:
+* The framework version may change in future. Check with your actual environment.
```
Ellerbach, line 110: Markdown 101: for enums, you need a new line before and after.

```suggestion

* The framework version may change in future. Check with your actual environment.
```

src/devices/Vcnl4040/AmbientLightSensor.cs
```diff
@@ -0,0 +1,359 @@
...
+        /// </summary>
+        public Illuminance ResolutionAsIlluminance => GetDetectionRangeAndResolution(IntegrationTime).Resolution;
+        #endregion
+
+        #region Interrupt
+
+        /// <summary>
+        /// Gets whether interrupts (INT-pin function) are enabled.
+        /// </summary>
+        public bool InterruptsEnabled
```
Ellerbach, line 249: Maybe better with `AreInterruptsEnabled`?

src/devices/Vcnl4040/Definitions/PowerState.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Vcnl4040.Definitions
+{
+    /// <summary>
+    /// Defines the set of ALS and SD power states.
+    /// Documentation: datasheet (Rev. 1.7, 04-Nov-2020 9 Document Number: 84274).
+    /// </summary>
+    public enum PowerState : byte
```
Ellerbach, line 10: why do you want to make this one public? You've been implementing a function to hide it with PowerOn?

src/devices/Vcnl4040/Internal/AlsConfRegister.cs
```diff
@@ -0,0 +1,124 @@
...
+        private bool _alsSdChanged = false;
+        private PowerState _alsSd = PowerState.PowerOff;
+        private AlsInterrupt _alsIntEn = AlsInterrupt.Disabled;
+        private AlsInterruptPersistence _alsPers = AlsInterruptPersistence.Persistence1;
+        private AlsIntegrationTime _alsIt = AlsIntegrationTime.Time80ms;
+
+        /// <summary>
+        /// Gets or sets the ALS integration time setting (ALS_IT).
+        /// </summary>
+        public AlsIntegrationTime AlsIt
```
Ellerbach, line 32: We always prefer a proper naming for public properties (even for internal classes). You can call is AlsIntegrationTime, same for the following ones, use directly the name of the enum, at least it's a bit easier to follow and understand what it means.

src/devices/Vcnl4040/Internal/PsConf1Register.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                _psDuty = value;
+                _pPsDutyChanged = true;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the interrupt persistence (PS_PERS).
+        /// </summary>
+        public PsInterruptPersistence PsPers
```
Ellerbach, line 45: for those as well, will be better with more descriptive naming.

src/devices/Vcnl4040/Internal/PsConf2Register.cs
```diff
@@ -0,0 +1,86 @@
...
+        private const byte PsIntMask = 0b0000_0011;
+        private bool _psHdChanged = false;
+        private bool _psIntChanged = false;
+        private PsOutputRange _psHd = PsOutputRange.Bits12;
+        private PsInterruptMode _psInt = PsInterruptMode.Disabled;
+
+        /// <summary>
+        /// Gets or sets the output range (PS_HD).
+        /// </summary>
+        public PsOutputRange PsHd
```
Ellerbach, line 26: here as well, I would go for a longer naming.

src/devices/Vcnl4040/Internal/PsConf3Register.cs
```diff
@@ -0,0 +1,143 @@
...
+        private PsMultiPulse _psMps = PsMultiPulse.Pulse1;
+        private PsSmartPersistenceState _psSmartPers = PsSmartPersistenceState.Disabled;
+        private PsActiveForceMode _psAf = PsActiveForceMode.Disabled;
+        private PsActiveForceModeTrigger _psTrig = PsActiveForceModeTrigger.NoTrigger;
+        private PsSunlightCancellationState _psScEn = PsSunlightCancellationState.Disabled;
+
+        /// <summary>
+        /// Gets or sets the multi pulse setting (PS_MPS).
+        /// </summary>
+        public PsMultiPulse PsMps
```
Ellerbach, line 35: I won't point all those but again, I feel it's better with a longer naming.

src/devices/Vcnl4040/Vcnl4040Device.cs
```diff
@@ -0,0 +1,135 @@
...
+{
+    /// <summary>
+    /// Represents a VCNL4040 device
+    /// </summary>
+    public class Vcnl4040Device : IDisposable
+    {
+        /// <summary>
+        /// I2C bus address
+        /// </summary>
+        public static int DefaultI2cAddress = 0x60;
```
Ellerbach, line 20: As it's a default address and it's not supposed to change, why not using a const?

Ellerbach (commented):
All up, looks really good. Few comments.

src/devices/Vcnl4040/Internal/AlsConfRegister.cs
fehdem: I would like to ask you to reconsider this. Here's why:

The register classes represent the low-level internals of the device. They encapsulate technical details. Regular users of the binding will never come into contact with these details, as all functions are available through a high-level API. That's why the register classes are marked as "internal." For 99.9% of all users, these classes are irrelevant.

However, if someone does use, extend, or debug the binding at the register level, in my opinion, it is much better for understanding if a direct connection can be made between the code and the datasheet. Personally, I don't want to take the detour through a datasheet but rather see it directly in the code (debugger). This seems much clearer to me than a (supposedly) descriptive name for a register or bit group.

Therefore I really would like to stick with the actual names of the things.

Ellerbach: If those are the names in the datasheet, as it's internal, I'm fine.

src/devices/Vcnl4040/README.md
Ellerbach: Well, most people will look at the few lines of code with the explanation to understand. So maybe just the simple example you have added to show that indeed it's simple for basic usage.

fehdem: Ok, agreed. I'll add an example section to the README.md.

src/devices/Vcnl4040/Definitions/PowerState.cs
fehdem: You're right. That's a mistake. This will be internal.

src/devices/Vcnl4040/AmbientLightSensor.cs
fehdem: Ok.

src/devices/Vcnl4040/Internal/PsConf1Register.cs
fehdem: We'll keep it as is, as discussed for AlsConfRegister.

src/devices/Vcnl4040/Vcnl4040Device.cs
fehdem: Agreed.

src/devices/Vcnl4040/samples/Explorer/.gitignore
```diff
@@ -0,0 +1 @@
+output
```
pgrawehr, line 1: Why is this required? Test output should be below the `bin` folder, not in the source folder.

src/devices/Vcnl4040/samples/Simple/.gitignore
```diff
@@ -0,0 +1 @@
+output
```
pgrawehr, line 1: Same here: Sample output should be below the `bin` folder, so that no extra ignore is necessary.

src/devices/Vcnl4040/samples/test_setup_schematic.png
pgrawehr, line 1: The wiring has a small bug: You should connect VCC to the 3.3V output of the Pi. The 3.3V pin is the _output_ of the internal voltage regulator of the VCNL4040 breakout board. Additionally, C1 has no value.

pgrawehr (commented):
Very nice and very complete binding, thank you! I've added a few questions and improvement suggestions.

src/devices/Vcnl4040/Vcnl4040Device.cs
```diff
@@ -0,0 +1,135 @@
...
+            new InterruptFlagRegister(_i2cDevice).Read();
+        }
+
+        /// <summary>
+        /// Verifies whether a functional I2C connection to the device exists and checks the identification for
+        /// device recognition. An exception is raised, if the communication doesn't work or the identification is incorrect.
+        /// </summary>
+        /// <exception cref="IOException">Non-functional I2C communication.</exception>
+        /// <exception cref="IncompatibleDeviceException">Incompatible device detected.</exception>
+        public void VerifyDevice()
```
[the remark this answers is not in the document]
fehdem: I agree. Actually I thought putting anything into the context of the constructor that goes beyond the local setup of the class, particularly if it takes some time to complete, would be considered a bad practice in the IOT project. 
So, I'm happy to change this.

src/devices/Vcnl4040/Vcnl4040Device.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        private const int CompatibleDeviceId = 0x0186;
+        private readonly InterruptFlagRegister _interruptFlagRegister;
+        private readonly IdRegister _idRegister;
+        private I2cDevice _i2cDevice;
+
+        /// <summary>
+        /// Gets the ambient light sensor of the VCNL4040 device.
+        /// </summary>
+        public AmbientLightSensor AmbientLightSensor { get; private init; }
```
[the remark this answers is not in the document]
fehdem: Agreed. That was just a mistake.

src/devices/Vcnl4040/IncompatibleDeviceException.cs
```diff
@@ -0,0 +1,21 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Iot.Device.Vcnl4040
+{
+    /// <summary>
+    /// Exception that indicates that an incompatible device was found.
+    /// </summary>
+    public class IncompatibleDeviceException : Exception
```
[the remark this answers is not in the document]
fehdem: Agreed.

src/devices/Vcnl4040/ProximityInterruptConfiguration.cs
```diff
@@ -0,0 +1,39 @@
...
+/// (object approaching from the proximity zone), or in both cases.
+/// In these instances, the corresponding flags are set.
+/// Alternatively, the Logic Output mode can be selected. In this mode, the INT pin is automatically controlled.
+/// The pin is set (INT = L) when an object approaches above the upper threshold.
+/// It is reset (INT = H) when the object moves away above the lower threshold.
+/// In this mode, the interrupt flags are NOT set. Detection in the host software is not possible.
+/// </param>
+public record ProximityInterruptConfiguration(
+    ushort LowerThreshold,
+    ushort UpperThreshold,
```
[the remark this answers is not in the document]
fehdem: Unfortunately there is no physical unit like (milli-)meters or so for the sensor readings. It's just "counts". The reason is that it measures the distance (proximity) basing on the amount of light reflected from the surface of the object. The higher the amount of light received the closer the object. But, every surface has different characteristics (and ambient light would play a role, too). So its impossible to derive any physical distance from the pure amount of counts.

src/devices/Vcnl4040/ProximitySensor.cs
```diff
@@ -0,0 +1,346 @@
...
+        #endregion
+
+        #region Measurement
+
+        /// <summary>
+        /// Gets the current proximity sensor reading.
+        /// Note: if active force mode is enabled reading this property implicitly triggers
+        /// the measurement for once cycle.
+        /// </summary>
+        public int Reading
```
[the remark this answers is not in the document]
fehdem: This isn't more expensive than any other property that has access through I2C. The reading is always available, even if the actual measurement process an the device takes a significant time. So it would stick to the property. But, I agree, that "Distance" or "Proximity" would be a more precise name for it.

src/devices/Vcnl4040/samples/Explorer/.gitignore
fehdem: This is an artifact of my own deployment script. I'll remove it.

src/devices/Vcnl4040/samples/Simple/.gitignore
fehdem: This is an artifact of my own deployment script. I'll remove it.

src/devices/Vcnl4040/samples/test_setup_schematic.png
fehdem: Thank you. You're right. I made this mistake during the transfer to Fritzing.

src/devices/Vcnl4040/Internal/AlsDataRegister.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Iot.Device.Vcnl4040.Internal
+{
+    /// <summary>
+    /// ALS output data register
+    /// Note: this is a read-only register.
+    /// Command code / address: 0x09 (LSB, MSB)
+    /// Documentation: datasheet (Rev. 1.7, 04-Nov-2020 9 Document Number: 84274).
+    /// </summary>
+    internal class AlsDataRegister : DataRegister
```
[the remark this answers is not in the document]
fehdem: I really want to stick with these classes. They provide a clear layer between the high level, rather functional interface, and the device level. This "pattern" separates between the device level and the higher level. I always model the device close to "as it is" and assemble from the high(er) level functions.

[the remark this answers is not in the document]
fehdem: Ok, I really got you wrong. Thanks for clarification. Abstraction is exactly what I intended and also consistency across all registers.

Ellerbach:
@fehdem looking at the build, you may have to adjust couple of visibility, some I feel I asked you to change :-) Once done, the build should be green and all looks good to me.

pgrawehr:
> @fehdem looking at the build, you may have to adjust couple of visibility, some I feel I asked you to change :-) Once done, the build should be green and all looks good to me.

I suppose there's some issue with the <InternalsVisibleTo> setting ignored somewhere.

fehdem:
> @fehdem looking at the build, you may have to adjust couple of visibility, some I feel I asked you to change :-) Once done, the build should be green and all looks good to me.



> > @fehdem looking at the build, you may have to adjust couple of visibility, some I feel I asked you to change :-) Once done, the build should be green and all looks good to me.
> 
> I suppose there's some issue with the `InternalsVisibleTo` setting ignored somewhere.

Unfortunately I missed re-exeucting the tests before committing, after a changed the visibility of enum PowerState from public to internal. However, I don't get why it isn't working with the enum, but with other classes of the Vncl4040 assembly. 
Previously all enums have been public. But, as Laurent @Ellerbach correctly pointed out some could be internal as they do not appear in the public API.

fehdem:
Even more strange to me: the Enum CommandCode has an internal visibility. It is next to PowerState in same assembly and even same namespace. CommandCode is used in the tests assembly as well, hence, it should cause the same issue, but it doesn't.

fehdem:
Ok, I think it's pretty clear. Some of the test methods have the enum PowerState as a parameter (for InlineData of Theory). 
I see two options to solve this:
a) keep the few enums, which do not show up in the public binding API, as public instead of internal
b) change the tests, so that the internal enums would be only used within the test methods. This would require introducing "proxy enums" for the InlineData and a mapping to the actual enum inside the test method.

Any other idea?

fehdem:
@Ellerbach @pgrawehr , shall I keep the visibilities as they're or rework the tests? From my perspective, the former is not disadvantageous for the clarity of the binding interface from the user's point of view, as other enums are also public (because they are also used in the public API). A rework of the tests is also conceivable but would require introducing alternative test-specific enums that I use as parameters of the InlineData of the theory and then map to the actual enums in the test method. Feasible but relatively complex and not elegant.

fehdem:
@Ellerbach @pgrawehr : disregard my previous comment. Adapting the public test methods without the internal enums is straightforward. I'll rework the tests and change all enums, not utilized by the public interface, to internal visibility.

pgrawehr:
Very good. You still have some markdown issues, though. To easily detect those locally, build using the following command (this will also run the markdown checks)

```
build -pack
```
(Windows only)

fehdem:
> Very good. You still have some markdown issues, though. To easily detect those locally, build using the following command (this will also run the markdown checks)
> 
> ```
> build -pack
> ```
> 
> (Windows only)

I ran the command, but there is nothing in the output that gives me a hint. 
I guess I found one potential issue and fixed it with the last push.
Really, I wish I could simply fix the issue by knowing the specific markdown rules for this project.

pgrawehr:
I'm sorry, I mixed up something. The correct commands to run the markdown linter locally are (from the root directory):

```
npm install -g markdownlint-cli
markdownlint -c  .markdownlint.json .
```

I've commited my changes, so it should now work.

pgrawehr (approved):

fehdem:
@pgrawehr Thank you! The hint with the markdown lint tool is very useful! I'm working on another binding and I will run it before creating a pull request.

Ellerbach (approved):
Thanks again for this high quality PR. It looks good for me. You can also directly install markdownlint on your system and install extensions on VS Code for example. Markdown like any other language have some rule as well ;-)

PR #2180, opened 2023-11-25
Adding virtual GpioController and virtual GpioPin

Ellerbach:
Adding virtual GpioController and virtual GpioPin. This allows to recreate a GpioController from different GpioPin created form potentially very different GpioController.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2180)


src/System.Device.Gpio/System/Device/Gpio/GpioPin.cs
```diff
@@ -17,6 +17,14 @@ public class Gpio​Pin
             _pinNumber = pinNumber;
         }
 
+        /// <summary>
+        /// Only for compatibility reason and heritage. Do not use in normal usage.
+        /// </summary>
+        protected GpioPin()
```
pgrawehr, line 23: If we have this as protected, shouldn't the other ctor also be protected? Apparently subclassing this class seems to have some use cases.

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,318 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="VirtualGpioController"/> class.
+        /// </summary>
+        /// <param name="numberingScheme">The numbering scheme used to represent pins provided by the controller.</param>
+        public VirtualGpioController(PinNumberingScheme numberingScheme)
+        {
+            // Nothing on purpose, as only logical is suported
```
pgrawehr, line 39: Maybe throw if not logical?

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="VirtualGpioController"/> class.
+        /// </summary>
+        /// <param name="numberingScheme">The numbering scheme used to represent pins provided by the controller.</param>
+        /// <param name="driver">The driver that manages all of the pin operations for the controller.</param>
+        public VirtualGpioController(PinNumberingScheme numberingScheme, GpioDriver driver)
+        {
+            throw new NotImplementedException();
```
pgrawehr, line 49: Can't we just delete this? A ctor that always throws is quite pointless...

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,318 @@
...
+                throw new InvalidOperationException($"Can not set a mode to pin {pinNumber} because it is not open.");
+            }
+
+            return _pins[pinNumber].GetPinMode();
+        }
+
+        /// <inheritdoc/>
+        public override bool IsPinModeSupported(int pinNumber, PinMode mode)
+        {
+            return _pins[pinNumber].IsPinModeSupported(mode);
```
pgrawehr, line 141: Could add a helper method to throw a specifc exception if pinNumber is out of range (here and elsewhere)

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+        /// <summary>
+        /// Checks if a specific pin is open.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        /// <returns>The status if the pin is open or closed.</returns>
+        public new bool IsPinOpen(int pinNumber)
```
pgrawehr, line 149: Better make the base method virtual. Otherwise the behavior could be undefined if using the base class.

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,318 @@
...
+        {
+            return _openPins.ContainsKey(pinNumber);
+        }
+
+        /// <summary>
+        /// Opens a pin and sets it to a specific mode.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        /// <param name="mode">The mode to be set.</param>
+        public new GpioPin OpenPin(int pinNumber, PinMode mode)
```
pgrawehr, line 159: You _need_ to make these virtual, or the whole thing will not work when the base `GpioController` is inserted into a binding. You could probably also get it to work by only overriding `OpenPinCore`, which is already intended to be overriden.

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,318 @@
...
+            _openPins.TryAdd(pinNumber, new VirtualGpioPin(_pins[pinNumber], pinNumber));
+            return _openPins[pinNumber];
+        }
+
+        /// <inheritdoc/>
+        public override ComponentInformation QueryComponentInformation()
+        {
+            ComponentInformation self = new ComponentInformation(this, "Virtual GPIO Controller");
+
+            // PinCount is not added on purpose, because the property throws NotSupportedException on some hardware
```
pgrawehr, line 201: Comment doesn't apply here, the list of available pins for the virtual controller is known.

src/devices/Gpio/README.md
```diff
@@ -7,6 +7,26 @@ This project contains some **full function(PULL-UP, PULL-DOWN)** generic GPIO dr
 * For Allwinner SoCs: [SunxiDriver](Drivers/Sunxi/README.md)
 * For Rockchip SoCs: [RockchipDriver](Drivers/Rockchip/README.md)
 
+## Usage of vivirtual GpioController and virtual GpioPin
+
+When you have multiple GpioControllers and want to use them together in a specific binding, the `VirtualGpioController` allows to create a GpioController from GpioPin already open from different controllers. USage is straight forward and classes fully compatible with a normal `GpioController`.
```
pgrawehr, line 12: Check for typos.
```suggestion
When you have multiple GpioControllers and want to use them together in a specific binding, the `VirtualGpioController` allows to create a GpioController from a set of `GpioPin`s already open from different controllers. Usage is straight forward and the class is fully compatible with a normal `GpioController`.
```

pgrawehr (commented):
Good start. I've found some possible pitfalls you might need to look into.

src/devices/Gpio/README.md
```diff
@@ -7,6 +7,26 @@ This project contains some **full function(PULL-UP, PULL-DOWN)** generic GPIO dr
 * For Allwinner SoCs: [SunxiDriver](Drivers/Sunxi/README.md)
 * For Rockchip SoCs: [RockchipDriver](Drivers/Rockchip/README.md)
 
+## Usage of vivirtual GpioController and virtual GpioPin
```
joperezr, line 10:
```suggestion
## Usage of virtual GpioController and virtual GpioPin
```

joperezr:
Not sure if this proposal was discussed in a meeting I wasn't part of, but I have some questions about the reason why this would be needed. I thought in general we agreed on two models:
- Cases were all of the pins for a specific device come from the same controller: These cases should just be solved by having the device take in a gpioController in one of its constructors, that has access to all of those pins.
- Cases were pins used by a device come from different controlelrs: These cases should just be sulved by having the device also have a constructor that takes in individual GpioPins. 

The whole reason why we added GpioPin was so that we wouldn't need something like this (which seems to be an approach that uses an emmulation to mix the two scenarios above) since devices can just take individual pins, and we also agreed all devices would get constructors that took Gpio Pins in favor of this. Am I missing something here?

Ellerbach:
@joperezr yes, we discussed in one of the call. As we have a lot of bindings to adjust, idea is to have a transparent way to use a virtual GPIO controller! So having a virtual controller will solve this issue of having to add all those new constructors and adjust all the internal codes to support GpioPin. Because, it's not just adding ability to have a GpioPin in the constructor, it's about adjusting all the logic in all the bondings to replace GpioController.Write, Read, etc by the GpioPin.Write, Read, etc.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -416,7 +416,7 @@ protected virtual void Dispose(bool disposing)
     }
 
     /// <inheritdoc/>
-    public void Dispose()
+    public virtual void Dispose()
```
krwq, line 419: this shouldn't be virtual (only the one with bool flag should)

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -131,7 +131,7 @@ protected virtual void OpenPinCore(int pinNumber)
     /// </summary>
     /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
     /// <param name="mode">The mode to be set.</param>
-    public GpioPin OpenPin(int pinNumber, PinMode mode)
+    public virtual GpioPin OpenPin(int pinNumber, PinMode mode)
```
krwq, line 134: [Triage] Let's try to reduce number of virtual OpenPin methods, if possible only OpenPinCore

Ellerbach: I've been looking at overwriting only the OpenPinCore but the issue is that the core logic with adding the pin into the lopen list and adding the gpio pin as well is done in the OpenPin. And that is what I want to override. And as the function returns a pin that won't be store properly otherwise, I don't really have a nice option for that!

pgrawehr: Fine for me. I mean if there's an use if this is virtual, I don't bother changing it.

Ellerbach: I think I found a good compromise and reviewed a bit the internal elements. So, I only made one of the OpenPin virtual, removed the others, reused some of the logic from the GpioPin dictionary. So it's cleaner.

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        /// <inheritdoc/>
+        public override int PinCount => _pins.Count;
+
+        /// <summary>
+        /// Closes an open pin.
+        /// If allowed by the driver, the state of the pin is not changed.
+        /// </summary>
+        /// <param name="pinNumber">The pin number in the controller's numbering scheme.</param>
+        public override void ClosePin(int pinNumber)
```
pgrawehr, line 96: Implementation or description is missleading, as you don't actually close the underlying pin.

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,279 @@
...
+            }
+
+            _gpioPins.TryRemove(pinNumber, out _);
+        }
+
+        /// <summary>
+        /// Disposes this instance and closes all open pins associated with this controller.
+        /// </summary>
+        /// <param name="disposing">True to dispose all instances, false to dispose only unmanaged resources</param>
+        protected override void Dispose(bool disposing)
```
pgrawehr, line 110: Update documentation to better match the behavior

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,279 @@
...
+            _pins.Clear();
+            _gpioPins.Clear();
+        }
+
+        /// <inheritdoc/>
+        public override PinMode GetPinMode(int pinNumber)
+        {
+            if (!IsPinOpen(pinNumber))
+            {
+                throw new InvalidOperationException($"Can not set a mode to pin {pinNumber} because it is not open.");
```
pgrawehr, line 128: Error message is wrong. We're _getting_ a mode.

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,279 @@
...
+            _gpioPins.TryAdd(pinNumber, pin);
+            return _gpioPins[pinNumber];
+        }
+
+        /// <inheritdoc/>
+        public override ComponentInformation QueryComponentInformation()
+        {
+            ComponentInformation self = new ComponentInformation(this, "Virtual GPIO Controller");
+
+            self.Properties["OpenPins"] = string.Join(", ", _gpioPins.Select(x => x.Key));
```
pgrawehr, line 163: Use self.AddSubComponent() to add the drivers of the pin collection.

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,279 @@
...
+            }
+
+            _callbackEvents.TryAdd(pinNumber, new PinEvent(eventTypes, callback));
+        }
+
+        private void PinValueChanged(object sender, PinValueChangedEventArgs pinValueChangedEventArgs)
+        {
+            // Get GpioPin associated with the real life pin
+            // Note that if we have multiple pins with the same pin number, we will only take the first one in the list
+            var pins = _pins.Where(m => m.Value.PinNumber == pinValueChangedEventArgs.PinNumber);
```
pgrawehr, line 236: You could possibly check whether the sender is equal to the controller of the pin to solve this.

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,279 @@
...
+            if (_callbackEvents.Count == 0)
+            {
+                _pins[pinNumber].ValueChanged -= PinValueChanged;
+            }
+        }
+
+        /// <inheritdoc/>
+        public override WaitForEventResult WaitForEvent(int pinNumber, PinEventTypes eventTypes, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
```
pgrawehr, line 270: This should be straight-forward, shouldn't it?

src/devices/Gpio/Drivers/VirtualGpioPin.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Device.Gpio;
+
+namespace Iot.Device.Gpio.Drivers
```
pgrawehr, line 6: It's only internal, but this looks like the wrong namespace.

src/devices/Gpio/Drivers/VirtualGpioPin.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// A virtual GpioPin to allow creating a new GpioPin and a new pin number
+    /// </summary>
+    internal class VirtualGpioPin : GpioPin
```
pgrawehr, line 11: Also, do we really need this class? It does not differ from its base class, and isn't even used for type-testing anywhere, it seems.

pgrawehr (commented):
Some comments and questions still. 

Some unit tests would be nice.

src/devices/Gpio/Drivers/VirtualGpioController.cs
```diff
@@ -0,0 +1,279 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Adds a new pin number with an associated <see cref="GpioPin"/>.
+        /// </summary>
+        /// <param name="pinNumber">The pin number.</param>
+        /// <param name="gpioPin">The <see cref="GpioPin"/> to add.</param>
+        public void Add(int pinNumber, GpioPin gpioPin) => _pins.TryAdd(pinNumber, gpioPin);
+
```
raffaeler, line 76: Considering the complex numbering of the Beaglebone black discussed recently, I would consider two alternatives to the use of integer for the virtual pin numbering.
- Option 1: A string 
- Option 2: An enum. This could be done by making the controller generic and using Enum as constraint for the generic type (which is new in C# 7.3)

In both cases this could help identifying the pins with any friendly name, which is in use on any board.
Going beyond, you can also define a json map loaded in the project with all the pins getting customized names, [similar to this example](https://github.com/dotnet/iot/blob/main/src/devices/BoardLed/BoardLed.cs).

I love the idea of virtualizing pins, but IMO it should/can be more versatile.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -33,7 +33,11 @@ public class GpioController : IDisposable
     /// If a pin element exists, that pin is open. Uses current controller's numbering scheme
     /// </summary>
     private readonly ConcurrentDictionary<int, PinValue?> _openPins;
-    private readonly ConcurrentDictionary<int, GpioPin> _gpioPins;
+
+    /// <summary>
+    /// The GPIO pins open in the GpioController.
+    /// </summary>
+    protected readonly ConcurrentDictionary<int, GpioPin> _gpioPins;
```
joperezr, line 40: If we want to include it in the next release, let's mark this as experimental. Also, if this will be renamed to be protected, we should rename to not have an `_` as suffix, since that is convention for private fields.

src/System.Device.Gpio/System/Device/Gpio/GpioPin.cs
```diff
@@ -8,13 +8,40 @@ namespace System.Device.Gpio
...
+        /// Gets or sets the <see cref="GpioDriver"/>.
+        /// </summary>
+        internal GpioDriver Driver { get; set; }
+
+        /// <summary>
+        /// Creates an instance of GPIO Pin with a new pin number based on an existing <see cref="GpioDriver"/>.
+        /// </summary>
+        /// <param name="pinNumber">The GPIO pin number.</param>
+        /// <param name="driver">The GpioDriver.</param>
+        protected internal Gpio​Pin(int pinNumber, GpioDriver driver)
```
joperezr, line 28: These two APIs should also be Experimental.

src/devices/Gpio/Drivers/VirtualGpioController.cs
Ellerbach: I would love to but the original controller is not available. Only the GpioPin is available.

src/devices/Gpio/Drivers/VirtualGpioController.cs
Ellerbach: no because I don't have access to the Driver neither. If we want to expose it, then yes, it can indeed be done

src/devices/Gpio/Drivers/VirtualGpioPin.cs
Ellerbach: great catch, removed the "driver" part

src/devices/Gpio/Drivers/VirtualGpioPin.cs
Ellerbach: yes because of the constructor. It's the way to call it in a clean way

src/devices/Gpio/Drivers/VirtualGpioController.cs
Ellerbach: I've checked all this and if we want to do it we have to keep in mind few things if we want to have done all to the bottom:

* changing int to string will have an impact on performances as comparing strings and int is slower
* adjustments can be done and it will require to change the GpioPin itself to store the string instead of the pin number otherwise it will complicated much more everything. And indeed will complicated the internal manipulation

Alternative:

* if we don't want to use string at the lower level, the possibility is to generate a random, positive, non already existing pin number from the list
* For this implementation, I will then have to add a new constructor, a new list and new methods using the string on top of the int (getting the int will also be possible anyway)

src/devices/Gpio/Drivers/VirtualGpioController.cs
pgrawehr: Oh, got it. Would need the Driver property to be at least protected.

src/devices/Gpio/Drivers/VirtualGpioController.cs
raffaeler: With regards to the "alternative" solution, one possibility is to provide the GpioPin an optional `IDictionary<string, int>` to decode the string into the number that makes sense for the board.
As it is optional, if you want pure performance, you can just skip it.
If you want to use it, the `GpioPin` can use the overload that takes a string and try resolving the pin through the dictionary.

Does this make sense to you?

Ellerbach: The core issue with this is that it can then be used only when in your own code you are calling this. As all bindings takes a GpioController and pin numbers.
So the solution with alternative is the one compatible with everything in that case. It will indeed require a dictionary and few more functions but ultimately, what is going to be manipulated are just in numbers. So you'll still be able to keep the string in the virtual controller but under the hook and for compatibility reasons, it's going to be pure uint.
As an example on how it can looks like:

```csharp
// Note that we can add couple of other constructors
var virtualGpio = new VirtualGpioController();
virtualGpio.Add("A42", gpioPinA);
virtualGpio.Add("B26", gpioPinB);
var myButtonPin = virtualGpio.GetPinNumber("A42");
var button = New Button(buttonPin: myButtonPin, gpio = virtualGpio);
```

In all cases, I would make those adjustments in a separate PR.

src/devices/Gpio/Drivers/VirtualGpioController.cs
Ellerbach: I've been playing with this as well and it's not that straight forward. I may continue a bit but it's not easy and requires to create a virtual driver as well, things like this.

pgrawehr:
Superseeded by #2368

Issue #2179, opened 2023-11-22
libgpiod v2 support


huesla:
**Preface**
In order not to make this post even more complicated than it already is I will try to simplify it a bit.
It is mainly about libgpiod versions. Since these show an inconsistency I would like to clear this up first.

libgpiod has a 

1. documented version (see [releases ](https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/refs/)): In this post I mark this with the 'v' prefix, e.g. 'v2.1'
2. library version: I mark this with the ".so" prefix: e.g. '.so.3.1.0'

libgpiod versions became inconsistent from 'v1.1'. See the following table which lists the most important documented versions with library version:

v1.0.2  ->  .so.1.0.2     (last .so.1.x library)
v1.1     ->  .so.2.0.0     (first occurrence of inconsistency, first .so.2.x library)
v1.6.4  ->  .so.2.2.2     (last .so.2.x library)
v2.0     ->  .so.3.0.0     (first .so.3.x library)
v2.1     ->  .so.3.1.0     (latest .so.3.x libray)

_Checked by building each released libgpiod version, then compared to the built library version._

**Hello all,**
I would like to use the LibGpiodDriver for libgpiod v2.1 on the Raspberry Pi5.
I had a little look into the code and noticed a few things.

As you can see `Interop.libgpiod.cs` tries to bind against `libgpiod.so.2` which can be resolved to the latest available minor version e.g. `.so.2.2.2`.

Since this is current master (591188c) I assume that v2 versions (.so.3) are not yet supported.

I would be interested to know if there are future plans for this.
If this is not the case I would be willing to contribute, although it is not yet clear to me what the versioning strategy in this repo is to support binding native libraries.

What I mean is this: If you look at the code you can see that in the class `Interop.libgpiod.cs` it tries to bind the library `libgpiod `or `if >= NET6 libgpiod.so.2`.
Before NET6 `libgpiod` would tried be resolved, which gives a different result depending on the installed libgpiod version, (huh?).
With NET6 or higher, `libgpiod.so.2` would tried to be resolved, i.e. all versions from `.so.2.0.0` to `.so.2.2.2`.

Nevertheless: The bind functions in the class are the same. (No `#if NET6_0_OR_GREATER` pre processor directive)

Now you can find the method `IsLibgpiodVersion1_5orHigher()` in`LibGpiodDriver.cs`which is used to provide an additional method (`gpiod_line_bias`) via 'if branching'.
[...]

PR #2184, opened 2023-12-01
Add libgpiodv2 support

huesla:
Fixes: https://github.com/dotnet/iot/issues/2179
Hi there,
This PR aims to integrate version 2 of libgpiod
> C library and tools for interacting with the linux GPIO character device (gpiod stands for GPIO device)

**Structure**
Basically, my strategy was to extend as much functionality as possible while touching as little existing code as possible.
A lot has changed in the API between v1 and v2 of libgpiod. For example, the control of GPIOs is purely via so-called request objects. If you want to learn more about the concepts, see:
1. overview https://www.youtube.com/watch?v=EcUULioruj0&t=1s
2. docs https://libgpiod.readthedocs.io/en/latest/modules.html

As the new functions are very different from the existing ones, I have decided to differentiate between V1 and V2 in the namespace or class name. (What changes together should go together). This concerns the Interop class, as well as the driver and event handler.

The new structure from the lower to the higher level is shown as follows:

**Binding**
_Interop.libgpiod_
Maps all methods of the current libgpiod API, structured in the same way as in the docs.

_Proxies_
libgpiod offers so-called "opaque data classes". These are objects whose fields are not visible and can only be changed via methods. libgpiod divides the API logically according to objects (or modules) see [docs](https://libgpiod.readthedocs.io/en/latest/modules.html). To represent these in C#, there are now so-called proxy classes. Such a class offers the same methods as the controlled gpiod object, but in a more C#-friendly way.
`LibGpiodProxyBase` is the base class and provides methods for thread security.
(more information on [the threading contract of libgpiod](https://lore.kernel.org/linux-gpio/CAEydidkPUC1DDw40vQP22XG=ECnttLxmmaUCN2q-Xr+Gq0aadA@mail.gmail.com/T/#t)) and wraps any exception from the native call in `GpiodException` (easier for clients to handle).
I think when you know the concept of a libgpiod module, looking at the corresponding proxy class should give a clear picutre in what it does (Basically it only forwards stuff).

_Handles_
Each proxy class works with a `SafeHandle`, the object that holds the pointer to the gpiod object. SafeHandle comes from the framework and makes sure to call `ReleaseHandle()` once, which frees the gpiod object.

_Enums_
Constants see [docs](https://libgpiod.readthedocs.io/en/latest/group__line__defs.html)

_LibGpiodV2Driver_
This class is the "bridge" between `GpioController` and the binding. Each driver instance manages one gpiochip. 
It implements the usual methods and manages proxies, mainly the `LineRequest` to control GPIOs. The methods of the driver were implemented to have as little traffic as possible, e.g. reuse existing `LineRequest` objects etc.

_LibGpiodV2EventObserver_
Is closely coupled with LibGpiodV2Driver and manages EdgeEvent subscriptions / callbacks.

_GpioController_
Existing class that instantiates the driver. LibGpiodDriverFactory helps to load the correct version suitable for the system. If libgpiodv2 is available, v2 is loaded, otherwise v1 or v0.

**Testing**
Tested with manual tests and `GpioControllerTestBase` tests on Raspberry Pi 3/4/5 with latest RPI OS (lite).
I had to adapt the gpiochip number with Pi5 to 4, instead of 0 on Pi3/4. I am not sure how you treat different boards in unit tests, so I left 0 as default, because Pi5 is not yet as popular.

**Testing (continued)**
The current tests of `GpioControllerTestBase` use delays, which provokes flakyness and makes the tests slow. 
I have an improvement branch that uses reset events instead of delays. Again I tested it on the Pi 3/4/5... the time got improved alot and I did not encounter flakyness anymore:

Pi3: 50.2 sec -> 32.2 sec
Pi4: 42.8s -> 19.3s
Pi5:  40.3s -> 13.2s
_(measured only once)_

I could make a seperate PR for this after this one, depending on how the integration of this will go.
The real question is whether other drivers can keep up... would probably have to be tested with your CI pipeline.

**Commits**
The commits are chronologically logically ordered.

**Advanced use**
It should be noted that the current GpioController/Driver architecture imposes restrictions on the libgpiodv2 functionality. For example, information such as ChipInfo, LineInfo, EventInfo, Event sequence numbers etc. are not available. Also functions like setting Debounce Period, EventBuffer size etc. are not available. (One of my use cases is, for example, the time-accurate detection of edges, and the time info is in the edge event as Ns timestamp).
In order not to change the existing architecture, I have stuck to it and made proxies etc. `internal`.
However, this also means that this functionality cannot be used externally.
My suggestion is therefore to make it public. This allows advanced users extended functionality if they need it. Normal users can continue to use the GpioController API.
Since it would probably be a bit tedious (even for advanced users) to manage proxy objects directly, and to use every fine grained function, it would make sense to create an abstraction that would simplify this and cover the most relevant use cases. The code could look much nicer then, for example:

```
var request = new GpioChip(4).PrepareRequest()
    .SetConsumer("me")
    .AddLineSettings(lineSettings => 
    {
        lineSettings.Lines = new { 12, 18 };
        lineSettings.Direction = LineDirection.Input;
        lineSettings.EdgeDetection = EdgeDetection.Both;
    })
    .DoReqeust();

Task.Run(() =>
{
    foreach (var event in request.GetNextEvent())
    {
        ...
    }
};

request.SetLine(12, High);
request.SetLine(12, Low);
```
_(By the way, this example is inspired by the Python wrapper for libgpiodv2)_

I would of course also make documentation for a few examples.

I look forward to comments and opinions :)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2184)

EDIT (@krwq): adding link to the issue


huesla:
@dotnet-policy-service agree

pgrawehr:
Thanks for the submission!
Reviewing this will take a while, in particular because it touches the core parts of our library. So please be patient. 

Can you please specify on what system you ran your tests and what systems you expect to work with the new library? Are there any systems where the old driver is no longer supported and only the new one will work?

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/V1/LibGpiodV1Driver.cs
```diff
@@ -5,22 +5,24 @@
...
 using System.Runtime.InteropServices;
 using System.Collections.Concurrent;
+using System.Device.Gpio.Drivers;
+using System.Device.Gpio.Interop.Unix.libgpiod.V1;
 using System.Diagnostics;
-using System.Net.NetworkInformation;
+using LibgpiodV1 = Interop.LibgpiodV1;
 
-namespace System.Device.Gpio.Drivers;
+namespace System.Device.Gpio.System.Device.Gpio.Drivers.Libgpiod.V1;
```
pgrawehr, line 13: This namespace name repeats itself. Once "System.Device.Gpio" is sufficient.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/V1/LibGpiodV1Driver.cs
```diff
@@ -5,22 +5,24 @@
...
 
 /// <summary>
-/// This driver uses the Libgpiod library to get user-level access to the gpio ports.
+/// This driver uses libgpiod v0/1 to get user-level access to the gpio ports.
 /// It superseeds the SysFsDriver, but requires that libgpiod is installed. To do so, run
 /// "sudo apt install -y libgpiod-dev".
 /// </summary>
-public class LibGpiodDriver : UnixDriver
+public class LibGpiodV1Driver : UnixDriver
```
pgrawehr, line 20: This rename is a major breaking change, and the class name is used directly by clients, so I think this shouldn't be renamed, even if it makes the names a bit missleading.

huesla:
> Thanks for the submission! Reviewing this will take a while, in particular because it touches the core parts of our library. So please be patient.
All right, I will :)
> Can you please specify on what system you ran your tests and what systems you expect to work with the new library? Are there any systems where the old driver is no longer supported and only the new one will work?

If you ask me then the answer is Raspberry Pi. It was also Pi's 3,4 and 5 on which I ran the tests.
Basically, the libgpiod C# drivers are not directly tied to a specific system, because wherever libgpiod is supported, the drivers are too (and that is on systems where gpiolib is supported and GPIO devices are available).

On the question of whether the V2 driver can replace the V1 driver: It depends :-)
What distinguishes V2 from V1, apart from a more sophisticated API, are of course the features, such as sequence number mapping to edge events, bias setting, kernel clock setting etc...
These features could be offered publicly for advanced users of dotnet iot. Otherwise, however, it probably depends mostly on which libgpiod the known distros deliver. It will probably take some time until v2 is delivered by default.

It's also not the goal of this PR to replace the v1 driver, rather to "prepare for the future" and have a v2 implementation in place, whose features can be used.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/V1/LibGpiodV1Driver.cs
huesla: Yeah... I was a bit unsure whether this is a convention in this repo, so I went with the file location. I'm totally fine with changing it to your convention.

Documentation/raspi-libgpiodv2.md
```diff
@@ -0,0 +1,99 @@
+# Use libgpiodv2 on a Raspberry Pi
```
joperezr, line 1: Thanks a bunch for adding docs and instructions :). That said, I wonder if we should instead make this guide also consider previous versions of libgpiod. For instance, someone who is not familiar with the native library and just want to  use that driver, they may incorrectly think that they have to go through these installation steps in order to do so, which is not true as most distros already come with libgpiod, just not the newer version.

I'm not suggesting this becomes a big libgpiod guide where all of that is explained, (nor that it is your fault that we don't have good docs for this today :P), but since we are introducing some docs with the PR we should make sure these aren't confusing for users who just want to use the libgpiod driver.

Documentation/raspi-libgpiodv2.md
```diff
@@ -0,0 +1,99 @@
...
+
+If you want to control GPIOs using libgpiod version 2 the library must be installed.
```
joperezr, line 3: > libgpiod version 2 the library must be installed

This can also be confusing, since the version 2 of the library can be confused with v1.1, or v1.6.4. (as you helpfully explained in the issue). I wonder if some of that info you logged in the issue would help having documented too.

Documentation/raspi-libgpiodv2.md
joperezr: And, final doc comment, but it would be good if we find somewhere in our existing docs to link to this page as I don't think it is currently being linked in your changes.

src/System.Device.Gpio/Interop/Unix/libgpiod/V1/SafeChipHandle.cs
```diff
@@ -1,9 +1,13 @@
...
 // The .NET Foundation licenses this file to you under the MIT license.
 
+// Disable these StyleCop rules for this file, as we are using native names here.
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+
 using System.Runtime.InteropServices;
+using LibgpiodV1 = Interop.LibgpiodV1;
 
-namespace System.Device.Gpio;
+namespace System.Device.Gpio.Interop.Unix.libgpiod.V1;
```
joperezr, line 10: I understand your logic here, but 7 levels of namespace hirerarchy might be pushing it (even when only having non-public types), so I'd suggest to stay with something like `System.Device.Gpio.Interop.GpiodV1` or something along those lines.

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/GpiodException.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Disable these StyleCop rules for this file, as we are using native names here.
+#pragma warning disable SA1300 // Element should begin with upper-case letter
```
joperezr, line 5: As shared in a different comment, if we change the namespace to make it shorter we can also avoid needing to do this in every file. It is nice to have the namespace and library name match, but not a hard requirement.

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/LastErr.cs
```diff
@@ -0,0 +1,22 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Disable these StyleCop rules for this file, as we are using native names here.
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+
+using System.Runtime.InteropServices;
+
+namespace System.Device.Gpio.Interop.Unix.libgpiod.V2;
+
+internal static class LastErr
```
joperezr, line 11: NIT: Can't this be also used for V1?

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Binding/Enums/GpiodEdgeEventType.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Disable these StyleCop rules for this file, as we are using native names here.
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+
+namespace System.Device.Gpio.Interop.Unix.libgpiod.V2.Binding.Enums;
```
joperezr, line 7: NIT: Another comment for deep hirerarchy depth, I love your separation of types into folders logically, but that doesn't help much for intellisense and usability. I'd suggest to just keep everything on the same root V2 namespace we pick. Folder structure can remain to have things separate as you have here.

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Binding/Handles/EdgeEventSafeHandle.cs
```diff
@@ -0,0 +1,44 @@
...
+    protected override bool ReleaseHandle()
+    {
+        LibgpiodV2.gpiod_edge_event_free(handle);
+        return true;
+    }
+
+    public override bool IsInvalid => handle == IntPtr.Zero;
+}
+
+internal class EdgeEventNotFreeable : EdgeEventSafeHandle
```
joperezr, line 33: >EdgeEventNotFreeable [](http://example.com/codeflow?start=15&length=20)

NIT: Move to new file?

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Binding/Handles/LineInfoSafeHandle.cs
```diff
@@ -0,0 +1,44 @@
...
+    protected override bool ReleaseHandle()
+    {
+        LibgpiodV2.gpiod_line_info_free(handle);
+        return true;
+    }
+
+    public override bool IsInvalid => handle == IntPtr.Zero;
+}
+
+internal sealed class LineInfoSafeHandleNotFreeable : LineInfoSafeHandle
```
joperezr, line 33: >LineInfoSafeHandleNotFreeable [](http://example.com/codeflow?start=22&length=29)

NIT: Can we also move this to a new file?

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/LibGpiodProxyBase.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// </summary>
+    protected static void TryCallGpiod(Action action)
+    {
+        try
+        {
+            action.Invoke();
+        }
+        catch (Exception e) when (e is not GpiodException)
+        {
+            throw new GpiodException("Exception while calling libgpiod", e);
```
joperezr, line 27: I believe that this may cause issues with the lock. It is convention that `TryXxx` methods  do not throw and for them to instead return `bool success`. This is specially important here, since you are calling these methods inside a lock statement, so you have to be careful of those methods not handling exceptions correctly as you could end up in an inconsistent state or unanticipated behavior.

joperezr: cc: @stephentoub can you please take a look at this class? The TL;DR is that sub types of this base will want to have a lot of its methods to not execute concurrently, so to protect against it this is introducing a lock and some helper methods (TryCallGpiodLocked) that will try to ensure that these methods get executed in a thread-safe manner. I'm sure this is a common pattern, do we have precedent for suggestions for these cases?

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
```diff
@@ -654,7 +655,7 @@ private void Initialize()
             }
 
             // try and open /dev/gpiomem
-            fileDescriptor = Interop.open(GpioMemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
+            fileDescriptor = global::Interop.open(GpioMemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
```
joperezr, line 658: Can you help me understand why we now need to use the global namespace here? What is clashing with this?

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+    private static int? _driverVersionToLoad;
+
+    /// <summary>
+    /// Tries with best effort to find installed libgpiod libraries and loads the latest version.
+    /// <see href="https://learn.microsoft.com/en-us/dotnet/standard/native-interop/native-library-loading#custom-import-resolver"/>
+    /// </summary>
+    /// <remarks> On NET Core 3.1 or NET5+ SetDllImportResolver would be a better solution</remarks>
+    public static GpioDriver Create(int chipNumber)
+    {
+        lock (_lockObject)
```
joperezr, line 34: This is a static constructor on a static class, I don't think that we need to introduce locks go guarantee for this to only run once, unless I'm missing something.

joperezr: duh! clearly missing something haha, this is not a constructor but instead a Create method. It does make it feel expensive though to have to use a lock everytime someone calls Create. Can we consider having a static constructor that essentially finds out which version to use and cache t hat, and then all subsequent calls to Create can avoid locking?

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+            2 => new LibGpiodV2Driver(chipNumber),
+            _ => throw new ArgumentOutOfRangeException(nameof(driverVersionToLoad), $"Unexpected libgpiod driver version: '{driverVersionToLoad}'")
+        };
+    }
+
+    private static int GetDriverVersionToLoad()
+    {
+        HashSet<string> foundLibraryFiles = new();
+
+        foreach (string searchPath in _librarySearchPaths)
```
joperezr, line 60: I honestly don't love this approach of having some magic places where we look for things, it is code that will likely cause issues in the future. Is there a PInvoke present on all versions that return the version number of the library? If so, I'd rather us do a discovery attempt where we try PInvoking to get that info, and similar to the previous approach use NativeLibrary.TryLoad to do the discovery. Having a list like this is very fragile, and likely to be broken for some scenarios.

joperezr:
Thanks a bunch for the contribution @huesla, I went through it and added some feedback (might seem like a lot but most are just either NITs or questions) but this is overall looking really good. Thanks a lot again for taking the time to contribute this.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/V1/LibGpiodV1Driver.cs
joperezr: Yeah, as @pgrawehr points out, unfortunately we cannot take this breaking change as it would have high impact (a lot of folks are creating GpioControllers today like: `... = new GpioController(new LibGpiodDriver())`). I so much regret that we didn't make LibGpiodDriver a Factory pattern as that would have been ideal, and we would just have V1 and V2 internal concrete implementations that would be returned by a Create method. Anyway, that ship has sailed now, and in terms of design, I don't think callers would ever want to select whether they use V1 or V2 of libgpiod driver, so no point on making changes in exposing those classes. From the consumer point of view, they should only select that they want LibGpiodDriver, but then for that to internally figure out if it should use V1 or V2 depending on what native library is installed.

For that reason, our public API should probably just keep LibGpiodDriver as the one entry point class, and for that to redirect internally to V2 stuff if needed.

pgrawehr: I think we should be careful with adding new automatic detection here, this has mostly caused issues (remember the mess we created with the RaspberryPi3Driver?) We should try to move the detection to separate classes going forward, e.g. to the `Board` namespace.

src/System.Device.Gpio/Interop/Unix/libgpiod/V1/SafeChipHandle.cs
huesla: Didn't know your convention so I went with file location. Will change it to your suggestion

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/LibGpiodProxyBase.cs
huesla: For clarification: libgpiodv2 is thread aware but not thread safe. That means libgpiod objects may be accessed from any thread but only one at a time. Also I found it useful to wrap any exception, to not require much exception handling on caller side, and bloating docs that has to list every possible exception.
This base class implements these aspects, for code use. I do not see any issue regarding locks.
I am open to rename the methods though to not clash with the Try pattern :)

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/Chip.cs
```diff
@@ -0,0 +1,293 @@
...
+                $"{nameof(ChipInfo)}: {ChipInfo}, {nameof(Path)}: {Path}, {nameof(FileDescriptor)}: {FileDescriptor}, {nameof(LineInfos)}: {string.Join("\n", LineInfos)}";
+        }
+    }
+
+    #region Dispose
+
+    private readonly object _isDisposedLock = new();
+    private bool _isDisposed;
+
+    public void Dispose()
```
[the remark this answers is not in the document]
huesla: The SafeHandles are freed by the framework. My thought here was to only implement IDisposable where the user really requires it, which was in my case for automated tests (to unreserve gpio lines at teardown, to not clash with other tests)...

src/System.Device.Gpio/System/Device/Gpio/Drivers/RaspberryPi3LinuxDriver.cs
huesla: So this was necessary after moving for example the class `SafeChipHandle` to it's file location namespace, so from `namespace System.Device.Gpio;` to `namespace System.Device.Gpio.Interop.Unix;` (I did this be consistent but since you do not have the convention to have the namespace same as file location I may just revert this)
I have not found yet why after that the "global" was needed...

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
huesla: Yeah that makes sense. Let's see on other comments on how we will move on with this factory

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
huesla: I don't like it either but it was the best thing I could come up with. It basically is based on the file system hierarchy standard of linux, that most distros follow.
If I could choose the framework to do this, I would, but `NativeLibrary` does not seem to be available. Is this because netstandard 2.0 is targeted?

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/V1/LibGpiodV1Driver.cs
huesla: I can't say much about that. Except that it is not impossible that distros will provide libgpiodv2 by default at some point.
By then at the latest there should be automatic detection or the possibility for users to determine the version.

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/LastErr.cs
huesla: Er... yes. Basically anywhere where` GetLastWin32Error()` is used. I thought it would be good that when migrating to >=NET7, this feature is also used.

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/Chip.cs
```diff
@@ -0,0 +1,293 @@
...
+    /// <param name="devicePath">File system path to the chip device, e.g. '/dev/gpiochip4'</param>
+    /// <exception cref="GpiodException">The chip does not exist or an unexpected error happened while opening chip</exception>
+    /// <seealso href="https://libgpiod.readthedocs.io/en/latest/group__chips.html#ga25097f48949d0ac81e9ab341193da1a4"/>
+    public Chip(string devicePath)
+    {
+        _handle = TryCallGpiod(() => LibgpiodV2.gpiod_chip_open(Marshal.StringToHGlobalAuto(devicePath)));
+
+        if (_handle.IsInvalid)
+        {
+            throw new GpiodException($"Could not open gpio chip at path '{devicePath}': {LastErr.GetMsg()}");
```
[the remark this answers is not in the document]
huesla: True. I tried bringing attention to this with documentation, but it doesn't seem enough. I can make a factory. I thought about `static T Create<T>() where T : LibGpiodProxyBase`, what do you think?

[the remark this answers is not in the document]
huesla: Wait this doesn't work with arguments. Have to find something else

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
joperezr: maybe, honestly I don't think Netstandard2.0 is getting us much here, so we should probably consider dropping it. Main place where it is still useful is for devices like raspberry pi nano that can run Mono but not .NET Core.

huesla: I see following options:
1. on .NET standard use the loading approach of this class, on  >= NET Core 3.0 use NativeLibrary
2. drop netstandard support and only use NativeLibrary (unsure about impact)

With either option we should consolidate the code of LibgpiodV1.Interop static ctor with the new approach.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/V1/LibGpiodV1Driver.cs
huesla: I have pushed changes that contain the solution of keeping LibGpiodDriver and it's namepsace to not break things. In the constructor it will resolve the proper driver version and act as a bridge to it. How the automatic detection is implemented is not yet final obviously and is discussed in another comment

krwq:
[Triage] Is it possible to install both libgpiod and libgpiod2 side by side? If not then we should try to make existing implementation work (it could use new code as implementation detail or derive).

huesla:
> [Triage] Is it possible to install both libgpiod and libgpiod2 side by side? If not then we should try to make existing implementation work (it could use new code as implementation detail or derive).

Usually different versions of a library can coexist. I am not sure what you mean with the second sentence

raffaeler:
> Usually different versions of a library can coexist. I am not sure what you mean with the second sentence

During the triage we were discussing whether the user of the library should be able or not to decide to bind against libgpiod v1 or v2.
- If both libraries can coexist, there may be multiple reason to pick one or the other (for example, the v2 is installed after the software is working, but you still want to rely an app on v1 for whatever reason).
- If instead the v2 installation process automatically uninstall v1, there is no reason for the user to decide to use v1 or v2. Our library should instead detect and use the currently installed library.
This is what happens now when libgpiod is not installed at all. In this case the sysfs driver is automatically picked and this process is transparent to the user. This is what the second sentence was referring to.

Since you are actively working on this PR (thanks for that :-) ) we want to know for sure whether both libraries can coexists on the same installation as this implies a different strategy as I hope to have explained above.

joperezr:
And just to add to what @raffaeler explained, this will be helpful on whether we should make LibgpiodV2 driver public or keep it as an implementation detail. If you find that the library can't be installed side by side with other version, then there is no point in making the v2 driver public, as consumers will mainly only want to select to  use libgpiod and we should internally figure out which driver to use (v1 or v2). If, on the other hand, the versions can live side by side, then we should probably keep the driver public and let the user decide if they choose.

joperezr:
I think a safe bet would be to start by keeping it internal (we can always make it public later), in favor of not over complicating what we expose as well as allowing existing apps to just automatically retarget to v2 when the version in their OS is updated.

huesla:
@joperezr It really depends on how the user installs libgpiod. Probably most users will use a package manager that "normally" replaces older library files. I would have at least tested it with APT, but there is no v2 package yet. 
In the rare case that the user installs the library manually, 2 versions of it may exist in the filesystem.
So to answer your question about whether 2 versions can exist: maybe, but in most cases probably not.
I think the more important argument for making the driver version public or not is the question of whether the features of v2 should be public (as discussed). 
Since you do not share this desire (yet?), it will make no difference to the user whether they choose the v1 or v2 driver.
So I agree that it's best to leave it internal and load the appropriate version dynamically under the hood.
I would even go so far as to have v1 selected by default if multiple versions are installed. This is more in line with the existing function.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/V1/LibGpiodV1Driver.cs
huesla: I think we can agree on the current implementation that does not contain breaking changes. If you have comments on automatic driver version loading, please make another comment, I will resolve this conversation

Documentation/raspi-libgpiodv2.md
huesla: I think we can agree on the user not having to select a specific driver version, loading it automatically under the hood. I will change these docs to only give necessary information about libgpiod.

huesla:
As you can see LibGpiodDriver is now "hiding" the fact which driver version is loaded. Still the types `LibGpiodV1Driver` and `LibGpiodV2Driver` are public. They are used in tests and cannot simply be made internal, except if we use "InternalsVisibleTo". I gave this a shot locally, but the compiler complains:
`error CS1726: Friend assembly referen
ce 'System.Device.Gpio.Tests' is invalid. Strong-name signed assemblies must specify a public key in their InternalsVis
ibleTo declarations`
Which route would you go here?

huesla:
I found another solution. Making tests use the `LibGpiodDriverFactory` which is driver version aware, and marking the versioned drivers internal. This is the same factory that `LibGpiodDriver` uses, you know, the thing that is not a driver but a factory...

pgrawehr:
> I found another solution. Making tests use the `LibGpiodDriverFactory` which is driver version aware, and marking the versioned drivers internal. This is the same factory that `LibGpiodDriver` uses, you know, the thing that is not a driver but a factory...

That is a good solution I think, particularly because you moved the actual factory to its own class. I'd only make that API a bit more versatile, so that it is future-proof.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// Creates and returns an instance of the libgpiodv1 driver.
+    /// </summary>
+    /// <param name="chipNumber">The number of the GPIO chip to drive.</param>
+    /// <param name="waitEdgeEventsTimeout">Timeout to wait for edge events. Primarily used for testing.</param>
+    public static UnixDriver CreateV2Driver(int chipNumber, TimeSpan? waitEdgeEventsTimeout = null)
+    {
+        return CreateLibgpiodV2DriverInstance(chipNumber, waitEdgeEventsTimeout);
+    }
+
+    private static UnixDriver InstantiateDriver(int driverVersionToLoad, int chipNumber)
```
pgrawehr, line 69: I'd rename this to an overload of `Create` and make it public. That's more generic than havig a `CreateV1Driver`, `CreateV2Driver`, `CreateVxxxDriver` and so on.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+        return new LibGpiodV1Driver(chipNumber);
+    }
+
+    private static LibGpiodV2Driver CreateLibgpiodV2DriverInstance(int chipNumber, TimeSpan? waitEdgeEventsTimeout = null)
+    {
+        Chip chip = LibGpiodProxyFactory.CreateChip(chipNumber);
+        return new LibGpiodV2Driver(chip);
+    }
+
+    private static int GetDriverVersionToLoad()
```
pgrawehr, line 90: (this is independent of the internal implementation) I'll consider making this public, too. So the user knows which version is going to be preferred. And then maybe return an instance of `System.Version` so the user doesn't do anything stupid with the returned value.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
huesla: Hm not sure about this... It would require to also abstract the method parameters. I'm not a fan of this, considering that the number of possible versions remains manageable over a longer period of time. Honestly I don't really expect/want the user to call this (it got introduced for tests primarily :poop:)

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/Chip.cs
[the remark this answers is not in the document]
huesla: There is now a proxy factory class that takes over creation of proxy objects that previously called libgpiod in the constructor

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/LibGpiodProxyBase.cs
huesla: As it turns out reading and writing concurrently is in most cases not a problem see [conversation](https://lore.kernel.org/linux-gpio/CAEydidkEikZciGD_EzS95tMXznPH=OFP__Q2Bkk6HrQ+sNhkxQ@mail.gmail.com/T/#t)
Changed it so only writes are now locked.

krwq:
[Triage] We've discussed the design a bit and we think it should look like following:

```csharp
partial class LibGpiodDriver
{
   public LibGpiodVersion Version { get; protected set; }
   // existing
   // public LibGpiodDriver Create(int chip); // env var to pick default
   
   public LibGpiodDriver Create(int chip, LibGpiodVersion version);

   public LibGpiodVersion[] GetAvailableVersions();
}

enum LibGpiodVersion
{
    LibGpiodV1,
    LibGpiodV2,
}
```

Environment to control version
DOTNET_IOT_LIBGPIOD_VERSION=<enumName LibGpiodVersion>
specified => always use explicit version but do not allow fallback
not specifed => latest

huesla:
Cool stuff. I can see which path you want to take. Make V1 and V2, as well as the implicit selection of these, internal.
The user can pick the version with an ENV variable. Unit tests use the ENV variable to target a versioned driver independently of the access modifier :+1:.

However, the attached example still raises some questions, which I will try to describe as precisely as possible:

Are the methods `Create()` and `GetAvailableVersions()` in the provided example meant to be static... shall they replace the constructor? I would guess so.
Note that in current main branch `LibGpiodDriver` has no `Create()` method, only a constructor taking the chip number.

Second, I am not sure whether I fully get the class structure, it could mean one of the following:
1. `LibGpiodDriver` is a partial class that contains implementations for both `V1` and `V2` (maybe in different files), where each method requires an "if" to call either implementation V1 or V2 (etc.)
2. `LibGpiodDriver` is a base class (partial needed?) that has `V1`/`V2` as sub classes, both internal, that override all base methods. `LibGpiodDriver` cannot provide any implementation at that point (which normally calls for being declared abstract, but again this would be a breaking change)
3. `LibGpiodDriver`, `V1` and `V2` are independent classes (no inheritance, no partial). LibGpiodDriver instantiates one of V1/V2 and simply forwards calls to it.

I would go with 3. either breaking or non breaking:

Non breaking:

```c#
public class LibGpiodDriver : UnixDriver
{ 
    private readonly UnixDriver _versionedLibgpiodDriver;
	
    // exists
    public LibGpiodDriver(int chipNumber = 0)
    {
        // if env var set select that version, if found ok, if not found, fail
        // else pick version automatically based on installed library
        // instantiate _versionedLibgpiodDriver with proper driver implementation (other class)
    }
	
    // new
    public LibGpiodDriver(int chipNumber = 0, LibGpiodVersion version)
    {
        // choose selected version, or fail if not found
        // instantiate _versionedLibgpiodDriver with proper driver implementation (other class)
    }

    public LibGpiodVersion Version { get; protected set; }

    public LibGpiodVersion[] GetAvailableVersions() => { ... };

    protected internal override int PinCount => _versionedLibgpiodDriver.PinCount;
    ...
}
```

Breaking:

```c#
public class LibGpiodDriver : UnixDriver
{ 
    private UnixDriver _versionedLibgpiodDriver;

    private LibGpiodDriver() { ... }

    // static method instead of constructor
    public LibGpiodDriver Create(int chip)
    {
        // if env var set select that version, if found ok, if not found, fail.
        // else pick version automatically based on installed library
        // instantiate _versionedLibgpiodDriver with proper driver implementation (other class)
     }

    public LibGpiodDriver Create(int chip, LibGpiodVersion version)
    {
        // choose selected version, or fail if not found
        // instantiate _versionedLibgpiodDriver with proper driver implementation (other class)
    }
	
    public LibGpiodVersion Version { get; protected set; }

    public LibGpiodVersion[] GetAvailableVersions() => { ... };
	
    protected internal override int PinCount => _versionedLibgpiodDriver.PinCount;
    ...
}
```

pgrawehr:
> Cool stuff. I can see which path you want to take. Make V1 and V2, as well as the implicit selection of these, internal. The user can pick the version with an ENV variable. Unit tests use the ENV variable to target a versioned driver independently of the access modifier 👍.
> 

Unit tests don't need to use the ENV variable, they can just use the new ctor.

> 
>     3. `LibGpiodDriver`, `V1` and `V2` are independent classes (no inheritance, no partial). LibGpiodDriver instantiates one of V1/V2 and simply forwards calls to it.
> 
> 
> I would go with 3. either breaking or non breaking:
> 
> Non breaking:
> 
> ```cs
> public class LibGpiodDriver : UnixDriver
> { 
>     private readonly UnixDriver _versionedLibgpiodDriver;
> 	
>     // exists
>     public LibGpiodDriver(int chipNumber = 0)
>     {
>         // if env var set select that version, if found ok, if not found, fail
>         // else pick version automatically based on installed library
>         // instantiate _versionedLibgpiodDriver with proper driver implementation (other class)
>     }
> 	
>     // new
>     public LibGpiodDriver(int chipNumber = 0, LibGpiodVersion version)
>     {
>         // choose selected version, or fail if not found
>         // instantiate _versionedLibgpiodDriver with proper driver implementation (other class)
>     }
> 
>     public LibGpiodVersion Version { get; protected set; }
> 
>     public static LibGpiodVersion[] GetAvailableVersions() => { ... };
> 
>     protected internal override int PinCount => _versionedLibgpiodDriver.PinCount;
>     ...
> }
> ```
> 

Sorry, yea, we missed some points here (unfortunately, the team's weekly telephone call is often interrupted by various other meetings, so we need to be quick). I do think Case 3 (non-breaking) is exactly what we meant to have. We have `Create()` methods in other places, but since here, the ctor is already public, we should stick to that scheme. 

The method `LibGpiodVersion[] GetAvailableVersions()` _should_ be static, though. Otherwise it is of little use, since no instance of `LibGpiodDriver` can be created when not knowing whether it is available at all.

huesla:
> Unit tests don't need to use the ENV variable, they can just use the new ctor.

True, almost missed that!
I like the proposed solution and will implement it soon.

huesla:
I just pushed new changes. As discussed the driver now has additional constructor taking the driver version, the driver factory is now internal and considers  an environment variable.
@pgrawehr while writing the code I couldn't find sense in `GetAvailableVersions`... what should it return? All the values contained in the LibGpiodDriverVersion enum?
Or did you deliberately call it "LibGpiodVersion" and not "LibGpiodDriverVersion" i.e. you are talking about the libgpiod version and not the driver version

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
huesla: We have agreed on changes to the factory that obsolete this, so I am going to close this

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
huesla: We have agreed on changes to the factory that obsolete this, so I am going to close this

pgrawehr:
> while writing the code I couldn't find sense in `GetAvailableVersions`... what should it return? 

The intent was to have a function that returns the driver versions that can (most likely) be instantiated. So it would return V1 when the V1 .so exists and/or V2 when the V2 .so exists. Basically the translated return value of `GetInstalledLibraries`.

huesla:
@pgrawehr makes sense. I added it now

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/LastErr.cs
huesla: Smth to be done about this?

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/LibGpiodProxyBase.cs
huesla: @joperezr The TryXXX methods are now renamed, and locks are only held on reads. Is it ok for you to resolve this?

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/GpiodException.cs
```diff
@@ -3,17 +3,29 @@
 
 namespace System.Device.Gpio.Libgpiod;
 
-internal class GpiodException : Exception
+/// <summary>
+/// Exception in the context of calling libgpiod
+/// </summary>
+public class GpiodException : Exception
```
pgrawehr, line 9: GPIO communication errors so far throw `IOException`, so this should probably derive from that. 

Additionally, the namespace is wrong, should be the same as the public LibGpiodDriver.

huesla: changed namespace/file location and let it derive from IOException

huesla:
> /azp run dotnet.iot

The documentation should be fixed now. I don't see why tests are failing though, is libgpiodv2 installed on the build agent?

pgrawehr:
> > /azp run dotnet.iot
> 
> The documentation should be fixed now. I don't see why tests are failing though, is libgpiodv2 installed on the build agent?

Unfortunately, there are more documentation issues (yes, the linter is touchy). See https://github.com/dotnet/iot/pull/2168#issuecomment-1826429276 for a tip on how to test that locally.

When that doesn't work, the actual build isn't even started, so we don't see that result. And no, it's very unlikely the runners have v2 installed. They have not been updated in a very long time.

huesla:
@pgrawehr hopefully I could fix all the md issue this time

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
```diff
@@ -0,0 +1,161 @@
...
+            LibGpiodDriverVersion.V2 => new LibGpiodV2Driver(LibGpiodProxyFactory.CreateChip(chipNumber)),
+            _ => throw new ArgumentOutOfRangeException(nameof(version), version, null)
+        };
+    }
+
+    private static IEnumerable<string> GetInstalledLibraries()
+    {
+        HashSet<string> foundLibrariesFileName = new();
+
+        foreach (string searchPath in _librarySearchPaths)
```
pgrawehr, line 133: This hardcoded search path list is a bit ugly. Can we use $path instead?

pgrawehr (commented):
API looks fine to me now. I did take a look at some of the implementation details and found some possible improvements. Especially the disposal should be reviewed carefully. See also PR https://github.com/dotnet/iot/pull/1861 for a known issue with the existing driver (which we shouldn't copy to the new).

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Binding/Enums/GpiodLineBias.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Device.Gpio.Libgpiod.V2;
+
+/// <seealso href="https://libgpiod.readthedocs.io/en/latest/group__line__defs.html"/>
+internal enum GpiodLineBias
+{
+    AsIs = 1,
+    Unknown,
```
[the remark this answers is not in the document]
huesla: It is based on the original C implementation, but I guess being explicit can be more clear

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Binding/Handles/ChipSafeHandle.cs
```diff
@@ -0,0 +1,23 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.InteropServices;
+using LibgpiodV2 = Interop.LibgpiodV2;
+
+namespace System.Device.Gpio.Libgpiod.V2;
+
+internal class ChipSafeHandle : SafeHandle
+{
+    public ChipSafeHandle()
```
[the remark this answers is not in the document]
huesla: No, tests have showed that after freeing the chip, any other resource related to it, like a request still is valid and can be used

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
```diff
@@ -0,0 +1,161 @@
...
+    private static readonly LibGpiodDriverVersion? _driverVersionSetByEnvVar;
+
+    /// <remarks>
+    /// Driver version that is picked by this factory when no version is explicitly specified. Null when libpgiod is not installed.
+    /// </remarks>
+    private static readonly LibGpiodDriverVersion? _automaticallySelectedDriverVersion;
+
+    private static readonly string[] _librarySearchPaths = { "/lib", "/usr/lib", "/usr/local/lib" }; // Based on Linux FHS standard
+
+    static LibGpiodDriverFactory()
```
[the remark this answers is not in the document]
huesla: In another comment there was a performance concern about the previous implementation that used a static method with a lock, so I moved this into a static constructor. I don't think we should move away from `static`, because executing this logic only once to cache the result makes sense for performance reasons. I do understand your concern though about the `TypeLoadException`, which is not really nice, so I would vote for the previous approach, using a static method instead.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
huesla: $PATH does not point to libraries (usually) but to directories containing executables.
In [this comment](https://github.com/dotnet/iot/pull/2184#discussion_r1418766214) we discussed what to do about this uglyness but have not come up with a better approach (yet)

huesla:
> API looks fine to me now. I did take a look at some of the implementation details and found some possible improvements. Especially the disposal should be reviewed carefully. See also PR #1861 for a known issue with the existing driver (which we shouldn't copy to the new).

I have carefully read #1849 and PR #1861 to understand the root problem and make sure it does not sneak into V2. So I tried reproducing the issue with the provided code, but I wasn't able to cause a segfault. Unsurprisingly, because the example expects the controller objects to be garbage collected (it calls GC.Collect) even though these objects live in the same scope (?)
So I tried other scenarios, but even the simplest one where first the chip is freed, then the line is released, no segfault occured.

I somehow have the feeling that either the analysis or the provided reproduction code of the issue isn't 100% right.

Nonetheless, I can't see resource release issues in the V2 implementation as it heavily relies on SafeHandles (and freeing only once).

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/Chip.cs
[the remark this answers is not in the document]
huesla: The safe handles can be freed independently from each other, so it should™ work. I made the proxies all disposable now to match the pattern

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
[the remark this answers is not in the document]
pgrawehr: [Triage] We discussed this briefly and we think the best solution would be to merge the two concepts and make the _class_ a singleton. So have all the fields and methods non-static, but offer a static `Instance` property that creates that instance exactly once. Like this, the overhead of scanning is required only once, while any exceptions will still be deterministic and hence catchable. You'll find some hints about singleton implementation [here](https://csharpindepth.com/articles/singleton) (just in case). 

Sorry this takes so many rounds.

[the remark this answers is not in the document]
huesla: No worries! I'm always open for improvements and learnings. It's true that the Singleton pattern can be an alternative to a static class, though I think it won't fully solve the problem, but only defer the root problem. 

```c#
class Factory
{
    private Factory()
    {
      if (someReason)
          throw new Exception();
    }

    public static Factory Instance { get; } = new Factory();
}
```

When accessing `Instance`, again a `TypeInitializationException` is thrown, for which there exists no sane reason.
I think the best solution here is to make sure not to throw when any static initializer is called.
I have combined your Singleton idea with my thoughts, and pushed changes that utilize `Lazy`. This prevents `TypeInitializationException` and only propagates e.g. PermissionDenied etc.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
```diff
@@ -33,83 +33,95 @@ internal static class LibGpiodDriverFactory
...
+        _automaticallySelectedDriverVersion = new Lazy<LibGpiodDriverVersion?>(GetAutomaticallySelectedDriverVersion, LazyThreadSafetyMode.PublicationOnly);
+        _driverVersionEnvVarValue = new Lazy<string?>(GetDriverVersionEnvVarValue, LazyThreadSafetyMode.PublicationOnly);
+        _driverVersionSetByEnvVar = new Lazy<LibGpiodDriverVersion?>(GetDriverVersionSetByEnvVar, LazyThreadSafetyMode.PublicationOnly);
+    }
 
-        _automaticallySelectedDriverVersion = DriverCandidates.Any() ? DriverCandidates.Max() : null;
+    /// <summary>
+    /// Singleton instance.
+    /// </summary>
+    public static LibGpiodDriverFactory Instance { get; } = new();
```
pgrawehr, line 76: This singleton pattern is not thread safe. Not sure it would matter here, but better use one of the templates I added to the last post.

raffaeler: I think it is thread safe as it is similar to the following which is thread safe:
```
private static LibGpiodDriverFactory _instance = new();
public LibGpiodDriverFactory Instance => _instance;
```

The static fields are initialized lazily, as soon as anything related to that class is ever used.
Other singleton patterns are not necessarily thread safe, but the static fields initialization is guaranteed to be atomically initialized from the CLR.

pgrawehr: True, you're right. But it's not lazy, as indicated in some other comment.

raffaeler: It is, until you use the `LibGpiodDriverFactory` for any reason.
As soon as the `JIT` sees that type in a method, it triggers its initialization and therefore the instance is created.

There are other, slightly more complex, implementations of the Singleton pattern which are thread-safe and use weak initialization. But I don´t believe it is worth the effort for this type.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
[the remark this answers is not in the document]
pgrawehr: Yes, the way you wrote it here is not safe about TypeInitialisationException. You can either use Lazy, or something llike this:

```csharp
public static Factory Instance
{
    get
    {
        lock (_lock)
        {
            if (_instance == null)
                _instance = new Factory();
        }
        return _instance;
    }
}
```

This way, the constructor is only called when the factory is accessed the first time and not when the class is accessed the first time. 

Note the difference between
```csharp
public static Factory Instance { get; } = new Factory();
```
and
```csharp
public static Factory Instance { get { return new Factory(); } };
```

These two look similar, but their semantics is very different: The first creates a static automatic property, which is initialized using an auto-generated static constructor, while the second creates a calculated property without a backing field.

pgrawehr:
There's some problem with the build too: You can ignore the Windows builds for now (what's failing there is some strange compatibility check), but the Linux Helix tests (which run the code on actual Raspberry Pi 3's) fails with an error about LibGpiod not found. You should take a look at that.

huesla:
> There's some problem with the build too: You can ignore the Windows builds for now (what's failing there is some strange compatibility check), but the Linux Helix tests (which run the code on actual Raspberry Pi 3's) fails with an error about LibGpiod not found. You should take a look at that.

The LibGpiodDriverV2Tests expect libgpiodv2 (.so3) to be installed, otherwise they fail. Who can/may take care of that?

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
raffaeler: I believe we are saying the same things:
- The initialization is lazy because it is done only after the type is used (which is good IMO)
- That specific implementation of the singleton pattern is thread safe, so there is no need to use `lock`.

As soon as a different implementation is chosen (static method creating the instance), the use of `lock` is needed to guarantee the thread safety.

Also, I agree with Patrick that the goal here is to move away from the static constructor because, should a `TypeInitializationException` occur inside it, the exception argument would not provide enough clear clues to understand the culprit and, of course, will make the entire app immediately fail.

raffaeler: I risk to go OT now, but let's look at the following snippet run with both  .NET Framework 4.7 and .NET 8.
This uses only "old" C# features to make it run with .NET Framework as well:
```csharp
using System;
public class C {
    private static C _instance = new C();
    public static C Instance { get { return _instance; } }

    public static void C1() { Console.WriteLine("C1"); }
    private C() { Console.WriteLine("Instance created"); }
}

public class Program
{
    public static void Main()
    {
        Console.WriteLine("Main");
        var p = new Program();
        p.M1();
        p.M2();
    }
    
    public void M1() { Console.WriteLine("M1"); }
    public void M2()
    {
        Console.WriteLine("M2 {0}", typeof(C));
        C.C1();
        C.Instance.ToString();
    }   
}
``` 

.NET Framework (regardless the version 4.7 or 4.8)
```
Main
M1
M2 C
Instance created
C1
```

.NET (regardless the version) >= 3.x:
```
Main
M1
M2 C
C1
Instance created
``` 

The point is that this implementation is always lazy. In .NET (core) the instance is not created until you really want to access the instance method.
If you just stick with static members, the instance is never created at all.

huesla: We are on the same page. I did not know this subtle difference between the dotnet versions, although it doesn't really matter here because `Instance` is the only public static member, so regardless of whether `get { ... }` or static constructor is used, only accessing `Instance`  will cause initialization (or maybe I am missing something?).
To summarize a bit: this factory should perform the initialization logic 1. only once 2. lazy and 3. thread safe. The current implementation utilizes the `Lazy` type to achieve 2. 3. and `static` for 1. When referencing `Instance`, the (automatically generated) static constructor gets called that initializes the `Lazy` members. This will not throw and consequently `TypeInitializationException` is avoided.
When we compare this solution with one that uses `get { ... }` and a lock, I only see performance being the last factor left, but I don't think it's worth the effort going there. This class will most likely not be used in hot spots, unless someone has the idea to create millions of  driver instances for some reason 😃

raffaeler: I'm not sure about the thread safety of the getter initialization. It probably is, I would rather prefer the following, which is a one-liner having all the requirements and leave no doubts :-)
```csharp
public static readonly LibGpiodDriverFactory Instance = new();
```

pgrawehr:
> 
> The LibGpiodDriverV2Tests expect libgpiodv2 (.so3) to be installed, otherwise they fail. Who can/may take care of that?

We can look into that (nobody has touched those in a while, so I dono), but the issue is not with V2, but with V1. Here's the relevant error:

```
 System.Device.Gpio.Tests.LibGpiodV1DriverTests.AddCallbackRemoveAllCallbackTest [FAIL]
      System.Device.Gpio.Drivers.GpiodException : No suitable libgpiod library found for LibGpiodDriverVersion.V1. Supported versions: libgpiod.so.0, libgpiod.so.1, libgpiod.so.2
      Searched paths: /lib, /usr/lib, /usr/local/lib
      Stack Trace:
        /_/src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs(128,0): at System.Device.Gpio.Drivers.LibGpiodDriverFactory.CreateInternal(LibGpiodDriverVersion version, Int32 chipNumber)
        /_/src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs(106,0): at System.Device.Gpio.Drivers.LibGpiodDriverFactory.Create(Int32 chipNumber, LibGpiodDriverVersion driverVersion)
        /_/src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriver.cs(38,0): at System.Device.Gpio.Drivers.LibGpiodDriver..ctor(Int32 gpioChip, LibGpiodDriverVersion driverVersion)
        /_/src/System.Device.Gpio.Tests/LibGpiodV1DriverTests.cs(22,0): at System.Device.Gpio.Tests.LibGpiodV1DriverTests.GetTestDriver()
        /_/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs(320,0): at System.Device.Gpio.Tests.GpioControllerTestBase.AddCallbackRemoveAllCallbackTest()
```

huesla:
> We can look into that (nobody has touched those in a while, so I dono), but the issue is not with V2, but with V1. Here's the relevant error:
> ...

The exception message is a bit misleading, since it does not include the installed library file names. I will add this in my next commit.
It complains about V1 and also V2 so there is definitely an issue with finding the library and thus with the factory, because I assume this previously worked. We could try to find out where the actual library is installed on the build agent and why the factory doesn't find it, but the framework in the previous implementation could. Something like `sudo find / -iname "libgpiod.so*`
If it turns out that the library is installed not in an expected directory, we could think further what to do.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
huesla: I might as well just use the other approach. I find it's readability better :)

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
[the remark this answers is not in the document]
huesla: As mentioned in another comment it would be safe due to initializing only Lazy members, but your approach is more readable so let's go with that one.

huesla:
Open points: 
1. Who ever has the rights, should investigate why the tests fail on this pipeline, i.e. find out why the factory implementation does not find libgpiod in the standard paths (where is libgpiod installed?)
2. Depending on 1. find out what to do with the factory implementation

pgrawehr:
@huesla I've tried to fix the remaining issues, but the thing with the library search path is complicated. On my Rpi4 (with 64 bit Raspberry Pi OS) the correct path for `libgpiod.so.2` is `/lib/aarch64-linux-gnu/libgpiod.so.2`. The only way to get to this path appears to be the output of `ldconfig -p`. This tool shows the search path of the dynamic linker. I'm not sure why this is different for you. Which operating system are you using?

huesla:
@pgrawehr thanks for your investigation.
I just did a fresh install of 32/64 bit RPI OS on Pi4/Pi5. On both Pi's the result was:
- 32 Bit: /usr/lib/arm-linux-gnueabihf/libgpiod.so.2.2.2
- 64 Bit: /usr/lib/aarch64-linux-gnu/libgpiod.so.2.2.2

I think on my previous system setup the libraries ended up in "/usr/lib" because I may have compiled and installed them manually only. But under Debian there is [multi arch support](https://askubuntu.com/a/558527).
An easy fix would be to make the search recursive, going downwards starting from "/usr/lib", wdyt?

pgrawehr:
Sounds like a plan. I was first thinking about parsing the output of ldconfig, but that's quite ugly to do. Of course, recursively iterating the subfolders of /usr/lib could theoretically find the wrong library (e.g. the 32bit library for a 64 bit process), but normally you would have either both or none, I suppose.

huesla:
@pgrawehr the search is now recursive. Should we enable tests again?

huesla:
@pgrawehr the V1 tests succeeded, can we run V2 aswell?

pgrawehr:
@huesla No, because the devices have not been upgraded yet. That may take some time, but I will push that further. I don't have the necessary permissions myself. When the old driver works, it's fine for me to merge anyway.

pgrawehr (approved):
Looks good to me now.

Ellerbach (approved):
Great work and great collaboration in this PR. I've been following the different threads and progress.

src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriverFactory.cs
```diff
@@ -0,0 +1,247 @@
...
+using System.Device.Gpio.Libgpiod.V2;
+using System.IO;
+using System.Linq;
+
+namespace System.Device.Gpio.Drivers;
+
+/// <summary>
+/// Driver factory for different versions of libgpiod.
+/// </summary>
+internal sealed class LibGpiodDriverFactory
```
krwq, line 16: Let's see if we can find some existing pattern for searching for right `.so`

Documentation/gpio-linux-libgpiod.md
```diff
@@ -0,0 +1,109 @@
...
+1. constructor of LibGpiodDriver:
+
+   ```c#
+   new LibGpiodDriver(chipNumber, LibGpiodDriverVersion.V1)
+   ```
+
+2. Environment variable:
+
+   ```shell
+   export DOTNET_IOT_LIBGPIOD_DRIVER_VERSION=V1 // or V2...
```
krwq, line 65: might be a good idea to mention the env var somewhere in the ctor xml doc

krwq (approved):
LGTM as long as we open P0 issue to improve .so searching logic

huesla:
> Great work and great collaboration in this PR. I've been following the different threads and progress.

Thanks to all for the collaboration aswell. If there is anything else to improve, let me know

pgrawehr:
@huesla We had a call on Thursday and decided to merge this now and then address some of the remaining issues separately. The two points that were still disputed was the library search algorithm and the best way to expose this new functionality. 

For this, I have created #2271 and #2272.

In no way, this should be criticism on your work, this is just trying to allign it with large scale standards.

PR #2240, opened 2023-12-16
Compiler multi threading support

pgrawehr:
Various updates to the Arduino Compiler:
- Supports latest .NET 6.0 SDK/Runtime
- Supports multi-threading (tasks, threads, including async calls)
- Added many low-level functions, in particular for file I/O
- Improved memory management

Known issues:
- The new version appears to be quite a bit slower than the old one, reason is unclear
- Use of PSRam isn't working properly yet

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2240)


Ellerbach (approved):
A lot of work placed here and not easy to review! All up, it seems good to me.

PR #2257, opened 2023-12-26
Update OpenHardwareMonitorHttp.cs (Fixes a Edge case where sensor value = NaN)

TheMaskedMan00:
Fix error on sensor value NaN

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2257)

### DIFF
--- a/src/devices/HardwareMonitor/OpenHardwareMonitorHttp.cs
+++ b/src/devices/HardwareMonitor/OpenHardwareMonitorHttp.cs
@@ -96,7 +96,7 @@ namespace Iot.Device.HardwareMonitor
             try
             {
                 string fullJson = _httpClient.GetStringAsync(_uri + "api/rootnode").Result;
-                fullJson = fullJson.Replace("\"Value\": NaN", "\"Value\": 0"); //this fixes an error I encounter where a sensors value is NaN causing the entire thing to crash.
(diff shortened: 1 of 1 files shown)


pgrawehr:
Thanks for the fix. Can you give a stack trace of the crash? I just want to make sure we fix it in the right place.

src/devices/HardwareMonitor/OpenHardwareMonitorHttp.cs
```diff
@@ -96,6 +96,7 @@ private void GetFullDataAndDecode(bool refreshSensorList)
             try
             {
                 string fullJson = _httpClient.GetStringAsync(_uri + "api/rootnode").Result;
+                fullJson = fullJson.Replace("\"Value\": NaN", "\"Value\": 0"); //this fixes an error I encounter where a sensors value is NaN causing the entire thing to crash.
```
pgrawehr, line 99:
```suggestion
                fullJson = fullJson.Replace("\"Value\": NaN", "\"Value\": 0"); // this fixes an error I encounter where a sensors value is NaN causing the entire thing to crash.
```

pgrawehr (approved):

PR #2270, opened 2024-01-23
fix exception when disposing Ft232h device with SPI manager initialized

jb1685:
**Summary of problem:**
When I initialize the Ft232h as a SPI device and then dispose of it, an InvalidOperationException error is thrown because we are disposing of the same elements from the list that we are iterating over.

**Resolution:**
Make a copy of the "_managers" list by using "ToList" so that the iteration continues as we dispose from the target list

**Example:**
image
Path: /Users/joshbender/DocumentsLocal/Github/iot/src/devices/Ft232H/samples/Program.cs<br/>

image
<br/>Path: /Users/joshbender/DocumentsLocal/Github/iot/src/devices/Board/Board.cs<br/>

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/2270)

### DIFF
--- a/src/devices/Board/Board.cs
+++ b/src/devices/Board/Board.cs
@@ -248,7 +248,7 @@ namespace Iot.Device.Board
                     bus.Value.Dispose();
                 }
 
-                foreach (var bus in _managers)
+                foreach (var bus in _managers.ToList())
                 {
                     bus.Dispose();
                 }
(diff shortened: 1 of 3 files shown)


jb1685:
@dotnet-policy-service agree

src/devices/Board/Board.cs
```diff
@@ -248,7 +248,7 @@ protected virtual void Dispose(bool disposing)
                     bus.Value.Dispose();
                 }
 
-                foreach (var bus in _managers)
+                foreach (var bus in _managers.ToList())
```
Ellerbach, line 251: I don't think the ToList() is necessary here.

Ellerbach (approved):
Thanks for the fix. Looks good, please check the comment on the ToList, I don't think it's necessary.

PR #2278, opened 2024-02-11
Reorganize dependencies and infrastructure for ACS

pgrawehr:
Make sure the release build only uses the official package, so that it can be used in client applications that don't build Iot.Device.Common from scratch. Due to the way the compiler works, the client application must be built against the exactly same library the compiler itself is built. 

Also updates some test settings
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2278)


.github/workflows/arduino.yml
```diff
@@ -17,12 +17,16 @@ jobs:
   build:
     # The type of runner that the job will run on
     runs-on: windows-latest
+    strategy:
```
joperezr, line 20: I was under the impression that this workflow was conditioned to only run on PRs that touched the relevant directories (Arduino, or ArduinoCsCompiler). Can we change it so that is the case?

.gitignore
```diff
@@ -57,7 +57,6 @@ BenchmarkDotNet.Artifacts/
 project.lock.json
 project.fragment.lock.json
 artifacts/
-**/Properties/launchSettings.json
```
joperezr, line 60: We could consider keeping this here, and just forcibly adding the launchSettings.json file you are adding which will track it anyway.

build.proj
```diff
@@ -36,6 +36,7 @@
       <_ProjectsToPackage Include="$(MSBuildThisFileDirectory)src\Iot.Device.Bindings\Iot.Device.Bindings.csproj" />
       <_ProjectsToPackage Include="$(MSBuildThisFileDirectory)src\System.Device.Gpio\System.Device.Gpio.csproj" />
       <_ProjectsToPackage Include="$(MSBuildThisFileDirectory)src\Iot.Device.Bindings.SkiaSharpAdapter\Iot.Device.Bindings.SkiaSharpAdapter.csproj" />
+      <_ProjectsToPackage Include="$(MSBuildThisFileDirectory)tools\ArduinoCsCompiler\Frontend\Frontend.csproj"/>
```
joperezr, line 39: This will make it so that this tool is marked as shipping, signed, etc. Do we actually need to produce the package from here? Given this tool is more of a "tool that we have inside our repo for your benefit" kind of thing, I'd rather have the scripts call pack into the project directly.

joperezr: I have nothing against marking the tool as isPackable, but we don't really have a good mechanism today for shipping tools that are just utilities and not really supported from our side, so I would suggest that for having folks using the tool, we keep the installation vehicle to be "run a script which will build the project, pack it, and install it for you so you can use it" kind of thing.

.github/workflows/arduino.yml
pgrawehr: That is handled within the build - when those parts have not changed, only a limited number of tests are run. Only if they did change, the full test suite runs.

.gitignore
pgrawehr: I wasn't sure whether this file should be generally commited, but I guess you're right.

pgrawehr:
@joperezr Added extra notices as discussed.

.github/workflows/arduino.yml
```diff
@@ -17,12 +17,16 @@ jobs:
...
+        matrix:
+            configuration: [Debug, Release]
```
joperezr, line 22: Do we really need the Debug configuration here? Can't we just use the Release one?

tools/ArduinoCsCompiler/Frontend/Frontend.csproj
```diff
@@ -9,20 +9,38 @@
...
     <RootNamespace>ArduinoCsCompiler</RootNamespace>
     <Product>Arduino CS Compiler</Product>
-    <Description>A translation compiler that retargets C# applications so they can run on a microcontroller using the unnamed firmware.</Description>
+    <Description>A translation compiler that retargets C# applications so they can run on a microcontroller using a special firmware. See Readme for further information.</Description>
     <Authors>The .NET Foundation community</Authors>
     <NeutralLanguage>en</NeutralLanguage>
+    <IsPackable>True</IsPackable>
+    <PackAsTool>True</PackAsTool>
+    <ToolCommandName>dotnet-acs</ToolCommandName>
+    <PackageOutputPath>..\..\..\artifacts\packages\$(Configuration)\Shipping</PackageOutputPath>
```
joperezr, line 18: Why do you need to set this? This should be done automatically by Arcade.

tools/ArduinoCsCompiler/Frontend/Frontend.csproj
```diff
@@ -9,20 +9,38 @@
...
+  </ItemGroup>
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.9.1" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\..\..\src\devices\Arduino\Arduino.csproj" />
-    <ProjectReference Include="..\..\..\src\devices\Arduino\samples\ApiChecker\Arduino.sample.csproj" />
     <ProjectReference Include="..\ArduinoCsCompiler.csproj" />
   </ItemGroup>
+  <ItemGroup Condition="$(Configuration)=='Debug'">
```
joperezr, line 36: Can you add a comment here to explain the distinction of why Debug and Release are different?

tools/ArduinoCsCompiler/Runtime/MiniInterop.Kernel32.cs
```diff
@@ -608,6 +626,68 @@ public static UInt32 GetTimeZoneInformation(out TIME_ZONE_INFORMATION lpTimeZone
...
+                    idx++;
+                }
+
+                return bytesToConvert;
+            }
+
+            [ArduinoImplementation]
+            internal static int GetLeadByteRanges(int codePage, byte[] leadByteRanges)
+            {
+                /*
```
joperezr, line 674: Is it intentional that this method is all commented out?

tools/ArduinoCsCompiler/samples/BlinkingLed/BlinkingLed.csproj
```diff
@@ -12,12 +12,15 @@
     <Compile Include="Program.cs" />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="$(Configuration)=='Debug'">
```
joperezr, line 15: I don't fully follow here why we use different P2Ps depending on the build config. Why can't we always just reference iot.device.bindings?

tools/ArduinoCsCompiler/Frontend/Frontend.csproj
```diff
@@ -9,20 +9,38 @@
     <AssemblyName>acs</AssemblyName>
     <RootNamespace>ArduinoCsCompiler</RootNamespace>
     <Product>Arduino CS Compiler</Product>
-    <Description>A translation compiler that retargets C# applications so they can run on a microcontroller using the unnamed firmware.</Description>
+    <Description>A translation compiler that retargets C# applications so they can run on a microcontroller using a special firmware. See Readme for further information.</Description>
     <Authors>The .NET Foundation community</Authors>
     <NeutralLanguage>en</NeutralLanguage>
+    <IsPackable>True</IsPackable>
+    <PackAsTool>True</PackAsTool>
```
joperezr, line 16: We should also make sure that this tool stays in preview and doesn't go stable (since it is experimental). You can accomplish that by adding the property: `<SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>` in the .csproj. This way, even when the rest of the repo builds stable packages, this tool will remain as prerelease.

joperezr (commented):
Left a few comments/questions, but looks good overall.

build.proj
pgrawehr: Already discussed: It is required that the tool links to the officially released versions of Iot.Device.Bindings and System.Device.Bindings, otherwise the reference handling during compilation will not work.

.github/workflows/arduino.yml
pgrawehr: I guess so.

tools/ArduinoCsCompiler/Frontend/Frontend.csproj
pgrawehr: I don't know why, but it doesn't work otherwise (the package ends up in the local project's output folder)

tools/ArduinoCsCompiler/Runtime/MiniInterop.Kernel32.cs
pgrawehr: Yes. I often take the original implementation and then simplify it so that it meets the minimal requirements for our purpose. The MCU runtime doesn't support changing codepages.

tools/ArduinoCsCompiler/samples/BlinkingLed/BlinkingLed.csproj
pgrawehr: Because the dependencies get really ugly for debugging and development in that case, as you would need to either link against the officially released package (which does not include the changes one might want to test) or rebuild the full package for every build. I've tried it, but it was impossible to work with.

tools/ArduinoCsCompiler/Frontend/Frontend.csproj
pgrawehr: I did that, but it doesn't seem to do anything. Can you verify with a release build?

joperezr (approved):

joperezr:
Ok, so I now understand why the build is still producing stable packages even when you applied the property, looks like all projects under tools (and turns out also the ones under devices) aren't using Arcade at all, and I'm working on some changes on top of your PR to fix that but that is unblocking a can of worms unfortunately. I'll be able to push some changes soon, but we may want to reconsider holding the release for this as this would introduce risk as I'm essentially changing how all of these things are built.

joperezr:
cc @pgrawehr ^^

tools/ArduinoCsCompiler/Frontend/Frontend.csproj
```diff
@@ -9,20 +9,44 @@
     <AssemblyName>acs</AssemblyName>
```
joperezr, line 9: I just noticed this project is targeting net6.0, why is that? Should it be net8.0 instead? `net6.0` is out of support so we'd want to update that now.

tools/ArduinoCsCompiler/Frontend/Frontend.csproj
```diff
@@ -9,20 +9,44 @@
...
-    <Description>A translation compiler that retargets C# applications so they can run on a microcontroller using the unnamed firmware.</Description>
+    <Description>A translation compiler that retargets C# applications so they can run on a microcontroller using a special firmware. See Readme for further information.</Description>
     <Authors>The .NET Foundation community</Authors>
     <NeutralLanguage>en</NeutralLanguage>
+    <IsPackable>True</IsPackable>
+    <PackAsTool>True</PackAsTool>
+    <ToolCommandName>dotnet-acs</ToolCommandName>
+    <PackageOutputPath>..\..\..\artifacts\packages\$(Configuration)\Shipping</PackageOutputPath>
+    <Version>1.0.2</Version>
+    <!-- Stay in beta mode -->
```
joperezr, line 20: I used the version you were using, but is there a reason why we are shipping 1.0.2 if we have never shipped a stable 1.0.0? Should we move this back to 1.0.0?

joperezr:
Ok, so did a whole bunch of changes in 361fdf7 to fix tools and devices directories not using Arcade to build. From that, I started getting a bunch of style and analyzer errors, so that's all of the changes in the *.cs files, mainly addressing those.

@pgrawehr could you pull my changes and build locally and validate everything looks good to you? If you want to do a local "Official stable build" you can run `build.cmd -pack -ci /p:OfficialBuildId=20240613.99 /p:DotnetFinalVersionKind=Release` which should generate stable versions of packages, except for acs package, that should stay unstable.

tools/ArduinoCsCompiler/Frontend/Frontend.csproj
pgrawehr: The entire library still targets 6.0, so this is consistent. But more importantly is the fact that the runtime used affects the compilation. Because acs relies heavily on the internal runtime implementations (eg the native methods called) updating this requires considerable effort and testing. 

So this is something for a future update (I'm looking into a slightly different approach there, though)

pgrawehr:
@joperezr Not good. The build now fails with

```
D:\a\1\s\.dotnet\sdk\7.0.403\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(221,5): error NU5017: Cannot create a package that has no dependencies nor content. [D:\a\1\s\tools\ArduinoCsCompiler\Frontend\Frontend.csproj]
```
That package shouldn't be empty...

joperezr:
Oh interesting, not sure why I didn't see that locally. That is because it is trying to create a symbol package (as that is the default on the repo) and this is a tool package. My latest commit should fix that.

joperezr:
I do still think we should retarget the tool and have it target 8.0

joperezr:
Not sure why the GH action is failing though, I think for that one it's best if you take a peek.

pgrawehr:
> Not sure why the GH action is failing though, I think for that one it's best if you take a peek.

That happened because the location of the `bin` folder has changed. Previously, the bin folders where below the projects, now they're all below the `artifacts` folder. Thus the build script was not able to find the binaries to execute.

pgrawehr:
> I do still think we should retarget the tool and have it target 8.0

Have you read [my comment](https://github.com/dotnet/iot/pull/2278#discussion_r1638813759) above? I quickly tried, and it doesn't even build, because the entire project still builds using .NET 7.0 (see global.json). I really wouldn't do such an update as part of this PR.

joperezr:
Oh missed that. Yeah I agree, let's get this merged then and I can put up a follow up PR that moves us to 8.0 infra

pgrawehr:
@joperezr Can you do the merge? I cannot, because the helix build keep failing. We really need to do something about this unstable ADC test.

joperezr:
We should just disable that test as it is not good to setup a pattern to merge on red.

PR #2287, opened 2024-02-29
Revert "Add libgpiodv2 support (#2184)"

pgrawehr:
This reverts commit 3fb06eb5fefbea23306cd85d67bfcff7b987065b. Could be the reason for recent test failures

update: 1/1 green so far https://github.com/dotnet/iot/pull/2287#issuecomment-1971740169

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2287)


pgrawehr:
Marked DNM, as this is just as a test whether the change causes the increased test failure rate.

krwq:
green before rerun

huesla:
@krwq I try find the issue, but I need a similar setup. What hardware/OS does the CI run tests on?

huesla:
And is it always the SysFsDriver tests? How did tests behave before this PR, 100% green?

pgrawehr:
@huesla The tests where not 100% green, but maybe 90%, now they're at 50% or less. 
It seems to mostly be the sysfs driver, but that could be a coincidence. I suspect that there's a situation where now a pin or GpioController is not properly disposed, which is known to cause problems when switching between sysfs and libgpiod. (see eg #1992) The error can also happen if a previous test run exited unexpectedly, as the devices are not rebooted between tests. 

The hardware in use are Raspberry Pi 3B's with a custom test print attached. The schematic can be found [here.](https://github.com/dotnet/iot/tree/main/src/System.Device.Gpio.Tests/schematics). For the GPIO tests, you should only need the connection between pins 5 and 6. I haven't done that in a while, but running the test suite should be as easy as running

```
dotnet test src\System.Device.Gpio\System.Device.Gpio.sln
```

on the Pi.

huesla:
@pgrawehr @krwq I think I found the issue, I have created #2291 that should fix it. I tested it also on a RPI3.

pgrawehr:
Closing this, as issue seems to be resolved with #2291

PR #2291, opened 2024-03-09
Do not pass static zero IntPtr to epoll wait. Dispose driver in tests…

huesla:
See [here](https://github.com/dotnet/iot/pull/2287)
This PR fixes a bug introduced by the merge of #2184.
The bug was introduced by an unintentional change in the behaviour of the SysFsDriver, that utilizes the native epoll_wait function to discard not needed events.
Originally the binding of `epoll_wait` was implemented as
`public static extern int epoll_wait(int epfd, out epoll_event events, int maxevents, int timeout);`
which does not fully match the native function
`int epoll_wait(int epfd, struct epoll_event *events, int maxevents, int timeout);`
because `events` is a pointer to an array.

Some new functionality in #2184 relies on epoll_wait, what prompted me to correct the binding.
As a consequence I had to update calls in the SysFsDriver, but mistakenly passed IntPtr.Zero to epoll_wait, which does not seem to result in the same behaviour. The fix is to pass an actual buffer instead.

This PR also includes some additional "using" statements in the GpioControllerTestBase because during testing, some sporadic occurrences of PermissionDeniedException occured. It complained about not having root permission for changing GPIO mode of used GPIOs. I'm pretty sure that was because of not yet disposed SysFsDriver instance/s.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2291)

### DIFF
--- a/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
+++ b/src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
@@ -28,7 +28,7 @@ public abstract class GpioControllerTestBase
     [InlineData(false)]
     public void PinValueStaysSameAfterDispose(bool closeAsHigh)
     {
-        var driver = GetTestDriver();
+        using var driver = GetTestDriver();
         if (driver is SysFsDriver)
         {
             // This check fails on the SysFsDriver, because it always sets the value to 0 when the pin is opened (but on close, the value does stay high)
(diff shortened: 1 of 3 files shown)


pgrawehr (approved):
Thanks for this fix and the quick reaction. That looks much greener now 🟢

Ellerbach (approved):
Thanks for the fix, indeed, seems to fix the issues :-)

PR #2305, opened 2024-04-18
Use development branch, so we get the ESP32 IDK 3.X

pgrawehr:
The build was still using the 2.X line, which is now incompatible with the firmware
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2305)


Ellerbach (approved):
Thanks for the fix

PR #2309, opened 2024-05-01
Add LPS22HB device

famblard86:
Adding LPS22HB based on existing LPS25H 

https://www.st.com/resource/en/datasheet/lps22hb.pdf

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2309)

### DIFF
--- a/src/devices/Lps22hb/Lps22hb.cs
+++ b/src/devices/Lps22hb/Lps22hb.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Buffers.Binary;
+using System.Device;
 using System.Device.I2c;
 using System.Device.Model;
 using System.Diagnostics;
@@ -96,7 +97,8 @@ namespace Iot.Device.Lps22hb
         public void ResetDevice()
         {
             WriteByte(Register.CTRL_REG2, 0b0100);
-            while (true)
+            int timeoutLoop = 10;
+            while (timeoutLoop > 0)
             {
                 byte reset = Read(Register.CTRL_REG2);
 
@@ -104,6 +106,9 @@ namespace Iot.Device.Lps22hb
                 {
                     break;
                 }
+
+                DelayHelper.DelayMilliseconds(10, true);
+                timeoutLoop -= 1;
             }
         }
 
--- a/src/devices/Lps22hb/Lps22hb.cs
+++ b/src/devices/Lps22hb/Lps22hb.cs
@@ -44,7 +44,7 @@ namespace Iot.Device.Lps22hb
 
             if (!TryRead())
             {
-                throw new Exception("Device not found");
+                throw new IOException("Lps22hb not found");
             }
 
             if (outputRate == OutputRate.PowerDownMode)
--- a/src/devices/Lps22hb/Lps22hb.cs
+++ b/src/devices/Lps22hb/Lps22hb.cs
@@ -52,6 +52,8 @@ namespace Iot.Device.Lps22hb
                 throw new NotImplementedException("One shot mode is not supported");
             }
 
+            ResetDevice();
+
             // Hight resolution with Normal mode = 0 p43
             byte resolution = Read(Register.RES_CONF);
             resolution &= 0b10;
@@ -89,10 +91,21 @@ namespace Iot.Device.Lps22hb
         }
 
         /// <summary>
-        /// Temperature
+        /// Reset the device
         /// </summary>
-        [Telemetry]
-        public Temperature Temperature => Temperature.FromDegreesCelsius(ReadInt16(Register.TEMP_OUT_L) / 100f);
+        public void ResetDevice()
+        {
+            WriteByte(Register.CTRL_REG2, 0b0100);
+            while (true)
+            {
+                byte reset = Read(Register.CTRL_REG2);
+
+                if ((reset & 0x04) == 0)
+                {
+                    break;
+                }
+            }
+        }
 
         /// <summary>
         /// Read temperature. A return value indicates whether the reading succeeded.
--- a/src/devices/Lps22hb/Lps22hb.cs
+++ b/src/devices/Lps22hb/Lps22hb.cs
@@ -1,0 +1,144 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Buffers.Binary;
+using System.Device.I2c;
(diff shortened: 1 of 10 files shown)


src/devices/Lps22hb/README.md
```diff
@@ -0,0 +1,12 @@
...
+Some of the applications mentioned by the datasheet:
+
+- Altimeter and barometer for portable devices
+- GPS applications
+- Weather station equipment
+- Sport watches
+
+## References
+
+- <https://www.st.com/resource/en/datasheet/lps22hb.pdf>
```
Ellerbach, line 12: can you add a quick usage in the read me? It' s a great practice even if there is the sample.
Also anything not implemented in this binding? You could list what is and is not implemented.

Ellerbach, line 12:
```suggestion
- [LPS22HB datasheet](https://www.st.com/resource/en/datasheet/lps22hb.pdf)
```

src/devices/Lps22hb/Register.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </summary>
+        CTRL_REG2 = 0x11,
+
+        /// <summary>
+        /// Control3
+        /// </summary>
+        CTRL_REG3 = 0x12,
+
+        /// <summary>
+        ///  FIFO configuration register
```
Ellerbach, line 40:
```suggestion
        /// FIFO configuration register
```

src/devices/Lps22hb/Register.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </summary>
+        RPDS_L = 0x18,
+
+        /// <summary>
+        /// Resolution register
+        /// </summary>
+        RES_CONF = 0x1A,
+
+        /// <summary>
+        ///  Interrupt register
```
Ellerbach, line 62:
```suggestion
        /// Interrupt register
```

src/devices/Lps22hb/Register.cs
```diff
@@ -0,0 +1,93 @@
...
+        PRESS_OUT_XL = 0x28,
+
+        /// <summary>
+        /// 16-bit Temperature output registers
+        /// TEMP_OUT_L (0x2B), TEMP_OUT_H (0x2C)
+        /// </summary>
+        TEMP_OUT_L = 0x2B,
+
+        /// <summary>
+        ///  Filter reset register
```
Ellerbach, line 89:
```suggestion
        /// Filter reset register
```

Ellerbach (commented):
Looks great! Thanks for this addition.

src/devices/Lps22hb/Lps22hb.cs
```diff
@@ -0,0 +1,144 @@
...
+            if (i2cDevice == null)
+            {
+                throw new ArgumentNullException(nameof(i2cDevice));
+            }
+
+            _i2c = i2cDevice;
+
+            if (!TryRead())
+            {
+                throw new Exception("Device not found");
```
pgrawehr, line 43: Do not throw this generic exception type. Use IOException or similar. In this case, the exception can only happen if there is a device on the provided address but not the expected one, as i2cDevice.Read() in itself throws an exception if there's no answer from the bus.

src/devices/Lps22hb/Lps22hb.cs
```diff
@@ -0,0 +1,144 @@
...
+            };
+
+            return BinaryPrimitives.ReadUInt32LittleEndian(b);
+        }
+
+        /// <summary>
+        /// Temperature
+        /// </summary>
+        [Telemetry]
+        public Temperature Temperature => Temperature.FromDegreesCelsius(ReadInt16(Register.TEMP_OUT_L) / 100f);
```
pgrawehr, line 81: Consider using `TryReadTemperature()`/`TryReadPressure()` methods instead of these properties. The call can fail. Also, using properties with direct access to the hardware may lead to debugger instability, which is at least a nuissance.

src/devices/Lps22hb/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+## Features bindings
+
+- [x] Interface I2c
+- [x] Read pressure (24 bit)
+- [x] Read temperature (16 bit)
+- [x] Continus mode (25Hz)
+- [x] BDU
+- [x] Enable Low-pass filter
```
pgrawehr, line 42: Looks a bit too good to me, when these items are checked but none of them are really available in the API as far as I can tell. What is BDU anyway?

pgrawehr (commented):
Looks mostly good. Some suggestions.

src/devices/Lps22hb/README.md
famblard86: I added more available configuration.
_The BDU (Block Data Update) is used to inhibit the update of the output registers between the reading of upper, middle and lower register parts_
p20 => https://www.st.com/resource/en/application_note/an4833-measuring-pressure-data-from-sts-lps22hb-digital-pressure-sensor-stmicroelectronics.pdf

pgrawehr (approved):

src/devices/Lps22hb/Lps22hb.cs
```diff
@@ -0,0 +1,207 @@
...
+            return BinaryPrimitives.ReadUInt32LittleEndian(b);
+        }
+
+        /// <summary>
+        /// Reset the device
+        /// </summary>
+        public void ResetDevice()
+        {
+            WriteByte(Register.CTRL_REG2, 0b0100);
+            while (true)
```
Ellerbach, line 99: We should add a timeout otherwise, this is never ending or a cancellation token with a default value or a pattern with a number of addend and a delay, see https://github.com/dotnet/iot/blob/8460898df2061230a5d45832a56a14d731d117e5/src/System.Device.Gpio.Tests/TimeoutHelper.cs#L8
or even something like this: https://github.com/dotnet/iot/blob/8460898df2061230a5d45832a56a14d731d117e5/src/System.Device.Gpio.Tests/TimeoutHelper.cs#L8
Somehow your choice but you definitely need to addess the infinite loop problem.

famblard86: I arbitrarily added 100ms max, I've never seen a single loop I think it's enough.
Pipeline looks broken?

Ellerbach (approved):
Thanks all good and now build is repaired, we are go to merge!

PR #2311, opened 2024-05-05
Fix access violation issue in libgpiod v1 driver

### LINKED ISSUE #1849: Segfault in finalization with LibGpiodDriver

pgrawehr:
Fixes #1849 

After the cause is finally verified, this removes the SafeHandle base class from SafeLineHandle and changes it to a normal class. The line(s) are always disposed together with their chips, and thus do not need to be explicitly disposed. Lines (individual pins) must not be disposed after the chip is disposed. 

This also adds various tests to make sure exceptions are thrown when attempting to use disposed drivers.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2311)


src/System.Device.Gpio/System/Device/Gpio/Drivers/Libgpiod/LibGpiodDriver.cs
```diff
@@ -144,7 +144,7 @@ public override ComponentInformation QueryComponentInformation()
     /// <inheritdoc/>
     protected override void Dispose(bool disposing)
     {
-        _driver.Dispose();
+        _driver?.Dispose();
```
krwq, line 147: should you null `_driver`

src/System.Device.Gpio.Tests/GpioControllerTestBase.cs
```diff
@@ -595,6 +595,34 @@ void Callback(object sender, PinValueChangedEventArgs pinValueChangedEventArgs)
...
+    {
+        var controller = new GpioController(PinNumberingScheme.Logical, GetTestDriver());
+        var pin6 = controller.OpenPin(InputPin, PinMode.Input);
+        controller.Dispose();
+        bool correctExceptionSeen = false;
+        try
+        {
+            pin6.Read();
+        }
+        catch (Exception x) when (x is InvalidOperationException || x is ObjectDisposedException)
```
krwq, line 609: we should unify to not throw InvalidOperationException

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -233,6 +233,11 @@ public virtual PinMode GetPinMode(int pinNumber)
     /// <returns>The status if the pin is open or closed.</returns>
     public bool IsPinOpen(int pinNumber)
     {
+        if (_driver == null)
```
joperezr, line 236: Why are we only making this check here as opposed to everywhere else in this class? For instance, it's not in IsPinModeSupported and that is accessing the driver

src/System.Device.Gpio/Interop/Unix/libgpiod/V1/SafeLineHandle.cs
```diff
@@ -7,31 +7,53 @@
 namespace System.Device.Gpio.Libgpiod.V1;
 
 /// <summary>
-/// Pointer to a pin.
+/// Pointer to a pin (Not a real SafeLineHandle, because we need to align its finalization with the owning object)
 /// </summary>
-internal class SafeLineHandle : SafeHandle
+internal sealed class SafeLineHandle : IDisposable
```
joperezr, line 12: NIT: Given this is no longer an actual SafeHandle, we should likely rename this to something like `LineHandle` or something.

src/System.Device.Gpio/Interop/Unix/libgpiod/V1/SafeLineHandle.cs
```diff
@@ -7,31 +7,53 @@
...
 
     /// <summary>
-    /// Release the lock on the line handle. <see cref="LibgpiodV1.gpiod_line_release"/>
+    /// Release the lock on the line handle. <see cref="Interop.LibgpiodV1.gpiod_line_release"/>
     /// </summary>
     public void ReleaseLock()
     {
-        ReleaseHandle();
+        // Contrary to intuition, this does not invalidate the handle (see comment on declaration)
+        Interop.LibgpiodV1.gpiod_line_release(_handle);
```
joperezr, line 46: Assuming this will make it so that `IsInvalid` method will return true? Does it set the handle to IntPtr.Zero or to InvalidHandleValue?

src/System.Device.Gpio/Interop/Unix/libgpiod/V1/SafeLineHandle.cs
```diff
@@ -7,31 +7,53 @@
...
     }
 
-    public override bool IsInvalid => handle == IntPtr.Zero || handle == LibgpiodV1.InvalidHandleValue;
+    public bool IsInvalid => _handle == IntPtr.Zero || _handle == Interop.LibgpiodV1.InvalidHandleValue;
+
+    public void Dispose()
+    {
+        if (_handle != IntPtr.Zero)
+        {
+            Interop.LibgpiodV1.gpiod_line_release(_handle);
```
joperezr, line 55: Whould this call ReleaseLock instead? And should ReleaseLock also set `_handle` to IntPtr.Zero?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
joperezr: If it is because we expect all other method's to call this, we should probably add Debug.Asserts in all of the rest of the methods that ensure that driver is in fact not null after calling IsPinOpen.

src/System.Device.Gpio/CompatibilitySuppressions.xml
```diff
@@ -13,6 +13,12 @@
     <Left>lib/netstandard2.0/System.Device.Gpio.dll</Left>
     <Right>lib/net6.0-windows10.0.17763/System.Device.Gpio.dll</Right>
   </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
```
joperezr, line 17: I wouldn't expect this to show up as we should have the finalizer in both places. Can you please check why this is needed?

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
pgrawehr: Yep, overlooked the IsPinModeSupported method. Fixed now.

src/System.Device.Gpio/Interop/Unix/libgpiod/V1/SafeLineHandle.cs
pgrawehr: No, see comment. This just internally decreases some kind of reference counter and does not invalidate the handle. IIRC this is needed to change the mode of a pin without closing it.

src/System.Device.Gpio/Interop/Unix/libgpiod/V1/SafeLineHandle.cs
pgrawehr: They're separate because they serve different purposes. Here, we really want to close the line (from an user point of view), the other method is a special case because gpiod_line_release needs to be called before a change of the pin mode is possible. Technically, the handle is never invalidated until the owning chip is closed.

src/System.Device.Gpio/CompatibilitySuppressions.xml
pgrawehr: I have no clue, but looks like a bug of the tool to me. We have _added_ a finalizer, so why should there be a warning that it's now missing? I can't see why this would be a breaking change, anyway.

pgrawehr: I verified the generated IL. The finalize method is also present in the netstandard2.0 assembly.

pgrawehr:
@joperezr I unfortunately have no idea why we get that report about the missing finalizer. It is there in all places.

krwq (approved):

PR #2316, opened 2024-05-20
Add SSD1309 Device using non-traditional SPI

AdamJSchofield:
Adds support for SSD1309 displays such as [this one](https://www.waveshare.com/1.51inch-transparent-oled.htm). Datasheet for the SSD1309 available [here](https://www.crystalfontz.com/controllers/datasheet-viewer.php?id=342).

These transparent displays have many potential applications but the coolest is mini volumetric displays like [this one](https://www.youtube.com/watch?v=sVv1oc14X1w) (not my video).

Implementation Notes:

- This display supports I2C and SPI. The SPI implementation suggested here uses additional GPIO pins for DC (Data/Command) and RST (Reset) as described in the datasheet.
- When the DC pin is pulled LOW, data read into the command buffer of the driver. When the DC pin is pulled HIGH, data will be written to display memory.
- Cycling the RST pin from HIGH -> LOW -> HIGH will reset the display.
- A new `PreRenderBitmap` method was added to the base `Ssd13xx` class. This is useful because it may not always be desired to send frames to the display immediately. An example of this is the simulation I included which uses one thread to buffer/render display-ready frames and another to send them to the display at a fixed rate. Obviously this only works for multi-core boards.
   
Sample Notes:
Included is a sample simulation for falling sand. It looks pretty cool on the transparent OLED I tested with. My 4-core RaspberryPi Zero 2 was able to maintain 45 FPS without dropping frames. Rendering to a SkiaSharp bitmap is the bottleneck. Much faster speeds are possible with a large buffer or by using a 2D byte-array as the render state instead of drawing to a BitmapImage.

If desired, a `PreRenderBitmap` method could be added to the `GraphicDisplay` base class allowing for the `Ssd1309Simulation` class to be extended for many other displays.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2316)


AdamJSchofield:
@dotnet-policy-service agree

src/devices/Ssd13xx/Ssd1309.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        /// <summary>
+        /// Send a command to the display controller.
+        /// </summary>
+        /// <param name="command">The command to send to the display controller.</param>
+        private void SendCommand(ICommand command)
+        {
+            Span<byte> commandBytes = command?.GetBytes();
+
+            _gpioController?.Write(_dc, PinValue.Low);
```
AdamJSchofield, line 70: The main difference between this and the existing I2C implementation is that the DC pin is set to LOW for commands and HIGH for data. the SSD1306 with I2C uses a control byte here.

joperezr:
@AdamJSchofield I noticed you marked this as work in progress and it is still a draft. Is this ready for review though? Is there anything we can do to help making it ready for review?

AdamJSchofield:
> @AdamJSchofield I noticed you marked this as work in progress and it is still a draft. Is this ready for review though? Is there anything we can do to help making it ready for review?

Hi! I have since got it working and am putting together a final draft of the PR with a working example in the samples project for RaspberryPi boards. I hit a small snag last night but I expect I'll be able to push that up and mark it ready for review in the next few days.

I'm also happy to close this PR/issue and open a new one when its actually ready if that helps you all stay organized. I jumped the gun a bit but I've since worked out most of the kinks. I may do that anyway since the nature of the PR will be more of a contribution with documentation and less of a cry for help 😆

joperezr:
Great to hear you have got it working now! No need to close this PR, we just need to mark it ready for review once you are ready, and then we can jump and take a peek 😃

AdamJSchofield:
Hi @joperezr @pgrawehr - I've marked this as ready for review. Let me know what you think!

src/devices/Ssd13xx/Ssd1309.cs
```diff
@@ -0,0 +1,227 @@
...
+            {
+                throw new ArgumentNullException(nameof(command), "Argument is either null or there were no bytes to send.");
+            }
+
+            // TODO: Is this needed at all if data/command is controlled via GPIO instead of a control byte?
+            Span<byte> writeBuffer = SliceGenericBuffer(commandBytes.Length);
+
+            commandBytes.CopyTo(writeBuffer.Slice(0));
+
+            if (_spiDevice != null)
```
Ellerbach, line 85: _spiDevice from the base should be internal and because it's internal, it should become a property so that it make things easier when it comes to accessing it.

src/devices/Ssd13xx/Ssd1309.cs
```diff
@@ -0,0 +1,227 @@
...
+                throw new InvalidOperationException("No SPI device available or it has been disposed.");
+            }
+        }
+
+        /// <summary>
+        /// Init 128x64
+        /// </summary>
+        protected virtual void Initialize()
+        {
+            SendCommand(new SetDisplayOff());
```
Ellerbach, line 139: possible to get couple of comments on the commands where there are magic numbers? It will make the maintainability a bit better

src/devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -23,10 +24,15 @@ public abstract class Ssd13xx : GraphicDisplay
...
         /// Underlying I2C device
         /// </summary>
-        protected I2cDevice _i2cDevice;
+        protected I2cDevice? _i2cDevice;
 
         /// <summary>
-        /// Constructs instance of Ssd13xx
+        /// Underlying SPI device
+        /// </summary>
+        protected SpiDevice? _spiDevice;
```
Ellerbach, line 32: This one should be Internal and as a property.

src/devices/Ssd13xx/samples/Program.cs
```diff
@@ -2,116 +2,248 @@
...
-/// </summary>
-public class Program
+namespace Ssd13xx.Samples
 {
     /// <summary>
-    /// Entry point
+    /// Demo entry point
     /// </summary>
-    public static int Main(string[] args)
+    public class Program
```
Ellerbach, line 25: rather than adjusting this sample to try to cover all the cases, maybe it's better to have another sample focusing on this new screen?

Ellerbach (commented):
Thanks for the addition! couple of comments in a very quick review

src/devices/Ssd13xx/Ssd1309.cs
AdamJSchofield: Sure, I did not add any new commands but instead used the Ssd1306 commands with the namespace prefix removed here for less confusion. I will go through the commands and make sure they are all documented and relevant to the Ssd1309 though. For example, I don't think the Ssd1309 has a charge pump like the Ssd1306 but I need to confirm.

src/devices/Ssd13xx/samples/Program.cs
AdamJSchofield: I grappled with this a bit but that's definitely the better approach 👍

src/devices/Ssd13xx/Ssd1309.cs
AdamJSchofield: I addressed this by removing the magic numbers and using the default values which have comments describing their meaning

src/devices/Ssd13xx/samples/Program.cs
AdamJSchofield: I restructured the sample projects to be broken out by protocol. The diff is a bit messy as a result but this is the folder structure:
image

src/devices/Ssd13xx/Ssd1309.cs
AdamJSchofield: In the latest commit, I made the I2cDevice internal as well. I'm not a huge fan of the pattern with both devices in the same class, but I plan to add SPI support for more Ssd13xx displays so I will think of some ways to improve that.

pgrawehr (approved):
Just a few questions left. 

Sorry I forgot that I was assigned to this PR. Nice work!

src/devices/Ssd13xx/Ssd1306.cs
```diff
@@ -60,7 +60,7 @@ private void SendCommand(ICommand command)
             // Be aware there is a Continuation Bit in the Control byte and can be used
             // to state (logic LOW) if there is only data bytes to follow.
             // This binding separates commands and data by using SendCommand and SendData.
-            _i2cDevice.Write(writeBuffer);
+            I2cDevice?.Write(writeBuffer);
```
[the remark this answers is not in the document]
pgrawehr: Clarified myself

src/devices/Ssd13xx/Ssd1309.cs
```diff
@@ -0,0 +1,226 @@
...
+
+        /// <summary>
+        /// Reset device by cycling RST pin to LOW and resting at HIGH
+        /// </summary>
+        protected virtual void Reset()
+        {
+            _rstGpioPin.Write(PinValue.Low);
+            Thread.Sleep(100);
+            _rstGpioPin.Write(PinValue.High);
+            Thread.Sleep(100);
```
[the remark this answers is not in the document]
AdamJSchofield: This method was intended to just execute the specific sequence to reset the controller according to the datasheet. But practically, it could be useful to include other reset-related tasks. Calling `Reset` and then `Initialize` separately in a "re-initialization" procedure feels more intuitive to me personally, but I can see advantages to both.

pgrawehr:
@AdamJSchofield Thanks for this contribution

PR #1927, opened 2022-09-18
The call to Dns.GetHostEntry() may fail on MacOS

pgrawehr:
... only ever seen there.

Saw at least one instance of this crash during a test run:
```
Der aktive Testlauf wurde abgebrochen. Grund: Der Testhostprozess ist abgestürzt. : Unhandled exception. System.Net.Internals.SocketExceptionFactory+ExtendedSocketException (00000005, 0xFFFDFFFF): nodename nor servname provided, or not known
   at System.Net.Dns.GetHostEntryOrAddressesCore(String hostName, Boolean justAddresses, AddressFamily addressFamily, ValueStopwatch stopwatch)
   at System.Net.Dns.GetHostEntry(String hostNameOrAddress, AddressFamily family)
   at System.Net.Dns.GetHostEntry(String hostNameOrAddress)
   at Iot.Device.Nmea0183.NmeaUdpServer.UdpClientStream.Read(Byte[] buffer, Int32 offset, Int32 count) in /Users/patrick/projects/iot/src/devices/Nmea0183/NmeaUdpServer.cs:line 258
   at System.IO.StreamReader.ReadBuffer()
   at System.IO.StreamReader.ReadLine()
   at Iot.Device.Nmea0183.NmeaParser.Parser() in /Users/patrick/projects/iot/src/devices/Nmea0183/NmeaParser.cs:line 129
   at System.Threading.Thread.StartCallback()
```

I was not able to reproduce the issue after the fix, however the MacOS build is still very unstable on the build servers.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/iot/pull/1927)

PR #2318, opened 2024-05-25
Fix network port scan for MacOS

pgrawehr:
Started failing without code changes on our end.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2318)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -150,7 +150,7 @@ stages:
     steps:
     - script: ./build.sh --ci
         --configuration $(BuildConfiguration)
-        --prepareMachine
+        --prepareMachine /maxCpuCount:1
       displayName: Build
 
     - task: PublishBuildArtifacts@1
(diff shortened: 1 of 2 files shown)


pgrawehr:
@joperezr Please have a look at this. It fixes the unit test that started to fail on MacOS without any changes on our end. 
Could be a new instance of https://github.com/actions/runner-images/issues/1335 causing #1927 one more.

azure-pipelines.yml
```diff
@@ -150,7 +150,7 @@ stages:
     steps:
     - script: ./build.sh --ci
         --configuration $(BuildConfiguration)
-        --prepareMachine
+        --prepareMachine /maxCpuCount:1
```
joperezr, line 153: Can we add a tracking issue and a comment here to make sure we remember to remove this in the future? 

Also can we make this only set in macos? That way we wont be setting this in distros that don't need that (and that are more important like Linux) where we could be unintentionally hiding race condition issues.

joperezr (approved):
Left a comment but other than that this looks good.

azure-pipelines.yml
joperezr: Nevermind, as you pointed out this is only on macos.

PR #2319, opened 2024-05-31
add support for Mifare Classic with 7-byte UIDs

jdbruner:
Mifare Classic cards can have either 4-byte or 7-byte UIDs (serial numbers). The existing implementation only supports cards with 4-byte UIDs (which are far more common). This PR adds support for cards with 7-byte UIDs.

There are two sets of changes in this PR:
1. Update ```SetCapacity``` to match an ATQA of either 0x0004 (4-byte UID) or 0x0044 (7-byte UID).
2. Change the serialization of authentication commands in ```Serialize``` to work with either length of the ```SerialNumber```. The arguments to the ```AuthenticationA``` and ```AuthenticationB``` commands are the command byte, block number, key (6 bytes), and last four digits of the serial number (which is the entire serial number in the case of a 4-byte UID). This code was also refactored so that the same code is used for both commands (differing only in the key that is used).

Tested with Mifare Classic 1K (4-byte UID), Mifare Classic 1K (7-byte UID), and Mifare Classic 4K (4-byte UID) cards using the PN532 sample code (and a PN532 transceiver :-)).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2319)

### DIFF
--- a/src/devices/Card/Mifare/MifareCard.cs
+++ b/src/devices/Card/Mifare/MifareCard.cs
@@ -584,7 +584,7 @@ namespace Iot.Device.Card.Mifare
             // Type of Mifare can be partially determined by ATQA and SAK
             // https://www.nxp.com/docs/en/application-note/AN10833.pdf
             // Not complete
-            if (ATQA == 0x0004)
+            if (ATQA == 0x0004 || ATQA == 0x0044)
             {
                 if (SAK == 0x08)
                 {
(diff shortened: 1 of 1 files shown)


Ellerbach (approved):
Looks good, thanks for the improvement! And thanks for confirming you've been tested it.

PR #2321, opened 2024-06-04
Fixing the Segmentation Fault in libgpiod V1

### LINKED ISSUE #2313: GPIO input gives segmentation fault on BeagleBone Black in v3.1.0
**Describe the bug**

Opening a pin for input on a BeagleBone Black gives a Segmentation Fault in v3.1.0. The same code works in v3.0.0

**Steps to reproduce**
```csharp
using System.Device.Gpio;

GpioController controller = new();
GpioPin buttonPin = controller.OpenPin(67, PinMode.Input);
```
code compiled with the following and then copied over via SCP:
```
dotnet publish --runtime linux-arm --self-contained
```

**Expected behavior**
An input pin should be opened and value can later be read.

**Actual behavior**
Segmentation fault as the pin is opened. Output pins work fine.

**Versions used**
dotnet version 8.0.204 (also tried 6.0.421)
BeagleBone Debian 11.7 minimal (although seems to happen on other versions too)
System.Device.Gpio version 3.1.0

Reverting System.Device.Gpio to 3.0.0. fixes the issue.

raffaeler:
Fixes #2313 

This PR fixes the Segmentation Fault occuring when a wrong gpioChip is passed to the `libgpiod` driver.
In a Raspberry PI this is typically `0` whichi is the default but on other boards there may be the need to specify a different number or even more than one (different I/Os are mapped on different chips).

The problem was related to the `SafeHandle` being non-null but invalid, therefore any following PInvoke to the native `libgpiod` library resulted in the fault.

After the change, the application calling the driver with a wrong `gpioChip` receive this exception message which can be handled in a `catch` statement.
``` 
System.IO.IOException: Unable to find a chip, error code: 2
```

The `libgpiod` V2 manages the `gpioChip` differently and it should already ok.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2321)


Ellerbach (approved):
Looks good to me! Nice fix :-)

pgrawehr (approved):
We should have thought of this much earlier...

Issue #2271, opened 2024-01-26
Improve library detection algorithm in LibGpiodDriverFactory


pgrawehr:
The current detection algorithm in LibGpidDriver factory is somewhat risky, in that it assumes certain default paths for libraries. A more standard approach would be to just try to use PInvoke and catch the error.

krwq:
[Triage] We want to repeat logic in System.DirectoryServices.Protocols in dotnet/runtime: https://github.com/dotnet/runtime/blob/ff3c5314badbb0952bc86325baf12f2dd6255f6f/src/libraries/Common/src/Interop/Linux/OpenLdap/Interop.Ldap.cs#L70-L99

krwq:
we should also look at https://github.com/dotnet/iot/issues/2272 as part of this issue as that code might not be needed anymore

krwq:
[Triage] We need to change https://github.com/dotnet/iot/blob/main/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs#L507

krwq:
[Triage] this was alsready done

PR #2323, opened 2024-06-06
Marking Libgpiod v2 api as experimental

joperezr:
cc: @raffaeler @Ellerbach @pgrawehr @krwq 

Relates to https://github.com/dotnet/iot/issues/2271
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2323)

### DIFF
--- a/src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/Chip.cs
+++ b/src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/Chip.cs
@@ -2,8 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Generic;
-using System.Runtime.InteropServices;
 using System.Device.Gpio.Drivers;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
 using LibgpiodV2 = Interop.LibgpiodV2;
 
 namespace System.Device.Gpio.Libgpiod.V2;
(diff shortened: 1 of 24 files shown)


src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/Chip.cs
```diff
@@ -13,6 +14,7 @@ namespace System.Device.Gpio.Libgpiod.V2;
 /// allows callers to retrieve information about each line, watch lines for state changes and make line requests.
 /// </summary>
 /// <seealso href="https://libgpiod.readthedocs.io/en/latest/group__chips.html"/>
+[Experimental(DiagnosticIds.SDGPIO0001, UrlFormat = DiagnosticIds.UrlFormat)]
```
pgrawehr, line 17: Why do we have to add this attribute to internal classes? They're not part of the API.

pgrawehr (approved):

Ellerbach (approved):
looks good

src/System.Device.Gpio/Interop/Unix/libgpiod/V2/Proxies/Chip.cs
joperezr: Good question, short answer is that this way we know if any stable API depends on any unstable API.

Long answer: when marking some of the new public APIs as experimental, the internal code we added in libgpiod started to throw warnings as we were calling experimental APIs. There are two ways of getting rid of those warnings, you either suppress them (either in the callsite or just add it to NoWarn so they are suppressed for the whole project) or you mark the caller also as Experimental, which would just bubble the warning one level up the callstack. I chose the latter as otherwise we wouldn't know if any stable public facing API depended on any nom stable public facing API, so I marked each of the classes that showed warnings Experimental all the way up till O reached public API, and only suppressed the warnings there. This way, it requires a conscious decision to depend on experimental stuff from non experimental stuff.

joperezr: For instance, I know that today, the only two existing public APIs that are stable which may depend on any of the new experimental codepaths are UnixDriver.Create and LibGpiod constructor, because those are the only two places where I am suppressing the warnings.

pgrawehr (approved):

PR #2324, opened 2024-06-08
SenseHat LED matrix text rendering model

logicaloud:
- Support text rendering into the LED matrix
- Apply device model to commands and properties

A critical eye would be good to judge whether the applied device interface, command and property attributes are in keeping with the intent of the device model.

### DIFF
--- a/src/devices/SenseHat/SenseHatLedMatrix.cs
+++ b/src/devices/SenseHat/SenseHatLedMatrix.cs
@@ -7,12 +7,14 @@ using System.Device.Gpio;
 using System.Diagnostics;
 using System.Drawing;
 using System.Runtime.CompilerServices;
+using System.Device.Model;
 
 namespace Iot.Device.SenseHat
 {
     /// <summary>
     /// Base class for SenseHAT LED matrix
     /// </summary>
+    [Interface("SenseHat LED Matrix")]
     public abstract class SenseHatLedMatrix : IDisposable
     {
         /// <summary>
@@ -31,6 +33,45 @@ namespace Iot.Device.SenseHat
         /// Number of pixels per column
         /// </summary>
         protected const int NumberOfPixelsPerColumn = 8;
+
+        /// <summary>
+        /// Lazily intialized pixel font reference. Initialized when rendering text.
+        /// </summary>
+        protected SenseHatTextFont? _pixelFont = null;
+
+        /// <summary>
+        /// Lazily initialized text render state. Initialized when rendering text.
+        /// </summary>
+        protected SenseHatTextRenderMatrix? _textRenderMatrix = null;
+
+        /// <summary>
+        /// Text color when rendering text. Null for "no color".
+        /// </summary>
+        protected Color _textColor = Color.DarkBlue;
+
+        /// <summary>
+        /// Text background color when rendering text. Null for "no color".
+        /// </summary>
+        protected Color _textBackgroundColor = Color.Black;
+
+        /// <summary>
+        /// Text rotation, counterclockwise.
+        /// </summary>
+        protected SenseHatTextRotation _textRotation = SenseHatTextRotation.Rotate_0_Degrees;
+
+        /// <summary>
+        /// Text scroll speed when the rendered text does not fit the 8x8 LED matrix
+        /// </summary>
+        protected double _textScrollPixelsPerSecond = 1;
+
+        /// <summary>
+        /// Timer used for text animation (scrolling). Lazily initialized as required.
+        /// </summary>
+        protected System.Timers.Timer? _textAnimationTimer = null;
+
+        // Lock object to prevent i2c disposal during render.
+        // "Terminate" must be called for clean termination of text animation.
+        private object _lockTextRender = new();
 
         /// <summary>
         /// Constructs SenseHatLedMatrix instance
@@ -81,12 +122,14 @@ namespace Iot.Device.SenseHat
         /// Write colors to the device
         /// </summary>
         /// <param name="colors">Array of colors</param>
+        [Command]
         public abstract void Write(ReadOnlySpan<Color> colors);
 
         /// <summary>
         /// Fill LED matrix with a specific color
         /// </summary>
         /// <param name="color">Color to fill the device with</param>
+        [Command]
         public abstract void Fill(Color color = default(Color));
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


src/devices/SenseHat/SenseHatTextRotation.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.SenseHat
+{
+    /// <summary>
+    /// Text rotation when rendering text, counterclockwise.
```
Ellerbach, line 7: on the counterclockwise : you have degrees in the enum, so not sure how this comments align

src/devices/SenseHat/SenseHatLedMatrix.cs
```diff
@@ -95,9 +138,216 @@ public static int PositionToIndex(int x, int y)
...
+            if (renderMatrix != null && renderMatrix.Text.Length > 1)
+            {
+                renderMatrix.ScrollByOnePixel();
+                RenderText();
+            }
+        }
+
+        private void RenderText()
+        {
+            lock (_lockTextRender)
```
Ellerbach, line 308: that's a bit expensive as a lock, any way to lock less that everything?
If my understanding is correct, the only 2 critical parts is _textRenderMatrix, so you may want to lock and check it  on when calling _textRenderMatrix.IsPixelSet

Ellerbach (commented):
Nice work, couple of quick comments

src/devices/SenseHat/SenseHatLedMatrix.cs
```diff
@@ -32,6 +34,45 @@ public abstract class SenseHatLedMatrix : IDisposable
...
+        /// </summary>
+        protected SenseHatTextRenderMatrix? _textRenderMatrix = null;
+
+        /// <summary>
+        /// Text color when rendering text. Null for "no color".
+        /// </summary>
+        protected Color _textColor = Color.DarkBlue;
+
+        /// <summary>
+        /// Text background color when rendering text. Null for "no color".
```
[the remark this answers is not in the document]
logicaloud: I have removed the "null" comment.

src/devices/SenseHat/SenseHatLedMatrix.cs
```diff
@@ -32,6 +34,45 @@ public abstract class SenseHatLedMatrix : IDisposable
...
+        /// </summary>
+        protected SenseHatTextFont? _pixelFont = null;
+
+        /// <summary>
+        /// Lazily initialized text render state. Initialized when rendering text.
+        /// </summary>
+        protected SenseHatTextRenderMatrix? _textRenderMatrix = null;
+
+        /// <summary>
+        /// Text color when rendering text. Null for "no color".
```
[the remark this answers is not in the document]
logicaloud: I have removed the "null" comment.

src/devices/SenseHat/SenseHatLedMatrix.cs
```diff
@@ -95,9 +138,216 @@ public static int PositionToIndex(int x, int y)
...
+        /// Text color when rendering text.
+        /// </summary>
+        [Property]
+        public Color TextColor
+        {
+            get
+            {
+                return _textColor;
+            }
+            set
```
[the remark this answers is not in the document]
logicaloud: It does not seem to mind here for `get { ... } set { ... } `, maybe similar to `if { ... } else { ... }`. I have not checked the rules - should there be an empty line? I can update if required.

src/devices/SenseHat/SenseHatTextRotation.cs
logicaloud: Not sure if I understand... The enum is named `Rotate_90_degrees`, `Rotate_180_degrees`, etc., so in the mathematical sense indicates this to be a positive (counterclockwise) rotation, but is it not helpful to point this out in the comment? Could you give an example of what kind of comment you had in mind?

src/devices/SenseHat/SenseHatLedMatrix.cs
logicaloud: Yes, that was a thorn in my eye also. It was necessary because the text animation timer could kick in mid-frame, and with every `SetPixel` the underlying device (i.e. i2c bus) is accessed. If the device is disposed mid frame, then it will cause an exception. Besides that, frame consistency is an issue.

I have refactored the locking approach by introducing a render state class holding a consistent set of parameters, and by using the `Write` method at the end of `RenderText`. This still requires a lock but it is brief and performance is much better.

You are probably going to ask whether the frame buffer needs to be allocated for each `RenderText`. It could be avoided with an additional lock, but I think it's OK as it is. What do you think?

src/devices/SenseHat/SenseHatTextRotation.cs
Ellerbach: By convention, the rotations are clockwise so if it's 90 degrees, it's 90 degrees clockwise. 270 degrees would be 90 degrees counterclockwise then.

src/devices/SenseHat/SenseHatLedMatrix.cs
Ellerbach: Yes, it's definitely better from my point of view!

src/devices/SenseHat/SenseHatTextRotation.cs
logicaloud: Sorry, still confused. The angle of rotation currently currently follows the mathematical convention (counterclockwise, i.e., https://en.wikipedia.org/wiki/Rotation_matrix; see "Direction"). Could you point to the convention that should be followed for the IoT library?

At the moment you get this:

image

Ellerbach: my bad. You're right. All good then!

src/devices/SenseHat/SenseHatLedMatrix.cs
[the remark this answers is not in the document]
logicaloud: I assume this is resolved, please let me know otherwise.

src/devices/SenseHat/SenseHatLedMatrix.cs
```diff
@@ -32,6 +34,51 @@ public abstract class SenseHatLedMatrix : IDisposable
         /// </summary>
         protected const int NumberOfPixelsPerColumn = 8;
 
+        /// <summary>
+        /// Lazily intialized pixel font reference. Initialized when rendering text.
+        /// </summary>
+        protected SenseHatTextFont? _pixelFont = null;
```
krwq, line 40: I do like the fact that we're extending this to support text scrolling etc but I'd prefer this logic was reusable to any LED matrix. Here is how I'd imagine this would work:
- this class remains mostly untouched (unless you need to add extra API or whatever to make text work then those changes are ok but nothing text related)
- text scrolling / setting is an abstraction which takes abstract display (or couple of Action/Func for each needed operation)

the usage would be more or less like this (you might have better ideas but this is bare minimum to isolate this utility):

```csharp
SenseHatLedMatrix matrix = ...;
LedMatrixTextScroller textScroller = new(width, height, (x, y, color) => matrix.DrawPixel(x, y, color));
textScroller.Start();
textScroller.SetText("hello");
textScroller.Stop();
```

Note: on first version I don't expect you to integrate anywhere else but if anyone else asks or tries to add it I'll point them in this direction so they can re-use this code

src/devices/SenseHat/SenseHatLedMatrix.cs
```diff
@@ -95,9 +144,300 @@ public static int PositionToIndex(int x, int y)
...
+
+                RenderText(renderState);
+            }
+        }
+
+        /// <summary>
+        /// Text scroll speed in pixels per second.
+        /// </summary>
+        [Property]
+        public double TextScrollPixelsPerSecond
```
krwq, line 273: note: if you're isolating this class but would like to utilize PropertyAttribute (although I don't ask you to do it in this PR) with the class you'd need to create a wrapper for the TextScroller, i.e. SenseHatLedMatrixTextScroller which would be just a tiny class which allow you to only do text writing (drawing and text scroller shouldn't be used together anyway because it will just look like bogus pixels showing up)

src/devices/SenseHat/SenseHatLedMatrix.cs
```diff
@@ -95,9 +144,300 @@ public static int PositionToIndex(int x, int y)
...
+                        frameBuffer[frameIndex] = textRenderState.TextColor;
+                    }
+                    else
+                    {
+                        frameBuffer[frameIndex] = textRenderState.TextBackgroundColor;
+                    }
+                }
+            }
+
+            // Lock write so that SenseHat disposal does not interfere
```
krwq, line 433: it's ok if someone disposes matrix prematurely as long as you get reasonable exception - we do not want to pollute the code just because someone might do weird things with the hardware

src/devices/SenseHat/SenseHatLedMatrix.cs
```diff
@@ -95,9 +144,300 @@ public static int PositionToIndex(int x, int y)
...
+            if (_textAnimationTimer != null)
+            {
+                _textAnimationTimer.Stop();
+                _textAnimationTimer.Elapsed -= TextAnimationTimer_Elapsed;
+                _textAnimationTimer.Dispose();
+                _textAnimationTimer = null;
+            }
+        }
+
+        private void TextAnimationTimer_Elapsed(object? sender, System.Timers.ElapsedEventArgs e)
```
krwq, line 365: Rather than timer consider allowing user pick when drawing happens. I.e.:

```
TextScroller scroller = ....;

while (true)
{
  scroller.Draw(); // note that you should keep track of time separately inside of this method so it doesn't move inconsistently when user calls this at not equal intervals
  Thread.Sleep(20);
}
```

this kind of design is usually more

krwq: And note - if you make this design the way I described above you can still use Timer in your app but it won't be necessary for scroller to have dependency on it - you'd just call DrawFrameIfNecessary inside of the timer callback

krwq: (of course pick better names if you have any idea but ideally it should be straight forward to understand what the method is doing by just looking at the name)

src/devices/SenseHat/SenseHatLedMatrix.cs
krwq: The callback could also draw entire buffer or whatever, pick a design which feels easiest to use from user perspective but is not unreasonably slow - I think for small matrix like this it doesn't matter if you draw by pixel or screen but for larger one by pixel might be a bit too slow but if library is not meant for large screens then that argument goes away - I think reasonably largest screen would be 64x64 so that gives you about 4k calls per 20ms which is around 200k operations per second - that might be good enough for Raspberry Pi 3. If we pick DrawPixel now worst case scenario we can always add extra overload to the scroller in the future with entire screen callback and just translate DrawPixel into DrawBuffer callback (basically you'd wrap user passed callback with your own which utilizes that to draw a screen). If we do reverse it will be much harder to simplify this later

krwq:
@logicaloud thanks for this work, it's really awesome! I added comments how we can leverage this outside of SenseHat as well

logicaloud:
@krwq Thank you for your review. I'll go through each point and comment separately.

src/devices/SenseHat/SenseHatLedMatrix.cs
logicaloud: The "text scroller" in this instance is very simple: It increments a value (in the timer events) that is the horizontal offset position of the pre-rendered text (in `SenseHatTextRenderMatrix`) from where bits should be copied to the LED matrix display. It may be possible to generalize the use for small, single-line LED displays. Larger displays will probably lead to different design choices. My suggestion would be to possibly add an issue "generalize text rendering and scrolling for single line LED displays" or similar, that could be picked up by someone who has such a device, otherwise there may be a bit of guessing of what is feasible. This pull request could go into a different branch in the mean time, but I think it would not be too much work to refactor for generalization later. I'd be happy to test the generalized solution on SenseHat if/when available.

logicaloud: There was a noticeable difference (smoother scrolling) when setting the pixels with a single `Write` operation vs multiple `SetPixel` (Pi 4, 8GB), although I can't recall whether I have visually compared in Debug mode only.  Maybe optimization here could also be left until implementation for another LED display is attempted?

src/devices/SenseHat/SenseHatLedMatrix.cs
logicaloud: The design objective was to keep things simple for the user, and text simply starts scrolling as soon as the there is more than one character to display, using the default `TextScrollPixelsPerSecond` (1 pixel per second). Separating the scrolling out is a possibility of course. Something to consider for a possible generalization maybe, that it should be possible to retain "automatically scroll long text" for tiny LED displays, which could be a property set in a base class with whatever default makes sense for the LED display at hand.

src/devices/SenseHat/SenseHatLedMatrix.cs
logicaloud: It's the disposal of the I2C bus that interferes in this case. Currently this works without exception:

```
using (var sh = new SenseHat()) 
{
    sh.LedMatrix.SetText("abc");
    // ... do something
}
```

In that case the text scrolling automatically starts because the text is long. At the end of the `using`, the `SenseHat` disposes the I2C bus. Without precautions, the text rendering could attempt to access the disposed I2C bus because it is in the middle of rendering a new frame. The `Lock` around `Write` in `SenseHatLedMatrix` prevents access to a disposed I2C device.

With the current API design (text automatically scrolls) it would not be obvious to the user that something needs to be reset before disposing the `SenseHat`. One could give up simplicity and ask the user to explicitly enable and disable text scrolling. Or demand that text is set to 'empty' before disposing. Or any exception in the text rendering method could be ignored. My preference would be to retain the lock for the simplest use without surprises; this would be for SenseHat on Pi with its small LED matrix. It is possible that generalizations for other LED displays require a different rendering method or make other design choices, eliminating the lock in their approach. - What do you think?

src/devices/SenseHat/SenseHatLedMatrix.cs
logicaloud: Maybe this could be an extended feature in the future, where the user can toggle between automatic scrolling and manual scrolling? For SenseHat, there is probably much gain in pausing text because then only one character (or parts of two) are visible at a time, but it would be a good thing to have for longer displays.

krwq (changes requested):
@logicaloud we've discussed this during our last triage meetings but given these utilities are being added only to SenseHat and that's not the only small screen implementation we have in the library it feels a bit out of place. This should be made as generic purpose if we're going to accept this change. There is nothing here which is SenseHat specific therefore adding implementation only in SenseHat doesn't feel right.

logicaloud:
@krwq Thanks for the review. I'd be happy to work with authors of other small screen display implementations to generalize the approach. Since I only have the Sensehat, I can't really test anything else. One approach could be to have some base classes with abstract methods or properties and then implement anything device specific for the various displays. 

I'm not really sure what the best way forward is from here and I won't be offended if you simply rejected the pull request at this stage.

logicaloud:
@krwq The implemented SenseHat text rendering is very limited. It handles a single line of text in an 8x8 matrix. The width of the matrix could be easily extended, but there is no provision for adjusting the height (and I cannot test that). There could be a `ISingleTextLineMatrixDisplay` interface for this special case but I'm not sure if that is the best approach.

I had a look at the existing `RgbLedMatrix` device. This device handles rendering into matrices of varying heights and widths, supports bitmap and primitive rendering (i.e., rectangle, circle), and allows for chaining of matrix displays. The SenseHat matrix display could become a specialization of a new common base class of the `RgbLedMatrix` device.

The `RgbLedMatrix` device renders into a memory back buffer before rendering the buffer in a device dependent way. That means, much of the rendering logic is device independent and could be pulled into the new base class without much effort.

So ideally, if `RgbLedMatrix` owners / authors could assist in putting this base class together, then the SenseHat LED matrix could inherit the same text, primitive, and bitmap rendering capabilities (as could other matrix displays).

PR #2325, opened 2024-06-11
Detect Raspberry 3B Plus with upgraded Firmware

fntc:
Raspberry 3B Plus when upgraded with new firmware will not be detected correctly. This change adds the detection of the new firmware version 20D4

Hardware        : BCM2835
Revision        : a0**20d4**
Serial          : 00000000687e3c1d
Model           : Raspberry Pi 3 Model B Plus Rev 1.4


raffaeler (approved):
Thank you for the PR!

Ellerbach (approved):
Thanks, that looks good!

PR #2345, opened 2024-08-19
Fix a possible crash when disposing ADS1115 device

pgrawehr:
In case a communication error happens during dispose, the driver would crash.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2345)

### DIFF
--- a/src/devices/Ads1115/Ads1115.cs
+++ b/src/devices/Ads1115/Ads1115.cs
@@ -289,6 +289,8 @@ namespace Iot.Device.Ads1115
                     {
                         throw;
                     }
+
+                    Thread.Sleep(100);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/devices/Ads1115/Ads1115.cs
```diff
@@ -261,19 +262,36 @@ private void SetConfig()
...
+                    writeBuff[2] = 0xFF;
+                    _i2cDevice.Write(writeBuff);
+                    success = true;
+                }
+                catch (IOException)
+                {
+                    if (retries-- < 0)
+                    {
+                        throw;
+                    }
```
Ellerbach, line 291: would it makes sense to wait few milliseconds?

Ellerbach (approved):
Looks good. Thanks a lot for the fix!

PR #2350, opened 2024-08-27
New Device: LED segment display 5641AS

jabberhams:
New device `LedSegmentDisplay5641AS` added. The device has fewer display characteristics (no colon or degrees) and while other devices in the Display category were driven by backpacks over I2C, this code is full GPIO only.  In addition to the device, a dedicated pin scheme `LedSegmentDisplay5641ASPinScheme` is for mapping board pins to device pins, and a helper class translates `Font`s to segment assignments.  Readme also updated.

My first contribution so looking for constructive feedback where necessary.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2350)


jabberhams (commented):
Fixed markdown syntax for local image

src/devices/Display/LedSegmentDisplay5641AS.cs
```diff
@@ -0,0 +1,190 @@
...
+            _shouldDispose = gpioController is null || shouldDispose;
+
+            _segments = new(pinScheme.SegmentCountPerDigit);
+            _digits = new(pinScheme.DigitCount);
+        }
+
+        /// <summary>
+        /// Opens GPIO pins defined in <see cref="LedSegmentDisplay5641ASPinScheme"/> as <see cref="PinMode.Output"/>
+        /// </summary>
+        public void Open()
```
Ellerbach, line 54: Why do you have the Open and Close functions? Can Open be done in the constructor and Close in the Dispose? So it's really fully transparent for the user? 
You can keep the function and use them but then change the visibility in private in that case.
I may have miss something why you need an Open and Close as public

src/devices/Display/LedSegmentDisplay5641AS.cs
```diff
@@ -0,0 +1,190 @@
...
+        private bool _disposedValue;
+
+        private List<GpioPin> _segments;
+        private List<GpioPin> _digits;
+
+        /// <inheritdoc />
+        public int NumberOfDigits => _pinScheme.DigitCount;
+
+        /// <inheritdoc />
+        public Segment this[int address] { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
```
Ellerbach, line 36: is there any way to recreate a Segment? So that it's make easy for the user and kind of easier?

src/devices/Display/README.md
```diff
@@ -1,10 +1,14 @@
-# HT16K33 - LED Matrix Display Driver
+# LED Displays
```
Ellerbach, line 1:
```suggestion
# LED Displays - 5641AS, HT16K33 and LED Matrix Display Driver
```

Ellerbach (commented):
Thanks, couple of questions and remarks. It looks good all up.

src/devices/Display/LedSegmentDisplay5641AS.cs
jabberhams: No reason, just carryover from testing. I will make the changes.

src/devices/Display/LedSegmentDisplay5641AS.cs
jabberhams: This has been updated in latest push - see  https://github.com/dotnet/iot/pull/2350/commits/e3f4cc05e0a007a5090d7955380df7bc77bbafa4

Ellerbach (approved):
Thanks, looks good. Good addition!

jabberhams:
@Ellerbach there is a build failure in `Arduino_Compiler_CI` - is there a way to requeue it? Seems like it is unrelated or ephemeral.

pgrawehr:
@jabberhams Thanks for the contribution. Sorry for the delayed update, but the team has been quite busy with RL recently.

PR #2359, opened 2024-10-23
Fix GrovePi binding timing issue when read sensors

### LINKED ISSUE #2337: GrovePi device binding fails to read sensor values
**Describe the bug**

When using the `DigitalRead` and `AnalogRead` functions included in the [device binding for GrovePi](https://github.com/dotnet/iot/tree/main/src/devices/GrovePi) to read sensor values, the functions return `-1` instead of the sensor output, making it impossible to read sensor values.

Using the [Python library](https://github.com/DexterInd/GrovePi/tree/master/Software/Python) provided by the device manufacturer, correct values can be read.

When reproducing the issue using `AnalogRead`, the binary data obtained from `I2CDevice.Read` is `23 255 255`, and in the Python binding, `23` is declared as `data_not_available_cmd` which triggers a retry condition.

In dotnet, by modifying the code to perform a retry when the first byte of the response is `23`, correct sensor values can be read.

If this issue and the proposed fix are deemed appropriate, I would like to submit a PR.

**Steps to reproduce**

1. Connect a Raspberry Pi model 4B with the GrovePi board and a push-button module.
2. Execute the following minimal C# code, based on the sample code:
3. The values written to the standard output will be -1 regardless of the button operation.
```csharp
I2cConnectionSettings i2CConnectionSettings = new(1, GrovePi.DefaultI2cAddress);
using GrovePi grovePi = new GrovePi(I2cDevice.Create(i2CConnectionSettings));

Console.WriteLine($"Manufacturer: {grovePi.GrovePiInfo.Manufacturer}");
Console.WriteLine($"Board: {grovePi.GrovePiInfo.Board}");
Console.WriteLine($"Firmware version: {grovePi.GrovePiInfo.SoftwareVersion}");

var buttonPin = GrovePort.AnalogPin0;
grovePi.PinMode(buttonPin, PinMode.Input);

for (int i = 0; ; i++)
{
    var value = grovePi.AnalogRead(buttonPin);
    Console.WriteLine($"{i}: button={value}");

    Thread.Sleep(100);
}
```

**Expected behavior**

The AnalogRead function should return the sensor values. internal `I2CDevice.Read` function should return responses like `3 3 255` (indicating maximum voltage level) or `3 0 0` (indicating minimum voltage level).

**Actual behavior**

The AnalogRead function returns -1. The internal `I2CDevice.Read` function returns a response of `23 255 255` .

**Versions used**

Compiled and executed on dotnet8 running on a Raspberry Pi 4. The GrovePi board firmware used is the latest version 1.4.

```
.NET SDK:
 Version:           8.0.303
 Commit:            29ab8e3268
 Workload version:  8.0.300-manifests.c915c39d
 MSBuild version:   17.10.4+10fbfbf2e

Runtime Environment:
 OS Name:     debian
 OS Version:  11
 OS Platform: Linux
 RID:         linux-arm64
 Base Path:   /home/pi/.dotnet/sdk/8.0.303/

.NET workloads installed:
There are no installed workloads to display.

Host:
  Version:      8.0.7
  Architecture: arm64
  Commit:       2aade6beb0

.NET SDKs installed:
  8.0.303 [/home/pi/.dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 8.0.7 [/home/pi/.dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 8.0.7 [/home/pi/.dotnet/shared/Microsoft.NETCore.App]

Other architectures found:
  None

Environment variables:
  DOTNET_ROOT       [/home/pi/.dotnet]

global.json file:
  Not found
```

2427dkusiro:
Fixes #2337

## Summary
Fix timing issue in the GrovePi Binding logic for sensor reading.

## Detail
As mentioned in the issue I raised, there is a bug in the current GrovePi Binding that prevents it from correctly reading sensor values on the Raspberry Pi 4. To fix this bug, I have modified the code to wait and retry when an invalid response is received from the device. Additionally, I've standardized the implementation of the `DigitalRead` method with other sensor readings so that it can also benefit from this bug fix.

## Validation
I have confirmed that the changes in this PR work as intended by testing with sample code from the GrovePi directory. I connected a button, potentiometer, relay, ultrasonic distance sensor, and temperature & humidity sensor to both Raspberry Pi 3 and 4.

IMG_20241023_163606

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2359)

### DIFF
--- a/src/devices/GrovePi/GrovePi.cs
+++ b/src/devices/GrovePi/GrovePi.cs
@@ -17,7 +17,7 @@ namespace Iot.Device.GrovePiDevice
     /// </summary>
     public class GrovePi : IDisposable
     {
-        private const byte MaxRetries = 4;
+        private const byte MaxRetries = 10;
         private readonly bool _shouldDispose;
         private I2cDevice _i2cDevice;
 
--- a/src/devices/GrovePi/GrovePi.cs
+++ b/src/devices/GrovePi/GrovePi.cs
@@ -175,20 +175,13 @@ namespace Iot.Device.GrovePiDevice
         public PinValue DigitalRead(GrovePort pin)
         {
             WriteCommand(GrovePiCommand.DigitalRead, pin, 0, 0);
-            try
-            {
-                var data = ReadCommand(GrovePiCommand.DigitalRead, pin);
-                if (data is null || data.Length < 2)
-                {
-                    return (PinValue)(-1);
-                }
-
-                return (PinValue)data[1];
-            }
-            catch (IOException)
+            var data = ReadCommand(GrovePiCommand.DigitalRead, pin);
+            if (data is null || data.Length < 2)
             {
                 return (PinValue)(-1);
             }
+
+            return (PinValue)data[1];
         }
 
         /// <summary>
--- a/src/devices/GrovePi/GrovePi.cs
+++ b/src/devices/GrovePi/GrovePi.cs
@@ -7,6 +7,7 @@ using System.Device.Gpio;
 using System.Device.I2c;
 using System.IO;
 using System.Threading;
+
 using Iot.Device.GrovePiDevice.Models;
 
 namespace Iot.Device.GrovePiDevice
@@ -100,7 +101,7 @@ namespace Iot.Device.GrovePiDevice
                     return;
                 }
                 catch (IOException ex)
-                 {
+                {
                     innerEx = ex;
                     tries++;
                     Thread.Sleep(10);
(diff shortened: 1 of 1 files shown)


2427dkusiro:
@dotnet-policy-service agree

src/devices/GrovePi/GrovePi.cs
```diff
@@ -165,27 +175,20 @@ public void WriteCommand(GrovePiCommand command, GrovePort pin, byte param1, byt
...
-                {
-                    // Give it another try
-                    innerEx = ex;
-                    tries++;
-                    Thread.Sleep(10);
+                    return (PinValue)(-1);
                 }
-            }
 
-            throw new IOException($"{nameof(DigitalRead)}: Failed to read byte with command {GrovePiCommand.DigitalRead}", innerEx);
```
Ellerbach, line 188: Why are you removing the throw here and replace with the -1 logic? Throwing is a good way to say that things did not work properly.

2427dkusiro: @Ellerbach 
I initially considered making this change because the existing [`AnalogRead` method](https://github.com/dotnet/iot/pull/2359/files#diff-499d0f2c3b3d578696ec4fefd1a30c638568203e461f54021ef624576baf384eL210) was designed to return -1 instead of throwing an exception. However, I have adjusted the behavior at commit 93c630371654591b84a5d69241ad368607f2acaa accordingly.

PR #2362, opened 2024-11-14
FTx: Invert Data and direction position when setting up outputs for SPI mode

ProtectedVariable:
The opcode specification says that data comes before direction. The current code prevents the clock pin from outputting the clock signal.

I've tested with a FT232H and the previous version made PIN D0 output a constant 3.3v signal. With the changes it outputs the clock signal

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2362)


Ellerbach (approved):
Looks good. Thank you!

Ellerbach:
@pgrawehr the Arduino build is failing again. Does that make sense to make this pipeline manual? So, we trigger it only when there is something related to it? In that case, it's not at all supported on that specific platform.

pgrawehr:
@ProtectedVariable Thanks for the fix!

PR #2355, opened 2024-09-30
Use Arcade 8.0, use .NET 8.0, remove old features

pgrawehr:
Fixes #2353 and #2354
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2355)

PR #2368, opened 2024-12-27
Add virtual gpio controller, attempt 2

pgrawehr:
Added virtual Gpio Controller with a slightly modified approach
Note: Fails to compile as it requires #2355 (currently uses a method only available in .NET 8.0)

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2368)


Ellerbach (approved):
Sounds better than my original approach. Great to see you managed to find this alternative! Thanks a lot for moving forward this very old PR!

PR #2376, opened 2025-01-18
Fix calculations of BMM150

pgrawehr:
The calculated headings (both corrected and uncorrected) of the BMM150 Magnetometer where wrong due to an incorrect sign extension. The compensation functions also had sign errors. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2376)


src/devices/Arduino/ArduinoBoard.cs
```diff
@@ -96,6 +96,7 @@ public ArduinoBoard(string portName, int baudRate)
         {
             _dataStream = null;
             _serialPort = new SerialPort(portName, baudRate);
+            _serialPort.ReadTimeout = int.MaxValue - 10;
```
krwq, line 99: add a comment on `- 10` please 😄

src/devices/Bmm150/Bmm150.cs
```diff
@@ -133,23 +134,36 @@ private void Initialize()
...
-        /// Get the device information
+        /// Calibrate the magnetometer.
+        /// Please make sure you are not close to any magnetic field like magnet or phone
+        /// Please make sure you are moving the magnetometer all over space, rotating it.
         /// </summary>
-        /// <returns>The device information</returns>
-        public byte GetDeviceInfo() => ReadByte(Bmp180Register.INFO);
+        /// <param name="numberOfMeasurements">Number of measurement for the calibration, default is 100</param>
+        // https://platformio.org/lib/show/12697/M5_BMM150
+        [Obsolete("Not a reliable way of calibration. Use a calibration determined with MagneticDeviationCorrection instead")]
```
krwq, line 143: can we remove all the obsolete methods?

raffaeler (approved):

pgrawehr:
> Hum, there is still something I don't get, as with the previous method, you were able to really have a great space repartition (see the pictures). If you don't apply a proper calibration like this, you'll mainly get improper numbers. Unfortunately, on other platforms and in the samples online, the calibration is most of the time not done properly. So, please generate a similar diagram with all the 3 axes and please make sure you really get them centered.

A calibration is only possible if you know what the values should be, as with this sensor, the expected reading when not in motion is not 0 (we can't switch off the earth's magnetic field). So to properly calibrate such a sensor, you need some kind of reference, E.g. a GNSS sensor and an attitude sensor. I've started to investigate into this topic, but that is a lot of math and I couldn't find the proper practical documentation on this yet. In either case, that's something that needs a separate PR.

PR #2377, opened 2025-01-25
Use RandomNumberGenerator.Fill in Credit Card code

krwq:
Note that this PR will fail until https://github.com/dotnet/iot/pull/2355 is merged and it doesn't seems worth to ifdef or use alternative API at this point.

Since this code is relying on randomness we should use the proper randomness here. Also that code analyzers don't complain about it :-)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2377)

### DIFF
--- a/src/devices/Card/CreditCard/CreditCard.cs
+++ b/src/devices/Card/CreditCard/CreditCard.cs
@@ -5,6 +5,7 @@ using System;
 using System.Buffers.Binary;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


krwq:
Draft because this will fail until we merge #2355

Ellerbach (approved):
looks good! Pull main and place the PR as ready.

PR #2405, opened 2025-07-13
Fixes #2404 - Execute interrupt handler for all 8 pins

### LINKED ISSUE #2404: Interrupt handler for Mcp23xxx only handles interrupts on first 4 pins
**Describe the bug**

In the Mcp23xxx driver (used by Mcp23008, Mcp23017, etc) the interrupt handler has a loop where it checks if the mask is matching the pending interrupts, but the loop checks only 4 of the total 8 pins, so interrupts are never working on pins 5-8 (for either port A and B on Mcp23017). This is the code:
```
            int mask = 1;
            int pin = 0;

            while (mask < 0x10)
            {
                if ((interruptPending & mask) != 0)
                {
                    CallHandlerOnPin(pin + offset, newValues & mask);
                }

                mask = mask << 1;
                pin++;
            }
```

**Steps to reproduce**

Hook up a button to pin 5-8 and you'll never receive the interrupts.

**Expected behavior**

Interrupts should received

**Actual behavior**

Nothing

**Versions used**

4.0.1

HakanL:
Check all 8 pins if the mask is matching the pending interrupts

### DIFF
--- a/src/devices/Mcp23xxx/Mcp23xxx.cs
+++ b/src/devices/Mcp23xxx/Mcp23xxx.cs
@@ -746,7 +746,7 @@ namespace Iot.Device.Mcp23xxx
             int mask = 1;
             int pin = 0;
 
-            while (mask < 0x10)
+            while (mask <= 0xF0)
             {
                 if ((interruptPending & mask) != 0)
                 {


src/devices/Mcp23xxx/Mcp23xxx.cs
```diff
@@ -746,7 +746,7 @@ private void InterruptHandler(object sender, PinValueChangedEventArgs e)
             int mask = 1;
             int pin = 0;
 
-            while (mask < 0x10)
+            while (mask <= 0xF0)
```
krwq: should this be `mask < 0xF0` or `mask & 0x0F != 0`?

krwq: btw. is there a datasheet page we can take a look at here? (if so can you add a comment?)

HakanL: @krwq This is the datasheet: https://ww1.microchip.com/downloads/en/DeviceDoc/MCP23008-MCP23S08-Data-Sheet-20001919F.pdf
and on page 16 (1.6.8 - Interrupt flag INTF register) you'll see that each of the 8 bits corresponds to a pin (for the 16-pin chip there's a secondary register).
What this change does is change so the mask iterates through 0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80 where before it was just 0x01, 0x02, 0x04, 0x08 (which was the bug). You're correct that is could be `mask < 0xF0` (but not `mask & 0x0F != 0` as that wouldn't properly evaluate the correct stopping point), I just went with this, basically the `mask` will be 0x100 when the while loop should stop.

raffaeler: The 0xF0 is pretty weird as the sequence never has more than a bit set.
I would have rather changed the loop in a do-while exiting at 0x80 :-)

HakanL: @raffaeler That's fine, when I changed it I was thinking the existing code that checked for 0x10 was because it went from 0x0F to 0x10 and that's why I just "extended" that and it tested out fine, but you're correct it's only doing one bit at a time, the more correct check is 0x80. I will push a change.

HakanL: @krwq Or you wanted me to add the datasheet link in the commit?

raffaeler: Yes, datasheet and paragraph please, for the future readers.

HakanL: @raffaeler Done!

raffaeler: Thank you!

raffaeler:
@HakanL Sorry for the merge taking so long. We are seeing some issues in the pipeline, but this should get solved very soon.

HakanL:
> @HakanL Sorry for the merge taking so long. We are seeing some issues in the pipeline, but this should get solved very soon.

Yeah I saw those, was worried my change caused them to fail, but my change was tiny so I couldn't see how it would not go through the builds.

PR #2408, opened 2025-08-01
Use only one interrup callback for one chip.

tolobis:
Prevent multiple callbacks on the interrupt pin. Only use one inpterrupt callback.
Fixing typos.
Code is test with a Tca9555 and Tca9554

### DIFF
--- a/src/devices/Tca955x/Tca955x.cs
+++ b/src/devices/Tca955x/Tca955x.cs
@@ -168,7 +168,7 @@ namespace Iot.Device.Tca955x
             {
                 if (mode != PinMode.Input && mode != PinMode.Output && mode != PinMode.InputPullUp)
                 {
-                    throw new ArgumentException("The Mcp controller supports the following pin modes: Input, Output and InputPullUp.");
+                    throw new ArgumentException("The Tca955x controller supports the following pin modes: Input, Output and InputPullUp.");
                 }
 
                 byte polarityInversionRegister = GetRegisterIndex(pinNumber, Register.PolarityInversionPort);
@@ -434,7 +434,12 @@ namespace Iot.Device.Tca955x
 
             _interruptPins.Add(pinNumber, eventType);
             _interruptLastInputValues.Add(pinNumber, Read(pinNumber));
-            _controller.RegisterCallbackForPinValueChangedEvent(_interrupt, PinEventTypes.Falling, InterruptHandler);
+
+            // Only register the callback if this is the first add callback
+            if (_interruptPins.Count == 1)
+            {
+                _controller.RegisterCallbackForPinValueChangedEvent(_interrupt, PinEventTypes.Falling, InterruptHandler);
+            }
 
             _eventHandlers[pinNumber] = callback;
         }
@@ -452,7 +457,12 @@ namespace Iot.Device.Tca955x
             {
                 _interruptPins.Remove(pinNumber);
                 _interruptLastInputValues.Remove(pinNumber);
-                _controller.UnregisterCallbackForPinValueChangedEvent(_interrupt, InterruptHandler);
+
+                // Only remove interrup callback if there are no more interrup pins active
+                if (_interruptPins.Count == 0)
+                {
+                    _controller.UnregisterCallbackForPinValueChangedEvent(_interrupt, InterruptHandler);
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/devices/Tca955x/Tca955x.cs
```diff
@@ -434,7 +434,12 @@ protected override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEve
 
             _interruptPins.Add(pinNumber, eventType);
             _interruptLastInputValues.Add(pinNumber, Read(pinNumber));
-            _controller.RegisterCallbackForPinValueChangedEvent(_interrupt, PinEventTypes.Falling, InterruptHandler);
+
+            // Only register the callback if this is the first add callback
```
krwq: Can you improve a comment a bit to clarify that we need to register interrupt once for host but this logic is for extender interrupts therefore the single handler will handle all extender interrupts registrations

krwq: (same for removal - it seems like many of us were confused about what's being fixed initially)

src/devices/Tca955x/Tca955x.cs
```diff
@@ -434,7 +434,12 @@ protected override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEve
...
+            if (_interruptPins.Count == 1)
```
joperezr: This class is not necessarily thread safe, but these checks are very likely to fail in any concurrent environment. Is it worth it to add a lock before these checks?

krwq: you can `lock (_interruptHandlerLock)` as we do elsewhere for all register/deregister

src/devices/Tca955x/Tca955x.cs
tolobis: I updated the comment so it should be now more clear why I register the callback for the interrupt only once

src/devices/Tca955x/Tca955x.cs
tolobis: @joperezr @krwq Thank you. I didnt see that in the first time. I added a lock around it.

src/devices/Tca955x/Tca955x.cs
```diff
@@ -432,9 +432,22 @@ protected override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEve
...
+                // The TCA955x family of I/O expanders shares a single INT (interrupt) output pin for all input ports in input mode.
+                // Any rising or falling edge on any input pin configured as input will trigger this shared interrupt.
+                // Therefore, we only need to register the interrupt handler once, regardless of how many input pins are monitored.
+                // This avoids redundant registrations and ensures efficient handling of input changes.
+                if (_interruptPins.Count == 1)
+                {
+                    _controller.RegisterCallbackForPinValueChangedEvent(_interrupt, PinEventTypes.Falling, InterruptHandler);
+                }
+
+            }
```
krwq: I think `_eventHandlers[pinNumber] = ...` shuold also be inside the lock

src/devices/Tca955x/Tca955x.cs
```diff
@@ -432,9 +432,22 @@ protected override void AddCallbackForPinValueChangedEvent(int pinNumber, PinEve
...
+                // Register the interrupt callback only once, when the first input pin is added.
+                // The TCA955x family of I/O expanders shares a single INT (interrupt) output pin for all input ports in input mode.
+                // Any rising or falling edge on any input pin configured as input will trigger this shared interrupt.
+                // Therefore, we only need to register the interrupt handler once, regardless of how many input pins are monitored.
+                // This avoids redundant registrations and ensures efficient handling of input changes.
+                if (_interruptPins.Count == 1)
+                {
+                    _controller.RegisterCallbackForPinValueChangedEvent(_interrupt, PinEventTypes.Falling, InterruptHandler);
+                }
+
```
krwq, line 449: nit: extra enter

krwq (approved):
LGTM with small comment

tolobis:
> I think this should just sign up when the driver is created/disposed. See my implementation in #2417
> 
> Sorry, I didn't see this PR until just now.

The #2417 PR looks good to me. Didn't see the other bugs. If you implement the interrupt lock we can close this PR an merge yours.

felixcollins:
I got carried away and made the TCA955x work better. I dropped my other PR and opened another #2420

PR #2412, opened 2025-08-11
Add kernel timestamp support

michael-betz:
The linux kernel provides high resolution timestamps for received CAN frames. 

This pull requests adds `CanRaw.GetLastTimeStamp()` to retrieve the timestamp of the last read frame.

The timestamp is in micro-seconds unix time.

Tested with linux virtual CAN device (no hardware needed).

```bash
# Add virtual CAN interface called `vcan0`
sudo ip link add dev vcan0 type vcan
sudo ip link set up vcan0

dotnet run --project src/devices/SocketCan/samples/SocketCan.Samples.csproj receive vcan0

# in another window
cansend vcan0 123#DEADBEEF

# Send 10 packets back to back
for i in {1..10}; do cansend vcan0 123#DEADBEEF; done

# Console output
  Can.Samples.csproj receive vcan0
  Listening for any id
  Ts: 2025-08-24T17:59:17.9521580Z Id: 0x123 [SFF]: DEADBEEF
  Ts: 2025-08-24T17:59:18.2795050Z Id: 0x123 [SFF]: DEADBEEF
  Ts: 2025-08-24T17:59:19.2140580Z Id: 0x123 [SFF]: DEADBEEF
```
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2412)

### DIFF
--- a/src/devices/SocketCan/CanRaw.cs
+++ b/src/devices/SocketCan/CanRaw.cs
@@ -125,7 +125,7 @@ namespace Iot.Device.SocketCan
         /// <param name="id">Recipient identifier</param>
         /// <param name="timestamp">UTC time when this frame was received with microsecond resolution</param>
         /// <returns></returns>
-        public bool TryReadFrame(Span<byte> data, out int frameLength, out CanId id, out DateTime timestamp)
+        public bool TryReadFrame(Span<byte> data, out int frameLength, out CanId id, out DateTimeOffset timestamp)
         {
             bool ret = TryReadFrame(data, out frameLength, out id);
             timestamp = Interop.GetLastTimeStamp(_handle);
--- a/src/devices/SocketCan/Interop.cs
+++ b/src/devices/SocketCan/Interop.cs
@@ -152,7 +152,7 @@ namespace Iot.Device.SocketCan
             return ifr.ifr_ifindex;
         }
 
-        public static unsafe DateTime GetLastTimeStamp(SafeHandle handle)
+        public static unsafe DateTimeOffset GetLastTimeStamp(SafeHandle handle)
         {
             // From Linux 6.12.33+deb13-amd64 x86_64
             const uint SIOCGSTAMP = 0x8906;
@@ -164,11 +164,8 @@ namespace Iot.Device.SocketCan
                 throw new IOException("Could not get socketcan timestamp");
             }
 
-            DateTime dt_utc = DateTimeOffset.FromUnixTimeSeconds(tv.tv_sec)
-                            .AddTicks(tv.tv_usec * 10) // 1 tick = 100 ns
-                            .UtcDateTime;
-
-            return dt_utc;  // UtcDateTime
+            // tv_usec is in [us]. DateTimeOffset in ticks of [100 ns], that's why * 10 is needed
+            return DateTimeOffset.FromUnixTimeSeconds(tv.tv_sec).AddTicks(tv.tv_usec * 10);
         }
 
         [StructLayout(LayoutKind.Sequential)]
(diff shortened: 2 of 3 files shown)


src/devices/SocketCan/Interop.cs
```diff
@@ -149,6 +152,27 @@ private static unsafe int GetInterfaceIndex(int fd, string name)
...
+            int ret = Ioctl4((int)handle.DangerousGetHandle(), SIOCGSTAMP, ref tv);
+            if (ret == -1)
+            {
+                throw new IOException("Could not get socketcan timestamp");
+            }
+
+            return ((ulong)tv.tv_sec * 1000000 + (ulong)tv.tv_usec);  // unix time in [us]
+        }
+
+        internal unsafe struct TimeVal
```
krwq, line 172: there should be a comment that `long` is correct type only because we're using this in conjunction with latest kernel. Please double check this actually works with 32-bit OS-es (if not then please make sure to display correct error message)

krwq: note, you don't need to test it - just double check what type is defined in the latest kernel's .h file, if it's `int64_t` or equivalent we're good if it's platform specific like `int` then likely not

michael-betz: Okay, I did some digging. In C, the ioctl call returns:

```C
       struct timeval {
           time_t       tv_sec;   /* Seconds */
           suseconds_t  tv_usec;  /* Microseconds */
       };
```

`time_t` can be a signed integer of 32 bit or 64 bit, depending on the platform.

In C#, I map this to a `long` type which is always a 64 bit signed integer.

If I run this on a 64 bit system, everything fits bit by bit.

If I run this on a 32-bit system, the C# runtime will promote the 32-bit return value to `long` properly.

Please correct me if I'm wrong.

I will look into testing this on a 32 bit system.

raffaeler: Is `DateTime` the correct for this timestamp?
When I see a `DateTime`, I expect this represent the "real" moment when the event occurred. I also expect to be able to compare this value with other `DateTime` instances coming from other sensors.

If instead this timestamp represents the time elapsed since the OS started, IMO the `TimeSpan` type would be more appropriate.

michael-betz: The kernel does indeed return the absolute value of the system clock when the CAN frame was received (microseconds since the 1.1.1970). So I think `DateTime` is a good type for this.

raffaeler: Good to know, thanks

michael-betz:
> You should also make sure that when older version of kernel is used the error message tells you what version of kernel is required to use the new API.

Okay, I did some research on this. 
SocketCAN was merged into the mainline Linux kernel in version 2.6.25, which was released in April 2008.
To me it looks like this initial version already supported timestamps via `ioctl(SIOCGSTAMP)`.

So seems like, if socketcan works, there is a very good chance that getting timestamps will also work.

src/devices/SocketCan/CanRaw.cs
```diff
@@ -117,6 +117,21 @@ public bool TryReadFrame(Span<byte> data, out int frameLength, out CanId id)
...
 
+        /// <summary>
+        /// Reads frame from the bus with the timestamp when it was received.
+        /// </summary>
+        /// <param name="data">Data where output data should be written to</param>
+        /// <param name="frameLength">Length of the data read</param>
+        /// <param name="id">Recipient identifier</param>
+        /// <param name="timestamp">UTC time when this frame was received with microsecond resolution</param>
+        /// <returns></returns>
+        public bool TryReadFrame(Span<byte> data, out int frameLength, out CanId id, out DateTime timestamp)
```
krwq: This must be DateTimeOffset on public APIs - it must contain the time zone. Internally do whatever to convert

krwq (approved):

PR #2416, opened 2025-08-28
Add GitHub Actions workflow for publishing nightly packages

joperezr:
FYI: @pgrawehr @krwq @raffaeler 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2416)

### DIFF
--- a/.github/workflows/publish-nightly.yml
+++ b/.github/workflows/publish-nightly.yml
@@ -1,0 +1,60 @@
+name: Publish Nightly Packages
+
+on:
+  push:
+    branches: [ main ]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish-nightly:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0 # Ensure we have full git history for versioning
+    
+    - name: Build and pack
+      run: ./build.sh -restore -build -pack
+      
+    - name: Find package files
+      id: find-packages
+      run: |
+        # Find all nupkg files in artifacts/packages
+        packages=$(find artifacts/packages -name "*.nupkg" -not -name "*.symbols.nupkg" | tr '\n' ' ')
+        echo "Found packages: $packages"
+        echo "packages=$packages" >> $GITHUB_OUTPUT
+        
+        # Count packages to verify we have the expected ones
+        package_count=$(echo $packages | wc -w)
+        echo "Package count: $package_count"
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-nightly.yml
```diff
@@ -0,0 +1,64 @@
+name: Publish Nightly Packages
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
joperezr:
```suggestion
```

joperezr:
@raffaeler @krwq @pgrawehr I've tested this and it works well and does what we need. There is one caveat that I found, which is that apparently GitHub Package feeds require authentication, unlike AzDO feeds that can be used anonymously. This is not super ideal, but we can wait for feedback and if we hear that it is a main drawback then we can look for another alternative. In general, people using these feeds in real repos work around it by adding a CI step (or prereq for local builds) where the token is added to the nuget.config for things to restore correctly.

raffaeler:
@joperezr thanks for the super-fast work :)

> apparently GitHub Package feeds require authentication

This may be a blocker for people wishing to automate the CI in their libraries/app using our nightly builds
While it is definitely a welcome solution on the short term, I believe it would be better to add the AzDO in parallel to this.

raffaeler:
BTW we should definitely move the `Arduino_Compiler_CI` out of the default workflow to make the CI run faster
/cc @pgrawehr

pgrawehr:
> BTW we should definitely move the `Arduino_Compiler_CI` out of the default workflow to make the CI run faster
> /cc @pgrawehr 

It's not part of the required tests, it just runs in parallel. I'd like to have it faster, but I don't currently know how.

raffaeler:
> > BTW we should definitely move the `Arduino_Compiler_CI` out of the default workflow to make the CI run faster
> > /cc @pgrawehr
> 
> It's not part of the required tests, it just runs in parallel. I'd like to have it faster, but I don't currently know how.

@joperezr what about creating a new workflow for the `Arduino_Compiler_CI` only that can be run by @pgrawehr manually when he modifies the source code? Or maybe you have a better solution :)

raffaeler:
> That would be super simple, it is literally just a one line change (need to remove the pull request trigger from the existing workflow). If @pgrawehr agrees, we can do the quick change.

@joperezr @pgrawehr 
If I remember well, we discussed about the opportunity to create a separate NuGet package for the Arduino code. In this case the potentially new pipeline for Arduino could run only when the PR affects its files, right?

PR #2417, opened 2025-09-01
Tca955x improvements

felixcollins:
I found some bugs in Tca955x driver. This is actually a pretty important driver as it works with many of the i2c io expanders.

Bug 1. The value for the port configuration register offset was wrong. Most likely a cut and paste error. 

Bug 2: The gpio optionally used for the interrupt pin on the ioexpander was subscribed for change every time one of the i/o pins on the expander was subscribed. That could have lead to mysterious multiple firings of events. Only need to sign up once on construction.

Bug 3: When the pin changed callback was called it reported the event type as the events that were subscribed, not the event that happened.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/iot/pull/2417)

### DIFF
--- a/src/devices/Tca955x/Tca955x.cs
+++ b/src/devices/Tca955x/Tca955x.cs
@@ -436,6 +436,7 @@ namespace Iot.Device.Tca955x
         {
             if (_controller == null)
             {
+                // We could offer a polling solution here instead.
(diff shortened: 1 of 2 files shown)


felixcollins:
@dotnet-policy-service agree company="Intranel"

PR #2420, opened 2025-09-04
Improve tca955x driver - (and add block GPIO access)

felixcollins:
This PR has numerous bug fixes and improvements to the TCA955x driver. 

In particular, the handling of the INT pin on the chip is vastly improved. The event handler for the falling edge on that pin now delegates to a Task. This does the work of reading the chip via i2c and calling any signed up events for the pins of the chip. While all this is going on, if there are any more than one further falling edge of the INT pin, they are ignored. If there is at least one then the check of the chip input registers and event firing is redone at the end. This behaviour prevents blocking the GpioDriver that is servicing the INT pin. Blocking could (especially in the case of libgiod v1) cause a queue up the events and subsequent delivery of them long after the actual inputs on the chip have settled.

The TCA955x driver now just does the one or two reads of the i2c that are needed, when the INT pin asserts, rather than repeating that read for every pin that has an event handler. Similar to this change, the GpioController now does a block read of its GpioDriver if the GpioDriver implementation supports block reads of I/O by implementing IGpioDriverBlockAccess. (This should probably be in a separate PR - sorry about that) 

Other fixes:
Report event type of pin in event rather than reporting the types signed up for.
A few more tests and fixed some redundant/incorrect tests.
Fixed reporting of supported pin types.
Fixed pin type of INT pin to support Input or InputPullup and default to Input.
Improved the block io read/write to reduce i2c traffic
Fixed incorrect register address constant
Spelling improved

### DIFF
--- a/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
@@ -31,6 +31,7 @@ public class GpioController : IDisposable
     /// </summary>
     private readonly ConcurrentDictionary<int, PinValue?> _openPins;
     private readonly ConcurrentDictionary<int, GpioPin> _gpioPins;
+    private readonly bool _driverHasBlockAccess = false; // Cache the block access capability. Doing it this was trades one bool of memory for a little performance gain.
     private GpioDriver _driver;
 
     /// <summary>
@@ -50,6 +51,10 @@ public class GpioController : IDisposable
     public GpioController(GpioDriver driver)
     {
         _driver = driver;
+        if (_driver is IGpioDriverBlockAccess)
+        {
+            _driverHasBlockAccess = true;
+        }
 
 #pragma warning disable CS0612 // PinNumberingScheme is obsolete
         NumberingScheme = PinNumberingScheme.Logical;
@@ -464,9 +469,16 @@ public class GpioController : IDisposable
     /// <param name="pinValuePairs">The pin/value pairs to write.</param>
     public void Write(ReadOnlySpan<PinValuePair> pinValuePairs)
     {
-        for (int i = 0; i < pinValuePairs.Length; i++)
+        if (_driverHasBlockAccess)
         {
-            Write(pinValuePairs[i].PinNumber, pinValuePairs[i].PinValue);
+            ((IGpioDriverBlockAccess)_driver).Write(pinValuePairs);
+        }
+        else
+        {
+            for (int i = 0; i < pinValuePairs.Length; i++)
+            {
+                Write(pinValuePairs[i].PinNumber, pinValuePairs[i].PinValue);
+            }
         }
     }
 
@@ -476,10 +488,17 @@ public class GpioController : IDisposable
     /// <param name="pinValuePairs">The pin/value pairs to read.</param>
     public void Read(Span<PinValuePair> pinValuePairs)
     {
-        for (int i = 0; i < pinValuePairs.Length; i++)
+        if (_driverHasBlockAccess)
         {
-            int pin = pinValuePairs[i].PinNumber;
-            pinValuePairs[i] = new PinValuePair(pin, Read(pin));
+            ((IGpioDriverBlockAccess)_driver).Read(pinValuePairs);
+        }
+        else
+        {
+            for (int i = 0; i < pinValuePairs.Length; i++)
+            {
+                int pin = pinValuePairs[i].PinNumber;
+                pinValuePairs[i] = new PinValuePair(pin, Read(pin));
+            }
         }
     }
 
--- a/src/System.Device.Gpio/System/Device/Gpio/IGpioDriverBlockAccess.cs
+++ b/src/System.Device.Gpio/System/Device/Gpio/IGpioDriverBlockAccess.cs
@@ -1,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Device.Gpio
+{
+    /// <summary>
+    /// Interface to mark drivers that support block access to GPIO pins.
+    /// Block access in this case meaning the ability to read/write multiple pins in one call.
+    /// </summary>
+    public interface IGpioDriverBlockAccess
+    {
+        /// <summary>
+        /// Write the given pins with the given values.
+        /// </summary>
+        /// <param name="pinValuePairs">The pin/value pairs to write.</param>
+        public void Write(ReadOnlySpan<PinValuePair> pinValuePairs);
+
+        /// <summary>
+        /// Read the given pins with the given pin numbers.
+        /// </summary>
+        /// <param name="pinValuePairs">The pin/value pairs to read.</param>
+        public void Read(Span<PinValuePair> pinValuePairs);
+    }
+}
--- a/src/devices/Tca955x/Tca955x.cs
+++ b/src/devices/Tca955x/Tca955x.cs
@@ -31,13 +31,15 @@ namespace Iot.Device.Tca955x
 
         private object _interruptHandlerLock = new object();
 
-        // This task processes the i2c reading of the io expander in the background to
-        // avoid blocking the interrupt handler too long. While it is running, further
+        // This task processes the i2c reading of the io expander in a background task to
+        // avoid blocking the interrupt handler for too long. While it is running, further
         // incoming events for the Interrupt pin are ignored but if there are any, then we
         // will read the device a second time just in case we missed the most current state.
         // This is the best we can do given the limitations of the i2c bus speed and the fact that
         // the device can clear and reassert the INT pin regardless of whether we have read
-        // the expander state over i2c.
+        // the expander state over i2c. Bear in mind that the interrupt processing routine
+        // also synchronously calls out to any user event handlers that have been registered
+        // and they could take an arbitrary amount of time to complete.
         private Task? _interruptProcessingTask = null;
         // Set to true if an interrupt occurred while we were already processing one.
         // If so, we need to do another read when we finish the current one.
@@ -54,7 +56,7 @@ namespace Iot.Device.Tca955x
         /// Maximum number of addresses configurable via A0, A1, A2 pins.
         /// </summary>
         public const byte AddressRange = 7;
-        
+
         /// <summary>
         /// Represents the number of bits in a byte.
         /// </summary>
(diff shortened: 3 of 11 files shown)


src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
```diff
@@ -31,6 +31,7 @@ public class GpioController : IDisposable
     /// </summary>
     private readonly ConcurrentDictionary<int, PinValue?> _openPins;
     private readonly ConcurrentDictionary<int, GpioPin> _gpioPins;
+    private readonly bool _driverHasBlockAccess = false; // Cache the block access capability. Doing it this was trades one bool of memory for a little performance gain.
```
pgrawehr: At least for the case where this is true, you have no performance gain at all, in particular because you're doing the cast to `IGpioDriverBlockAccess` anyway. I'd rather store the casted interface and then do a `_driverBlockAccess != null` test.

src/System.Device.Gpio/System/Device/Gpio/IGpioDriverBlockAccess.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Device.Gpio
+{
+    /// <summary>
+    /// Interface to mark drivers that support block access to GPIO pins.
+    /// Block access in this case meaning the ability to read/write multiple pins in one call.
+    /// </summary>
+    public interface IGpioDriverBlockAccess
```
pgrawehr: I'm not convinced this is a good approach, in particular because you're adding logic to the GpioController which is very specific, currently even specific to this binding. 

I would rather move the implementation of `public void Read(Span<PinValuePair> pinValuePairs)` to `GpioDriver` base class (with the default implementation just doing the foreach loop)

src/devices/Tca955x/Tca955x.cs
```diff
@@ -64,10 +89,16 @@ protected Tca955x(I2cDevice device, int interrupt = -1, GpioController? gpioCont
...
+                var currentIntPinMode = _controller.GetPinMode(_interrupt);
+                // The INT pin is open drain active low so we need to ensure it is configured as an input,
+                // but it could be configured with pullup or use an external pullup.
+                if (currentIntPinMode != PinMode.Input && currentIntPinMode != PinMode.InputPullUp)
                 {
                     _controller.SetPinMode(interrupt, PinMode.Input);
                 }
+
+                // This is only be done once as there is only one INT pin interrupt for the entire ioexpander
+                _controller.RegisterCallbackForPinValueChangedEvent(_interrupt, PinEventTypes.Falling, InterruptHandler);
```
pgrawehr, line 107: `_controller` can still be null at this point (if _interrupt == -1), so this would crash.

src/devices/Tca955x/Tca955x.cs
```diff
@@ -373,36 +421,104 @@ protected override PinMode GetPinMode(int pinNumber)
...
+            // event handlers that the consumer of the library has signed up, we are likely
+            // to miss edges while we are doing that anyway. Dropping interrupts in this
+            // case is the best we can do and prevents flooding the consumer with events
+            // that could queue up in the INT gpio pin driver.
             lock (_interruptHandlerLock)
             {
-                if (_interruptPins.Count > 0)
+                if (_interruptProcessingTask == null)
+                {
+                    _interruptProcessingTask = ProcessInterruptInTask();
```
pgrawehr, line 444: The code is very complex and it's hard to imagine all possible execution paths. (Note for instance that the started task will just run to the lock initially, having to wait for this method to end, which requires at least 2 task switches)

I think this will be simpler if just the callback happens in a task, but not the readout of the new values. I don't think that would perform worse than this solution (if we aren't fast enough to read out the values after each interrupt, we're lost anyway).

src/devices/Tca955x/Tca955x.cs
```diff
@@ -511,7 +632,29 @@ void InternalHandler(object sender, PinValueChangedEventArgs pinValueChangedEven
 
         /// <inheritdoc/>
         protected override bool IsPinModeSupported(int pinNumber, PinMode mode) =>
-            (mode == PinMode.Input || mode == PinMode.Output || mode == PinMode.InputPullUp);
+            (mode == PinMode.Input || mode == PinMode.Output);
+
+        /// <inheritdoc/>
+        protected override void Dispose(bool disposing)
+        {
+            if (_shouldDispose)
```
pgrawehr, line 651: You need to do the `UnregisterCallbackForPinValueChangedEvent` first and then wait for any active Interrupt callback tasks. Otherwise the interrupt callback might still be called on a disposed device.

pgrawehr (changes requested):
Some suggestions and a few possible issues found.

src/System.Device.Gpio/System/Device/Gpio/GpioController.cs
felixcollins: Thanks, I'll change.  I would rather have two implementations of GpioController, one for block-gpio and one for non-block-gpio but the architecture prevents this.

src/System.Device.Gpio/System/Device/Gpio/IGpioDriverBlockAccess.cs
felixcollins: Gpio is at the hardware level, block-oriented, even reading memory-mapped gpio in a micro is done a byte at a time. The current, looping implementation of the Read Span overload is very inefficient, I think this deserves to be fixed where possible. Do you mean we should take the Read Span out of GpioController? That would be a breaking API change. How about I introduce a subclass of GpioController called "BlockGpioController"? This could only be constructed with a GpioDriver supporting IGpioDriverBlockAccess. Then library users can choose to use it if they care about efficiency.

src/devices/Tca955x/Tca955x.cs
felixcollins: `_controller = gpioController ?? new GpioController();`  This was in there before I started modifying the driver. Now that I look, I see that the parameterless constructor of GpioController does not return a working GpioController. I'm not sure what the point of it is. Should we just insist on a non-null GpioController if interrupt is not -1?

src/devices/Tca955x/Tca955x.cs
felixcollins: Thank you for looking through this. It is complex, but the problem is that the i2c reading is very slow relative to the rate that the ioexpander chip can generate interrupts. Any pin-toggle can cause a new interrupt even if the chip has not been read to service the last interrupt. I think this is bad behaviour of the chip but I can't change that. This is not an academic exercise, I'm watching the scope and trying to make the driver behave better when there is a noisy pin on the ioexpender. It has made a huge difference - the i2c now only gets 4 transactions (16 bit i/o expander) for a short burst of input toggles. Previously the interrupts would queue up, resulting in a whole heap of unnecessary i2c reads.  I'll move the mirroring of the pins and values out of the task (they will be added to the closure for the task lambda). That will remove the initial lock in the task.

src/System.Device.Gpio/System/Device/Gpio/IGpioDriverBlockAccess.cs
felixcollins: Sorry, I think I see what you mean now.  You mean just add a public Read Span overload to GpioDriver with a default looping implementation, then override it in the TCA955x.  That does seem better.

src/devices/Tca955x/Tca955x.cs
felixcollins: I made the construction requirements more explicit.

src/devices/Tca955x/Tca955x.cs
felixcollins: Agreed. Can you check my impl please?

felixcollins:
@pgrawehr oh dear... I've broken the building of several other GpioDriver subclasses. Shall I go through and add override to those other drivers? Or add another intermediate abstract class GpioBlockDriver?

src/System.Device.Gpio/System/Device/Gpio/IGpioDriverBlockAccess.cs
pgrawehr: Yes, that was the idea. The GpioController would then just forward the call. No breaking Api change.

Note that the methods in GpioDriver should be `internal protected virtual`, as only the controller should call them.

src/devices/Tca955x/Tca955x.cs
pgrawehr: `new GpioController()` returns a working copy on the Raspberry Pi or some flavors of linux only. On other systems, this in fact returns a dummy instance. At the moment, all bindings have this fallback behavior if gpioController is null. I'm not a fan of this either, but it's been there from the beginning and it would be a major breaking change to make that argument mandatory.

src/devices/Tca955x/Tca955x.cs
```diff
@@ -80,6 +80,12 @@ protected Tca955x(I2cDevice device, int interrupt = -1, GpioController? gpioCont
                 throw new ArgumentOutOfRangeException(nameof(device), $"Address should be in Range {DefaultI2cAddress} to {DefaultI2cAddress + AddressRange} inclusive");
             }
 
+            if ((_interrupt != -1 && gpioController is null) ||
+                (_interrupt == -1 && (gpioController is not null)))
+            {
+                throw new ArgumentException("gpioController and interrupt must be set together.");
+            }
```
pgrawehr: I'm ok with this solution, but it's a bit of a concept break (see my comment elsewhere), so I need to ask the team about it.

pgrawehr (approved):
Looks good to me now, thanks a lot for this contribution. Since it includes a core class change, I will have to discuss this in a team meeting. Might not work out this week, though.

src/devices/Tca955x/Tca955x.cs
felixcollins: Okay, so It should just make a default GpioController and let it fail elsewhere if it is not actually able to get events from that pin. I'll change it back. I agree we should endeavour to not break the existing API. I'm wondering if I should actually drop this PR and fork the TCA955x driver. There are a couple of other i2c gpio drivers that are very similar already. Perhaps we should make a more generic one and people can migrate to it if they want the better behaviour and bug fixes? The new one could demand a working GpioController in the ctor right from the start. I would probably get rid of the locks too.

src/devices/Tca955x/Tca955x.cs
felixcollins: I commented on the other comment location. Please consider my proposal to put in a whole new I2CExpanderGpioDriver. We can of course accept this PR and then go for another to introduce the new driver.

src/devices/Tca955x/Tca955x.cs
pgrawehr: No, I think it's fine like it is now and we should merge it. You can still propose a new set of changes later, but this is already an improvement. 
And yes, there are a few very similar devices (namely the MCP23xxx series). I haven't checked whether they could be unified, though.

src/devices/Tca955x/Tca955x.cs
pgrawehr: Looks correct.

PR #2425, opened 2025-09-16
TCA955x - making address check optional so this driver can be used with other c…

felixcollins:
The Tca955x constructor validates that the address of the passed in i2c device is in the correct range for the Tca955x family. This driver is useful for other register compatible devices that have different addressing schemes. This PR optionally turns off this check so that advanced users can deploy it for other devices.


raffaeler (approved):

joperezr (approved):
Thanks for the change @felixcollins. It is of course a binary breaking change but not a source breaking change. I think due to the value of it we should take it. Thanks!

felixcollins:
@joperezr Is there a policy for this? It seems pretty unlikely that someone would upgrade to a newer version of the library without testing their existing application. All they need to do is recompile against the new lib. I guess I could have added a new constructor but that seems messy. The whole driver probably needs a rethink really (the lock is scary to me). It would be nice to make a UniversalGpioExpander driver that supports all the similar chips.

